PR #6, opened 2026-03-09
React profiler tools

stachbial:
⚒️🎯 ❌🧢  Introducing React Profiler Tools!

What it aims to do?
- starts react profiling session via metro
- listens to commit tree events
- extracts the **hot commits** which make the app run slow [same criteria as commits changing colors from green to yellowish and red in the profiler UI + annotation if the commit exceeds 16ms making the app perform below 60FPS]
- links the heavy renders with the cause (resolves the render reason from "parent" to actual component)
- gets the source tree of components of interest to refer to them in the summary
- outputs the findings in agent-friendly markdown

- the agent then presents the findings with its own conclusions and links to specific components in user's codebase

I paid a lot of tokens creating and testing it out...

  <details>
  <summary>✨EXAMPLE 👌😫😮‍💨</summary>

**App:** Expensify
**Input prompt:** 

"profile this scenario with radon-lite mcp : go to inbox tab -> concierge  -> scroll to the top -> then back to the 
 bottom -> send one message -> go back to home. i think that the tabs are switching slowly and have some concerns about chat  rendering performance as well. do not use keyboard unless the text input is focused then send it with the green button "
 
**Agents output (Sonnet 4.6): **
⏺ Here's the full analysis:                             
                                                                                                                              
  ---                                                                                                                         
  Performance Report                                             
                                                                                                                              
  🔴 Critical: Chat Initial Mount — 387ms → 197ms → 173ms burst (t≈10.7–12.8s)
                                                                                                                              
  Opening Concierge triggered 5 consecutive hot commits as ReportActionItems batch-mounted progressively (22 → 32 → 42 → 48   
  items). Each batch re-rendered the entire navigator tree on top. This is the main cause of the perceived tab/chat           
  sluggishness.                                                                                                               
                                                                 
  Root: AppNavigator re-renders from context changed on every single commit — 51 renders at a fixed 64.99ms cost each (3,314ms
   total). Something upstream in the context tree is changing frequently, and AppNavigator isn't bailing out.
                                                                                                                              
  ---                                                                                                                         
  🔴 Critical: Send Message — 243ms (t≈108.4s)                   
                                                                                                                              
  Sending one message caused PureReportActionItem ×27 + ReportActionItem ×27 to re-render — all existing messages were
  re-rendered. The trigger was props: policies changing. Every item gets policies as a prop that is likely a new object       
  reference on each Onyx update, forcing the full list to re-render instead of just appending the new message.
                                                                                                                              
  ---                                                            
  🔴 High: AppNavigator — 64.99ms on every commit                
                                                                                                                              
  Same self-cost every time = it's never bailing out. React Compiler has optimized it, so the problem is at the callsite:
  something passed to AppNavigator creates a new reference on every context update (inline object, selector without           
  deep-equal, etc.).                                             
                                                                                                                              
  ---                                                            
  🟡 ReportActionAvatars — singleAvatarContainerStyle prop instability
                                                                                                                              
  ReportActionAvatars rendered 1,174 times with props: singleAvatarContainerStyle as the trigger. This is an inline style
  object being created fresh on every parent render. Memoizing it at the call site would eliminate these re-renders entirely. 
                                                                 
  ---                                                                                                                         
  🟡 App component — compiler missed it (⚠️ )                     
                                                                 
  App re-rendered 102 times (1,206ms total) due to parent re-renders, but the React Compiler didn't memoize it. Run npx 
  react-compiler-healthcheck on src/App.tsx to find what's blocking compilation.                                              
  
  ---                                                                                                                         
  🟡 TextCommentFragment — 1,146 renders (1,818ms total)         
                                                                                                                              
  Compiler-optimized but still re-rendering from context. Likely subscribing to a broad context that changes on every
  scroll/commit event. Should subscribe to a narrower selector.                                                               
                                                                 
  ---                                                                                                                         
  Top Priorities                                                 
  image
                                                                 
  Note: all timings are dev mode (~3× slower than production). Divide by 3 for rough production estimate. Full report saved to
   rn-devtools-debug/profiler-report.md. 

  </details>


pFornagiel (approved):
Just make sure that `MEMORY.md` is actually needed and whether we should include it in the project for us. Other than that it looks alright, if needed we will iterate as we go further in the development.

PR #15, opened 2026-03-10
feat: Port network tools from `Radon IDE` to `Argent`

latekvo:
This PR ports the following tools from Radon IDE:

- `view-network-logs`
- `view-network-request-details`

Also new blueprint is added:  `network-inspector.ts`, i previously modified the js debugger one but i since reverted that approach.

### Tests succeeded:

image

**NOTE:** Majority of all LoC are pulled verbatim from Radon IDE. The only new parts are the monkeypatching logic.

I'd actually really want your opinion about the `fetch` and `XMLHttpRequest` call monkeypatching approach. It sounds bad, but it does work.


latekvo:
Merged because it does not modify any existing code. New blueprint & new tools. Tested both extensively. 

<sup>(the last sentence is a double-entrende, "both" refers to "both the blueprint and the new tools" as well as "both the new tools").</sup>

PR #1, opened 2026-02-27
feat: add the registry

pFornagiel:
This pull request introduces the new `registry`, which implements a service manager and tool registry for the project. The registry provides API for managing service instances nad invoking tools. The changes include the full implementation of the registry, error handling, typed events, utility functions.

Registry implementation:

* Added the complete `Registry` class in `packages/registry/src/registry.ts`, supporting service resolution, tool registration and invocation, lifecycle management, and dependency tracking.
* Defined core types for services, tools, events, and URNs in `packages/registry/src/types.ts`, enabling interaction with the registry and its components.

Error handling and utilities:

* Provided utility functions for URN parsing and Zod-to-JSON Schema conversion in `packages/registry/src/urn.ts` and `packages/registry/src/zod-to-json-schema.ts`, respectively. [[1]](diffhunk://#diff-058bf7956f16fad9cd3e3aa9ad701c288a4adc58147a0317c4a3752cfb26a644R1-R14) [[2]](diffhunk://#diff-c991baed0abf7412eb13a4430b3fb99a7317789b1744b510b9c0de513e7c8394R1-R38)
* Included a registry logger in `packages/registry/src/logger.ts` for observing lifecycle events in the server.

Documentation and configuration:
* Updated `README.md` with instructions for running the tools server and registry, and integrating with the frontend UI.

PR #21, opened 2026-03-12
Ios instruments profiler

stachbial:
- added ios instruments profiling tools
- renamed previous profiling tools to "react-profiler" 
- enhanced interoperability/ parallelism of react-profiler and ios-instruments (u can still use them as standalone if u want tho)
- added profiler-query tools: to interact and drill down further into specific data from the profiler output after the initial analysis/summary
- enabled loading old profiler data if present - useful for comparison if in new context

Todo in later feature: 
- integrate modern react profiler tools like "performance" tab available from 0.83

**Check out this demo** 😫👌❌🧢⬇️
https://drive.google.com/file/d/1aI5ijV5zFI6Usek7S2-06Ihml-DXGyNE/view?usp=sharing

<details>
<Summary>Example ✨</summary>

**App:** Expensify
**Model:** Claude Sonnet 4.6

**Prompt:** "analyse this app scenario:  inbox->  concierge chat (tap y:0.21, the screen wont probably appear on first ss and thats what we are investigating) -> swipe down 1 time -> swipe back to inbox Using argent mcp profiling tools (react + iOS). use only screenshots to navigate the app or my  hints. i feel like tab switching  and chat opening is sluggish and i  want to investigate the case further using queries from argent mcp "

**Response (after doing the scenario via tools): **

Here's the full analysis across both profilers:

---

### TL;DR — Root causes of sluggishness

The chat open (~8s delay observed) is driven by **incremental mount batching on the JS thread**, not a main thread hang. No iOS UI hangs were recorded. The JS thread was simply overwhelmed with work.

---

### React Profiler — Hot commits breakdown

**21 of 54 commits were ≥16ms** across the 67s session. The chat open phase (t=0–5.4s) alone produced 9 heavy commits:

| Commit | Duration | Time | What happened |
|--------|----------|------|----------------|
| #0 | **1536ms** 🔴 | t=0.0s | Initial mount — 328 components, full chat tree |
| #1 | **236ms** 🔴 | t=0.8s | Re-render cascade (context change) |
| #5 | **79ms** 🔴 | t=1.4s | Re-render cascade |
| #12 | **241ms** 🔴 | t=2.2s | 10 more `ReportActionItem`s loaded (32 total) |
| #15 | **349ms** 🔴 | t=3.1s | 10 more items (42 total) |
| #16 | **112ms** 🔴 | t=3.4s | `policies` prop re-render |
| #22 | **243ms** 🔴 | t=4.3s | 10 more items (52 total) |
| #24 | **108ms** 🔴 | t=4.8s | 4 more items + `TaskPreview` mount |
| #26 | **81ms** 🔴 | t=5.2s | 2 more items, `TaskPreview` props change |

**Pattern**: Messages are loaded in batches from Onyx, each batch triggering a full tree re-render. The chat didn't appear until all these commits settled (~8s total).

After the "Tap Concierge" annotation (+15.2s into session), the next React commits don't appear until +33s — **an 18 second gap** with no React work, suggesting the navigation was pending on data availability.

---

### Top component offenders

| Component | Renders | Total cost | Problem |
|-----------|---------|------------|---------|
| `SceneView` ⚠️ | 412 | **4091ms** | Not compiler-optimized — re-renders on every context change |
| `AppNavigator` | 35 | **2730ms** | 77.99ms flat per render — some context is always dirty |
| `BaseHTMLEngineProvider` | 35 | **1154ms** | 32.96ms flat — re-initializes on every commit |
| `ReportActionItem` ×42–60 | 1000s | ~1000ms+ | `props: policies` causes re-renders — unstable selector |
| `TRenderEngineProvider` ⚠️ | 35 | **350ms** | Not compiler-optimized |

**Critical**: `ReportActionItem` / `PureReportActionItem` consistently re-render due to `props: policies`. The policies object is likely recreated on every Onyx subscription tick instead of being memoized at the selector level.

---

### iOS Instruments — CPU & Memory

**JS thread: 65.2% of all CPU** (13234ms sampled). Two call chains dominate:
1. Normal Hermes JS execution via `RuntimeScheduler_Modern::runEventLoop` — expected
2. **CDP evaluation overhead** (458+ samples of `CDPAgentImpl::handleCommand` → JSON serialization) — this is dev-mode debugger protocol cost, not present in production

**Main thread: only 3.24%** via `folly::json::escapeStringImpl` (inspector ↔ Metro serialization). The UI thread was **not blocked** — this confirms the sluggishness is purely JS-thread latency, not a native render stall.

**Memory leaks — notable items:**
- `Malloc 16 Bytes` ×210,990 = **3.2MB** — large number of tiny allocations not freed (likely JS/bridge churn)
- `RCTMarkdownStyle` ×3 + `RCTMarkdownUtils` ×4 + `MarkdownParser` ×4 + `MarkdownTextFieldObserver` ×4 + `MarkdownFormatter` ×4 — **clear retain cycle in the Markdown native module** (`react-native-live-markdown`)
- `UIDeviceRGBColor` ×36 — color objects leaking, likely from theme-related native calls
- `hermes::vm::JSTypedArrayBase::createBuffer` — a Hermes TypedArray buffer not freed

---

### Root causes ranked by impact

1. **Incremental Onyx batch mounting** — each batch of loaded messages triggers a full reconcile of `SceneView` → all 400+ `View`/`Pressable`/`Hoverable` children. The chat isn't "ready" until all batches land.

2. **`SceneView` not compiler-optimized** — re-renders 412 times (4 seconds+ of render time). React Compiler couldn't optimize it — likely has a conditional hook or prop mutation somewhere.

3. **`policies` prop instability** — `ReportActionItem` re-renders are consistently blamed on `props: policies`. The selector returning the policies object creates a new reference every time, causing unnecessary renders on all visible chat items.

4. **`AppNavigator` context** — triggers a 78ms re-render on every single commit (35 times). Whatever context it consumes is changing constantly.

5. **Markdown native module retain cycle** — `MarkdownParser`/`RCTMarkdownStyle`/`MarkdownFormatter` all leaking together = retain cycle in `react-native-live-markdown` native side.

---

### Suggested next drill-downs

```
profiler-commit-query mode=by_index commit_index=0      # Full breakdown of the 1536ms mount
profiler-cpu-query mode=component_cpu component=SceneView  # What CPU work happens during SceneView renders
profiler-stack-query mode=leak_stacks object_type=RCTMarkdownStyle  # Markdown retain cycle stack
```

> **Dev mode note**: All timing values are ~3× slower than production. Divide by 3 for real-world estimates (e.g. the 1536ms initial mount → ~512ms in prod).
</details>


packages/tool-server/src/utils/ios-profiler/pipeline/xml-parser.ts
```diff
@@ -0,0 +1,379 @@
...
+// ---------------------------------------------------------------------------
+
+function decodeXml(str: string): string {
+  return str
+    .replace(/&lt;/g, "<")
+    .replace(/&gt;/g, ">")
+    .replace(/&amp;/g, "&")
+    .replace(/&quot;/g, '"')
+    .replace(/&apos;/g, "'");
+}
```
latekvo, line 16: The `&` has to be last, because it can be a part of other XML expressions.

For example with string: `&amp;lt;tag&amp;gt;`
The expected output is: `&lt;tag&gt;`
But with this function the output is: `<tag>`

```suggestion
function decodeXml(str: string): string {
  return str
    .replace(/&lt;/g, "<")
    .replace(/&gt;/g, ">")
    .replace(/&quot;/g, '"')
    .replace(/&apos;/g, "'")
    .replace(/&amp;/g, "&");
}
```

packages/tool-server/src/tools/profiler/ios-instruments/ios-instruments-start.ts
```diff
@@ -0,0 +1,209 @@
...
+          reject(new Error(`Failed to attach to iOS process: ${errorOutput}`));
+        }
+      });
+
+      xctraceProcess.on("close", (code: number | null) => {
+        console.log("✨✨✨ xctrace process CLOSED with code:", code);
+      });
+
+      xctraceProcess.on("error", (err: Error) => {
+        console.log("✨✨✨ xctrace process ERROR:", err.message);
```
latekvo: Do we want `console.log`s here? Especially `xctraceProcess.on("close", ...` and `xctraceProcess.on("error", ...` look like something you only used for debugging.

packages/tool-server/src/tools/profiler/ios-profiler/ios-profiler-start.ts
```diff
@@ -0,0 +1,209 @@
...
+      .slice(0, 15);
+    const outputFile = path.join(
+      debugDir,
+      `ios-instruments-${timestamp}.trace`,
+    );
+
+    api.appProcess = appProcess;
+    api.traceFile = outputFile;
+
+    return new Promise((resolve, reject) => {
```
latekvo, line 136: I think this promise will hang infinitely if `xctrace` never finishes. Do you think a timeout would make sense here?

packages/tool-server/src/tools/profiler/react/react-profiler-analyze.ts
```diff
@@ -0,0 +1,226 @@
...
+      sessionFiles: {
+        sessionId: sessionPaths.sessionId,
+        cpuProfile: sessionPaths.cpuProfilePath,
+        commits: sessionPaths.commitsPath,
+      },
+    };
+
+    // Warn only when hook was genuinely not installed (null).
+    // hotCommitIndices === [] means hook was installed but 0 commits — that's valid, not an error.
+    if (commitTree === null || hotCommitIndices === null) {
```
latekvo: `commitTree === null` is never `null`:

```suggestion
    if (hotCommitIndices === null) {
```

but you could've meant:


```suggestion
    if (commitTree.commits.length === 0 || hotCommitIndices === null) {
```

packages/tool-server/src/tools/profiler/combined/profiler-combined-report.ts
```diff
@@ -0,0 +1,334 @@
...
+          const { hang, overlappingCommits } = corr;
+          const topCommit = overlappingCommits[0]!;
+
+          lines.push(
+            `### ${hang.hangType} at ${hang.startTimeFormatted} (${hang.durationMs}ms) ↔ Commit #${topCommit.commit.commitIndex} (${topCommit.commit.totalRenderMs}ms)`,
+          );
+          lines.push("");
+
+          // Explain the ratio
+          const ratio = topCommit.commit.totalRenderMs / hang.durationMs;
```
latekvo: I couldn't find a place in code that explicitly ensures `hang.durationMs > 0`.

```suggestion
          const ratio = hang.durationMs > 0 ? topCommit.commit.totalRenderMs / hang.durationMs : 0;
```

packages/tool-server/src/tools/profiler/ios-profiler/ios-profiler-stop.ts
```diff
@@ -0,0 +1,68 @@
...
+    const api = services.session as IosInstrumentsSessionApi;
+
+    if (!api.profilingActive || !api.xctracePid || !api.traceFile) {
+      throw new Error(
+        "No active iOS profiling session found. Call ios-instruments-start first.",
+      );
+    }
+
+    const pidToKill = api.xctracePid;
+    process.kill(pidToKill, "SIGINT");
```
latekvo, line 45: It's expected that `kill` will throw if `pidToKill` does not exist. I think it would be good to handle this case.

packages/tool-server/src/tools/profiler/react/react-profiler-start.ts
latekvo: This is somewhat out of scope of this PR (is it?), but doesn't the code contradict this comment? Comment says to inject the hook before calling `Profiler.start`, and in the code it's the other way around.

packages/tool-server/src/utils/ios-profiler/config.ts
```diff
@@ -0,0 +1,18 @@
...
+  "Yoga",
+  "Hermes",
+  "hermes::",
+  "JSI",
+  "main",
+  "facebook::",
+  "React",
+  "jsi::",
+  "HermesRuntime",
+];
```
latekvo, line 11: Isn't this ***wayyy*** too broad? Detecting strings that contain `main` as RN framework.

packages/tool-server/src/utils/ios-instruments/render.ts
```diff
@@ -0,0 +1,339 @@
...
+  const bottlenecksTotal = payload.bottlenecks.length;
+
+  const report =
+    bottlenecksTotal === 0
+      ? renderAllClear(payload, input.exportErrors)
+      : renderFullReport(payload, input.exportErrors);
+
+  const reportFile = traceFile ? deriveReportPath(traceFile) : null;
+  if (reportFile) {
+    await writeReport(reportFile, report);
```
latekvo: `writeReport` returns `false` when it fails. This is not handled here.

packages/tool-server/src/utils/ios-profiler/pipeline/xml-parser.ts
```diff
@@ -0,0 +1,379 @@
...
+  const weightDefRe = /<weight\s+id="(\d+)"[^>]*>(\d+)<\/weight>/g;
+  let wm;
+  while ((wm = weightDefRe.exec(xml)) !== null) {
+    weightRegistry.set(wm[1], parseInt(wm[2], 10));
+  }
+
+  // Second pass: resolve refs
+  let sampleIdx = 0;
+  for (const row of rows) {
+    if (sampleIdx >= samples.length) break;
```
latekvo, line 135: The first pass skips rows with `sample-time ref`
This second pass does not while still relying on the same sample indexes. I'm pretty sure this will break.

packages/tool-server/src/utils/react-profiler/pipeline/00-cpu-correlate.ts
```diff
@@ -0,0 +1,222 @@
...
+    if (Math.abs(diff) > 1000) {
+      clockOffsetMs = diff;
+    }
+  }
+
+  // Build absolute timestamps for each sample
+  const timestampsMs = new Float64Array(samples.length);
+  let accumulatedUs = startTime;
+  for (let i = 0; i < samples.length; i++) {
+    if (i > 0) accumulatedUs += timeDeltas[i]!;
```
latekvo: `[i]!` is sketchy because `i` is bound to len of `samples` not `timeDeltas`.


```suggestion
    accumulatedUs += timeDeltas[i] ?? 0;
```

latekvo (changes requested):
Just a few inlines, great job overall 👍

pFornagiel:
The `react-native-profiler` skill should not be added to the top-level `.claude` folder, as it is mean for our internal tooling. Same goes for `agents` directory, but that could be an artefact we've overlooked earlier,

packages/tool-server/src/tools/profiler/ios-profiler/ios-profiler-stop.ts
```diff
@@ -0,0 +1,68 @@
...
+    await new Promise<void>((resolve) => {
+      const checkInterval = setInterval(() => {
+        try {
+          process.kill(pidToKill, 0);
+        } catch {
+          clearInterval(checkInterval);
+          resolve();
+        }
+      }, 1000);
+    });
```
latekvo, line 57: Use can use `withResolvers` to remove the excessive nesting and make this slightly more readable.


```suggestion
    const { xctracePromise, resolveXctrace } = Promise.withResolvers<void>();
    
    const checkInterval = setInterval(() => {
      try {
        process.kill(pidToKill, 0);
      } catch {
        clearInterval(checkInterval);
        resolveXctrace();
      }
    }, 1000);
    
    await xctracePromise;
```

packages/tool-server/src/utils/ios-profiler/pipeline/xml-parser.ts
stachbial: I disagree, we check for the full tag annotation, so there is no possibility to double decode as you are suggesting. Wrote a test case to confirm 😁

packages/tool-server/src/tools/profiler/combined/profiler-combined-report.ts
stachbial: The corelation between a hang and a react commit would not happen if there was no hang, and a hang can never be 0 therfore `hang.durationMs > 0 ? topCommit.commit.totalRenderMs / hang.durationMs : 0;` would rather be a dead code

stachbial: I have added the 0 check for security tho

latekvo: Yeah i kind of figured the existence of `hang` implies `durationMs` is >0, but i couldn't find any checks that ensure it earier on in the code, so it made me wonder if there is a potential for some edge cases. Well it's resolved either way, so thanks.

packages/tool-server/src/utils/ios-profiler/pipeline/xml-parser.ts
latekvo: I could be misunderstanding this, but isn't picrel an example of the current function in fact being broken?  

image

packages/tool-server/src/tools/profiler/ios-profiler/ios-profiler-stop.ts
stachbial: Promise.withResolvers is not available in our current ts ES2022 lib (>=ES2024)

packages/tool-server/src/utils/ios-profiler/pipeline/xml-parser.ts
stachbial: oh, u mean when we run the function multiple times on the same string. I guess I can patch that, but it's not a real use case in our scenario. Nevertheless I agree - it should be deterministic

packages/tool-server/src/tools/profiler/ios-profiler/ios-profiler-stop.ts
latekvo: We do have `"ESNext"` in our included TS libs though, so it should work. Does it not?

```
  "compilerOptions": {
    ...
    "jsx": "react-jsx",
    "lib": ["es6", "dom", "ESNext"],
    "module": "es2022",
    ...
  },
```

latekvo: Nevermind i was looking at the wrong repository haha 😅

packages/tool-server/src/utils/ios-profiler/pipeline/xml-parser.ts
latekvo: >oh, u mean when we run the function multiple times on the same string

No, the picture i sent shows the function being ran exactly once and still producing the wrong output.
But yeah, it's patched now so thx

packages/skills/skills/ios-instruments/SKILL.md
```diff
@@ -0,0 +1,138 @@
+---
+name: ios-instruments
```
latekvo: nitpick, but maybe it'd make sense to rename this, and all references to just `ios-profiling`? 
The name `ios-instruments` is somewhat vague.

packages/skills/skills/ios-instruments/SKILL.md
```diff
@@ -0,0 +1,138 @@
...
+
+## 2. Tool Overview
+
+| Tool                      | Purpose                                                                                     |
+| ------------------------- | ------------------------------------------------------------------------------------------- |
+| `ios-instruments-start`   | Start xctrace recording on a booted simulator or device. Captures CPU, hangs, and leaks.    |
+| `ios-instruments-stop`    | Stop xctrace, export trace data to XML files (timestamped, persist on disk).                |
+| `ios-instruments-analyze` | Parse exported XML and return structured bottleneck payload (CPU hotspots, UI hangs, leaks). |
+| `profiler-stack-query`    | Drill into parsed data: hang stacks, function callers, thread breakdown, leak details.       |
+| `profiler-load`           | List and reload previous trace sessions from disk for re-investigation.                      |
```
latekvo: Let's keep using `ios-*` prefix if these tools are related. Or if they are not related, maybe let's move them to another `SKILL.md` file?

packages/skills/skills/ios-instruments/SKILL.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Quick Reference
+
+| Action                          | Tool                      |
+| ------------------------------- | ------------------------- |
+| Start iOS Instruments recording | `ios-instruments-start`   |
+| Stop iOS Instruments            | `ios-instruments-stop`    |
+| Analyze iOS Instruments trace   | `ios-instruments-analyze` |
+| Drill into hangs/CPU/leaks      | `profiler-stack-query`    |
+| Reload previous trace session   | `profiler-load`           |
```
latekvo: All other occurences of `profiler-*` were renamed to `react-profiler-*`. On the other hand all the references to iOS profiling were named `ios-instruments-*`.

Are these `profiler-stack-query` & `profiler-load` tools meant to be grouped with either of these existing toolsets, or do they belong to a separate `profiler-*` group?

stachbial: the querying tools are used for both types of profiling outputs at this moment

PR #25, opened 2026-03-17
feat: Network inspector tool

latekvo:
This tool has been previously reverted in #18 due to a crash occuring when both the network inspector and js debugger were ran sequentially in a single tool-server session.

This PR reimplements `network-inspector` while resolving the aforementioned bug.


latekvo:
Tested, works well and without any interference between the debugger and this tool

PR #33, opened 2026-03-19
feat: compact ios profiler report + ensure wallAnchor for combined report

stachbial:
Stability improvements for combined report not to exceed context character limit.
The full data still remains accessible via profiler-query tools.

### DIFF
--- a/packages/tool-server/src/blueprints/react-profiler-session.ts
+++ b/packages/tool-server/src/blueprints/react-profiler-session.ts
@@ -184,10 +184,8 @@ export const reactProfilerSessionBlueprint: ServiceBlueprint<
     return {
       api: state,
       dispose: async () => {
-        // No CDP calls here — Profiler.stop is issued explicitly in react-profiler-stop,
-        // and the new session's factory re-enables the domain.
-        // Keeping dispose free of I/O ensures TERMINATING→IDLE completes as a microtask,
-        // before the next MCP request (macrotask) can trigger a resolve.
+        // Profiler.stop is called explicitly in react-profiler-stop before disposal.
+        await cdp.send("Profiler.disable").catch(ignore);
       },
       events,
     };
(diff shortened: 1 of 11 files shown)


packages/tool-server/src/blueprints/react-profiler-session.ts
```diff
@@ -173,20 +177,17 @@ export const reactProfilerSessionBlueprint: ServiceBlueprint<
...
       dispose: async () => {
-        if (state.profilingActive) {
-          await cdp.send("Profiler.stop").catch(ignore);
-          state.profilingActive = false;
-        }
-        await cdp.send("Profiler.disable").catch(ignore);
+        // No CDP calls here — Profiler.stop is issued explicitly in react-profiler-stop,
+        // and the new session's factory re-enables the domain.
+        // Keeping dispose free of I/O ensures TERMINATING→IDLE completes as a microtask,
+        // before the next MCP request (macrotask) can trigger a resolve.
```
latekvo: Does this comment say that automatic profiler disposal is not necessary because `react-profiler-stop` exists?
Why are we even removing this `dispose` body, will keeping it break anything?

packages/tool-server/src/tools/profiler/react/react-profiler-stop.ts
```diff
@@ -282,8 +282,9 @@ Call react-profiler-start first, then exercise the app, then call this.`,
         totalReactCommits: api.totalReactCommits,
       };
 
-      cacheProfilerPaths(api.port, sessionPaths);
-      api.sessionPaths = sessionPaths;
+    cacheProfilerPaths(api.port, sessionPaths);
+    api.sessionPaths = sessionPaths;
+    api.disposeSession();
```
latekvo: use a formatter pls

latekvo (approved):
Nice, just 2 comments then :shipit:

packages/tool-server/src/blueprints/react-profiler-session.ts
latekvo: Btw:

>// Keeping dispose free of I/O ensures TERMINATING→IDLE completes as a microtask,
>// before the next MCP request (macrotask) can trigger a resolve.

This sounds super flaky and obscure, also I'm not sure how it is relevant to the dispose function body being no-oped

packages/tool-server/src/tools/profiler/react/react-profiler-stop.ts
latekvo: bruh
image

PR #38, opened 2026-03-23
feat: enable physical device for iOS profiling

stachbial:
- renamed `list-simulators` to `list-devices` which behaves as previously with optional arg that triggers `xcrun devicectl list ....` run 
- links it to ios-profiler-start tool


latekvo:
I would much prefer a single tool that lists both devices and simulators, then CLEARLY labels which device is a simulator and which is a physical device. Then the agent should have it CLEARLY explained what are the limitations & reasons for using a simulator over a device and vice versa.

Unless there are arguments against this approach, my arguments are:
- We have to clearly explain the differences between the approaches either way, we don't want the agent to confuse these options
- Having separate tools still is more confusing than having one no matter how clearly you explain the difference with seemingly no quantifiable reason for having these tools split
- More tools take up more context, having a single one takes up less context.

stachbial:
> I would much prefer a single tool that lists both devices and simulators, then CLEARLY labels which device is a simulator and which is a physical device. Then the agent should have it CLEARLY explained what are the limitations & reasons for using a simulator over a device and vice versa.
> 
> Unless there are arguments against this approach, my arguments are:
> 
> * We have to clearly explain the differences between the approaches either way, we don't want the agent to confuse these options
> * Having separate tools still is more confusing than having one no matter how clearly you explain the difference with seemingly no quantifiable reason for having these tools split
> * More tools take up more context, having a single one takes up less context.

I hear you, that was my initial thought actually. Then I thought: due to the fact that most of Argent's features are heavily bound to simserver, querying the physical devices all the time would seem redundant and steering it with just a flag - harder to grasp for the agent. The standalone tool is rather easier to distinguish in such use case. 
Regarding the simulator vs physical device differences explanation, I am not very found of that approach - I think we should always default to simulator and talk about/work with real devices if the user explicitly says so in the prompt (current implementation).

stachbial:
> > I would much prefer a single tool that lists both devices and simulators, then CLEARLY labels which device is a simulator and which is a physical device. Then the agent should have it CLEARLY explained what are the limitations & reasons for using a simulator over a device and vice versa.
> > Unless there are arguments against this approach, my arguments are:
> > 
> > * We have to clearly explain the differences between the approaches either way, we don't want the agent to confuse these options
> > * Having separate tools still is more confusing than having one no matter how clearly you explain the difference with seemingly no quantifiable reason for having these tools split
> > * More tools take up more context, having a single one takes up less context.
> 
> I hear you, that was my initial thought actually. Then I thought: due to the fact that most of Argent's features are heavily bound to simserver, querying the physical devices all the time would seem redundant and steering it with just a flag - harder to grasp for the agent. The standalone tool is rather easier to distinguish in such use case. Regarding the simulator vs physical device differences explanation, I am not very found of that approach - I think we should always default to simulator and talk about/work with real devices if the user explicitly says so in the prompt (current implementation).

Eventually, a unified `list-devices` tool was made with physical device fetch optionally executed based on boolean arg. cc: @latekvo @pFornagiel

docs/argent_guide.md
```diff
@@ -0,0 +1,31 @@
+## 1. Install in the target project
+
+```bash
+cd /path/to/your-project
+npm install /path/to/software-mansion-argent-0.3.1.tgz
+```
+
+Use `--force` flag if needed.
```
latekvo: Are there any known cases where `npm install --force` would be necessary?

docs/cli-usage.md
```diff
@@ -0,0 +1,268 @@
...
+@software-mansion:registry=https://npm.pkg.github.com
+//npm.pkg.github.com/:_authToken=<GITHUB_PAT>
+```
+
+Replace `<GITHUB_PAT>` with a GitHub Personal Access Token that has the
+`read:packages` scope. Alternatively, use `npm login`:
+
+```bash
+npm login --registry=https://npm.pkg.github.com --scope=@software-mansion
+```
```
latekvo: This method did not work for me. Does it work for you?

docs/distribution-guide.md
```diff
@@ -0,0 +1,305 @@
...
+4. Re-runs the full configure step (MCP entries, permissions, file copies) regardless of
+   whether a version upgrade happened, ensuring workspace files are always fresh.
+
+### What `remove` does
+
+1. Removes the `argent` entry from all three MCP config files.
+2. Removes the `mcp__argent` permission from `.claude/settings.json`.
+3. Runs `npm uninstall @software-mansion/argent`.
+4. With `--prune`: deletes `.claude/skills`, `.claude/agents`, `.claude/rules`, and
+   `.cursor/rules`.
```
latekvo: Other places mention `argent uninstall` but not `argent remove`. This file mentions only `argent remove` but not `argent uninstall`.

Am I missing something or is this intentional?

docs/npm-migration-guide.md
```diff
@@ -0,0 +1,275 @@
...
+
+---
+
+### 7. Org setup on npmjs.com (prerequisite)
+
+Before the first publish, the `@software-mansion` scope must exist on npmjs:
+
+1. Create an npm organization at
+   [npmjs.com/org/create](https://www.npmjs.com/org/create) named
+   `software-mansion` (must match the `@software-mansion` scope).
```
latekvo: Come on...

docs/npm-migration-guide.md
```diff
@@ -0,0 +1,275 @@
...
+- **`packages/mcp/src/cli/update.ts`** — already fixed to not pass
+  `--registry` in install commands.
+- **`packages/mcp/src/cli/uninstall.ts`** — no registry references.
+- **`packages/mcp/src/cli/utils.ts` (other than `getLatestVersion`)** —
+  `globalInstallCommand` and `globalUninstallCommand` have no registry flags.
+- **`packages/mcp/src/cli/mcp-configs.ts`** — editor config adapters are
+  registry-agnostic.
+- **MCP server entry** — uses `argent-mcp` binary name, unrelated to registry.
+- **Skills, rules, agents** — bundled assets, no registry dependency.
+- **`bin/` directory** — native binary, no registry dependency.
```
latekvo: This is probably unnecessary. Also I'm certain this will become stale very soon and no one will be updating this list

docs/npm-migration-guide.md
```diff
@@ -0,0 +1,275 @@
...
+
+---
+
+## Migration order
+
+1. Create the npm org / claim the scope on npmjs.com.
```
latekvo: Same here

packages/mcp/scripts/bundle-tools.cjs
```diff
@@ -105,3 +111,16 @@ if (fs.existsSync(RULES_SRC)) {
...
+if (fs.existsSync(AGENTS_SRC)) {
+  fs.cpSync(AGENTS_SRC, AGENTS_DEST, { recursive: true });
+  const count = fs
+    .readdirSync(AGENTS_SRC, { withFileTypes: true })
+    .filter((e) => e.isFile() && e.name.endsWith(".md")).length;
+  console.log(
+    `✓ Copied ${count} agent(s) → ${path.relative(process.cwd(), AGENTS_DEST)}`,
+  );
+} else {
+  console.warn(`⚠ Agents source not found at ${AGENTS_SRC} — skipping copy`);
```
latekvo: I'm pretty sure this should be a fatal exception and not just a warn

latekvo (commented):
Nice. Probably also adjust the title to reflect what's being done in this PR.

latekvo:
I'm so lost right now. What did I just review?  The comments are here but the diff is from a different PR? 

edit:

hmm could this be the cause?
image

stachbial:
> I'm so lost right now. What did I just review? The comments are here but the diff is from a different PR?
> 

@latekvo  Yep It looks like You reviewed a different PR, probably a a stale diff in your cache 👀

packages/tool-server/src/blueprints/js-runtime-debugger.ts
```diff
@@ -132,7 +132,9 @@ export const jsRuntimeDebuggerBlueprint: ServiceBlueprint<JsRuntimeDebuggerApi,
 
     await cdp.evaluate(DISABLE_LOGBOX_SCRIPT).catch(ignore);
 
-    await sourceMaps.waitForPending();
+    // Let source maps load in the background – consumers that need them
+    // (e.g. debugger-set-breakpoint) already call waitForPending() themselves.
+    void sourceMaps.waitForPending();
```
latekvo: Let's add an empty catch in case this promise is rejected for whatever reason.

packages/tool-server/src/utils/ios-device.ts
```diff
@@ -0,0 +1,341 @@
...
+        model:
+          d.hardwareProperties.marketingName ??
+          d.hardwareProperties.productType ??
+          "Unknown",
+        osVersion: d.deviceProperties.osVersionNumber,
+        connectionType: d.connectionProperties.transportType,
+      }));
+  } catch {
+    return [];
+  } finally {
```
latekvo: Since this is a direct to-agent output, i'd prefere there to be at least a bit of explanation that an error just happened. 
Returning empty list will likely confuse the agent, might even result in the agent never wanting to use the tool ever again (via memories).

latekvo (commented):
Great work, i wish we could merge this really soon ❤️ 
Just noticed i had these 2 comments pending for the last week - they're just nitpicks.

filip131311:
This is very old we can always revisit it in time

latekvo:
Testing and reworking https://github.com/software-mansion/argent/pull/369/changes at this very moment. Once it ships we'll re-open this one.

PR #69, opened 2026-04-01
feat(tool-server): npm update checker (ARG-79)

latekvo:
## Summary
- Adds a background npm update checker
- On startup: queries the npm registry for the latest version of `@swmansion/argent`
- Every hour: re-checks for updates
- Stores the result in a runtime flag (`updateAvailable`, `latestVersion`) via `getUpdateState()`
- Silently fails on network errors — never crashes the server

## Scope
For now the changes **only the update-checking logic**. No user-facing behavior is wired up yet.

## Test plan
- [x] Unit tests cover: update detected, no update, network failure, periodic recheck, dispose cleanup, non-200 responses
- [x] All 6 tests pass
- [x] `getUpdateState()` returns correct values verified via mocked unit tests

Closes ARG-79

### DIFF
--- a/packages/tool-server/src/utils/update-checker.ts
+++ b/packages/tool-server/src/utils/update-checker.ts
@@ -1,0 +1,107 @@
+import https from "node:https";
+import { version as currentVersion } from "../../package.json";
+
+const PACKAGE_NAME = "@argent/tool-server";
+const CHECK_INTERVAL_MS = 60 * 60 * 1000; // 1 hour
+const REQUEST_TIMEOUT_MS = 10_000;
+
+export interface UpdateState {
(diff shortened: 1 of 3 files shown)


latekvo:
This PR only adds the polling logic. Actual risky part is done in #74

packages/tool-server/src/utils/update-checker.ts
```diff
@@ -0,0 +1,107 @@
+import https from "node:https";
+import { version as currentVersion } from "../../package.json";
+
+const PACKAGE_NAME = "@argent/tool-server";
```
latekvo, line 4: This is a mistake, it will be fixed in #74

PR #70, opened 2026-04-01
improve: rewrite tool descriptions to meet SpiderShield quality criteria

latekvo:
Rewrites the `description` field of all 53 tool definitions in `packages/tool-server/src/tools/` to satisfy SpiderShield's 6 scoring criteria:

- **`has_action_verb`** — description starts with a strong action verb
- **`has_scenario_trigger`** — contains "Use when …"
- **`has_param_docs`** — key parameters documented inline
- **`has_param_examples`** — concrete example values (UDIDs, bundle IDs, ports, coordinates)
- **`has_error_guidance`** — documents failure conditions and what to do
- **`has_return_docs`** — documents the return shape

Result: **3.3 → 9.9 / 10** average description score. `gesture-pinch` and `gesture-rotate` are capped at 9.3 by SpiderShield's disambiguation penalty: both tools take the same parameter set (`center`, `scale`/`angle`, `duration`) and SpiderShield scores down any two tools whose descriptions share more than 50% of their content words — no amount of rephrasing can overcome this since the parameters themselves are what's being described.

Also updates the CI gate threshold from 10.0 to 9.9 to match the achievable score.

Only `description` fields and the CI threshold were changed. No logic, schemas, or other fields were touched.

### DIFF
--- a/packages/tool-server/src/tools/debugger/debugger-component-tree.ts
+++ b/packages/tool-server/src/tools/debugger/debugger-component-tree.ts
@@ -513,11 +513,22 @@ const zodSchema = z.object({
 
 export const debuggerComponentTreeTool: ToolDefinition<z.infer<typeof zodSchema>, string> = {
   id: "debugger-component-tree",
-  description: `Inspect the current React Native screen as a compact component tree. Use when you need tap coordinates or element names for React Native UI interactions.
-Accepts: port (default 8081), onScreenOnly, maxNodes, includeSkipped.
-Each component lists its name, text, and tap coordinates, e.g. Button "Submit" (tap: 0.50,0.72).
-Returns a text tree with component names and normalized tap coordinates for gesture-tap.
-Fails if Metro debugger is not connected or the app is not running.`,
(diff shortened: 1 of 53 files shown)


latekvo:
I'm still working on this, I think the SpiderShield requirementes might be wrong but i'm still working to see if that's true.
For example: `has_param_docs` really insentivizes duplication of param declarations inside the description which is redundant and token-wasteful.

latekvo:
This looks promising - but I'll have to benchmark if the changes improve tool usage rates.

These updates would go hand-in-hand with the tool usage rate tester we've had in Radon, **but** i don't want to spend too much time on this so i'll come back to it after i'm done with more crucial features

c.c. @pFornagiel tagging you to keep you up to date on this

latekvo:
The original attempt overfit to 2 criteria that i later had to completely disable due to their poor implementation. 
The current attempt looks really solid, the tool descriptions are also more consistant.
I'd personally prefer for the descriptions to be shorter, but that'll be done in the next PR, as i know from experience that it really isn't trivial to shorten the descriptions without them being overfit to this length criteria.

PR #71, opened 2026-04-01
ci: replace SpiderShield scoring with custom weights, drop param criteria

latekvo:
Explanation why: https://github.com/teehooai/spidershield/issues/1

Replaces SpiderShield's built-in `description_score` with a custom reweighting of its per-tool boolean outputs in the CI check step.

## What changed

Two SpiderShield criteria are excluded from scoring:

- **`has_param_docs`** — redundant: MCP agents already receive the full `inputSchema` alongside every tool description in `tools/list` responses; re-documenting parameters in the description is not required by the MCP spec or Anthropic's tool-writing guidelines
- **`has_param_examples`** — input notation varies per harness (some use JSON, some use positional args, some use natural language); penalises valid descriptions that correctly defer examples to the schema

The remaining six criteria are rescaled proportionally so the maximum achievable score stays 10.0:

| Criterion | Old weight | New weight |
|---|---|---|
| `has_action_verb` | 1.5 | 2.0 |
| `has_scenario_trigger` | 2.5 | 3.5 |
| `has_return_docs` | 1.0 | 1.5 |
| `has_error_guidance` | 1.0 | 1.0 |
| `disambiguation` | 1.0 | 1.5 |
| `length` | 0.5 | 0.5 |
| **total** | **10.0** | **10.0** |

Threshold is set to **9.9** — the current average on the `improve/tool-description-quality` branch. `gesture-pinch` and `gesture-rotate` score 9.12 due to an unavoidable disambiguation penalty (both tools have identical parameter sets), but the average stays above threshold.

## No changes to tool files or other scripts

### DIFF
--- a/.github/workflows/tool-description-quality.yml
+++ b/.github/workflows/tool-description-quality.yml
@@ -45,55 +45,84 @@ jobs:
           python3 -c "
           import json, sys
 
+          # Custom scoring — SpiderShield criteria we exclude:
+          #   has_param_docs    (redundant: MCP agents already receive the full inputSchema)
+          #   has_param_examples (notation varies per harness; punishes valid descriptions)
+          #
+          # Remaining weights scaled proportionally to keep max at 10.0:
+          #   has_action_verb      2.0  (was 1.5)
+          #   has_scenario_trigger 3.5  (was 2.5, bumped — most impactful for tool selection)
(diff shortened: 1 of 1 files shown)


latekvo:
Explanation: https://github.com/teehooai/spidershield/issues/1

PR #74, opened 2026-04-03
feat: add update-argent tool and per-response update notification

latekvo:
## Summary

- Adds an `update-argent` MCP tool that agents can call (with explicit user consent) to apply a pending update — spawns `npx @swmansion/argent update --yes` in a detached process with a 2-second grace period so the HTTP response flushes before SIGTERM arrives
- Injects a `note` field into every successful tool-server HTTP response when an update is available; the MCP server prepends this note as a text content block before the tool result, ensuring the AI sees the notice on its next tool call
- The note text is explicit about consent requirements: the AI must never apply an update without the user asking for it in the current conversation, and lists both the manual command (`npx @swmansion/argent update`) and the `update-argent` tool as the only two approved paths
- Includes the update-checker polling logic (hourly npm check, `getUpdateState()` runtime flag) that this feature builds on

## Note on scope

This PR includes the update-checker infrastructure (polling + runtime flag) as well as the two effectors. The notification mechanism has no user-facing impact until an actual update is available — the `note` field is omitted entirely when `updateAvailable` is false.

### DIFF
--- a/packages/mcp/src/mcp-server.ts
+++ b/packages/mcp/src/mcp-server.ts
@@ -82,7 +82,7 @@ export async function startMcpServer(): Promise<void> {
   async function callTool(
     name: string,
     args: unknown
-  ): Promise<{ result: unknown; outputHint?: string }> {
+  ): Promise<{ result: unknown; outputHint?: string; note?: string }> {
     const tools = await fetchTools();
     const meta = tools.find((t) => t.name === name);
     const res = await fetchWithReconnect(() => `${TOOLS_URL}/tools/${name}`, {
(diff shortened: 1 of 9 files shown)


latekvo:
Tested locally using Verdaccio local npm registry.
The agent noticed new versions both on startup and during the session.

Added tools:

```
- update-argent - applies the update (spawns argent update --yes detached, 2s delay)
- dismiss-update - suppresses the notification for N hours (accepts hours param)
```

What the agent sees when an update drops:

```
NOTE: An update for Argent is available (0.3.5 -> 0.3.7).

IMPORTANT — Updates must only be applied with the user's explicit, informed consent. Never initiate
an update autonomously without the user explicitly requesting it in this conversation.

How to update:
  - Manual (user-initiated): run `npx @swmansion/argent update` in the terminal
  - Agent-assisted (requires explicit user consent): call the `update-argent` tool — only after the
user has explicitly asked you to update Argent in this conversation
  - To postpone this reminder: call the `dismiss-update` tool with the number of hours to suppress
it

ACTION REQUIRED: Save a note (via TODO, task list, or memory) that an Argent update to v0.3.7 is
available so you can remind the user about it later. This notification will not repeat for the next
30 minutes.
```

PR #79, opened 2026-04-07
feat: default local MCP/skills install with custom path support

pFornagiel:
## Summary

- Changes CLI `init` to default to local installation of MCP and skills, with an option for a custom path
- Adds path validation in the `init` function, resolving and checking paths before use
- Improves `isGloballyInstalled` check using `which`/`where` instead of `npm list -g --json`
- Refactors `runShellCommand` to accept a structured `ShellCommand` object instead of a raw string
- Renames "argent" to "Argent" (capitalized) in installer-facing user messages (where it does not refer to `argent` command but the branding name itself)

### DIFF
--- a/packages/mcp/src/cli/init.ts
+++ b/packages/mcp/src/cli/init.ts
@@ -16,31 +16,28 @@ import {
   getLatestVersion,
   detectPackageManager,
   globalInstallCommand,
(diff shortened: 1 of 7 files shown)


filip131311 (approved):
lgtm

PR #80, opened 2026-04-07
fix: Rework descriptions to fix failing CIs

latekvo:
This PR adjust descriptions to align with SpiderShield's description requirements.

It also reduces the fail threshold to `9.0/10.0`


packages/tool-server/src/tools/native-devtools/native-network-logs.ts
```diff
@@ -21,13 +21,11 @@ type Result =
 
 export const nativeNetworkLogsTool: ToolDefinition<Params, Result> = {
   id: "native-network-logs",
-  description: `View network requests captured at the native NSURLProtocol level.
-
-Unlike the JS-level network inspector (view-network-logs), this captures ALL network
-traffic from the app including native modules, Swift/Objective-C networking, and
-background transfers that bypass JS fetch.
-
-If requiresRestart is returned: call restart-app then retry.`,
```
pFornagiel, line 30: `If status is restart_required: call restart-app then retry.`

I believe we should leave this information in on the native-devtools, to be more sure that agent knows what to do next after failing.

latekvo: I agree. Done in [82652e3](https://github.com/software-mansion/argent/pull/80/commits/82652e36ab0b55520f79de230bb068f08343726d)

filip131311 (changes requested):
conflicts

PR #81, opened 2026-04-07
fix: add node to types in tsconfig to resolve missing Node.js typing

latekvo:
This PR fixes the linting issues visible on the attached pictures.

When lib is explicitly set in `tsconfig.json` (e.g., `"lib": ["ES2022"]`), TypeScript does not auto-resolve `@types/node` even if the package is installed. We need "types": ["node"] in `compilerOptions` to explicitly include it.

image

image

image


filip131311 (approved):
lgtm

PR #82, opened 2026-04-07
fix: move simulator-server binary to native-devtools-ios and fix esbuild bundling

pFornagiel:
## Summary
- Moves the `simulator-server` binary from the repo root into `packages/native-devtools-ios/bin/`, co-locating it with the other native artifacts (dylibs). Updates all path references in docs, scripts, and source code accordingly.
- Adds the missing `@argent/native-devtools-ios` esbuild alias in `bundle-tools.cjs`, so the bundle step resolves from TypeScript source (consistent with how `@argent/registry` is already handled). This fixes `npm run pack:mcp` failing from a clean state, and eliminates the risk of silently bundling stale `dist/` artifacts from a previous build.


filip131311 (changes requested):
I remove the old sim server in my describe pr and added one build from radon ide main this pr will probably conflict with that quietly which seems bad

pFornagiel:
Fixed the merge conflicts, its fine

PR #85, opened 2026-04-07
chore: update README files for production release

pFornagiel:
## Summary

- Overhaul root and \`packages/mcp\` README files for a polished, user-facing presentation
- Replace internal developer docs (architecture, API reference, launch configs) with concise installation guide, capabilities overview, CLI reference, and supported editors table
- Add branding assets (banner image, placeholder GIF) and Software Mansion attribution

## README sync mechanism

The root \`README.md\` is the **single source of truth**. \`packages/mcp/README.md\` is always a verbatim copy of the root file — it is never edited directly.

Syncing is handled by \`packages/mcp/scripts/sync-readme.cjs\`, a tiny Node script that reads the root README and overwrites the package one:

\`\`\`
root README.md  ──(sync-readme.cjs)──▶  packages/mcp/README.md
\`\`\`

The script is wired into two npm lifecycle hooks in \`packages/mcp/package.json\`:

| Hook | Command | When it runs |
|------|---------|--------------|
| \`sync-readme\` | \`node scripts/sync-readme.cjs\` | Called explicitly via \`npm run sync-readme\` or as part of \`npm run build\` |
| \`prepack\` | \`node scripts/sync-readme.cjs\` | Runs automatically before \`npm pack\` / \`npm publish\`, ensuring npm always ships the latest README |

This means the package README stays in sync with the root without any manual copying — both during local builds and on every publish.


packages/mcp/README.md
```diff
@@ -1,99 +1,101 @@
...
 
-### Installing from a tarball
+This command triggers an installation wizard which:
 
-If you have a pre-built `.tgz` (e.g. from CI or `npm pack`), no registry auth is needed:
+- Installs `@swmansion/argent` globally
+- Detects your editor and registers the MCP server
+- Copies skills, rules, and agent definitions into your workspace
+
+Prefer a manual install?
```
latekvo: This blends in with the other steps grouped under the `#### Run 'init' in your project` header. 

```suggestion
#### Prefer a manual install?
```

README.md
```diff
@@ -1,144 +1,99 @@
...
-  -H "Content-Type: application/json" -d '{}'           # invoke a tool
-```
+> **Tip:** Once installed, ask your assistant _"What can Argent do?"_ - it will walk you through all capabilities available.
 
-## Installing in a project
 
-The `argent` package is distributed via [GitHub Packages](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-npm-registry) under the `@swmansion` scope. Because the source repository is private, you must authenticate before installing.
+<br/>
+<p align="center">
+  <img width="80%" src="assets/placeholder.gif" alt="argent banner" width="100%" />
```
latekvo: Duplicated width parameter.

```suggestion
  <img src="assets/placeholder.gif" alt="argent banner" width="100%" />
```

latekvo: That's the demo video, not the banner.

```suggestion
  <img width="80%" src="assets/placeholder.gif" alt="video demonstration of argent being used" width="100%" />
```

README.md
```diff
@@ -1,144 +1,99 @@
...
-| `--show-touches` / `--no-show-touches` | `SHOW_TOUCHES` | `true`  | Touch pointer overlay |
 
-### Quick reference (curl)
+## CLI Reference
 
-```bash
-curl http://localhost:3000/simulators/running              # list running simulators
+| Command         | Description                                                 |
+| --------------- | ----------------------------------------------------------- |
+| `argent init`   | Install globally and configure MCP in the current workspace |
```
latekvo: `argent install` is not mentioned anywhere in the CLI reference, despite being the command that's used in the "How to install" section above (which i think is incorrect either way, right?).

README.md
```diff
@@ -1,144 +1,99 @@
...
+#### Run `init` in your project
 
-## Building and testing
+From your project root:
 
 ```bash
-npm run build                          # tsc --build (all packages)
-npm test -w @argent/registry           # registry unit tests (vitest)
-npm test -w @argent/tool-server        # tool-server tests (vitest)
+npx @swmansion/argent install
```
latekvo: Header says "Run `init` in your project", while the command uses `install` not `init`.

I think `init` was meant to be used here.

latekvo (changes requested):

README.md
pFornagiel: It is referenced in the "How to install" section, as it is provided in the promotional video. `install` is an alias for `init`, same goes for `uninstall` being an alias for `remove`. Can change it for consistency to `init` or add information about CLI aliases to the table with usage.

README.md
pFornagiel: Same as the comment above

README.md
latekvo: 1. If they're an alias then it should be noted in the CLI reference. Otherwise we have undocumented commands that only we know are aliases of eachother.
2. The "Want to install manually?" section uses `init` instead of `install` which does suggest they should not be used interchangeably. If they're aliases i'd use `install` in both places, unless there's a good reason not to do that.

README.md
latekvo: Responded above.

latekvo:
> The root `README.md` is the single source of truth. `packages/mcp/README.md` is always a verbatim copy of the root file — it is never edited directly.

> Syncing is handled by `packages/mcp/scripts/sync-readme.cjs`, a tiny Node script that reads the root README and overwrites the package one

Can't we just symlink the two files instead? Or is there a reason why we aren't doing that?

latekvo (approved):
Discussed in person, thanks for the work ❤️

PR #86, opened 2026-04-08
feat: Use `launchd` instead of a subprocess

latekvo:
This PR changes the tool server launch method from subprocess execution to true daemon via `launchd`.
This change allows the process to persist even after its parent (terminal, VSCode, Cursor) has been killed.

### Let's hold off reviewing and merging this until the release

I also have a suspicion that this might be causing some crashes - that's what seemed to happened in my preliminary tests.


latekvo:
For now throwing this back into backlog

PR #87, opened 2026-04-08
feat: add accessibility permission gate during installation

latekvo:
Require macOS Accessibility permissions before proceeding with argent init. 
Shows a prominent banner, counts down 3 seconds, triggers the native system permission dialog, and polls for the result. 

- If the user accepts, a green success banner is shown. 
- If denied, installation is refused with manual instructions and an option to auto-open System Settings. In non-interactive mode (--yes), performs a quick check and exits if permissions are missing.


latekvo:
This was an alternate approach that we decided not to use.

PR #96, opened 2026-04-09
feat: enhance debugger API with app and device context

pFornagiel:
## Summary

- Added `appName` and `logicalDeviceId` fields to `JsRuntimeDebuggerApi` interface, populated from the selected Metro target's title and React Native logical device ID
- Propagated `appName` and `logicalDeviceId` to all debugger tool responses: `debugger-connect`, `debugger-status`, `debugger-evaluate`, `debugger-inspect-element`, `debugger-log-registry`, `debugger-reload-metro`
- `debugger-component-tree` now prepends a device context header line (`device | app | udid`) to the returned tree string

This was added because of the observation I made - when trying to reload a metro server with two available servers running, the agent sometimes mixes up some ports. This change is the best-effort approach to deal with this edge-case.

### DIFF
--- a/packages/tool-server/src/blueprints/js-runtime-debugger.ts
+++ b/packages/tool-server/src/blueprints/js-runtime-debugger.ts
@@ -89,6 +89,8 @@ export interface JsRuntimeDebuggerApi {
   port: number;
   projectRoot: string;
   deviceName: string;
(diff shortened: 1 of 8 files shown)


filip131311 (approved):
nice

PR #101, opened 2026-04-09
feat: add MCP config adapter for Cline

latekvo:
## Summary
- Add a new `Cline` MCP config adapter for the Cline VS Code extension (`saoudrizwan.claude-dev`)
- Global-only adapter storing config at VS Code's `globalStorage/saoudrizwan.claude-dev/settings/cline_mcp_settings.json`
- Uses standard `{ mcpServers: { ... } }` format (same as Cursor/Windsurf/Gemini)
- Cross-platform path resolution for macOS and Linux

## Test plan
- [x] All 69 existing + new tests pass (`npx vitest run test/cli/mcp-configs.test.ts`)
- [x] TypeScript compiles cleanly (`npx tsc --noEmit`)
- [x] Adapter registry updated from 7 to 8 adapters
- [x] Tests cover: write format, remove, non-existent remove, empty-server remove, projectPath (null), globalPath, preserve other servers

Resolves ARG-104

### DIFF
--- a/packages/mcp/src/cli/mcp-configs.ts
+++ b/packages/mcp/src/cli/mcp-configs.ts
@@ -466,6 +466,73 @@ const codexAdapter: McpConfigAdapter = {
   },
 };
 
+// ── Cline adapter ───────────────────────────────────────────────────────────
+// Format: { mcpServers: { argent: { command, args, env } } }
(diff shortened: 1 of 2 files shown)


latekvo:
Deferring this for now. Will come back to it after release.

PR #102, opened 2026-04-09
feat: add MCP config adapter for JetBrains IDEs

latekvo:
Adds a `JetBrains` MCP config adapter supporting IntelliJ IDEA, WebStorm, PyCharm, GoLand, Rider, CLion, PhpStorm, RubyMine, DataGrip, RustRover, Aqua, DataSpell, Fleet, and Android Studio

- Only works with local `.idea/mcp.json` project level configs for now

Resolves ARG-103


latekvo:
Postponing till `1.0.0`

packages/mcp/src/cli/mcp-configs.ts
```diff
@@ -604,6 +604,129 @@ const codexAdapter: McpConfigAdapter = {
...
+  "CLion",
+  "PhpStorm",
+  "RubyMine",
+  "DataGrip",
+  "RustRover",
+  "Aqua",
+  "DataSpell",
+  "Fleet",
+  "AndroidStudio",
+];
```
latekvo: Nearly each of these has a unique global config directory. So for now at least we're limiting ourselves to project config only, since on the project scope they all support the `.idea/` config directory.

PR #106, opened 2026-04-09
feat: add MCP config adapter for Hermes Agent

latekvo:
## Summary
- Add `hermesAdapter` to `packages/mcp/src/cli/mcp-configs.ts` supporting Hermes (Nous Research AI coding agent CLI) with YAML-based config format (`mcp_servers` key)
- Add `readYaml`/`writeYaml` helpers to `packages/mcp/src/cli/utils.ts` using the `yaml` package
- Add full test coverage for write, remove, preserve-other-servers, path resolution, and edge cases

https://github.com/user-attachments/assets/62b2a24c-61c9-4453-8fc6-bc00cfbaecc7

<br>

https://github.com/user-attachments/assets/9125c0ec-cb20-462e-99ba-1f5e9555a0c1

Before this PR, adapters with only one config path (only local or only global) without were installing regardless of user's scope selection. This PR makes this fact more visible, allowing the user to disable the undesired installation targets. 

Would appreciate a UI opinion, i went with A because B was scary in my opinion, but something in between could also work:

A | B
---|---
image|image

## Test plan
- [x] All 69 tests pass in `mcp-configs.test.ts` (7 new Hermes tests + updated adapter count test)
- [x] TypeScript compiles with `tsc --noEmit`

### DIFF
--- a/packages/mcp/src/cli/init.ts
+++ b/packages/mcp/src/cli/init.ts
@@ -206,11 +206,22 @@ export async function init(args: string[]): Promise<void> {
   if (nonInteractive) {
     selectedAdapters = detected.length > 0 ? detected : ALL_ADAPTERS;
   } else {
-    const choices = ALL_ADAPTERS.map((a) => ({
-      value: a,
-      label: a.name,
-      hint: detectedNames.includes(a.name) ? "detected" : undefined,
-    }));
+    const choices = ALL_ADAPTERS.map((a) => {
+      const parts: string[] = [];
+      if (detectedNames.includes(a.name)) parts.push("detected");
+      const hasProject = a.projectPath(process.cwd()) != null;
+      const hasGlobal = a.globalPath() != null;
+      if (!hasProject && hasGlobal) {
+        parts.push(pc.italic(pc.cyan(`ⓘ  will be installed into ${a.name}'s global config`)));
+      } else if (hasProject && !hasGlobal) {
+        parts.push(pc.italic(pc.cyan(`ⓘ  will be installed into ${a.name}'s project config`)));
+      }
+      return {
+        value: a,
+        label: a.name,
+        hint: parts.length > 0 ? parts.join(", ") : undefined,
+      };
+    });
 
     p.log.message(pc.dim("  Use arrow keys to move, space to toggle, enter to confirm."));
 
--- a/packages/mcp/src/cli/init.ts
+++ b/packages/mcp/src/cli/init.ts
@@ -310,6 +310,16 @@ export async function init(args: string[]): Promise<void> {
         } catch (err) {
           mcpResults.push(`${pc.red("x")} ${adapter.name}: ${pc.dim(String(err))}`);
         }
+      } else if (scope !== "global" && adapter.globalPath()) {
+        const fallback = adapter.globalPath()!;
+        try {
+          adapter.write(fallback, mcpEntry);
+          mcpResults.push(
+            `${pc.green("+")} ${adapter.name} ${pc.dim(`(global fallback: ${fallback})`)}`
+          );
+        } catch (err) {
+          mcpResults.push(`${pc.red("x")} ${adapter.name}: ${pc.dim(String(err))}`);
+        }
       } else {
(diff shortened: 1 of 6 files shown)


packages/mcp/src/cli/init.ts
```diff
@@ -299,6 +300,15 @@ export async function init(args: string[]): Promise<void> {
...
         }
+      } else if (scope !== "global" && !localPath && globalPath) {
+        try {
+          adapter.write(globalPath, mcpEntry);
+          mcpResults.push(
+            `${pc.green("+")} ${adapter.name} ${pc.dim(`(global-only fallback: ${globalPath})`)}`
+          );
+        } catch (err) {
+          mcpResults.push(`${pc.red("x")} ${adapter.name}: ${pc.dim(String(err))}`);
+        }
```
latekvo: We want to strictly adhere to `scope`. We do not want a global fallback when `local` fails.

```suggestion
```

latekvo:
Appears to work, reviewing code and polishing final issues
image

packages/mcp/src/cli/init.ts
```diff
@@ -310,6 +310,16 @@ export async function init(args: string[]): Promise<void> {
...
+      } else if (scope !== "global" && adapter.globalPath()) {
+        const fallback = adapter.globalPath()!;
+        try {
+          adapter.write(fallback, mcpEntry);
+          mcpResults.push(
+            `${pc.green("+")} ${adapter.name} ${pc.dim(`(global fallback: ${fallback})`)}`
+          );
+        } catch (err) {
+          mcpResults.push(`${pc.red("x")} ${adapter.name}: ${pc.dim(String(err))}`);
+        }
```
latekvo, line 333: This `if-else` already handled `global->local` fallback.
This PR only adds the third `local->global` fallback case handling.

User is notified about this fallback mechanism at the agent selection stage - that's also where they can choose to disable integration with a specific agent.

packages/mcp/src/cli/utils.ts
```diff
@@ -81,6 +82,29 @@ export function writeToml(filePath: string, data: Record<string, unknown>): void
...
+  }
+  return doc;
+}
+
+export function writeYaml(filePath: string, doc: Document): void {
+  fs.mkdirSync(path.dirname(filePath), { recursive: true });
+  // lineWidth: 0 disables hard-wrap so long user strings (e.g. multi-line
+  // quoted personalities in ~/.hermes/config.yaml) stay on the lines they
+  // were on. Default would re-wrap at column 80.
+  fs.writeFileSync(filePath, doc.toString({ lineWidth: 0 }));
```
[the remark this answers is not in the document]
latekvo: Your smarter brother said:

> The feedback is technically invalid for the current yaml package. I just checked every case — empty
>   Document, populated Document, parseDocument on input with no trailing newline, Document after setIn
>   mutations — and doc.toString({ lineWidth: 0 }) always returns a string ending in \n. This is
>   documented behavior of the yaml package (2.x guarantees trailing newline on toString).
> 
>   The reason writeJson and writeToml append \n explicitly is that JSON.stringify and smol-toml's
>   stringify genuinely don't — that's not a stylistic consistency choice, it's compensating for a real
>   gap. YAML has no such gap.

packages/mcp/src/cli/init.ts
```diff
@@ -206,11 +206,22 @@ export async function init(args: string[]): Promise<void> {
...
+        parts.push(pc.italic(pc.cyan(`ⓘ  will be installed into ${a.name}'s global config`)));
+      } else if (hasProject && !hasGlobal) {
+        parts.push(pc.italic(pc.cyan(`ⓘ  will be installed into ${a.name}'s project config`)));
+      }
+      return {
+        value: a,
+        label: a.name,
+        hint: parts.length > 0 ? parts.join(", ") : undefined,
+      };
+    });
```
latekvo, line 224: This fragment adds the blue info text:

Image

filip131311 (approved):
blue italic option is nice

PR #112, opened 2026-04-13
fix: Removing skills removed skills external to argent

filip131311:


packages/mcp/src/cli/uninstall.ts
```diff
@@ -154,10 +446,12 @@ export async function uninstall(args: string[]): Promise<void> {
...
 
     if (pruneResults.length > 0) {
-      p.note(pruneResults.join("\n"), "Pruned Directories");
+      p.note(pruneResults.join("\n"), "Pruned Argent Content");
+    } else {
+      p.log.info(pc.dim("No Argent-owned skills, rules, or agents found to remove."));
     }
   } else {
-    p.log.info(pc.dim("Kept skills, rules, and agents directories. Pass --prune to remove them."));
+    p.log.info(pc.dim("Kept Argent-owned skills, rules, and agents. Pass --prune to remove them."));
```
latekvo: When `argent uninstall` is ran without the `--prune` flag, argent skills will stay, which is something we don't want.
Or am i wrong on either of these statements?

packages/mcp/src/cli/uninstall.ts
```diff
@@ -91,51 +364,70 @@ export async function uninstall(args: string[]): Promise<void> {
   if (shouldPrune) {
```
latekvo, line 363: Would really prefer a guard clause here, this if-else is bearly readable

packages/mcp/src/cli/uninstall.ts
```diff
@@ -91,51 +364,70 @@ export async function uninstall(args: string[]): Promise<void> {
...
-        }
+    const bundledSkillNames = getBundledSkillNames(SKILLS_DIR);
+    const skillTargets: CleanupTarget[] = [
+      { targetDir: path.join(projectRoot, ".claude", "skills"), label: ".claude/skills" },
+      { targetDir: path.join(projectRoot, ".cursor", "skills"), label: ".cursor/skills" },
+      { targetDir: path.join(projectRoot, ".agents", "skills"), label: ".agents/skills" },
+      { targetDir: path.join(homedir(), ".claude", "skills"), label: "~/.claude/skills" },
+      { targetDir: path.join(homedir(), ".cursor", "skills"), label: "~/.cursor/skills" },
+      { targetDir: path.join(homedir(), ".agents", "skills"), label: "~/.agents/skills" },
+    ];
```
latekvo: Hardcoded duplications for home and for local + i'm pretty sure we store the actual global config path in the adapter definitions + we should use the Path class constructor instead of string literals.

Sidenote: Have you tested if this will work in a project when someone is in: `myProject/src/` and runs `argent uninstall`?

packages/mcp/src/cli/uninstall.ts
```diff
@@ -91,51 +364,70 @@ export async function uninstall(args: string[]): Promise<void> {
...
+        sourceDir: RULES_DIR,
+        targets: [
+          { targetDir: path.join(projectRoot, ".claude", "rules"), label: ".claude/rules" },
+          { targetDir: path.join(projectRoot, ".cursor", "rules"), label: ".cursor/rules" },
+          { targetDir: path.join(projectRoot, ".gemini", "rules"), label: ".gemini/rules" },
+          { targetDir: path.join(homedir(), ".claude", "rules"), label: "~/.claude/rules" },
+          { targetDir: path.join(homedir(), ".cursor", "rules"), label: "~/.cursor/rules" },
+          { targetDir: path.join(homedir(), ".gemini", "rules"), label: "~/.gemini/rules" },
+        ],
+        contentLabel: "rule",
```
latekvo, line 407: Same here, 20 nearly identical lines.

latekvo (approved):
Just a nitpick: the `mcp-configs.test.ts` removal tests were relaxed from asserting `"argent is gone but other entries remain"` to just `expect(fs.existsSync(configPath)).toBe(false)`

Rest is fine, thank you for the pr 👍

PR #113, opened 2026-04-13
feat: Show offline banner during argent init

latekvo:
## Summary
- Detects offline state during `argent init` and gracefully degrades
- Skills step: warns and skips to manual when offline; uses cached `npx skills` with `--no-install` when available
- Update check: bounded by a 3s timeout so it doesn't hang offline

### DIFF
--- a/packages/mcp/src/cli/utils.ts
+++ b/packages/mcp/src/cli/utils.ts
@@ -95,6 +95,17 @@ export function getLatestVersion(): string {
   return result.trim();
 }
 
+export function isSkillsCliAvailable(): boolean {
+  try {
+    execSync("npx --no-install skills --version", {
+      stdio: ["ignore", "ignore", "ignore"],
+    });
+    return true;
+  } catch {
+    return false;
+  }
+}
+
 export async function isOnline(timeoutMs = 1500): Promise<boolean> {
   const host = new URL(NPM_REGISTRY).hostname;
   const lookup = new Promise<boolean>((resolve) => {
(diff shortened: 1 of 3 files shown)


latekvo:
Tested. Banner only displays when:
1. Device is offline
2. AND the required packages are missing

<br>
<br>

image

packages/mcp/src/cli/utils.ts
```diff
@@ -128,13 +129,46 @@ export function getInstalledVersion(): string | null {
   }
 }
 
+const PROBE_TIMEOUT_MS = 3_000;
```
latekvo, line 132: I believe this is an acceptable wait time for the edge case where someone is offline and is still running `argent init`.

From my measurements:
- `isOnline()` probes DNS, for me it takes `100-200ms` every time
- `isSkillsCliAvailable()` probes local fs via npm, takes `0.5-1s`
- `getLatestVersion()` talks over https, on super super slow connections it could hit the wall, but it doesn't actually regress any features, only defers the available update to a later date. Nothing else happens.

filip131311: why not just increes the probe time for getLatestVersion specifically to lets say 5_000 ms

filip131311 (approved):
approved but left a question that is actually a suggestion

packages/mcp/src/cli/utils.ts
latekvo: Couple of things:
1. The user physically waits for this timer to pass
2. Hitting this particular timer just causes the argent not to update on installation if there is an update available.
3. It's unlikely the timer will be hit in the first place, that requires 3k ping

I'm okay with rising it to 5s but only after we add loader spinner that's visible to the user - i'll do that in a separate PR because it has to be done in a couple of places and would make this PR couplefolds larger in both scope and code.

PR #114, opened 2026-04-13
fix: add `project_root` arg to `flow-start-recording`

latekvo:
## Summary

Before this PR we were calling `process.cwd()` from `tool-server`. This was bad because `tool-server` can be launched from anywhere, not always the project cwd. Also it frequently caused `getGitRoot()` to fail: `fatal: not a git repository`

**`flow-start-recording`** now takes a required `project_root` parameter.

`react-profiler-analyze` already does the same thing: required parameter `project_root` so the profiler knows which project to analyze. This PR applies the same pattern to flow tools.

Also moved all flows from `project/.argent/` to `project/.argent/flows`.

### How the flow looks now:

```
flow-start-recording  { name: "my-flow", project_root: "/Users/dev/MyApp", executionPrerequisite: "..." }
flow-add-echo  { message: "Step label" }
flow-add-step  { command: "gesture-tap", args: "..." }
flow-finish-recording  {}
flow-execute   { name: "my-flow", prerequisiteAcknowledged: true }
```


filip131311 (approved):

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -1,32 +1,47 @@
...
 
 // ── Paths ────────────────────────────────────────────────────────────
 
-async function getGitRoot(): Promise<string> {
-  const { stdout } = await execFileAsync("git", ["rev-parse", "--show-toplevel"]);
-  return stdout.trim();
+// ── Active session state ─────────────────────────────────────────────
+
+let activeFlowName: string | null = null;
+let activeProjectRoot: string | null = null;
```
latekvo, line 12: This does not support multiple concurrent flow creation sessions unfortunately. This is a broader issue that i plan to address in future - it's out of the scope of this PR.

latekvo:
Reviewed and tested on multiple parallel projects. All flows were written to the expected locations.

PR #117, opened 2026-04-13
Require device_id for debugger/profiler; React profiler session metadata; skill updates

pFornagiel:
## Description

This PR aims to resolve the issue where the profiler sessions would affect one-another if started on separate devices. The cause was probably moving the profiling files out of the current project folder, where the agent lacked context about where the profiling session came from, as it was marked by port alone.

In order to facilitate this:
- the calls to **profiler, debugger and network-inspector** now require **port and device_id (UUID)** in order to distinguish between different devices and properly take care of reading and caching the profiling session files
- this change should also help with other debugger tools, giving the agent more context when it comes to working with more than one device simultaneously
- also added more metadata when it comes to listing available profiler session files, for more reference to agent

For reference, this is the usual flow of profiling flow (with now added `device_id` identifier):
- Start: react-profiler-start ties to port + device_id, starts Hermes CPU sampling over CDP, and captures React commits in JS for the recording window.
- Stop: react-profiler-stop ends sampling, fetches the CPU profile plus a filtered commit set (hot commits ≥16ms ± margin), writes *_cpu.json / *_commits.json to the debug dir, caches those paths under port+device_id, then disposes the live profiler session.
- Analyze / reload: react-profiler-analyze reads those files via the cache and runs the offline pipeline to produce the markdown report (and may write a CPU index for queries); profiler-load (load_react) can rehydrate the same cache from disk later so query tools work without a new capture.

### Blueprints and tool contracts

- **`device_id` is required** across the JsRuntimeDebugger and ReactProfilerSession blueprints and the tools that depend on them (Metro debugger, network inspector, React profiler start/stop/load/analyze, combined report, CPU summaries, fiber tree, renders, component source, profiler queries).
- **React profiler session** is created with payload `port:deviceId`, ties the session to the same device identity as the debugger, and exposes **`appName`**, **`deviceName`**, Hermes version, detected architecture, and profiling/session path metadata on the session API.
- **Profiler load / analyze paths** and session debug dumps were adjusted so **reported field names match the runtime and project details** actually available from the session.

### Tests

- **`network-inspector-blueprint.test`** updated for the required `device_id` argument and related blueprint behavior.

### Agent skills (documentation)

- **`argent-metro-debugger`** and **`argent-react-native-profiler`** skills (including reference docs) now **spell out `device_id` usage**, connection expectations, and profiler diagnostics so scripted flows align with the tool schemas.


filip131311 (approved):
thank you

PR #118, opened 2026-04-14
chore: Remove duplicate method

latekvo:
Both methods are identical. One checks for `IDLE || TERMINATING` the other for `TERMINATING || IDLE`.
I removed one of them.


filip131311 (approved):
fiar

PR #120, opened 2026-04-14
feat(mcp): support progressive tool loading via _meta hints

latekvo:
MCP standard allows MCP servers to add "meta" hints to each declared tool. Claude Code looks for:

- `_meta["anthropic/alwaysLoad"]` — a request to keep this tool's full schema visible on every turn rather than load it on demand.
- `_meta["anthropic/searchHint"]` — a short phrase that helps the client surface the tool when a user request matches it, without needing the full description loaded into context.

This PR threads both hints end-to-end through the Argent stack and sets them conservatively on a small core of tools. Clients that don't consume these keys simply ignore them per the MCP spec, so nothing changes for them.

This PR focuses more on the `alwaysLoad` hint, `searchHint` exists but we need more benchmarking to decide how  we want to apply it to the tools with hidden descriptions (and also if it's worth the time spent, since this is a claude-only feature).

before|after
---|---
image|image

**Curation — only the eight tools the model reaches for on practically every turn opt in:**
- `describe`
- `debugger-component-tree`
- `screenshot`
- `gesture-tap`
- `gesture-swipe`
- `launch-app`
- `restart-app`
- `run-sequence`


filip131311 (commented):
we have a tool for restarting the app we shoud add it to the curated list: 
describe
debugger-component-tree
screenshot
gesture-tap
gesture-swipe
launch-app
run-sequence
 

as it is vital for othere tools that depend on controling app start to work

filip131311:
BTW: nice find

latekvo:
Deferring till `1.0.0` release

pFornagiel:
Seems like the most important toolings are included, good job, approve from my side.

Note: it would be interesting to see whether adding a seldom-used tool (like `flow-add-step`) would make agent more happy using this tool or make it overuse the tool. This should not be implemented in production, but it would be interesting to watch if you find some free time to do it.

pFornagiel (approved):

PR #122, opened 2026-04-15
chore: update argent-private submodule with build-native-binaries workflow

filip131311:
## Summary

- Updates `packages/argent-private` submodule to latest `main`
- Picks up the new `build-native-binaries.yml` GitHub Actions workflow that builds, signs, and publishes dylibs and ax-service to `argent-private-releases`

## Test plan
- [ ] Verify submodule points to the correct commit


latekvo:
> Updates packages/argent-private submodule to latest main

Perhaps we could consider a CI that'd sync it every time it updates/releases?
From what i've seen adding a CI like that is nearly free, Opus manages it. Just a question of whether we want to do that or not.

filip131311:
@latekvo not now but next week I would love a stable solution

PR #123, opened 2026-04-15
fix: drain stdout before process.exit in tool-server CLI paths

latekvo:
## Summary

This PR fixes:
- `SyntaxError: Unterminated string in JSON at position 8170 (line 191 column 1275)` during `argent init`. 

Root cause: 
- `packages/tool-server/src/index.ts` writes the ~61 KB tool list to `process.stdout` and then synchronously calls `process.exit(0)`. 
- When stdout is a pipe, `process.exit()` does not drain `WriteStream` buffer, so the tail of the JSON gets discarded.

Fixed by awaiting buffer write completion before calling `process.exit`.

### DIFF
--- a/packages/tool-server/src/index.ts
+++ b/packages/tool-server/src/index.ts
@@ -82,14 +82,21 @@ export function getAvailableTools(): Array<{
   });
 }
 
-function printUsage(stream: NodeJS.WriteStream): void {
-  stream.write(
+function usageText(): string {
+  return (
     "Usage: tool-server <command>\n\n" +
-      "Commands:\n" +
-      "  start                        Start the tool server\n" +
-      "  -t, --get-available-tools    Print available tools as JSON and exit\n" +
-      "  -h, --help                   Show this menu\n"
+    "Commands:\n" +
+    "  start                        Start the tool server\n" +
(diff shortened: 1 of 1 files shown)


latekvo:
Tested, both the installation and usage goes well.
image

packages/tool-server/src/index.ts
```diff
@@ -82,29 +82,33 @@ export function getAvailableTools(): Array<{
...
+    "  start                        Start the tool server\n" +
+    "  -t, --get-available-tools    Print available tools as JSON and exit\n" +
+    "  -h, --help                   Show this menu\n"
   );
 }
 
+// process.exit() does not drain Node's WriteStream buffer when stdout/stderr
+// is a pipe, so large writes must be flushed via the write callback before
+// exiting.
+function writeAndExit(stream: NodeJS.WriteStream, chunk: string, code: number): void {
```
filip131311, line 98: this helper seems exesive this is basically the same as 

`process.stdout.write(JSON.stringify(getAvailableTools(), null, 2) + "\n", () => process.exit(0));`

but it is fine

filip131311 (approved):
👍

PR #130, opened 2026-04-16
fix: add .catch() to watcherReady promise chain

filip131311:
## Summary

- `watcherReady.then(...)` in the tool server had no `.catch()` handler, so if `app.listen()` threw (e.g. port already in use), it produced an unhandled promise rejection that could crash the process.
- Added a `.catch()` that logs the error to stderr and exits with code 1, so failures surface cleanly instead of silently crashing.

## Test plan

- [ ] Start the tool server normally and confirm it still listens and logs as before.
- [ ] Start a second instance on the same port and verify it logs `[tool-server] Failed to start: ...` and exits with code 1 instead of an unhandled rejection.


filip131311:

PR #131, opened 2026-04-16
fix: log stdout/stderr errors instead of rethrowing

filip131311:
## Summary

- Throwing inside a Node.js event listener creates an uncaught exception that crashes the process
- Non-EPIPE `stdout` errors are now logged to `stderr` instead of rethrowing
- Non-EPIPE `stderr` errors fall back to writing to `stdout` (wrapped in try/catch since both streams may be broken)

## Test plan

- [ ] Verify the tool-server still silently ignores EPIPE errors when the parent process stops reading
- [ ] Simulate a non-EPIPE stdout error and confirm it is logged to stderr without crashing
- [ ] Simulate both streams broken and confirm the process does not crash


filip131311:

PR #132, opened 2026-04-16
fix: catch errors in TypedEventEmitter listeners

filip131311:
## Summary

- A throwing listener in `TypedEventEmitter.emit()` would stop iteration of remaining listeners and propagate the error to the caller.
- During service teardown this could skip cleanup steps and leak resources.
- Each listener is now isolated with a try/catch so all listeners run even if one throws. Errors are logged to stderr rather than swallowed silently.

## Test plan

- [ ] Verify that when one listener throws, subsequent listeners still execute
- [ ] Verify error messages appear on stderr with event name and error details
- [ ] Confirm existing emit behavior is unchanged when no listeners throw


filip131311:

PR #133, opened 2026-04-16
fix: handle rejected promises in simulator watcher

filip131311:
## Summary
- Fire-and-forget `initSimulator` calls in `startSimulatorWatcher` could produce unhandled promise rejections if an unexpected error occurred before the internal try block.
- Added explicit `.catch(() => {})` to the `initSimulator` calls in the forEach loop and to the `poll(false)` call in setInterval.

## Test plan
- [ ] Verify the tool-server starts and polls for simulators without errors.
- [ ] Confirm that booting/shutting down simulators during polling does not surface unhandled rejections.


filip131311:

PR #134, opened 2026-04-16
fix: handle rejected promises in update checker

filip131311:
## Summary

- Fire-and-forget `check()` calls in `startUpdateChecker()` are async but were not guarded with `.catch()`. While `fetchLatestVersion()` is internally defensive, `check()` also calls `process.stderr.write()` which can throw if stderr is broken, producing an unhandled promise rejection.
- Added `.catch(() => {})` to both the initial `check()` call and the one inside `setInterval` to silently swallow any unexpected errors, since update checks must never crash the server.

## Test plan

- [ ] Verify the tool server starts without regressions
- [ ] Confirm no unhandled rejection warnings appear when stderr is unavailable


filip131311:

PR #135, opened 2026-04-16
fix: tool server crash recovery for MCP server

filip131311:
## Summary

- **MCP CLI**: global error handlers log but don't `process.exit(1)` when running as the MCP server — the long-lived stdio process must survive tool server crashes
- **Tool server**: on fatal errors, attempts graceful `registry.dispose()` (with 5s timeout) before exiting so child processes (simulator servers, AX daemons) are not orphaned
- **`fetchWithReconnect`**: retries up to 5 times with exponential backoff (500ms → 8s, ~15.5s total) giving a restarted tool server time to boot
- **`ListToolsRequestSchema`**: wrapped in try/catch, returns empty tool list on failure instead of propagating raw errors to Claude Code
- **Health monitor**: 30s interval pings the tool server and preemptively triggers `reconnect()` if it's down — so the next tool call doesn't pay the full restart latency
- **Defense-in-depth**: `TypedEventEmitter.emit()` catches per-listener errors, fire-and-forget promises get `.catch()` handlers, stdout/stderr error handlers log instead of rethrowing

## Test plan

- [ ] Start MCP server via `argent mcp`, kill tool server PID (`~/.argent/tool-server.json`), invoke a tool — should recover automatically
- [ ] Kill tool server, wait 30s, check stderr for health monitor reconnect log, then invoke tool — should succeed immediately
- [ ] Verify `list_tools` returns empty array (not an error) when tool server is unreachable
- [ ] Verify tool server logs cleanup attempt on `kill -TERM` (graceful shutdown path)
- [ ] `npx tsc --noEmit` passes

### DIFF
--- a/packages/mcp/src/mcp-server.ts
+++ b/packages/mcp/src/mcp-server.ts
@@ -43,12 +43,24 @@ export async function startMcpServer(): Promise<void> {
   }
 
   async function fetchWithReconnect(getUrl: () => string, init?: RequestInit): Promise<Response> {
-    try {
-      return await fetch(getUrl(), init);
-    } catch {
-      await reconnect();
-      return fetch(getUrl(), init);
+    const MAX_RETRIES = 5;
+    let lastError: unknown;
+
+    for (let attempt = 0; attempt <= MAX_RETRIES; attempt++) {
+      try {
+        return await fetch(getUrl(), init);
+      } catch (err) {
+        lastError = err;
+        if (attempt === MAX_RETRIES) break;
+        if (attempt === 0) {
+          // First failure: trigger reconnect (spawns new server if dead)
+          await reconnect();
+        }
+        // Exponential backoff: 500ms, 1s, 2s, 4s, 8s (~15.5s total)
+        await new Promise((r) => setTimeout(r, 500 * Math.pow(2, attempt)));
+      }
     }
+    throw lastError;
   }
 
   const LOG_FILE = process.env.ARGENT_MCP_LOG ?? `${homedir()}/.argent/mcp-calls.log`;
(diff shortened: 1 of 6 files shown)


packages/mcp/src/mcp-server.ts
```diff
@@ -43,12 +43,24 @@ export async function startMcpServer(): Promise<void> {
...
+        return await fetch(getUrl(), init);
+      } catch (err) {
+        lastError = err;
+        if (attempt === MAX_RETRIES) break;
+        if (attempt === 0) {
+          // First failure: trigger reconnect (spawns new server if dead)
+          await reconnect();
+        }
+        // Exponential backoff: 500ms, 1s, 2s, 4s, 8s (~15.5s total)
+        await new Promise((r) => setTimeout(r, 500 * Math.pow(2, attempt)));
```
pFornagiel: Do we want to keep trying for 15 seconds? I would halve the time, but that is personal

pFornagiel:
At this point, if we do a silent catch on `initSimulator`, can we add a log inside it as well? Seems like scope of PR, potentially helps later (`line 24` function):

```ts
async function initSimulator(
  registry: Registry,
  watchedUdids: Set<string>,
  udid: string
): Promise<void> {
  watchedUdids.add(udid);
  try {
    await registry.resolveService(`${NATIVE_DEVTOOLS_NAMESPACE}:${udid}`);
  } catch (err) { // actually log the error
    watchedUdids.delete(udid);
    process.stderr.write(
      `[simulator-watcher] initSimulator failed for ${udid}: ${err instanceof Error ? err.message : err}\n`
    );
  }
}

```

filip131311:
done

PR #137, opened 2026-04-16
fix: crash recovery review issues from #135 + other findings

latekvo:
## Summary

Follow-up fixes for #135 (tool-server crash recovery):

- **`shutdown()` accepts exit code** — `crashShutdown` now exits with code 1 instead of 0; SIGINT/SIGTERM wrapped in arrow functions to prevent Node passing the signal string as `exitCode`
- **`shutdown()` awaits `server.close()`** before `process.exit()` so in-flight HTTP responses finish cleanly, with a `PROCESS_TIMEOUT_MS` safety timeout to prevent hangs
- **`forceExit` timer keeps event loop alive** — removed `.unref()` from `crashShutdown`'s force-exit timer
- **`GET /tools` touches idle timer** — health-check polling now keeps the tool-server alive between tool calls
- **`PROCESS_TIMEOUT_MS` constant** — replaces magic `5_000` numbers in shutdown/crash paths

## Tests

- `fetch-timeout.test.ts` — verifies `fetchWithReconnect` timeout and retry behavior
- `idle-timeout.test.ts` — new case verifying `GET /tools` resets the idle timer


packages/tool-server/src/index.ts
```diff
@@ -54,13 +54,17 @@ export function start(): void {
...
     httpHandle.dispose();
     await registry.dispose();
-    server?.close();
-    process.exit(0);
+    if (server) {
+      const forceExit = setTimeout(() => process.exit(exitCode), 5_000);
+      await new Promise<void>((resolve) => server!.close(() => resolve()));
+      clearTimeout(forceExit);
+    }
+    process.exit(exitCode);
```
latekvo, line 68: Prevents hangs. Makes sense.

packages/tool-server/src/index.ts
```diff
@@ -93,8 +97,9 @@ export function start(): void {
     });
 
   // ── Lifecycle ─────────────────────────────────────────────────────
-  process.on("SIGINT", shutdown);
-  process.on("SIGTERM", shutdown);
+  // `process.on` passes "SIGINT" as the first arg. Passing it would cause a TypeError crash. Using `() =>` ignores it.
+  process.on("SIGINT", () => shutdown());
+  process.on("SIGTERM", () => shutdown());
```
latekvo, line 103: Tested with a minimal repro. `process.on` calls the 2nd argument with the first argument.
Without arrow functions it crashed, since `shutdown()` does not expect `"SIGINT"` or any other string literal as it's first argument.

packages/tool-server/src/index.ts
```diff
@@ -10,16 +10,16 @@ export function start(): void {
...
   function crashShutdown(label: string, detail: string): void {
     process.stderr.write(`[tool-server] ${label}: ${detail}\n`);
     if (shuttingDown) return; // avoid re-entrant shutdown
     shuttingDown = true;
+    // 5s grace period for registry.dispose() to clean up child processes
     const forceExit = setTimeout(() => process.exit(1), 5_000);
-    forceExit.unref();
     if (shutdown) {
-      shutdown().catch(() => process.exit(1));
+      shutdown(1).catch(() => process.exit(1));
```
latekvo, line 23: This is inside a `crashShutdown` - I think we want to return `1` regardless of if the `shutdown` throws or not.

packages/tool-server/src/http.ts
```diff
@@ -55,6 +55,7 @@ export function createHttpApp(registry: Registry, options?: HttpAppOptions): Htt
   });
 
   app.get("/tools", (_req: Request, res: Response) => {
+    idleTimer.touch();
```
latekvo, line 58: This is minor. It bumps the 30 minute shutdown timer on the tool server. We are already doing this on `POST "/tools/:name"`, now we're also doing it on `GET "/tools"`

filip131311:
1) 

> stderr error handler falls back to stdout for non-EPIPE errors (safe — launcher only regex-matches the first stdout line)

I belive this is very wrong mcp server is consumed by clients that talk to it via stdout so this might corupt the protocol 

2) 

some unit tests are failing 


3) 
> fetchWithReconnect wraps each attempt in an AbortController with a configurable timeout (ARGENT_FETCH_TIMEOUT_MS, default 30s) — a hanging tool-server no longer wedges the MCP parent indefinitely 

in one place you increest the time otu to 10 s and here you introduce a 30s time out will i talo not break the "retry multiple times" functionality? 



I general can you meke this more structered so its not so confuseing

packages/tool-server/src/index.ts
```diff
@@ -54,13 +55,17 @@ export function start(): void {
...
     updateChecker.dispose();
     stopWatcher();
     httpHandle.dispose();
     await registry.dispose();
-    server?.close();
-    process.exit(0);
+    if (server) {
+      const forceExit = setTimeout(() => process.exit(exitCode), PROCESS_TIMEOUT_MS);
+      await new Promise<void>((resolve) => server!.close(() => resolve()));
+      clearTimeout(forceExit);
```
pFornagiel, line 66: Are we sure here that server is gonna survive in the Promise and not change to `undefined`?

latekvo: Yes, the only two possible paths from `if (server)` are:
1. Server closes, promise resolves
2. Server doesn't close, program forcefully exits with `.exit()`

I can't see a possible moment where `server` could turn nullish.

pFornagiel:
Right, looks good imo 👍

pFornagiel (approved):

PR #144, opened 2026-04-16
fix: npx init skips global install

filip131311:
## Summary
- When running `npx @swmansion/argent init`, npx temporarily puts the `argent` binary on PATH inside its cache (`_npx` directory)
- `isGloballyInstalled()` uses `which argent`, finds the npx-cached binary, and returns `true` — so the global install step is skipped entirely
- After npx exits, that PATH entry is gone and `argent` command is not found
- Fix: check if the resolved path contains `_npx` and treat it as not globally installed

## Test plan
- Run `npx @swmansion/argent init` on a machine without argent installed globally
- Verify it now prompts to install globally instead of skipping to the update check


filip131311:
closing in favor of an alternative solution

PR #146, opened 2026-04-17
docs: add privacy footer to README

latekvo:
## Summary

- Adds a Privacy section to the README documenting that Argent does not collect, transmit, or phone home any user data.
- Notes local-only MCP stdio integration with the agent and the `127.0.0.1`-bound internal tool server.
- Discloses the single outbound call (the npm registry version check) so the claim is accurate rather than absolute.

## Test plan

- [ ] Render the README on GitHub and verify the new section appears between Supported Editors and License.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ argent init
 
 ## Privacy
 
-Argent does not collect or transmit any user data. 
+Argent does not collect or transmit any user data.
 No telemetry, no analytics, no crash reporting.
 
 - Argent integrates with your agent locally over MCP stdio.
--- a/README.md
+++ b/README.md
@@ -86,6 +86,14 @@ argent init
 | Gemini CLI  | `.gemini/settings.json`                                       |
 | Codex CLI   | `.codex/config.yaml`                                          |
 
+## Privacy
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -86,6 +86,14 @@ argent init
 | Gemini CLI  | `.gemini/settings.json`                                       |
 | Codex CLI   | `.codex/config.yaml`                                          |
 
+## Privacy
+
+Argent does not collect, transmit, or phone home any user data. No telemetry, no analytics, no crash reporting.
+
+- Argent integrates with your agent (Claude Code, Cursor, etc.) locally over MCP stdio.
+- Its internal tool server binds to `127.0.0.1` only and is not reachable from outside your machine.
+- The only outbound network call is an optional version check against the public npm registry (`registry.npmjs.org`), which sends no user data and fails gracefully if blocked.
```
latekvo:
```suggestion
Argent does not collect or transmit any user data. 
No telemetry, no analytics, no crash reporting.

- Argent integrates with your agent locally over MCP stdio.
- Its internal tools are not reachable from outside your machine.
- The only outbound network call we make is the version check against our public npm package, which sends no user data and fails gracefully if blocked.
```

filip131311 (approved):
👍

PR #151, opened 2026-04-21
fix: actually update skills on argent update and use semver-aware comparison

filip131311:
## Summary
- `argent update` used `installed !== latest` when deciding whether an upgrade was available — same bug pattern as #150, but in the update command. A locally-installed prerelease would be offered a downgrade. Switched to `isNewerVersion` (semver-aware).
- The "skills" step in the update flow only ran `npx skills check`, which reports available updates but never applies them. Replaced it with `npx skills update -y`, so skills are actually updated when the user agrees.
- `argent update --yes` / `-y` now updates skills unattended (previously it skipped the skills prompt entirely, so scripted/CI upgrades silently fell behind on skills).

## Test plan
- [x] `npx vitest run` in `packages/mcp` — 223 tests pass
- [ ] Run `argent update` against a locally installed prerelease — confirm "Already on the latest version" instead of a downgrade prompt
- [ ] Run `argent update --yes` — confirm skills update step runs without prompting
- [ ] Run `argent update` interactively — confirm skills update step prompts and runs `npx skills update -y` on accept


pFornagiel:
Changes seem to be sound and work - let me know if you want to add the flag discussed internally for testing or just go forward and merge - will give you approval if the second option is so

packages/mcp/src/cli/utils.ts
```diff
@@ -35,6 +36,58 @@ export const SKILLS_DIR = resolveBundledDir("skills");
...
+  }
+}
+
+// Locations of the skills-CLI lock files. These paths mirror the skills CLI
+// (v1.5.x) so we can detect which scopes already track argent skills and only
+// re-sync those. Project lock lives next to the user's working tree, global
+// lock under XDG_STATE_HOME or ~/.agents/.
+export function getProjectSkillLockPath(cwd: string = process.cwd()): string {
+  return path.join(cwd, "skills-lock.json");
+}
```
pFornagiel, line 62: This one can be a errorous on the second thought. We have already defined a custom `resolveProjectRoot` in `utils.ts` function for exactly this purpose (vs just using the `cwd`), which uses some heuristics (such as folders location and .git directories) to determine where the root is. 

Note that this discrepancy does not only happen here - it is also in `update.ts` logic that was implemented earlier. Did not cause issues so far, but you may take a look and determine which approach you find more correct. We should stick to single approach everywhere.

filip131311: i use the "better" utill consistantly now

PR #152, opened 2026-04-21
feat: add MCP config adapter for opencode

latekvo:
This PR adds MCP config adapter for OpenCode

At one point OpenCode worked presumably because it read `.claude` or some other config (i've had like 13 of them on my machine). That doesn't seem to be the case anymore.

before|after
---|---
image|image
image|image


packages/argent-installer/src/mcp-configs.ts
```diff
@@ -670,6 +670,89 @@ const hermesAdapter: McpConfigAdapter = {
...
+  },
+
+  addAllowlist(root: string, scope: "local" | "global"): void {
+    const configPath = scope === "global" ? this.globalPath() : this.projectPath(root);
+    if (!configPath) return;
+    const config = readJson(configPath);
+    const tools = (config.tools ?? {}) as Record<string, unknown>;
+    tools[OPENCODE_ALLOWLIST_PATTERN] = true;
+    config.tools = tools;
+    writeJson(configPath, config);
```
latekvo, line 742: Tested, opencode allows this `argent*` glob pattern for allowlisting our tools.

PR #155, opened 2026-04-21
feat(tool-server): prefix stdout/stderr lines with ISO timestamp

latekvo:
Adds timestamp to every `tool-server.log` line.

`mcp-calls.log` already has timestamps, so no changes.

The tz is UTC+0 because that's what we already use for the `mcp-calls.log`

Works:

image

### DIFF
--- a/packages/tool-server/src/index.ts
+++ b/packages/tool-server/src/index.ts
@@ -8,6 +8,12 @@ import { startUpdateChecker } from "./utils/update-checker";
 const PROCESS_TIMEOUT_MS = 5_000;
 
 export function start(): void {
+  for (const stream of [process.stdout, process.stderr] as const) {
+    const orig = stream.write.bind(stream);
+    stream.write = ((chunk: string | Uint8Array, ...rest: unknown[]) =>
+      orig(`[${new Date().toISOString()}] ${chunk}`, ...(rest as []))) as typeof stream.write;
+  }
+
   // ── Global error handlers ─────────────────────────────────────────
   // The tool server should exit on uncaught errors (state may be corrupted),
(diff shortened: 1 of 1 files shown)


packages/tool-server/src/index.ts
```diff
@@ -8,6 +8,12 @@ import { startUpdateChecker } from "./utils/update-checker";
 const PROCESS_TIMEOUT_MS = 5_000;
 
 export function start(): void {
+  for (const stream of [process.stdout, process.stderr] as const) {
+    const orig = stream.write.bind(stream);
+    stream.write = ((chunk: string | Uint8Array, ...rest: unknown[]) =>
+      orig(`[${new Date().toISOString()}] ${chunk}`, ...(rest as []))) as typeof stream.write;
+  }
```
pFornagiel, line 29: Honestly, I would:
- extract this into a separate function (like `initializeStdioTimestampWrapper`) that initialises this with proper docstring, which justifies that does not cause any corruption, as we are overwriting builtins after all
- also maybe introduce some kind of environmental variable, which could easly turn it off if we were to test whether this causes problems in some future use (like `WRAP_STDIO_DISABLED`)

pFornagiel (dismissed):

latekvo:
Tested - still works after [ed4c649](https://github.com/software-mansion/argent/pull/155/commits/ed4c649046febef866374529f5b9f0d555e58998).

filip131311 (approved):
We realy should go through the console not process.stderr.write and overide the console not the write

latekvo:
Merging as discussed in person. This is a low priority matter.

PR #157, opened 2026-04-23
fix(describe): bump argent-private with WKWebView describe fix

latekvo:
The native ax-service now recurses into `primaryApp.children` beyond the flat `explorerElements` pass, so WKWebView web content (labels from the WebContent process) is surfaced alongside the host app's VoiceOver targets.

Details: https://github.com/software-mansion/argent-private/pull/7


filip131311 (approved):
point to main and its nice thank you

PR #166, opened 2026-04-28
fix: redbox on `debugger-evaluate`

### LINKED ISSUE #160: ERROR  [Invariant Violation: `new NativeEventEmitter()` requires a non-null argument.]
Never seen this error until running Argent on my app. I rebuilt it with `npx expo run:ios` and was still facing the same issues.

It still works and the agent is continue by dismissing it every time but this still continues to show up without fail.

I suspect this is happening when using the `debugger-evaluate` tool

Image

latekvo:
Fixes #160

On some setups `debugger-evaluate` tool call triggered a redbox.

This PR fixes that issue.

- We wrap `LBData.addException` to short-circuit while `LBData.isDisabled()` is true.
- `LB.ignoreAllLogs(false)` then re-enables fatals naturally — no uninstall needed.

## Details

`LogBox.ignoreAllLogs(true)` only flips an observer flag — it doesn't prevent `LogBoxData.addException` from enqueueing fatal logs or the inspector from auto-opening for them. Wrapping `addException` (rather than replacing it with a no-op) means uncaught top-level throws like `TurboModuleRegistry.getEnforcing('SegmentFetcher')` (common in Expo `expo run:ios` builds where the native module isn't linked) no longer surface as a redbox while LogBox is disabled — and re-enabling LogBox restores them.

A handful of lines in `disable-logbox.ts`; the rest of the diff is a new test file (no tests existed for this script before).

## Test plan

- [x] `npm test` in `packages/tool-server` — all pass, including 5 new vitest cases covering the wrapper, the re-enable path, the no-double-wrap guard, the no-`__r` early-exit, and the ErrorUtils-fallback path
- [x] `npm run build` clean
- [x] **End-to-end repro on iOS simulator** (iPhone 16 Pro, RN 0.81.5, Expo 54 dev-client): an async uncaught throw via `debugger-evaluate` reproduced the issue's exact redbox under the unpatched script; under the patched script the redbox was suppressed, and after `LB.ignoreAllLogs(false)` a subsequent throw surfaced again as expected.

### DIFF
--- a/packages/tool-server/src/utils/debugger/scripts/disable-logbox.ts
+++ b/packages/tool-server/src/utils/debugger/scripts/disable-logbox.ts
@@ -1,11 +1,17 @@
 /**
  * IIFE that scans the Metro module registry for the LogBox module
  * and calls `ignoreAllLogs(true)` to suppress the yellow/red overlay,
- * monkey-patches `LogBoxData.addException` to drop uncaught exceptions
- * (the redbox path used by `TurboModuleRegistry.getEnforcing(...)` and
- * other top-level module-load throws — `ignoreAllLogs` only flips an
+ * wraps `LogBoxData.addException` so it short-circuits while LogBox is
+ * disabled (the redbox path used by `TurboModuleRegistry.getEnforcing(...)`
+ * and other top-level module-load throws — `ignoreAllLogs` only flips an
  * observer flag and does not block `addException`), then clears any
  * already-queued LogBox entries.
+ *
+ * The wrapper consults `LBData.isDisabled()` on each call so a later
+ * `LB.ignoreAllLogs(false)` restores fatal-redbox visibility — important
+ * for devs who disconnect the debugger and keep developing. When
+ * `isDisabled` is unavailable, the wrapper defaults to no-op (matching
+ * the prior behavior on RN versions without that API).
  *
  * Uses `__r.getModules()` (available in DEV) to iterate only
  * already-initialized modules, avoiding forced evaluation of unloaded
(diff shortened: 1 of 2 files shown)


pFornagiel:
I do not believe this is what we want.

The logbox fix was implemented here the way it was done in Radon in order to supress the messages at the bottom of app (like `"Open debugger to view warnings"`) which obscured agent's work field and caused it to accidentally tap on boxes which would open debugger - thus causing further problems with debugger connection, etc.

Right now, if we disable the behaviour of Redbox appearing on errors, the agent is simply left confused why the box did not appear if the error surfaced - same goes for detection, if something goes wrong agent has no source of knowing.

Fatal logs SHOULD obscure the screen and cause agent to debug what went wrong. If I interpret this change the wrong way, you may correct me, but at the time I believe this PR should be closed in the current state, or change to fix the issue it mean to solve, not silently ignore.

pFornagiel:
I would advise digging into the actual cause of the [issue](https://github.com/software-mansion/argent/issues/160) instead of silently ignoring. This is an issue with native modules, that people have encountered historically and it may be worth actually reproducing it.

filip131311:
I agree with @pFornagiel that this does not seem like a solution to the problem

latekvo:
As discussed in person, we'll not be proceeding with this fix.

I initially assumed that `debugger-evaluate` is an isolated environment that we do not want affecting the main app. With that assumption, ignoring redboxes would be desired.

This assumptions turned out to be false - the agent sometimes opts to use the `debugger-evaluate` tool for app interactions or even for navigating around the app - we'll not be preventing the agent from doing so for now, hence we also want to keep the current redbox behaviour.

PR #168, opened 2026-04-28
fix: preserve old Zed config

### LINKED ISSUE #165: `argent init` overwrites global Zed settings.json with no backup
## Summary
  Running `npx @swmansion/argent init` replaced my entire `~/.config/zed/settings.json` with an argent-only stub,
   wiping every existing setting (theme, icon theme, fonts, vim mode, assistant config, agent_servers, etc.). No
  backup was created — the original file is simply gone. There is no prompt, no warning in the terminal output,
  and nothing in the docs indicating the file will be replaced rather than patched.

  ## Reproduction
  1. Have an existing `~/.config/zed/settings.json` with custom settings.
  2. Run `npx @swmansion/argent init` in any project.
  3. Open `~/.config/zed/settings.json` — everything you had is gone, replaced with an argent stub.

  ## Expected
  preserving all other top-level keys and comments. If a safe merge isn't possible, it should:
  - Refuse to write and tell the user what to add manually, **or**
  - Prompt for confirmation and write a timestamped backup (e.g. `settings.backup-2026-04-28T11-02.json`) before
  overwriting.

  ## Actual
  The file is replaced wholesale with:
  ```json
  {
    "context_servers": {
      "argent": {
        "source": "custom",
        "command": "argent",
        "args": ["mcp"],
        "env": { "ARGENT_MCP_LOG": "/Users/<me>/.argent/mcp-calls.log" }
      }
    },
    "agent": {
      "tool_permissions": { "default": "allow" }
    }
  }
```
  No backup file is written by argent. The original config is unrecoverable unless the user happens to have
  version control, an editor undo buffer, or an unrelated pre-existing backup. In my case I only partially
  recovered because an 8-month-old settings_backup.json from a prior Zed migration happened to be sitting in the
  same directory — pure luck, and it was missing every change I'd made since August.

  Environment

  - macOS (darwin 25.2.0)
  - Zed (current stable)
  - argent: latest via npx @swmansion/argent init

### PR DESCRIPTION
**Pull request by latekvo**
Closes #165.

`argent init` was wiping users' `~/.config/zed/settings.json` - Zed's settings format is JSONC, so `readJson` used `JSON.parse` and silently fell back to `{}` on any parse failure, after which the Zed adapter would overwrite the file with an argent-only stub.

### DIFF
--- a/packages/argent-installer/src/utils.ts
+++ b/packages/argent-installer/src/utils.ts
@@ -7,6 +7,12 @@ import semver from "semver";
 import { PACKAGE_NAME, NPM_REGISTRY } from "./constants.js";
 import { parse as parseToml, stringify as stringifyToml } from "smol-toml";
 import { Document, parseDocument } from "yaml";
+import {
+  applyEdits as applyJsoncEdits,
+  modify as modifyJsonc,
+  parse as parseJsonc,
+  type JSONPath,
+} from "jsonc-parser";
 
 // ── Package root resolution ───────────────────────────────────────────────────
 // At runtime this module ships in two shapes:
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: packages/argent-installer/src/utils.ts [hunk 1]
```diff
@@ -174,11 +174,147 @@ export function writeYaml(filePath: string, doc: Document): void {
...
+}
+
+/**
+ * Read a JSON-with-Comments file (line and block comments plus trailing
+ * commas). Used for editors whose settings format is JSONC — currently Zed.
+ *
+ * Comments and trailing whitespace are not preserved when the caller writes
+ * back via {@link writeJson}; data is.
+ */
+export function readJsonc(filePath: string): Record<string, unknown> {
```
filip131311, line 255: no need for custom parser https://www.npmjs.com/package/jsonc-parser

filip131311 (changes requested):
seems like easy enough fix we don't need to implement the jsonc parser by hand tho we can just use existing one https://www.npmjs.com/package/jsonc-parser

packages/argent/scripts/bundle-tools.cjs
```diff
@@ -35,6 +35,14 @@ const ALIASES = {
   "@argent/cli": CLI_ENTRY,
 };
 
+// esbuild on platform:"node" defaults mainFields to ["main","module"], which
+// picks UMD entries that use runtime `require("./impl/...")`. Those requires
+// can't be statically resolved during bundling and fail at runtime because
+// the impl/ folder isn't shipped next to the bundle. Prefer "module" so we
+// pick ESM entries with static imports that get fully inlined. Affects e.g.
+// jsonc-parser, which ships both UMD (main) and ESM (module).
+const MAIN_FIELDS = ["module", "main"];
```
latekvo, line 44: @filip131311 This looks like solution to a possibly hallucinated problem. If it targets a real issue, please make the comment reasonably short.

filip131311 (approved):

packages/argent/scripts/bundle-tools.cjs
filip131311: turns out to be real problem

PR #170, opened 2026-04-30
fix: opencode detect alternative filenames

latekvo:
OpenCode supports:
- project-scope: `opencode.jsonc` and `opencode.json`; 
- global-scope: `config.json` (legacy/migrated), `opencode.json`, and `opencode.jsonc`

This PR adds support for these alternative filenames.
This PR also adds `jsonc` support.

image


filip131311 (approved):
looks legitt

PR #182, opened 2026-05-04
feat: add `argent server start` for long-lived tool-server

filip131311:
## Human summary: 

This adds `argent server start` command to enable us to spawn tools server in arbitrary location. we want this to enable using remote tools-server via our mcp not only local ones. 

## Summary

- Adds `argent server start` to spawn a tool-server that does **not** auto-shutdown on idle, with configurable bind port and host — the missing piece for running tool-server on a remote/dedicated host.
- Foreground by default so process supervisors (systemd, Docker, supervisord) own the lifecycle; `--detach` keeps the existing daemon-spawn behavior. Refuses to clobber a healthy running server unless `--force`.
- Phase 2 (consumer-side `--tools-url` flag pointing the MCP / CLI at a remote tool-server) is now a one-flag change away — `tools-client.ts` and `mcp-server.ts` already honor `ARGENT_TOOLS_URL`. Out of scope for this PR.

## What changed

- **`tool-server`** — read `HOST` env var (default `127.0.0.1` so the existing auto-spawn path is unaffected), bind to it, attach an `error` listener on `app.listen` so `EADDRINUSE` / `EACCES` exit cleanly instead of routing through `uncaughtException`.
- **`argent-tools-client`** — broaden the readiness regex to match any host; extend `buildToolsServerEnv` with optional `{ host, idleTimeoutMinutes }`; record `host` in `~/.argent/tool-server.json` (backward-compat: missing field defaults to `127.0.0.1`); export `spawnToolsServer`, `findFreePort`, `isToolsServerHealthy`, `isToolsServerProcessAlive`, the state read/write/clear helpers, and `formatToolsServerUrl` so the CLI can reuse them.
- **`argent-cli`** — new `server start [flags]`: `--port`/`-p` (default 3001, `0` = pick free), `--host` (default `127.0.0.1`), `--idle-timeout` (default `0` = never), `--detach`/`-d`, `--force`, `--help`. Warns when host is non-loopback. Foreground mode forwards `SIGINT`/`SIGTERM` to the child and writes the state file so local `argent run` / `argent mcp` can attach.
- **`argent` dispatcher** — pass `BUNDLED_RUNTIME_PATHS` to `server()` so the new subcommand can find the bundled tool-server.

## Test plan

- [x] `argent server --help` lists the new `start` subcommand
- [x] `argent server start --help` shows full flag reference
- [x] `argent server start --detach --port 0` → status alive/healthy, `/tools` reachable, state file populated with `host` field
- [x] Second `argent server start --detach` refuses (`exit 1`) with helpful message
- [x] `argent server start --detach --force` kills the old PID and starts a new one
- [x] Foreground `argent server start --port 0` runs with inherited stdio; child PID is the actual tool-server (not the wrapper); `argent server stop` from a sibling shell tears down both
- [x] Existing auto-spawn path (`argent tools`) still works — verifies the regex broadening + default-host change is backward-compatible
- [x] `tool-server` (444 tests) + `argent-tools-client` (1) + `argent` dispatcher (1) suites pass
- [ ] Manual smoke on `--host 0.0.0.0` — could not exercise locally (sandbox declined the public-interface bind); the warning + bind path is identical to loopback besides the listen address


packages/tool-server/src/index.ts
```diff
@@ -62,6 +62,9 @@ export function start(): void {
 
   // ── Config ────────────────────────────────────────────────────────
   const PORT = parseInt(process.env.PORT ?? "3001", 10);
+  // HOST defaults to loopback so the local auto-spawn path stays safe.
+  // `argent server start --host 0.0.0.0` is the opt-in for remote exposure.
+  const HOST = process.env.HOST ?? "127.0.0.1";
```
pFornagiel: I would change both the `PORT` and `HOST` env varibales in all the code references to `ARGENT_HOST` and `ARGENT_PORT`. There is a chance someone will have defined `HOST` env variable in their shell and this will cause hard-to-reproduce problems.

packages/tool-server/src/index.ts
```diff
@@ -103,16 +106,25 @@ export function start(): void {
...
-      server = httpHandle.app.listen(PORT, "127.0.0.1", () => {
+      server = httpHandle.app.listen(PORT, HOST, () => {
         const addr = server!.address();
         const boundPort = typeof addr === "object" && addr ? addr.port : PORT;
-        process.stdout.write(`Tools server listening on http://127.0.0.1:${boundPort}\n`);
-        process.stderr.write(`  GET  http://127.0.0.1:${boundPort}/tools\n`);
-        process.stderr.write(`  POST http://127.0.0.1:${boundPort}/tools/:name\n`);
+        process.stdout.write(`Tools server listening on http://${HOST}:${boundPort}\n`);
+        process.stderr.write(`  GET  http://${HOST}:${boundPort}/tools\n`);
+        process.stderr.write(`  POST http://${HOST}:${boundPort}/tools/:name\n`);
```
pFornagiel: Tiny NIT, because it is only user-facing, but if host is ipv6 the authority part should be written in square brackets like this:
```
http://[::1]:3001
```

packages/argent-cli/src/server.ts
```diff
@@ -97,13 +89,312 @@ function logsCmd(follow: boolean): void {
...
+  }
+
+  await new Promise<void>((resolve) => {
+    child.on("exit", (code, signal) => {
+      process.removeListener("SIGINT", onInt);
+      process.removeListener("SIGTERM", onTerm);
+      if (stateWritten) {
+        clearToolsServerState().catch(() => {
+          /* non-fatal */
+        });
```
pFornagiel, line 454: The `writeToolServerState` is asynchronous, so there could be a situation when:
- parent spawns child tool-server
- parent starts async `writeToolsServerState`, which is not awaited
- child may exit quickly  (bind fail for example)
- exit handler runs and only clears state if `stateWritten === true`
- if write resolves after exit handler, stale state file remains, which can cause agent confusion as the data will be referencing a server that has failed at startup

or it may be fine, hard to tell, but this seems like a real edge-case

pFornagiel:
Taking it from here

pFornagiel:
**Note on idle-timeout default change**

tldr; nothing changes on the local argent usage, if someone were to run pure binary they will get a server that never times-out - this is not expected use case and has been done for consistency with how the server works now.

The old `DEFAULT_IDLE_TIMEOUT_MINUTES = 30` lived inside the tool-server binary itself, so any direct invocation of the binary without `ARGENT_IDLE_TIMEOUT_MINUTES` set would auto-shutdown after 30 minutes.

The new default is `0` (run forever). This is intentional - `argent server start` is designed for daemon/supervisor use cases where the process lifecycle is managed externally so an implicit auto-shutdown would be surprising there.

The MCP auto-spawn path is unaffected: it still explicitly passes `ARGENT_IDLE_TIMEOUT_MINUTES=30` via `AUTOSPAWN_IDLE_TIMEOUT_MINUTES`. The change only affects users who were invoking the tool-server binary directly without setting the env var.

pFornagiel:
Tested, the only functionality which does not work are the screenshots, but this is out of scope for this PR - will have to add a patch for sending them reliably through the network later on

PR #183, opened 2026-05-04
fix(skills): prefix bare skill references with 'argent-'

latekvo:
All skills are currently exposed as `argent-***`. In the markdown though, we refer to them as just `***`. This may confuse the agent when it tries reading `skills/***` instead of `skills/argent-***`.

This PR changes all skill references in docs to the correct `argent-***` form. 

## Why?
There's no downside to this fix. Keeping it as-is potentially introduces some confusion especially if when the user is running a weaker model or if we encounter namespace conflicts.


filip131311 (approved):
LGTM

PR #148, opened 2026-04-17
feat: Android emulator support

latekvo:
### For reviewers - what to focus on:

The diff is ~6k LoC; this breakdown should help triage where to spend attention:

| Bucket | Lines | Effort |
| --- | --- | --- |
| **Net-new Android logic**: `boot-device.ts` (660) + `adb.ts` (399) + `uiautomator-parser.ts` (292) | 1351 | **Read carefully** |
| Blueprint `DeviceInfo` refactor: `simulator-server`, `ax-service`, `native-devtools` | 365 | Read |
| Tool `index.ts` / `types.ts` sprawl across ~20 tools (capability gates, `services()`, descriptions) | ~965 | Skim — watch for cross-file consistency |
| `platforms/android.ts` impls (stub → real implementation) | ~426 | Skim |
| Skill/docs prose, profiler rename ripple | ~1000 | Skim |
| **`platforms/ios.ts` deletions** — 12 files, 0 additions; logic moved into unified handlers | 684 | **Skip — mechanical** |

### Just works thanks to either sim server or metro:
- `gesture-*`
- `button`, `keyboard`
- `screenshot`
- `run-sequence`, `flow-*`
- `debugger-*`
- `react-profiler-*`

### Adjusted to work on android too:

| tool | android | ios |
| --- | --- | --- |
| `describe` | `uiautomator dump` → `DescribeNode` | `AXServiceApi` |
| `launch-app` | `cmd package resolve-activity` + `am start -W` | `simctl launch` |
| `restart-app` | `cmd package resolve-activity` + `am force-stop` + `am start -W` | `simctl terminate` + `simctl launch` |
| `reinstall-app` | `adb install -r` (+`-g`/`-d`) | `simctl uninstall` + `simctl install` |
| `open-url` | `am start -a VIEW -d <url>` | `simctl openurl` |
| `list-devices` | `adb devices` + `emulator -list-avds` | `xcrun simctl list --json` |
| `boot-device` | snapshot probe → `-force-snapshot-load` hot boot with screencap-alive guard, falls back to cold boot on any failure | `simctl boot` + `bootstatus -b` |

### Will not work on android for now:
- `native-devtools-*` (7 total)
- `native-profiler-*` (3 total — was `ios-profiler-*`)
- `profiler-stack-query`, `profiler-combined-report` (depend on `native-profiler-*`)
- `paste` — unregistered on both platforms for now.

PR #185, opened 2026-05-04
feat: clean up `describe` output

latekvo:
Stacked on #148 — review after that lands.

# STILL WORK IN PROGRESS

## Why

Real captures from the iPhone 16 Pro home screen and a Pixel emulator launcher:

| Platform | Before | After | Δ pretty | Δ minified |
| --- | --- | --- | --- | --- |
| Android (uiautomator) | 32 nodes, depth 14, 17.9 KB pretty | 23 nodes, depth 6, 8.0 KB pretty | **−56 %** | −37 % |
| iOS (AX-service) | 18 nodes, depth 1, 4.2 KB pretty | 15 nodes, depth 1, 3.0 KB pretty | **−28 %** | −34 % |

The Android tree had 5+ FrameLayout / LinearLayout wrappers with no `content-desc` / `resource-id` chained around every actually-meaningful leaf, and a launcher-chain (`Screen → id/content → id/launcher → id/drag_layer`) of single-child same-bounds containers. The iOS tree had every widget element repeated verbatim (Map, "MONDAY, 04 MAY", "No events today" each appeared twice).

## What

A shared post-processor `compressDescribeTree` runs after both platform adapters (AX + native-devtools fallback on iOS, uiautomator on Android). It applies four rules, all idempotent:

1. **Drop pure-noise structural wrappers.** `View` / `ViewGroup` / `FrameLayout` / `LinearLayout` / `RelativeLayout` / `AXGroup` / etc. with no `label`, `identifier`, or `value` are removed and their children promoted up.
2. **Collapse same-frame single-child structural wrappers.** Forward `label` / `identifier` / `value` to the surviving child if it has none. Refuses the collapse when wrapper and child carry **differing labels or values** (e.g. `AXGroup label="Calendar widget" → AXButton label="Open"` is preserved as two layers — both labels are real). Identifier conflicts collapse with the child winning, since the chains they appear in (`id/content → id/launcher → id/drag_layer`) are platform-internal noise.
3. **Deduplicate structurally-identical sibling subtrees.** Hash on `(role, frame, label, identifier, value, recursive children)` — kills the iOS widget-repeat noise without folding distinct items that happen to share a frame.
4. **Round normalized frame components to 4 decimals.** ≈ 0.1 px on a 1080-wide screen, well under any tap-target tolerance. Replaces the AX service's `0.07960199005`-style float-noise tails.

The synthetic root is always preserved (single-root contract). Semantic roles (`Button`, `TextField`, `StaticText`, `Image`, `Switch`, `CheckBox`, `RadioButton`, `ScrollView`, `WebView`, `AXButton`, `AXTextField`, `AXStaticText`, `AXLink`, `AXImage`, `AXHeading`, `AXTabBar`, `AXAdjustable`) are never dropped — agents can still reason about what to do with each element.

### Behavior change worth flagging

iOS describe previously returned `source: "ax-service"` whenever AX returned any elements. Now an AX response containing only labelless trait-less elements (which adapt to bare `AXGroup` leaves) compresses to an empty tree and triggers the existing native-devtools fallback. That's almost always a strict improvement (the fallback gives a richer, app-scoped tree) but it is a real behavioral expansion. Pinned with a regression test in `describe-tool.test.ts`.

## Test plan

- [x] Unit: 24 new tests in `describe-compress.test.ts` covering wrapper promotion, same-frame collapse, conflict refusal, sibling dedup, frame rounding, idempotence, no-op shapes, and input-mutation safety.
- [x] Unit: 1 new test in `describe-tool.test.ts` pinning the AX-empty-after-compression → native fallback path.
- [x] Full tool-server suite: 627 pass (was 620, +7 new). Registry suite: 57 pass.
- [x] `npx prettier --check .` clean for changed files.
- [x] Real-world before/after measured on captured iOS + Android dumps (numbers above).
- [ ] Visual regression: I have not run this against a live RN app yet — would be good for a reviewer to spot-check that targeting elements via `describe` still works in their usual flow.

## Stack

Branch is one commit + one post-review-fix commit on top of `feat/android-emulator-support` (PR #148). No diverging history; once #148 lands, GitHub will auto-retarget this PR to `main`.

### DIFF
--- a/packages/tool-server/src/tools/describe/compress.ts
+++ b/packages/tool-server/src/tools/describe/compress.ts
@@ -72,9 +72,26 @@ function framesEqual(a: DescribeFrame, b: DescribeFrame): boolean {
 // Targeting info (label / identifier) on the wrapper is forwarded to the
 // surviving child if the child has none, so e.g. `id/launcher → unnamed View`
 // becomes a single `View id=launcher` instead of vanishing the id entirely.
+//
+// Refuse to collapse only when both layers carry differing *labels* or
+// *values*: an `AXGroup label="Calendar widget" → AXButton label="Open"`
+// stack at the same bounds describes two semantically distinct entities and
+// folding it to just the button would silently lose the widget context.
+// Identifier conflicts between layers are common platform-internal noise —
+// the launcher chain `id/content → id/launcher → id/drag_layer` carries a
+// distinct resource-id at every level even though no agent would target the
+// outer two. We tolerate that by keeping the child's identifier and dropping
+// the wrapper's; the merge function below preserves the wrapper's id only
+// when the child has none.
 function shouldCollapseSingleChildWrapper(node: DescribeNode, only: DescribeNode): boolean {
   if (!STRUCTURAL_ROLES.has(node.role)) return false;
   if (!framesEqual(node.frame, only.frame)) return false;
+  if (node.label !== undefined && only.label !== undefined && node.label !== only.label) {
+    return false;
+  }
+  if (node.value !== undefined && only.value !== undefined && node.value !== only.value) {
+    return false;
+  }
   return true;
 }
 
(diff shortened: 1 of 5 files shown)


latekvo:
added to android branch

PR #188, opened 2026-05-05
fix installer update version detection under npx

pFornagiel:
## Summary
- fix `argent update` version detection when invoked via `npx @swmansion/argent`, so it compares against the actually installed global binary instead of the transient npx-cached package
- add `getGloballyInstalledVersion()` that resolves the real global binary path (`which -a` / `where`), follows symlinks, and reads the owning `package.json`
- add regression tests covering staged global-install layouts and npx-style transient paths to prevent false "already on latest" results


filip131311 (approved):
LGTM

PR #190, opened 2026-05-05
feat: [android-WIP] enhance DescribeNode interface

pFornagiel:
As stated, for reference implementation attempt


latekvo:
discussed in person - added to android branch

PR #191, opened 2026-05-06
fix: 100% CPU usage on broken stdio

latekvo:
One of our users reported occasional 100% CPU usage stale MCP server process.

I wasn't able to reproduce this bug directly in Codex. We only have non-codex repros that are able to consistantly reproduce a bug that *looks* like the one reported to us.

The bug in question is: Our error handler causes errors which throws it into an infinite loop.
 
image

### More details:

When `argent mcp` runs as an orphaned process whose stderr pipe is broken (e.g. parent died), every `stderr.write` from the uncaughtException handler emits an async 'error' event. Without an 'error' listener that becomes another uncaughtException, runs the handler again, and so on forever.

Move the handlers into `fatal-handlers.ts` and add three guards:
- 'error' listeners on stdout/stderr exit cleanly when stdio is broken, before the failure becomes another uncaughtException
- try/catch around `stderr.write` so a sync write failure exits instead of escaping
- try/catch around the formatter so a throwing `.stack` getter or `toString` (the production trace pointed at defaultPrepareStackTrace) can't take down the handler

Idempotent: a second `installFatalHandlers` call is a no-op.


pFornagiel (approved):
Not fond of attaching listeners to stdio or stderr, but this seems fine given the MCP will not be functional on such error anyways. We may wish to have some logging when this happens, though, before we do `exit(0)` - to be able to verify what went wrong if something does unexpectedly break.

PR #197, opened 2026-05-11
feat(describe): render tree as text to cut token cost

filip131311:
## Summary
- Move the JSON tree out of the describe tool's response and add an in-between layer (`format-tree.ts`) that converts it to a sectioned text rendering. On a typical iOS screen the new `description` field is ~6× cheaper on the wire than the previous nested tree once braces, keys, and indentation are counted.
- Public response shape becomes `{ description, source, should_restart? }`. All existing pruning (Android v2 trimmer in `uiautomator-parser.ts`, iOS frame clamp in `ios-ax-adapter.ts`) is untouched — the new layer only re-presents what the platform adapters already produced.
- Header now spells out the coordinate space and the tap-point formula so an agent can act on a frame without further docs.

## How it renders
Two modes picked automatically from the tree shape:
- **flat** (iOS ax-service): bucket children by `frame.y` into Top bar / Body / Composer / Bottom; split the bottom zone into keyboard rows by rounded y and classify each row (letters / shift / return / globe-dictate).
- **nested** (Android uiautomator, iOS native-devtools fallback): iterative DFS with depth indentation, preserving `clickable` / `scrollable` / `scrollHidden` / `disabled` / `password` flags inline.

Example (Messages with keyboard up):
```
Source: ax-service
Mode: flat
Coordinates are normalized [0,1] fractions of the screen (x, y, width, height), not pixels — pass them straight to gesture-tap / gesture-swipe / gesture-pinch, which expect this same space. To tap an element, use its centre: tap_x = frame.x + frame.width / 2, tap_y = frame.y + frame.height / 2.

— Top bar —
  AXButton     "Messages"  (0.040, 0.071, 0.109, 0.050)
  AXButton     "+1 (888) 555-1212"  (0.267, 0.134, 0.466, 0.037)

— Body / content —
  AXStaticText "iMessage ￼ Encrypted"  (0.040, 0.156, 0.920, 0.030)
  ...

— Bottom row 1: letters —
  AXImage      "Q"  (0.012, 0.675, 0.098, 0.062)
  ...
```

## Files
- `packages/tool-server/src/tools/describe/contract.ts` — split into `DescribeTreeData` (internal) and `DescribeResult` (public)
- `packages/tool-server/src/tools/describe/index.ts` — `withDescription(data)` wrapper applied to both platform handlers
- `packages/tool-server/src/tools/describe/format-tree.ts` — the renderer (new)
- `packages/tool-server/src/tools/describe/platforms/{ios,android}/index.ts` — return `DescribeTreeData`
- `packages/tool-server/test/describe-tool.test.ts` — assertions migrated to `result.description` / `result.source` / `result.should_restart`
- `packages/tool-server/test/describe-format-tree.test.ts` — renderer unit tests (new): flat zoning, nested indentation + flags, embedded-newline escaping, empty tree

## Test plan
- [x] `npm test` in `packages/tool-server` — 49 files / 482 tests passing locally
- [x] `tsc --build` at workspace root — clean
- [x] Live MCP sanity check: `argent run describe --udid <booted> --json` returns the new shape; tested against Messages (full keyboard surface), Safari (chrome-only, expected), and a minimal in-process WKWebView host. Existing pruning behaviour unchanged in every case.

### DIFF
--- a/packages/tool-server/src/tools/describe/index.ts
+++ b/packages/tool-server/src/tools/describe/index.ts
@@ -1,9 +1,24 @@
 import { z } from "zod";
 import type { Registry, ToolCapability, ToolDefinition } from "@argent/registry";
-import type { DescribeResult } from "./contract";
+import type { DescribeResult, DescribeTreeData } from "./contract";
 import { dispatchByPlatform } from "../../utils/cross-platform-tool";
 import { describeAndroid, androidRequires } from "./platforms/android";
 import { iosRequires, describeIos } from "./platforms/ios";
+import { formatDescribeTree } from "./format-tree";
+
+// In-between layer between the per-platform adapters (which still own all
+// pruning — the Android v2 trimmer in uiautomator-parser stays untouched) and
+// the public DescribeResult. The internal `tree` is converted to a token-
+// efficient text rendering here and then dropped, so the caller (LLM) never
+// pays for the JSON tree.
+function withDescription(data: DescribeTreeData): DescribeResult {
+  const out: DescribeResult = {
+    description: formatDescribeTree(data.tree, { source: data.source }),
+    source: data.source,
+  };
+  if (data.should_restart) out.should_restart = data.should_restart;
+  return out;
+}
 
 const zodSchema = z.object({
   udid: z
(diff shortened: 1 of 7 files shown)


latekvo:
What criteria is used for grouping interactible elements by the section they belong to (e.g. `Top bar`, `Body / content`, `Bottom row 1: letters`?

This could really explode the output unnecessarily, for example if you had a list, would every such list element be recognized as a section heading, as is the case with `Bottom row 1: letters` in the example you provided in the description? Or would it be flattened into a single `Body / content`?

What about nested elements - list with two elements that would classify as a section heading?

packages/tool-server/src/tools/describe/index.ts
latekvo, line 1: Stale description, still says, it still says `"Returns a JSON tree of UI elements"` and `"Use frame.x + frame.width/2 as the tap X coordinate"` which is wrong.

latekvo:
Apologies for the force push. I have an **identical** PR open ([link](https://github.com/software-mansion/argent/pull/157)), thought it was the one i was working with.

PR #198, opened 2026-05-11
feat: per-step delayMs on flow steps

latekvo:
- Tool steps in a flow YAML may carry an optional `delayMs` that sleeps that long *before* the step runs; `flow-add-step` exposes the same field.
- New `longRunning?: boolean` field on `ToolDefinition`; when set, the MCP fetch wrapper bypasses its 30s per-request timeout (`fetchTimeoutMs: null`) so long invocations aren't aborted. `run-sequence` and `flow-execute` opt in.

### DIFF
--- a/packages/tool-server/src/tools/flows/flow-add-step.ts
+++ b/packages/tool-server/src/tools/flows/flow-add-step.ts
@@ -1,6 +1,6 @@
 import { z } from "zod";
 import type { Registry, ToolDefinition } from "@argent/registry";
-import { getFlowPath, getActiveFlow, appendStep } from "./flow-utils";
+import { getFlowPath, getActiveFlow, appendStep, type FlowStep } from "./flow-utils";
 
 const zodSchema = z.object({
   command: z.string().describe('MCP tool name (e.g. "tap", "screenshot", "launch-app")'),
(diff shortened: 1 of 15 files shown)


packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -10,6 +10,12 @@ const zodSchema = z.object({
     .describe(
       'Tool arguments as a JSON string, e.g. \'{"udid": "ABC", "x": 0.5, "y": 0.3}\'. Omit for tools with no arguments.'
     ),
+  delayMs: z
+    .number()
+    .int()
+    .min(0)
+    .optional()
+    .describe("Milliseconds to sleep before executing this step during replay."),
```
latekvo, line 18: For `run-sequence`, the delay is *after* each tool, as in case of `run-sequence` the agent **anticipates** the required delays for each of the tools.

In case of `flow-add-step`, the delays **have to** be "before executing this step", as the required delay (if there is any necessary) is **observed**, not anticipated .

pFornagiel:
Looks alright, but skill update or tool description in the MCP would be nice for agent

latekvo:
> skill update or tool description

yep, missed that - done in [41a0e21](https://github.com/software-mansion/argent/pull/198/commits/41a0e21695cdf4340e6d593ce2e5eede8e9a4e00)

pFornagiel (approved):

PR #199, opened 2026-05-11
fix(registry): emit JSON Schema type for ZodDefault fields

filip131311:
## Summary

`zodTypeToJsonSchema` in `packages/registry/src/zod-to-json-schema.ts` did not handle `z.ZodDefault`. Any tool input field declared as `z.X().default(...)` fell through to `{}`, dropping the type. Downstream, the MCP harness sent values for those fields as strings, and the tool-server's strict Zod validators rejected them with `Expected number, received string` / `Expected boolean, received string`.

Also fixes a related issue in the same file: `isOptional` returned `false` for `ZodDefault`, so defaulted fields were pushed into `required` even though the caller doesn't have to provide them.

Tools that misbehave in practice today because of this and start working after this fix:
- `stop-metro` — `port`
- `react-profiler-start` — `port`, `force`, `sample_interval_us`
- `debugger-component-tree` — `onScreenOnly`, `includeSkipped`

(Reproduced against the published `argent 0.6.1` MCP server while profiling an Expo app — the `argent run <tool> --args '<json>'` CLI path is unaffected because it doesn't go through the published JSON Schema.)

## Changes

- `packages/registry/src/zod-to-json-schema.ts`: handle `ZodDefault` (recurse into `_def.innerType`, attach `default: defaultValue()`); treat `ZodDefault` as optional.
- `packages/registry/tests/zod-to-json-schema.test.ts`: new file. Covers `z.number()`, `z.boolean()`, `z.string().optional()`, `z.number().default(...)`, `z.boolean().default(...)`, `z.coerce.number().int().positive().default(...)`, and a mixed-object happy path.

## Test plan

- [x] `npx vitest run` in `packages/registry` — 64 passed, 0 failed (4 files; the new test file contributes 7 cases).
- [ ] Reviewer sanity check: load the rebuilt MCP server and confirm a tool with a `.default(...)` numeric/boolean field now publishes the correct \`type\` in its input schema.

### DIFF
--- a/packages/registry/src/zod-to-json-schema.ts
+++ b/packages/registry/src/zod-to-json-schema.ts
@@ -24,6 +24,9 @@ function zodTypeToJsonSchema(type: z.ZodTypeAny): Record<string, unknown> {
   if (type instanceof z.ZodNumber) return { type: "number" };
   if (type instanceof z.ZodBoolean) return { type: "boolean" };
   if (type instanceof z.ZodOptional) return zodTypeToJsonSchema(type.unwrap());
+  if (type instanceof z.ZodDefault) {
+    return { ...zodTypeToJsonSchema(type._def.innerType), default: type._def.defaultValue() };
+  }
   if (type instanceof z.ZodArray) {
     return { type: "array", items: zodTypeToJsonSchema(type.element) };
   }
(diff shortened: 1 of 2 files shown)


pFornagiel (approved):
Tested, works, good enough for me

PR #201, opened 2026-05-12
fix(profiler): verbose errors when target app cannot be profiled

latekvo:
If you point the profiler or inspect-element tool at an app that can't actually be profiled (usually a release build with React DevTools stripped out), the tools currently blow up with a cryptic `Cannot read property 'X' of undefined` and leave the session in a weird half-started state. This PR makes them say what went wrong and what to do about it instead.

Profiler and inspector tools used to crash with generic "Cannot read property 'X' of undefined" TypeErrors whenever they hit a runtime without `__REACT_DEVTOOLS_GLOBAL_HOOK__` (release/production builds where DevTools is stripped) or against a partial/malformed Hermes CPU profile. Surface the diagnosis up front instead, and stop leaving a half-initialised session for downstream tools to trip over.

# Technical details

- `inspect-at-point.ts`: guard the injected script for missing hook / renderer / fiber root and report through the existing `__argent_callback` error channel; export `INSPECT_NO_DEVTOOLS_HOOK_ERROR`, `INSPECT_NO_RENDERER_ERROR`, `INSPECT_NO_FIBER_ROOT_ERROR` so callers can recognise the same diagnosis without string-matching bespoke phrasings.
- `react-profiler-start.ts`: replace the terse "production build" one-liner with the verbose `NO_DEVTOOLS_HOOK_ERROR` (rebuild in dev mode, three likely causes, concrete `run-ios` / Expo dev client hints). Export `NO_RENDERERS_ATTACHED_ERROR` for the distinct "hook present but no renderer registered" state.
- `react-profiler-stop.ts`: refuse to stop when `api.profilingActive` is false (start failed before reaching the sampler), and validate the Hermes profile shape (`samples` / `nodes` / `timeDeltas` arrays) before handing it to the analyse pipeline.
- `dump.ts` (`readCpuProfile`): validate shape inline at load so `profiler-load` of an older / partial dump fails fast with an actionable message instead of crashing `buildCpuSampleIndex` later.
- `react-profiler-renders.ts` / `react-profiler-fiber-tree.ts`: branch on the in-runtime error code so "no `__REACT_DEVTOOLS_GLOBAL_HOOK__`" gets the rebuild-in-dev advice while "no renderers attached to hook" gets the wait-for-render / let-`react-profiler-start`-bootstrap advice — separating two failure modes that the old `HOOK_MISSING_MESSAGE` collapsed into one misleading message.

### Tests

- New `inspect-at-point-guards.test.ts` — eval the injected IIFE against a controlled `globalThis` to verify each guard fires the verbose diagnostic and routes through `__argent_callback`.


hubgan (approved):
Looks good. Just make sure we don’t reintroduce the bug fixed [here](https://github.com/software-mansion/argent/pull/317) during the rebase.

PR #202, opened 2026-05-13
ci: typecheck test files in CI

latekvo:
We've had lots of type errors in our tests, scripts and other non-core code despite all CI being green. This PR fixes that.

This PR also removes some redundant features like `@ts-check` inline directives which have been replaced by proper ts config files.

<details>

## AI Summary

Audit of every CI workflow against every package's `package.json` and `tsconfig` surfaced several gaps where CI did partial coverage. This PR closes them and fixes every pre-existing type error the new gates surfaced.

## Gaps closed

| # | Gap | Fix |
|---|---|---|
| 1 | `tsc --build` only covers `src/**`; test files (`test/**`, `tests/**`) never typecheck. Vitest transforms tests with esbuild — strips types instead of checking them. | New `tsconfig.test.json` per test-bearing package (`composite: false`, `noEmit`, `rootDir: "."`, includes `src/**` + `test/**` (or `tests/**`) + `vitest.config.ts`). New `typecheck:tests` npm script per package. New CI step `npm run typecheck:tests --workspaces --if-present`. |
| 2 | Build/publish scripts (`scripts/*.{cjs,mjs}`, `packages/argent/scripts/*.cjs`, `packages/skills/scripts/install.js`) — JS, never typechecked. `sync-readme.cjs` runs in `prepack`; `postinstall.cjs` runs on every install; `bundle-tools.cjs` is build-critical. A typo only surfaces at runtime. | Root `tsconfig.scripts.json` with `allowJs` + `checkJs` + `noImplicitReturns` covers all script paths. Root `typecheck:scripts` npm script. New CI step. Per-file `// @ts-check` comments removed — tsconfig drives inclusion. |
| 3 | `argent-cli` had no `typecheck:tests` script. When tests get added there they'd silently bypass the gate. | Added `tsconfig.test.json` + `typecheck:tests` script. `--if-present` would have skipped it anyway, so this is future-proofing. |
| 4 | Local `prettier --check .` walked into the `argent-private` submodule and reported violations there. CI didn't see this because `actions/checkout@v4` doesn't init submodules — so local `prettier --check` did not match CI behavior. | New `.prettierignore` excludes the submodule plus `dist/`, `node_modules/`, lock files, `tsbuildinfo`. No-op in CI; aligns local with remote. |

## Test type errors fixed

These were already present. The `npm test` passed because esbuild stripped the offending types:

- **argent-installer**: widen `scope` literal type for dead-code-elimination guard; non-null `addAllowlist!`/`removeAllowlist!` (optional methods on adapter).
- **argent-mcp**: add `.js` extensions for NodeNext module resolution.
- **registry**: broaden `StaticBlueprintResult` blueprint api from `{ id; deps? }` to `Record<string, unknown>` so existing tests can pass arbitrary api shapes.
- **tool-server**:
  - non-null `zodSchema!` on tool defs that always carry one (`launchAppTool`, `restartAppTool`, etc.)
  - cast sentinel `"ignored"` strings to `DeviceInfo` in factory-rejection tests
  - match the 4-arg `dispatchByPlatform<IosServices, AndroidServices, Params, Result>` signature (tests still passed the old 3-arg shape)
  - `assertFlowRunResult` type guard for the `FlowRunResult | FlowPrerequisiteNotice` union before `.steps` access
  - swap broken `typeof import("supertest").default` for static `import supertest` (supertest is `export = supertest`)
  - fill in the 4 fields missing from a `NativeProfilerSessionApi` mock (`xctraceProcess`, `recordingTimedOut`, `recordingExitedUnexpectedly`, `lastExitInfo`)
  - cast a vitest `Mock` to a plain function at one direct call site where the `Procedure | Constructable` union loses callability
  - fix `update-checker.test.ts` reading `../../package.json` (off-by-one — went past tool-server; was relying on a Vite resolver quirk)
  - replace one cross-package `../../../registry/src/index` import with `@argent/registry` so registry's private `services` field isn't compared between src and dist declarations

All fixes are type-level. Full `npm test --workspaces --if-present` still reports **991 passed across 87 test files**.

</details>


filip131311 (approved):
sure

PR #204, opened 2026-05-13
feat(debugger): dismiss-logbox tool + redbox-safe LogBox suppression

pFornagiel:
## Summary

ARG-187 — hide the React Native LogBox banner reliably across the whole RN session without ever dismissing an open fullscreen redbox.

- **Layer 1** — fix a latent bug in `DISABLE_LOGBOX_SCRIPT`: removed the `LBData.clear()` call, which routes through `setSelectedLog(-1)` and would dismiss an open fullscreen redbox. The banner is gated by `_isDisabled` alone, so clearing the data store was never necessary.
- **Layer 2** — new `dismiss-logbox` MCP tool. Resolves the existing `JsRuntimeDebugger:port:device_id` service (auto-connects if needed) and re-evaluates the disable script. `alwaysLoad: true` so the agent can call it before any other debugger tool. Mirrors `debugger-evaluate`'s URN/services shape.
- **Layer 3** — re-inject the disable script on every new JS execution context. Added `executionContextCreated` to the typed `CDPClientEvents` map and emitted it in `handleMessage`. The blueprint subscribes with a debounced two-stage re-eval (1500 ms + 3500 ms) so cmd-R / shake reload / fast-refresh full reload / `DevSettings.reload` / `Page.reload` all recover the disabled state without the agent needing to re-call.
- **Layer 4** — skill prompts updated so the agent learns about the tool atomically with its existence: `argent-device-interact`, `argent-react-native-app-workflow` (new §1.4 + quick-ref row), and `argent-metro-debugger` (tool-table row + prereqs paragraph).

The fullscreen redbox shown for fatal/uncaught errors is preserved end-to-end.

## Note for reviewers

`.claude/rules/argent.md` is gitignored in this repo. The plan's `<tapping_rule>` LogBox callout edit lives only in my local working tree and is not in this PR. Apply it manually to your local `.claude/rules/argent.md` or skip it — the three shipping skill files cover the same guidance.

## Test plan

Automated (`pnpm --filter @argent/tool-server test`):
- [x] 11 new tests pass; full suite still green (72 files / 653 tests).
- [x] `disable-logbox-script.test.ts` — asserts the script calls `ignoreAllLogs(true)`, never `.clear()`, and remains an IIFE.
- [x] `dismiss-logbox.test.ts` — id, `alwaysLoad`, default port 8081, required `device_id`, URN resolution.
- [x] `cdp-execution-context-event.test.ts` — typed `executionContextCreated` event fires + generic `event` still fires (no regression).
- [x] `tsc --noEmit` clean.

Manual verification (Bluesky / iPhone 17 Pro / iOS 26.2 / Metro :8082):
- [x] **V0 premise** — `DevSettings.reload` wipes a `globalThis` marker and flips `LBData.isDisabled()` from `true` → `false`. Confirms the JS-context-reset condition Layer 3's listener recovers from.
- [x] **Layer 1 fix** — fixed disable script (no `.clear()`) leaves an open fullscreen redbox visible; explicit `LBData.clear()` dismisses it (proves the bug existed and the fix addresses it).
- [x] **Banner suppression** — `addLog({level:'error',…})` produces the bottom banner with an enabled LogBox; fixed disable script hides it; subsequent `addLog` while disabled produces no banner.

Not directly observed via MCP (would require Node-side CDP subscription / MCP server rebuild):
- [ ] `Runtime.executionContextCreated` actually firing in this RN runtime — covered by the unit test against a mock WS loop, premise verified via state-reset proxy, ship-as-designed per the plan's "if it doesn't fire on this RN version, Layer 3 is a no-op but does no harm."
- [ ] `dismiss-logbox` end-to-end through the MCP transport — the underlying `cdp.evaluate(DISABLE_LOGBOX_SCRIPT)` path was exercised directly via `debugger-evaluate`.

### DIFF
--- a/packages/skills/skills/argent-device-interact/SKILL.md
+++ b/packages/skills/skills/argent-device-interact/SKILL.md
@@ -108,7 +108,7 @@ Read the exact error and choose the action that matches it:
 
 Coordinates: `0.0` = left/top, `1.0` = right/bottom.
 
-Before tapping near the bottom of the screen in React Native apps, check that "Open Debugger to View Warnings" banners are not visible — tapping them breaks the debugger connection. Close them with the X icon if present.
+When working with a React Native app, call the `dismiss-logbox` tool once near the start of your interaction session — this hides the bottom-screen LogBox banner (yellow warnings + red non-fatal errors) and keeps it hidden across JS reloads, without affecting the fullscreen redbox for fatal errors. **Never tap the banner or its X icon** — accidental hits open the fullscreen debugger overlay. If `describe` still shows an `AXGroup` near the bottom of the screen with a label that starts with `"! "`, call `dismiss-logbox` again (it's idempotent).
 
 ### gesture-swipe — Straight-line gesture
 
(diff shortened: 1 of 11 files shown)


filip131311:
Old draft @pFornagiel you can open it back when you come back

PR #229, opened 2026-05-19
feat(tools): implement screenshot-diff tool

hubgan:
## Summary

Adds a new `screenshot-diff` tool for comparing mobile screenshots during UI QA and visual regression checks. The tool can compare two saved PNGs, or compare one saved PNG against one live full-resolution device capture, then returns a compact text summary and generated diff artifacts.

## Context

Mobile UI validation often needs more than a single screenshot: small layout, color, typography,
copy, clipping, and regression changes are easy to miss manually. This feature gives agents a
deterministic before/after comparison path while keeping `screenshot-diff` as supporting evidence.

## What Changed

- Registered a new `screenshot-diff` tool in the tool server.
- Supports saved PNG comparisons and one-sided live capture via `captureBaseline` or
`captureCurrent`.
- Adds PNG diffing with fixed internal defaults, changed-region detection, status-bar-band
ignoring, full-size diff output, and downscaled context diff output.
- Adds optional OCR/text analysis for moved, appeared, disappeared, changed, and restyled text.
- Updates MCP content rendering to attach the downscaled context diff image plus the text summary.

1. Capture a stable full-resolution baseline.
2. Perform an interaction or apply a UI change.
3. Compare the current screen against the baseline.
4. Review a summarized diff plus generated artifacts.


pFornagiel (commented):
I think I would prefer moving the screenshot diff explanation from `agant-device-interact` skill to some other skill or a reference file. Anthropic also suggests such granuality:

https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices

pFornagiel (commented):
When it comes to `agent-test-ui-flow` the splitting is less apparent, but maybe we could also add QA as additional resource instead of adding more information to the exitsing skill to avoid context bloat? This requires testing.

packages/tool-server/src/tools/screenshot-diff/index.ts
```diff
@@ -0,0 +1,195 @@
...
+  outputHint: "screenshot-diff",
+  capability,
+  services: (params): Record<string, ServiceRef> => {
+    if ((params.captureBaseline || params.captureCurrent) && params.udid) {
+      return {
+        simulatorServer: simulatorServerRef(resolveDevice(params.udid)),
+      };
+    }
+    return {};
+  },
```
pFornagiel: NIT: Not a fan of conditional services object, but it seems like this should not cause issues. We could also simply enforce the sim-server even if we are doing a screen-diffing on saved image. In the standard flow, the user will need the simulator-server running one way or another, so adding it as dependency unconditionally should not matter much.

packages/skills/rules/argent.md
```diff
@@ -57,6 +59,12 @@ Decision order:
...
   Call `screenshot` separately only for a baseline before any action or after a delay.
+- Use normal downscaled screenshots for UI context. Do not treat `screenshot { scale: 1.0 }`
+  as a general readability or tapping aid. Use `scale: 1.0` with `includeImageInContext: false`
+  only when saving baseline/current PNGs for visual regression, then compare with `screenshot-diff`.
+- For explicit visual-regression, screenshot-diff, or before/after comparison requests, use `screenshot-diff` unless no stable comparable screenshots can be produced;
+- For implementation tasks that change visible mobile UI, consider `screenshot-diff` as supporting visual evidence when the affected screen has stable before/after states and the expected result is pixel-visible. Skip it when structural checks, logs, network evidence, or tests answer the question better, or when dynamic content makes comparison noisy.
+- For `screenshot-diff`, provide `udid` and exactly one input for each side. Common flow: saved `baselinePath` plus live `captureCurrent: true`. Never set both `captureBaseline` and `captureCurrent` in one call.
```
pFornagiel: Saying "provide `udid`" and likewise are redundant, given that this is a parameter that is already surfaced to the agent via the tool description inside `screenshot-diff/index.ts`.

packages/skills/rules/argent.md
```diff
@@ -57,6 +59,12 @@ Decision order:
 - Before calling any gesture tool for the first time, use ToolSearch to load its schema.
 - Interaction tools (`gesture-tap`, `gesture-swipe`, `gesture-pinch`, `gesture-rotate`, `gesture-custom`, `launch-app`, etc.) return a screenshot automatically.
   Call `screenshot` separately only for a baseline before any action or after a delay.
+- Use normal downscaled screenshots for UI context. Do not treat `screenshot { scale: 1.0 }`
+  as a general readability or tapping aid. Use `scale: 1.0` with `includeImageInContext: false`
+  only when saving baseline/current PNGs for visual regression, then compare with `screenshot-diff`.
+- For explicit visual-regression, screenshot-diff, or before/after comparison requests, use `screenshot-diff` unless no stable comparable screenshots can be produced;
+- For implementation tasks that change visible mobile UI, consider `screenshot-diff` as supporting visual evidence when the affected screen has stable before/after states and the expected result is pixel-visible. Skip it when structural checks, logs, network evidence, or tests answer the question better, or when dynamic content makes comparison noisy.
```
pFornagiel: NIT: The screenshot caveat is fine, but poitns below seem like they could shortened and added it to the `skill_routing` category. If agent behaves right we can keep it here, but we should consider not bloating the context window too much.

packages/tool-server/src/tools/screenshot-diff/screenshot-diff-ocr.ts
```diff
@@ -0,0 +1,404 @@
...
+      blocks: parseTesseractTsv(stdout),
+    };
+  } catch {
+    return {
+      status: "unavailable",
+      provider: "tesseract",
+      blocks: [],
+    };
+  }
+}
```
pFornagiel, line 90: This one seems to be a blocker - we do not provide tesseract binary anyhow as far as I can see. The user trying to use this tool, without tesseract on the system, will just see tool failures. 

There is no mention on additional dependency in the readme, tool description or wherever, users will consistently face tool failures. We should provide a packed binary ourselves using our CI build system (releases from `argent-private`) or, for example, use some npm package which has that dependency built-in. 

Unfortunately, this is a blocker in my opinion.

pFornagiel (changes requested):

packages/tool-server/src/tools/screenshot-diff/screenshot-diff.ts
```diff
@@ -0,0 +1,829 @@
...
+          regions.splice(j, 1);
+          joinedThisPass = true;
+          break outer;
+        }
+      }
+    }
+  }
+
+  return regions;
+}
```
pFornagiel, line 715: How many regions can we have given worst-case scenario? If we take `n` as the number of regions this loop runs in `O(n^3)`. Have you verified that we are fine with it and this wont potentially block amid tool call?

packages/skills/skills/argent-device-interact/SKILL.md
```diff
@@ -48,23 +48,24 @@ Common schemes: `messages://`, `settings://`, `maps://?q=<query>`, `tel://<numbe
...
+| Drag & drop       | `gesture-custom`  | Complex drag interactions                                                        |
+| Pinch/zoom        | `gesture-pinch`   | Two-finger pinch with auto-interpolation                                         |
+| Rotation          | `gesture-rotate`  | Two-finger rotation with auto-interpolation                                      |
+| Custom gesture    | `gesture-custom`  | Arbitrary touch sequences, optional interpolation                                |
+| Hardware key      | `button`          | Home, back, power, volume, appSwitch, actionButton                               |
+| Type text (fast)  | `paste`           | iOS only. Form fields — uses clipboard                                           |
+| Type text         | `keyboard`        | iOS+Android. Fallback when paste fails; supports Enter, Escape, arrows           |
+| Rotate device     | `rotate`          | Orientation changes                                                              |
+| Visual comparison | `screenshot-diff` | Compare saved baseline/current PNGs, or saved baseline plus live current capture |
```
filip131311: this is literaly not device interaction

packages/skills/skills/argent-test-ui-flow/SKILL.md
```diff
@@ -20,14 +20,24 @@ Once a platform is chosen, the per-platform setup skill takes over:
 
 All interactions go through argent MCP tools. Ensure the simulator/emulator is ready before starting.
 
-1. **Baseline screenshot**: Call `screenshot` to see the current UI state.
+For implementation tasks that modify visible UI, this workflow can also serve as a visual acceptance path.
+
+### QA preflight
+
+Before running a QA flow, classify each expected result as visual, structural, runtime/log/network, or mixed. For detailed evidence selection, read `references/qa-evidence.md`. For exact screenshot-diff use cases, parameter choices, and full-resolution screenshot guidance, use the `argent-screenshot-diff` skill.
```
latekvo: There isn't a good reason to have a non-conditional* "read this" inside a skill, this sort of nesting does not optimize any part of the skill while requiring an additional action (read) from the agent.

Also the term `QA` is new in context of `flow`, it seems hallucinated. 

*(non-conditional because `"QA flow"` seem to apply to every usage of `flow`, right?)

packages/tool-server/src/tools/screenshot-diff/index.ts
```diff
@@ -0,0 +1,184 @@
...
+  const currentPath = params.captureCurrent
+    ? await captureLiveInput({
+        api: services.simulatorServer as SimulatorServerApi,
+        outputDir: params.outputDir,
+        name: "current",
+        rotation: params.rotation,
+        signal: options?.signal,
+        captureScreenshot,
+      })
+    : params.currentPath!;
```
latekvo, line 142: Do i understand correctly that these files are saved as `/tmp/baseline.png` and `/tmp/current.png`?
If that's the case, we want to at least append a hash like `/tmp/baseline-0f75d.png`, so that the agent doesn't overwrite past comparisons in case it wants to check them. This is also done to avoid potential naming collisions, since these two names are very likely to collide.

packages/tool-server/src/tools/screenshot-diff/screenshot-diff-summary.ts
latekvo, line 1: Not sure if this is something we can trivially fix, but both the old and the new images have the same color in the comparison, making it unclear which one is the baseline and which one is the current result. 

If this is a complex fix, let's leave it as-is, but if it's trivial, would be nice to have both the image layers be distinct.

packages/tool-server/src/tools/screenshot/index.ts
```diff
@@ -16,7 +16,14 @@ const zodSchema = z.object({
...
     .describe(
-      "Scale factor (0.01-1.0). Defaults to ARGENT_SCREENSHOT_SCALE env var, or 0.3 if unset. Use 1.0 for full resolution."
+      "Scale factor (0.01-1.0). Defaults to ARGENT_SCREENSHOT_SCALE env var, or 0.3 if unset. Use 1.0 only when saving full-resolution PNG artifacts."
+    ),
+  includeImageInContext: z
+    .boolean()
+    .optional()
+    .default(true)
+    .describe(
+      "Whether the MCP adapter should attach the screenshot image to the agent context. Set false to save the file path only."
```
latekvo: Is there *any* istuation where we would want this to be `false` for any reason?

I could've missed something, but if the answer is no then adding arguments with no purpose really clutters up the context and wastes model's [attention](https://www.3blue1brown.com/lessons/attention/).

packages/tool-server/package.json
```diff
@@ -13,12 +13,15 @@
     "typecheck:tests": "tsc --noEmit -p tsconfig.test.json"
   },
   "dependencies": {
-    "@argent/registry": "file:../registry",
     "@argent/native-devtools-ios": "file:../native-devtools-ios",
+    "@argent/registry": "file:../registry",
     "@clack/prompts": "^1.1.0",
+    "@tesseract.js-data/eng": "^1.0.0",
```
latekvo: Just wanted to confirm - is there any place where we're using this OCR? This is potentially a very heavy download (100s of MBs) IIRC.

packages/tool-server/src/tools/screenshot/index.ts
```diff
@@ -25,6 +32,7 @@ type Params = z.infer<typeof zodSchema>;
 interface Result {
   url: string;
   path: string;
+  includeImageInContext: boolean;
```
latekvo: Why are we adding `includeImageInContext` to the output of this function? It's already in the input.

latekvo (changes requested):
Good job, i only focused on places where this tool affects existing logic etc, did not go over the algorithm in depth.

packages/tool-server/package.json
filip131311: let's just remove it if i understan correctly it is here because it was not always on the system argent is in but it works fine without it we can just rertur results without it if it is not present (with no errors)

packages/tool-server/src/tools/screenshot/index.ts
filip131311: should be removed screenshots are important tool we need to give the agent as little changes of halucenating args here as posible

packages/argent-mcp/src/content.ts
```diff
@@ -4,15 +4,43 @@
...
+      blocks.push({
+        type: "image" as const,
+        data: buf.toString("base64"),
+        mimeType: "image/png" as const,
+      });
+    }
+
+    blocks.push({ type: "text" as const, text: result.summary });
+    return blocks;
+  }
```
latekvo: `toMcpContent` is a very low level abstraction, we should not be cramming tool-specific sphaghetti into it please.

packages/skills/skills/argent-test-ui-flow/SKILL.md
hubgan: I removed the unconditional "QA preflight"

packages/tool-server/src/tools/screenshot-diff/index.ts
hubgan: I added suffix to avoid naming collisions

packages/tool-server/src/tools/screenshot-diff/screenshot-diff-summary.ts
hubgan: I changed the diff rendering so changed pixels are green when the current image is brighter and red when current image is darker. It is not perfect, but I don't think we can actually trivially say which one is the baseline and which one is the current result.

packages/tool-server/src/tools/screenshot/index.ts
hubgan: Yes, when capturing a full-resolution PNG purely as a screenshot-diff baseline, we don't want to attach large files to the agent context. Kept the parameter and tightened its description to make that the only documented use case.

packages/tool-server/package.json
hubgan: We need to choose one of two behaviors. Either we keep the OCR dependencies so it is expect to always work, or we make OCR optional and allow users without Tesseract to skip OCR text analysis gracefully.

packages/tool-server/src/tools/screenshot/index.ts
hubgan: I removed it from `Result` interface.

5.png
latekvo, line 1: I think this is a testing artifact.

6-context-diff.png
latekvo, line 1: Same

packages/argent-mcp/src/content.ts
latekvo: Reference to how we have these abstractions implemented in Radon: https://github.com/software-mansion/radon/blob/main/packages/vscode-extension/src/ai/mcp/toolOutputConverters.ts

There they are extremely simple, with their only role being conversion of image/text data to MCP compatible format.

packages/tool-server/src/tools/screenshot-diff/screenshot-diff.ts
hubgan: I optimized the function a bit [here](https://github.com/software-mansion/argent/pull/229/changes/f473d076dd73f0a24cc30f954b4aa871b70d68b5). I also [capped the regions at 5,000](https://github.com/software-mansion/argent/pull/229/changes/46d1dada8b6a5dc28af441241ffe0182f743ef81), since anything above that can be considered a noisy image. The tests I performed on images with about 4,900 regions finish in around 150 ms.

pFornagiel (commented):
Just make sure to resolve the current merge conflicts, other than that - looks good to me overall, thank you for implementing this 👍

filip131311 (approved):
thank you

pFornagiel (approved):

latekvo (approved):
Looks alright

PR #230, opened 2026-05-19
Fetch argent skills from GitHub for portable lockfiles

### LINKED ISSUE #208: The `argent init` skills installation breaks skills.lock files checked into repos
The `argent init` script seems to go and install the skills with the `skills` CLI from the node_modules where argent is installed. In a `skills-lock.json` file that is shared into a repo, this breaks it for everyone but the person who sets it up because the lockfile saves the fully qualified path, i.e. `My/Specific/User/Path/` which obviously doesn't work for everyone else.

I would prefer that the setup script instead installed the skills straight from the repo, i.e. `skills add software-mansion/argent` which works fine regardless of whether you're installing `argent` and the skills at the repo level or user level.

pFornagiel:
### Description

Install and sync argent skills from a GitHub-pinned source (`software-mansion/argent/packages/skills/skills#v<version>`) instead of the local bundle when possible, so `skills-lock.json` stays portable across machines. Falls back to the bundled skills directory when offline or the version is unknown.

### Fixes #208

### DIFF
--- a/packages/argent-installer/src/init.ts
+++ b/packages/argent-installer/src/init.ts
@@ -14,6 +14,7 @@ import {
   SKILLS_DIR,
   RULES_DIR,
   AGENTS_DIR,
+  buildArgentSkillsSource,
   getInstalledVersion,
   getLatestVersion,
   isGloballyInstalled,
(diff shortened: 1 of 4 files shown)


filip131311 (approved):
good change :thank-you"

PR #233, opened 2026-05-19
fix: stop simctl spawn loop on iOS init failure

### LINKED ISSUE #210: [0.7.1] cpu at 100%, endless amount of `simctl` launched
Hi guys,

using `0.7.1`, I left my mac alone for a while and when i came back i saw about ~500 simctl instances eating up the cpu. 

I asked codex what is spawning it and responded: 

```
The process names point pretty strongly at the culprit already:
  argent mcp is running, and the simctl spawn ...
  DYLD_INSERT_LIBRARIES ... @swmansion/argent/... commands are
  Argent’s iOS simulator instrumentation. I’m doing one parent-
  process check now to make that airtight
```

No matter if I force quitted the processes they kept spawning. The only solution was to kill the argent mcp server.

pFornagiel:
Fixes #210 — `argent mcp` pinning CPU to 100% and spawning hundreds of `simctl` processes that survive `kill -9`.

## The problems

1. **No retry cap.** The simulator watcher polls every 10s; on any failure it forgets the UDID and re-classifies it as "new" on the next tick. Each retry runs 4–5 `xcrun simctl spawn` calls in series. A persistently failing simulator produces an unbounded loop — hundreds of `entered ERROR state` log lines and a steady process pile-up.

2. **No timeouts on `simctl spawn`.** A degraded `CoreSimulatorService` makes simctl block forever. The factory `await`s the hung call, never rejects, never retries, but the stuck child process stays alive.

## How this fixes it

- **Bounded retries with terminal "given up" state.** Each `NativeDevtools` service instance now tracks `{attempts, lastError, givenUp}`. After 3 consecutive init failures the service stops trying and reports `init_failed` to any caller. The watcher keeps polling but skips UDIDs in the given-up state.

- **10s timeout on every `simctl spawn` call** in `ensureEnv`, `ensureAccessibilityEnabled`, and `ensureAutomationEnabled`. Healthy CSS responds in ~300ms; 10s is well above any plausible legit latency and well below "hung forever." A hung simctl now rejects cleanly, increments the attempt counter, and feeds the bounded retry instead of accumulating children.

- **Tools surface `init_failed` as a typed result** instead of throwing. `launch-app`, `boot-device`, `native-devtools-status`, and the rest of the native tools now return a structured error with the failure reason and a hint to restart the simulator, instead of throwing an opaque exception.

- **Factory no longer throws on first env-init failure.** Previously a single failed `simctl` call during factory creation leaked the socket server. The factory now tolerates init failure; the watcher's bounded retry handles the recovery.

Verified on live simulator and against test simulating `simctl` timeouts. Happy path works as before, `restart-needed` is still signalled and the new failure path works as well.


filip131311 (approved):
nice

PR #235, opened 2026-05-19
fix(init): prefer local devDep bundle paths over running module's

filip131311:
## Summary

Fixes a pre-existing leak that surfaces when `argent init` is run via `npx @swmansion/argent` (the default install recipe in the README). The module-relative `SKILLS_DIR` / `RULES_DIR` / `AGENTS_DIR` resolve to the npx cache directory in that case:

```
~/.npm/_npx/008dffcbe9d304e1/node_modules/@swmansion/argent/skills
```

When init shells out to `npx skills add <SKILLS_DIR>`, the skills CLI records that path verbatim in `skills-lock.json`:

```json
{
  "skills": {
    "argent-android-emulator-setup": {
      "source": "/Users/me/.npm/_npx/008dffcbe9d304e1/node_modules/@swmansion/argent/skills",
      "sourceType": "local",
      ...
    }
  }
}
```

That path is per-developer, per-npx-invocation, and gets evicted from the cache over time. The lock file becomes broken on the original machine (eventually) and is broken from day one for any teammate who picks up a committed copy.

## Fix

New helper `resolveProjectBundledDir(projectRoot, kind)` in `utils.ts`:

- Returns `<projectRoot>/node_modules/@swmansion/argent/<kind>` when that directory exists.
- Falls back to the module-relative `SKILLS_DIR` / `RULES_DIR` / `AGENTS_DIR` otherwise.

init's Step 2 (skills) and Step 3 (rules & agents) now use this helper instead of the module-relative constants. The skills CLI sees a stable, version-pinned project-local path; teammates' `npm install` provisions the same path on every checkout.

The check is intentionally loose — no `package.json` dep declaration required. The running init is itself argent code; finding bundled assets at the project-local path is the strongest possible signal that they should be used. That happens whenever a teammate has argent listed as a dependency (workspace member, regular devDep, or a manual `npm i -D @swmansion/argent`), which is exactly the case where the bug matters.

## Why this isn't part of the team-share PR

The bug exists for **any** `npx`-driven init, not just the `--devdep` flow. Splitting it out so it can land independently:

- A team using the existing `--scope local` (without `--devdep`) can ship this fix without waiting on the team-share PR (#232).
- The team-share PR stays focused on the user-facing feature it advertises.

## Tests

`@argent/installer`: 237 passed (+6 new):

- `resolveProjectBundledDir` returns the local devDep path when present (for `skills` / `rules` / `agents`).
- Returns the module-relative fallback when no local devDep exists.
- Partial-bundle case (only one of the three dirs locally) resolves per-kind correctly.
- Defensive guard: a regular file at the expected path falls back rather than mis-reporting.

## Test plan

- [ ] Run `npx ./swmansion-argent-X.Y.Z.tgz init` in a project that has `@swmansion/argent` declared as a devDep with `node_modules/@swmansion/argent/skills` populated. Verify the resulting `skills-lock.json` `source` field contains `<projectRoot>/node_modules/@swmansion/argent/skills`, not `~/.npm/_npx/...`.
- [ ] Run `npx ./swmansion-argent-X.Y.Z.tgz init` in a project that does NOT have argent locally — verify the existing behavior is preserved (`source` points at whatever the running module resolves to; no functional regression for users who run init once and don't commit the lock file).

### DIFF
--- a/packages/argent-installer/src/init.ts
+++ b/packages/argent-installer/src/init.ts
@@ -11,9 +11,6 @@ import {
   type McpConfigAdapter,
 } from "./mcp-configs.js";
 import {
-  SKILLS_DIR,
-  RULES_DIR,
-  AGENTS_DIR,
   getInstalledVersion,
   getLatestVersion,
   isGloballyInstalled,
@@ -23,6 +20,7 @@ import {
   detectPackageManager,
   globalInstallCommand,
   formatShellCommand,
+  resolveProjectBundledDir,
   resolveProjectRoot,
   type ShellCommand,
 } from "./utils.js";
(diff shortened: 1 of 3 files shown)


filip131311:
Superseded by main's #230, which switched argent init's skills source from the module-relative SKILLS_DIR to a GitHub-pinned URL (`software-mansion/argent/packages/skills/skills#v<version>`). That removes the npx-cache leak this PR was created to fix for the common online + known-version case. The standalone fix here is now only relevant for the offline / unknown-version fallback, which is too niche to maintain as a separate code path. Closing.

PR #239, opened 2026-05-20
fix(tool-server): resolve project-local argent before spawning update

filip131311:
## Summary

Fixes a silent ENOENT in the MCP `update-argent` tool that left users on the team-share (`--devdep`) workflow stuck in an infinite "update available" loop.

### The bug

`packages/tool-server/src/tools/system/update-argent.ts` was unconditionally spawning bare `argent` via PATH:

```ts
const child = spawn("argent", ["update", "--yes"], {
  detached: true,
  stdio: "ignore",
});
child.unref();
```

For a user who installed argent as a project devDependency, the binary lives at `<projectRoot>/node_modules/.bin/argent` and isn't on PATH. The spawn fails with ENOENT — but three things made the failure invisible:

1. `detached: true` + `stdio: "ignore"` + `child.unref()` → parent stops observing the child as soon as `spawn()` returns; non-zero exits vanish.
2. No `on('error', ...)` handler → the synchronous ENOENT `'error'` event hit Node's default path with no diagnostic output (because of #1).
3. The tool had already returned `"Argent update initiated..."` to the user **2 seconds before** the spawn fired, so the response confirmed success while the action silently failed.

Net effect: user clicks "update" → confirmation message → nothing happens → next session shows the same notification → repeat forever.

### Fix

`resolveArgentBinary(cwd)`:
- Probes `<cwd>/node_modules/.bin/argent` (and `argent.cmd` on Windows).
- If present, returns its absolute path AND tells the spawn to run with `cwd = projectRoot` so `argent update`'s own project-root logic (lockfile probe, dep declaration check) sees the same root that `argent init` did.
- Falls back to plain `"argent"` PATH lookup otherwise — preserves the historical behavior for globally-installed users.

Plus minimal error surfacing in the spawn site:
- `child.on('error', ...)` writes failures to stderr so a future regression leaves a trace in the tool-server log instead of disappearing.
- The handler also resets the module-level `updateScheduled` flag — without that, a failed spawn would wedge the tool permanently into the "already in progress" branch until the tool-server restarted.

### Scope

Intentionally minimal:
- No walk-up-the-tree to find `node_modules` — `argent init --devdep` commits a literal `./node_modules/.bin/argent` to `.mcp.json`, so a flat probe matches what's actually installed.
- No `package.json` dep-declaration check — the runtime question is "is there a usable binary here right now". The strict `isLocallyInstalled` (PR #232) is the right check for *config* writes; for spawning, "file exists" is sufficient and the error path catches any remaining mistakes.
- Does not address other pre-existing limitations of `argent update` itself — those live in `@argent/installer` and are covered by PR #232.

### Tests

`@argent/tool-server`: 691 passed (+6 new).

| New test | What it covers |
| --- | --- |
| PATH fallback | `argent` (no override, no cwd) when there's no local install |
| local-devDep resolution | absolute path + `cwd: projectRoot` when `node_modules/.bin/argent` exists |
| error handler registration | `child.on('error', ...)` is wired up (regression guard for the silent-failure bug) |
| stderr surfacing | replays an ENOENT and asserts stderr writes |
| updateScheduled flag reset | replay ENOENT, verify the next call doesn't get the "already in progress" reject |
| resolveArgentBinary unit | direct contract test for the resolver |

## Test plan

- [ ] Manual: in a project with `@swmansion/argent` as a devDep but no global install, trigger the MCP update notification, accept it, and confirm `argent update` actually fires (check that the devDep entry in `package.json` bumps).
- [ ] Manual: in a project with only a global install, repeat — confirm the spawn still routes to PATH and behaves identically to today.
- [ ] Manual: in a project with neither install (broken state), trigger the tool, then inspect the tool-server stderr log to confirm the ENOENT is now logged instead of silently swallowed.

## Related

Builds conceptually on PR #232 (team-share devDep flow), but does not depend on it — this fix is self-contained and can land independently against `main`. After both land, the end-to-end "MCP update notification → user accepts → argent updates → MCP configs preserved" flow works for both topologies.

### DIFF
--- a/packages/tool-server/src/tools/system/update-argent.ts
+++ b/packages/tool-server/src/tools/system/update-argent.ts
@@ -1,8 +1,52 @@
 import { spawn } from "node:child_process";
+import * as fs from "node:fs";
+import * as path from "node:path";
 import type { ToolDefinition } from "@argent/registry";
 import { getUpdateState } from "../../utils/update-checker";
 
 let updateScheduled = false;
+
+/**
+ * Resolve which `argent` binary to spawn for the update.
+ *
+ * History: this tool used to call `spawn("argent", ...)` unconditionally,
+ * which does a PATH lookup. That works for globally-installed argent,
+ * but fails silently with ENOENT for users who installed argent as a
+ * project devDependency (the team-share flow) — the binary lives at
+ * `<projectRoot>/node_modules/.bin/argent` and is not on PATH. Combined
+ * with `detached + stdio: "ignore" + unref + no error listener`, the
+ * failure left no trace and the same "update available" notification
+ * kept reappearing every session.
+ *
+ * Resolution order:
+ *   1. `<cwd>/node_modules/.bin/argent` (or `argent.cmd` on Windows) if
+ *      it exists on disk. The MCP server is launched by the editor with
+ *      cwd at the project root, so this is the right place to look.
+ *   2. Plain `"argent"` PATH lookup. Preserves the historical behavior
+ *      for globally-installed users.
+ *
+ * We do NOT walk up the tree looking for a node_modules — keeping it
+ * simple matches what the install side (`argent init --devdep`) commits
+ * to .mcp.json (`./node_modules/.bin/argent`, relative to project root).
+ */
+export function resolveArgentBinary(cwd: string = process.cwd()): {
+  binary: string;
+  spawnCwd?: string;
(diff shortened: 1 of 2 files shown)


filip131311:
Folded into #232 (cherry-picked as commit 34d06a2). Closing as superseded — both fixes will land together in the team-share PR.

PR #240, opened 2026-05-20
fix(describe): surface dialogs hosted by remote-view services

### LINKED ISSUE #207: StoreKit "Sign in to Apple Account" dialog causes Argent to stumble
When the simulator launches the app, I see a prompt to sign in to an apple account on the device. This causes problems because it is overlaying other content as a modal that first needs to be dismissed. Argent struggles because the dialog is not selectable in the way other buttons rendered in our app might be. The agent eventually figures out it can just trigger a press action based on x/y coordinates, but it takes up a ton of time to do so. 

Btw, loving how easy argent is to use. Your install instructions don't mention this, but I've been enjoying modeling it under `devDependencies` and committing the local mcp server configuration into my repo itself. Makes it possible to have whole team sharing their Argent setup.

filip131311:
## Summary

Bumps the \`argent-private\` submodule to pull in the ax-service daemon fix in [software-mansion/argent-private#9](https://github.com/software-mansion/argent-private/pull/9). The radon-lite side of this PR is just the submodule pointer — the actual source change is one if-statement in \`ax_service.m\`.

\`describe\` returns nothing when a \`_UIRemoteViewController\`-hosted overlay is on top of the foreground app — most visibly, the \`StoreKitUISceneService\` Apple ID auth sheet from the reporter's case. Root cause: \`+[AXElement primaryApp]\` returns nil while the remote-view service is frontmost, so the daemon never reaches \`collectExplorerElements\`. The submodule PR adds a single fallback: when \`primaryApp\` is nil, look up the AXElement via \`+[AXElement applicationAtCoordinate:<screen-centre>]\` instead, which returns the *hosting* app's AXElement regardless of which process owns the frontmost view. AX bridges the XPC boundary, so the hosting app's \`explorerElements\` walk surfaces the remote-view's content in the same flat-leaf shape the daemon already emits.

## Why this PR vs. #234

I closed [#234](https://github.com/software-mansion/argent/pull/234) after empirical reproduction showed the TS-side \`systemElements\` fallback there had no effect on the reporter's case — the daemon's \`isSystemAppShowingAnAlert\` gate doesn't fire for XPC-hosted sheets either, so \`systemElements\` was empty too. The reachability gap was at the daemon layer (\`primaryApp\` selector returning nil), not the TS adapter. This PR addresses that root cause.

## Empirical verification

Built the patched daemon + ran against the reporter's exact trigger on an erased iOS 26.5 sim (News → "Get News+" → "Get Started" → "Apple Account Sign in Requested" sheet):

- **Old daemon** (md5 \`d5240e8…\`, currently shipped): raw describe socket returned \`{ "alertVisible": false, "screenFrame": {…} }\` — no \`elements\` field, agent has no labels.
- **Patched daemon** (md5 \`75fdffcc…\`, dev build): 42 elements, all dialog content labelled: heading "Apple Account Sign in Requested", email field, "Forgot password?", **"Sign In" at normalized (0.500, 0.414)**, **"Cancel" at (0.500, 0.482)**, plus 37 keyboard keys.

The agent can now tap by label.

## Risk

- No wire-format change, no TS/adapter change. \`@argent/tool-server\` and the describe handler consume the same shape they did before.
- Fallback gates on \`primaryApp == nil\`, so every working case stays on its existing path.
- If \`applicationAtCoordinate:\` also returns nil at the centre, the daemon emits no \`elements\` — same as current behaviour. The existing native-devtools fallback in \`describeIos\` still kicks in afterwards.

## Open items before un-drafting

- [ ] Merge https://github.com/software-mansion/argent-private/pull/9 (the source change lives there)
- [ ] Publish a new \`argent-main\` build to \`software-mansion-labs/argent-private-releases\` so \`npm run download:native-binaries\` picks up the patched \`bin/ax-service\`
- [ ] Re-verify \`argent run describe\` end-to-end against the same dialog with the released binary
- [ ] One regression-style smoke test: an in-process dialog (e.g. Wallet location prompt) to confirm the existing \`primaryApp\` path still fires for the unchanged cases

## Notes

- Daemon binary lives in \`packages/native-devtools-ios/bin/ax-service\` but is \`.gitignore\`d; the release pipeline supplies it. This PR cannot ship the new binary in-repo for the same reason.
- I empirically introspected the AXRuntime surface inside the simulator before committing to this approach — \`+systemWideElement\` and \`+systemEventDispatchElement\` segfault from a non-UI process, so coordinate lookup is the practical option that doesn't require a daemon-process redesign.

Fixes #207


filip131311:
Rebased the submodule branch onto current `argent-private/main` to pick up [argent-private#7](https://github.com/software-mansion/argent-private/pull/7) (WKWebView deep-walk + dedup). Submodule pointer here advanced from `c3dc463` → `7ec7643`. The two changes touch different regions of `ax_service.m` (#7 refactors `collectExplorerElements` and `collectLeafElements`; this PR adds a `primaryApp == nil` fallback in `handleCommand`'s `describe` branch) so the rebase was clean — no conflicts.

Re-verified the daemon empirically on the same dialog state after the rebase. Element count went from 42 to **82** — same dialog content surfaced (`"Apple Account Sign in Requested"`, `"Forgot password?"`, `"Sign In"`, `"Cancel"`, keyboard keys, etc.) plus the additional WebView-style children #7 now picks up. The fixes compound cleanly: #7 makes `collectExplorerElements` surface more from any given AXElement; this PR makes sure we *have* an AXElement when `primaryApp` is nil.

Rebuilt daemon md5 after rebase: `81fac1b79830a783ef83075b298ace35`.

PR #241, opened 2026-05-20
fix(update): only refresh editors that have argent configured

### LINKED ISSUE #195: `update` adds skills for editors I don't use
When I run `npx @swmansion/argent update` in a repository where I already use Argent it adds skills to editors that I have explicitly unchecked during `npx @swmansion/argent init`.

For example during `init` I explicitly uncheck gemini, cursor etc:

Image

When I run `update` it adds `.gemini` skills because it detects that I have it present on my computer.

Image

which is pretty annoying.

filip131311:
## Summary

- `argent update` was probing the filesystem for editor dirs (`.gemini`, `.cursor`, ...) and writing MCP entries, allowlists, rules, and agents into every detected editor — including ones the user explicitly unchecked during `init`.
- Switch update to read the actual config file for each adapter and only refresh scopes where an argent MCP entry already exists. Editors without one are skipped end-to-end: no MCP write, no allowlist, no rules/agents copy.
- Adds `hasArgentEntry(configPath)` to every adapter and a `findConfiguredAdapterScopes()` helper that drives the new branch.

Closes #195

## User experience & product implications

**Before:** running `npx @swmansion/argent update` could resurrect editors the user had explicitly unchecked during `init`. For example, a Claude Code-only user with `~/.gemini` on disk would find `.gemini/settings.json`, `~/.gemini/rules/`, and `~/.gemini/agents/` populated after every update.

**After:** update only touches editors where argent's MCP entry is already present in that scope's config. The `init` choice is now authoritative across runs.

**Alternatives considered:**
- *Persist the init selection in a side file.* Heavier — adds a new state file, migration story for existing users, and a recovery story when the file is missing. The MCP entry itself is already a perfect record of "the user opted into this editor in this scope," so we use that.
- *Keep `detectAdapters()` but invert the gemini/etc detection rule.* Same root cause (treating editor-dir presence as consent); just shifts the bug to other adapters.

**Backwards compatibility / schema changes:**
- No tool input/output schema changes. The interface change is internal to the installer package.
- New `hasArgentEntry` method on `McpConfigAdapter` is required (not optional). External consumers of `McpConfigAdapter` would need to implement it, but the interface isn't part of the public package API.
- For an existing user who runs `update` *and has never run `init`*, nothing happens to their editor configs (correct: nothing to refresh until they opt in). They'll still get the global package version bump and the skills lockfile refresh.

**Follow-ups this unlocks:**
- Same predicate could power a future `argent status` command that summarizes which editors are configured.
- Opens the door to a `--add-editor <name>` flag on `update` for users who want to opt new editors in without re-running `init`.

## Test plan

- [x] `npm test -w @argent/installer` — 257 tests pass, including 21 new ones covering `hasArgentEntry` per adapter and the `findConfiguredAdapterScopes` semantics.
- [x] `npm run build` — clean tsc build for the whole workspace.
- [x] `npx prettier --check` on touched files.
- [ ] Manual: in a workspace where `init` was run with gemini unchecked but `~/.gemini` exists, run `argent update` and confirm `~/.gemini/settings.json` is not modified and no `~/.gemini/rules/` or `~/.gemini/agents/` directories are created.
- [ ] Manual: in a workspace where `init` configured Cursor + Claude Code, run `argent update` and confirm both still get refreshed (MCP entry, allowlist, rules, agents).
- [ ] Manual: run `argent update` in a workspace with no editor ever configured — confirm it still performs the package upgrade and skills refresh, just skips MCP/rules/agents.


packages/argent-installer/test/mcp-configs.test.ts
```diff
@@ -1313,6 +1314,117 @@ describe("injectCodexRules / removeCodexRules", () => {
...
+
+describe("hasArgentEntry — non-mutating per-adapter predicate", () => {
+  for (const adapter of ALL_ADAPTERS) {
+    describe(adapter.name, () => {
+      it("returns false when the config file does not exist", () => {
+        const configPath =
+          adapter.projectPath(tmpDir) ?? adapter.globalPath() ?? path.join(tmpDir, "nope");
+        // Adapter-provided paths point into tmpDir for project / homedir for
+        // global; the test's only invariant is "file doesn't exist yet".
+        if (fs.existsSync(configPath)) fs.rmSync(configPath, { force: true });
```
latekvo, line 1342: This test will remove the real config when `projectPath` is nullish (multiple such cases)

latekvo (approved):
Solid change, thanks for fixing this. **Please fix the one comment before merging**

PR #243, opened 2026-05-21
fix(describe): wire dialogElements fallback for XPC-hosted prompts (iOS 26+ TCC dialogs)

filip131311:
## Summary

Wires through the new \`dialogElements\` wire field added by [argent-private#10](https://github.com/software-mansion/argent-private/pull/10), so \`describe\` can render iOS 26+ TCC permission prompts (\`Allow X to use your location?\` and friends) instead of returning an empty ROOT.

Three changes:

1. **\`packages/tool-server/src/blueprints/ax-service.ts\`** — declare \`dialogElements?: AXDescribeElement[]\` on \`AXDescribeResponse\` and forward it from \`describe()\`.
2. **\`packages/tool-server/src/tools/describe/platforms/ios/ios-ax-adapter.ts\`** — when the primary \`elements\` walk produces zero usable children, use \`dialogElements\` as a fallback.
3. **\`packages/argent-private\` submodule pointer** — bumped to the branch shipping the daemon change.

Two tests added (Maps-style prompt landing in the fallback, defensive elements-wins precedence).

## Repro before / after

iPhone Air iOS 26.5 simulator, Maps with \`xcrun simctl privacy ... reset location com.apple.Maps\` to force the prompt.

**Before (current main, raw daemon socket):**
\`\`\`json
{ \"alertVisible\": true, \"elements\": [], \"systemElements\": [] }
\`\`\`
→ \`describe\` source=\`ax-service\` returns \`ROOT  AXGroup\` with no children.

**After (this PR + argent-private#10):**
\`\`\`
Source: ax-service
  AXStaticText \"Allow 'Maps' to use your location?\"  (0.177, 0.343, 0.647, 0.048)
  AXStaticText \"Your location is used to show your position on the map, ...\"  (0.177, 0.400, 0.647, 0.089)
  AXButton \"Allow Once\"  (0.142, 0.513, 0.716, 0.055)
  AXButton \"Allow While Using App\"  (0.142, 0.577, 0.716, 0.055)
  AXButton \"Don't Allow\"  (0.142, 0.641, 0.716, 0.055)
\`\`\`

## Why a separate field and adapter fallback (not folded into \`elements\` / \`systemElements\`)

- \`elements\` has a fixed meaning: result of the primary AX walk on the resolved foreground app. The cross-app sweep that catches XPC dialogs is a different lookup; folding it into \`elements\` would conflate two unrelated sources.
- \`systemElements\` is gated on \`isSystemAppShowingAnAlert\` and sourced from Springboard. The dialogElements sweep enumerates *all* running apps. Different gate, different source.
- Keeping \`dialogElements\` separate lets us surface the source distinction in describe output later if useful, and keeps the wire format honest.
- The adapter only consults \`dialogElements\` when \`elements\` is empty, mirroring the daemon's own gating (\`alertVisible\` AND prior walks empty). So in-process \`UIAlertController\` dialogs never get duplicated, and post-merge builds with both gates behave identically.

## Why this isn't the previously-closed PR #234

[#234](https://github.com/software-mansion/argent/pull/234) wired \`systemElements\` (Springboard tree). Empirically that field is also \`[]\` for the Maps-style TCC dialog because the prompt's buttons aren't in Springboard's tree either. The daemon needed a fourth lookup path; that's what argent-private#10 adds. This PR's TS-side wiring is independent of the systemElements question — happy to land both, but they target different dialog classes.

## Companion PR

[**argent-private#10**](https://github.com/software-mansion/argent-private/pull/10) — daemon change in \`ax_service.m\`. This radon-lite PR is **blocked on** that one landing AND the \`build-native-binaries\` workflow republishing the daemon binary. Without those, the daemon never emits \`dialogElements\` and the adapter fallback is a no-op (but harmless).

## Test plan

- [x] \`npx vitest run packages/tool-server/test/describe-ax-adapter.test.ts\` — 100 pass (2 new).
- [ ] After daemon binary republished: trigger Maps location prompt on iOS 26.5 sim, call \`describe\` from the MCP tool, confirm the five elements above appear under \`source: ax-service\`.
- [ ] Confirm a normal app screen (no alert) still returns \`source: ax-service\` with its usual content.
- [ ] Confirm an in-process \`UIAlertController\` dialog still lands in \`elements\`, no duplication.

### DIFF
--- a/packages/tool-server/src/blueprints/ax-service.ts
+++ b/packages/tool-server/src/blueprints/ax-service.ts
@@ -46,6 +46,13 @@ export interface AXDescribeResponse {
   alertVisible: boolean;
   screenFrame?: { width: number; height: number };
   elements: AXDescribeElement[];
+  // Populated by the daemon (post argent-private#10) only when alertVisible
+  // is true and the primaryApp + systemApplication walks were both empty.
+  // Sourced from `[AXElement currentApplicationsIgnoringSiri]` — the last-
+  // resort sweep that catches XPC-hosted dialogs (e.g. iOS 26+ TCC
+  // permission prompts) which live outside primaryApp and Springboard
+  // alike. Empty/missing on daemon builds that predate that change.
+  dialogElements?: AXDescribeElement[];
 }
 
 export interface AXServiceApi {
@@ -276,6 +283,7 @@ export const axServiceBlueprint: ServiceBlueprint<AXServiceApi, DeviceInfo> = {
           alertVisible: result.alertVisible ?? false,
           screenFrame: result.screenFrame,
           elements: result.elements ?? [],
+          dialogElements: result.dialogElements ?? [],
         };
       },
 
--- a/packages/tool-server/src/tools/describe/platforms/ios/ios-ax-adapter.ts
+++ b/packages/tool-server/src/tools/describe/platforms/ios/ios-ax-adapter.ts
@@ -35,9 +35,23 @@ export function adaptAXElement(el: AXDescribeElement): DescribeNode | null {
 }
 
 export function adaptAXDescribeToDescribeResult(response: AXDescribeResponse): DescribeNode {
-  const children = response.elements
+  let children = response.elements
     .map(adaptAXElement)
     .filter((n): n is DescribeNode => n !== null);
+
+  // XPC-hosted dialogs (iOS 26+ TCC permission prompts: "Allow X to use
+  // your location?" etc.) are invisible to every in-app AX walk — neither
+  // primaryApp, applicationAtCoordinate, nor systemApplication reach them.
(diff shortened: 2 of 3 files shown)


filip131311:
Closing in favor of the simpler fix: the iOS 26.5 dialog regression is unlocked by a single `defaults write com.apple.Accessibility IgnoreAXServerEntitlements -bool true` (plus a one-time SpringBoard kickstart). The new fix sits next to the existing `AutomationEnabled` write in `ensureAutomationEnabled`. The `dialogElements` wire-protocol approach this PR proposed isn't necessary — the daemon's existing code path works once the AX server stops gating it. New PR: TBD. Companion daemon PR software-mansion/argent-private#10 also being closed.

PR #244, opened 2026-05-21
fix(describe): unlock iOS 26.5 SpringBoard-hosted dialogs via IgnoreAXServerEntitlements

filip131311:
## Summary

iOS 26.5 tightened cross-process AX access: SpringBoard's AX server rejects MIG queries from non-UIApplication clients with kAXError \`-25215\`. Our \`ax-service\` daemon is a \`simctl spawn\`-launched CLI binary, so it fails the new check, and \`describe\` returns an empty ROOT for any SpringBoard-hosted dialog — most visibly the TCC permission prompts (\`Allow X to use your location?\`, camera, contacts, photos, microphone, notifications). On iOS 26.4 the same daemon code works fine.

Apple ships a debug switch in \`AccessibilityUtilities\` for exactly this case: the \`com.apple.Accessibility/IgnoreAXServerEntitlements\` preference. When set, the server skips the entitlement check, and our existing \`-[AXElement explorerElements]\` walk returns the dialog tree as expected.

This PR wires the pref next to the existing \`AutomationEnabled\` write in \`ensureAutomationEnabled\`, plus a one-time SpringBoard kickstart so SB re-reads its AX-server config. The pref persists on the sim's disk, so subsequent ax-service spawns detect it's already set and skip the kickstart — avoiding the disruptive app-icon reflow on every argent run.

## Before / after on iOS 26.5

**Before** (raw socket dump for the Maps location prompt):
\`\`\`
{ \"alertVisible\": true, \"elements\": [], \"systemElements\": [], ... }
\`\`\`
→ describe returns \`source: ax-service ROOT  AXGroup\` with no children.

**After**:
\`\`\`
Source: ax-service
  AXStaticText \"Allow 'Maps' to use your location?\"
  AXStaticText \"Your location is used to show your position on the map, ...\"
  AXButton \"Allow Once\"
  AXButton \"Allow While Using App\"
  AXButton \"Don't Allow\"
  ...
\`\`\`

Matches iOS 26.4 behaviour. Same daemon binary, no entitlement claims, no resigning.

## What I changed

Single file: \`packages/tool-server/src/blueprints/ax-service.ts\` — added \~50 lines to \`ensureAutomationEnabled\` that:

1. Read the current value of \`IgnoreAXServerEntitlements\` from \`com.apple.Accessibility\`.
2. If it isn't \`1\` already: \`defaults write\` it true, then \`launchctl kickstart -k system/com.apple.SpringBoard\` so SB re-reads.
3. \`kickstart\` emits a warning about user/foreground service identifier on iOS 26 — tolerated with \`.catch(() => undefined)\`; the restart still happens.

## Cost characterisation

| Scenario | Cost |
|---|---|
| Fresh sim never touched by argent | One SpringBoard restart (~3-5s, apps reflow, status bar blinks) |
| Sim with the pref already on disk (post-first-run, post-reboot, etc.) | Zero — just one \`defaults read\` |
| Sim shut down when argent first targets it | Pref is written, SB picks it up on next boot — no restart needed |
| iOS < 26.5 sim | Pref-write is harmless; \`_shouldValidateEntitlements\` doesn't read it on older versions |

The kickstart fires at most once per sim, ever. The pref is per-sim in CoreSimulator data; \`simctl erase\` wipes it (next run sets it again).

## Background — why this is the right layer

Briefly investigated:

- **Daemon code-path tweaks** (enumerating \`currentApplicationsIgnoringSiri\`, fresh \`_AXUIElementCreateAppElementWithPid\`, etc.) — all hit the same server-side gate and return \`-25215\`. Cf. closed software-mansion/argent-private#10 and software-mansion/argent#243.
- **Sign daemon with private AX entitlements** — sim AMFI rejects (\`Security policy issue\` 163) for anything stricter than what we already have.
- **Re-host daemon as UIApplication** — would pass the new check but is a substantial refactor (Info.plist, scene config, background-task management for the socket server, install/launch instead of spawn). Not worth it when a one-line debug-pref unlocks the same outcome.

The IgnoreAXServerEntitlements path is the only one that's small, contained, and reversible.

## Test plan

- [x] \`vitest run\` in tool-server passes (175/175)
- [x] \`tsc --noEmit\` clean on tool-server
- [x] Manual verify on iPhone Air iOS 26.5: Maps location prompt now returns 5 AX elements from \`describe\` with \`source: ax-service\`
- [x] Manual verify on iPhone 17 Pro iOS 26.4: still works (pref-write is a no-op for the gating logic on that version)
- [ ] CI / unit coverage for the new conditional kickstart path

## Files

- \`packages/tool-server/src/blueprints/ax-service.ts\` — extended \`ensureAutomationEnabled\`

### DIFF
--- a/packages/tool-server/src/blueprints/ax-service.ts
+++ b/packages/tool-server/src/blueprints/ax-service.ts
@@ -123,6 +123,69 @@ export async function ensureAutomationEnabled(udid: string): Promise<void> {
     ],
     { timeout: SIMCTL_SPAWN_TIMEOUT_MS }
   );
+
+  // iOS 26.5+ gates cross-process AX queries — the SpringBoard-hosted AX
+  // server rejects MIG requests from non-UIApplication clients (the ax-service
+  // daemon is a `simctl spawn`-launched CLI) with kAXError -25215, so
+  // `describe` returns an empty ROOT for any dialog presented by SpringBoard
+  // (TCC permission prompts: "Allow X to use your location?" etc.).
+  //
+  // Apple ships a debug switch for exactly this: AccessibilityUtilities reads
+  // `com.apple.Accessibility/IgnoreAXServerEntitlements` and skips the server-
+  // side entitlement check when it's true. SB caches the value at AX-server
+  // init, so a freshly-set pref doesn't take effect until SB re-reads — which
+  // means we have to kickstart SB the FIRST time we set it. The pref persists
+  // on the sim's disk, so subsequent ax-service spawns see it already true and
+  // skip the kickstart (avoiding the disruptive app-icon reflow on every run).
+  const isAlreadySet = await execFileAsync(
+    "xcrun",
+    [
+      "simctl",
+      "spawn",
+      udid,
+      "defaults",
+      "read",
+      "com.apple.Accessibility",
+      "IgnoreAXServerEntitlements",
+    ],
+    { timeout: SIMCTL_SPAWN_TIMEOUT_MS }
+  )
+    .then(({ stdout }) => stdout.trim() === "1")
+    .catch(() => false);
+
+  if (!isAlreadySet) {
+    await execFileAsync(
+      "xcrun",
+      [
+        "simctl",
+        "spawn",
+        udid,
+        "defaults",
+        "write",
+        "com.apple.Accessibility",
+        "IgnoreAXServerEntitlements",
+        "-bool",
+        "true",
+      ],
+      { timeout: SIMCTL_SPAWN_TIMEOUT_MS }
+    );
+    // launchctl emits a warning ("Please switch to user/foreground/...") but
+    // still restarts the process; tolerate non-zero exit because the restart
+    // is what we want.
+    await execFileAsync(
+      "xcrun",
+      [
+        "simctl",
+        "spawn",
+        udid,
+        "launchctl",
+        "kickstart",
+        "-k",
+        "system/com.apple.SpringBoard",
+      ],
+      { timeout: SIMCTL_SPAWN_TIMEOUT_MS }
+    ).catch(() => undefined);
+  }
 }
 
 async function killExistingDaemon(socketPath: string): Promise<void> {
--- a/packages/tool-server/src/blueprints/ax-service.ts
+++ b/packages/tool-server/src/blueprints/ax-service.ts
@@ -1,5 +1,8 @@
 import * as net from "node:net";
 import * as fs from "node:fs";
+import * as fsAsync from "node:fs/promises";
+import * as os from "node:os";
+import * as path from "node:path";
 import { promisify } from "node:util";
 import { execFile, ChildProcess } from "node:child_process";
 import {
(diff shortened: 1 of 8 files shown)


packages/tool-server/src/blueprints/ax-service.ts
```diff
@@ -123,6 +123,53 @@ export async function ensureAutomationEnabled(udid: string): Promise<void> {
...
+        "defaults",
+        "write",
+        "com.apple.Accessibility",
+        "IgnoreAXServerEntitlements",
+        "-bool",
+        "true",
+      ],
+      { timeout: SIMCTL_SPAWN_TIMEOUT_MS }
+    );
+    // launchctl warns ("Please switch to user/foreground/...") but still restarts SB;
```
latekvo: ~~What is `SB`?~~ 

Found it: `SpringBoard`

packages/tool-server/src/blueprints/ax-service.ts
```diff
@@ -123,6 +123,53 @@ export async function ensureAutomationEnabled(udid: string): Promise<void> {
...
     { timeout: SIMCTL_SPAWN_TIMEOUT_MS }
   );
+
+  // iOS 26.5+: SpringBoard's AX server rejects MIG queries from non-UIApplication clients
+  // (this `simctl spawn`-launched CLI) with kAXError -25215, so `describe` sees an empty
+  // ROOT for any SB-hosted dialog (TCC prompts etc.). The debug pref
+  // `com.apple.Accessibility/IgnoreAXServerEntitlements` disables the check, but SB caches
+  // it at AX-server init — so we kickstart SB only on first set. The pref persists on disk,
+  // so later runs skip the disruptive icon-reflow restart.
+  const isAlreadySet = await execFileAsync(
```
latekvo: Could we possibly cache this shell call instead of making it on every single `ensureAutomationEnabled` invokation?

packages/tool-server/src/blueprints/ax-service.ts
```diff
@@ -123,6 +123,53 @@ export async function ensureAutomationEnabled(udid: string): Promise<void> {
     ],
     { timeout: SIMCTL_SPAWN_TIMEOUT_MS }
   );
+
+  // iOS 26.5+: SpringBoard's AX server rejects MIG queries from non-UIApplication clients
```
latekvo: Do i understand correctly that the `execFileAsync` above (`xcrun simctl ... write com.apple.Accessibility`) will effectively no-op on `iOS 26.5+`? If that's the case, can we only call it on relevant iOS versions?

latekvo (approved):
Thanks for the fix! ❤️ 

My concerns regard optimality of this solution. During the review i noticed this code is called infrequently anyways, so feel free to disregard these comments.

packages/tool-server/src/blueprints/ax-service.ts
filip131311: i dont wanna do version checks if i don't have to they might end up flacky (radon expirience)

packages/tool-server/src/blueprints/ax-service.ts
filip131311: it's cheap and rare and safer that way

PR #249, opened 2026-05-22
feat: Linux support

### LINKED ISSUE #246: argent doesn't run on linux
The README lists Android as adb-only, but on Linux any argent tool that goes through simulator-server fails because the bundled binaries are macOS-only. The JS dispatcher itself loads and the adb-based discovery tools work fine.

Repro on NixOS x86_64 with @swmansion/argent 0.8.0, against a booted Android emulator (emulator-5554):

- list-devices: works
- describe: works (uiautomator)
- launch-app: works (am start)
- screenshot: fails with "[SimulatorServer:emulator-5554] simulator-server exited with code before becoming ready"
- gesture-tap: fails the same way

What the npm package ships:

- bin/simulator-server: Mach-O universal x86_64+arm64
- bin/ax-service: Mach-O arm64
- dylibs/*.dylib (Darwin only by format)

CONTRIBUTING.md also lists macOS as required and notes the binaries as arm64 macOS, which contradicts the Android-only path the README documents.

For Android specifically, the gesture and screenshot tools could route through adb directly (input tap, input swipe, exec-out screencap -p) instead of simulator-server. That would unblock Linux without shipping any new native binaries.

latekvo:
This PR makes Linux work.

# AI Summary 

<details>
Brings Argent up on Linux as a host: per-platform npm bundle, renderer
selection that actually works on hybrid-GPU / Wayland / containerized
Linux, hot-boot guarding that survives the SwiftShader composite gap, a
real Linux preflight, and a Wayland-headless E2E job in CI. macOS stays on
the path it was on before (no `-gpu` flag was passed; AVD config decided);
nothing about the iOS Simulator behavior changes.

The diff is broader than "ship the Linux artifact" — calling it out below
because several pieces affect macOS too and reviewers should be primed for
that rather than skim past expecting "just packaging."

### What's actually in here

**npm packaging (the original framing — still load-bearing)**

| File | Change |
|---|---|
| `packages/argent/scripts/argent-simulator-server.cjs` (new) | Node dispatcher exposed as the `argent-simulator-server` bin entry. Joins `process.platform` into the bin path and execs. Forwards SIGTERM/SIGINT/SIGHUP parent→child so non-TTY supervisors can't orphan the binary. |
| `packages/argent/scripts/bundle-tools.cjs` | Copies per-platform binaries from `packages/native-devtools-ios/bin/<platform>/` into `packages/argent/bin/<platform>/`. darwin required, linux best-effort. |
| `packages/argent/package.json` | `bin.argent-simulator-server` points at the dispatcher. |
| `packages/native-devtools-ios/src/index.ts` | Resolver joins `process.platform` into the simulator-server bin path. iOS-only dylibs and `ax-service` throw `"requires a macOS host"` on Linux callers so they get the root cause, not "file not found". |
| `scripts/download-simulator-server.sh` | Downloads `simulator-server-argent-{macos,linux}` into platform-keyed subdirs. **Purges each platform dir before its download attempt** so a failed download can't leave stale binaries in place from a prior run. |
| `scripts/dev.cjs` | Local dev copies the current host's binary into `bin/<process.platform>/` to match the resolver layout. |
| `.github/workflows/publish{,-next}.yml` | Replaces single `test -x darwin/...` with a per-platform inventory loop. darwin always required; Linux gated by `ARGENT_REQUIRE_LINUX_BINARY=1` (flip once radon#118 lands). |

**Linux runtime — emulator boot path**

| File | Change |
|---|---|
| `packages/tool-server/src/tools/devices/boot-device.ts` | `selectGpuMode()`: Linux defaults to `-gpu swiftshader` (`auto` lands on lavapipe with ~10× cold-boot slowdown; `-gpu host` produces silent black-window failures on Optimus / NVIDIA+libglvnd / Wayland-on-hybrid / containers). macOS keeps `auto` (ANGLE→Metal). `ARGENT_EMULATOR_GPU_MODE` overrides, **validated against `emulator -help-gpu`'s whitelist** so a typoed value fails fast instead of burning the whole boot budget. `ARGENT_EMULATOR_NO_WINDOW=1` appends `-no-window` for headless contexts (CI / containers / Wayland sessions with no Qt wayland plugin). |
| `packages/tool-server/src/tools/devices/boot-device.ts` (cont.) | **Cold-boot:** `awaitFirstRealFrame` polls SwiftShader's lockscreen composite (PNG byte-count via `screencap -p`); blocks `booted:true` until a real frame shows up. Without this, callers chaining `boot-device → screenshot` get an all-black PNG because `sys.boot_completed=1` fires 5–60 s before the first composite. |
| `packages/tool-server/src/tools/devices/boot-device.ts` (cont.) | **Hot-boot:** `assertScreencapAlive` polls within an 8 s budget instead of single-probing. The earlier single-probe caught the sticky-blank state but also killed the emulator on every transient snapshot-restore blank, forcing "every boot is cold" cycles. Sticky vs transient is discriminated by whether the budget exhausts. |
| `packages/tool-server/src/utils/linux-preflight.ts` (new) | `linuxBootDiagnostics()` checks `/dev/kvm` access. `diagnoseAvdSizing()` warns when a target AVD's `hw.ramSize`/`vm.heapSize` is below the floor an aggressive RN/Expo app needs (2G/228M avdmanager defaults wedge under load). |
| `packages/tool-server/src/utils/android-binary.ts` | `defaultAndroidRoots()` probes typical SDK install locations when `ANDROID_HOME`/`ANDROID_SDK_ROOT` are unset — **cross-platform**, benefits macOS launches from GUI-launched MCP hosts (Claude Code Desktop etc.) where shell rc env doesn't propagate. |

**Cross-platform fixes that landed alongside**

| File | Change |
|---|---|
| `packages/argent-installer/src/utils.ts` | `getLatestVersion()` now redirects `npm view`'s stderr to `ignore` so `EBADDEVENGINES` warnings from a consumer project's `package.json` (e.g. Bluesky's pnpm pin) don't leak to the user's terminal. **Cross-platform fix**, not Linux-specific. |

**CI**

| File | Change |
|---|---|
| `.github/workflows/wayland-e2e.yml` (new) | Boots an AVD through the tool-server under headless Weston; asserts `hw.gpu.mode = swiftshader` and that screenshots + gesture-tap work without a display. Currently runs on `pull_request` for the affected paths + `workflow_dispatch`. Pinned to `radon-daily` for the simulator-server until radon#118 ships a Linux asset to `radon-main`. |

### Tests added in this PR

- `packages/argent/test/dispatcher.test.ts` — e2e against tmpdir-staged dispatcher + fake `<platform>/simulator-server`: missing-binary error path, exit-code passthrough, argv passthrough, SIGTERM forwarding.
- `packages/tool-server/test/native-devtools-ios-resolver.test.ts` — `requireDarwin` gates for dylibs and `ax-service` on Linux, per-platform `simulator-server` path resolution, missing-binary error, `ARGENT_NATIVE_DEVTOOLS_DIR` override.
- `packages/tool-server/test/linux-preflight.test.ts` — KVM diagnostic shape + `diagnoseAvdSizing` (ram low, heap low, both, neither, missing keys).
- `packages/tool-server/test/boot-device-hotboot.test.ts` — extended with `it.each([linux→swiftshader, darwin→auto])` so a single CI runner exercises both GPU branches (was runner-platform-dependent, regressions could pass on both runners individually). New cases for the wedged-frame fallback (under fake timers because the polling budget is short but real-time enough to exceed vitest's 5 s default), the GPU-mode whitelist rejection, and the cold-boot composite-wait timeout.
- `packages/tool-server/test/android-binary.test.ts` — extended for `defaultAndroidRoots` probe ordering and HOME pinning.

### Verification

- `tsc --build`: clean apart from pre-existing `@argent/native-devtools-android` import errors unrelated to this branch.
- `vitest run` in `packages/tool-server` and `packages/argent`: 689 + 11 = 700 tests passing; the 4 pre-existing test-file load failures (native-devtools-android missing) reproduce on `main` and are out of scope.
- `npm pack -w @swmansion/argent`: produces `swmansion-argent-0.8.0.tgz` with the per-platform layout.
- Wayland-e2e: triggered on this branch — link will be in a PR comment once the run completes.
- Manually verified end-to-end on a Wayland-on-hybrid-GPU developer laptop: AVD boots through the tool-server with `-gpu swiftshader`, screencap returns real frames, gesture tools work.

### Dependencies / what still gates this

- **radon#118** must land before the Linux artifact ships through the default channel. The download script tolerates the asset being absent today (best-effort), so this PR is technically safe to merge before — Linux installs just won't get a working `simulator-server` until the upstream artifact is published. Once it lands:
  - Drop the `radon-daily` arg in `.github/workflows/wayland-e2e.yml`.
  - Flip `ARGENT_REQUIRE_LINUX_BINARY` to `1` in both publish workflows.
- **wayland-e2e is not a required status check** today (no branch protection on `main` requires it). Worth turning on before merging anything else that touches the Linux boot path.

### Related

- radon#118 — produces the `simulator-server-argent-linux` artifact this PR consumes.
- argent-private#11 — cross-builds the iOS Simulator dylibs on Linux. Strictly orthogonal: those dylibs are macOS-runtime-only regardless of where they're built, and this PR keeps them under `dylibs/` with a clean error path on Linux callers.

</details>

Fixes #246


packages/tool-server/src/tools/devices/boot-device.ts
```diff
@@ -76,8 +77,35 @@ const STAGE_BUDGET = {
...
+// hosts — argent's screencap-based screenshot tool reports success while the
+// developer sees a black window. `swiftshader` (emulator's bundled CPU
+// renderer) sidesteps both traps and is indistinguishable from `host` on
+// modern multi-core machines. `ARGENT_EMULATOR_GPU_MODE` overrides. macOS
+// uses `auto` (resolves to ANGLE→Metal, hardware-accelerated).
+function selectGpuMode(): string {
+  const override = process.env.ARGENT_EMULATOR_GPU_MODE;
+  if (override && override.trim()) return override.trim();
+  return process.platform === "linux" ? "swiftshader" : "auto";
+}
```
latekvo, line 160: This was the primary blocker for this PR so far. I still don't have a very understanding of whether it's better to use `auto` or `swiftshader` - i found multiple broken setups for both of these, but for general reliability `swiftshader` appears to be a better choice than `auto`.

latekvo:
Triggered the wayland-e2e workflow on this branch for one of the final real-Linux verification pass:

▶ https://github.com/software-mansion/argent/actions/runs/26757963527

Follow-up commits in this push (all reviewer-feedback fixes addressed before flipping ready):

| Commit | Fixes |
|---|---|
| `4bac42c` | publish workflows: per-platform inventory + ARGENT_REQUIRE_LINUX_BINARY gate |
| `28e9600` | dispatcher SIGTERM/SIGINT/SIGHUP forwarding + e2e tests |
| `04dfe53` | native-devtools-ios resolver tests (requireDarwin gates + per-platform path) |
| `0a5266d` | boot-device: hot-path polling + ARGENT_EMULATOR_GPU_MODE whitelist |
| `f7cd32d` | hotboot tests: parametrize platform-dependent GPU assertions across linux + darwin |
| `5c858a4` | download-simulator-server.sh: purge platform dir before download (no stale binaries) |
| `119fd48` | wayland-e2e: rephrase radon-daily pin as inline note |

latekvo:
wayland-e2e went green on the first try: https://github.com/software-mansion/argent/actions/runs/26757963527

packages/native-devtools-ios/src/index.ts
```diff
@@ -26,19 +27,28 @@ export const bootstrapDylibPathTcp = () =>
 export const nativeDevtoolsDylibPathTcp = () =>
   requireDylibIn(DYLIB_TCP_DIR, "libNativeDevtoolsIos.dylib");
 
-const BIN_DIR = process.env.ARGENT_SIMULATOR_SERVER_DIR ?? path.join(__dirname, "..", "bin");
+// simulator-server is a host-side binary that talks to both iOS Simulators
+// (macOS) and Android emulators (any host with `adb`). Each platform's
+// binary lives in its own subdirectory of bin/ — keyed by `process.platform`
+// — so a single npm package can ship both without colliding filenames.
+function platformBinDir(): string {
```
filip131311, line 53: I expect this might clash with recent changes to the BIN_DIR structure made by @jwajgelt to allow for binaries that connect via tcp instead of unix sockets can you make sure that the structure is uniform and the ful matrix [platform x connection type] is supported https://github.com/software-mansion/argent/pull/270

filip131311 (commented):
I ask inline fo importand validation to perform

packages/native-devtools-ios/src/index.ts
latekvo: Yup, this had to be accounted for. Done in [62b699d](https://github.com/software-mansion/argent/pull/249/commits/62b699d5c68519cf45647529d5f2633badc9a8b4), [53b481b](https://github.com/software-mansion/argent/pull/249/commits/53b481ba9050ea1a258e258cefe0f9d4819c1917)

packages/native-devtools-ios/src/index.ts
```diff
@@ -5,8 +5,21 @@ import * as fs from "node:fs";
...
+const BIN_DIR = process.env.ARGENT_SIMULATOR_SERVER_DIR ?? path.join(__dirname, "..", "bin");
 const DYLIB_TCP_DIR = process.env.ARGENT_NATIVE_DEVTOOLS_TCP_DIR ?? path.join(DYLIB_DIR, "tcp");
 
+// iOS Simulator only runs on macOS, so the dylibs that get injected into it
+// and the ax-service that gets `simctl spawn`d into it are only ever usable
+// on darwin hosts. Throw with a clear, root-cause message so a Linux user
+// invoking these accidentally doesn't get a confusing "file not found".
+function requireDarwin(what: string): void {
+  if (process.platform !== "darwin") {
+    throw new Error(
```
filip131311, line 17: is it handled anywhere ???? 

nit: if anything it should be `assert`

filip131311 (commented):
the native devtoolls are not only tool working only on mac os there are profiling paths that lead to ios and probably so more how do we deal with it? 

do we advertise all the tools via tool server on linux are there any tools that specificaly targent ios and should be excluded from GET <toolserver-url>/tools?

filip131311 (approved):
nice one

PR #251, opened 2026-05-22
fix(android): stop forcing cold boot on every emulator launch

latekvo:
## Summary

`hasDefaultBootSnapshot` enforced a 60s mtime-skew guard between `snapshot.pb` and `ram.bin`. In practice the emulator updates `snapshot.pb` metadata on every load (load count, last-loaded timestamp) even with `-no-snapshot-save`, so the two mtimes drift apart by hours or days after a few hot-boot sessions and the guard rejects every valid snapshot — forcing a cold boot every time.

The bug hits macOS too; Linux just makes it painful because cold boot is slow there. Confirmed against my own `Pixel_6_API_34`, whose `snapshot.pb` is 4 days newer than `ram.bin` while the snapshot itself is fully usable.

## What changed

- **Drop the mtime-skew guard.** `-check-snapshot-loadable` + the hot-boot spawn's `-force-snapshot-load` already make truncated / corrupt `ram.bin` fail loudly via `attemptBoot`'s `earlyExitError` race. The on-disk pre-check only needs to confirm both files exist and `ram.bin` is non-empty.
- **Resolve the `.avd` folder via the AVD's `.ini` `path=` line**, matching what the emulator binary itself does (`external/qemu/android/android-emu/android/avd/util.cpp`). Snap-installed Studio on Linux and users who relocate AVDs to faster disks both put `.avd` outside the convention root, where the old `<root>/<name>.avd` assumption silently failed.
- **Honor `$ANDROID_USER_HOME`, `$XDG_CONFIG_HOME`, `$ANDROID_SDK_HOME`** in priority order alongside `$ANDROID_AVD_HOME` and `$HOME/.android/avd`, mirroring the emulator binary's own lookup precedence.
- **Reject whitespace-only and relative `path=` values.** The regex's non-greedy `(.+?)` still captures a single space for `path=   `, which would have leaked past `if (!avdPath)` as truthy; a relative `path=` would have resolved against `process.cwd()` and silently mis-located snapshots.

## Test plan

- [x] 18 new unit tests in `avd-snapshot.test.ts` covering: priority order across all five env-var candidates, relocated AVDs, drifted mtimes (4-day skew explicitly regression-tested), zero-length / missing / directory-shaped files, whitespace-only paths, relative paths, bad-then-good fallthrough.
- [x] Full `npm run test -w @argent/tool-server` — 703 passing.
- [x] `npm run typecheck:tests -w @argent/tool-server` — clean.
- [x] `npm run build` from workspace root — clean.
- [x] `npx prettier --check` on changed files — clean.
- [x] End-to-end repro against a real local AVD: drifted `Pixel_3a_API_34`'s `snapshot.pb` mtime to 5 days newer than `ram.bin` (the exact pathology in the wild), drove `boot-device` through this branch's built code, observed `qemu-system-aarch64 ... -force-snapshot-load -no-snapshot-save` in the spawned process — i.e. the hot-boot path was taken with a snapshot the old check would have rejected.

## Risk

Pre-check is now more lenient on disk shape. If `ram.bin` is non-empty but truncated/corrupt, the hot-boot spawn still catches it via `-force-snapshot-load` (loud early child-exit) and `attemptBoot` falls back to cold boot. So the worst case has not regressed; it just surfaces at spawn-time instead of pre-spawn.

### DIFF
--- a/packages/tool-server/src/tools/devices/boot-device.ts
+++ b/packages/tool-server/src/tools/devices/boot-device.ts
@@ -518,7 +518,23 @@ async function bootAndroidImpl(params: { avdName: string; bootTimeoutMs: number 
   if (!hasSnapshot) {
     hotBootFailureReason = "no default_boot snapshot exists";
   } else {
-    const probe = await checkSnapshotLoadable(params.avdName, "default_boot");
+    // `-gpu auto` overrides `hw.gpu.enabled=no` in AVDs created via
+    // `avdmanager create avd` (its default), which would otherwise force the
+    // emulator onto lavapipe/swangle software rendering even when the host
+    // has a usable Vulkan ICD. macOS hosts hit this less because their AVDs
+    // are usually created through Android Studio with hardware GPU on.
+    //
+    // The exact same flag MUST be passed to `checkSnapshotLoadable` too: the
+    // probe resolves a renderer based on its own argv plus the AVD's config,
+    // so a probe without `-gpu auto` ends up on a different renderer than the
+    // boot does. The snapshot saved during a `-gpu auto` boot then trips the
+    // probe's "different renderer configured" check and routes every hot-boot
+    // attempt through the cold-boot fallback. Sharing one source of truth
+    // here keeps the two argvs in lockstep.
+    const RENDERER_ARGS = ["-gpu", "auto"] as const;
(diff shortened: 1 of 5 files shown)


filip131311 (approved):
seems a reasonable change won't it fail in some exotic edge case when `"-gpu", "auto"`  argument hides from us that  `probe.laodable` should be false?

PR #257, opened 2026-05-25
fix(init): stop cramming homedir into mcp config

### LINKED ISSUE #238: argent init writes absolute /Users/<username>/… paths into committed MCP configs
After running argent init in a project, every generated MCP config has a literal home path baked into the ARGENT_MCP_LOG env var, e.g.:

```
"env": {
  "ARGENT_MCP_LOG": "/Users/<username>/.argent/mcp-calls.log"
}
```

This shows up in all six IDE-specific files that argent init writes:

.mcp.json (Claude Code)
.cursor/mcp.json
.vscode/mcp.json
.zed/settings.json
.gemini/settings.json
.codex/config.toml

These files are meant to be committed and shared with the team, so every collaborator who pulls the repo ends up with my **username** hardcoded into their MCP environment. The path doesn't exist on their machine, and the diff isn't portable.

latekvo:
`argent init` was writing `/Users/<username>/.argent/mcp-calls.log` into every generated MCP config (Cursor, Claude Code, VS Code, Windsurf, Zed, Gemini, Codex, Hermes, opencode). Those files get committed and shared, so collaborators ended up with the original author's username pinned into their environment.

Drop the env var from `buildMcpEntry` and have each adapter omit the `env`/`environment` field when the entry carries no env vars. The MCP server already falls back to `${homedir()}/.argent/mcp-calls.log` at runtime, so logging keeps working for every user on their own machine.

Fixes #238

### DIFF
--- a/packages/argent-installer/src/mcp-configs.ts
+++ b/packages/argent-installer/src/mcp-configs.ts
@@ -35,7 +35,7 @@ function getAvailableToolIds(): string[] {
 export interface McpServerEntry {
   command: string;
   args: string[];
-  env: Record<string, string>;
+  env?: Record<string, string>;
 }
 
 export interface McpConfigAdapter {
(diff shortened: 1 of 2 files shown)


filip131311 (approved):
Thank you!!!

PR #261, opened 2026-05-26
fix: prevent returning empty screenshots

latekvo:
## What broke
`toMcpContent` base64-encoded whatever bytes the simulator-server URL fetch returned and labelled them `image/png` with no checks. A 404 on the `/media/...` endpoint (file missing) returns an empty body — that became `data: ""` shipped as a PNG, which the model API rejects with "Image could not be processed". Claude Code then loops on the error and burns the session limit (cf. anthropics/claude-code#60334). Reproduces directly:

```
$ curl -sI http://127.0.0.1:<port>/media/missing.png
HTTP/1.1 404 Not Found
content-length: 0
```

## What changed
One helper added: `fetchPngBytes(url)` returns the bytes only if:
- `fetch` resolves (catches ECONNREFUSED / DNS / abort)
- `res.ok` is true
- the buffer starts with the 8-byte PNG signature

If any check fails, `toMcpContent` emits a single text placeholder instead of an image block. Diff: **+103 / −27** across `content.ts` + tests.

## E2E verification
Tested against the real MCP stdio JSON-RPC interface (not just unit tests):

**Happy path** — real Android emulator, real tool-server:
```
[1] initialize: argent v0.5.3
[3a] screenshot returned 2 block(s)
     [0] image: mime=image/png dataLen=66968 validPngSig=true
     [1] text: Saved: /var/folders/.../media/430131000-...png
HAPPY PATH overall: PASS
```

**Failure path** — mock tool-server returning a URL pointing at a closed port:
```
mock on :56529
returned 1 block(s)
  [0] text: (Screenshot unavailable: no valid PNG at http://127.0.0.1:1/stale.png. Take a new screenshot.)
hasImage=false, hasPlaceholder=true
FAILURE PATH: PASS
```

## Test plan
- [x] `npx vitest run` — 1048/1048 pass (5 new regression tests in `content.test.ts` cover 404, non-PNG body, fetch-throw, and the no-url fallback)
- [x] `npm run build` — tsc clean
- [x] `npx prettier --check` — clean
- [x] Bundled MCP server rebuilt and driven via real JSON-RPC stdio against (a) a real Android screenshot and (b) a mock tool-server reproducing the bug

### DIFF
--- a/packages/argent-mcp/src/content.ts
+++ b/packages/argent-mcp/src/content.ts
@@ -8,18 +8,44 @@ export type ContentBlock =
   | { type: "text"; text: string }
   | { type: "image"; data: string; mimeType: string };
 
+const PNG_SIGNATURE = Buffer.from([0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a]);
+
+// Fetch image bytes and confirm they actually start with a PNG signature.
+// Without this check, a 404 (file missing), an HTML error page, or any other
+// non-PNG response would be base64'd and labelled `image/png`, which the
+// model API rejects with "Image could not be processed" (issue #255).
+async function fetchPngBytes(url: string): Promise<Buffer | null> {
+  try {
+    const res = await fetch(url);
+    if (!res.ok) return null;
+    const buf = Buffer.from(await res.arrayBuffer());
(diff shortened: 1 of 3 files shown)


filip131311 (approved):
if you think its okay just merge it

PR #262, opened 2026-05-27
fix: profiler attribution on older React DevTools backends

pFornagiel:
## Summary
- Resolve profiler display-name lookup across both `getDisplayNameForElementID` and legacy `getDisplayNameForFiberID` backends.
- Add a fallback diagnostic (`displayNameApiAvailable` / `unattributed_reason`) so fully-unattributed captures are reported as an API-surface mismatch instead of a transient lifecycle race.
- Extend multi-renderer script coverage for modern, legacy, mixed, and missing display-name accessor scenarios.


filip131311 (approved):
radon flashbacks nice work

PR #269, opened 2026-05-27
feat(boot-device): pass canonical launch-hardening flags

latekvo:
## Summary

argent's boot-device currently spawns `emulator -avd <name> -gpu auto` plus the snapshot flags — nothing else. Five flags every MCP-driven boot benefits from were missing; add them as `LAUNCH_HARDENING_ARGS` and spread into the snapshot probe, hot boot, and cold boot.

### Performance (3 flags)

- `-noaudio` — guest skips pulseaudio init; saves one thread and ~50 MB RSS.
- `-no-boot-anim` — skips the Pixel boot animation, a major CPU spike on software-rendered GPU modes for ~30 s.
- `-netfast` — disables qemu network shaping; pure overhead for MCP use.

Measured on a 4-core i5-6300U with a 4096 MB / 228 MB-heap AVD, warm-cache cold boot, no other workload, `-gpu swiftshader`:

| | Boot time | Δ |
|---|---|---|
| Baseline (no extras) | 66 s | — |
| + 3 perf flags | 49 s | −26% |
| + all 5 (incl. dialog suppression) | 47 s | −29% |

qemu RSS at +20 s post-boot drops ~190 MB (5%) with the perf flags. `android-emulator-runner` (the canonical CI launcher) defaults to the same three.

### Dialog suppression (2 flags)

- `-crash-report-mode never` — keeps emulator crashes from popping a Qt consent dialog that blocks the next boot until a human dismisses it. Crash dumps still land in `/tmp/android-unknown/emu-crash-*.db`; only the modal popup is suppressed.
- `-no-metrics` — same for the metrics-collection consent dialog. Unrelated to any argent profiler tool: those run guest-side via Perfetto/simpleperf or via Metro CDP, none of which depend on the emulator's host-side Google telemetry.

### Snapshot parity

`-noaudio` and `-netfast` change qemu device topology. The snapshot's saved state has to match the runtime config the next boot uses or the snapshot validator rejects perfectly loadable snapshots — the same class of bug as the prior `-gpu auto` parity work that introduced `RENDERER_ARGS`. Plumb the hardening args into the same three call sites (probe `extraArgs`, hot spawn, cold spawn) so the probe agrees with both boot paths.

## Test plan

- [x] Unit tests: extended `boot-device-hotboot.test.ts` with two new cases asserting the flags reach the hot-boot and cold-boot spawn argv; tightened the "same renderer args" parity test to also require `-noaudio`/`-netfast` on the probe. **712 tests pass.**
- [x] Live-emulator verification on Linux: actual emulator boot with all 5 flags completes in 47 s, no flag-rejection warnings in the emulator log.
- [ ] Confirm no regression on a macOS test host (please run locally before merging — I can only verify Linux here).
- [ ] Confirm CI green.

Pre-existing test failure in `android-binary.test.ts` on `origin/main` reproduces without this change — host-side `$PATH` leakage in the test setup, not in scope for this PR.

### DIFF
--- a/packages/tool-server/src/tools/devices/boot-device.ts
+++ b/packages/tool-server/src/tools/devices/boot-device.ts
@@ -70,6 +70,40 @@ type BootDeviceResult =
   | { platform: "android"; serial: string; avdName: string; booted: true }
   | NativeDevtoolsInitFailedResult;
 
+// Flags every boot-device launch should always pass. Two purposes:
+//
+//   - Performance: `-noaudio` skips guest pulseaudio init (one thread, ~50 MB
+//     RSS); `-no-boot-anim` skips the Pixel boot animation, which is a major
+//     CPU spike on software-rendered GPU modes; `-netfast` disables network
+//     shaping (latency/speed simulation), pure overhead for MCP use cases.
+//     Measured on a 4-core Skylake host with a 4096 MB / 228 MB-heap AVD:
+//     warm-cache cold boot drops 66 s → 49 s (~25%), qemu RSS at +20 s drops
+//     ~190 MB. android-emulator-runner (the canonical CI launcher) passes the
+//     same three by default for the same reasons.
+//
+//   - Dialog suppression: `-crash-report-mode never` keeps emulator crashes
+//     from popping a Qt consent dialog that blocks the next boot until a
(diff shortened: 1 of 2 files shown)


filip131311 (approved):
semper in altum

PR #270, opened 2026-05-28
feat: make ax-service a socket client, add optional TCP connection to native tools

jwajgelt:
- flips the server-client relationship between `tool-server` and `ax-service` launched on iOS simulators: now, `ax-service` connects to a listener in `tool-server`
- adds an option to connect to `ax-service` / native devtools dylib via TCP instead of UNIX sockets

Corresponds to [argent-private PR #12](https://github.com/software-mansion/argent-private/pull/12)


latekvo (commented):
Pls fix CI failures before merging

packages/native-devtools-ios/src/index.ts
```diff
@@ -5,20 +5,29 @@ import * as fs from "node:fs";
...
+export const nativeDevtoolsDylibPath = () =>
+  requireDylibIn(DYLIB_DIR, "libNativeDevtoolsIos.dylib");
+export const keyboardPatchDylibPath = () => requireDylibIn(DYLIB_DIR, "libKeyboardPatch.dylib");
+
+export const bootstrapDylibPathTcp = () =>
+  requireDylibIn(DYLIB_TCP_DIR, "libArgentInjectionBootstrap.dylib");
+export const nativeDevtoolsDylibPathTcp = () =>
+  requireDylibIn(DYLIB_TCP_DIR, "libNativeDevtoolsIos.dylib");
 
 const BIN_DIR = process.env.ARGENT_SIMULATOR_SERVER_DIR ?? path.join(__dirname, "..", "bin");
```
filip131311, line 29: you cna change to DYLIB_DIR for consistnacy not that agent will care

packages/tool-server/src/blueprints/ax-service.ts
```diff
@@ -12,28 +13,39 @@ import {
...
 } from "@argent/registry";
-import { axServiceBinaryPath } from "@argent/native-devtools-ios";
+import { axServiceBinaryPath, axServiceBinaryPathTcp } from "@argent/native-devtools-ios";
 import { SIMCTL_SPAWN_TIMEOUT_MS } from "../utils/simctl-config";
 
 const execFileAsync = promisify(execFile);
 
 export const AX_SERVICE_NAMESPACE = "AXService";
 
+export type AXServiceTransport = "unix" | "tcp";
```
filip131311, line 23: i have a wieered instinct to tell you that this would be better as an enum, but again, will it help the agent

filip131311 (approved):
my agent tested the normal scenario (no tcp) and it is no regresion there so im fine with it

PR #271, opened 2026-05-28
feat: Argent Lens

latekvo:
Adds a UI for proposing and choosing between multiple agent-generated variants on a running app.

**MCP tools** (`packages/tool-server/src/tools/variants/`)
- `propose_variant` — non-blocking; agent stages a variant for an on-screen element (text/label/identifier/role matcher).
- `await_user_selection` — one blocking call; parks until the human submits, then returns the picks + free-form annotations + a global comment.

**Preview UI** (`packages/ui`)
- Floating cards anchored to live element frames via the existing `describe` tree, with connector lines.
- Comment-mode inspector: hover spotlight (blur+dim everything else), right-click any element to pin a comment.
- Sticky padded hitbox + trajectory-aware dismissal so focus doesn't drop on cursor jitter.
- Footbar [Complete] [Clear], confirm modal for destructive actions, green toast on submit.
- Single source of feel constants in `theme.css`.

**Electron host** (`packages/preview-window`)
- New workspace package. Frameless transparent BrowserWindow at fixed size; the squeeze open/close is a CSS `transform: scaleY` on `<html>`, GPU-composited — no OS-level resize.
- Tool-server spawns it on `await_user_selection` park, reuses it across re-awaits, and asks it to animate-close on submit (1s grace so the toast is visible).
- Stdio-line IPC (no extra ports). Ephemeral throughout — the renderer hits the tool-server's already-ephemeral `/preview/*` HTTP.


filip131311 (approved):
ship it

PR #273, opened 2026-05-29
fix(tool-server): persist update-note suppression across restarts

latekvo:
## Summary

`dismiss-update` (and the auto-suppression after delivering an update note) only stored `suppressUntil` in a module-scoped variable, so the dismissal evaporated the moment the tool-server process exited — which the launcher does after 30 min of idleness, on machine reboot, and on `npx @swmansion/argent update`. Users asking to silence the note for 24 h would see it reappear well within the window, contradicting the tool's own description.

This PR persists `suppressUntil` to `~/.argent/update-suppression.json` (alongside the existing `tool-server.json` / `tool-server.log`) and reads it back on module init, so the dismissal survives restarts.

- `suppressUpdateNote` now writes through to disk and throws on write failure (matches `dismiss-update`'s documented "Fails if … the suppression state cannot be persisted").
- The HTTP auto-suppress call (after delivering a note alongside an unrelated tool result) is wrapped in try/catch — a disk-write failure there must not 500 the user's tool call. Worst case: the note repeats next request.
- Missing/malformed/wrong-typed file all degrade to "no suppression" — never crashes the server.

## Test plan

- [x] `npx vitest run test/update-checker.test.ts` — 19/19 pass (10 original + 9 new persistence tests).
- [x] `test/http-update-note.test.ts` + `test/http-tools-meta.test.ts` — 14/14 pass (mock update-checker; persistence is transparent).
- [x] Full `@argent/tool-server` suite — 726/727 pass. The single failure (`test/android-binary.test.ts:73`, `/usr/bin/adb` resolution) is pre-existing on `main` and unrelated.
- [x] `npm run typecheck:tests` clean, `npm run build` clean, `prettier --check` clean.
- [x] Restart simulated via `vi.resetModules()` + re-import → suppression loaded back from disk.

### DIFF
--- a/packages/tool-server/src/http.ts
+++ b/packages/tool-server/src/http.ts
@@ -188,7 +188,13 @@ export function createHttpApp(registry: Registry, options?: HttpAppOptions): Htt
         const { updateAvailable, currentVersion, latestVersion } = getUpdateState();
         const shouldNotify = updateAvailable && !isUpdateNoteSuppressed();
         if (shouldNotify) {
-          suppressUpdateNote(AUTO_SUPPRESS_MS);
+          // Best-effort: a persistence failure here must not fail the user's tool call.
+          // Worst case: the note appears again on the next request.
+          try {
+            suppressUpdateNote(AUTO_SUPPRESS_MS);
+          } catch {
+            // ignore
+          }
         }
         res.json({
           data,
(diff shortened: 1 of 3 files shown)


filip131311 (approved):
nice one

PR #274, opened 2026-06-01
feat: add telemetry

hubgan:
## Summary

Adds anonymous, opt-out telemetry for Argent installer, update/uninstall flows, CLI telemetry controls, tool-server lifecycle, and tool invocation health signals.
## What Changed
- Added a new private `@argent/telemetry` package
- Introduced typed telemetry events with runtime property allowlisting/sanitization.
- Added lazy anonymous identity creation in `~/.argent/telemetry-id`.
- Added consent handling via: `argent telemetry enable`, `argent telemetry disable`, `argent telemetry status`
- Added local debug logging with `ARGENT_TELEMETRY_DEBUG=1`.
- Tracks installer funnel events for init, global install/update decisions, editor selection, allowlist decisions, skill installation, rules/agents copy, update, and uninstall.
- Tracks tool-server start/stop and tool invocation/complete/fail events.
- Adds per-tool invocation IDs so request metadata can be correlated without relying on ordering.
- Records only coarse platform metadata (`ios` / `android`) for device tools
- Ensures short-lived commands drain telemetry with bounded shutdown before exit.
- Wires telemetry package into bundling and injects full Argent CLI version metadata at build time.
- Added central `FAILURE_CODES` and `FailureSignal` helpers in `@argent/registry`.
- Attached structured, sanitized failure metadata across registry errors, tool-server tools/services, HTTP validation/dependency gates, installer flows, and `argent run`.
- Added telemetry allowlisting for safe failure fields only: static error codes/stages, coarse area/kind/command, bounded exit codes, allowlisted signals/spawn codes, and coarse network failure classes.
- Split CLI-local `argent run` failures into `cli:run_fail` so they are not double-counted as tool failures.
- Added installer telemetry for failed package actions, update/uninstall failures, rules/agents copy results, and skill refresh results.
- Made telemetry consent writes atomic and kept telemetry status side-effect free via `peekAnonId()`.
- Adds a coarse, anonymous signal to tool telemetry: which AI coding tool is driving the MCP server (Claude Code, Codex, Cursor, Gemini, VS Code, Windsurf, Zed, opencode, Copilot, or other).

## Tests Added / Updated

- Telemetry unit coverage for consent precedence, identity creation, base props, sanitization, PostHog config, debug logging, shutdown, opt-out, and forget behavior.
- Registry/tool-server tests for invocation IDs, platform metadata boundaries, startup failure flushing, and tool lifecycle telemetry.
- Installer tests for final telemetry drain, update action classification, and uninstall identity cleanup ordering.


packages/telemetry/src/ci-detect.ts
```diff
@@ -0,0 +1,54 @@
...
+    const defs = Array.isArray(vendor.env) ? vendor.env : [vendor.env];
+    return defs.every((def) => checkEnv(env, def));
+  });
+}
+
+export function isCi(env: NodeJS.ProcessEnv = process.env): boolean {
+  if (env.CI === "false") return false;
+  if (GENERIC_CI_ENV_VARS.some((name) => Boolean(env[name]))) return true;
+  return isKnownVendorCi(env);
+}
```
pFornagiel, line 58: The check here can do lots of false-positives from my perspective. The ordering should be: 
1. `env.CI == "false`
2. `isKnownVendorCi(env)`
3. `GENERIC_CI_ENV_VARS`

Also, the `GENERIC_CI_ENV_VARS` can be a little too genereic - `BUILD_ID`, `BUILD_NUMBER`, `RUN_ID` could simply be present in normal setup. Additionally, I am not sure whether ENV variables mentioning `CI` are not sometimes present on the users end. Can we do a bit more research here, or provide some source as to why this will not produce too many false-positives?

packages/telemetry/src/events.ts
```diff
@@ -0,0 +1,184 @@
...
+export interface InstallationEditorsSelectProps {
+  /** Bounded list of adapter names — sanitizer caps to 16 elements. */
+  editors: string[];
+  detected_editor_count: number;
+  scope: "local" | "global" | "custom";
+}
+
+export interface InstallationAllowlistDecisionProps {
+  is_enabled: boolean;
+  applicable_adapter_count: number;
```
pFornagiel: What is the use of the Allowlist adapter count? The user just opts in / out, and such boolean itself would be more informative imo.

packages/telemetry/src/events.ts
```diff
@@ -0,0 +1,184 @@
...
+  has_offline_cache: boolean;
+}
+
+export interface InstallationSkillInstallResultProps {
+  is_success: boolean;
+}
+
+export interface InstallationRulesAgentsCopyProps {
+  copied_count: number;
+}
```
pFornagiel: This is not very informative as well.

packages/telemetry/src/events.ts
```diff
@@ -0,0 +1,184 @@
...
+
+export interface InstallationSkillInstallProps {
+  method: "default" | "interactive" | "manual";
+  is_online: boolean;
+  has_offline_cache: boolean;
+}
+
+export interface InstallationSkillInstallResultProps {
+  is_success: boolean;
+}
```
pFornagiel: I feel like this is more of a "is Github / npx skills operating properly" rather than "this is something we wish to keep track of to get better view of tool usage". I do not think it fits.

packages/telemetry/src/events.ts
```diff
@@ -0,0 +1,184 @@
...
+  duration_ms: number;
+}
+
+export interface ToolFailProps {
+  tool: string;
+  tool_invocation_id: string;
+  platform?: "ios" | "android";
+  duration_ms: number;
+}
+
```
pFornagiel: Up to discussion: Would it not be better to consolidate this into `ToolInvokeProps` entry with information about `success` / `failure`? Or would that be less clear / problematic to implement?

pFornagiel (commented):
The `README.md` right now states:

```
Argent does not collect or transmit any user data. No telemetry, no analytics, no crash reporting.
```

This will have to be updated as well

pFornagiel:
Possibly found something - at the moment it seems like every time when we do `run-sequence` and possibly `flow` we are emitting standalone events, even though they are  part of these tools invocations. Maybe there would be a way to mark those events as executed using `flow` or `run-sequence` or not gather them? Up to discussion.

pFornagiel:
Additionally, maybe we could also add information whether the source of invocation was `MCP` or `CLI`. If you believe this is a scope for another PR - it is also fine.

packages/telemetry/src/events.ts
hubgan: I agree, I dropped `applicable_adapter_count` and keep just the `is_enabled` boolean on `allowlist_decision`.

packages/telemetry/src/events.ts
hubgan: You're right I removed `installation:rules_agents_copy` event entirely

packages/telemetry/src/events.ts
hubgan: Yeah I removed `installation:skill_install_result` event.

packages/telemetry/src/ci-detect.ts
hubgan: The generic list, the CI === "false" bypass, and the vendor table are all taken verbatim from `ci-info` v4.4.0. Added a comment noting this. Kept BUILD_ID/BUILD_NUMBER/RUN_ID for parity. Narrowing them risks false negatives (Jenkins/TeamCity set only these), and this just gates a telemetry property, not behavior. CI === "false" is the escape hatch for anyone hitting a false positive. Ordering is a no-op — checks 2 and 3 are an OR, and the current order matches ci-info's.

latekvo (approved):
Reviewed the core telemetry path and the privacy surface — no High/Medium issues.

packages/telemetry/src/posthog.ts
```diff
@@ -0,0 +1,64 @@
...
+  const config = resolveConfig();
+  if (!config.isUsable) {
+    client = null;
+    return null;
+  }
+
+  const opts = {
+    host: POSTHOG_HOST,
+    disableGeoip: true,
+    requestTimeout: 3000,
```
latekvo, line 43: The bounded shutdown this client feeds into doesn't actually bound process exit when the network is unreachable. `shutdown()`'s internal `Promise.race` returns within the timeout, but posthog-node flushes on a retry backoff whose timer is not `unref`'d, so the orphaned drain keeps the Node event loop alive long after the race has resolved. I built the CLI from this branch and ran `argent telemetry disable` against an unreachable host: it prints "Telemetry disabled." and then hangs ~22s before the process exits (the opt-out is already persisted at that point, so the time is pure dead wait). The same applies to a successful `argent init` — its completion path returns to the event loop instead of calling `process.exit`, so a first-time setup on an offline/flaky network hangs ~22s after reporting success. Only the paths that call `process.exit` explicitly (the tool-server, `argent run`) are unaffected.

packages/argent-installer/src/init.ts
```diff
@@ -61,511 +116,686 @@ function extractFlag(args: string[], flag: string): string | null {
...
+      const packageActionStartedAt = performance.now();
+      track("installation:global_install_decision", { decision: "already_installed" });
+      await trackPackageAction("already_installed", packageActionStartedAt, true);
+      let latest: string | null = null;
+      const spinner = p.spinner();
+      spinner.start("Checking for updates...");
+      try {
+        latest = getLatestVersion();
+      } catch {
+        await trackPackageAction(
```
latekvo: The `already_installed` package_action is emitted just above (line 262). When the update check throws here because the registry is unreachable, a second package_action is emitted for the same init step, labelled `update_skipped` with `INSTALL_REGISTRY_CHECK_FAILED`. That double-counts the step and conflates a failed update *check* with the user explicitly choosing to skip an available update. Driving `init()` with the registry call throwing produces both `{action: already_installed}` and `{action: update_skipped}` from one run. This emission was previously removed for this exact reason and reintroduced in the structured-failure change — the comment that replaced it spelled out why a failed check shouldn't reuse `update_skipped`.

packages/telemetry/src/identity.ts
```diff
@@ -0,0 +1,110 @@
...
+    let fd: number;
+    try {
+      fd = fs.openSync(tmpPath, "wx", 0o600);
+    } catch (err) {
+      if ((err as NodeJS.ErrnoException).code === "EEXIST") continue;
+      throw err;
+    }
+    try {
+      fs.writeSync(fd, uuid);
+      fs.fsyncSync(fd);
```
latekvo: If `writeSync` or `fsyncSync` throws here (e.g. ENOSPC / EIO), the temp file created by `openSync` just above is orphaned: the `unlinkSync` cleanup lives in the `finally` of the `link` block further down, which is only reached after the write succeeds. The function then throws out of the retry loop leaving a `.telemetry-id.tmp.<pid>.<uuid>` file behind. Confirmed by forcing `fsyncSync` to throw — a leftover temp file remains and no final id is published. Low impact (subsequent reads tolerate stray temp files), but this path doesn't clean up after itself.

packages/telemetry/src/ai-identity.ts
```diff
@@ -0,0 +1,74 @@
...
+  "zed",
+  "opencode",
+  "copilot",
+  "other",
+] as const;
+
+export type AiClient = (typeof AI_CLIENTS)[number];
+
+/** Bounded free-form client name, used for the long tail of unrecognized tools. */
+export const AI_CLIENT_NAME_PATTERN = /^[A-Za-z0-9 ._-]{1,80}$/;
```
latekvo: This pattern accepts username- and hostname-shaped strings — `alice`, `john.doe`, `alice-macbook.local`, `DESKTOP-AB12CD` all match (paths, URLs and emails are correctly rejected). The value comes from the MCP client's self-reported `clientInfo.name` and is only forwarded for unrecognised clients, but a non-standard client that names itself after the machine or user would have that string captured. Worth confirming the free-form name's residual cardinality/PII exposure is acceptable versus recording only the bounded `other` bucket.

packages/telemetry/src/sanitize.ts
```diff
@@ -0,0 +1,234 @@
...
+const FAILURE_SIGNAL_NAME = oneOf(FAILURE_SIGNAL_NAMES);
+const FAILURE_SPAWN_CODE = oneOf(FAILURE_SPAWN_CODES);
+const NETWORK_FAILURE = oneOf(NETWORK_FAILURES);
+
+const AI_CLIENT = oneOf(AI_CLIENTS);
+const AI_CLIENT_NAME = matches(AI_CLIENT_NAME_PATTERN, 80);
+
+const AI_TELEMETRY = {
+  ai_client: AI_CLIENT,
+  ai_client_name: AI_CLIENT_NAME,
```
latekvo: `ai_client` and `ai_client_name` are validated independently, so the sanitizer doesn't enforce the invariant documented in ai-identity.ts that the free-form name is only carried when `ai_client === "other"`. Both current producers gate the name on `other`, so it isn't a live leak, but this central allowlist — the layer meant to be the backstop — would keep `ai_client_name` even paired with a recognised `ai_client` (verified: `{ai_client: "codex", ai_client_name: "..."}` passes both through).

packages/telemetry/src/base-props.ts
```diff
@@ -0,0 +1,60 @@
...
+import { isCi } from "./ci-detect.js";
+
+// Build-time version metadata injected by esbuild; source tests fall back to "0.0.0".
+declare const ARGENT_CLI_VERSION: string | undefined;
+
+// Process-local session id. Never persisted or reused across Node processes.
+let SESSION_ID: string = randomUUID();
+
+function readCliVersion(): string {
+  const fromDefine = (globalThis as any).ARGENT_CLI_VERSION;
```
latekvo: This `globalThis.ARGENT_CLI_VERSION` read is dead in the bundled artifact. esbuild's `define` substitutes the bare identifier on the next line (the bundle folds it to `if (true) return "<version>"`) but not this member expression, so `fromDefine` is always `undefined` at runtime — confirmed in the built bundle. Version injection still works via the bare-const branch; this branch never executes.

Telemetry.md
```diff
@@ -0,0 +1,111 @@
...
+- which Argent tools are invoked, and whether they succeeded or returned an error;
+- which AI coding tool is driving Argent;
+- how long a process ran in the terminal or in CI, and which Argent component emitted the event;
+- start and stop of the Argent tool-server, its uptime, the number of tools used, and the reason it stopped.
+
+### Environment
+
+- Argent version, Node.js version, operating system, processor architecture;
+- whether the process runs in an interactive terminal and whether it runs in a CI environment;
+- whether Argent is used in connection with Android or iOS.
```
latekvo: This lists only Android and iOS, but the platform set now also includes `chromium` (it's in `PLATFORMS` and the sanitizer allowlist), so the coarse platform of a Chromium tool invocation is collected as well. The documented data and the collected data have drifted here.

README.md
```diff
@@ -129,12 +130,15 @@ argent init
...
-- Argent integrates with your agent locally over MCP stdio.
-- Its internal tools are not reachable from outside your machine.
-- The only outbound network call we make is the version check against our public npm package, which sends no user data and fails gracefully if blocked.
+You can opt out at any time:
+
+```bash
+argent telemetry disable   # check status with: argent telemetry status
+```
+
+For the full details — see the [Argent Privacy Notice (Telemetry)](https://github.com/software-mansion/argent/blob/main/Telemetry.md).
```
latekvo, line 141: This "full details" link points at `Telemetry.md` on GitHub, while the in-product first-run notice sends users to `https://swmansion.com/legal/argent/privacy-notice/`. Two different canonical destinations for the same privacy notice — worth aligning them.

packages/tool-server/src/tools/profiler/react/react-profiler-cpu-summary.ts
```diff
@@ -133,8 +133,14 @@ Fails if react-profiler-stop has not been called or no CPU profile is stored.`,
   async execute(_services, params) {
     const sessionPaths = getCachedProfilerPaths(params.port, params.device_id);
     if (!sessionPaths?.cpuProfilePath) {
-      throw new Error(
-        "No CPU profile stored. Call react-profiler-start, exercise the app, then react-profiler-stop."
+      throw new FailureError(
+        "No CPU profile stored. Call react-profiler-start, exercise the app, then react-profiler-stop.",
+        {
+          error_code: FAILURE_CODES.REACT_PROFILER_NO_CPU_PROFILE,
```
latekvo, line 139: `REACT_PROFILER_NO_CPU_PROFILE` is also emitted in react-profiler-stop.ts for a different condition, and the two sites disagree on `error_kind` (`validation` here, `unknown` there). They stay distinguishable via `failure_stage`, so the only effect is telemetry bucketing precision — noting it for consistency.

packages/registry/src/errors.ts
```diff
@@ -1,11 +1,215 @@
...
+export const FAILURE_SPAWN_CODES = ["EACCES", "ENOENT", "EPERM", "ETIMEDOUT"] as const;
+
+export type FailureSpawnCode = (typeof FAILURE_SPAWN_CODES)[number];
+
+export const NETWORK_FAILURES = [
+  "timeout",
+  "connection_refused",
+  "connection_reset",
+  "invalid_response",
+  "unavailable",
```
latekvo: `unavailable` is in this allowlist but no code path ever sets `network_failure: "unavailable"` — the producers emit `timeout` / `connection_refused` / `connection_reset` / `invalid_response` / `other`. Dead allowlist entry.

latekvo (commented):
Full pass over the telemetry change — findings inline below. Most are low-severity consistency/cleanup items; two are worth prioritising: the offline shutdown hang (`telemetry disable` / `init`) and the duplicate `package_action` emitted on an offline already-installed `init`. All findings were reproduced against the CLI built from this branch or the real modules, not just code review.

packages/tool-server/src/index.ts
```diff
@@ -52,12 +59,34 @@ export function start(): void {
...
-    if (shuttingDown) return; // avoid re-entrant shutdown
-    shuttingDown = true;
+    // A second fatal event must not re-run teardown or schedule a second timer.
+    if (crashing) return;
+    crashing = true;
+    shutdownReason = "crash";
+    finalExitCode = 1;
+    shutdownFailureSignal = {
+      error_code:
+        label === "Unhandled rejection"
```
latekvo: The failure code here is chosen by comparing the human-readable `label` argument against the literal `"Unhandled rejection"`. It's correct today because the two call sites pass matching labels, but it couples the emitted failure code to a log-prefix string — a later tweak to the label wording would silently misclassify the crash code.

latekvo (approved):
Re-reviewed after the latest fixes. The earlier findings are addressed:

- Offline shutdown no longer hangs the process — capping fetch retries bounds the worst-case tail to a single request timeout (verified locally: ~22s → ~4s offline).
- The offline `init` path no longer double-emits `package_action`; the registry-unreachable branch is now silent, and non-interactive installs emit the matching `update_decision`.
- The temp identity file is cleaned up on every failure path after `openSync`.
- The free-form AI client name is gone entirely (with a regression test asserting it is never carried), so the permissive pattern and the missing cross-field invariant are both moot.
- Dead `globalThis` version read removed; base props memoized with a test seam.
- Crash classification is passed explicitly instead of being derived from the log label; the unused `unavailable` network-failure value and `INSTALL_REGISTRY_CHECK_FAILED` code are removed; the reused profiler error code now reports a consistent `error_kind`; the Telemetry.md platform list includes Chromium.

One non-blocking leftover: README still links to the in-repo `Telemetry.md` while the in-product notice points at the hosted privacy page — two destinations for the same "full details" that can drift.

Build and the telemetry / registry / installer / tool-server suites pass locally; CI is green.

PR #277, opened 2026-06-01
feat(remote): artifact transport for files produced by a remote tool-server

filip131311:
> ⚠️ **Stacked on unmerged work.** This PR's intended base, `@pFornagiel/remote-server-init`, no longer exists on origin, so it targets `main`. The diff therefore includes the 31 commits of the remote-server-init feature (link/unlink, idle-shutdown, launcher) that haven't landed in `main` yet. **My change is the single commit `a3d4c963`** ("feat(remote): artifact transport …"); review that commit, or rebase onto the real base once it's restored.

---

## Why

Files produced on the **tool-server host** (screenshots, profiler exports) were surfaced to the agent as host paths / `127.0.0.1` media URLs. The MCP client (`argent-mcp`) then did a *second* fetch of that localhost media URL to get the bytes. When the tool-server runs **remotely**, that URL resolves to the client's own machine — nothing there — so screenshots fail with `(Screenshot unavailable: no valid PNG …)`. Profiler exports have the same shape: paths that only exist on the remote host.

## What

A reusable **artifact boundary layer** so any file-producing tool works regardless of where the tool-server runs:

**Phase 0 — foundation**
- **tool-server** (`src/artifacts.ts`, `http.ts`): in-process artifact registry + `GET /artifacts/:id` streaming route. `registerArtifact(hostPath)` returns a wire handle (`{ __argentArtifact, id, filename, mimeType, size }`) instead of a raw path.
- **argent-mcp** (`src/artifacts.ts`, `content.ts`, `mcp-server.ts`): a client materializer deep-walks tool results for handles, downloads each over the remote-aware `TOOLS_URL`, writes it into a **structured cache**, and rewrites the result to the **local** path (+ inline image rendering).

**Cache layout** (override root with `ARGENT_ARTIFACTS_DIR`):
```
~/.argent/artifacts/<project>/<session>/<device>/<filename>
```
- `project` = `basename(cwd)` + short hash of the full path (readable + collision-safe)
- `session` = minted once per argent-mcp process (groups a run, easy GC)
- `device` = udid/serial when the artifact is device-scoped, omitted otherwise

**Phase 1 — screenshot** returns an image artifact; `content.ts` renders inline from the downloaded bytes with **no second `127.0.0.1` fetch**. Fixes `auto-screenshot` and `flow-execute` screenshots via the same path.

**Phase 2 (partial) — `native-profiler-stop`** `exportedFiles` (XML) become downloadable artifacts materialized client-side.

## Deferred (follow-ups)
- **`.trace` bundle** — it's a directory (Instruments bundle); streaming needs archiving (no archiver dep yet). `native-profiler-stop` now returns a `traceFileNote` explaining it stays on the profiling host.
- **Phase 3 — project source on remote**: `react-profiler-component-source` and `react-profiler-analyze` read the user's source tree on the tool-server host. Class-A profiler analysis chains already work remotely (server-internal). Tracked separately.

## Compatibility
- Legacy `{ url, path }` screenshot rendering retained as a fallback when no artifact context is present (older tool-servers, unit tests).

## Tests
- tool-server: artifact registry (infer/override), `/artifacts/:id` (200/404/410), updated `stop-recovery` assertions. **723 pass.**
- argent-mcp: materializer (download/rewrite/nested/failure/device-scoping), artifact-aware `toMcpContent`. **65 pass.**
- Both packages typecheck (src + tests); prettier clean.

## Notes
- Routes around the `argent-private` native simulator-server binary (which emits the `127.0.0.1` media URL) rather than changing it — no submodule change.


filip131311:
Superseded by #278, which targets the correct base (feat/argent-server-start) with a clean single-commit diff. The original base @pFornagiel/remote-server-init was deleted from origin (squash-merged into feat/argent-server-start as #200).

PR #287, opened 2026-06-01
fix(argent): declare + ship tree-sitter so react-profiler-component-source finds components

latekvo:
## Problem

`react-profiler-component-source` returned `found:false` for **every** component in a standard TS/TSX Expo project.

## Two root causes (both verified live against the Bluesky app)

1. **tree-sitter never declared/shipped.** `@swmansion/argent` didn't depend on `tree-sitter`/`tree-sitter-typescript`, so the parser `require()` threw, was swallowed, and the AST index was always empty. Fixed by adding them as runtime deps and keeping them `external` in the CJS tools bundle so the native addons are `require()`d at runtime (not inlined). Verified: `HostingProvider`, `LoggedOutLayout`, `Login` now resolve with correct file:line.

2. **memo()/forwardRef() components were skipped.** The AST matcher's `variable_declarator` branch only indexed `arrow_function`/`function_expression` value nodes; a wrapped component (`const X = memo(...)` / `forwardRef(...)` / `React.memo(...)`) has a `call_expression` value node, so it was never indexed — and profiler findings are disproportionately memo-wrapped, i.e. exactly the components users look up. Now recognises memo/forwardRef wrapper calls (incl. `React.`-prefixed and nested `memo(forwardRef(...))`) and indexes the declared name; `memo` sets `isMemoized`. Verified live: `ThreadItemReadMore`, `CenteredView`, `ScrollView`, `BookmarkButton`, `Logo` now resolve.

## Tests
`ast-index.test.ts` covers the idiomatic forms (default-export/plain/arrow) and the memo/forwardRef/React.memo/nested wrappers; the wrapped-form test fails on pre-fix code. Rebased onto latest main (re-merged the `bundle-tools.cjs` helper refactor; bundle still externalizes tree-sitter). CI green.

### DIFF
--- a/packages/tool-server/src/utils/react-profiler/pipeline/06-resolve/ast-index.ts
+++ b/packages/tool-server/src/utils/react-profiler/pipeline/06-resolve/ast-index.ts
@@ -141,6 +141,29 @@ function nodeContainsCall(node: TreeSitterNode, callName: string): boolean {
   return false;
 }
 
+/**
+ * Detect a React component wrapper call: memo(...) / forwardRef(...) /
+ * React.memo(...) / React.forwardRef(...) (also nested, e.g. memo(forwardRef(...)),
+ * since the outer callee is what we match). Components declared as
+ * `const X = memo(...)` / `const X = forwardRef(...)` have a call_expression
+ * value node, so without recognising these they would be missed entirely --
+ * and profiler-flagged components are disproportionately memo-wrapped.
+ */
+function reactWrapperCall(node: TreeSitterNode | undefined): {
+  isWrapper: boolean;
+  isMemo: boolean;
+} {
+  if (!node || node.type !== "call_expression") return { isWrapper: false, isMemo: false };
+  const callee = node.children[0];
+  let name: string | undefined;
+  if (callee?.type === "identifier") name = callee.text;
+  else if (callee?.type === "member_expression")
+    name = callee.children[callee.children.length - 1]?.text;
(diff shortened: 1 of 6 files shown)


packages/tool-server/src/utils/react-profiler/pipeline/06-resolve/ast-index.ts
```diff
@@ -204,20 +227,23 @@ export async function buildAstIndexWithDiagnostics(projectRoot: string): Promise
...
+          (valueNode.type === "arrow_function" ||
+            valueNode.type === "function_expression" ||
+            wrapper.isWrapper)
         ) {
           const componentName = nameNode.text;
           if (!index.has(componentName)) {
             index.set(componentName, {
               file,
               line: nameNode.startPosition.row + 1,
               col: nameNode.startPosition.column,
```
hubgan, line 243: I wouldn't do that in this PR but I think it's worth mentioning that if two components same have name only one survives. For example in bluesky app there is a `List` component in `List.tsx` and another in `List.web.tsx`. We keep only the first one it happens to stumble across while scanning folders, and throws the rest away. So when we ask for List, we might get the wrong versions code.

hubgan (approved):

packages/tool-server/src/utils/react-profiler/pipeline/06-resolve/ast-index.ts
latekvo: Will open a new PR

PR #290, opened 2026-06-02
feat: add structured failure telemetry

hubgan:
## Summary

- Added central `FAILURE_CODES` and `FailureSignal` helpers in `@argent/registry`.
- Attached structured, sanitized failure metadata across registry errors, tool-server tools/services, HTTP validation/dependency gates, installer flows, and `argent run`.
- Added telemetry allowlisting for safe failure fields only: static error codes/stages, coarse area/kind/command, bounded exit codes, allowlisted signals/spawn codes, and coarse network failure classes.
- Split CLI-local `argent run` failures into `cli:run_fail` so they are not double-counted as tool failures.
- Added installer telemetry for failed package actions, update/uninstall failures, rules/agents copy results, and skill refresh results.
- Made telemetry consent writes atomic and kept telemetry status side-effect free via `peekAnonId()`.

## Tests
- Added/updated tests for registry failure signals, telemetry sanitization, registry listener forwarding, CLI run failure telemetry, HTTP preflight failure telemetry, and uninstall failure telemetry.


latekvo:
**`init()` no longer flushes telemetry on an unexpected throw** (`packages/argent-installer/src/init.ts`)

The base branch wrapped the whole `init()` body in `try { … } catch (err) { await finalizeInitTelemetry(); throw err; }`. This PR removed that outer wrapper, so `finalizeInitTelemetry()` (the drain) now runs only on the success tail and a few classified failure paths. Any unhandled throw (file I/O, `copyRulesAndAgents`, the online check, a clack prompt) propagates to `main().catch()` in `cli.ts`, which only does `console.error` + `process.exit(1)` — so `installation:cli_init_complete` is never emitted and previously-buffered events are dropped unflushed.

**`uninstall()` has the same regression** (`packages/argent-installer/src/uninstall.ts`)

Base wrapped the body in `try { … } catch (err) { await finalizeUninstallTelemetry(…); throw err; }`; this PR removed it too. An unclassified throw in the prune/cleanup section or a prompt skips `finalizeUninstallTelemetry`, so `installation:cli_uninstall_complete` is lost and the buffered `cli_uninstall_start` never flushes. `update.ts` kept its equivalent wrapper, so update is unaffected — which makes the divergence look unintentional.

PR #291, opened 2026-06-02
feat: initial sim-remote support

jwajgelt:
## Summary

Adds **initial support for remote iOS simulators** ("sim-remote") — simulators that run in the cloud (radon-cloud orchestrator) and are driven from the dev's machine over a QUIC/MoQ tunnel, alongside the existing local iOS simulators. A remote sim is exposed as a new device platform, `ios-remote`, and the existing interaction/inspection/profiling tools work against it through the same tool surface.

To get there without scattering `isRemote` branches across every tool, the iOS code paths are refactored behind a small set of host/transport/backend abstractions so each tool stays transport-agnostic.

## How it works

- **Discovery** — `list-devices` enumerates remote sims via the `sim-remote` CLI in addition to local `simctl`. They classify as `platform: "ios-remote"`. Enumeration is best-effort: if `sim-remote` isn't installed or the user isn't logged in, it silently returns none.
- **Transport** — remote interaction goes over a MoQ relay (WebTransport/QUIC). `moq-client.ts` subscribes to the server-published `simulator` broadcast (catalog/video/screenshot) and publishes a `control` track carrying protobuf-encoded input commands (touch / key / button / rotate / wheel / screenshot). WebTransport is provided in Node by `@fails-components/webtransport`, polyfilled lazily on first use.
- **Native devtools over the tunnel** — unix sockets don't bridge the QUIC tunnel, so on remote sims `ax-service` and `native-devtools` are forced onto TCP with OS-assigned ephemeral ports, and a reverse proxy is wired before the in-sim daemon dials back.

## Key components

New (`packages/tool-server/src/`):
- `utils/sim-remote.ts` — thin wrapper around the `sim-remote` CLI (simctl verbs, `moq-info`, proxy), propagating CLI stderr verbatim.
- `utils/moq-client.ts` — pure-JS MoQ client for the remote simulator-server relay.
- `utils/datachannel-proto.ts` + `proto/datachannel.proto` — vendored protobuf schema + encoders for the control channel.
- `utils/ios-host.ts` — `IosHost` abstraction (`localIosHost` / `remoteIosHost`, picked by `pickIosHost(device)`) capturing the local-vs-remote differences (transport, env setup, proxying, `requiresTcp`).
- `utils/simctl-backend.ts` — strategy for simctl verbs so one iOS handler serves both `xcrun simctl` (local) and `sim-remote simctl` (remote).
- `utils/ax-prefs.ts` — AX-pref helpers extracted from `ax-service`.
- `tools/{launch-app,open-url,reinstall-app,restart-app}/platforms/ios-remote.ts` + shared handlers.

Refactors:
- **Strategy pattern** to limit `isRemote`/platform branching inside tool and blueprint implementations.
- `simulator-client.ts` gains a transport abstraction (`SimulatorServerTransport`) so call sites stay transport-agnostic; local sims back it with WebSocket+HTTP, remote sims with the MoQ client.
- `ax-service` / `native-devtools` blueprints updated for ephemeral TCP ports and the `ios-remote` platform.
- Dependency checks: tools that pick a platform path internally now gate on the right CLI (`sim-remote` vs `xcrun`).

## New dependencies

`@moq/lite`, `@moq/net`, `@fails-components/webtransport`, `@fails-components/webtransport-transport-http3-quiche`, `protobufjs`.

## Testing

- Existing unit suite passes (CI green); the native-devtools status/describe tests now stub the platform dep-check so they exercise their logic, not the host toolchain.
- E2E device-smoke jobs (local iOS sim, Android, Chromium, remote VVD) pass.

---

<sub>Branch rebased onto latest `main` (linear history). `sim-remote` is opt-in — only required when driving a remote iOS simulator.</sub>


filip131311 (approved):
I think this should be okay thank you <3

PR #292, opened 2026-06-03
fix(installer): suppress npm stderr in getLatestVersion

latekvo:
## Summary

- Adds `stdio: ["ignore", "pipe", "ignore"]` to the `execSync` call in `getLatestVersion()`, matching the pattern already used in `getGlobalBinaryPath`
- Prevents npm warnings (notably `EBADDEVENGINES`) from printing to the user's terminal when argent checks for updates inside a project whose `package.json` triggers engine mismatch warnings — e.g. Bluesky's repo

## Stacking

## Test plan

- [ ] `npm run test -w packages/argent-installer` passes


latekvo:
Larger issue - running `argent init` in bluesky throws lots of errors preventing skills from being installed. I originally thought these errors are not critical, but they are.

PR #293, opened 2026-06-03
feat(android-binary): probe OS-default SDK install locations when env vars are absent

latekvo:
## Summary

- Extends `androidRoots()` to fall back to OS-default Android SDK install paths after `ANDROID_HOME` / `ANDROID_SDK_ROOT` come up empty
- Covers Android Studio defaults (`~/Library/Android/sdk` on macOS, `~/Android/Sdk` on Linux) and common system-wide paths (`/opt/android-sdk`, `/usr/lib/android-sdk`, etc.)
- Fixes the case where an MCP server spawned by a GUI process (Claude Code desktop, VS Code) doesn't inherit the shell-exported `ANDROID_HOME`, causing argent to silently fail to find `adb`/`emulator` even on a correctly configured machine

## Stacking

## Test plan

- [ ] New tests in `android-binary.test.ts` cover the Studio-default paths and the `ANDROID_HOME` priority invariant
- [ ] `npm run test -w packages/tool-server` passes

### DIFF
--- a/packages/tool-server/src/utils/android-binary.ts
+++ b/packages/tool-server/src/utils/android-binary.ts
@@ -1,6 +1,7 @@
 import { execFile } from "node:child_process";
 import { access } from "node:fs/promises";
 import { constants as fsConstants } from "node:fs";
+import { homedir } from "node:os";
 import { join } from "node:path";
 import { promisify } from "node:util";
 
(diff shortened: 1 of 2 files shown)


filip131311 (approved):
git commit

PR #294, opened 2026-06-03
feat(boot-device): Linux boot diagnostics for KVM access and AVD sizing

latekvo:
## Summary

- Adds `linux-preflight.ts` with two non-blocking diagnostics run at boot time on Linux hosts:
  - `/dev/kvm` readability check — warns when KVM is missing or inaccessible, which causes 10–50× TCG software-emulation slowdown
  - AVD `hw.ramSize` / `vm.heapSize` floor check — warns when values are below the threshold where Android's Watchdog kills `system_server` under Hermes+Metro+swiftshader load
- Both warnings name the exact fix command; neither throws nor blocks
- Wired into `boot-device.ts` via `linuxBootDiagnostics(avdName)` immediately after `ensureDep` succeeds

## Stacking

## Test plan

- [ ] `linux-preflight.test.ts` covers the pure `diagnoseAvdSizing` helper for all four sizing combinations
- [ ] `npm run test -w packages/tool-server` passes

### DIFF
--- a/packages/tool-server/src/tools/devices/boot-device.ts
+++ b/packages/tool-server/src/tools/devices/boot-device.ts
@@ -24,6 +24,7 @@ import {
   waitForBootCompleted,
 } from "../../utils/adb";
 import { ensureDep } from "../../utils/check-deps";
+import { linuxBootDiagnostics } from "../../utils/linux-preflight";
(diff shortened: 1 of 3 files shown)


filip131311 (approved):
ok

PR #297, opened 2026-06-03
feat(tool-server): add disable-update-notification flag

pFornagiel:
## Summary

- Adds a `disable-update-notification` feature flag to `configuration-core` so users can permanently suppress the agent-facing "update available" note via `argent enable disable-update-notification`.
- Wires the flag into the tool-server HTTP layer through a small `update-reminder` helper; `argent update` and the `update-argent` tool are unchanged, and temporary dismiss/auto-suppress behavior stays as-is.
- Adds HTTP integration test coverage for the flag-gated update note.

Stacked on `@pFornagiel/env-to-feature-flags`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3702,7 +3702,8 @@
       "version": "0.9.0",
       "dependencies": {
         "@argent/configuration-core": "file:../configuration-core",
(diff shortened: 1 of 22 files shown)


filip131311:
this seems important but is very stale

PR #305, opened 2026-06-03
fix(boot-device): don't tear down a healthy-but-slow emulator on the final boot attempt

latekvo:
## Problem

On Linux (headless, software-GPU) I hit `boot-device` repeatedly failing with
`PackageManager did not respond ... Emulator has been terminated`, even though
the emulator had booted fine. Root cause: **Stage 5** of `attemptBoot` ran a
single `pm path android` probe with a 10 s timeout and, on failure, killed the
emulator and threw.

On a loaded host (or a freshly `-wipe-data`'d image still finishing its
first-boot package scan), `pm` can take well over 10 s to answer while the guest
is perfectly healthy — it already reached `sys.boot_completed` and registered
with adb. On the **final cold attempt** there is no fallback, so a momentarily
slow PackageManager turned a usable emulator into a hard boot failure (and
destroyed it, so the next try started from a cold boot again).

## Fix

- Stage 5 now **retries** `pm path android` within a budget instead of a single
  10 s window.
- Threads a `tearDownIfUnready` flag through `attemptBoot`:
  - **hot-boot attempt** — tight ~10 s budget, tear down on failure so we still
    fall through to the cold boot.
  - **final cold attempt** — retry up to 45 s and, if PM is still slow, **return
    the emulator as booted rather than destroying it**. gRPC screenshots/gestures
    work without PM, and killing it just guarantees failure with nothing to fall
    back to. A confirmed QEMU crash (mid-probe or via the early-exit racer) still
    tears down and rethrows the real cause.

Also adds a note to the `argent-android-emulator-setup` skill: hard-killing qemu
(`pkill -9`) leaves the userdata image dirty — the usual cause of a cold boot
that hangs for minutes with runaway writes — and `-wipe-data` resets it.

## Tests

`packages/tool-server/test/boot-device-hotboot.test.ts` — added a regression
test asserting the final cold attempt returns the emulator as `booted` (and does
**not** issue `adb emu kill`) when PM stays slow. All 28 boot-device tests pass.

## Out of scope / follow-ups (found during the same Linux session)

- `simulator-server` (Rust) **panics** (`android_emulator.rs:104 "should be
  running"`) when asked to attach to a pre-existing emulator launched with
  `-read-only` that never registered gRPC discovery. Better to return a clear
  "can't attach, reboot via Argent" error. Separate (cross-language) change.
- `screenshot-diff` with `captureCurrent: true` fails reproducibly with
  `"server response missing url or path"` while plain `screenshot` and
  file-vs-file diff both work.
- `react-profiler-component-source` returns `found: false` for valid top-level
  `export function` components.

### DIFF
--- a/packages/skills/skills/argent-android-emulator-setup/SKILL.md
+++ b/packages/skills/skills/argent-android-emulator-setup/SKILL.md
@@ -26,4 +26,4 @@ Pass the Android serial as `udid` to the unified interaction tools — `gesture-
 - Serials are the adb device id. iOS UDIDs and Android serials are not interchangeable, but you do NOT need to tell the tools which platform — dispatch is automatic.
 - `describe` on Android returns a shallower tree than iOS (no accessibility-service equivalent), but covers most tap-target discovery.
 - `reinstall-app` on Android always installs with `-g` so first-launch runtime permissions are pre-granted.
-- To kill the emulator when you're done, run `adb -s <serial> emu kill` from a shell.
+- To stop the emulator, run `adb -s <serial> emu kill` from a shell (clean shutdown). Never `pkill -9`/`kill -9` qemu — a hard kill leaves the userdata image dirty, after which cold boots can hang for many minutes doing recovery (runaway writes, `boot_completed` never flips). If an image gets into that state, boot once with `-wipe-data` to reset it.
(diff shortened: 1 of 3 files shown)


latekvo:
Partially overlaps #285. For now leaving both open, will inspect more closely once either gets merged.

filip131311 (approved):

PR #312, opened 2026-06-08
fix(native-profiler): fall back to host all-processes Time Profiler when the simulator Instruments tap is broken

filip131311:
## Problem

On Xcode 26.x / Instruments 16, `xctrace` can no longer package a `--device <udid> --attach <app>` **simulator** trace — `coreprofilesessiontap` fails during packaging. This breaks native profiling on the simulator entirely: `native-profiler-start` either errors out or produces an unusable trace.

## Fix

When the device-attach recording fails **on a simulator**, transparently fall back to a host `--all-processes` Time Profiler recording and scope the results to the app's host PID.

Because Leaks/Allocations cannot target "All Processes", this fallback path is **CPU-only by construction** — hangs/leaks are intentionally not captured, and their absence is no longer surfaced as an export failure. Physical devices keep the original behavior (their processes aren't host processes, so there's nothing to fall back to — the original error is surfaced).

## What changed

- **`native-profiler-start`** — resolve the app's host PID, retry device-attach (bounded, for cold-start races), then fall back to a host-wide Time Profiler scoped to that PID. Clears stale `.trace` bundles between attempts. Returns `mode` + `processFilterPid`.
- **`pipeline`** — `filterSamplesByPid` scopes a host-wide trace down to the app using the `(AppName, pid: N)` thread-fmt tag (anchored on the closing paren so `111` never matches `1110`).
- **`native-profiler-analyze`** — emits an **"Analysis Inconclusive"** report (never "All clear") when the only data source produced no readable / app-matching samples, so an empty fallback trace isn't mistaken for a clean run. Suppresses hangs/leaks "export failed" noise in fallback mode.
- **metadata sidecar** (`metadata.ts`) — persists `mode` / `processFilterPid` / `appProcess` next to the trace, so a *different* tool-server process can reconstruct app-scoped analysis on `profiler-load load_native`.
- **`export`** — `ExportOptions.keys` restricts export to CPU in fallback mode.
- **`native-profiler-stop`** — writes the sidecar and threads mode/pid through the result.

## Tests

- `host-fallback.test.ts` — PID scoping (drops unrelated host processes) + inconclusive reporting when CPU is empty/unreadable.
- `stop-recovery.test.ts` — updated for the new `exportIosTraceData` signature.

### DIFF
--- a/packages/tool-server/src/blueprints/native-profiler-session.ts
+++ b/packages/tool-server/src/blueprints/native-profiler-session.ts
@@ -35,6 +35,18 @@ export interface NativeProfilerParsedData {
   memoryLeaks: MemoryLeak[];
 }
 
+/**
+ * How the active (or last) recording was captured.
+ *  - `device-attach`: the normal path — xctrace `--device <udid> --attach <app>`,
+ *    full fidelity (CPU + hangs + leaks).
+ *  - `host-all-processes`: fallback used when the simulator-targeted Instruments
+ *    tap is broken (Xcode 26.x / Instruments 16 cannot package `--device`
+ *    simulator traces — `coreprofilesessiontap` fails). We record the host with
+ *    `--all-processes` Time Profiler instead and scope results to the app's host
+ *    PID. CPU-only: Leaks/Allocations cannot target "All Processes".
+ */
+export type NativeProfilerRecordingMode = "device-attach" | "host-all-processes";
+
 export interface NativeProfilerSessionApi {
   deviceId: string;
   appProcess: string | null;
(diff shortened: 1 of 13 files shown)


filip131311:
Closing in favor of #333, which takes a cleaner approach to the attach problem: attach by **PID** instead of `CFBundleExecutable`.

Root cause clarified — Xcode 26.5's `xctrace --attach` matches the app **display name**, not `CFBundleExecutable` (verified against the 26.3 toolchain), which is a *separate* regression from the `coreprofilesessiontap` finalize hang this PR was targeting.

Note for whoever revisits this: #333 fixes **attach** only. The host `--all-processes` Time Profiler fallback prototyped here is still the only thing that produces data on stock 26.5 (the finalize hang remains). If we want working CPU capture on 26.5, that fallback (rebased onto the current split-platform structure and using `--attach <pid>` for its device-attach tier) should be revived as a follow-up.

PR #316, opened 2026-06-09
fix(tool-server): make debugger-component-tree work on New Arch / multi-renderer apps

CreatlV:
## Summary

`debugger-component-tree` returns **"No visible components found on screen"** on React Native **New Architecture (Fabric / bridgeless)** apps that also use a secondary React reconciler such as `react-native-skia` — even when the app is fully rendered and CDP is connected. This PR fixes the two root causes so the tool returns the real component tree with names, testIDs, and accurate tap coordinates.

## Root cause

The injected fiber-walk script (`packages/tool-server/src/utils/debugger/scripts/component-tree.ts`) made two assumptions that don't hold on modern RN:

1. **Hardcoded renderer id.** `hook.getFiberRoots(1)` assumes the React Native renderer is DevTools renderer id `1`. Libraries that ship their own React reconciler (`react-native-skia`, `react-native-svg`, `react-three-fiber`, …) register a renderer too and frequently take id `1`. On such apps `getFiberRoots(1)` returns only that library's roots (e.g. Skia shapes), so the walk never reaches the app UI → `No visible components`.

2. **`nativeFabricUIManager.measure` under bridgeless.** Layout was read via `nativeFabricUIManager.measure(stateNode.node, …)`. Under bridgeless, `globalThis.nativeFabricUIManager` is an **empty object** (no `.measure`), so every measure threw and all rects became `null` — components were then filtered out (and the few kept by testID reported no usable rect).

## Fix

- **Renderer-agnostic root selection:** enumerate `hook.renderers`, collect roots from every renderer, and walk the mounted root with the largest fiber subtree (the real app UI) instead of hardcoding id 1.
- **Fabric layout via the public host instance:** measure through `stateNode.canonical.publicInstance.measureInWindow()` (a `ReactNativeElement`) — the **same approach this repo's `inspect-at-point.ts` already uses** — falling back to `nativeFabricUIManager.measure(shadowNode, …)` when that global is functional (non-bridgeless).
- **Hang guard:** a per-measure 1200 ms timeout so `Promise.all` can't stall on a measure callback that never fires (detached / off-screen nodes).

No change to the output schema or the Paper (old-architecture) path.

## How to reproduce

On a Fabric + bridgeless RN app that also renders `react-native-skia`:

```
argent run debugger-connect       --device_id <udid> --port <metro>
argent run debugger-component-tree --device_id <udid> --port <metro>
# before: "No visible components found on screen."
```

The DevTools hook reports multiple renderers — e.g. `getFiberRoots(1)` is the Skia reconciler (only canvas nodes), while the real app UI lives in a different react-native renderer that the hardcoded id never reaches.

## Testing

- Full `npm run build` (`tsc --build`, project references) passes; the modified file typechecks under `strict`.
- Verified live against a **React 19.2.0 / Fabric (bridgeless)** app whose DevTools hook exposes 3 renderers (id 1 = `react-native-skia` `0.0.1`; id 3 = `react-native` `19.2.0` holding the real ~2,900-fiber UI tree).

Before (unpatched): `No visible components found on screen.` — even with `--includeSkipped` and after forcing a fresh commit.

After (this change) — names + testIDs + accurate coords (example genericized):

```
Screen: 440x956
OnboardingQuestion "<question text>"            (tap: 0.50,0.17)
  SelectableCard [testID=onboarding.option.yes]  (tap: 0.50,0.84)
    Text "Yes"                                    (tap: 0.50,0.84)
  SelectableCard [testID=onboarding.option.no]    (tap: 0.50,0.93)
Pressable [testID=screen.back-button]             (tap: 0.09,0.09)   ← top-left ✓
TouchableOpacity [testID=dev-settings.trigger]    (tap: 0.06,0.97)   ← bottom-left ✓
```

Coordinates were spot-checked against the visible layout (back button top-left, an off-screen/hidden trigger bottom-left).

## Notes

`inspect-at-point.ts` has the same hardcoded `getFiberRoots(1)` / `renderers[0]` renderer assumption (its Fabric layout path already uses `publicInstance`, which is what motivated this fix). I scoped this PR to `component-tree`; happy to follow up on the inspect path separately.

### DIFF
--- a/packages/tool-server/src/utils/debugger/scripts/component-tree.ts
+++ b/packages/tool-server/src/utils/debugger/scripts/component-tree.ts
@@ -5,8 +5,14 @@
  * (testID, accessibilityLabel, visible text).
  *
  * Supports both Fabric (new architecture) and Paper (old architecture).
- * On Fabric, uses nativeFabricUIManager.measure with the shadow node.
- * On Paper, falls back to UIManager.measureInWindow with native tags.
+ * On Fabric, measures via the public host instance (ReactNativeElement)
+ * measureInWindow, falling back to nativeFabricUIManager.measure with the
+ * shadow node. On Paper, falls back to UIManager.measureInWindow with native tags.
+ *
+ * Renderer selection is renderer-agnostic: apps that use a secondary reconciler
+ * (react-native-skia, react-native-svg, etc.) register their own renderer, so we
+ * enumerate every renderer and walk the mounted root with the largest fiber
+ * subtree rather than assuming the React Native renderer is id 1.
  *
  * Measurement is async-safe: on Paper (where measureInWindow is async),
  * the script collects all candidates first, then batch-measures them via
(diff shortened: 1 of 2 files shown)


filip131311 (approved):
THank you for this I will merge it and i expect it will land in the next version <3

PR #317, opened 2026-06-09
fix(tool-server): select the correct renderer in debugger-inspect-element

CreatlV:
## Summary

Companion to #316. `debugger-inspect-element` has the same hardcoded-renderer assumption as the component-tree script, so it inspects the wrong React renderer on multi-renderer apps.

## Root cause

`makeInspectScript` (`packages/tool-server/src/utils/debugger/scripts/inspect-at-point.ts`) picks:

```js
var renderer = Array.from(hook.renderers.values())[0];
var roots = hook.getFiberRoots(1);
var root = Array.from(roots)[0];
```

On React Native **New Architecture** apps that also use a secondary React reconciler (`react-native-skia`, `react-native-svg`, `react-three-fiber`, …), that library registers its own renderer and frequently takes id `1`. `renderer` then points at the Skia/SVG reconciler and `getFiberRoots(1)` returns only its roots, so `renderer.rendererConfig.getInspectorDataForViewAtPoint(...)` runs against the wrong tree and returns no app component at the touch point.

## Fix

Enumerate `hook.renderers`, walk each renderer's roots, and select the **(renderer, root)** pair with the largest fiber subtree (the real app UI) before inspecting — then call `getInspectorDataForViewAtPoint` on that renderer. Falls back to the previous first-renderer behavior if enumeration yields nothing.

Unlike the component-tree script, the Fabric layout path here was already correct (it uses `hostFiber.stateNode.canonical.publicInstance`), so no measurement change is needed — this PR is purely the renderer/root selection.

## Testing

- Full `npm run build` (`tsc --build`) passes; the modified file typechecks under `strict`.
- Same environment as #316: a React 19.2 / Fabric (bridgeless) app whose DevTools hook exposes a `react-native-skia` renderer at id 1 alongside the real `react-native` UI renderer. With the fix, inspection resolves against the app renderer instead of the Skia one.

No change to the output schema or the Paper path.

### DIFF
--- a/packages/tool-server/src/utils/debugger/scripts/inspect-at-point.ts
+++ b/packages/tool-server/src/utils/debugger/scripts/inspect-at-point.ts
@@ -21,9 +21,37 @@
 export function makeInspectScript(x: number, y: number, requestId: string): string {
   return `(function() {
   var hook = window.__REACT_DEVTOOLS_GLOBAL_HOOK__;
-  var renderer = Array.from(hook.renderers.values())[0];
-  var roots = hook.getFiberRoots(1);
-  var root = Array.from(roots)[0];
+
+  // Pick the renderer + root that hosts the real app UI. Secondary reconcilers
+  // (react-native-skia, react-native-svg, ...) register their own renderer and
+  // often take id 1, whose roots contain only that library's nodes. Walk every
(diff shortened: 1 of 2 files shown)


filip131311 (approved):
thank you <3

PR #318, opened 2026-06-09
docs: add dynamic banner serving to readme

p-malecki:
Adds dynamic banner serving to the readme, server-side content swaps update banner without repo changes. Banners are currently hidden.


filip131311 (approved):
thank you

PR #325, opened 2026-06-11
feat: record AI client metadata for telemetry

hubgan:
## Summary

Adds a coarse, anonymous signal to tool telemetry: which AI coding tool is driving the MCP server (Claude Code, Codex, Cursor, Gemini, VS Code, Windsurf, Zed, opencode, Copilot, or other).

## How it works

1. Source of truth — the MCP `initialize` handshake clientInfo.name, read at runtime via `Server.getClientVersion()`. New `ai-identity.ts` in `@argent/telemetry` canonicalizes that raw name to a known slug. Unrecognized tools become `other` plus a bounded, sanitized free-form name.
2. Cross-process hop — telemetry lives in the tool-server, but `clientInfo` lives in the MCP server process. The MCP server forwards the identity as `X-Argent-AI-Client` / `X-Argent-AI-Client-Name` request headers.
3. Re-validation at the boundary — the tool-server re-checks both headers against the same allowlist/pattern the sanitizer enforces, so a misbehaving client can't inject arbitrary telemetry values. The free-form name is retained only when ai_client === "other".
4. Threading through events — `AiTelemetryProps` is mixed into the tool event types and merged in the registry listener and tool-server stop handler via the shared aiTelemetryFromMeta helper, so keys are omitted entirely when absent.

## Privacy

It records only which tool is connected. It is validated at the `mcp-server`, re-validated at the `tool-server`, and enforced a third time by the `sanitizer` allowlist before anything reaches PostHog. Path-like or oversized names are dropped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,12 +130,15 @@ argent init
 
 ## Privacy
 
-Argent does not collect or transmit any user data.
-No telemetry, no analytics, no crash reporting.
+Argent collects anonymous, opt-out usage and diagnostic telemetry to help us prioritise features and fix what breaks. It is minimal by design.
 
-- Argent integrates with your agent locally over MCP stdio.
-- Its internal tools are not reachable from outside your machine.
-- The only outbound network call we make is the version check against our public npm package, which sends no user data and fails gracefully if blocked.
+You can opt out at any time:
+
+```bash
+argent telemetry disable   # check status with: argent telemetry status
+```
+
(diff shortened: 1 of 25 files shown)


latekvo:
**Privacy-notice link 404s** (`README.md:141`)

The link points to `…/blob/main/TELEMETRY.md`, but the file added in this PR is `Telemetry.md`. GitHub blob paths are case-sensitive, so the link is a guaranteed 404 — the one place a user is pointed to in order to understand what telemetry collects is broken.

PR #328, opened 2026-06-12
feat: support arm64 Linux simulator-server binary

filip131311:
## Summary

The only Linux simulator-server binary argent ships today is x86_64 (`simulator-server-argent-linux` — verified: `ELF 64-bit … x86-64`). On arm64 Linux hosts (Lima/UTM VMs on Apple Silicon, arm cloud instances, Raspberry Pi-class devices) the package has no usable binary — the x86_64 ELF would only fail with `ENOEXEC` at spawn time.

This PR teaches the whole binary pipeline about a **host platform key**: `process.platform` everywhere, except arm64 Linux which maps to `linux-arm64`. macOS keeps a single `darwin/` dir because its binary is universal (lipo); Linux binaries are single-arch ELFs, so each arch gets its own platform-keyed directory. x86_64 Linux keeps the existing `linux` key for backward compatibility.

## Changes

- **`@argent/native-devtools-ios` resolver** — new exported `hostPlatformKey()`; `simulatorServerBinaryPath()` resolves `bin/linux-arm64/` on arm64 Linux and names the searched key in its error message.
- **`argent-simulator-server.cjs` dispatcher** (npm `bin` entry) — mirrors the same key (duplicated deliberately; the dispatcher ships as a standalone file).
- **`bundle-tools.cjs`** — `linux-arm64` added to the supported host platform dirs; copied into the package when staged, warn-and-skip otherwise (same as x86_64 Linux).
- **`scripts/download-simulator-server.sh`** — downloads `simulator-server-argent-linux-arm64` into `bin/linux-arm64/`; tolerates the asset being absent (warn + skip) until the upstream release ships it, so publishing keeps working in the meantime.
- **`scripts/dev.cjs`** — copies the current host's binary by key.
- **publish / publish-next / build-package-artifact workflows** — `linux-arm64` added to the binary inventory; gated by the existing `ARGENT_REQUIRE_LINUX_BINARY` flag (now covering both Linux keys via a `linux*` glob).

## Tests

- Resolver: new arch-aware suite — `linux`+`arm64` → `bin/linux-arm64`, `linux`+`x64` → `bin/linux`, `darwin`+`arm64` → `bin/darwin`, and the missing-binary error names `linux-arm64`.
- Dispatcher: end-to-end test faking `process.platform`/`process.arch` in a child process to exercise the real resolution path on any host.
- Full suite: 146 files / 1651 tests pass. `tsc --build` clean.
- `download-simulator-server.sh` exercised against the live `radon-main` release: downloads darwin + linux, warn-and-skips the not-yet-published arm64 asset.

## Dependencies

Companion PR software-mansion/radon#126 adds the `ubuntu-24.04-arm` build leg that publishes `simulator-server-argent-linux-arm64` to `simulator-server-releases`. This PR is safe to merge first (the download script and verify steps tolerate the missing asset); arm64 hosts start working once a release containing the new asset is cut.

### DIFF
--- a/.github/workflows/build-package-artifact.yml
+++ b/.github/workflows/build-package-artifact.yml
@@ -54,7 +54,7 @@ jobs:
             p="packages/argent/bin/${plat}/simulator-server"
             if [[ -x "$p" ]]; then
               size=$(wc -c <"$p")
-              echo "  ok ${plat}: ${p} (${size} bytes)"
+              echo "  ok ${plat}: ${p} (${size} bytes; $(file -b "$p"))"
             else
               echo "  missing ${plat}: ${p}"
               if [[ "$plat" == "darwin" ]]; then missing=1; fi
--- a/.github/workflows/build-package-artifact.yml
+++ b/.github/workflows/build-package-artifact.yml
@@ -50,7 +50,7 @@ jobs:
           set -e
           missing=0
           echo "simulator-server inventory:"
-          for plat in darwin linux; do
+          for plat in darwin linux linux-arm64; do
             p="packages/argent/bin/${plat}/simulator-server"
             if [[ -x "$p" ]]; then
               size=$(wc -c <"$p")
@@ -58,7 +58,7 @@ jobs:
             else
               echo "  missing ${plat}: ${p}"
               if [[ "$plat" == "darwin" ]]; then missing=1; fi
-              if [[ "$plat" == "linux" && "${ARGENT_REQUIRE_LINUX_BINARY}" == "1" ]]; then missing=1; fi
+              if [[ "$plat" == linux* && "${ARGENT_REQUIRE_LINUX_BINARY}" == "1" ]]; then missing=1; fi
             fi
           done
           [[ "$missing" == "0" ]] || { echo "::error::required simulator-server binary missing"; exit 1; }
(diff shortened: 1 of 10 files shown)


filip131311:
**End-to-end validation against real binaries** — the companion workflow (software-mansion/radon#126) was dispatched from its branch and published a complete test release including `simulator-server-argent-linux-arm64`. Ran `scripts/download-simulator-server.sh` against that tag:

```
Downloading simulator-server-argent-macos → bin/darwin/simulator-server
  ✓ arch ok: Mach-O universal binary with 2 architectures: [x86_64…] [arm64]
Downloading simulator-server-argent-linux → bin/linux/simulator-server
  ✓ arch ok: ELF 64-bit LSB pie executable, x86-64, …
Downloading simulator-server-argent-linux-arm64 → bin/linux-arm64/simulator-server
  ✓ arch ok: ELF 64-bit LSB pie executable, ARM aarch64, …
```

The downloaded arm64 binary was also executed on an aarch64 Linux VM (Lima): `--help` and `fingerprint` both succeed.

The `✓ arch ok` lines come from 150c0f23, which adds a `file(1)`-based architecture gate to the download script: a present-but-wrong-arch asset now fails the download hard (a wrong-arch binary is worse than a missing one — the resolver would pick it and the user would get a bare ENOEXEC). Missing assets keep the existing warn-and-skip behavior. The publish workflows' inventory also logs `file` output per binary.

PR #337, opened 2026-06-15
fix(ios-profiler): make memory-leak reporting honest (drop fake --hal flag, demote unattributed leaks, tests)

latekvo:
## Context

This began as "add native iOS memory profiling," but the feature **already exists** — the iOS native profiler captures and reports memory **leaks** end-to-end (export → parse → aggregate → render → `leak_stacks`). An audit plus real-device verification then surfaced genuine defects, fixed here.

> **Course-correction worth noting:** an early static read suspected the leak parser/xpath were broken. A real capture disproved that — xctrace's `Leaks` *track-detail* export uses self-closing attribute rows (a different shape from the CPU/hangs *schema tables*), and the parser matches it correctly. So this PR does **not** touch parsing/xpath; the real issues are below.

### 1. Drop the fabricated `--hal` flag (`export.ts`)
`xctrace export` has no `--hal` option — verified against xctrace 16 and every subcommand's help. On xctrace ≥ 15 the exporter passed it, so the **first** export attempt always failed and silently fell back to the plain export: a wasted failing subprocess per leak export, plus a false "required by newer xctrace versions" claim in code + docs. Leaks export exactly like hangs (one `xctrace export --xpath` of the `Leaks` track detail), so the special case and the now-unused `getXctraceVersion` are removed.

### 2. Stop flagging unattributed leaks as RED
Argent records via `xctrace --attach`, which gives the Leaks/Allocations instruments **no malloc-stack history**, so most leaks come back unattributed: responsible frame `<Call stack limit reached>`, empty library, generic `Malloc N Bytes`/`xpc_*` types. (CPU/hang stacks symbolicate fine; leaks largely don't under attach.) The pipeline marked them **all** RED, so idle Safari reported 19 RED "leaks" (1646 system allocations) — benign noise as confirmed bugs.

- `aggregateLeaks` tags each group `attributed` (real responsible frame) and sets severity RED (attributed) / **YELLOW** (unattributed); attributed leaks sort first. New `isLeakAttributed()` centralises the check.
- The report shows attributed leaks in the full table; unattributed ones collapse into one low-confidence YELLOW summary with guidance. `leak_stacks` and the suggested-fix / next-step hints fire only for attributed leaks.

### 3. Tests + docs
- New `ios-leak-pipeline.test.ts` over `parseLeaksXml` + `aggregateLeaks` (the leak path had **zero** coverage), built from a real xctrace 16 capture — pins the self-closing-attribute-row format and the attributed/unattributed severity split.
- Reference doc + `argent-native-profiler` skill: corrected the `--hal` claim, documented the `--attach` attribution limitation, and updated the severity rules.

## Verification

Ran the real pipeline + render over actual xctrace-16 captures (Argent template, `--attach`) on three apps:

- **Bluesky** (social-app, complex app, ~85s onboarding/sign-in churn): **9 leak groups, exactly 1 genuinely attributed** — `hermes::vm::JSTypedArrayBase::createBuffer` (library `hermes`) → rendered RED in the table with a fix suggestion + `leak_stacks` next-step; the other 8 (345 objects, 642 KB) collapse into one YELLOW line. **This mixed case is the key proof:** a blanket filter would have *hidden a real Hermes leak*, so demote-don't-drop is the correct design.
- **Idle Safari:** 1646 rows → 19 groups, all unattributed → one YELLOW summary (was 19 RED rows).
- **Active pokemon RN app (~30s):** 1 unattributed group → one YELLOW summary (was 1 RED); CPU hotspots symbolicated normally, confirming the capture itself works.
- `ios-leak-pipeline.test.ts`: 6/6 pass. Changed files type-clean + Prettier-clean. CI green.


packages/tool-server/src/tools/profiler/query/profiler-stack-query.ts
```diff
@@ -277,15 +277,29 @@ function renderLeakStacksIos(
   const totalBytes = sorted.reduce((s, l) => s + l.totalSizeBytes, 0);
   const totalCount = sorted.reduce((s, l) => s + l.count, 0);
 
+  const unattributedCount = sorted.filter((l) => !l.attributed).length;
+
   const lines: string[] = [
```
hubgan, line 292: On line 275 there's this code:

```
const sorted = [...filtered].sort((a, b) => b.totalSizeBytes - a.totalSizeBytes).slice(0, topN);
```

aggregateLeaks hands this list in attributed-first order, but the line re-sorts by size only and slices to `top_n`. Unattributed system noise is larger than a real app leak, so with > `top_n` groups a smaller attributed leak gets truncated out entirely. I think we should sort it by attributed-first and then by size.

latekvo: Good catch — fixed in 807652eb. `renderLeakStacksIos` now sorts attributed-first and then by size before the `top_n` slice, so a smaller attributed leak can no longer be truncated out by larger unattributed system noise. Added a regression test (`profiler-leak-stacks.test.ts`) covering both the truncation case and the within-group size ordering.

hubgan (approved):

PR #338, opened 2026-06-15
fix(profiler): make Android native-profiler analysis accurate

latekvo:
## Summary

Found while exercising the new Android (Perfetto) native profiler end-to-end on an emulator: the **measurement** is solid, but the **interpretation** layer didn't account for the emulator environment or its own signals, so it surfaced misleading findings/advice.

This makes the Android analysis honest:

- **CPU hotspots:** on the emulator the hottest frames are the goldfish/QEMU GPU-transport pipe and Linux kernel syscalls (`goldfish_pipe_read_write`, `gup_fast_fallback`, `do_syscall_64`), not app code. They were presented as actionable app hotspots with `reduce view hierarchy depth` advice. Now they're classified and labelled as emulator/kernel overhead, with advice that says so (and to re-profile on a real device); real app frames get actionable advice.
- **UI hangs:** advice was always `move heavy work to background queue`, even when the printed main-thread state breakdown showed the thread was **sleeping** (state `S`) for the whole hang — i.e. a GPU/vsync wait, not CPU work. Advice is now derived from the state breakdown (blocked / runnable / executing).
- **`hang_stacks`:** an off-CPU (sleeping) hang has no on-CPU stack, so the section was silently empty. It now explains the empty result and points at the state breakdown.

## Why

The advice directly contradicted the tool's own state-breakdown data, and the CPU section was dominated by emulator noise — both erode trust in the report.

## Test plan

- New `test/android-perfetto/analysis-advice.test.ts` and `hang-stacks-offcpu.test.ts`.
- `npm test -w @argent/tool-server` — all pass; `tsc --build` + `typecheck:tests` clean.

> Part of a set of native-profiler fixes found in the same session. Touches the shared `render.ts` advice paths and the Android `hang_stacks` renderer.


packages/tool-server/src/utils/profiler-shared/native-frame-class.ts
```diff
@@ -0,0 +1,81 @@
...
+  /\bvfs_(read|write|fsync)\b/,
+  /\bgup_/,
+  /get_user_pages/,
+  /\bhandle_mm_fault\b/,
+  // pthread / low-level lock internals — usually the leaf of a lock/futex wait.
+  /__pthread_internal/,
+  /__pthread_mutex/,
+  /__lll_/,
+  /\bfutex_/,
+];
```
hubgan, line 78: `SYSTEM_FRAME_PATTERNS` misses some arm64 kernel frames. The patterns are x86-centric, so on an arm64 emulator several kernel leaves slip through and get classified as app. In the result we're getting misleading advice.

hubgan (changes requested):

packages/tool-server/src/utils/profiler-shared/native-frame-class.ts
latekvo: fixed in [19c15f5](https://github.com/software-mansion/argent/pull/338/commits/19c15f52c2d7cf1acf82a3665a052f8d96b296f7), [89c5584](https://github.com/software-mansion/argent/pull/338/commits/89c5584c900fb13342ccf75cfae432e86ce8e700)

hubgan (approved):

PR #339, opened 2026-06-15
fix(profiler): demangle native C++ symbols in profiler reports

latekvo:
## Summary

`profiler-stack-query` Android stacks come straight from perf, which stores frame names **mangled** — so `function_callers` / `hang_stacks` output reads like:

```
_ZN16GrDrawingManager5flushE6SkSpan... <- _ZL25libutil_thread_trampolinePv.__uniq.2265... <- __start_thread
```

This adds a conservative, best-effort Itanium demangler and applies it when rendering those stacks (display-only — SQL matching upstream still uses the raw mangled names). The same stack now reads:

```
GrDrawingManager::flush <- libutil_thread_trampoline <- __start_thread
```

## What it handles

- Nested names (`_ZN...E`) and plain function names; the argument list is dropped.
- Strips LLVM/GCC internal suffixes (`.__uniq.*`, `.llvm.*`, `.part.*`, `.cold`).
- **Conservative:** templates, substitutions, and kernel C symbols it can't confidently parse are returned **verbatim** — it never corrupts a name.

## Test plan

- New `test/android-perfetto/demangle.test.ts` (10 cases incl. fallbacks).
- Updated `function-callers.test.ts` to expect the demangled leaf in the substring-match note.
- `npm test -w @argent/tool-server` — all pass; build + `typecheck:tests` clean.

> Part of a set of native-profiler fixes. Note: this and #338 both touch the `hang_stacks` sample renderer — whichever merges second needs a trivial rebase.

### DIFF
--- a/packages/tool-server/src/utils/ios-profiler/render.ts
+++ b/packages/tool-server/src/utils/ios-profiler/render.ts
@@ -243,11 +243,11 @@ function renderFullReport(
       if (b.topCallChains && b.topCallChains.length > 0) {
         lines.push(`**Call chains:**`);
         for (const { chain, count } of b.topCallChains) {
-          lines.push(`- (${count}×) \`${chain.join(" > ")}\``);
+          lines.push(`- (${count}×) \`${chain.map(demangleSymbol).join(" > ")}\``);
         }
         lines.push(``);
       } else if (b.topCallChain.length > 0) {
-        lines.push(`**Call chain:** \`${b.topCallChain.join(" > ")}\``);
+        lines.push(`**Call chain:** \`${b.topCallChain.map(demangleSymbol).join(" > ")}\``);
         lines.push(``);
       }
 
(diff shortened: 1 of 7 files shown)


packages/tool-server/src/utils/ios-profiler/render.ts
```diff
@@ -236,7 +237,7 @@ function renderFullReport(
       : cpuHotspots;
     for (const b of hotspotDetailSlice) {
       lines.push(``);
-      lines.push(`### \`${b.dominantFunction}\` (${b.thread})`);
+      lines.push(`### \`${demangleSymbol(b.dominantFunction)}\` (${b.thread})`);
       lines.push(``);
 
       if (b.topCallChains && b.topCallChains.length > 0) {
```
hubgan, line 243: Call chains at 246 and 250 line are not demangled

latekvo: thanks for pointing this out, fixed in [7cde9d2](https://github.com/software-mansion/argent/pull/339/commits/7cde9d2faaf816ae264a3513ef5f521763805acf)

hubgan (approved):

PR #340, opened 2026-06-15
fix(profiler): validate app_process and flag stale Android traces

latekvo:
## Summary

Two native-profiler session-correctness gaps found while testing on Android:

- **`native-profiler-start` accepted any `app_process`.** Perfetto scopes `linux.perf` to the cmdline but never checks it exists, so a typo'd / not-installed package recorded an empty trace whose failure only surfaced at **analyze** time — and was misattributed to "non-debuggable / missing `<profileable>`". An explicit `app_process` is now validated up front (installed package or running process), failing fast with the real reason. Auto-detection is unchanged.
- **No staleness signal.** The profiler session lives for the whole tool-server run, so `native-profiler-analyze` happily reported on a trace recorded days earlier (in an unrelated working session) with nothing flagging it. Analyze now emits a "stale trace" warning when the recording's wall-clock start is clearly from an earlier session.

## Test plan

- New `test/android-perfetto/session-correctness.test.ts` (freshness helper, app_process validation incl. device-error path, render pass-through).
- `npm test -w @argent/tool-server` — all pass; build + `typecheck:tests` clean.

> Scoped to Android (the platform tested); the staleness helper is shared and could be wired into the iOS analyze path too. Part of a set of native-profiler fixes.


packages/tool-server/src/tools/profiler/native-profiler/platforms/android.ts
```diff
@@ -193,5 +206,9 @@ export async function analyzeNativeProfilerAndroid(
     payload,
     traceFile: hostTracePath,
     exportErrors: pipelineResult.exportErrors,
+    // wallClockStartMs is the recording's start time (set at start, persisted in
+    // the metadata sidecar, restored by profiler-load). A large gap to "now"
+    // means we're analyzing a trace from an earlier session, not a fresh capture.
+    freshnessNote: formatTraceFreshness(api.wallClockStartMs, Date.now()) ?? undefined,
   });
```
hubgan, line 213: it seems that iOS analyze path uses the same api object and the same renderer, but doesn't pass the note. I think freshnessNote should be also added to `ios.ts` file.

latekvo: Fixed in 22a1537f9110c09acf793a8be0adf1370910d1b3.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
```diff
@@ -456,5 +457,9 @@ export async function analyzeNativeProfilerIos(
     payload,
     traceFile: api.traceFile,
     exportErrors,
+    // wallClockStartMs is the recording's start time, set at start on the
+    // session that persists for the whole tool-server run. A large gap to "now"
+    // means analyze is reusing a trace from an earlier capture, not a fresh one.
+    freshnessNote: formatTraceFreshness(api.wallClockStartMs, Date.now()) ?? undefined,
   });
```
hubgan: Not sure if that should be part of this PR but definitely worth noticing, `wallClockStartMs` isn't persisted to disk on iOS the way Android does it. The iOS never persists it and there's no iOS sidecar to read. So loading an old iOS session leaves `wallClockStartMs` at `null`, which results in `formatTraceFreshness` returning null.

hubgan (approved):

PR #343, opened 2026-06-15
fix(profiler): enable native stack-query and combined-report on Android

latekvo:
## Problem

`profiler-stack-query` and `profiler-combined-report` reject Android devices with:

> Tool '…' is not supported on android emulator (no android support declared)

even though **both tools already contain complete, working Android implementations**:

- `profiler-stack-query` has `executeAndroid` (queries the Perfetto `.pftrace` via `runAndroidStackQuery`) and an `api.platform === "android"` dispatch branch.
- `profiler-combined-report` has an `api.platform === "android"` branch using `loadAndroidCombinedData` and `buildPerfettoAnchor`.

The only thing gating Android was the `capability` declaration: `{ apple: { simulator: true, device: true } }` with no `android` block. `assertSupported` (`packages/tool-server/src/utils/capability.ts`) throws `UnsupportedOperationError` for any platform without a declared block — so it rejected the device *before* the Android code path could run.

This was a broken workflow: the sibling `native-profiler-analyze` (which **does** declare Android support) and `profiler-load`'s `load_native` output both tell Android users to drill down with `profiler-stack-query` — which then refused.

## Fix

Add `android: { emulator: true, device: true, unknown: true }` to the `capability` of both tools, matching `native-profiler-analyze`. The existing `apple` block is unchanged. No code in the handlers needed to change — the Android paths were already there.

### `leak_stacks` on Android

`leak_stacks` mode is iOS-only. `executeAndroid` already guards it: it returns a clean guidance message (`"Memory leak detection is not yet supported on Android. … adb shell am dumpheap"`) *before* touching the trace, so enabling the capability does not let `leak_stacks` reach any crashing code path. A test pins this behavior.

## Tests

New `packages/tool-server/test/android-perfetto/profiler-query-android-capability.test.ts`:
- both tools are supported on Android emulator / device / unknown,
- both tools are still supported on Apple simulator + device,
- `profiler-stack-query` `leak_stacks` on Android returns the iOS-only guidance message without querying the trace.

## Verification

- `npm test -w @argent/tool-server` — 131 files, 1158 tests passed.
- `npm test --workspaces` — all package suites green.
- `npm run build` (`tsc --build`) — clean.
- `prettier --write` on changed files — no changes.

### DIFF
--- a/packages/tool-server/src/tools/profiler/combined/profiler-combined-report.ts
+++ b/packages/tool-server/src/tools/profiler/combined/profiler-combined-report.ts
@@ -25,7 +25,7 @@ import { readCpuProfile, readCommitTree } from "../../../utils/react-profiler/de
 
 const zodSchema = z.object({
   port: z.coerce.number().default(8081).describe("Metro server port"),
-  device_id: z.string().describe("iOS Simulator or device UDID"),
+  device_id: z.string().describe("iOS Simulator/device UDID or Android serial"),
 });
 
 interface HangCommitCorrelation {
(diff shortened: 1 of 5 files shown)


hubgan:
It doesn't seem to be linked to this PR but native-profiler is not working on iOS.

- native-profiler-start returns {status: "recording"} and a .trace bundle with a populated Trace1.run/ is written to disk so capture starts and data is recorded.
- The xctrace process then exits on its own
- Every subsequent native-profiler-stop fails with: `No active native profiling session found. Call native-profiler-start first.`
- Because stop never runs, no _raw_*.xml is exported, so the entire downstream chain is dead on iOS: native-profiler-analyze, profiler-stack-query (all modes), and the native half of profiler-combined-report.

I guess that should be another PR, but I wanted to address it here.

The android changes looks good.

filip131311 (approved):
Nice catch

PR #344, opened 2026-06-15
fix(debugger): make inspect-element work on old-arch RN (RN 0.81 bridge)

latekvo:
## Problem

`debugger-inspect-element` returned `{"error":"no host fiber"}` at every coordinate on a React Native 0.81.5 app (Hermes, old architecture / legacy bridge, React Compiler). Verified live on a real device.

## Root causes (three layers, each verified live)

1. **Host-fiber resolution** - `findHostFiber` only recognised old-arch host fibers via `stateNode.canonical`, but RN 0.81 bridge host fibers expose `stateNode._nativeTag` with no `canonical`, so it found nothing and threw "no host fiber". Fixed by also matching `_nativeTag` (mirrors the working `component-tree.ts` getHostInfo). On the real tree: 87/87 host fibers had `_nativeTag` and 0 had `canonical`.
2. **Wrong anchor** - even with the host fiber resolved, the first host fiber is a sibling subtree that does not contain the touch point, so the renderer's native `findSubviewIn` resolved nothing for *every* coordinate (touchedViewTag undefined). Fixed by anchoring at the **FiberRoot container** (its `containerTag`), the universal ancestor. Fabric and older Paper (no containerTag) keep the host-fiber anchor.
3. **Changed data shape** - RN 0.81 no longer returns `data.closestInstance`/`hierarchy`; it returns `data.componentStack` (a React stack string). The old extraction produced empty results. Now parsed into items; bundle source locations are symbolicated by the tool layer exactly like `_debugStack` frames.

## Result

Live on a real RN 0.81 app, inspecting a point now returns the real component hierarchy (170 frames: `SplashScreen`, `ErrorBoundary`, `LoggedOut`, `NativeStackNavigator`, `HomeTabNavigator`, `SceneView`, `NavigationProvider`, ...) with source locations.

## Safety / scope

- Fabric (new-arch) and iOS paths unchanged; canonical-publicInstance handling is byte-for-byte preserved (incl. the lazily-null Fabric case).
- Distinct from #284 (symbolication fallback - a different, later failure mode).
- Unit tests added for: `_nativeTag` host-fiber resolution, container anchoring, componentStack parsing, and the Fabric null-publicInstance no-regression case. The test harness evaluates the generated script, so it exercises the real emitted code.
- Validated end-to-end against a live emulator, not only unit tests.

### DIFF
--- a/packages/tool-server/src/utils/debugger/scripts/inspect-at-point.ts
+++ b/packages/tool-server/src/utils/debugger/scripts/inspect-at-point.ts
@@ -168,15 +168,19 @@ export function makeInspectScript(x: number, y: number, requestId: string): stri
     return null;
   }
 
-  // RN 0.81+ getInspectorDataForViewAtPoint no longer returns a walkable
-  // data.closestInstance; instead it returns data.componentStack -- a React
-  // component-stack STRING whose lines look like:
+  // On the RN runtimes observed live (RN 0.81.x + Hermes, legacy bridge),
+  // getInspectorDataForViewAtPoint does not return a walkable data.closestInstance
+  // -- closestInstance is absent and data.componentStack (a React component-stack
+  // STRING) carries the result instead. Lines look like:
   //   "    at View (http://.../index.bundle//...:10685:19)"
   //   "    at RCTView (<anonymous>)"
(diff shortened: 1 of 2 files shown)


filip131311 (approved):
this works too

PR #345, opened 2026-06-16
fix(installer): soften host project engine gate for npx skill install (#298)

### LINKED ISSUE #298: Running `argent init` in bluesky throws on skill install
Running `argent init` in bluesky throws errors preventing skills from being installed. 

Reproduced on Linux.

Will add more details - don't have the linux machine on hand at this moment.

_Originally posted by @latekvo in https://github.com/software-mansion/argent/issues/292#issuecomment-4613947112_

filip131311:
Fixes #298.

## Problem

`argent init` installs skills by shelling out to `npx skills add …`. Because `npx` is `npm exec`, npm first evaluates the nearest `package.json`'s `engines` / `devEngines` gate against the user's machine **before** running anything.

[Bluesky's `package.json`](https://github.com/bluesky-social/social-app/blob/main/package.json) declares:

```json
"devEngines": { "runtime": { "name": "node", "version": "^24.15.0", "onFail": "download" } }
```

When the user's Node doesn't satisfy `^24.15.0` (common on Linux), npm aborts with **`EBADDEVENGINES` (exit 1)** — even with `onFail: "warn"`/`"download"` (npm doesn't honor `"download"` and treats a runtime mismatch as a hard error). The skills CLI never runs, so **skills don't install**. On a machine running Node 24.x it only prints a warning, which masked the failure — that's why it reproduced on Linux but not for everyone.

This breaks all three argent code paths that go through `npx`: `isSkillsCliAvailable()`, the `init.ts` skill install, and `refreshArgentSkills()`. (`npm install -g` for argent's own self-update is **not** affected — only `npm exec`/`npx` consults the host project's gate.)

> #292 only suppressed the cosmetic npm stderr in `getLatestVersion`; this is the "larger issue" the reporter flagged — the actual install failure.

## Why `--force`

npm has **no flag to disable a `devEngines` check**. `engine-strict` governs only the `engines` field, there is no `--ignore-engines`, and the npm maintainers have not added a skip option ([npm/cli#8003](https://github.com/npm/cli/issues/8003), [npm/cli#7888](https://github.com/npm/cli/issues/7888)). The only documented lever that lets the command proceed is **`--force`**, which downgrades the gate from a hard error to a non-fatal warning.

## Fix

Add a small `withNpmForce()` helper and prepend `--force` to every argent `npx` invocation. It is **scoped to argent's own skills commands** (it is never applied to the user's project installs), and argent already discards the child's stderr on success, so the `npm warn using --force` notice never reaches the user. The displayed and manual-fallback commands stay clean.

- `src/utils.ts` — `withNpmForce()` helper; `isSkillsCliAvailable()` probe gains `--force`.
- `src/skills.ts` — `refreshArgentSkills` add / fallback-add / remove wrapped.
- `src/init.ts` — skill install spawned with `--force`.
- Tests updated in `test/utils.test.ts` and `test/skills.test.ts`.

Net: **+34 / −9**, no temp dirs, no new runtime state.

## Verification

- Full installer suite: **263 tests pass**; `tsc --noEmit` clean.
- End-to-end against the **real `bluesky-social/social-app` repo**, driving the compiled `dist`, on **Node 22.22.3 (npm 10)** and **Node 24.16.0 (npm 11)**:

  | Node | before (`npx skills add`) | after (`npx --force skills add`) |
  |---|---|---|
  | 22.22.3 (< floor) | exit 1, `EBADDEVENGINES`, **0 skills** | exit 0, **11 skills** ✅ |
  | 24.16.0 (≥ floor) | exit 0, 11 skills (noisy warn) | exit 0, **11 skills** ✅ |

## Test plan

- [x] `npm run test -w packages/argent-installer`
- [x] Manual: real Bluesky repo on Node < 24.15 installs skills without `EBADDEVENGINES`

### DIFF
--- a/packages/argent-installer/src/init.ts
+++ b/packages/argent-installer/src/init.ts
@@ -25,6 +25,7 @@ import {
   globalInstallCommand,
   formatShellCommand,
   resolveProjectRoot,
+  withNpmForce,
   type ShellCommand,
 } from "./utils.js";
 import { refreshArgentSkills, formatSkillRefreshSummary } from "./skills.js";
@@ -495,9 +496,13 @@ export async function init(args: string[]): Promise<void> {
       skillsArgs.push("--skill", "*", "-y");
     }
 
-    const npxArgs = offlineWithCache ? ["--no-install", ...skillsArgs] : skillsArgs;
+    const baseArgs = offlineWithCache ? ["--no-install", ...skillsArgs] : skillsArgs;
+    // The spawned command carries `--force` to soften the host project's npm
+    // engine gate (see withNpmForce / issue #298). The displayed and
+    // manual-fallback commands stay clean so users see the real `npx skills`.
+    const npxArgs = withNpmForce(baseArgs);
 
-    p.log.info(`Running: ${pc.dim("npx")} ${pc.cyan(npxArgs.join(" "))}`);
+    p.log.info(`Running: ${pc.dim("npx")} ${pc.cyan(baseArgs.join(" "))}`);
 
     const spinner = p.spinner();
     if (skillsMethod === "default") {
--- a/packages/argent-installer/src/skills.ts
+++ b/packages/argent-installer/src/skills.ts
@@ -7,6 +7,7 @@ import {
   getProjectSkillLockPath,
   listArgentSkillsInLock,
   listBundledSkills,
+  withNpmForce,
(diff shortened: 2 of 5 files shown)


latekvo (approved):
Discussed in person, thanks for the PR.

PR #347, opened 2026-06-16
feat: `GET /artifacts/` endpoint

sjchmiela:
EAS is going to poll this endpoint for new artifacts and upload as session artifacts for later user inspection.

I'm open to better ways to integrate, this seems the easiest to merge and move forward.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3914,6 +3914,7 @@
       "name": "@argent/tool-server",
       "version": "0.11.0",
       "dependencies": {
+        "@argent/configuration-core": "file:../configuration-core",
         "@argent/native-devtools-android": "file:../native-devtools-android",
         "@argent/native-devtools-ios": "file:../native-devtools-ios",
         "@argent/registry": "file:../registry",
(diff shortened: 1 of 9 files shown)


filip131311 (commented):
I have a few minor issues with it but it looks okay: 

1) it will show "dead" paths resulting in 410 response once you try to get the resources, i believe you might want to check if the resource is there when compiling the list.
2) I am not sure I like exposing a full artifact list by default and while I can not pin point the exact attack vector that could you this I would prefer you used our `argent flags` system (agent can tell you all about it) to hid it, on you server enabling this would be as trivial as running `argent enable <flag_name>` before starting the server, or even embedding the config file into the server image itself.

sjchmiela:
2. Added feature flag! For tests mocked `configuration-core` which seemed simpler than overriding the whole HOME and whatnots, happy to follow-up if this should follow convention used in other places.

---

1. I think one can make an argument that _technically_ the backing file can get removed after consumer fetches the list, but before it downloads it, so consumer needs to handle 410s either way, so we can maybe just return all artifacts and let the consumer download what it can? I, for one, am happy to just handle 410s. What do you think?

filip131311 (approved):
THank you for this contribution hopefully it will land in tomorrows release

PR #350, opened 2026-06-17
feat(ios-profiling): add @argent/ios-profiling — xctrace-free iOS Simulator profiler

filip131311:
## What

Adds a new **standalone** package `@argent/ios-profiling` (`packages/argent-ios-profiling`). It is **not wired into the tool-server yet** — nothing imports it; adopting it is a deliberate follow-up.

It replaces the broken xctrace capture+export on **Xcode 26.4 / 26.5 / 26.6 / 27.0**, where `xctrace record --device <sim>` hangs on finalize (the simulator `coreprofilesessiontap` finalize regression — verified: 26.3 build `17C529` finalizes fine, 26.4+ hangs). A native binary drives the Instruments server directly over **DTX** and a **pure-TypeScript** parser emits the same `xctrace export`-format XML that `runIosProfilerPipeline` already consumes, so the whole downstream pipeline is unchanged.

## Coverage (all 5 of Argent's template vectors)

| Vector | Source |
|---|---|
| Time Profiler (CPU) | `coreprofilesessiontap` PET time trigger @ 1 kHz (DTX) |
| Hangs | derived (main-thread on-CPU runs ≥250 ms) |
| Leaks | in-sim `simctl spawn leaks` |
| Memory | `sysmontap` (DTX) |
| Allocations | in-sim `simctl spawn heap` |

- **No Python / no third-party deps.** The kdebug `RAW_VERSION2` parser is a pure-TS port of the `pykdebugparser` callstack path, validated **byte-for-byte** against the Python reference (identical 36,028-callstack output) — zero runtime dependency on `pymobiledevice3`.
- Native binaries build locally via `scripts/build.sh` (clang → `bin/darwin/`), `bin/` gitignored like `native-devtools-ios/bin`; `PREBUILT_IOS_PROFILER_BIN_DIR` supports non-macOS CI.

## Validation vs xctrace-26.3 (last-good)

- **Leaks / Allocations / Memory:** byte-for-byte identical on an idle app (11 leaks / 1888 B; 63,639 heap nodes / 8,197,136 B; footprint 54.3 MB == OS `footprint`/`vmmap`).
- **CPU:** the `tk:1` PET time trigger + `si` (ns) sample interval + PERF-class-only `kdf2` filter give **931 Hz vs xctrace 1037 Hz** on a deterministic 100%-CPU workload (both 100% on the same hot symbol); idle apps correctly yield ~0 samples.
- **Real app (ParadiseGallery scrolling):** identical top hotspot (Hermes interpreter ~8%) and supporting cast. Known caveat (documented in README): on bursty multi-threaded workloads absolute sample density is ~5× sparser than xctrace though the relative ranking matches — a separate follow-up.

## Checks
- `prettier --write` (repo formatter) — clean
- `tsc --noEmit` — clean
- (repo has no ESLint config, so none to run)

## Not included
- Wiring into the tool-server `native-profiler-start/stop` path (intentional follow-up).

### DIFF
--- a/packages/argent-ios-profiling/objc_src/capture.m
+++ b/packages/argent-ios-profiling/objc_src/capture.m
@@ -5,47 +5,65 @@
 // report on stderr and exit non-zero so the TS layer never trusts a partial file.
 #import "conn.h"
 
-int main(int argc, char **argv){
-  if(argc<4){ fprintf(stderr,"usage: capture <udid> <seconds> <out.bin>\n"); return 1; }
+int main(int argc, char **argv) {
+  if (argc < 4) {
+    fprintf(stderr, "usage: capture <udid> <seconds> <out.bin>\n");
+    return 1;
+  }
   double secs = atof(argv[2]);
-  @autoreleasepool{
+  @autoreleasepool {
     loadInstrumentsFrameworks();
-    int spid=-1; id conn=connectLocalHub(&spid);
-    if(!conn){ fprintf(stderr,"capture: could not connect to DTServiceHub\n"); return 2; }
-    ((void(*)(id,SEL))objc_msgSend)(conn, sel_registerName("resume"));
-    ((void(*)(id,SEL,id))objc_msgSend)(conn, sel_registerName("_notifyOfPublishedCapabilities:"),
-        (@{@"com.apple.private.DTXBlockCompression": @2, @"com.apple.private.DTXConnection": @1}));
+    int spid = -1;
+    id conn = connectLocalHub(&spid);
+    if (!conn) {
+      fprintf(stderr, "capture: could not connect to DTServiceHub\n");
+      return 2;
+    }
+    ((void (*)(id, SEL))objc_msgSend)(conn, sel_registerName("resume"));
+    ((void (*)(id, SEL, id))objc_msgSend)(conn, sel_registerName("_notifyOfPublishedCapabilities:"),
+                                          (@{
+                                            @"com.apple.private.DTXBlockCompression" : @2,
+                                            @"com.apple.private.DTXConnection" : @1
+                                          }));
 
-    id ch = ((id(*)(id,SEL,id))objc_msgSend)(conn, sel_registerName("makeChannelWithIdentifier:"),
+    id ch = ((id (*)(id, SEL, id))objc_msgSend)(
+        conn, sel_registerName("makeChannelWithIdentifier:"),
         @"com.apple.instruments.server.services.coreprofilesessiontap");
-    if(!ch){ fprintf(stderr,"capture: could not open coreprofilesessiontap channel\n"); return 2; }
+    if (!ch) {
+      fprintf(stderr, "capture: could not open coreprofilesessiontap channel\n");
+      return 2;
+    }
 
-    NSString *outPath=[NSString stringWithUTF8String:argv[3]];
+    NSString *outPath = [NSString stringWithUTF8String:argv[3]];
     [[NSFileManager defaultManager] createFileAtPath:outPath contents:nil attributes:nil];
-    NSFileHandle *fh=[NSFileHandle fileHandleForWritingAtPath:outPath];
-    if(!fh){ fprintf(stderr,"capture: cannot open output file %s\n", argv[3]); return 2; }
+    NSFileHandle *fh = [NSFileHandle fileHandleForWritingAtPath:outPath];
+    if (!fh) {
+      fprintf(stderr, "capture: cannot open output file %s\n", argv[3]);
+      return 2;
+    }
 
-    __block long long totalBytes=0;
-    __block BOOL writeFailed=NO;
-    void (^h)(id) = ^(id msg){
-      id data=((id(*)(id,SEL))objc_msgSend)(msg, sel_registerName("data"));
-      if([data isKindOfClass:[NSData class]] && [data length]>0){
-        totalBytes+=[data length];
-        uint32_t len=(uint32_t)[data length];
-        @try{
+    __block long long totalBytes = 0;
+    __block BOOL writeFailed = NO;
+    void (^h)(id) = ^(id msg) {
+      id data = ((id (*)(id, SEL))objc_msgSend)(msg, sel_registerName("data"));
+      if ([data isKindOfClass:[NSData class]] && [data length] > 0) {
+        totalBytes += [data length];
+        uint32_t len = (uint32_t)[data length];
+        @try {
           [fh writeData:[NSData dataWithBytes:&len length:4]];
           [fh writeData:data];
-        }@catch(id e){
+        } @catch (id e) {
           // disk full / IO error — the stream would silently truncate, so flag it
           // and exit non-zero rather than leave a valid-looking partial file.
-          if(!writeFailed) fprintf(stderr,"capture: write failed: %s\n", [[e description]UTF8String]);
-          writeFailed=YES;
+          if (!writeFailed)
+            fprintf(stderr, "capture: write failed: %s\n", [[e description] UTF8String]);
+          writeFailed = YES;
         }
       }
     };
-    ((void(*)(id,SEL,id))objc_msgSend)(ch, sel_registerName("setMessageHandler:"), h);
+    ((void (*)(id, SEL, id))objc_msgSend)(ch, sel_registerName("setMessageHandler:"), h);
 
-    NSString *uuid=[[NSUUID UUID] UUIDString];
+    NSString *uuid = [[NSUUID UUID] UUIDString];
     // Periodic "Profile Every Thread" (PET) time trigger:
     //   tk=1  -> DTKPTriggerTime (periodic timer), NOT the kdebug-event trigger (tk=3,
     //            which only samples on syscalls/context-switches and so massively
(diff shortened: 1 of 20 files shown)


packages/argent-ios-profiling/src/leaks.ts
```diff
@@ -0,0 +1,113 @@
...
+  const ROOT = /^\s*(\d+)\s+\(([^)]+)\)\s+ROOT LEAK:\s+<([A-Za-z_][\w]*)[^>]*>\s+\[(\d+)\]/;
+  for (const line of txt.split("\n")) {
+    const m = ROOT.exec(line);
+    if (!m) continue;
+    const count = parseInt(m[1], 10);
+    const total = toBytes(m[2]);
+    const typ = m[3];
+    const objsz = parseInt(m[4], 10);
+    rows.push(
+      `<row leaked-object="${xmlEscape(typ)}" size="${total || objsz}" count="${count}" ` +
```
latekvo: **leaked bytes over-reported ×count.** `size` here is the group TOTAL bytes, but downstream (`pipeline/01-correlate.ts:121`) computes `totalSize = sizeBytes * count`, which expects size to be PER-OBJECT. Every multi-instance leak is inflated by ×count. Hidden by the idle-app validation (idle leaks are mostly count=1).

packages/argent-ios-profiling/src/leaks.ts
```diff
@@ -0,0 +1,113 @@
...
+    txt = stdout + stderr;
+  } catch (e: unknown) {
+    // `leaks` exits non-zero when it finds leaks; capture its stdout anyway
+    const err = e as { stdout?: string; stderr?: string };
+    txt = (err.stdout ?? "") + (err.stderr ?? "");
+  }
+
+  const rows: string[] = [];
+  // e.g. "  9 (1.25K) ROOT LEAK: <NSMutableDictionary 0x..> [32]  item count: 3"
+  const ROOT = /^\s*(\d+)\s+\(([^)]+)\)\s+ROOT LEAK:\s+<([A-Za-z_][\w]*)[^>]*>\s+\[(\d+)\]/;
```
latekvo: **leak type accuracy.** `[A-Za-z_][\w]*` stops at the first `.`, so Swift module-qualified names (`MyMod.MyClass`) are truncated to the module. Untyped raw-malloc-block leaks (no `<Type>`) are also silently dropped.

packages/argent-ios-profiling/src/leaks.ts
```diff
@@ -0,0 +1,113 @@
...
+      {
+        maxBuffer: 48 * 1024 * 1024,
+        timeout: 90_000,
+      }
+    );
+    txt = stdout;
+  } catch (e: unknown) {
+    txt = (e as { stdout?: string }).stdout ?? "";
+  }
+  const summary = txt.slice(txt.indexOf("All zones:"));
```
latekvo: **slice(-1) bug.** `indexOf("All zones:")` returns -1 when absent, and `slice(-1)` then returns the LAST char, not an empty string.

packages/argent-ios-profiling/src/parser/kdebug.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  // --- kd_header_v2 ---
+  // numThreads u32@4, pad(8+4), is_64bit u32@16, tick_frequency u64@20,
+  // pad(0x100), threadmap[numThreads] (each 32B: tid u64, pid u32, name char[20]),
+  // then a run of zero padding until the first record.
+  const numThreads = raw.readUInt32LE(4);
+  // version(4) + numThreads(4) + pad(8) + pad(4) + is_64bit(4) + tick_frequency(8) = 32
+  let off = 32;
+  off += 0x100; // pad(256) → threadmap at 288
+  for (let i = 0; i < numThreads; i++) {
```
latekvo, line 89: **unbounded threadmap loop crashes on truncated capture.** `numThreads` comes straight from the (untrusted) stream and bounds this loop with no `off + 32 <= raw.length` check — unlike the kd_buf loop just below, which does guard. A capture killed mid-write throws an opaque RangeError that rejects the whole profile (reproduced locally).

packages/argent-ios-profiling/src/parser/kdebug.ts
```diff
@@ -0,0 +1,161 @@
...
+    const end = off + 12;
+    let z = end;
+    while (z < off + 32 && raw[z] !== 0) z++;
+    const name = raw.toString("utf8", end, z);
+    threadsPids.set(tid, pid);
+    if (!pidsNames.has(pid)) pidsNames.set(pid, name);
+    off += 32;
+  }
+  // skip zero padding to the first record (kd_header_v2 `_pad` = GreedyRange of 0)
+  while (off < raw.length && raw[off] === 0) off++;
```
latekvo: **zero-padding skip silently drops the entire record stream ~1/256.** Assumes records never begin with 0x00, but the first record's field is a u64 timestamp. When its low byte is 0 (~1/256, uniform), this eats the byte, misaligns the 64-byte grid, and loses ALL samples — no error, threadmap still parses, so you get a valid-looking empty profile. Reproduced: identical streams differing only in the first timestamp's LSB → 1 callstack vs 0.

packages/argent-ios-profiling/src/parser/kdebug.ts
```diff
@@ -0,0 +1,161 @@
...
+    const debugid = raw.readUInt32LE(off + 48);
+    const eventid = debugid & 0xfffffffc;
+    const func = debugid & 0x3;
+    const tid = Number(raw.readBigUInt64LE(off + 40));
+
+    if (eventid === PERF_EVENT) {
+      if (func === 1) {
+        // START — open a sample group for this thread
+        open.set(tid, {
+          ts: Number(raw.readBigUInt64LE(off)),
```
latekvo: **u64→number narrowing.** Timestamp narrowed to a JS number; lossy above 2^53 (x86 sims report ns → ~104 days uptime). Addresses correctly stay bigint.

packages/argent-ios-profiling/src/parser/emit.ts
```diff
@@ -0,0 +1,207 @@
...
+        `<backtrace id="${nid()}">${bt.join("")}</backtrace></row>`
+    );
+  }
+  const cpuXml =
+    `<?xml version="1.0"?>\n<trace-toc><run number="1"><data>` +
+    `<table schema="time-profile">${rows.join("")}</table></data></run></trace-toc>\n`;
+  fs.writeFileSync(`${outPrefix}_raw_cpu.xml`, cpuXml);
+
+  // Hangs: contiguous on-CPU runs of the busiest (≈main) thread ≥250ms.
+  const HANG = 250_000_000;
```
latekvo, line 169: **hang threshold may be in the wrong units (verify on 26.x).** `HANG = 250_000_000` assumes nanoseconds, but kd_buf timestamps are mach-absolute TICKS. On Apple Silicon (~41.7 ns/tick) that's ~10s, not 250ms → hangs badly under-detected. CPU ranking is unaffected (relative gaps), but absolute hang durations aren't. There's no test covering it.

packages/argent-ios-profiling/src/index.ts
```diff
@@ -0,0 +1,137 @@
...
+    path.dirname(opts.outPrefix),
+    `.iosprof-${process.pid}-${Date.now()}.kdbg`
+  );
+  try {
+    // 1) native capture: drive coreprofilesessiontap → length-framed kdebug stream
+    await execFileAsync(
+      bin("ios-profiler-capture"),
+      [opts.udid, String(opts.durationSec), rawFile],
+      {
+        maxBuffer: 8 * 1024 * 1024,
```
latekvo, line 84: **no timeout on the native child (the hang you're routing around).** No `timeout` on the capture exec (and `captureMemory` at line 124), unlike leaks/heap which set 90s. If the binary stalls on DTX connect or stop/finalize, `captureProfile` hangs forever with no kill.

latekvo (commented):
hmm yes very reviewable

packages/argent-ios-profiling/objc_src/capture.m
```diff
@@ -0,0 +1,102 @@
...
+    if(!conn) return 2;
+    ((void(*)(id,SEL))objc_msgSend)(conn, sel_registerName("resume"));
+    ((void(*)(id,SEL,id))objc_msgSend)(conn, sel_registerName("_notifyOfPublishedCapabilities:"),
+        (@{@"com.apple.private.DTXBlockCompression": @2, @"com.apple.private.DTXConnection": @1}));
+
+    id ch = ((id(*)(id,SEL,id))objc_msgSend)(conn, sel_registerName("makeChannelWithIdentifier:"),
+        @"com.apple.instruments.server.services.coreprofilesessiontap");
+    printf("coreprofile channel=%p\n", ch);
+
+    printf("a1: channel ok\n");
```
latekvo: **debug scaffolding shipped.** Spike-grade printf tracing throughout (a1/a2/a3, [msg #N], config dumps, RESULT at line 97). Harmless (the stream goes to the file, not stdout).

packages/argent-ios-profiling/objc_src/capture.m
```diff
@@ -0,0 +1,102 @@
...
+      id obj =((id(*)(id,SEL))objc_msgSend)(msg, sel_registerName("object"));
+      unsigned long mt=((unsigned long(*)(id,SEL))objc_msgSend)(msg, sel_registerName("messageType"));
+      if(anyMsg<=5) printf("[msg #%d] class=%s type=%lu dataLen=%lu obj=%s\n", anyMsg,
+          object_getClassName(msg), mt,
+          [data isKindOfClass:[NSData class]]?(unsigned long)[data length]:0,
+          obj?object_getClassName(obj):"nil");
+      if([data isKindOfClass:[NSData class]] && [data length]>0){
+        dataMsgs++; totalBytes+=[data length];
+        uint32_t len=(uint32_t)[data length];
+        @try{ [fh writeData:[NSData dataWithBytes:&len length:4]]; [fh writeData:data]; }@catch(id e){}
```
latekvo: **write errors swallowed.** `@catch(id e){}` silently drops write failures (disk full / IO) → the stream truncates yet the binary can still exit 0.

packages/argent-ios-profiling/objc_src/simmem.m
```diff
@@ -0,0 +1,49 @@
...
+      id top=nil; @try{ top=[NSKeyedUnarchiver unarchiveObjectWithData:d]; }@catch(id e){}
+      if(!top) return;      NSArray *arr = [top isKindOfClass:[NSArray class]] ? top : @[top];
+      for(id e in arr){
+        if(![e isKindOfClass:[NSDictionary class]]) continue;
+        NSDictionary *procs=e[@"Processes"];
+        if(![procs isKindOfClass:[NSDictionary class]]) continue;
+        id row=procs[@(tpid)] ?: procs[[@(tpid) stringValue]];
+        if([row isKindOfClass:[NSArray class]] && [row count]>=2){
+          // procAttrs=[pid,physFootprint,memResidentSize]; row may or may not include pid
+          NSUInteger base=([row count]>=3)?1:0;
```
latekvo: **footprint/resident column heuristic (verify on 26.x).** `base = ([row count] >= 3) ? 1 : 0` guesses whether the row includes the pid column; if a 26.x sysmontap row carries 3 cols without a leading pid, footprint/resident get mislabeled. Couldn't verify on Xcode 16.4.

packages/argent-ios-profiling/objc_src/simmem.m
```diff
@@ -0,0 +1,49 @@
...
+        @"com.apple.instruments.server.services.sysmontap");
+    // procAttrs order MUST match how we read the row below
+    NSArray *procAttrs=@[@"pid",@"physFootprint",@"memResidentSize",@"cpuUsage"];
+    __block int dumped=0;
+    ((void(*)(id,SEL,id))objc_msgSend)(ch, sel_registerName("setMessageHandler:"), ^(id o){
+      // sysmontap update payload = OS_dispatch_data wrapping a binary plist.
+      id pl=((id(*)(id,SEL))objc_msgSend)(o,sel_registerName("object"));
+      NSData *d = [pl isKindOfClass:[NSData class]] ? (NSData*)pl : nil;
+      if(!d) return;
+      id top=nil; @try{ top=[NSKeyedUnarchiver unarchiveObjectWithData:d]; }@catch(id e){}
```
latekvo: **deprecated API.** `unarchiveObjectWithData:` is deprecated since macOS 10.14 (the only compiler warning in the build).

packages/argent-ios-profiling/objc_src/conn.h
```diff
@@ -0,0 +1,47 @@
...
+  NSArray *devs=((id(*)(id,SEL))objc_msgSend)(ds,sel_registerName("availableDevices"));
+  NSString *want=[NSString stringWithUTF8String:udid];
+  for(id d in devs){id u=((id(*)(id,SEL))objc_msgSend)(d,sel_registerName("UDID"));
+    if([((id(*)(id,SEL))objc_msgSend)(u,sel_registerName("UUIDString")) caseInsensitiveCompare:want]==NSOrderedSame) return d;}
+  return nil;
+}
+
+// Returns a resumed DTXConnection to a freshly spawned host DTServiceHub (or nil).
+static inline id connectLocalHub(int *serverPidOut){
+  AuthorizationRef auth=NULL; AuthorizationCreate(NULL,kAuthorizationEmptyEnvironment,kAuthorizationFlagDefaults,&auth);
```
latekvo: **AuthorizationRef leak.** The AuthorizationRef created here is never freed (MRR). One-shot, so minor.

packages/argent-ios-profiling/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@argent/ios-profiling",
+  "version": "0.11.0",
+  "private": true,
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "scripts": {
```
latekvo, line 7: **no tests for the parity claims; CI can't see this package.** No `test` script, and the typecheck script is named `typecheck`, not `typecheck:tests`, so CI's `--if-present` hooks skip it. The byte-for-byte / 36,028-callstack claims have no committed fixture.

packages/argent-ios-profiling/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "composite": true,
```
latekvo, line 4: **composite package orphaned from the build graph.** `composite: true`, but the package isn't in the root tsconfig `references`, so `tsc --build` (used by every CI workflow) never compiles or typechecks it — a type error here would ship.

PR #351, opened 2026-06-17
feat(ios-profiler): attributable memory leaks via malloc_stack_logging

latekvo:
## What

Opt-in `malloc_stack_logging` flag on `native-profiler-start` that makes iOS memory leaks **attributable** — they come back with a real responsible frame + library instead of `<Call stack limit reached>`.

## Why

`native-profiler-start` attaches to the already-running app. iOS records malloc allocation backtraces only when `MallocStackLogging` is set **at process launch**, so an attached app has none: Instruments' Leaks scanner finds the leaked blocks but can't attribute them and emits the placeholder `<Call stack limit reached>` (that string is Instruments', not ours).

## How

When `malloc_stack_logging: true`, the profiler cold-launches the app **under** xctrace instead of attaching:

```
xctrace record --template <Argent.tracetemplate> --device <udid> \
  --env MallocStackLogging=1 --output <…> --no-prompt \
  --launch -- <App.app>
```

- `.app` path resolved via `simctl get_app_container`; the running instance is terminated first for a clean cold start.
- The existing Leaks export / parser / render path is reused **unchanged** — no parser changes.
- Default behaviour is untouched: attach to the running app (no relaunch, no overhead). The flag is opt-in because it restarts the app (loses state) and adds allocator overhead — for leak-attribution passes, not CPU/hang work.
- The report relabels unattributable leaks with a hint to re-run with `malloc_stack_logging: true`, instead of surfacing the raw placeholder.

## Evidence (real iPhone 16 sim, same scroll workload)

Before (attach) — every leak row:
```
responsible-frame="<Call stack limit reached>" responsible-library=""
```
After (`--env MallocStackLogging=1 --launch`) — same Leaks export:

| leaked-object | responsible-frame | library |
|---|---|---|
| Malloc 1008 B ×3 | `itanium_demangle::OutputBuffer::grow(...)` | libc++abi |
| Malloc 48 B ×1 | `hermes::vm::JSTypedArrayBase::createBuffer(...)` | hermes |

## Tests

- `test/ios-instruments/malloc-stack-logging.test.ts` — the malloc launch argv (`--env MallocStackLogging=1`, `--launch -- <app>` last, with and without `--notify-tracing-started`, `--no-prompt`) vs default `--attach`; terminate-first / get_app_container ordering; failed-start behavior; per-test `ARGENT_IOS_CAPTURE` scrubbing.
- `test/ios-instruments/leak-attribution-render.test.ts` — unattributable vs attributed leak rendering, capture-mode-aware hints.
- `test/ios-instruments/load-freshness.test.ts` — per-capture residue clearing on `profiler-load`, live/recovery-pending load refusals.
- `test/ios-instruments/stop-recovery.test.ts` — capture-mode pairing at stop (clean and recovery paths).
- `test/profiler-combined-leaks.test.ts`, `test/profiler-leak-stacks.test.ts`, `test/ios-leak-pipeline.test.ts` — shared caveat line, evidence-first hints, attribution-normalized grouping.
- Full `test/ios-instruments/` suite green (9 files / 70 tests); full tool-server suite green; `tsc --noEmit` + `tsc -p tsconfig.test.json` clean.

## Notes

E2E'd live twice during review (real iPhone 16 Pro / iPhone 17 Pro sims on Xcode 26.6): guard refusals fire before the app is touched, the `--device` cold launch's deadlock on degraded Xcode is confirmed (so the guard is warranted), and the malloc argv/terminate/relaunch behavior checked out on-device. Docs updated: `IOS_PROFILER_REFERENCE.md` and the `argent-native-profiler` skill.

## Review follow-ups (post-approval hardening)

Later commits address the remaining review threads and several subsequent multi-lens review passes:

- **Capture-mode is a real signal, end to end.** The session records the mode of the in-flight recording at start (`recordingMallocStackLogging`), pairs it with `exportedFiles` at stop (`mallocStackLogging`), and freezes it into `parsedData` at analyze/load — so the analyze report, the combined report, and the `leak_stacks` drill-down all name the capture mode of the data they actually render instead of inferring it (or mislabeling it after a newer capture starts). One shared `renderUnattributedLeaksNote()` keeps the wording identical across reports, and attribution evidence outranks the flag: a resolved frame proves the target ran under malloc stack logging however it was launched.
- **`profiler-load` restores a clean slate.** Loading clears all per-capture residue (capture mode, `cpuFilterPid`, `wallClockStartMs`, `traceFile`) so a loaded trace is never analyzed with a previous live capture's PID filter, freshness anchor, or report path — and refuses to load while a recording is in flight or a crashed capture awaits `native-profiler-stop`'s partial-trace recovery, which the load would otherwise orphan.
- **Failed starts are non-destructive.** Per-capture descriptors — and the recovery flags that gate `native-profiler-stop`'s partial-trace export — are stamped/reset only on a successful start, so a failed attempt leaves the previous capture's still-loaded exports fully described and a pending abnormal-end recovery intact.
- **Guards keep session state coherent.** `profiler-load`, `native-profiler-analyze`, and `profiler-combined-report` all refuse while a recording is in flight or a crashed capture awaits recovery, via one shared `isCaptureInFlight`/`inFlightGuardMessage` helper whose message distinguishes recording vs 10-min-cap timeout vs unexpected exit (so the stated cause matches what `native-profiler-stop` reports next). Proceeding would otherwise wedge the session (load) or render one capture's frozen data under another's live fields (analyze/combined).
- **The combined report anchors on a frozen clock.** The recording's start time is frozen into `parsedData` at analyze, and the iOS combined report anchors its hangs to that frozen value rather than the live session field — so a capture started after analyze can never shift a real hang↔commit correlation into "unmatched". Android keeps its live anchor (it re-derives hangs from the live trace). The combined-report guard's recovery advice routes through `native-profiler-analyze`, the only step that rebuilds `parsedData`.
- **Leak grouping is attribution-normalized.** `aggregateLeaks` keys on `(objectType, frame-normalized-the-way-isLeakAttributed-matches)`, so mixed unattributed spellings (`Unknown`, the sentinel, whitespace variants) can't split one group per type.
- **The malloc launch target is unambiguous.** Since the cold launch terminates the resolved app, `app_process` resolution prefers an exact `CFBundleExecutable` match and, when builds still tie (dev + prod sharing an executable/display name), refuses and points at the globally-unique `CFBundleIdentifier` — which it also accepts as an `app_process` — instead of killing whichever app the plist enumerates first.
- **Report tables escape `|`.** Demangled C++ frames (`operator|` included) are real headline content under malloc mode; every free-form cell across the leak, CPU-hotspot, function caller/callee, and thread-breakdown tables — symbol *and* thread/queue-name columns — is escaped so a stray pipe can't shift the row's columns.
- **Android start is non-destructive too.** `startNativeProfilerAndroid` now starts perfetto before mutating session state, mirroring iOS, so a failed start no longer burns a prior capped capture's pending partial-trace recovery.


packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
```diff
@@ -340,20 +433,46 @@ export async function startNativeProfilerIos(
...
+        "--output",
+        outputFile,
+        "--no-prompt",
+        "--env",
+        "MallocStackLogging=1",
+      ];
+      if (notify) {
+        xctraceArgs.push("--notify-tracing-started", notifyName);
+      }
+      xctraceArgs.push("--launch", "--", launchBundlePath);
```
hubgan: `platforms/ios.ts:444–464` builds the cold-launch argv with a **hardcoded `--device`**, completely bypassing `selectIosCaptureStrategy()`. But `capture-strategy/select.ts:50-53` classifies **Xcode 26.4–27.0 as degraded** ("the `--device` recording-start handshake deadlocks"), and the default path uses the `all-processes` fallback for exactly this reason.

**Reproduced E2E on Xcode 26.5, same host, same workload — twice:**

| Path | xctrace args | Trace bundle | Export | `analyze` result |
|---|---|---|---|---|
| Default (this PR untouched) | `--all-processes` (strategy-selected) | **510 MB** | ok | normal report |
| `malloc_stack_logging: true` (run 1) | `--device … --launch -- <App.app>` | **52 KB** | `Document Missing Template Error` | *"Analysis failed — no findings"* |
| `malloc_stack_logging: true` (run 2, clean) | same | **56 KB** | same | same |

The tool-server log printed `Xcode 26.5 has the xctrace --device recording-start deadlock; using the "all-processes" capture fallback` for the default path — then the malloc path used `--device` anyway. The 510 MB vs ~50 KB contrast **isolates the cause to `--device`**, ruling out an environmental/export issue (Android produced a healthy 25 MB Perfetto trace in the same session).

Net effect for a user on current Xcode: the app is **terminated (state lost)**, they wait through a slow instrumented launch, and they get **zero leaks instead of attributed ones** — the opposite of the feature's purpose, surfaced only as a generic "Analysis failed."

There is also a real **design gap**, not just a missing branch: the `all-processes` strategy is a host-wide attach-by-PID and *cannot* `--launch`, while `malloc_stack_logging` *requires* a cold launch. So "just route through the strategy" does not work. At minimum the feature must **detect the degraded Xcode range and fail fast with a clear message before terminating the app**, instead of silently producing an empty trace. (The `ARGENT_IOS_CAPTURE` escape hatch is also ignored by this path, so there is no override either.)

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
```diff
@@ -340,20 +433,46 @@ export async function startNativeProfilerIos(
...
+        params.device_id,
+        "--output",
+        outputFile,
+        "--no-prompt",
+        "--env",
+        "MallocStackLogging=1",
+      ];
+      if (notify) {
+        xctraceArgs.push("--notify-tracing-started", notifyName);
+      }
```
hubgan, line 618: The start gating resolves on the `--notify-tracing-started` signal, which fires even though the `--device` session records nothing. So `native-profiler-start` returns a confident `{status:"recording", pid}`, `native-profiler-stop` returns an artifact, and only `analyze` reveals the trace was empty — after the app was already terminated. The default attach path never terminates the app; this path has a strictly worse failure-recovery story (a failed start leaves the app dead with no restore).

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
```diff
@@ -300,30 +361,62 @@ export async function startNativeProfilerIos(
...
-      }
-    );
+
+  // Default flow attaches to the running app (preserves state, no overhead).
+  // malloc_stack_logging mode instead cold-launches the app *under* xctrace with
+  // MallocStackLogging=1 so the malloc library records allocation backtraces from
+  // the first allocation — without that, leaks are detected but unattributable
+  // ("<Call stack limit reached>"). `--env` is only honoured with `--launch`,
+  // which needs the .app path rather than the executable name or PID.
+  const useMallocStackLogging = params.malloc_stack_logging === true;
```
hubgan, line 502: Disclosed as a trade-off, but worth quantifying: under `MallocStackLogging=1` the Bluesky cold launch **never got past the splash screen in ~80 s** (every Hermes/RN startup allocation records a backtrace). Combined with #2, a user could reasonably conclude the tool hung. Consider setting expectations in the report/docs ("the app will be very slow to start under malloc stack logging").

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
```diff
@@ -224,6 +205,85 @@ function resolveExplicitApp(udid: string, name: string): DetectedApp {
...
+        info.ApplicationType === "User" &&
+        (info.CFBundleExecutable === appProcess || info.CFBundleDisplayName === appProcess)
+      ) {
+        return info;
+      }
+    }
+    throw new Error(
+      `No installed user app matching "${appProcess}" found on simulator ${udid}. ` +
+        `Pass the exact CFBundleExecutable or display name, or omit app_process to auto-detect the running app.`
+    );
```
hubgan: `getAppBundlePath` (`ios.ts:234-248`) and `resolveAppForLaunch` (`ios.ts:256-285`) throw plain `Error`, unlike every other failure in this file which uses `FailureError` with `error_code` / `failure_stage` / `failure_area`. These are reachable user-facing failures (no bundle path, no matching app, multiple running apps) that will be invisible to failure-code telemetry. Wrap them in `FailureError` for consistency.

packages/tool-server/test/ios-instruments/malloc-stack-logging.test.ts
```diff
@@ -0,0 +1,155 @@
...
+    vi.doUnmock("../../src/utils/ios-profiler/notify");
+    vi.doUnmock("../../src/utils/ios-profiler/startup");
+  });
+
+  function applyCommonMocks(spawnFn: unknown, execSyncFn: unknown) {
+    vi.doMock("child_process", () => ({
+      spawn: spawnFn,
+      execSync: execSyncFn,
+      execFile: vi.fn(),
+    }));
```
hubgan, line 122: `test/ios-instruments/malloc-stack-logging.test.ts` mocks `spawn`/`execSync`, so it verifies the assembled `--launch --env MallocStackLogging=1` argv and the terminate-first behavior — but **never runs real xctrace**, which is why a green suite (9 files / 44 tests, confirmed locally) coexists with a non-functional feature on current Xcode. Not a defect in itself, but the PR's "argv verified independently" confidence overweights what these tests can prove. The evidence table in the PR was likely captured on a non-degraded Xcode (≤26.3); please state the Xcode version used.

hubgan (changes requested):

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
latekvo: Fixed in 9e123359.

The malloc path now calls `selectIosCaptureStrategy()` at the top of the branch and, when it does not resolve to the `device` strategy (i.e. degraded Xcode 26.4–27.0 with no override), throws `FailureError(NATIVE_PROFILER_MALLOC_DEGRADED_XCODE)` **before** terminating the running app — so the app is never killed and you get an actionable message instead of an empty trace. `ARGENT_IOS_CAPTURE=device` is honored as the escape hatch you flagged. The `--all-processes` fallback genuinely can't `--launch`, so the guard refuses fast rather than trying to substitute it.

Added a deterministic test that mocks `xcodebuild -version` → 26.5 and asserts the refusal fires before any `simctl terminate`/`get_app_container`/spawn, plus a second test that `ARGENT_IOS_CAPTURE=device` forces the cold launch through.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
latekvo: Fixed in 9e123359.

Two parts: (1) the degraded-Xcode case now fails fast *before* the terminate (sibling thread), so the common "app killed, nothing recorded" path is gone; (2) for any other malloc start failure after the terminate, the app is now best-effort relaunched (`xcrun simctl launch`, via `execFileSync` with discrete argv) before the original error propagates — so a failed start no longer leaves the user's app dead with no restore.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
latekvo: Fixed in 9e123359.

Expectation is now set in both the `malloc_stack_logging` param description and `IOS_PROFILER_REFERENCE.md`: the app is noticeably slow to launch under malloc stack logging because every startup allocation records a backtrace.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
latekvo: Fixed in 9e123359.

`getAppBundlePath` and both `resolveAppForLaunch` throw sites now raise `FailureError` with native-profiler codes (`NATIVE_PROFILER_APP_BUNDLE_PATH_FAILED`, `NATIVE_PROFILER_LAUNCH_APP_NOT_FOUND`, and the existing `NATIVE_PROFILER_MULTIPLE_RUNNING_USER_APPS`) so these reachable user-facing failures are telemetry-classified like the rest of the file. Also added an empty-resolved-bundle-path guard.

packages/tool-server/test/ios-instruments/malloc-stack-logging.test.ts
latekvo: Fair point, and thanks for quantifying it. The evidence table in the PR was captured on a non-degraded Xcode (≤ 26.3), where `--device --launch` works and produces the attributed leaks shown.

The substantive risk you flagged — a green mock suite coexisting with a feature that silently does nothing on current Xcode — is now closed in 9e123359: on 26.4–27.0 the malloc path is rejected up front with a clear `FailureError` instead of recording an empty trace, and that refusal is itself covered by a deterministic test (mock `xcodebuild -version` → 26.5, assert the refusal happens before any app termination/spawn). The argv-shape tests stay as-is; the real-xctrace gap is inherent to not having Instruments in CI, but the degraded path is no longer a silent failure.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
```diff
@@ -300,60 +394,152 @@ export async function startNativeProfilerIos(
...
+          `terminate your app and then capture an empty trace. Re-run without malloc_stack_logging ` +
+          `(leaks are still detected, just unattributed), profile on a non-degraded Xcode, or set ` +
+          `ARGENT_IOS_CAPTURE=device to force the device path if you know it works on your host.`,
+        {
+          error_code: FAILURE_CODES.NATIVE_PROFILER_MALLOC_DEGRADED_XCODE,
+          failure_stage: "native_profiler_start_malloc_capability",
+          failure_area: "tool_server",
+          error_kind: "validation",
+        }
+      );
```
hubgan: This guard fires whenever the selected capture strategy isn't `device`, which also covers an operator setting `ARGENT_IOS_CAPTURE=all-processes` on a perfectly healthy Xcode. In that case both the message ("the active Xcode has the --device recording-start deadlock (26.4-27.0)") and the `NATIVE_PROFILER_MALLOC_DEGRADED_XCODE` code attribute the refusal to a degraded Xcode that isn't present - the actual cause is the forced strategy override. The refusal itself is correct, but the reported reason misleads anyone reading the error or filtering telemetry by that code.

hubgan (commented):
A few small notes on the malloc_stack_logging path - low-severity messaging/attribution details and a couple of branches the new tests don't cover. None are blocking; flagging for your consideration.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
```diff
@@ -300,60 +394,152 @@ export async function startNativeProfilerIos(
...
+  if (useMallocStackLogging) {
+    // malloc_stack_logging must cold-launch the app under `xctrace --device --launch`
+    // (only `--launch` honours `--env MallocStackLogging=1`). On Xcode 26.4–27.0 the
+    // `--device` recording-start handshake is broken (see capture-strategy), so this
+    // would terminate the running app and then capture an empty trace — the opposite
+    // of the feature's purpose, surfaced only as a downstream "Analysis failed". Refuse
+    // up front, BEFORE touching the running app, unless the operator forces the device
+    // path via ARGENT_IOS_CAPTURE=device. selectIosCaptureStrategy() encodes exactly
+    // this decision (env override → degraded-version detection → device default).
+    if (selectIosCaptureStrategy().name !== "device") {
```
hubgan: `selectIosCaptureStrategy()` is invoked here only for its return value, but on a degraded host it also writes "...using the 'all-processes' capture fallback. Override with ARGENT_IOS_CAPTURE=device." to stderr as a side effect, and this block then immediately throws the refusal. The resulting log sequence reads as though the fallback is about to run, directly followed by an abort - confusing when diagnosing why malloc_stack_logging was rejected.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
```diff
@@ -224,6 +205,118 @@ function resolveExplicitApp(udid: string, name: string): DetectedApp {
...
+    throw new FailureError(
+      `No installed user app matching "${appProcess}" found on simulator ${udid}. ` +
+        `Pass the exact CFBundleExecutable or display name, or omit app_process to auto-detect the running app.`,
+      {
+        error_code: FAILURE_CODES.NATIVE_PROFILER_LAUNCH_APP_NOT_FOUND,
+        failure_stage: "native_profiler_resolve_app_for_launch",
+        failure_area: "tool_server",
+        error_kind: "not_found",
+      }
+    );
```
hubgan, line 366: Neither error branch the malloc path adds is exercised by the tests - this `LAUNCH_APP_NOT_FOUND` throw (an explicit `app_process` matching no installed user app) nor the multiple-running-user-apps throw just below. The malloc tests always pass a name matching the single mocked app, so a regression that mis-wired either guard (wrong code/condition, or failing to throw) would go uncaught.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
```diff
@@ -421,7 +609,26 @@ export async function startNativeProfilerIos(
...
+    if (mallocRelaunchBundleId) {
+      try {
+        execFileSync("xcrun", ["simctl", "launch", params.device_id, mallocRelaunchBundleId], {
+          timeout: DETECT_RUNNING_APP_TIMEOUT_MS,
+          stdio: "ignore",
+        });
+      } catch {
+        // best-effort restore; surface the original start failure regardless
+      }
+    }
```
hubgan, line 716: In malloc mode `resolveAppForLaunch` accepts an explicitly-named app that is installed but not currently running (it doesn't require a live PID), and the preceding `simctl terminate` is then a no-op. If the cold-launch subsequently fails, this best-effort `simctl launch` foregrounds an app the user did not have running beforehand - the opposite direction from the "restore the app we killed" intent described just above.

packages/tool-server/test/ios-instruments/malloc-stack-logging.test.ts
```diff
@@ -0,0 +1,336 @@
...
+    // launch + env, not attach
+    expect(args).toContain("--launch");
+    expect(args).toContain("--env");
+    expect(args[args.indexOf("--env") + 1]).toBe("MallocStackLogging=1");
+    expect(args).not.toContain("--attach");
+    // the launched target is the resolved .app bundle, and must be the LAST args
+    // (everything after `--` is the launched command).
+    const dashIdx = args.indexOf("--");
+    expect(dashIdx).toBeGreaterThan(args.indexOf("--launch"));
+    expect(args[dashIdx + 1]).toBe("/Users/x/Library/.../MyApp.app");
```
hubgan, line 168: These assertions check `--launch`, `--env MallocStackLogging=1`, the absence of `--attach`, and the app path being the final argument, but never that `--device <udid>` is present or that the correct device id is threaded through. `--device` is the whole premise of the degraded-Xcode guard above, so a regression that dropped it or passed the wrong device id would still pass green here.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
latekvo: Fixed in 5aebc219.

The guard now resolves the strategy via a new side-effect-free `resolveIosCaptureStrategy()` that also returns *why* the strategy isn't `device`. When the cause is a forced `ARGENT_IOS_CAPTURE=all-processes` on an otherwise-healthy host (rather than a degraded Xcode), the refusal now names the override as the cause and carries a distinct `NATIVE_PROFILER_MALLOC_STRATEGY_OVERRIDE` code — so neither the message nor the telemetry code blames a degraded Xcode that isn't present. Covered by a test that forces `ARGENT_IOS_CAPTURE=all-processes` on Xcode 16.4 and asserts the override code plus a message containing `ARGENT_IOS_CAPTURE` and no `deadlock`/`26.4`.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
latekvo: Fixed in 5aebc219.

The malloc guard now calls the new side-effect-free `resolveIosCaptureStrategy()` instead of `selectIosCaptureStrategy()`, so it no longer writes the `using the "all-processes" capture fallback` line to stderr immediately before throwing the refusal (that fallback never runs on this path). `selectIosCaptureStrategy()` keeps the logging for the normal record flow, unchanged. A test asserts the degraded-Xcode refusal emits no `capture fallback` stderr write.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
latekvo: Fixed in 5aebc219.

Added two malloc-mode tests that exercise both branches: one passes an `app_process` matching no installed user app and asserts `NATIVE_PROFILER_LAUNCH_APP_NOT_FOUND` (plus that no terminate / get_app_container / spawn ran); the other passes no `app_process` with two running user apps and asserts `NATIVE_PROFILER_MULTIPLE_RUNNING_USER_APPS`. A mis-wired code/condition or a failure to throw would now be caught.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
latekvo: Fixed in 5aebc219.

`mallocRelaunchBundleId` is now assigned only *inside* the terminate `try`, after `simctl terminate` returns successfully — which only happens when the app was actually running. For an installed-but-not-running named app the terminate throws, so it stays null and a later cold-launch failure no longer foregrounds an app the user never had open; we only restore what we actually killed. Covered by a test where terminate throws (`found nothing to terminate`) and the cold launch then fails, asserting no `simctl launch` relaunch fires.

packages/tool-server/test/ios-instruments/malloc-stack-logging.test.ts
latekvo: Fixed in 5aebc219.

The malloc argv test now asserts `--device` is present and that the token following it equals the passed device id (`DEVICE-UDID`), so dropping `--device` or threading the wrong device id would fail the test.

packages/tool-server/test/ios-instruments/malloc-stack-logging.test.ts
latekvo: Follow-up in 5aebc219: the argv/branch coverage the mock suite can actually prove is now broader — the malloc path additionally asserts `--device <udid>` is threaded through, exercises the `LAUNCH_APP_NOT_FOUND` and `MULTIPLE_RUNNING_USER_APPS` error branches, and asserts the degraded-vs-forced-override refusal attribution. The real-xctrace gap remains inherent to not having Instruments in CI, as noted — but the substantive risk you flagged (a green mock suite coexisting with a silent no-op on current Xcode) stays closed. Resolving.

hubgan:
The feature is dormant on all current Xcode (26.4+), and there's no way to exercise it in CI. isDegraded() (capture-strategy/select.ts:84) flags 26.4+ and all 27+. The malloc path requires --device, so on every current Xcode the guard refuses by default, and forcing it produces an empty trace (E2E-confirmed). In practice the feature only delivers value on Xcode ≤ 26.3. This is inherent to Apple's xctrace regression and the guard is the right mitigation — but reviewers/users should know the feature ships inert on mainstream Xcode, and its core value (attribution) has no automated coverage against a real trace. Consider a tracking note for when Apple fixes --device so the bound can be narrowed.

hubgan:
Unattributed-leak render message is misleading in malloc mode. render.ts:417-421 hardcodes "Argent records via xctrace --attach, which has no malloc-stack history … For attributed stacks, capture with malloc stack logging enabled at launch." When a trace was captured with malloc_stack_logging:true but some leaks still come back unattributed, this tells the user to do the thing they just did, and names the wrong capture mode. Pre-existing (from #337) but newly reachable via this PR — the render layer has no signal of the capture mode. Low impact; worth a follow-up to thread capture-mode into the report.

hubgan:
Residual shell-injection surface in the relocated getInstalledApps. platforms/ios.ts still uses execSync(`xcrun simctl listapps ${udid} | plutil …`) with udid interpolated into a shell string. Pre-existing (this PR just extracted it to a function), and the PR's new helpers (getAppBundlePath, terminate, relaunch) correctly use execFileSync. device_id is semi-trusted, so risk is low — but the new code sets the better pattern and this straggler could adopt it.

packages/tool-server/src/utils/ios-profiler/pipeline/01-correlate.ts
hubgan, line 1: The new composite key looks like a space in the text diff, but the actual byte is a literal NUL (cat -v → ${leak.objectType}^@${leak.responsibleFrame}):

`const key = `${leak.objectType}\x00${leak.responsibleFrame}`;  // ^@ is NUL, not space`

Git/GitHub render the whole file as binary → git diff main...HEAD shows "Binary files … differ". The PR's single most important logic change (per-call-site grouping — the feature's whole point) is invisible to reviewers.

packages/tool-server/src/utils/ios-profiler/capture-strategy/select.ts
```diff
@@ -52,38 +86,68 @@ function isDegraded({ major, minor }: XcodeVersion): boolean {
...
+  }
+
+  const invalidOverride = override.kind === "invalid" ? override.raw : undefined;
+
+  const version = readActiveXcodeVersion();
+  if (version && isDegraded(version)) {
+    return {
+      strategy: allProcessesStrategy,
+      reason: { kind: "degraded-xcode", major: version.major, minor: version.minor },
+      invalidOverride,
```
hubgan, line 134: The malloc guard uses the side-effect-free resolveIosCaptureStrategy(), which (correctly) doesn't emit stderr — but that means the "ignoring unrecognised ARGENT_IOS_CAPTURE=…" warning that selectIosCaptureStrategy() prints on the normal path never fires here. A user who typos the override gets no feedback in malloc mode.

packages/tool-server/src/utils/ios-profiler/pipeline/01-correlate.ts
latekvo: Fixed in 88db2f6a.

The delimiter is now written as a unicode escape (the six characters backslash-u-0-0-0-0) instead of a raw NUL byte, so the file is text again and the per-call-site grouping change shows up as an ordinary diff: `git diff origin/main...HEAD -- 01-correlate.ts` now renders 15 insertions / 5 deletions rather than "Binary files ... differ". The runtime key is byte-identical (the grouping tests still pass unchanged), and a new regression test in `ios-leak-pipeline.test.ts` pins the collision-safety - two distinct `(objectType, responsibleFrame)` pairs that would share a space-joined key stay separate - so the escape can't be quietly swapped for a printable delimiter without the suite going red.

packages/tool-server/src/utils/ios-profiler/capture-strategy/select.ts
latekvo: Fixed in 88db2f6a.

Extracted `warnIfInvalidCaptureOverride(decision)` and call it from the malloc guard right after `resolveIosCaptureStrategy()`, so a typo'd `ARGENT_IOS_CAPTURE` now warns in malloc mode exactly as it does on the normal record path - and before the degraded-Xcode refusal, which otherwise advised setting the very var the user had just fumbled. Covered by a new test that sets `ARGENT_IOS_CAPTURE=devise` on a healthy Xcode and asserts the "ignoring unrecognised ARGENT_IOS_CAPTURE=..." line is emitted while the cold launch still proceeds.

packages/tool-server/src/tools/profiler/native-profiler/native-profiler-start.ts
```diff
@@ -29,6 +29,20 @@ const zodSchema = z.object({
...
+    .boolean()
+    .optional()
+    .describe(
+      "iOS-only. When true, cold-launches the app under the profiler with Malloc Stack Logging " +
+        "enabled so memory leaks carry an allocation backtrace (responsible frame + library). " +
+        "Without it, leaks are still detected but unattributable — Instruments reports " +
+        "'<Call stack limit reached>'. Trade-offs: this RESTARTS the app (current state is lost), " +
+        "adds memory/CPU overhead, and makes the app noticeably slow to launch (every startup " +
+        "allocation records a backtrace), so leave it off for pure CPU/hang profiling. Requires a " +
+        "non-degraded Xcode: on Xcode 26.4–27.0 the cold-launch path is broken, so the call is " +
```
hubgan: This description tells the model the cold-launch path is broken only on Xcode 26.4–27.0, but the degraded-Xcode check that gates this flag treats 26.4 and every version from 27 onward as degraded. So on an Xcode such as 27.5 the flag is documented here as usable while the call is actually refused up front, before the app is touched. The runtime refusal message now reads "Xcode 26.4 and later," which contradicts this text, and the same 26.4–27.0 range is repeated in IOS_PROFILER_REFERENCE.md. Because this describe() string is the contract the model reads when deciding whether to set malloc_stack_logging, the narrower range can lead it to expect attribution on newer Xcodes where the tool will instead reject the call.

hubgan (commented):
Reviewed the malloc_stack_logging changes and exercised them end-to-end against an iPhone 16 Pro simulator on Xcode 26.6 and a Pixel 3a emulator. The production logic and its safety guarantees hold up well: on this (degraded) Xcode the malloc path is refused before the running app is terminated - the app's PID was unchanged after the refusal - and forcing ARGENT_IOS_CAPTURE=all-processes yields a refusal that correctly names the override rather than blaming a degraded Xcode. I also confirmed the `--device` cold launch genuinely deadlocks on 26.6, so the guard is warranted, and on Android the flag is cleanly ignored (no regression). A few inline notes below, mostly around test coverage and docs.

packages/tool-server/test/ios-instruments/malloc-stack-logging.test.ts
```diff
@@ -0,0 +1,642 @@
...
+    const prev = process.env.ARGENT_IOS_CAPTURE;
+    delete process.env.ARGENT_IOS_CAPTURE;
+    try {
+      const spawnFn = vi.fn(() => new StartFakeChild());
+      const execSyncFn = vi.fn((cmd: string) => {
+        if (cmd.includes("xcodebuild")) return "Xcode 16.4\nBuild version 16F6";
+        if (cmd.includes("listapps")) return LISTAPPS_JSON;
+        return "";
+      });
+      const execFileSyncFn = vi.fn((_bin: string, args: string[] = []) => {
```
hubgan: This mock returns `""` for the `plutil` and `listapps` calls, so `resolveAppForLaunch` -> `getInstalledApps` reaches `JSON.parse("")` (which sits outside that function's try/catch) and throws a `SyntaxError` before `simctl terminate` is ever called. The test's two assertions - that an error surfaces, and that no relaunch happened - are both satisfied by that early `SyntaxError`, so the scenario it names (terminate is a no-op for a not-running app, therefore the killed-app relaunch is correctly suppressed) is never actually exercised. Running it with an instrumented mock confirms `terminate` is called zero times and the surfaced error is `SyntaxError: Unexpected end of JSON input`. The production behaviour itself is correct; only the coverage here is illusory. Separately, both relaunch tests carry `execSyncFn` branches for `xcodebuild`/`listapps`/`launchctl` that are now dead (those calls go through `execFileSync`), so the "non-degraded Xcode -> guard passes" comment is misleading: the guard passes because the unmocked `xcodebuild` reads as an undetermined version, not because of the mocked "Xcode 16.4".

packages/tool-server/src/utils/ios-profiler/render.ts
```diff
@@ -413,13 +413,25 @@ function renderFullReport(
     if (unattributedLeaks.length > 0) {
       const objs = unattributedLeaks.reduce((s, b) => s + b.count, 0);
       const bytes = unattributedLeaks.reduce((s, b) => s + b.totalSizeBytes, 0);
+      // When this same capture ALSO produced attributed leaks, the app was launched
+      // under malloc stack logging (the only way a responsible frame is recorded), so
+      // the unattributed remainder is pre-existing / freed-region reuse — telling the
+      // user to "re-run with malloc stack logging" would be advising the thing they
+      // just did. Only when nothing is attributed is the --attach hint apt. The render
+      // layer has no direct capture-mode signal, so it infers it from the attributed count.
+      const hint =
```
hubgan: This conditional - suppressing the "capture with malloc stack logging enabled at launch" advice once the capture has already attributed some leaks - was added here but not to `renderCombinedMemoryLeaks` in `tools/profiler/combined/profiler-combined-report.ts` (around line 404), whose own doc-comment states it mirrors this split. That renderer still emits the unconditional attach-era hint. A malloc-mode capture is now the only way to get attributed and unattributed leaks in the same report, so the combined report will list an attributed leak and then, directly below it, advise enabling malloc stack logging - the exact thing the user just did. The combined-report test exercises that mixed state but doesn't assert the hint wording, so the stale text stays green.

packages/tool-server/src/tools/profiler/native-profiler/native-profiler-start.ts
hubgan: This says the cold-launch path is broken on "Xcode 26.4-27.0", but `isDegraded` in `capture-strategy/select.ts` returns true for `major === 26 && minor >= 4` or any `major >= 27`, with no upper bound - so 27.1, 27.5, 28.x are rejected too. The runtime refusal message in `platforms/ios.ts` correctly says "Xcode 26.4 and later". Since this description is what an agent or user reads to decide whether to pass the flag, the "26.4-27.0" range implies anyone past 27.0 is fine and contradicts both the actual behaviour and the error they will hit. The same range wording also appears in `IOS_PROFILER_REFERENCE.md` and the top-of-file docstring in `select.ts`.

packages/tool-server/src/utils/ios-profiler/IOS_PROFILER_REFERENCE.md
```diff
@@ -56,14 +56,41 @@ All three are wired through `native-profiler-session` (per-device service, keyed
...
+     --env MallocStackLogging=1 --output <…> --no-prompt \
+     --launch -- <path/to/App.app>
+   ```
+   The `.app` path is resolved via `simctl get_app_container`, and any running
+   instance is terminated first for a clean cold start. Trade-off: this restarts
+   the app, adds allocator overhead, and makes the app slow to launch (every
+   startup allocation records a backtrace), so it stays opt-in — leave it off for
+   pure CPU/hang work, where attach (no relaunch, no overhead) is preferable.
+   **Degraded-Xcode guard:** the cold launch needs `xctrace --device`, which is
+   broken on Xcode 26.4–27.0 (the `--device` recording-start handshake records an
```
hubgan: Same range mismatch as the tool schema: `isDegraded` treats 26.4 and everything from 27 upward as degraded (no upper bound), so "26.4-27.0" understates it. A reader on, say, Xcode 27.5 would conclude the cold-launch path works, but it is refused up front.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
```diff
@@ -242,6 +224,118 @@ function resolveExplicitApp(udid: string, name: string): DetectedApp {
...
+        error_code: FAILURE_CODES.NATIVE_PROFILER_APP_BUNDLE_PATH_FAILED,
+        failure_stage: "native_profiler_resolve_app_bundle_path",
+        failure_area: "tool_server",
+        error_kind: "subprocess",
+        ...subprocessFailureMetadata(err, "xcrun_simctl"),
+      },
+      { cause: err instanceof Error ? err : new Error(String(err)) }
+    );
+  }
+  if (!appPath) {
```
hubgan, line 293: This empty-path branch, and the subprocess-error branch of `getAppBundlePath` just above it, have no test coverage - nothing drives `simctl get_app_container` returning an empty string or throwing. The whole `NATIVE_PROFILER_APP_BUNDLE_PATH_FAILED` path is therefore unexercised, unlike the other three new failure codes, which each carry an assertion.

packages/tool-server/src/tools/profiler/native-profiler/native-profiler-start.ts
latekvo: Fixed in 7b87427a.

`isDegraded()` has no upper bound (`major === 26 && minor >= 4` or `major >= 27`), so the schema now reads "on Xcode 26.4 and later the cold-launch path is broken" — matching the runtime refusal ("Xcode 26.4 and later") instead of the narrower "26.4–27.0" that implied 27.1+ was usable. The same reword landed in `IOS_PROFILER_REFERENCE.md`, the `select.ts` precedence docstring, and the `all-processes.ts` / `device.ts` / `types.ts` capture-strategy docstrings (plus the two `26.4–27.0` code comments in `platforms/ios.ts`). The `isDegraded()` implementation docstring already stated "from 26.4 onwards … ALL of 27+", so it was left as-is.

latekvo: Fixed in 7b87427a.

`isDegraded()` has no upper bound (`major === 26 && minor >= 4` or `major >= 27`), so the schema now reads "on Xcode 26.4 and later the cold-launch path is broken" — matching the runtime refusal ("Xcode 26.4 and later") instead of the narrower "26.4–27.0" that implied 27.1+ was usable. The same reword landed in `IOS_PROFILER_REFERENCE.md`, the `select.ts` precedence docstring, and the `all-processes.ts` / `device.ts` / `types.ts` capture-strategy docstrings (plus the two `26.4–27.0` code comments in `platforms/ios.ts`). The `isDegraded()` implementation docstring already stated "from 26.4 onwards … ALL of 27+", so it was left as-is.

packages/tool-server/src/utils/ios-profiler/IOS_PROFILER_REFERENCE.md
latekvo: Fixed in 7b87427a.

`IOS_PROFILER_REFERENCE.md` now reads "broken on Xcode 26.4 and later (… `isDegraded` treats 26.4 and every version from 27 up as broken, with no upper bound)", matching `isDegraded()` and the runtime message. The same reword was applied to the `malloc_stack_logging` tool schema and the `select.ts` / `all-processes.ts` / `device.ts` / `types.ts` docstrings so a reader on e.g. 27.5 no longer concludes the cold-launch path works.

packages/tool-server/src/utils/ios-profiler/render.ts
latekvo: Fixed in 7b87427a.

`renderCombinedMemoryLeaks` now mirrors `render.ts`'s conditional: when the capture attributed some leaks (only possible under malloc stack logging), the unattributed-remainder caveat names the active malloc capture ("Some leaks here were attributed, so malloc stack logging was active …") instead of advising "capture with malloc stack logging enabled at launch" — the thing the user just did.

The combined-report test now asserts the hint wording in both states: the attach-era hint when nothing is attributed, and the malloc-era hint (with `not.toContain("capture with malloc")`) in the mixed state. I confirmed the mixed-state assertion fails against the old unconditional text and passes after the fix.

packages/tool-server/test/ios-instruments/malloc-stack-logging.test.ts
latekvo: Fixed in 7b87427a.

Two parts:

**(1) The illusory relaunch test.** You're right — the mock returned `""` for the `listapps`/`plutil` calls, so `getInstalledApps` reached `JSON.parse("")` and threw `SyntaxError` before `simctl terminate` ever ran. I reproduced it with an instrumented copy of the exact mock: **terminate call count 0**, surfaced error `SyntaxError: Unexpected end of JSON input` — so the no-op-terminate → no-relaunch scenario the test names was never actually exercised (both assertions were satisfied by the early throw).

The mock now feeds a valid listapps plist (`plutil` → `LISTAPPS_JSON`, `listapps` → `<plist/>`) so `resolveAppForLaunch` resolves the installed-but-not-running app and the path reaches `terminate` (which throws `found nothing to terminate`, arming no relaunch). The test now additionally asserts `terminate` **was** attempted and the surfaced error is **not** a `SyntaxError`, alongside the existing no-relaunch assertion.

**(2) Dead `execSyncFn` branches.** Dropped the `xcodebuild`/`listapps`/`launchctl` `execSyncFn` branches in both relaunch tests and the forced-override test — those subprocesses go through `execFileSync` now, so the branches never fired and the "Xcode 16.4" comment misdescribed why the guard passed (it passed on an *undetermined* version, not 16.4). The relaunch / not-running tests now mock the non-degraded Xcode version **through `execFileSync`** so "non-degraded Xcode → guard passes" is real; the forced-override test short-circuits on the `ARGENT_IOS_CAPTURE=all-processes` env override before any `xcodebuild` read, so it just uses a bare `execSync` stub.

latekvo:
Status on the three top-level notes from 2026-07-02:

**[Unattributed-leak render message misleading in malloc mode](https://github.com/software-mansion/argent/pull/351#issuecomment-4863326807)** — resolved. `render.ts` was made capture-mode aware in 88db2f6a (the unattributed-leak note infers the mode from the attributed count and no longer advises enabling malloc stack logging when it was clearly already on). The parallel gap you re-flagged inline — `renderCombinedMemoryLeaks` still emitting the unconditional hint — is now fixed in 7b87427a, with the combined-report test asserting the wording in both states.

**[Residual shell-injection surface in getInstalledApps](https://github.com/software-mansion/argent/pull/351#issuecomment-4863331371)** — resolved. `getInstalledApps` was hardened to two discrete-argv `execFileSync` calls (`simctl listapps` piped into `plutil`) in ac28cfd / the main merge; there is no remaining `execSync` shell string on this path (the only `execSync` token left in `platforms/ios.ts` is a comment describing what was replaced).

**[Feature dormant on current Xcode + tracking note](https://github.com/software-mansion/argent/pull/351#issuecomment-4863324838)** — acknowledged; inherent to Apple's `xctrace --device` regression and the lack of Instruments in CI, and the guard is the intended mitigation. The in-code tracking marker lives in `capture-strategy/select.ts` ("When Apple fixes it, narrow this bound … force the original path on a known-good version via ARGENT_IOS_CAPTURE=device"). As part of 7b87427a the degraded-range wording across the schema, `IOS_PROFILER_REFERENCE.md`, and the capture-strategy docstrings now reads "26.4 and later" (no false upper bound), so nobody reading the docs concludes the feature is live on 27.x when the guard refuses it.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
latekvo: Fixed in bbd2f4e2.

Added two malloc-mode tests in `malloc-stack-logging.test.ts` covering both branches of `getAppBundlePath`:

- **`simctl get_app_container` throws** — asserts `NATIVE_PROFILER_APP_BUNDLE_PATH_FAILED`, that `simctl terminate` was **not** called (bundle-path resolution runs before the destructive terminate, so a failed container resolution leaves the running app untouched), and that `xctrace` never spawned.
- **`simctl get_app_container` returns a blank/whitespace path** — same three assertions (the empty-path guard rejects up front so the argv never becomes `xctrace --launch -- ""`).

So the whole `NATIVE_PROFILER_APP_BUNDLE_PATH_FAILED` path now carries assertions like the other three new failure codes. I confirmed the coverage is real by mutation: disabling the `if (!appPath)` guard makes the empty-path test fail (`err` is `null` — no throw), and moving `getAppBundlePath` below the terminate makes both tests fail on the "terminate not called" assertion; both pass with the shipped ordering.

hubgan (commented):
A few inline notes from a review + end-to-end pass on this branch (real iPhone 17 Pro, Xcode 26.6). The guard refusals, the malloc `--launch` argv, and the terminate/relaunch behavior all checked out live against the simulator — no correctness problems there. The comments below are the smaller things worth a look.

packages/tool-server/src/utils/ios-profiler/render.ts
```diff
@@ -413,13 +413,25 @@ function renderFullReport(
...
+        attributedLeaks.length > 0
```
hubgan: When a capture runs with `malloc_stack_logging` enabled but nothing ends up attributed — a short capture, or leaks that are all freed-region reuse / live in system libraries — `attributedLeaks.length` is 0 and this falls into the `else` branch. That branch states the capture was taken via `xctrace --attach` and suggests re-running with malloc stack logging enabled, but in this case malloc logging was on and the capture used `--launch`, not `--attach`. The capture mode is inferred from the output rather than from the flag that was actually passed, so the single case where the inference is wrong is exactly a real malloc run that happened to attribute nothing. Worth checking whether that path can surface the misleading caveat in practice.

packages/tool-server/src/tools/profiler/combined/profiler-combined-report.ts
```diff
@@ -404,13 +407,26 @@ export function renderCombinedMemoryLeaks(
...
     const bytes = unattributedLeaks.reduce((s, b) => s + b.totalSizeBytes, 0);
+    // Mirror render.ts's split: when this same capture ALSO produced attributed
+    // leaks, the app was launched under malloc stack logging (the only way a
+    // responsible frame is recorded), so the unattributed remainder is freed-region
+    // reuse / pre-recording noise — telling the user to "re-run with malloc stack
+    // logging" would be advising the thing they just did. Only when nothing is
+    // attributed is the --attach hint apt. Infer capture mode from the attributed
+    // count (the render layer has no direct capture-mode signal).
+    const hint =
+      attributedLeaks.length > 0
```
hubgan: Same inference as `render.ts`: a malloc-enabled capture that attributes zero leaks lands in the `else` branch and gets the `--attach` advice even though malloc logging was on. Separately, this hint block (both branches, ~5 lines) is a verbatim copy of the one in `render.ts` — the two have to be kept in sync by hand, so a wording change to one will silently diverge from the other.

packages/tool-server/src/tools/profiler/native-profiler/native-profiler-start.ts
```diff
@@ -29,6 +29,21 @@ const zodSchema = z.object({
...
+    .optional()
+    .describe(
+      "iOS-only. When true, cold-launches the app under the profiler with Malloc Stack Logging " +
+        "enabled so memory leaks carry an allocation backtrace (responsible frame + library). " +
+        "Without it, leaks are still detected but unattributable — Instruments reports " +
+        "'<Call stack limit reached>'. Trade-offs: this RESTARTS the app (current state is lost), " +
+        "adds memory/CPU overhead, and makes the app noticeably slow to launch (every startup " +
+        "allocation records a backtrace), so leave it off for pure CPU/hang profiling. Requires a " +
+        "non-degraded Xcode: on Xcode 26.4 and later the cold-launch path is broken, so the call is " +
+        "rejected up front (re-run without the flag, or set ARGENT_IOS_CAPTURE=device to override if the " +
```
hubgan, line 43: This frames `ARGENT_IOS_CAPTURE` only as an escape hatch to force the device path (`=device`). It doesn't mention the other direction: when `ARGENT_IOS_CAPTURE=all-processes` is set, `malloc_stack_logging` is rejected up front with its own failure code. An operator who keeps that variable exported globally for the normal capture path will hit a rejection this description doesn't lead them to expect.

packages/tool-server/src/utils/ios-profiler/pipeline/01-correlate.ts
```diff
@@ -129,17 +129,27 @@ export function isLeakAttributed(responsibleFrame: string): boolean {
...
+    // Group by object type AND responsible frame: with malloc_stack_logging the
+    // same object type can leak from several distinct call sites, and each site
+    // is its own finding — merging on type alone would attribute them all to
+    // whichever was parsed first and hide the rest. Under `--attach` (no stacks)
+    // every frame is the same sentinel, so this collapses back to per-type.
+    // The delimiter is a NUL, written as the `\u0000` escape rather than a
+    // raw NUL byte: a raw NUL makes git treat this whole source file as binary
+    // and hides the diff. Object types and demangled frames never contain a NUL,
+    // so the composite key can't collide the way a printable separator could.
+    const key = `${leak.objectType}\u0000${leak.responsibleFrame}`;
```
hubgan: The group key embeds `responsibleFrame` verbatim (untrimmed), while `isLeakAttributed` trims before matching. If the parser ever emits more than one unattributed spelling for the same object type — a missing frame defaulting to `Unknown` alongside the `<Call stack limit reached>` sentinel, or trailing whitespace — those share one attribution state but split into separate groups here. Downstream totals stay correct (all unattributed groups are summed into a single caveat), so the only visible effect is the `N unattributed leak group(s)` count. Flagging it because the comment just above claims `--attach` always collapses back to per-type.

packages/tool-server/test/ios-instruments/malloc-stack-logging.test.ts
```diff
@@ -0,0 +1,783 @@
...
+      malloc_stack_logging: true,
+    });
+
+    expect(result.status).toBe("recording");
+    expect(spawnFn).toHaveBeenCalledTimes(1);
+    const [bin, args] = spawnFn.mock.calls[0] as unknown as [string, string[]];
+    expect(bin).toBe("xctrace");
+
+    // launch + env, not attach
+    expect(args).toContain("--launch");
```
hubgan, line 154: This pins the malloc argv well, but two things in the real argv aren't covered. The notify mock makes `registerStartupNotify` return null in every test, so `--notify-tracing-started` is never present — which means the `--launch -- <path>` is-last guarantee is never actually exercised with the notify arg sitting in between (on a real host notify IS present and does precede `--launch`). And `--no-prompt` isn't asserted in the malloc branch, which hand-rolls its own argv apart from the strategy builder. A regression that dropped `--no-prompt` or ordered the notify arg after `--launch --` would still pass here. I confirmed the current argv is correct on a real device, so this is a coverage gap rather than a defect.

hubgan (approved):

packages/tool-server/src/utils/ios-profiler/render.ts
latekvo: Fixed in 737e56b9 — the capture mode is now threaded in for real instead of inferred: `native-profiler-start` stamps `mallocStackLogging` on the session, analyze passes it into the renderer, and a malloc capture that attributes nothing gets its own wording (no `--attach` claim, no advice to re-enable the flag that was just on). A session restored via profiler-load has no capture-mode sidecar, so it renders with `null` and keeps the attributed-count inference as the fallback — same behavior as before for that case.

packages/tool-server/src/tools/profiler/combined/profiler-combined-report.ts
latekvo: Fixed in 737e56b9 — the combined report now reads the session's real `mallocStackLogging` the same way render.ts does, and the duplication is gone: both reports emit the caveat via a single shared `renderUnattributedLeaksNote()` in render.ts, with a test pinning that the two emitted lines are byte-identical so a wording change can't silently diverge again.

packages/tool-server/src/tools/profiler/native-profiler/native-profiler-start.ts
latekvo: Fixed in 9aa05db1 — the description now covers the other direction too: `ARGENT_IOS_CAPTURE=all-processes` (e.g. kept exported globally for the normal capture path) rejects the flag up front since that fallback cannot cold-launch, and it names the remedy (unset it, or set it to `device`).

packages/tool-server/src/utils/ios-profiler/pipeline/01-correlate.ts
latekvo: Fixed in d0b6700b — the group key now uses the frame normalized exactly the way `isLeakAttributed()` matches it: trimmed, with every unattributed spelling ("", "Unknown", the `<Call stack limit reached>` sentinel) collapsed to one bucket. Reproduced first: an export mixing the sentinel, "Unknown" (the parser's default for a missing responsible-frame attribute), and a whitespace-padded sentinel for one object type produced three groups; after the fix it's one, so the per-type collapse the comment claims is now guaranteed rather than assumed. Attributed frames are also keyed trimmed, so whitespace variants of one call site merge too.

packages/tool-server/test/ios-instruments/malloc-stack-logging.test.ts
latekvo: Fixed in e2d9aff5 — added a notify-present test (the notify mock returns a live handle, so `registerStartupNotify` yields non-null) asserting `--notify-tracing-started <name>` precedes `--launch` and that `--launch -- <app>` remains the final three args with the notify pair in the argv, plus `--no-prompt` assertions in the malloc branch (both the notify-null and notify-present shapes). Both pass against the current production argv, matching what you saw on-device — coverage gap, not defect.

packages/tool-server/src/utils/ios-profiler/render.ts
latekvo: Follow-up: a subsequent review pass showed the start-time stamp could re-label the PREVIOUS capture's still-loaded exports when a new recording starts before its analyze. 1c6b5bdf moves the report-facing stamp to stop (paired with `exportedFiles`; the in-flight mode rides a separate `recordingMallocStackLogging`), and freezes the mode into `parsedData` for the drill-down consumers.

packages/tool-server/src/tools/profiler/combined/profiler-combined-report.ts
latekvo: Follow-up: with 1c6b5bdf the combined report now reads the mode frozen into `parsedData` at analyze/load time rather than the live session field, so a recording started after the analyze can no longer re-label the data being rendered.

PR #352, opened 2026-06-17
chore(eslint): fail CI on warnings and unused disable directives

latekvo:
Small follow-up to #349 (which added the ESLint gate). Makes the gate strict so it can't be silently bypassed or accumulate dead suppressions:

- **`reportUnusedDisableDirectives: "error"`** — a stale or unjustified `eslint-disable` is now a hard failure. (The tree had dead directives before the gate existed; #349's `--fix` removed them — this stops them creeping back.)
- **`--max-warnings 0`** in the `lint` script — a future `"warn"`-level rule can't pass CI silently.

Both are no-ops against the current tree (0 disable directives, 0 warnings) — pure guardrails.

Verified locally: `npm run lint` (strict), `tsc --build`, and `prettier --check` all pass.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -25,6 +25,14 @@ export default tseslint.config(
     ],
   },
 
+  // Keep the gate honest: a stale or unjustified `eslint-disable` is a hard
+  // error, so suppressions can't quietly pile up or be used to dodge a rule.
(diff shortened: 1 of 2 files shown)


filip131311 (approved):
sure why not

PR #353, opened 2026-06-17
feat(ios-profiling): signed-binary build + local-test pull

filip131311:
## What

The public half of getting `@argent/ios-profiling`'s native binaries **signed and shippable** through the existing signed-artifact pipeline — without putting any Apple identity in this public repo.

Stacked on #350 (the profiler package only exists on that branch). Paired with an `argent-private` PR that adds the build+sign+publish step.

### Pipeline (after both PRs)

```
software-mansion/argent (PUBLIC)            ── ObjC source stays here (open)
   │  read-only checkout by private CI
   ▼
software-mansion/argent-private (PRIVATE)   ── build-native-binaries.yml: build UNIVERSAL,
   │                                            codesign (Developer ID + hardened runtime + entitlements)
   ▼
software-mansion/argent-private-releases    ── ios-profiler-capture, ios-profiler-mem (per tag)
   │
   ├─ full release  → download-native-binaries.sh → bundle-tools → @swmansion/argent
   └─ local testing → download-ios-profiler.sh argent-daily → bin/darwin (no npm pack/publish)
```

Signed binaries flow one way: private CI → releases → consumers. Never committed to git, never built locally for distribution.

## Changes

- **`build.sh`** — build a **universal** binary (`-arch arm64 -arch x86_64`). It was host-arch-only, which would `ENOEXEC` on x86_64 / Rosetta hosts and ship a broken artifact. Now matches the dylib CI build matrix.
- **`ios-profiler.entitlements`** (new) — hardened-runtime entitlements (`com.apple.security.cs.disable-library-validation`) consumed by argent-private's `codesign --options runtime` step, so the signed binary can `dlopen` Xcode's Apple-signed private frameworks (CoreSimulator / DTX) under a different Team ID. Mirrors the existing `ax-service.entitlements` recipe.
- **`scripts/download-ios-profiler.sh`** (new) — pull the **signed** binaries from `argent-private-releases` into `bin/darwin/` for local testing (`argent-daily` for the prerelease dev loop), decoupled from any npm publish. Pairs with `build.sh`'s existing `PREBUILT_IOS_PROFILER_BIN_DIR` hook.

## Not included (deliberate follow-ups)

- Wiring `bin/darwin/ios-profiler-*` into the published `@swmansion/argent` (bundle-tools + publish verify + `files`) — only needed once the tool-server consumes the package (it's `private: true` and unwired today).

## ⚠️ Must verify before relying on signed artifacts

The entitlements set is **suspected-sufficient**, mirroring ax-service. PR #350 validated with clang's **ad-hoc** signature — Developer ID + hardened runtime is a different trust context. **Acceptance gate: download the CI-signed binary onto a clean second Mac and confirm `coreprofile channel=...` still connects.** If `disable-library-validation` is insufficient, the fallback is ad-hoc signing for these two (documented as non-notarized).

### DIFF
--- a/packages/argent-ios-profiling/ios-profiler.entitlements
+++ b/packages/argent-ios-profiling/ios-profiler.entitlements
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+  <!--
+    Entitlements applied when argent-private's build-native-binaries.yml signs
+    ios-profiler-capture / ios-profiler-mem with a Developer ID under hardened
+    runtime (`codesign --options runtime`).
+
+    disable-library-validation: the capture binary dlopen()s Xcode's private
+    frameworks (CoreSimulator, DTXConnectionServices, DVTInstruments*) to drive
+    the Instruments DTServiceHub. Those are Apple-signed under a *different* Team
+    ID than our Developer ID, so hardened runtime's library validation would
+    refuse the load and the DTX channel would never open. This entitlement lifts
+    that restriction for these two host-side tools.
+
+    NOTE: this is the suspected-sufficient set, mirroring ax-service's recipe.
+    It MUST be verified on a clean second Mac (download the CI-signed binary and
+    confirm `coreprofile channel=...` still connects) before relying on it for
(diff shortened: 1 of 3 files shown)


filip131311:
Consolidated into #350. All commits from this branch (signed-binary build + local-test pull, plus the entitlements XML-comment fix `dfd20a8`) are now in #350's branch `feat/argent-ios-profiling` — verified byte-identical and as a contained ancestor. GitHub auto-marked this MERGED once its commits became reachable from the base. The ios-profiler work now ships as a single PR (#350).

PR #357, opened 2026-06-17
fix(release): auto-sync the npm 'next' dist-tag to the highest published version

latekvo:
## Problem

The `next` dist-tag drifts. Right now:

```
npm dist-tag ls @swmansion/argent
latest: 0.11.0
next:   0.7.0-next.4   # 8 releases stale
```

A stable `npm publish` only moves `latest`, and `npm publish --tag next` moves `next` only when a *prerelease* ships. So once `0.7.0-next.4` was tagged `next`, every later stable release (0.7.0 → 0.11.0) left `next` frozen — hence the recurring "we set `next` once then forget" problem.

## Fix

`scripts/sync-next-dist-tag.mjs` re-points `next` at the **highest published version overall** (prerelease *or* stable, per SemVer precedence). It is:

- **Automatic** — runs as a final step in both `publish.yml` and `publish-next.yml`, so `next` self-heals on every publish.
- **Manual** — exposed via a new `Sync 'next' dist-tag` workflow (`workflow_dispatch`, with a `dry_run` toggle) to repair the tag **without republishing**. Running it now would move `next` 0.7.0-next.4 → 0.11.0.
- **Idempotent** — no-op when `next` already points at the max.
- **Dependency-free at runtime** — no `npm ci` needed in the lightweight manual job.

## Verification

- Live dry-run against the real registry resolves `next` → `0.11.0` (the current max) as expected.
- `scripts/sync-next-dist-tag.test.mjs` proves the hand-rolled SemVer comparator matches the `semver` library across every pair of the real version history plus prerelease edge cases. Wired into the unit-tests CI (`npm run test:scripts`); `semver` added as a devDep (test oracle only).
- `npm run lint`, `npm run typecheck:scripts`, and `prettier --check` all pass.

## Note (out of scope)

A stable `npm publish` of a *backport* (e.g. `0.8.2` while `0.11.0` is out) would still wrongly move `latest` — that's a separate, pre-existing npm behavior and not addressed here. The `next` sync handles backports correctly (it never regresses below the max).


hubgan (approved):

filip131311 (dismissed):
npm view argent versions is broken here

.github/workflows/publish.yml
```diff
@@ -95,3 +95,11 @@ jobs:
       - run: npm publish -w @swmansion/argent
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      # Keep `next` pointing at the highest published version (prerelease or
+      # stable). A stable `npm publish` only moves `latest`, so without this the
+      # `next` tag would freeze on the last prerelease. Idempotent + self-healing.
+      - name: Sync 'next' dist-tag to highest published version
+        run: node scripts/sync-next-dist-tag.mjs
```
hubgan, line 109: Post-publish sync step has no failure tolerance. The sync runs after npm publish. A transient npm view / dist-tag add failure would turn the release workflow red even though the publish already succeeded.

.github/workflows/publish-next.yml
```diff
@@ -105,3 +105,11 @@ jobs:
       - run: npm publish -w @swmansion/argent --tag next
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      # Re-point `next` at the highest published version. `npm publish --tag next`
+      # already tags this prerelease, but this also self-heals `next` if it was
+      # left stale on an older version. Idempotent.
+      - name: Sync 'next' dist-tag to highest published version
+        run: node scripts/sync-next-dist-tag.mjs
```
hubgan, line 120: Post-publish sync step has no failure tolerance. The sync runs after npm publish. A transient npm view / dist-tag add failure would turn the release workflow red even though the publish already succeeded.

.github/workflows/publish-next.yml
```diff
@@ -105,3 +105,11 @@ jobs:
...
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
hubgan, line 122: Immediately after publish, npm view may serve a cached packument missing the just-published version. In publish-next.yml this could theoretically move next backward from the just-tagged prerelease to the previous max.

.github/workflows/publish.yml
latekvo: Fixed in 541f9026.

Added `continue-on-error: true` to the post-publish sync step. By the time it runs, `npm publish` has already succeeded and is irreversible, and the sync itself is idempotent + self-healing (it re-runs on the next publish, and the manual `Sync 'next' dist-tag` workflow can repair it on demand). So a transient `npm view` / `dist-tag add` blip no longer turns an otherwise-successful release red.

The manual `sync-next-tag.yml` deliberately does **not** get `continue-on-error` — there its sole job is the sync, so a failure should stay visible.

.github/workflows/publish-next.yml
latekvo: Fixed in 541f9026.

Added `continue-on-error: true` to the post-publish sync step. By the time it runs, `npm publish` has already succeeded and is irreversible, and the sync itself is idempotent + self-healing (it re-runs on the next publish, and the manual `Sync 'next' dist-tag` workflow can repair it on demand). So a transient `npm view` / `dist-tag add` blip no longer turns an otherwise-successful release red.

The manual `sync-next-tag.yml` deliberately does **not** get `continue-on-error` — there its sole job is the sync, so a failure should stay visible.

.github/workflows/publish-next.yml
latekvo: Fixed in 541f9026.

The sync now only ever *advances* `next` — it never moves it backward. The move is gated by a new `shouldAdvanceTag(max, current)` predicate that fires only when `max` is strictly newer than the current tag (or the tag is unset/unparseable). If a stale CDN-cached packument omits the just-published prerelease, the computed `max` lags behind the freshly-tagged `next`, and the script now no-ops instead of demoting it.

Reproduced with a stubbed `npm` returning a stale `versions` list while `dist-tags.next` already pointed at the new prerelease:
- before: `Updating 'next': 0.12.0-next.0 -> 0.11.0` (backward move)
- after: `'next' (0.12.0-next.0) is already at or ahead of the highest published version (0.11.0); nothing to do.`

The self-heal-forward and idempotent paths are unaffected, and the predicate is covered by a unit test.

hubgan:
> npm view argent versions is broken here

I think it's working. `npm view @swmansion/argent versions --json ` is returning:

```
[
  "0.5.1",
  "0.5.2",
  "0.5.3",
  "0.6.0-next.1",
  "0.6.0-next.2",
  "0.6.0",
  "0.6.1",
  "0.7.0-next.4",
  "0.7.0",
  "0.7.1",
  "0.8.0",
  "0.8.1",
  "0.9.0",
  "0.10.0",
  "0.11.0",
  "0.12.0",
  "0.12.1"
]
```

PR #363, opened 2026-06-17
fix(argent): ship tree-sitter so react-profiler-component-source finds components

latekvo:
## Problem
`react-profiler-component-source` returns `found:false` for **every** component in a standard TS/TSX project. Root cause: `@swmansion/argent` never declared `tree-sitter` / `tree-sitter-typescript` as runtime deps, so the bundle's `require("tree-sitter")` throws, the error is swallowed, and the AST index is always empty.

Verified on the shipped 0.12.0 bundle: `tree-sitter` and `tree-sitter-typescript` are `MODULE_NOT_FOUND` from the installed tool-server.

## Fix
- Declare both as `dependencies` of `@swmansion/argent` and mark them `external` in the tools bundle (native `.node` addons can't be inlined by esbuild; they must resolve from the install at runtime).
- Index `memo()` / `forwardRef()` / `React.memo()` component declarations (their value node is a `call_expression`, previously skipped entirely — and profiler-flagged components are disproportionately memo-wrapped).
- Surface an actionable message when the parser can't load instead of a misleading "not found".

## Verification
- Installed the two deps into the shipped 0.12.0 package, restarted the tool-server, re-ran the tool: `LoggedOut → LoggedOut.tsx:35`, `SplashScreen → SplashScreen.tsx:23`, `Button`, `Text` all now resolve (were `found:false` for all).
- New `ast-index` unit test covers plain/default/arrow + memo/forwardRef/React.memo/nested forms and guards `treeSitterAvailable`.
- `tsc --build` clean, 1254 tool-server tests pass, eslint + prettier clean.

Supersedes #287 (that branch was based on a much older `main` and could not be rebased cleanly through the `bundle-tools.cjs` rewrite). Closes #287.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,6 +21,16 @@ updates:
       prefix: "chore(deps)"
       prefix-development: "chore(deps-dev)"
       include: "scope"
+    ignore:
+      # tree-sitter 0.25.0 ships no Node 24 prebuilds and its binding.gyp forces
+      # C++17 while Node 22+ v8 (cppgc) headers require C++20, so the native
+      # build fails `npm install`/`npm ci` on current Node. Stay on ^0.21.1,
+      # which ships N-API prebuilds (no compile). Patch updates stay allowed;
+      # drop this ignore once upstream restores prebuilt binaries.
+      - dependency-name: "tree-sitter"
+        update-types:
+          - "version-update:semver-minor"
+          - "version-update:semver-major"
     groups:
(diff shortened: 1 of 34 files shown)


latekvo:
Closing — superseded. The tree-sitter/electron externalization and the `tree-sitter`/`tree-sitter-typescript` runtime deps this aimed to add already landed on main (via #287). The net diff vs main is only a reorder of the `external` array (esbuild treats it as a set → no behavioral change) plus a comment, so there's nothing left to merge.

PR #365, opened 2026-06-17
fix(debugger): return object/array results from debugger-evaluate

latekvo:
## Problem
`debugger-evaluate` silently returns nothing for any expression that evaluates to an object or array. `CDPClient.evaluate` calls `Runtime.evaluate` without `returnByValue`, so a non-primitive result comes back as a RemoteObject reference with no `value` field, and the tool drops it (only primitives/strings came through).

## Fix
Pass `returnByValue` + `awaitPromise` (default on) so objects/arrays serialize by value and promise-returning expressions resolve to their value. `evaluateWithBinding` opts both off to keep its fire-and-forget binding path unchanged (it delivers its payload via a side binding, not the evaluate return).

## Verification
- New cdp-client unit tests assert `evaluate` sends `returnByValue:true` + `awaitPromise:true` and returns an object result, and that `evaluateWithBinding` sends both `false`.
- Full cdp-client + reconnect suites pass; 1254 tool-server tests pass; eslint + prettier clean.
- The real-Hermes `returnByValue` path is already exercised in production by `react-profiler-fiber-tree` / `react-profiler-renders`.

### DIFF
--- a/packages/tool-server/src/tools/debugger/debugger-evaluate.ts
+++ b/packages/tool-server/src/tools/debugger/debugger-evaluate.ts
@@ -19,7 +19,7 @@ export const debuggerEvaluateTool: ToolDefinition<
 > = {
   id: "debugger-evaluate",
   description: `Execute arbitrary JavaScript in the app's JS runtime via CDP — Hermes on iOS / Android, V8 on Chromium.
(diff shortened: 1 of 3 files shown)


hubgan (approved):
Circular objects now throw instead of returning nothing. {...; o.self=o} now throws TypeError: cyclical structure in JSON object; before it silently returned nothing. Many RN runtime objects are cyclic (fiber nodes, navigation refs, global). This is inherent to returnByValue and is arguably better (clear signal vs. silent no-op), but it's a behavior change not mentioned in the PR. A one-line note in the tool description would set expectations.

PR #367, opened 2026-06-18
feat(tool-server): add tool-server event log

sjchmiela:
We want to display a nicely parsed and rendered log of events that happened over a span of an Argent session.

Current log file is not designed to be machine-readable, so adding a JSONL-formatted log of events (gated by `tool-server-event-log`).

I eventually want this 😋 https://x.com/j3nnings/status/2069072177584849192

When enabled, the tool server writes Bunyan-compatible records to `ARGENT_EVENT_LOG` or `~/.argent/tool-server-events.jsonl`. The log captures tool-server lifecycle events and registry lifecycle events, including tool invocations and others.

To make messages nicer, added `interaction` object to tool definition. This allows tools to define a nicer message in log (print "Screenshot capture completed." instead of "Tool `screenshot` completed.") Adding these to the existing tools is a follow up PR matter I think.

### DIFF
--- a/packages/tool-server/package.json
+++ b/packages/tool-server/package.json
@@ -25,6 +25,7 @@
     "@fails-components/webtransport-transport-http3-quiche": "^1.6.3",
     "@moq/lite": "^0.2.4",
     "@moq/net": "^0.1.1",
+    "bunyan": "^1.8.15",
     "bytes": "^3.1.2",
     "dotenv": "^17.4.2",
     "express": "^5.2.1",
@@ -42,6 +43,7 @@
     "zod": "^4.0.0"
   },
   "devDependencies": {
+    "@types/bunyan": "^1.8.11",
     "@types/bytes": "^3.1.5",
     "@types/express": "^5.0.6",
     "@types/ini": "^4.1.1",
--- a/packages/tool-server/src/event-log.ts
+++ b/packages/tool-server/src/event-log.ts
@@ -22,11 +22,10 @@ export type EventLogRecord = { type: string; msg: string } & Record<
   EventLogValue | undefined
 >;
 
-type EventLogLevel = "debug" | "info" | "warn" | "error" | "fatal";
+type EventLogLevel = "info" | "warn" | "error" | "fatal";
 
 export interface ToolServerEventLog {
   filePath: string;
-  debug: (record: EventLogRecord) => void;
   info: (record: EventLogRecord) => void;
   warn: (record: EventLogRecord) => void;
   error: (record: EventLogRecord) => void;
@@ -78,7 +77,6 @@ export function createToolServerEventLog({
 
   return {
     filePath,
-    debug: (record) => log("debug", record),
     info: (record) => log("info", record),
     warn: (record) => log("warn", record),
     error: (record) => log("error", record),
--- a/packages/tool-server/src/event-log.ts
+++ b/packages/tool-server/src/event-log.ts
@@ -153,10 +153,10 @@ export function attachRegistryEventLogger(registry: Registry, eventLog: ToolServ
     });
   });
 
-  registry.events.on("toolInvoked", (toolId: string, toolInvocationId: string) => {
+  registry.events.on("toolInvoked", (toolId: string, toolInvocationId: string, msg: string) => {
     eventLog.info({
       type: "tool.invoked",
-      msg: `Tool ${toolId} was invoked.`,
+      msg,
       toolId,
       toolInvocationId,
     });
@@ -164,10 +164,10 @@ export function attachRegistryEventLogger(registry: Registry, eventLog: ToolServ
 
   registry.events.on(
     "toolCompleted",
-    (toolId: string, toolInvocationId: string, durationMs: number) => {
+    (toolId: string, toolInvocationId: string, durationMs: number, msg: string) => {
       eventLog.info({
         type: "tool.completed",
-        msg: `Tool ${toolId} completed in ${durationMs.toFixed(2)} ms.`,
+        msg,
         toolId,
         toolInvocationId,
         durationMs,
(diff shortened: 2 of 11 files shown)


packages/tool-server/src/index.ts
```diff
@@ -121,14 +125,26 @@ export function start(): void {
   // ── Bootstrap ─────────────────────────────────────────────────────
   const registry = createRegistry();
   attachRegistryLogger(registry);
+  const eventLog = isFlagEnabled("tool-server-event-log")
+    ? createToolServerEventLog({
+        filePath:
+          process.env.ARGENT_EVENT_LOG ||
+          path.join(homedir(), ".argent", "tool-server-events.jsonl"),
+      })
+    : null;
```
hubgan: Creating the event log is the first thing startup does, and it writes to disk right away. If that path can't be written to this throws, and nothing catches it, so the whole tool-server fails to start. The error lands in the crash handler, which immediately tries to set `shutdownReason`. Because we crashed before that line ran, the crash handler itself throws "Cannot access 'shutdownReason' before initialization", which hides the real cause and exits with code 7 instead of a clean shutdown. Since this is an optional, off-by-default feature, failing to open the log should just mean "run without the log," not kill the server.

packages/tool-server/src/event-log.ts
```diff
@@ -0,0 +1,157 @@
...
+  filePath: string;
+}
+
+function serializeError(error: Error): Record<string, unknown> {
+  const serialized = bunyan.stdSerializers.err(error) as Record<string, unknown>;
+  if (error.cause instanceof Error) {
+    serialized.cause = serializeError(error.cause);
+  }
+  return serialized;
+}
```
hubgan: On success the log only records the tool name and a duration, which is great. But on failure this serializes the error's full message, stack, and every nested cause. Tool errors are built from the failing subprocess command line, so the whole argument list ends up in the file.

packages/tool-server/package.json
```diff
@@ -13,6 +13,7 @@
     "typecheck:tests": "tsc --noEmit -p tsconfig.test.json"
   },
   "dependencies": {
+    "@argent/configuration-core": "file:../configuration-core",
```
hubgan, line 16: This package is already declared at line 21

packages/tool-server/src/event-log.ts
```diff
@@ -0,0 +1,157 @@
...
+      err: serializeError,
+    },
+    streams: [
+      {
+        level: "info",
+        type: "raw",
+        stream: {
+          write(record: object): void {
+            appendFileSync(filePath, JSON.stringify(record) + "\n");
+          },
```
hubgan, line 66: Each event opens the file, writes one line, and closes it - synchronously - and this runs for every tool invoke/complete before the tool's result is returned (and again per service state change, and per step inside run-sequence). Synchronous writes pause everything else the server is doing until the write finishes, and re-opening the file per line is the slow way to do it. On a network-mounted home directory this can stall all concurrent requests. I think it shouldn't be a problem, but it's worth looking into it.

packages/tool-server/src/event-log.ts
```diff
@@ -0,0 +1,157 @@
...
+  writeFileSync(filePath, "");
+
+  const logger = bunyan.createLogger({
+    name: "argent-tool-server",
+    serializers: {
+      err: serializeError,
+    },
+    streams: [
+      {
+        level: "info",
```
hubgan, line 61: The logger only records "info" and above, but the interface still exposes a `debug()` method. Anything sent through `debug()` is silently discarded, so the method looks usable but does nothing.

packages/tool-server/src/event-log.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  return {
+    filePath,
+    debug: (record) => log("debug", record),
+    info: (record) => log("info", record),
+    warn: (record) => log("warn", record),
+    error: (record) => log("error", record),
+    fatal: (record) => log("fatal", record),
+    dispose: () => {},
```
hubgan: `dispose()` is called during shutdown but its body is empty. That's technically fine right now (each write closes the file itself), but it implies a cleanup step that doesn't exist.

packages/tool-server/src/index.ts
sjchmiela: 👍 Wrapped in `try`-`catch`!

https://github.com/software-mansion/argent/blob/d4e28e0aad7e201c7cf9d429f614d14a8df87756/packages/tool-server/src/index.ts#L136-L142

packages/tool-server/src/event-log.ts
sjchmiela: I didn't know about "failure signal"! This, I believe, is much nicer.

https://github.com/software-mansion/argent/blob/4f934aa94627aba719b592de3a43c43ab27c3f1e/packages/tool-server/src/event-log.ts#L124-L138

packages/tool-server/package.json
sjchmiela: Fixed via rebase!

packages/tool-server/src/event-log.ts
sjchmiela: Yea, I went for simplicity, but maybe you're right. Replaced with a `WriteStream`, maybe better.

packages/tool-server/src/event-log.ts
sjchmiela: Removed `debug()`. Good call!

packages/tool-server/src/event-log.ts
sjchmiela: With the implementation moved to `WriteStream`, `dispose()` _does_ do stuff now!

filip131311 (approved):
Thank you I will merge it will ship in the Wednesday release

PR #369, opened 2026-06-18
feat(ios): control physical iOS devices over CoreDevice (experimental)

latekvo:
## Physical iOS device support (experimental)

Adds app-free control of a **physical iPhone (iOS 27+)** over Apple's CoreDevice
"remote control" services (the path Xcode's device window uses) via
[`pymobiledevice3`](https://github.com/doronz88/pymobiledevice3) — **no app installed on the
device, no WebDriverAgent, no code signing**. Gated behind a new `physical-ios-devices` flag.

Mirrors the existing physical-Android split: a real iPhone is a separate backend
(`blueprints/core-device.ts`), selected by `device.kind === "device"`, the same way
`android_device` differs from the emulator controller.

### Supported on physical iOS
`list-devices` (shows the iPhone as `kind: "device"`), `screenshot`, `gesture-tap`,
`gesture-swipe`, `button`, `launch-app` (via `devicectl`), and `describe` (the SpringBoard
home-screen grid — see below). Every other tool returns a clear "not supported on physical iOS"
error rather than mis-routing.

### Requirements (⚠️ needs sudo, once)
- **iOS 27+** — Apple gates host-driven input ("remote control") to 27.0+; below that the
  device returns `CoreDeviceError 9021` and only screenshots work.
- `pymobiledevice3` installed; iPhone connected, unlocked, trusted, Developer Mode on.
- A running CoreDevice tunnel: **`sudo pymobiledevice3 remote tunneld`** — this needs root
  *once* to create the tunnel interface; every Argent command then runs unprivileged.
- Enable the flag: `argent enable physical-ios-devices`.

The sudo/tunnel requirement is surfaced in the runtime error, the README, the flag
description, and the `list-devices` description.

### Performance: persistent CoreDevice sidecar
Interactions no longer spawn the `pymobiledevice3` CLI per call (~0.8s each, ~0.5s of which is
just `import pymobiledevice3`). A single long-lived helper per device connects the RSD tunnel and
opens the HID / screenshot services once, then serves newline-delimited JSON on stdio. Warm
`tap`/`button`/`swipe` drop from ~800ms to **~150-320ms (5x)**, measured live through the tool-server
on an iPhone 15 (iOS 27). The helper is a small Python program base64-embedded in
`coredevice-agent.ts` (survives esbuild, no build-pipeline wiring), materialized to a temp file and
run with pmd3's own venv interpreter; `core-device.ts` maps its responses (incl. the 9021 gate) to
`FailureError`s and disposes the process with the service.

### `describe` — the real on-screen accessibility tree (app-free, any app)
`describe` on a physical iPhone returns the device's **live accessibility tree** — the frontmost
app's elements (or the home screen) — read app-free via the iOS-26+ axAudit service over CoreDevice.
Labels, values, traits (mapped to roles) and reading order are exact.

The unlock: iOS 26 re-plumbed the accessibility daemon onto RSD and it now needs the **RSDCheckin**
handshake before it accepts DTX framing. pmd3's `DtxServiceProvider` opens RSD services with a raw
connect and *skips* RSDCheckin, so iOS 26/27 dropped it on the first byte — which had read as an
"Apple wall". The fix is to open the service through `start_lockdown_service` (which performs
RSDCheckin); the sidecar does that. (`installation_proxy` worked over the same tunnel because it
already used that path — the entitlement was never the problem.)

**Verified live (iPhone 15 / iOS 27):** `describe` Settings → read the "Other…" button's frame →
`gesture-tap` its centre → the Join-Wi-Fi form opened. On the home screen it lists every app + widget.

Frames: Apple exposes no per-element geometry on hardware, so exact rects come only from the
accessibility audit (a subset of elements, correlated to the tree by element id); the rest are
interpolated from reading-order neighbours — enough to tap a list row, with a hint to confirm precise
taps via `screenshot`. Pixel-exact in-app frames/taps would need an on-device XCUITest runner
(code-signing), which is a possible follow-up.

### Not supported yet (clean rejections)
keyboard/typing, pinch & rotate (multi-touch), `open-url`, `reinstall-app`, `restart-app`,
`native-*` (simulator-only mechanisms).

### Verification
- `tsc --build`, `typecheck:tests`, `typecheck:scripts`, `prettier --check .` all clean;
  **1207 tool-server + 38 configuration-core unit tests pass** (new suite covers
  classification, `devicectl` parsing, coord mapping).
- **E2E live on a physical iPhone 15 (iOS 27)** through the tool-server HTTP API:
  `list-devices` (kind:device), `screenshot`, `button` (woke device), `gesture-swipe`
  (Notification Center), `gesture-tap` (opened Messages).
- **Sidecar + describe re-verified live (iPhone 15, iOS 27) through the tool-server:**
  `list-devices` → single `kind:"device"` (no phantom sims); warm `gesture-tap` **155ms**,
  `button` **187ms**, `gesture-swipe` **323ms** (vs ~810ms per CLI call); `describe` → home-screen
  tree, and tapping "Settings" via its `describe` frame opened Settings. The in-app AX wall was
  reproduced against the real `axAuditDaemon` (see the `describe` section).
- **Clean-setup audit:** flag OFF hides the device; flag ON shows it; missing
  pymobiledevice3 → install hint; no tunnel → the `sudo … tunneld` error above.

### Notes
Interactions run through a persistent `pymobiledevice3` sidecar (one process per device); a
native re-implementation in the Rust simulator-server is still a possible follow-up. Env overrides:
`ARGENT_PYMOBILEDEVICE3`, `ARGENT_PMD3_TUNNELD_PORT`.


latekvo:
Describe doesn't work well yet. This is the primary blocker as of now

latekvo:
Will likely go with #538 instead. Sorry Opus, Fable is my new friend now.

latekvo:
Superseded by #538, which relocates physical-iOS control to radon's native `ios_device` simulator-server controller (software-mansion/radon#146) — no pymobiledevice3 sidecar, no root tunnel, plus live streaming/recording for free. Closing in favor of that approach.

PR #371, opened 2026-06-18
ci: device E2E smoke tests for every hosted-CI platform

latekvo:
## What

`wayland-e2e.yml` boots an Android AVD on Linux under headless Weston and asserts the full pipeline works without a display. It guards exactly one cell of the support matrix. This adds the same boot→screenshot→gesture-tap smoke test for every other (target × host) cell that can run on GitHub-hosted runners.

| target | host | job | status |
|---|---|---|---|
| Android emulator | Linux (KVM) | `wayland-e2e.yml` (existing) | ✅ |
| iOS simulator | macOS | `ios-sim-macos` | ✅ |
| Chromium / Electron | Linux (Xvfb) | `chromium-linux` | ✅ |
| Chromium / Electron | macOS | `chromium-macos` | ✅ |

## How

- **`scripts/e2e/drive-device.sh`** — shared driver. Boots one device through the tool-server, then asserts `booted:true` → screenshot has real (non-blank) pixels → `gesture-tap` round-trips. Every tool takes the device id as `udid` and screenshots come back as the same `data.image.hostPath` envelope on all platforms, so the body is uniform; each job just supplies the cell-specific boot JSON + device id.
- **`packages/tool-server/test/fixtures/electron-smoke-app/`** — minimal self-contained Electron app the Chromium jobs point `electronAppPath` at. It vendors its own electron via its own lockfile (kept out of the repo root lockfile); CI `npm ci`s it so `boot-electron` resolves `./node_modules/.bin/electron`.
- **`.github/workflows/e2e-device-smoke.yml`** — three jobs, `workflow_dispatch` + path-filtered `pull_request`. iOS downloads the darwin native binaries (injection must succeed for `bootIos` to report `booted:true`); Chromium-Linux runs under Xvfb with `--no-sandbox`. The tool-server is started with `TS_NODE_TRANSPILE_ONLY=1` so cold ts-node startup doesn't flakily exceed the readiness poll on loaded macOS runners (types are still enforced by the typecheck job).

## Known gaps (supported on real machines, not runnable on hosted runners)

- **Android emulator on macOS** — the arm64 emulator needs HVF (`-enable-hvf`), but hosted macOS runners are themselves VMs with no nested virtualization. Confirmed by running it: qemu dies with `HVF error: HV_UNSUPPORTED` ~18s into boot, regardless of GPU mode or RAM. The Android boot/screenshot/tap path is already regression-guarded on Linux via KVM (`wayland-e2e.yml`), so macOS adds no runnable coverage.
- **Physical devices** — real iOS (CoreDevice HID) and real Android (adb-USB) have no attached hardware on hosted runners.

Both would need self-hosted runners (a Mac exposing HVF / wired-up devices). Documented in the workflow header.

## Verification

- All checks green: the 3 device jobs + ESLint + Prettier + Unit tests.
- Locally (before CI): ran `drive-device.sh` end-to-end against a real tool-server → boot `booted:true`, screenshot **142 KB** (floor 20 KB), tap `tapped:true`, exit 0; and confirmed via the Argent stack that a tap at (0.5, 0.5) lands on the fixture's centered button and increments its counter.

### DIFF
--- a/.github/workflows/e2e-device-smoke.yml
+++ b/.github/workflows/e2e-device-smoke.yml
@@ -38,6 +38,12 @@ on:
       - "packages/tool-server/src/tools/devices/**"
       - "packages/tool-server/src/tools/gesture-tap/**"
       - "packages/tool-server/src/tools/screenshot/**"
+      # Transport layer that drives screenshot/gesture-tap for each platform —
+      # where the #391 screenshot regression lived, so changes here must re-run
+      # these smoke jobs too.
+      - "packages/tool-server/src/blueprints/simulator-server.ts"
+      - "packages/tool-server/src/blueprints/chromium-cdp.ts"
+      - "packages/tool-server/src/utils/simulator-client.ts"
       - "packages/tool-server/test/fixtures/electron-smoke-app/**"
       - "scripts/e2e/drive-device.sh"
       - ".github/workflows/e2e-device-smoke.yml"
--- a/.github/workflows/e2e-device-smoke.yml
+++ b/.github/workflows/e2e-device-smoke.yml
@@ -1,0 +1,238 @@
+name: Device E2E smoke
+
+# Boot-and-drive smoke tests for every (target × host) cell of the support
+# matrix that runs on GitHub-hosted runners. Each job boots ONE device through
+# the tool-server and asserts the headless interaction pipeline end-to-end via
+# scripts/e2e/drive-device.sh: booted:true → screenshot has real pixels →
+# gesture-tap round-trips.
+#
+#   ┌────────────────────┬───────────────┬──────────────────────────────────┐
+#   │ target             │ host          │ where                            │
+#   ├────────────────────┼───────────────┼──────────────────────────────────┤
+#   │ Android emulator   │ Linux (KVM)   │ wayland-e2e.yml (separate file:  │
+#   │                    │               │ also asserts gpu.mode=swiftshader)│
+#   │ iOS simulator      │ macOS         │ ios-sim-macos (this file)        │
+#   │ Chromium / Electron│ Linux (Xvfb)  │ chromium-linux (this file)       │
+#   │ Chromium / Electron│ macOS         │ chromium-macos (this file)       │
+#   └────────────────────┴───────────────┴──────────────────────────────────┘
+#
+# KNOWN GAPS — cells Argent supports on real machines but that GitHub-hosted
+# runners can't run, so they are intentionally not jobs here:
+#   - Android emulator on macOS: the arm64 emulator needs HVF (`-enable-hvf`),
+#     but hosted macOS runners are themselves VMs with no nested virtualization
+#     — qemu dies with "HVF error: HV_UNSUPPORTED" seconds into boot. The
+#     Android boot/screenshot/tap path is already regression-guarded on Linux
+#     via KVM (wayland-e2e.yml), so macOS adds no runnable coverage here.
+#   - Physical devices: real iOS (CoreDevice HID) and real Android (adb-USB)
+#     have no attached hardware on hosted runners.
+#   Both would need self-hosted runners (a Mac exposing HVF / wired-up devices).
+#
+# Manual + on-change trigger only — the macOS jobs are slow and billed.
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "packages/tool-server/src/tools/devices/**"
+      - "packages/tool-server/src/tools/gesture-tap/**"
+      - "packages/tool-server/src/tools/screenshot/**"
+      - "packages/tool-server/test/fixtures/electron-smoke-app/**"
+      - "scripts/e2e/drive-device.sh"
+      - ".github/workflows/e2e-device-smoke.yml"
+
+jobs:
+  # ── iOS simulator on macOS ────────────────────────────────────────────────
+  ios-sim-macos:
+    name: Boot + drive iOS simulator (macOS)
+    runs-on: macos-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v6
+        with:
+          node-version: "20"
+
+      - name: Install deps + build
+        run: |
+          npm ci
+          npm install @rollup/rollup-darwin-arm64 --no-save
+          npx tsc --build
+
+      - name: Download signed native binaries (darwin)
+        # iOS needs the simulator-server (screenshot/gesture-tap backend) AND the
+        # dylibs + ax-service: bootIos() only returns booted:true when native-
+        # devtools injection does NOT give up (boot-device.ts), so missing dylibs
+        # would surface here as a hard failure rather than a silent skip.
+        run: |
+          bash scripts/download-simulator-server.sh
+          bash scripts/download-native-binaries.sh
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Start tool-server
+        run: |
+          mkdir -p ~/.argent
+          : > ~/.argent/tool-server.log
+          cd packages/tool-server
+          # TS_NODE_TRANSPILE_ONLY skips ts-node's whole-program type-check (the
+          # typecheck job already enforces types). Without it a cold start on a
+          # loaded macOS arm runner can take >60s to bind and flakily times out.
+          nohup env ARGENT_PORT=3033 TS_NODE_TRANSPILE_ONLY=1 npx ts-node src/index.ts start \
+            > ~/.argent/tool-server.log 2>&1 &
+          for i in $(seq 1 60); do
+            curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:3033/tools | grep -q 200 \
+              && { echo "tool-server up t+${i}s"; exit 0; }
+            sleep 1
+          done
+          echo "tool-server failed to start"; cat ~/.argent/tool-server.log; exit 1
+
+      - name: Pick an available simulator UDID
+        run: |
+          UDID=$(xcrun simctl list devices available --json | python3 -c "import json,sys; d=json.load(sys.stdin)['devices']; print(next(dev['udid'] for rt in d for dev in d[rt] if dev.get('isAvailable') and 'iPhone' in dev['name']))")
+          echo "Chosen simulator: $UDID"
+          echo "UDID=$UDID" >> "$GITHUB_ENV"
+
+      - name: Boot + drive
+        run: bash scripts/e2e/drive-device.sh
+        env:
+          BOOT_JSON: '{"udid":"${{ env.UDID }}"}'
+          DEVICE_ID: ${{ env.UDID }}
+
+      - if: always()
+        uses: actions/upload-artifact@v4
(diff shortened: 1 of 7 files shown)


scripts/e2e/drive-device.sh
```diff
@@ -0,0 +1,87 @@
...
+fi
+
+echo "::group::screenshot"
+SHOT_JSON="$ARTIFACT_DIR/smoke-shot.json"
+curl -sS -m 60 -X POST "${BASE_URL}/screenshot" \
+  -H 'Content-Type: application/json' \
+  -d "{\"udid\":\"${DEVICE_ID}\"}" >"$SHOT_JSON"
+# The screenshot tool returns an ArtifactHandle ({ image: { hostPath, ... } });
+# the job runs co-located with the tool-server so hostPath is readable here.
+HOST_PATH=$(python3 -c "import json;print(json.load(open('$SHOT_JSON'))['data']['image']['hostPath'])")
```
hubgan: The screenshot step pipes the response straight into python `['data']['image']['hostPath']` without checking for an error field first. On any screenshot error the job dies with a raw Python KeyError traceback instead of a clean `::error::` - I hit this live; the real cause was buried. Detect `.error` and emit a proper annotation.

scripts/e2e/drive-device.sh
```diff
@@ -0,0 +1,87 @@
...
+T0=$(date +%s)
+BOOT_RESP=$(curl -sS -m "$BOOT_CURL_TIMEOUT" -X POST "${BASE_URL}/boot-device" \
+  -H 'Content-Type: application/json' -d "$BOOT_JSON")
+T1=$(date +%s)
+echo "boot-device ($((T1 - T0))s): $BOOT_RESP"
+echo "::endgroup::"
+echo "$BOOT_RESP" | grep -q '"booted":true' || {
+  echo "::error::boot-device did not report booted:true"
+  exit 1
+}
```
hubgan, line 46: grep -q `'"booted":true'` / `'"tapped":true'` is serialization-brittle (assumes compact JSON). Works with Express res.json, but reusing the python3 it already invokes would be more robust/consistent.

.github/workflows/e2e-device-smoke.yml
```diff
@@ -0,0 +1,238 @@
...
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "packages/tool-server/src/tools/devices/**"
+      - "packages/tool-server/src/tools/gesture-tap/**"
+      - "packages/tool-server/src/tools/screenshot/**"
+      - "packages/tool-server/test/fixtures/electron-smoke-app/**"
+      - "scripts/e2e/drive-device.sh"
+      - ".github/workflows/e2e-device-smoke.yml"
```
hubgan, line 49: The pull_request path filter covers `tools/devices`, `gesture-tap`, `screenshot`, the fixture and the script — but not the transport layer (`blueprints/simulator-server.ts`, `utils/simulator-client.ts`, `blueprints/chromium-cdp.ts`). That is exactly where the #391 screenshot regression lived, so a future regression there wouldn't auto-trigger these smoke jobs. Considerwidening the filter.

scripts/e2e/drive-device.sh
```diff
@@ -0,0 +1,87 @@
...
+if [ "$SLEEP_BEFORE_SHOT" -gt 0 ]; then
+  echo "Waiting ${SLEEP_BEFORE_SHOT}s for the compositor to paint the first frame..."
+  sleep "$SLEEP_BEFORE_SHOT"
+fi
+
+echo "::group::screenshot"
+SHOT_JSON="$ARTIFACT_DIR/smoke-shot.json"
+curl -sS -m 60 -X POST "${BASE_URL}/screenshot" \
+  -H 'Content-Type: application/json' \
+  -d "{\"udid\":\"${DEVICE_ID}\"}" >"$SHOT_JSON"
```
hubgan: Single-shot screenshot, no retry. It leans entirely on the tool-level #391 fix — which the author explicitly flagged as a temporary patch to be reverted. A tiny retry loop around the screenshot would be cheap belt-and-suspenders for a one-shot smoke.

.github/workflows/e2e-device-smoke.yml
```diff
@@ -0,0 +1,244 @@
...
+          DEVICE_ID: ${{ env.UDID }}
+
+      - if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: e2e-ios-sim-macos
+          path: |
+            ${{ runner.temp }}/smoke-shot.png
+            ${{ runner.temp }}/smoke-shot.json
+            ~/.argent/tool-server.log
```
hubgan, line 103: `actions/upload-artifact` resolves path: with `@actions/glob`, which does not do tilde expansion, so ~/.argent/tool-server.log matches nothing and `if-no-files-found: warn` silently swallows it. The shots above work only because they use ${{ runner.temp }}. Net effect: on a failed job the one artifact you actually need is missing.

The same goes for line 185 and 242

.github/workflows/e2e-device-smoke.yml
```diff
@@ -0,0 +1,244 @@
...
+#     — qemu dies with "HVF error: HV_UNSUPPORTED" seconds into boot. The
+#     Android boot/screenshot/tap path is already regression-guarded on Linux
+#     via KVM (wayland-e2e.yml), so macOS adds no runnable coverage here.
+#   - Physical devices: real iOS (CoreDevice HID) and real Android (adb-USB)
+#     have no attached hardware on hosted runners.
+#   Both would need self-hosted runners (a Mac exposing HVF / wired-up devices).
+#
+# Manual + on-change trigger only — the macOS jobs are slow and billed.
+
+on:
```
hubgan, line 32: Add a concurrency group to cap billed macOS minutes. The header notes the macOS jobs are "slow and billed," but pull_request fires two of them on every push to a matching PR with no cancellation, so rapid pushes stack up.

.github/workflows/e2e-device-smoke.yml
```diff
@@ -0,0 +1,244 @@
...
+          # typecheck job already enforces types). Without it a cold start on a
+          # loaded CI runner can take >60s to bind and flakily times out.
+          nohup env ARGENT_PORT=3033 TS_NODE_TRANSPILE_ONLY=1 npx ts-node src/index.ts start \
+            > ~/.argent/tool-server.log 2>&1 &
+          for i in $(seq 1 60); do
+            curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:3033/tools | grep -q 200 \
+              && { echo "tool-server up t+${i}s"; exit 0; }
+            sleep 1
+          done
+          echo "tool-server failed to start"; cat ~/.argent/tool-server.log; exit 1
```
hubgan: This block is duplicated verbatim in all three jobs (here, 154, 210), as is the npm ci + rollup-install + tsc --build boilerplate. Worth extracting into a composite action (e.g. .github/actions/start-tool-server) so the cells can't drift out of sync the TS_NODE_TRANSPILE_ONLY / port / poll logic only needs to live in one place.

scripts/e2e/drive-device.sh
```diff
@@ -0,0 +1,113 @@
...
+# An all-zero / uniform framebuffer PNG-compresses to a few KB; a real painted
+# screen is reliably larger. Guards against the boot "succeeding" but the
+# device never actually rendering.
+test "$SZ" -gt "$MIN_SHOT_BYTES" || {
+  echo "::error::screenshot too small (${SZ}B) — likely a blank/all-zero framebuffer"
+  exit 1
+}
+
+echo "::group::gesture-tap"
+TAP_RESP=$(curl -sS -m 30 -X POST "${BASE_URL}/gesture-tap" \
```
hubgan: gesture-tap has no retry, unlike screenshot's 3x loop just above (line 59). Low risk because the screenshot succeeded immediately before, so the device is painted but if the same first-frame/transport race (#391) ever surfaces on the tap, it's an unguarded hard fail. Either wrap it in the same small retry, or add a one-line comment stating the tap relies on the preceding successful screenshot as its readiness gate.

scripts/e2e/drive-device.sh
latekvo: Fixed in c44f3a45 — the screenshot step now detects an `error` field (and unparseable responses) and surfaces a clean annotation instead of a raw Python `KeyError`.

scripts/e2e/drive-device.sh
latekvo: Fixed in c44f3a45 — both the `booted` and `tapped` checks now parse the JSON with `python3` instead of grepping the serialized string.

.github/workflows/e2e-device-smoke.yml
latekvo: Fixed in c44f3a45 — the `pull_request` path filter now also covers the transport layer: `blueprints/simulator-server.ts`, `blueprints/chromium-cdp.ts`, and `utils/simulator-client.ts`.

scripts/e2e/drive-device.sh
latekvo: Fixed in c44f3a45 — the screenshot is now wrapped in a 3× retry loop that tolerates a transient error envelope before the first frame is available.

.github/workflows/e2e-device-smoke.yml
latekvo: Fixed in 6ab5fe1c — added a workflow-level `concurrency` group keyed on `${{ github.workflow }}-${{ github.ref }}` with `cancel-in-progress: true`, so rapid pushes cancel superseded (billed macOS) runs.

.github/workflows/e2e-device-smoke.yml
latekvo: Fixed in 6ab5fe1c — extracted the install/build + tool-server-start block into a shared composite action (`.github/actions/start-tool-server`) that all three jobs now use, so the `TS_NODE_TRANSPILE_ONLY` / port / readiness-poll logic lives in exactly one place. The rollup native binary is selected by `$RUNNER_OS`.

scripts/e2e/drive-device.sh
latekvo: Fixed in 6ab5fe1c — gesture-tap now uses the same 3× retry as the screenshot, with robust JSON parsing so an error/non-JSON reply retries instead of raising a raw Python traceback.

.github/workflows/e2e-device-smoke.yml
latekvo: I checked this before changing it, and `upload-artifact@v7` does expand a leading `~/`. It bundles `@actions/glob@^0.6.1`, whose pattern normalizer rewrites a leading `~/` segment to `os.homedir()` — the branch that checks `pattern === '~' || pattern.startsWith('~' + path.sep)` in its bundled `dist/index.js`.

I confirmed it empirically against 0.6.1: `glob.create('~/.argent/tool-server.log').glob()` resolves the search path to `/Users/<me>/.argent/tool-server.log` and matches the file. So `~/.argent/tool-server.log` does get uploaded — leaving it as-is. (The `.png` / `.json` use `${{ runner.temp }}` only because that's where drive-device.sh writes them.)

latekvo:
Merging due to purely CI changes. Holding this PR open is somewhat inconvenient for #421 and others

PR #372, opened 2026-06-18
fix(skills): degrade argent rule gracefully without argent installed

hubgan:
## Problem

Teams want to commit argent's `.claude/` content (the always-on rule + skills +
agents) into their repos so teammates share the same workflow. Today that
**breaks the agent for anyone who clones the repo without argent installed**.

The always-on rule (`packages/skills/rules/argent.md`) unconditionally asserted
*"Argent MCP tools are available in this project"* and routed the agent into
argent skills and the `argent` CLI. For a teammate with no argent install:
- the committed `.mcp.json` server (`command: "argent"`) can't launch, so no
  `mcp__argent__*` tools appear;
- the `argent` CLI isn't on PATH;
- but the rule still names argent the *preferred* way to work → the agent
  confidently calls missing tools and a missing binary → confusing failures.

## Change

- **Soften the opening claim** from an unconditional assertion to conditional
  phrasing ("If argent is installed and configured… otherwise see
  `<availability_check>`").
- **Add an `<availability_check>` block** that runs **once per session** before
  the first argent tool call / `argent` command:
  1. primary signal — are `mcp__argent__*` tools in the tool list?
  2. secondary probe — `command -v argent`.
  - If present: follow the rule normally (no-op).
  - If absent: treat it as an expected state (not an error to retry), don't
    attempt any argent workflow, and surface a one-time install hint
    (`npx @swmansion/argent init`).

## Scope

- Only `packages/skills/rules/argent.md` (16 insertions, 1 deletion).
- Later steps (an `argent doctor` command for version skew and global-vs-local
  install conflicts) are tracked separately and **not** in this PR.

## Verification

- No-argent path: in a checkout with the committed `.claude/` but argent not on
  PATH, an argent-triggering task ("take a screenshot of the app") makes the
  agent recognize argent is absent, print the install hint, and skip
  `mcp__argent__*` calls.
- Regression check: with argent installed, the "present" path is a no-op and
  normal argent behavior is unchanged.

### DIFF
--- a/packages/skills/rules/argent.md
+++ b/packages/skills/rules/argent.md
@@ -32,7 +32,7 @@ If argent IS available, ignore the rest of this block and follow this rule norma
 
 If argent is ABSENT, treat it as an expected state, not an error to retry. Do not call `mcp__argent__*` tools, do not run `argent` commands, and do not attempt any argent workflow. Tell the user once, and ask if you should continue without argent:
 
-> Argent isn't installed in this environment. To enable the mobile/Chromium tooling this repo is configured for, run `npx @swmansion/argent init` (or `npm i -g @swmansion/argent && argent init`).
+> Argent isn't installed in this environment. To enable the mobile/Chromium tooling this repo is configured for, run `npx @swmansion/argent init -y` (or `npm i -g @swmansion/argent && argent init -y`).
 > </availability_check>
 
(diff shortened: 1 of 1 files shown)


filip131311 (approved):
love it thank you

PR #366, opened 2026-06-18
feat(vega): Amazon Vega (Fire TV) platform support

finloop:
## Summary

Adds **Vega (Amazon Fire TV)** as a third Argent platform alongside iOS and Android. Vega apps are React Native 0.72 + Hermes running on a QEMU **Virtual Device (VVD)**, driven by a **D-pad tv-remote rather than touch**. This PR extends Argent's platform abstraction so the existing tool surface gates correctly on `vega`, and adds the Vega-specific tools needed to actually drive a Fire TV app end to end.

### Platform abstraction

- `Platform` gains `vega`; `DeviceKind` gains `vvd` (the QEMU VVD); `ToolCapability` gains a `vega { vvd, device }` block; `ToolDependency` gains `vega`.
- `assertSupported()` selects the `vega` capability matrix for Vega devices; `dispatchByPlatform()` gains an optional `vega` branch, and a Vega dispatch with no branch throws `NotImplementedOnPlatformError` (501).
- New `vega-cli` util (modeled on `adb.ts`) resolves `vega`/`kepler` from `PATH` or `~/vega/bin` with the same hardened exec/quoting; `check-deps` resolves and hints the `vega` dependency.
- The `simulator-server` blueprint is never wired for `vega` (Vega uses `adb` + the `vega`/`kepler` CLI, not the proprietary binary).

### How Vega is wired

- **Input / screenshot / describe** go over `adb` (input via `inputd-cli`, screenshots host-side via the Android emulator console, element tree via the on-device automation toolkit).
- **App lifecycle** goes over the `vega`/`kepler` CLI.
- v1 is **Virtual-Device-only**: every Vega capability is declared `vvd`. Physical Fire TV is out of scope and untested — the tool-server does not connect to or detect hardware, so `resolveDevice` classifies every `amazon-` serial as a VVD by shape. A physical device is therefore not classified as `device` and not rejected at the capability gate; supporting and gating real hardware is deferred to a version where it can be tested.
- Vega runs **React Native 0.72**, which predates the new React Native DevTools. The CDP/Hermes debugger surface (`debugger-*`, React/native profilers) depends on integration points the new DevTools provides, so **debugging and profiling are out of scope for this PR** — Vega profiling/crash analysis is handled by the `amazon-devices-buildertools-mcp` server instead.

### Also includes

- `argent-vega` skill documenting the full Vega workflow (target discovery, the describe → compute-path → `tv-remote` navigation loop, text injection).
- Vega VVD end-to-end CI workflow.

[...]

PR #374, opened 2026-06-18
feat(tv): Apple TV (tvOS) and Android TV support

scianek:
Adds end-to-end support for driving **Apple TV (tvOS)** and **Android TV (leanback)** targets through argent, using a focus-driven interaction model (TVs have no touchscreen — interaction is moving a focus highlight with the remote).

This PR was reworked after review to **reuse the existing cross-platform tool surface and the `tv-remote` tool from the Vega/Fire TV work (#366)** rather than ship a separate family of `tv-*` tools.

## What this delivers

TV targets are driven by the **same cross-platform tools as every other platform**, routed through the existing `dispatchByPlatform` fork mechanism — there are no dedicated `tv-describe`/`tv-navigate`/`tv-type`/`tv-set-focus` tools:

- **`describe`** — on a `runtimeKind: "tv"` target, returns the focus-driven view (focused + focusable elements) instead of a tap tree. On Android TV it auto-falls-back to the full `uiautomator` tree when the RN focus engine exposes nothing.
- **`tv-remote`** — the cross-platform remote/D-pad tool from #366, now extended with `ios` (Apple TV HID daemon) and `android` (Android TV adb keyevents) branches alongside the existing `vega` branch. Single button, a `repeat`, or a whole path (`["up","right","select"]`) in one call.
- **`keyboard`** — types into the focused field on a TV target (named keys are rejected there — they're navigation, which belongs to `tv-remote`).
- **`button`** stays hardware-only (phones/tablets); it is not a TV tool.

**Apple TV** runs two native daemons (in-sim AX service + host-side HID daemon), shipped via the `argent-private` submodule. **Android TV** reuses the same tool surface, adb-backed (`input keyevent`, `uiautomator dump`, `input text`).

## Remote vocabulary parity

`tv-remote` exposes the full 16-button vocabulary (`up`/`down`/`left`/`right`/`select`/`back`/`home`/`menu`/`playPause`/`rewind`/`fastForward`/`next`/`previous`/`volumeUp`/`volumeDown`/`mute`):

| | Apple TV | Android TV | Vega |
|---|---|---|---|
| D-pad / select / back / menu / home / playPause | ✅ | ✅ | ✅ |
| media-transport + volume/mute | ❌ rejected | ✅ | ✅ |

Media-transport/volume keys **genuinely work on Android TV** (real keycodes — verified live: `volumeUp` moved `STREAM_MUSIC`, `mute` toggled per `dumpsys audio`). On the **Apple TV simulator** they are **rejected with a clear error**: on-device testing confirmed the tvOS sim's HID stack silently drops Consumer-Control events, so returning success would be a lie.

## Key design points
- **tvOS UDIDs are UUID-shaped and indistinguishable from iOS by shape.** Handled with `runtimeKind` detection (tvOS runtime string for Apple TV; `pm list features` leanback/television for Android TV — *not* `ro.build.characteristics`, which lies on TV emulators).
- Converted launch-app / restart-app / screenshot / screenshot-diff / run-sequence from **eager** to **lazy** service resolution, so a tvOS target never spins up (and hangs on) the iOS-only simulator-server / native-devtools blueprints.
- Native injection selects the platform-matched (TVOSSIMULATOR) dylib slice; daemons recycle across sim reboots and app relaunches so injection and focus survive.
- SKILL.md footprint minimized per review: the two TV skills were collapsed into a single lean **`argent-tv-interact`** (~40 lines).

## Verification
- `npm run build`, ESLint, Prettier, and the full tool-server suite (**1565 tests**) all pass; CI green (incl. the Apple TV / Android TV / Vega e2e jobs).
- Core flows verified live against real apps on both an Apple TV 4K simulator and a Google ATV emulator.

## Dependency / merge order
Pins the `argent-private` submodule to the tip of its `feat/tv-support` branch (software-mansion/argent-private#20). **Merge that PR to `main` first, then bump the submodule pointer here to the merged SHA before merging this PR.**


filip131311 (changes requested):
The tools like tv-describe should realy just be "describe" and use our platform fork mechanism: 

Image

latekvo:
Thanks for this PR! 

As discussed in person, please:
1. Remove all redundant tools (`tv-describe`, `tv-type`, etc)
2. Use the `tv-remote` that's already implemented in #366
3. Use our platform forking mechanism (this applies to both the redundant tools as well as the `tv-remote` tool)

I'd also say:

4. Minimize SKILL.md footprint - i believe it's excessive. Skills are an expensive resource

But that's up to discussion.

latekvo (commented):
Good job overall. Left some inlines

packages/tool-server/src/blueprints/tv-control.ts
```diff
@@ -0,0 +1,391 @@
...
+      }
+      axProc = spawnAxDaemon(udid, axSock);
+      axProc.on("exit", onAxExit);
+      await waitForSocket(axSock, axProc, 15_000);
+    }
+
+    // Respawn only if the daemon process actually exited (rare — it normally
+    // outlives the app). Stale-cache recovery goes through `recycleAx` instead.
+    async function ensureAxAlive(): Promise<void> {
+      if (!axExited || disposed) return;
```
latekvo: `ensureAxAlive` returns early whenever `axExited` is false, without awaiting an in-flight respawn. `recycleAx()` (which `describe` calls after a post-launch empty window) runs `spawnFreshAx`, and `spawnFreshAx` sets `axExited = false` synchronously and then unlinks the socket before the new daemon has bound it. A second ax-backed call (`describe` / `hierarchy` / `setFocus`) that arrives during that window passes this guard and then connects to a socket that has just been removed and not yet re-created, so it fails with a connection error. The service instance is shared per device, so two concurrent calls on the same Apple TV — one of them triggering a recycle — are enough to reach this.

packages/tool-server/src/blueprints/android-tv-control.ts
```diff
@@ -0,0 +1,393 @@
...
+        await pressKey(direction);
+      },
+
+      async type(text: string): Promise<void> {
+        // `input text` treats a literal space as an argument separator, so
+        // spaces must be sent as %s. Quoting the whole token keeps every other
+        // shell metacharacter inert (adb re-parses the command through the
+        // device shell). Empty input is a no-op.
+        if (text.length === 0) return;
+        const encoded = text.replace(/ /g, "%s");
```
latekvo: This encoding relies on `adb shell input text` decoding `%s` back to a space, which is not round-trip safe. If the user's text already contains the literal two-character sequence `%s`, `input text` turns it into a space on the device, so the value typed differs from the value requested — silently, with no error surfaced to the caller.

packages/tool-server/src/blueprints/android-tv-control.ts
```diff
@@ -0,0 +1,393 @@
...
+        await adbShell(serial, `input text ${shellQuote(encoded)}`, { timeoutMs: 15_000 });
+      },
+
+      async setFocus(label: string): Promise<{ ok: boolean; message: string }> {
```
latekvo: `setFocus`, together with the geometry-driven D-pad walk and its supporting helpers (`PixelRect` / `centre` / `labelMatches` / `MAX_FOCUS_STEPS`), has no caller anywhere in the tool surface — the `tv-set-focus` tool was removed, and only `describe` / `navigate` / `type` / `recycleAx` are reached. `ping` and `hierarchy` on `TvControlApi` are likewise never invoked. This leaves a substantial block of code that ships but cannot run in production and is exercised only by its own unit tests.

packages/tool-server/src/blueprints/tv-control-types.ts
```diff
@@ -0,0 +1,92 @@
...
+  | "playPause"
+  | "rewind"
+  | "fastForward"
+  | "next"
+  | "previous"
+  | "volumeUp"
+  | "volumeDown"
+  | "mute";
+
+export const TV_DIRECTIONS: readonly TvDirection[] = [
```
latekvo: `TV_DIRECTIONS` is exported but has no references anywhere in the codebase.

packages/tool-server/src/blueprints/tv-control-types.ts
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * Keeping the types here (rather than in `tv-control.ts`) lets the Android
+ * backend depend on the contract without importing the iOS-only daemon-binary
+ * resolvers from `@argent/native-devtools-ios`.
+ */
+
+export interface TvElement {
+  label?: string;
+  frame?: { x: number; y: number; width: number; height: number };
+  tapPoint?: { x: number; y: number };
```
latekvo: `TvElement.tapPoint` is never assigned or read on the TV path: the Android `toTvElement` and the tvOS `describe` response do not populate it, and the focus-view renderer does not read it. A focus-driven TV surface has no tap coordinates, so this field carries no information.

packages/tool-server/src/utils/react-profiler/pipeline/06-resolve/ast-index.ts
```diff
@@ -171,6 +148,63 @@ function isWrappedInMemo(source: string, componentName: string): boolean {
...
+/**
+ * Classify a `variable_declarator` value node as a (possibly HOC-wrapped)
+ * React component. Recognizes a bare arrow/function expression as well as
+ * components wrapped in `memo` / `forwardRef` (including the `React.`-qualified
+ * forms and nesting like `memo(forwardRef(() => …))`). Before this, every
+ * `export const X = React.memo(...)` / `forwardRef(...)` component was silently
+ * absent from the index, so `react-profiler-component-source` returned
+ * `found: false` for them despite the source existing.
+ */
+function classifyComponentValue(node: TreeSitterNode | undefined): {
```
latekvo: This memo / forwardRef HOC-unwrapping change fixes `react-profiler-component-source` component indexing and is unrelated to Apple TV / Android TV support. Bundling it into this PR widens the change surface beyond the stated feature.

latekvo:
Please update the PR description. It's very stale

packages/tool-server/src/utils/adb.ts
```diff
@@ -310,6 +317,78 @@ async function readAvdName(serial: string): Promise<string | null> {
...
+ * tool can surface a clear error rather than driving an offline target).
+ *
+ * `resolveDevice` classifies by serial shape alone and tags every Android
+ * target `platform: "android"`; code paths that must branch on Android TV
+ * (the tv-* tools) call this for the real form factor. Parallels
+ * `getSimulatorRuntimeKind` on the iOS side.
+ */
+export async function getAndroidRuntimeKind(serial: string): Promise<"mobile" | "tv" | undefined> {
+  const cached = androidRuntimeKindCache.get(serial);
+  if (cached) return cached;
```
latekvo: `androidRuntimeKindCache` is keyed by adb serial and is never invalidated — once a serial resolves to `tv` or `mobile`, this early return hands back that verdict for the rest of the tool-server process. Unlike a simulator UDID, an emulator serial like `emulator-5554` is a positional console slot that the next emulator reclaims after the current one stops. So within one long-lived tool-server, booting an Android TV AVD, exercising a TV-aware tool (which populates this cache), then shutting it down and booting a phone AVD on the same serial leaves the phone classified as `tv`: `describe` routes it to the focus view, `keyboard` to `adb input text`, and `launch-app` resolves a LEANBACK_LAUNCHER activity the phone app doesn't declare. The reverse — a real TV cached as `mobile`, after which the TV-only services reject it — is equally reachable. Recovery needs a tool-server restart. The iOS twin cache is keyed by a stable UDID, so it isn't exposed to this.

packages/tool-server/src/utils/adb.ts
```diff
@@ -310,6 +317,78 @@ async function readAvdName(serial: string): Promise<string | null> {
...
+  // physical TV hardware; absent on the ATV emulator (hence the feature-list
+  // primary above), but harmless to check when the feature list came back empty.
+  const characteristics = await adbShell(serial, "getprop ro.build.characteristics", {
+    timeoutMs: ENRICH_TIMEOUT_MS,
+  }).catch(() => "");
+  const isTv = characteristics
+    .split(",")
+    .map((c) => c.trim().toLowerCase())
+    .includes("tv");
+  return isTv ? "tv" : "mobile";
```
latekvo: Both probes above swallow errors (`.catch(() => "")`), and an empty result falls through to a definitive `"mobile"` here. A `pm list features` that returns empty because the device is mid-boot, under load, or hit the 5s enrichment timeout is therefore indistinguishable from a real phone. Since `getAndroidRuntimeKind` caches the first non-undefined verdict for a `device`-state serial and never recomputes it, a genuine Android TV that is briefly slow on its first TV-tool call gets pinned to `mobile` for the rest of the process — its `describe` / `keyboard` then route to the touch backend and the TV-only services reject it, with no self-heal short of a restart.

packages/tool-server/src/utils/adb.ts
```diff
@@ -330,14 +409,15 @@ export async function listAndroidDevices(): Promise<AndroidDevice[]> {
...
-      const [model, sdk, avd] = await Promise.all([
+      const [model, sdk, avd, runtimeKind] = await Promise.all([
         adbShell(d.serial, "getprop ro.product.model", { timeoutMs: ENRICH_TIMEOUT_MS }).catch(
           () => ""
         ),
         adbShell(d.serial, "getprop ro.build.version.sdk", { timeoutMs: ENRICH_TIMEOUT_MS }).catch(
           () => ""
         ),
         readAvdName(d.serial),
+        readRuntimeKind(d.serial),
```
latekvo: `readRuntimeKind` now runs for every `device`-state target during this enrichment. For a non-TV device the leanback feature is absent, so it always makes both the `pm list features` call and the `getprop ro.build.characteristics` fallback. `listAndroidDevices` sits on paths that never consume `runtimeKind`: `list-devices` (alwaysLoad) and `findSerialByAvdName`, which polls every 1.5s through a boot. Every ready phone there now pays two extra adb round-trips per call, and during boot — when `pm` may not yet be answering — each `readRuntimeKind` can block up to the 5s enrichment timeout, stretching individual poll iterations.

packages/tool-server/src/tools/tv-remote/platforms/focus-remote.ts
```diff
@@ -0,0 +1,47 @@
...
+        `${bad.join(", ")} ${bad.length === 1 ? "is" : "are"} not supported on the Apple TV ` +
+          `simulator (its HID stack ignores media-transport / volume keys) — use up/down/left/` +
+          `right/select/back/menu/home/playPause, or drive media via the app's on-screen controls`
+      );
+    }
+  }
+
+  const api = await resolveTvApi(registry, device.id);
+  const pressed: RemoteButton[] = [];
+  for (const button of buttons) {
```
latekvo, line 56: This loop presses each button in sequence without checking the abort signal the framework threads through (`InvokeToolOptions.signal`, available as the handler's 4th argument). With `tv-remote` declared `longRunning` and accepting a path of up to 64 buttons × `repeat` 50, and each Apple TV / Android TV press being a separate daemon/adb round-trip, a sequence can run for minutes. If the client cancels or disconnects, the loop keeps firing presses at the held device to completion. (The Vega branch is unaffected — it injects the whole path in one round-trip.)

packages/tool-server/src/blueprints/tv-control.ts
```diff
@@ -0,0 +1,353 @@
...
+          screenFrame: r.screenFrame,
+        };
+      },
+
+      async navigate(direction: TvDirection): Promise<void> {
+        await sendJson(hidSock, `navigate ${direction}`);
+      },
+
+      async type(text: string): Promise<void> {
+        await sendJson(hidSock, `type ${text}`);
```
latekvo: `type` sends the whole string in one socket command and uses `sendLine`'s default 10s timeout. The HID daemon types roughly one keypress every ~40ms and only writes its reply after the entire string has been typed, so a long input — on the order of a couple hundred characters — exceeds 10s. `sendLine` then rejects with a timeout and destroys the socket while the daemon is still typing, so `keyboard` reports a hard failure for text that was in fact (largely) entered. `navigate` shares the default, but a single press is well within budget.

packages/tool-server/src/tools/screenshot/index.ts
```diff
@@ -66,100 +67,95 @@ const capability: ToolCapability = {
...
+  return file;
 }
 
-interface ChromiumServices {
-  chromium: ChromiumCdpApi;
-}
+export function createScreenshotTool(registry: Registry): ToolDefinition<Params, Result> {
+  return {
+    id: "screenshot",
+    description: `Capture a screenshot of the device screen (iOS simulator, Android emulator, Apple TV simulator, or Chromium app). Returns { url, path }; the MCP adapter renders it as a visible image unless the caller passed includeImageInContext: false.
```
latekvo: This description lists `iOS simulator, Android emulator, Apple TV simulator, or Chromium app` but drops Vega, which the previous wording included and which the tool still supports — the `vega` capability and the `device.platform === "vega"` capture branch both remain below. An agent reading this would conclude screenshots aren't available on a Vega target.

packages/native-devtools-ios/scripts/build.sh
```diff
@@ -74,6 +78,19 @@ if [[ -n "${PREBUILT_NATIVE_DEVTOOLS_IOS:-}" ]] && [[ -n "${PREBUILT_KEYBOARD_PA
...
+    && -n "${PREBUILT_NATIVE_DEVTOOLS_IOS_TVOS:-}" \
+    && -n "${PREBUILT_KEYBOARD_PATCH_TVOS:-}" \
+    && -n "${PREBUILT_INJECTION_BOOTSTRAP_TVOS:-}" ]]; then
+    echo "Using pre-built tvOS dylibs"
+    mkdir -p "${ROOT_DIR}/dylibs/tvos"
+    cp "$PREBUILT_NATIVE_DEVTOOLS_IOS_TVOS" "${ROOT_DIR}/dylibs/tvos/libNativeDevtoolsIos.dylib"
+    cp "$PREBUILT_KEYBOARD_PATCH_TVOS" "${ROOT_DIR}/dylibs/tvos/libKeyboardPatch.dylib"
+    cp "$PREBUILT_INJECTION_BOOTSTRAP_TVOS" "${ROOT_DIR}/dylibs/tvos/libArgentInjectionBootstrap.dylib"
+  fi
   exit 0
```
latekvo, line 112: This `exit 0` ends the prebuilt path (the documented non-macOS CI route) after copying the iOS and tvOS dylibs, but before the section that consults `PREBUILT_TVOS_AX_SERVICE` / `PREBUILT_TVOS_HID_DAEMON` (lines 286/300). Those two variables are advertised in the header (line 22) as prebuilt inputs, yet on the prebuilt path they're never read — so a build that supplies them still produces a tree with no `tvos-ax-service` / `tvos-hid-daemon`.

packages/tool-server/src/tools/describe/platforms/tv.ts
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  // Still empty after the transition window: on Apple TV the daemon may be
+  // holding a stale primaryApp cache from a killed app. Recycle it once and
+  // re-probe — a fresh daemon rebinds to the current foreground app, recovering
+  // a fully-rendered screen the stale cache reported as empty. (No-op on
+  // Android TV.)
+  if (isEmpty(res)) {
+    await api.recycleAx();
+    res = await api.describe();
```
latekvo, line 143: On an Android TV target `recycleAx()` is a documented no-op, so this re-probe repeats the exact uiautomator dump the retry loop just confirmed was empty — it cannot return anything different. When it is still empty, the Android branch below then calls `describeAndroid`, which dumps the hierarchy a third time. On the empty-focus path (a headline case for RN-focus-engine screens) a single `describe` can issue around five uiautomator dumps plus the retry sleeps.

packages/tool-server/src/blueprints/tv-control.ts
```diff
@@ -0,0 +1,353 @@
...
+      if (!axExited) return;
+      axRespawn = spawnFreshAx().finally(() => {
+        axRespawn = null;
+      });
+      await axRespawn;
+    }
+
+    // Force a fresh daemon regardless of liveness — drops a stale primaryApp
+    // cache. Coalesces concurrent callers onto a single respawn.
+    async function recycleAx(): Promise<void> {
```
latekvo, line 410: The `recycleAx` / `spawnFreshAx` / `ensureAxAlive` coalescing here (the single-flight `axRespawn` guard, the unlink-and-rebind, and the exit-during-respawn handling) is the most concurrency-sensitive logic in the change, but no test exercises this Apple TV blueprint — `describe-tv` and `tv-remote` only import the `TvControlApi` type and inject a mock, and nothing imports `tvControlBlueprint` (the Android backend's factory, by contrast, is exercised directly). A regression in the respawn coalescing or daemon cleanup would pass the suite.

latekvo (commented):
Reviewed the TV-support changes across the tool-server: the `describe` / `tv-remote` / `keyboard` routing, the two focus-driven backends, the eager→lazy service-resolution refactor, the native-binary packaging chain, and the new tests. Specific observations are left inline on the relevant lines.

packages/tool-server/src/utils/adb.ts
```diff
@@ -311,12 +318,150 @@ async function readAvdName(serial: string): Promise<string | null> {
...
+  const isTv = characteristics
+    .split(",")
+    .map((c) => c.trim().toLowerCase())
+    .includes("tv");
+  if (isTv) return "tv";
+
+  // Determinate "mobile" only when the characteristics probe actually answered.
+  // Both probes empty means we never got a reading — report indeterminate rather
+  // than a false "mobile".
+  return characteristics.trim() ? "mobile" : undefined;
```
latekvo: `readRuntimeKind` only returns `undefined` (indeterminate, uncached) when *both* `pm list features` and `ro.build.characteristics` come back empty. But on the Google ATV emulator `ro.build.characteristics` is `emulator` — non-empty, and without a `tv` token (the comment a few lines up and the discovery tests rely on exactly this). So in the boot / PMS-not-ready window, where `pm list features` is empty but `getprop` already answers `emulator`, this line returns a *definitive* `"mobile"`, which `resolveRuntimeKindCached` then caches keyed by the (boot-stable) serial+avdName.

Every later call returns the cached `"mobile"` even once PackageManager is up and would report leanback — the verdict is poisoned for the process lifetime (cleared only when the serial leaves the device list). So the first TV-tool call right after booting a real Android TV pins it to mobile: `describe`/`keyboard` route it to the touch backend, and the TV factory / `tv-remote` reject it as 'a mobile device', recoverable only by restarting the tool-server. This is the precise mislabel-then-cache the surrounding comment says it prevents; the protection only holds when characteristics is *also* empty, which never happens on an emulator.

Verified against the real `getAndroidRuntimeKind`: with `pm list features`='' and characteristics='emulator' it returns `"mobile"`, and it stays `"mobile"` on the next call after `pm` recovers to leanback.

packages/tool-server/src/utils/adb.ts
```diff
@@ -311,12 +318,150 @@ async function readAvdName(serial: string): Promise<string | null> {
...
+ * `resolveDevice` classifies by serial shape alone and tags every Android
+ * target `platform: "android"`; code paths that must branch on Android TV
+ * (the tv-* tools) call this for the real form factor. Parallels
+ * `getSimulatorRuntimeKind` on the iOS side.
+ *
+ * The readiness check goes through the cheap `adb devices` listing (no getprops)
+ * rather than the full `listAndroidDevices` enrichment; only the avdName getprop
+ * and — on a cache miss or a slot whose AVD changed — the feature probe run.
+ */
+export async function getAndroidRuntimeKind(serial: string): Promise<"mobile" | "tv" | undefined> {
```
latekvo, line 477: `getAndroidRuntimeKind` runs `listAndroidSerials()` (an `adb devices`) and `await readAvdName(serial)` (a `getprop`, two on a physical phone) on *every* call before it consults the cache, because `avdName` is part of the cache key — only the `pm list features` / characteristics probe is memoized. The per-interaction callers that were added in this PR — `describe` (platforms dispatch), `keyboard/platforms/android`, `launch-app/platforms/android`, and `restart-app/platforms/android` — therefore each pay ~2-3 net-new adb round-trips on every Android invocation, including the common phone case, which on `main` went straight to the uiautomator dump / simulator-server. The comment that this memoizes 'to keep the hot describe/navigate path off the … round-trips' only covers the feature probe; the readiness listing and avd-name getprop are not cached.

Verified: a second (cache-hit) `getAndroidRuntimeKind` call still issues `adb devices` and the avd_name getprop, skipping only `pm list features`.

packages/tool-server/src/blueprints/tv-control.ts
```diff
@@ -0,0 +1,472 @@
...
+      },
+
+      async navigate(direction: TvDirection): Promise<void> {
+        await sendJson(hidSock, `navigate ${direction}`);
+      },
+
+      async type(text: string): Promise<void> {
+        // Scale the socket timeout to the input length (see typeTimeoutMs) so a
+        // long string doesn't time out mid-type and report a false failure.
+        await sendJson(hidSock, `type ${text}`, typeTimeoutMs(text.length));
```
latekvo, line 471: `type` sends the whole string as one daemon command (`type ${text}`), and `sendLine` writes it as a single newline-terminated line. If `text` contains a `\n` (the `keyboard` tool passes `text` through unsanitized), the wire payload becomes two lines — e.g. `type foo\nbar\n`. The tvOS HID daemon is a one-line-per-connection server (read one line → reply → close, per the comment at the top of this file), so it reads only `type foo`, types `foo`, and the remainder (`bar`) is silently dropped on the device. `type` discards the reply and resolves cleanly, so `keyboard` reports the full string as typed. Any newline in typed text truncates the input at the first line while reporting success.

Verified against the real blueprint: `api.type('foo\nbar')` writes exactly `type foo\nbar\n` to the hid socket in a single write.

packages/tool-server/src/blueprints/tv-control.ts
```diff
@@ -0,0 +1,472 @@
...
+    // This also yields a clear error when someone passes an iPhone udid.
+    const sims = await listIosSimulators();
+    const match = sims.find((s) => s.udid === udid);
+    if (!match) {
+      throw new Error(
+        `${TV_CONTROL_NAMESPACE}: no available simulator with udid '${udid}'. Run list-devices to find a booted Apple TV.`
+      );
+    }
+    if (match.runtimeKind !== "tv") {
+      throw new Error(
```
latekvo: This wrong-target rejection (and its twin in `android-tv-control.ts`) throws a plain `Error`. `tv-remote` has no pre-dispatch TV-kind check — a regular iPhone simulator or a phone/tablet emulator passes the shape-based capability gate and reaches this factory, so pointing `tv-remote` at a phone throws here. The plain `Error` is wrapped as `ServiceInitializationError` → `ToolExecutionError`, and because it is neither `UnsupportedOperationError` nor `NotImplementedOnPlatformError`, `http.ts` maps it to **500**. The media/volume-key rejection in `focus-remote.ts`, by contrast, throws `UnsupportedOperationError` → **400**. So the same wrong-target call surfaces as a 500 (which reads to an agent as a transient server fault and invites retries) where the capability contract treats a wrong device class as a 400.

packages/tool-server/src/tools/describe/platforms/tv.ts
```diff
@@ -0,0 +1,170 @@
...
+ * agent uses one `describe` for every target and never has to know up front
+ * whether a UDID is a phone or an Apple TV.
+ */
+export async function describeTv(registry: Registry, device: DeviceInfo): Promise<DescribeResult> {
+  const api: TvControlApi = await resolveTvApi(registry, device.id);
+
+  // Ride out a brief post-launch / transition window where the focus engine
+  // hasn't populated yet (see EMPTY_RETRY_* / EMPTY_HINT).
+  let res = await api.describe();
+  for (let attempt = 1; attempt < EMPTY_RETRY_ATTEMPTS && isEmpty(res); attempt++) {
```
latekvo: On an Android TV running react-native-tvos — the headline case this file documents, where focus is RN-engine-managed and invisible to the OS accessibility tree — the focus set is empty in *steady state*, not just during a post-launch transition. This empty-retry loop still runs unconditionally: the initial `api.describe()` plus two retries are three full `uiautomator dump`s (each ~0.5-2s on a real device) interleaved with two 600ms sleeps, all returning empty and discarded, before the Android fallback issues a fourth dump that actually produces the rendering. The loop's premise ('ride out a brief transition window') does not hold for these screens, so every `describe` of one pays ~3 wasted dumps + ~1.2s, repeated on each navigate→describe cycle the agent runs.

The project's own `describe-tv.test.ts` encodes this: on the empty Android path it asserts `describeFn` is called 3 times before the `describeAndroid` fallback fires.

.github/workflows/publish.yml
```diff
@@ -81,6 +81,13 @@ jobs:
       - run: test -f packages/argent/dylibs/libNativeDevtoolsIos.dylib
       - run: test -f packages/argent/dylibs/libKeyboardPatch.dylib
       - run: test -f packages/argent/dylibs/libArgentInjectionBootstrap.dylib
+      # tvOS (Apple TV) binaries — fail the publish if any is missing rather
+      # than silently shipping a package where tv-* tools crash at runtime.
+      - run: test -x packages/argent/bin/darwin/tvos-ax-service
```
latekvo, line 86: These tvOS verification gates were added here and to `build-package-artifact.yml`, but not to `publish-next.yml`. That workflow runs the same `download-native-binaries.sh` (which deliberately skips and exits 0 when a release tag lacks the Apple TV assets) and the same `bundle-tools.cjs` (where the four tvOS assets are `required: false`, so a missing source only warns), then runs `npm publish --tag next`. With no `test -x bin/darwin/tvos-*` / `test -f dylibs/tvos/*` gate on that path, the `next` channel can publish a package missing the tvOS daemons/dylibs, after which any Apple TV tool throws 'tvos-ax-service binary not found' at runtime — exactly the failure these gates exist to prevent on the stable channel. (`grep -c tvos publish-next.yml` = 0; publish.yml and build-package-artifact.yml = 6 each.)

latekvo (commented):
Re-review of the latest round of fixes. Every prior thread except the Android runtime-kind one (left open with a note) is addressed; resolving those separately.

This pass surfaced a set of new issues, posted inline below — the most important is the Android runtime-kind detection, where the empty-`pm-list-features` fix doesn't actually cover the Google ATV emulator (it still resolves to a cached 'mobile'). Each finding was reproduced before posting.

packages/tool-server/src/tools/screenshot/index.ts
```diff
@@ -66,100 +69,95 @@ const capability: ToolCapability = {
...
+  const file = path.join(
+    os.tmpdir(),
+    `argent-tv-screenshot-${udid.slice(0, 8)}-${process.hrtime.bigint()}.png`
+  );
+  await execFileAsync("xcrun", ["simctl", "io", udid, "screenshot", file], { signal });
+  // tvOS captures are 4K (3840x2160). Downscale in place unless full-res was
+  // explicitly requested, mirroring the default 0.3 scale the iOS/Android path
+  // applies server-side.
+  if (scale < 1.0) {
+    await execFileAsync("sips", ["-Z", String(Math.round(3840 * scale)), file], { signal }).catch(
```
latekvo: `sips -Z round(3840 * scale)` assumes the capture is always 3840px wide, but the non-4K 'Apple TV' simulator (a standard Xcode device type) captures at 1920x1080. `-Z` caps the longest *actual* dimension to the target, so for a 1920-wide capture at the default scale 0.3 the target is 1152 → an effective scale of 0.60, twice the requested scale and twice what the iOS/Android/Vega paths (which scale against the real source dimensions) produce.

Verified on a booted non-4K Apple TV sim: native 1920×1080 capture → tool output 1152px wide (0.60×) instead of the intended 576px.

packages/tool-server/src/tools/tv-remote/platforms/focus-remote.ts
```diff
@@ -0,0 +1,58 @@
...
+  params: TvRemoteParams,
+  unsupported?: ReadonlySet<RemoteButton>,
+  options?: InvokeToolOptions
+): Promise<TvRemoteResult> {
+  const buttons = expandButtons(params.button, params.repeat);
+
+  if (unsupported?.size) {
+    const bad = [...new Set(buttons.filter((b) => unsupported.has(b)))];
+    if (bad.length) {
+      throw new UnsupportedOperationError(
```
latekvo, line 36: This rejection runs before `resolveTvApi`, so the device's runtime kind has not been verified when the message is produced. A regular iPhone simulator (classified as `ios`/simulator by UDID shape, passing the capability gate) called with a media/volume button throws 'not supported on the Apple TV simulator (its HID stack ignores …)' — asserting the target is an Apple TV when it is an iPhone. The same non-TV device thus returns two different and both-misleading errors depending on the button: a media key → this 'Apple TV simulator' message, a D-pad key → the factory's 'not a tvOS simulator' (a 500).

packages/tool-server/src/blueprints/tv-control.ts
```diff
@@ -0,0 +1,472 @@
...
+  timeoutMs: number
+): Promise<void> {
+  const deadline = Date.now() + timeoutMs;
+  let exited: number | null | undefined;
+  // Track exit only for the duration of this wait. Remove the listener on every
+  // exit path (success or timeout) — otherwise it stays attached to a daemon
+  // that lives ~1h, holding a closure over this resolved call, and a fresh one
+  // accumulates per respawn.
+  const onExit = (code: number | null) => (exited = code);
+  proc.once("exit", onExit);
```
latekvo, line 170: `waitForSocket` (and the factory's `onAxExit`/`onHidExit`) only listen for the child's `exit` event. A daemon that fails to *spawn* — a missing or non-executable `tvosHidDaemonBinaryPath()` / `xcrun`, which the missing `publish-next.yml` tvOS gate makes more likely — fires `error`, not `exit`. So the spawn failure goes undetected and this loop polls `existsSync` until the 15s deadline, then throws the generic 'Timed out waiting for tv daemon socket' rather than failing fast with the real cause. The sibling `ax-service.ts` listens for both `exit` and `error`.

Verified: `execFile` on a nonexistent binary emits `error` and never emits `exit`.

packages/tool-server/src/blueprints/tv-control.ts
latekvo: Fixed in 1aeff914 (hardened in fcaa7c5b). `ensureAxAlive` now awaits an in-flight `axRespawn` before the liveness check, and `sendAx` retries against the rebound socket; the new concurrency cases in `tv-control-blueprint.test.ts` cover this exact race.

packages/tool-server/src/blueprints/android-tv-control.ts
latekvo: Fixed in 8e36fa52. `type` no longer emits `%s`: spaces are sent as `KEYCODE_SPACE` and each run is split after every `%`, so a literal `%s` can never sit inside a single `input text` call — it round-trips verbatim now.

packages/tool-server/src/blueprints/android-tv-control.ts
latekvo: Fixed in ddab58d9. The dead `setFocus` / geometry-walk / `ping` / `hierarchy` code and their helpers were removed from both backends and from `TvControlApi`.

packages/tool-server/src/blueprints/tv-control-types.ts
latekvo: Fixed in ddab58d9. `TV_DIRECTIONS` was removed.

packages/tool-server/src/blueprints/tv-control-types.ts
latekvo: Fixed in ddab58d9. `TvElement.tapPoint` was removed.

packages/tool-server/src/utils/react-profiler/pipeline/06-resolve/ast-index.ts
latekvo: Fixed in 74c629fc. The memo/forwardRef AST-indexing change was reverted out of this PR (it now lives on `main` independently).

packages/tool-server/src/utils/adb.ts
latekvo: Fixed in 13891dd3. The runtime-kind cache is now keyed by serial+avdName and dropped when the device goes offline, so a reclaimed emulator slot re-probes instead of inheriting the previous AVD's verdict.

packages/tool-server/src/utils/adb.ts
latekvo: Fixed in 13891dd3. `listAndroidDevices` resolves `runtimeKind` only when `opts.runtimeKind` is set; `list-devices` is the sole opt-in caller and the boot-loop pollers stay on the cheap `listAndroidSerials`.

packages/tool-server/src/tools/tv-remote/platforms/focus-remote.ts
latekvo: Fixed in fcaa7c5b. The press loop now calls `options.signal?.throwIfAborted()` between presses (the signal is threaded through `dispatchByPlatform` → the platform impls → `pressFocusRemote`).

packages/tool-server/src/blueprints/tv-control.ts
latekvo: Fixed in fcaa7c5b. `type` scales the socket timeout via `typeTimeoutMs(len)` (10s base + 60ms/char), so a long string no longer times out mid-type.

packages/tool-server/src/tools/screenshot/index.ts
latekvo: Fixed in 976f04d4. The screenshot description lists Vega again.

packages/native-devtools-ios/scripts/build.sh
latekvo: Fixed in 976f04d4. The prebuilt path now copies `PREBUILT_TVOS_AX_SERVICE` / `PREBUILT_TVOS_HID_DAEMON` into `bin/` before the `exit 0`.

packages/tool-server/src/tools/describe/platforms/tv.ts
latekvo: Fixed in 13891dd3. The recycle + re-probe is now skipped on Android (`device.platform !== "android"`), so the empty-focus path goes straight to the uiautomator fallback instead of repeating the no-op dump. (The retry-loop dumps that still run on every empty Android describe are raised separately in the new review.)

packages/tool-server/src/blueprints/tv-control.ts
latekvo: Fixed in fcaa7c5b. `tv-control-blueprint.test.ts` now drives `tvControlBlueprint` directly — the respawn coalescing, the concurrent-describe-await-respawn race, the post-`ensureAxAlive` unlink retry, the dispose/respawn race, and factory-failure cleanup are all covered.

packages/tool-server/src/utils/adb.ts
latekvo: Unresolved as of eb74580d5b. The fix returns `undefined` (indeterminate, uncached) only when *both* `pm list features` and `ro.build.characteristics` are empty. On the Google ATV emulator `ro.build.characteristics` is `emulator` (non-empty, no `tv` token), so in the boot / PMS-not-ready window an empty `pm list features` still falls through to a definitive, cached `"mobile"` here — the exact mislabel-then-pin this raised, for the primary target. Reproduced and detailed in the new review (the `readRuntimeKind` inline comment).

packages/tool-server/src/tools/describe/platforms/ios/index.ts
```diff
@@ -35,6 +49,14 @@ export async function describeIos(
   device: DeviceInfo,
   params: DescribeIosParams
 ): Promise<DescribeTreeData> {
+  // tvOS short-circuit: the focus-engine accessibility tree is served by the
+  // tv-control daemons, not the iOS ax-service. Without this, describe would
+  // try to spawn ax-service inside the Apple TV sim, time out on the daemon
+  // connection, and degrade with the wrong (boot-device) hint.
+  if (await isTvOsSimulator(device.id)) {
```
latekvo: `describeIos` now runs `await isTvOsSimulator(device.id)` as its first step on every call. `isTvOsSimulator` → `getSimulatorRuntimeKind` → `listIosSimulators()` shells out to `xcrun simctl list devices --json`, and `getSimulatorRuntimeKind` caches only *successful* (device-found) lookups. `describeIos` is a hot path: the `describe` tool calls it for every regular iPhone, `await-ui-element` calls it on every poll iteration, and (per the comment here) the preview / match-element-frame internal callers invoke it directly.

Two consequences, both reproduced on macOS:

1. The `await-ui-element` suite — untouched by this PR — goes red: 6 tests fail with 'tree fetch did not complete within the 40/120ms wait budget'. They use a fake UDID that never matches a real sim, so the lookup is never cached and every poll re-runs the xcrun probe, blowing the per-fetch budget. Disabling only this `isTvOsSimulator` block makes the suite green again (37/37, and the file runs in 1.3s instead of 7.3s). The PR's 'full suite passes' claim holds only where xcrun is absent (Linux CI), where the probe fails fast.

2. In normal use, the first `describe` / `await-ui-element` / preview call for any iOS UDID now pays an extra `xcrun simctl list devices --json` (measured ~500ms on this machine) before any tree is fetched; only after a successful lookup is it cached.

Verified: `isTvOsSimulator(<fake-udid>)` took 493ms here, and the 6 await-ui-element failures disappear when only this probe is removed.

latekvo (commented):
One more finding from the same pass — a regression in the iOS describe hot path that also red-lights the await-ui-element suite on macOS. Reproduced before posting.

packages/skills/rules/argent.md
```diff
@@ -115,7 +115,11 @@ Prompt keywords: vega, fire tv, vvd, virtual device, d-pad
...
-When: Performing touch interactions, typing, pressing hardware buttons, launching/restarting apps, opening URLs, rotating device, taking standalone screenshots, or verifying a visible UI code change.
+When: Performing touch interactions, typing, pressing hardware buttons, launching/restarting apps, opening URLs, rotating device, taking standalone screenshots, or verifying a visible UI code change. Phone/tablet iOS and Android only — for a TV target (`runtimeKind: "tv"`, Apple TV or Android TV) use `argent-tv-interact` instead.
+
+TV (APPLE TV / ANDROID TV) INTERACTION
+Skill: `argent-tv-interact`
+When: Any task involving a TV target (a `list-devices` entry with `runtimeKind: "tv"` — an Apple TV simulator or an Android TV emulator) — booting the target, moving focus with the remote / D-pad, activating elements, typing into search fields, or reading TV focus state. A TV UI is focus-driven, not touch-driven: the `gesture-*` tools do NOT apply. Use `describe` to read focus, `tv-remote` for remote presses (up/down/left/right/select/back/menu/home), and `keyboard` to type (the same tools drive both platforms). For Amazon Fire TV (Vega) use `argent-vega` instead.
```
latekvo: 1. Your section contains redundant info about vega that is already contained few lines above
2. These two sections - vega, android tv and apple tv can be squashed into a single brief paragraph.

packages/skills/skills/argent-tv-interact/SKILL.md
```diff
@@ -0,0 +1,43 @@
...
+---
+
+# Argent TV (Apple TV + Android TV)
+
+## Critical
+
+- A TV is **focus-driven, not touch-driven.** Drive every interaction with `tv-remote`; never use `gesture-*` / coordinate taps — they don't apply.
+- **Always `describe` before navigating** to find the live focus (the `→`-marked / focused element) and your target. Never guess focus from a screenshot.
+- The same tools drive both platforms; dispatch is automatic from the id. Pass an Apple TV simulator UDID or an Android TV `serial` (from `list-devices`) as `udid`.
+- For Amazon Fire TV (Vega) see `argent-vega` instead — same `tv-remote` / `describe` model, different platform.
```
latekvo: I'm pretty sure we can remove `argent-vega` skill, or integrate it into `argent-tv-interact`, since `argent-tv-interact` already covers all TV platforms **including** vega.

latekvo (commented):
Minor request regarding skills. I think it's otherwise pretty much ready for merging.

Thank you a lot for taking your time to implement this feature ❤️

latekvo (approved):
Thanks, this is ready for merge :shipit:

We have a large mess with redundancy and inconsistency with how `platform`, `kind` and `runtimeKind` are used - but it's a matter for separate PR, it's not a blocker :)

filip131311 (approved):

PR #375, opened 2026-06-19
feat(datagen): schema-grounded synthetic training data for an Argent-native model

latekvo:
## What

`datagen/` — a deterministic pipeline that generates **high-quality, high-volume tool-use trajectories** for fine-tuning a model to drive the Argent toolkit (iOS sim / Android emu / Chromium). The primary deliverable is the **data**; a ~$500 4B proof model comes next, then the $50k run.

## Why not just prompt an LLM to write the conversations

That hallucinates tool schemas (the skills even reference a `paste` tool that isn't in the registry), invents ungrounded tap coordinates (the one thing Argent forbids), and approximates the observation formats. Training on that teaches the model to guess.

## Approach: simulate Argent, roll out the expert, gate hard

- **Ground truth** — all 67 tool schemas extracted from the live registry (`spec/tools.json`); real output serializers (`describe`, component-tree, network logs, profiler reports) ported in `format.ts`.
- **Gym** (`gym.ts` + `archetypes/`) — a text-simulator of devices + apps + screen graphs; each tool has a real transition function returning the exact observation.
- **Expert policy** (`expert.ts`) — the `argent.md` rules as executable behavior: list-devices first, discovery-before-tap, tap the centre of a *discovered* element, re-discover after navigation, component-tree for RN, `run-sequence` batching, profiling/flow/network procedures, and **recovery from injected failures** (tap miss, describe error, boot timeout, debugger drop).
- **Hard gates** (`validate.ts`) — schema (strict, unknown args rejected) + structure + device-order + policy + **coordinate grounding** (every tap must resolve to an element in the most recent discovery output). The grounding gate makes "non-hallucinated" a *checkable property* and also enforces discovery-before-tap.
- **NL layer** (`narrate.ts`) — varies only *phrasing*, never the grounded action/observation backbone (the safe place for later LLM paraphrase).

## Proof it works

- **Validator self-check** (`node src/selfcheck.ts`): corrupts a known-good trajectory 9 ways (ungrounded tap, missing discovery, schema break, unknown arg, out-of-range coord, wrong device order, unknown tool, un-offered tool) and asserts each is rejected — **9/9**. A gate that accepts everything is worthless; this keeps it honest.
- **Pilot** (`node src/generate.ts --n 800 --evalN 100`): **800 train + 100 eval, 0 rejected**, 40/67 tools (60%) exercised across 16 task families and iOS/Android/Chromium, ~9 tool calls/trajectory, ~10% with recovery. Deterministic (seeded) → byte-identical reruns; train/eval seed-disjoint.
- **Speed**: ~1,300 validated trajectories/sec single-threaded → ~1M in ~13 min, embarrassingly parallel beyond.
- OpenAI + ShareGPT exporters; committed `samples/` shows real output without running.

## Review pointers

- `datagen/DESIGN.md` — full rationale + path to the proof model and scale-up.
- `datagen/README.md` — quickstart, gates, extension pattern.
- `datagen/samples/` — rendered trajectories + pilot `stats.json`.

## Not included / next

- 27 niche tools still uncovered (listed in `stats.json#tools_never_used`) — extension is mechanical (adding the last 5 families took ~150 lines).
- Screenshots are textual placeholders (text SFT; reinforces "read structure, not pixels"). A vision variant would layer real images.
- No model trained yet — this PR is the data pipeline + validated pilot.

Runs on Node 24+ with no build step (native TS type-stripping). Depends only on `ajv`.


latekvo:
Uuuh 😶‍🌫️

PR #378, opened 2026-06-19
fix(native-profiler): remove shell interpolation from iOS profiler (device_id + trace export)

latekvo:
## Summary
The iOS native profiler reached `/bin/sh` by string-interpolating values into shell-executed commands — the classic command-injection shape. This PR removes the shell from **every** such call site in the iOS-profiler subsystem, converting them to argv-array `execFileSync`/`execFile` so no caller- or path-derived value is ever re-parsed by a shell.

### 1. `device_id` interpolation — gated behind the UDID-shape classifier (`native-profiler/platforms/ios.ts`)
The profiler interpolated the caller-supplied `device_id` (`udid`) directly into shell-interpreted `execSync` template strings:
- `execSync(`xcrun simctl spawn ${udid} launchctl list`)`
- `execSync(`xcrun simctl listapps ${udid} | plutil -convert json -o - -`)`

`device_id` is declared `z.string()` with no format constraint and `resolveDevice()` returns it unchanged. **However, this sink is not directly reachable with a hostile `device_id`** (thanks @hubgan for the catch): before the iOS platform module runs, `classifyDevice()` (`utils/device-info.ts`) routes by *shape* against an anchored hex-UDID regex `^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}$`. Any `device_id` carrying a shell metacharacter fails that (anchored, hex-only) match, classifies as **Android**, and takes the argv-based `adb -s <serial> shell …` path — it never reaches this iOS function. Conversely, any value that *does* classify as iOS is, by the regex, metacharacter-free. So on `main` the original `execSync` was **dormant / defense-in-depth**, gated behind the classifier invariant, not an exploitable path from a tool parameter.

The conversion is still worthwhile: it removes the code's reliance on that classifier invariant (if the shape regex were ever loosened, or the sink reached by a future refactor with a synthetic `platform: "ios"`, the shell would be back). Fixed by converting both call sites to `execFileSync` arg-arrays — the `simctl listapps | plutil` pipe is split into two in-process stages so nothing is shell-interpreted.

This anti-pattern previously had a fix on an old branch that was lost when the code moved from `tools/profiler/ios-profiler/ios-profiler-start.ts` to `native-profiler/platforms/ios.ts`.

### 2. Trace-export hardening — same anti-pattern (`utils/ios-profiler/export.ts`)
Six `xctrace export` / `xctrace version` commands were built as `/bin/sh` strings via `execSync`, interpolating the trace-file and output paths:
- `execSync(`xctrace export --input "${traceFile}" --output "${outPath}" --xpath '...'`)` (×5, incl. the CPU/leaks/hangs exports and `--toc` discovery)
- `execSync("xctrace version 2>&1 || true")`

`traceFile` is internally generated today (a timestamped path under the debug dir, only ever passed in by `native-profiler/platforms/ios.ts`), so this is **not currently reachable from a tool parameter** — but it is the identical anti-pattern sitting in the same subsystem, and #1's interpolation was likewise dormant. Converted the shared `execSyncWithTimeout` helper to an argv-based `execFileAsyncWithTimeout` (`execFile`, no shell) and pass every `xctrace` invocation as a discrete argv array.

### 3. `xcodebuild` version probe — same shell-string shape (`capture-strategy/select.ts`)
`readActiveXcodeVersion()` ran `execSync("xcodebuild -version")`. The command is a constant with no interpolated input, so this is **not an injection** — but it is the same `/bin/sh`-string shape this PR removes everywhere else in the subsystem. Converted to `execFileSync("xcodebuild", ["-version"])` so the iOS-profiler subsystem is uniformly shell-free.

## Audit
Swept the whole tool-server for the same class (caller-supplied tool parameter → shell-interpreted sink). **No other reachable injection found:**
- Host commands all use argv `execFile`/`spawn` (`adb`, `xctrace`, `emulator`, `simctl`) — serial / udid / avdName / bundleId are discrete argv elements.
- Device-shell commands that interpolate a caller value (`open-url` URL, `launch-app` / `restart-app` bundleId+component) are escaped through `shellQuote()` / single-quote escaping.
- The `command -v ${dep}` PATH probes in `check-deps.ts` / `android-binary.ts` interpolate only the closed `ToolDependency` enum (`adb | xcrun | emulator`), never a request value.

## Buffering note (post-review)
Splitting the `simctl listapps | plutil` shell pipe into two in-process `execFileSync` stages moved stdout buffering from the OS to Node, where each stage inherited Node's 1 MiB default `maxBuffer` — a large `listapps` could throw `ENOBUFS` (`NATIVE_PROFILER_APP_LIST_FAILED`) where the pipe streamed unbounded. All three iOS app-enumeration captures now pass the subsystem's `DEFAULT_EXEC_MAX_BUFFER` (256 MiB, from `run-with-timeout.ts`), matching how the exporter already avoids this class.

## Verification
- `npm run build` (tsc) ✓ · `typecheck:tests` ✓ · `prettier --check` ✓ · `eslint --max-warnings 0` ✓
- Regression tests assert hostile values are only ever discrete argv elements, that no `execSync` is used in the subsystem, and that every app-enumeration stage passes the raised `maxBuffer`. Reverting a fix back to `execSync` (or dropping a `maxBuffer`) makes its test fail — confirming each is a real guard.


hubgan:
`select.ts` `execSync("xcodebuild -version")` is safe but it is the same shape the PR is eliminating; converting it to `execFileSync("xcodebuild", ["-version"])` would make the subsystem uniformly shell-free. Pure consistency, no security impact.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
```diff
@@ -83,10 +83,11 @@ interface DetectedApp {
  * its host PID. The PID is the leading column of `launchctl list`; apps that are
  * registered but not running carry `-` there and are skipped.
  */
-function enumerateRunningUserApps(udid: string): { info: AppInfo; pid: number }[] {
+// Exported for the shell-injection regression test (native-profiler-ios-shell-injection.test.ts).
+export function enumerateRunningUserApps(udid: string): { info: AppInfo; pid: number }[] {
   let launchctlOutput: string;
   try {
-    launchctlOutput = execSync(`xcrun simctl spawn ${udid} launchctl list`, {
+    launchctlOutput = execFileSync("xcrun", ["simctl", "spawn", udid, "launchctl", "list"], {
```
hubgan, line 91: On reachability: this sink (and the `listapps` sink below) is only reached when the session platform resolves to iOS, which comes from `classifyDevice()` matching the anchored hex UDID shape `^[0-9A-Fa-f]{8}-...-[0-9A-Fa-f]{12}$`. A `device_id` that matches that shape cannot contain shell metacharacters, and any payload that does classifies as Android and takes the argv-based `adb` path instead. Sending a hostile `device_id` to `native-profiler-start` confirms this: it routes to `adb -s '<payload>' shell ...` and never reaches this function. So on current main the original `execSync` was gated behind the UDID-shape classifier rather than reachable as the summary states, i.e. dormant / defense-in-depth. The conversion is still worthwhile because it removes reliance on that classifier invariant, but the "reachable" severity framing in the description is worth adjusting.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
```diff
@@ -130,10 +131,18 @@ function enumerateRunningUserApps(udid: string): { info: AppInfo; pid: number }[
...
+    // installed apps. Stage 2: feed it to plutil over stdin to get JSON.
+    const rawPlist = execFileSync("xcrun", ["simctl", "listapps", udid], {
       encoding: "utf-8",
       timeout: DETECT_RUNNING_APP_TIMEOUT_MS,
     });
+    listAppsOutput = execFileSync("plutil", ["-convert", "json", "-o", "-", "--", "-"], {
+      encoding: "utf-8",
+      input: rawPlist,
+      timeout: DETECT_RUNNING_APP_TIMEOUT_MS,
+    });
```
hubgan, line 154: Splitting the old `simctl listapps | plutil` pipeline into two in-process `execFileSync` stages changes the buffering model. Previously the shell wired the pipe at the OS level, so Node only buffered plutil's final JSON while the intermediate OpenStep plist streamed unbounded. Here both `execFileSync` calls buffer their entire stdout in-process with no explicit `maxBuffer`, so each inherits Node's 1 MiB default. On a simulator with a large `listapps` output the first stage can throw `ENOBUFS` (surfaced as `NATIVE_PROFILER_APP_LIST_FAILED`) where the piped version would have succeeded. Output on an iPhone 16 Pro with 26 apps measures ~33 KB, so the window is narrow, but it is a behavior difference the shell pipe did not have, and it is inconsistent with `run-with-timeout.ts` in this same subsystem, which raises `maxBuffer` to 256 MiB precisely to avoid this class of failure.

hubgan (commented):
Reviewed the diff and exercised the native profiler start/stop/analyze flow end-to-end against a build of this branch on a simulator. Two notes inline.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
latekvo: Good catch — you're right, and I've corrected the description accordingly.

`classifyDevice()` (`utils/device-info.ts`) routes purely by shape against the anchored, hex-only regex `^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}$`. Any `device_id` carrying a shell metacharacter fails that match, classifies as **android**, and takes the argv `adb -s <serial> shell …` path — it never reaches this iOS function. Conversely, anything that *does* classify as iOS is, by the regex, metacharacter-free.

I verified it: `classifyDevice()` returns `"android"` for every payload I threw at it (`x; rm -rf /`, `` `id` ``, `$(whoami)`, `a | nc attacker 1`, `booted"; touch /tmp/pwned #`), and only a well-formed hex UDID classifies as `"ios"`. So on `main` the original `execSync` was **dormant / defense-in-depth** behind the classifier invariant, not directly reachable from a tool parameter. I've reframed section 1 from "reachable" to "gated behind the UDID-shape classifier". The conversion still stands on its own merit: it drops the code's reliance on that invariant (a loosened shape regex, or a future refactor reaching the sink with a synthetic `platform: "ios"`, would otherwise re-expose the shell).

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
latekvo: Fixed in bd8a1d2eea1852603b1a68664b3e423761c602c2.

You're right. The shell pipe buffered only plutil's final JSON at the OS level, whereas the two in-process `execFileSync` stages each capture full stdout under Node's 1 MiB default. I reproduced the class directly: `execFileSync("node", ["-e", 'process.stdout.write("x".repeat(2*1024*1024))'])` throws `ENOBUFS` with the default cap and succeeds with `maxBuffer: 256 MiB` (same for the piped-stdin plutil stage).

All three iOS app-enumeration captures (launchctl list, simctl listapps, plutil) now pass `DEFAULT_EXEC_MAX_BUFFER` (256 MiB) from `run-with-timeout.ts` — the same constant the exporter already uses to avoid exactly this. Extended the shell-injection regression test to assert every stage sets it; dropping one `maxBuffer` makes the test fail.

latekvo:
@hubgan re: the `select.ts` `execSync("xcodebuild -version")` consistency suggestion — Fixed in 5b72adc6116832fee93d21ac9465a8bccbc159c2.

Converted `execSync("xcodebuild -version")` → `execFileSync("xcodebuild", ["-version"])` so the iOS-profiler subsystem is uniformly shell-free, and added a guard test (`native-profiler-ios-capture-strategy-shell-safety.test.ts`) asserting the version probe never reaches `execSync`.

packages/tool-server/src/utils/ios-profiler/run-with-timeout.ts
```diff
@@ -31,15 +34,16 @@ export const DEFAULT_EXEC_MAX_BUFFER = 256 * 1024 * 1024;
...
+  options: ExecFileOptions = {}
 ): Promise<{ stdout: string; stderr: string }> {
   // `promisify` is resolved per-call (not at module load) so that test suites
-  // which `vi.doMock("child_process", …)` without an `exec` export can still
+  // which `vi.doMock("child_process", …)` without an `execFile` export can still
   // import this module — matching the original lazy `execSync` usage.
-  const execAsync = promisify(nodeExec);
-  const { stdout, stderr } = await execAsync(command, {
+  const execFileAsync = promisify(nodeExecFile);
+  const { stdout, stderr } = await execFileAsync(file, args as string[], {
```
hubgan, line 46: `...options` is spread after `timeout`/`maxBuffer`/`encoding` in this call, so any option a caller passes wins over the defaults. Since the purpose of this wrapper is to guarantee the 256 MiB buffer and the timeout (the ENOBUFS / event-loop-freeze guards documented in the comment above), a future caller that passes its own `maxBuffer` or `timeout` would silently defeat exactly those guards. Every current caller passes no options, so this is latent rather than an active bug — worth being aware of as the wrapper gains callers.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
```diff
@@ -130,9 +133,24 @@ function enumerateRunningUserApps(udid: string): { info: AppInfo; pid: number }[
...
-    listAppsOutput = execSync(`xcrun simctl listapps ${udid} | plutil -convert json -o - -`, {
+    // Two stages, piped in code rather than by /bin/sh, so the udid is never
+    // interpreted by a shell. Stage 1: ask simctl for the OpenStep plist of
+    // installed apps. Stage 2: feed it to plutil over stdin to get JSON.
+    //
+    // A shell pipe streamed the intermediate plist at the OS level; capturing
+    // each stage's stdout in-process instead means both inherit Node's 1 MiB
+    // default maxBuffer, so a large `listapps` (many installed apps) would throw
+    // ENOBUFS where the pipe succeeded. Match run-with-timeout.ts's 256 MiB.
+    const rawPlist = execFileSync("xcrun", ["simctl", "listapps", udid], {
```
hubgan, line 144: The `plutil` call just below gets a `--` end-of-options guard, but these `xcrun simctl … <udid>` invocations don't. A `udid` starting with `-` would be handed to simctl as a positional that could be parsed as an option flag rather than a device id. Not reachable today — `classifyDevice()`'s anchored UDID-shape regex means any value that reaches this iOS path is a bare hex UUID — but it's an asymmetry with the plutil call right below, given both consume an externally-derived value.

hubgan (approved):

packages/tool-server/src/utils/ios-profiler/run-with-timeout.ts
latekvo: Fixed in 64b80687f... You're right — the wrapper's entire purpose is to guarantee those two, and spreading `...options` last let a caller silently override them.

Reordered so `timeout` + `maxBuffer` apply AFTER `...options`, so a caller can no longer weaken them. I kept `maxBuffer` as a floor (`Math.max(options.maxBuffer ?? 0, DEFAULT_EXEC_MAX_BUFFER)`) rather than a hard set: a caller can still RAISE it for an even larger capture (a host-wide `--all-processes` export can exceed even the current sizes), but never drop below the 256 MiB guard — a hard cap would just be the opposite ENOBUFS footgun. Added a unit test (`native-profiler-ios-run-with-timeout.test.ts`) asserting a caller can't shrink `maxBuffer` or drop the `timeout` but can raise the buffer; reverting the option order makes it fail.

packages/tool-server/src/tools/profiler/native-profiler/platforms/ios.ts
latekvo: Fair observation on the asymmetry, but I'm going to leave the simctl calls as-is — three reasons:

1. **Structurally precluded, not merely "not reachable."** The udid on this path is gated by `classifyDevice`'s anchored regex `^[0-9A-Fa-f]{8}-…-[0-9A-Fa-f]{12}$`, so any value that reaches these `simctl` calls begins with a hex digit and cannot start with `-`. I verified it (same proof as the reachability thread above): every `-`-leading / metacharacter payload classifies as android and takes the adb path, so a `-`-leading udid can't arrive here at all.

2. **`simctl` has no `--` separator to add.** `simctl help` gives `listapps <device>` and `spawn [options] <device> <path> [argv…]` — the device is a bare positional with no documented end-of-options marker. Injecting an undocumented `--` before `<device>` risks the CoreSimulator parser treating `--` as the device id (or, for `spawn`, the executable) and breaking the invocation on real hardware. Unlike `plutil` — which documents `-- specifies that all further arguments are file names` — there's no verified-safe `--` form for simctl.

3. **The plutil `--` guards a different thing.** plutil's trailing `-` is a code-owned stdin sentinel; the `--` there defends that sentinel against a future arg-reorder, not against external input. The simctl udid is external but already constrained by (1) — so it isn't really the same asymmetry.

So I'd be adding an unverified, potentially command-breaking flag to guard a case the classifier already makes impossible. If that invariant were ever loosened, the right place to defend would be the classifier (or a sink-side UDID-shape assertion), not an undocumented simctl separator.

PR #389, opened 2026-06-22
feat: convert argent.md rules to argent skill to prevent always-included context noise

danieldunderfelt:
Currently Argent adds a rules file to the project that is always included in agent context, even when the user does not intend to use Argent. This is context noise, and can in some cases lead to worse agent performance for other tasks.

A better approach is to make an `Argent` skill that the agent can use autonomously when working with Argent. Most of the rules file content is already covered by the various skills, and the skill can be a lighter "router" skill that the agent uses to select the right Argent skill, in addition to providing some general Argent usage instructions.

I realize that this is a significant change, and I am not privy to the decisions or reasons involved in making the original rules file. However, having the always-included rules does significantly hinder my usage of and willingness to recommend Argent. I am of course open to discussion and further iteration on this!


filip131311:
Hello @danieldunderfelt,
thank you for taking the time to analize this problem and propose the solution,
we of course are aware of the issue and are trying to mitigate it in 2 ways: 

1) the default `argent init` command installs mcp and skills at "project level" (unless user chooses otherwise), that means that  the argent will only activate in projects that can benefit from it like mobile apps repositories or web apps.
2) we make it very easy to just run `argent remove` command that will clean up the project for the time of a task that should not use argent

We know that this is not perfect, but it is a compromise that mitigates another frustration our early users faced before we adopted "rules file" namely agents would not always load skills needed for the task even after installing argent, this was a broken promise on our part and we can not come back to that model. 

I am sorry for not noticing this PR sooner. 
Best 
Filip

filip131311:
closing, because of inactivity

PR #354, opened 2026-06-17
chore(deps): bump express to 5.x and migrate route params (fixes #314)

latekvo:
## What

Bumps `express` `^4.19.2` → `^5.2.1` and `@types/express` `^4.17.21` → `^5.0.6` in `@argent/tool-server`, with the source change Express 5 requires. This is the **express-5 half of dependabot #314**, which can't merge as-is because it breaks the build.

## Why #314 broke

Express 5 / `@types/express` 5 widen route param values from `string` to `string | string[]`, so every single-segment param read fails to typecheck — 9 `TS2345` errors across `artifacts.ts`, `http.ts`, `preview.ts`:

```
packages/tool-server/src/artifacts.ts(61,42): error TS2345: Argument of type 'string | string[]' is not assignable to parameter of type 'string'.
... (9 total, incl. preview.ts variant-image/describe routes added by Argent Lens)
```

## Fix

Narrow each param read with `as string`. Every tool-server route uses single-segment params, which path-to-regexp v8 always resolves to a string, so the cast is sound:

- `artifacts.ts` — `id`
- `http.ts` — `deviceId`, `name`
- `preview.ts` — `udid` (×2), `elementId`, `variantId`

No removed Express 5 APIs (`app.del`, `res.sendfile`, `req.param()`, numeric `res.send`/`res.json` status) are used, and no route relies on the changed wildcard/optional path syntax. Lockfile regenerated on **Node 20 / npm 10.8.2** to match the repo's package-lock toolchain.

## Verification (local)

- Reproduced the 9 `TS2345` errors on express 5, then `tsc --build` clean after the narrowings.
- Full CI-equivalent on **Node 20** (the unit-tests toolchain): `tsc --build`, `eslint --max-warnings 0`, `prettier --check`, `typecheck:tests`/`typecheck:scripts`, full unit suite incl. **1250 tool-server tests** — all green. `test/artifacts.test.ts` boots the real Express app via supertest and exercises `GET /artifacts/:id` under Express 5.
- `npm ci` from the committed lockfile succeeds; the lock is idempotent under `npm install --package-lock-only --ignore-scripts` (Node 20).
- **Live device E2E** (out-of-process tool-server under Express 5, booted iPhone 16 Pro Max): `GET /artifacts/:id` → `200 image/png` (331 KB PNG, ws + simulator-server); `GET /preview/simulator-server/:udid` → `200`; `GET /preview/variant-image/:elementId/:variantId` → route matched + params parsed. Unknown artifact id → `404`.

[...]

PR #390, opened 2026-06-22
fix(argent-lens): serve preview UI when installed under a dot-path (nvm/fnm/volta/asdf)

latekvo:
## Problem

The Argent Lens preview window fails to load its UI for anyone whose argent is installed under a path containing a **dot-segment** — most commonly Node managed by **nvm** (`~/.nvm/...`), but also fnm, volta, and asdf. Instead of the variant-selection UI the window shows a raw `NotFoundError: Not Found`, and `await_user_selection` then blocks forever.

Concretely, on such installs:
- `GET /preview/` (index.html) → **404**
- `GET /preview/theme.css` → **404**
- JSON routes (e.g. `GET /preview/simulators`) → 200

## Root cause

Regressed by the express 4 → 5 bump (#354), shipped in **v0.12.1**. Express 5's `send@1` defaults to `dotfiles: "ignore"`, which 404s any file path containing a dot-segment. The preview routes call `res.sendFile(absolutePath)` where `absolutePath` is the on-disk install location; under nvm that path includes the `.nvm` segment, so `send` refuses to serve it.

Reproduced in isolation against `express@5.2.1` / `send@1.2.1` (the shipped versions):

| call | result |
| --- | --- |
| `res.sendFile("/Users/…/.nvm/…/index.html")` | **404** |
| same file copied to a non-dot path | 200 |
| `res.sendFile(basename, { root: dir })` | 200 |
| any of the above on `express@4.22.1` | 200 |

The express-4 column confirms the bump is the regression; the `{ root }` column is the fix.

## Fix

Serve the preview UI via `sendFile(basename, { root: dir })` so the request path carries no dot-segment — `send` only applies its dotfile filter to the path *after* `root`, so the install location no longer matters. Factored into a small `serveUiFile` helper shared by the `index.html` and `theme.css` routes.

## Test

`preview-ui-dotfile.test.ts` serves the UI from a temp directory whose name starts with `.` and asserts 200 for both `index.html` and `theme.css`. It fails on the old `res.sendFile(p)` under express 5 and passes with the fix.

### DIFF
--- a/packages/tool-server/src/preview.ts
+++ b/packages/tool-server/src/preview.ts
@@ -34,6 +34,18 @@ function findUiFile(name: string): string | null {
   return null;
 }
 
+// Serve a resolved preview-UI file. We go through `sendFile` with an explicit
+// `root` instead of passing the absolute path directly: Express 5's `send`
+// defaults to `dotfiles: "ignore"`, which 404s any path containing a dot-segment.
+// argent is routinely installed under one (nvm's `~/.nvm/...`, fnm, volta,
+// asdf), so `sendFile(absolutePath)` silently fails there and the Lens preview
+// window can't load. Scoping to `{ root: dir }` makes the request path just the
+// basename — no dot-segment — so the file is served regardless of install path.
+export function serveUiFile(res: Response, filePath: string, contentType: string): void {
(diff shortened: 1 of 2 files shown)


latekvo:

PR #392, opened 2026-06-22
fix(argent-lens): anchor variants reliably right after navigation (warm-up retry + exact-match)

latekvo:
## Problem

A `propose_variant` made right after navigating to a screen — exactly what the Lens workflow does (navigate → screenshot → propose) — often fails to anchor the floating card to the element. The card shows up unanchored / in the "hidden" bucket ("didn't register against anything"), or it latches onto a same-text **distractor** (e.g. the bottom **"Favourites" tab** instead of the **"Favourites" header**).

## Root cause

`propose_variant` auto-captures the matched element's frame via `captureElementFrame`, which describes the device **once** and matches. But iOS' accessibility tree is **not ready** at propose time — measured against a real device:

- For the first **~1.25 s** after a screen appears, describe returns an essentially empty tree (`nodes=1–2`). A single describe there matches nothing → `null` frame → no anchor.
- The tree then lands **incrementally**: persistent nav chrome (the `"Favourites (5)"` tab) shows up *before* the screen's own content (the `"Favourites"` header). A single describe in that window matches the **tab**, not the header.

So the one-shot capture either gets nothing or the wrong (same-text) element.

## Fix

`captureElementFrame` now:
- **Retries across the warm-up window** instead of giving up on the first (empty) describe.
- **Holds out for an exact label/identifier hit** — the intended element — and only falls back to a substring match once the retry budget is spent. This skips the half-built tree where only a same-text distractor exists yet.

`findElementMatch` / `matchFrameInTree` now **tier exact matches above substring matches** (within a tier, smallest sane box still wins), so the exact `"Favourites"` header beats the `"Favourites (5)"` tab once both are present.

## Verification

- New `match-element-frame-cold-describe.test.ts` (6 cases): holds out through the cold + half-built-tab window to the exact header; returns immediately when warm; falls back to a substring match after the budget; returns null when absent; `findElementMatch` exact-beats-substring ranking.
- Updated two existing tests that asserted the old single-describe behavior (`{ attempts: 1 }`).
- All matcher tests green; monorepo `tsc --build` + `typecheck:tests` green.
- **Real device (pokemon app, iPhone 16 Pro Max):** reproduced the ~1.25 s empty-tree window by polling describe across a relaunch; confirmed the old single-shot returns `null` in that window; confirmed the new logic resolves the **exact header** (`y≈0.092`), not the tab (`y≈0.923`), on the Favourites screen.

## Note

A substring-only match (e.g. a list row matched by a partial name) now waits the full retry budget before falling back, since there is no way to tell "the exact element hasn't rendered yet" from "there is no exact element" without risking the distractor bug. Propose is not latency-critical; correctness is preferred. Agents are still best served by `identifier`/`label` matchers (already recommended in the skill).

### DIFF
--- a/packages/tool-server/src/utils/match-element-frame.ts
+++ b/packages/tool-server/src/utils/match-element-frame.ts
@@ -28,34 +28,52 @@ function normLabel(s: string | undefined): string {
 // mirrors the UI's `spotMaxFrameArea` guard.
 const MAX_FRAME_AREA = 0.85;
 
-function nodeMatches(n: DescribeNode, match: VariantMatch, needle: string): boolean {
-  if (!needle) return false;
+// Whether a node matches, and whether that match is EXACT (the whole normalized
+// label/identifier/value equals the needle) rather than a looser substring.
+// Exact hits win: a propose for "Favourites" should anchor the header
+// (label === needle), not the "Favourites (5)" tab (a same-text distractor).
+function matchNode(
+  n: DescribeNode,
+  match: VariantMatch,
+  needle: string
+): { exact: boolean } | null {
   const label = normLabel(n.label);
   const ident = normLabel(n.identifier);
   const value = normLabel(n.value);
   const role = (n.role || "").toLowerCase();
   switch (match.by) {
     case "label":
-      return label === needle || label.includes(needle);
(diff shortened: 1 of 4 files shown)


latekvo:

PR #393, opened 2026-06-22
fix(argent-lens): always offer to reveal off-screen choices

latekvo:
## Problem

In the Argent Lens window, the off-screen reveal prompt — both the **"Show them"** card and the collapsed **"N hidden — show"** pill — was hidden unless **every on-screen choice was already staged** (`onScreenStaged`). So when a proposal's element was off-screen (e.g. it lives on another tab) and the user hadn't yet staged all the *on-screen* choices, there was a hidden-choice count but **no affordance to reveal it** — the off-screen choice was unreachable.

## Fix

Show the reveal prompt whenever any element is off-screen:

```js
const showPrompt = offScreenCount > 0 && !vp.revealOffscreen && !vp.submitting;
```

i.e. drop the `(onScreenStaged || vp.offscreenPromptCollapsed)` gate from the visibility condition (and the now-unused `onScreenStaged`). `offscreenPromptCollapsed` still selects between the full card and the collapsed pill.

## Verified

Patched into the running global bundle's `preview-ui/index.html` and confirmed the tool-server serves the un-gated version (`GET /preview/` → 200, gate string gone). The edited file is prettier-clean under the repo's `prettier@3.8.4`.

### DIFF
--- a/packages/ui/index.html
+++ b/packages/ui/index.html
@@ -4296,24 +4296,14 @@
         const allChosen = vp.proposals.length > 0 && vp.proposals.every((p) => vp.staged[p.id]);
         footEl.classList.toggle("ready", allChosen && !vp.submitting);
 
-        // Off-screen proposals stay folded away (the "N hidden" badge) until the
-        // user opts in. Once every ON-screen choice is staged, offer to reveal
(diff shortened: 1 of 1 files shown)


latekvo:

PR #394, opened 2026-06-22
fix(argent-lens): preview-UI serving, variant matcher, and off-screen reveal

latekvo:
Three independent reliability fixes for **Argent Lens** (the `propose_variant` / `await_user_selection` variant-review flow and its preview window). They touch disjoint files and were validated separately; combined here into one PR.

---

### 1. Preview window 404s on dot-path installs (nvm/fnm/volta/asdf)

The Lens window showed a raw `NotFoundError` instead of the variant UI whenever argent was installed under a path containing a dot-segment (most commonly nvm, `~/.nvm/...`). `GET /preview/` and `/preview/theme.css` returned 404 while JSON routes worked.

Cause: the express 4 → 5 bump (#354, shipped in v0.12.1). Express 5's `send@1` defaults to `dotfiles: "ignore"`, which 404s any file path containing a dot-segment, and the preview routes pass the absolute install path to `res.sendFile`. Proven in isolation: express 5.2.1 `sendFile` of the file → 404; express 4.22.1 → 200; `sendFile(basename, { root: dir })` → 200.

Fix: serve the preview UI via `sendFile(basename, { root: dir })` (a `serveUiFile` helper) so the request path carries no dot-segment. Regression test serves the UI from a temp dir whose name starts with `.`.

> Note: the underlying `send@1` dot-path behaviour is a property of the express-5 migration, not of Lens — these two routes are simply the only `sendFile` calls in the tool-server today.

### 2. Variant fails to anchor right after navigation

A `propose_variant` made immediately after navigating to a screen (exactly the navigate → screenshot → propose workflow) often failed to anchor: the card floated/hid ("didn't register"), or latched onto a same-text distractor (e.g. a bottom-bar tab instead of a header).

Cause: `captureElementFrame` describes the device **once** at propose time, but iOS' accessibility tree is empty for ~1.25 s after a screen appears and then fills incrementally (nav chrome before screen content) — measured on a real device.

Fix: `captureElementFrame` now retries across the warm-up window and **holds out for an exact label/identifier hit** (the intended element), falling back to a substring match only after the budget. `findElementMatch` / `matchFrameInTree` rank exact matches above substring ones. Adds `match-element-frame-cold-describe.test.ts`; two existing tests updated for the new retry signature.

### 3. Off-screen choices were unreachable

In the preview window, the off-screen reveal prompt (the "Show them" card and the "N hidden — show" pill) was hidden unless **every on-screen choice was already staged** (`onScreenStaged`). So a proposal whose element was off-screen showed a hidden-choice count with no affordance to surface it.

Fix: show the prompt whenever any element is off-screen (`offScreenCount > 0 && !revealOffscreen && !submitting`), dropping the `onScreenStaged` gate.

---

Supersedes #390, #392, #393 (same commits, combined).

### DIFF
--- a/packages/tool-server/src/preview.ts
+++ b/packages/tool-server/src/preview.ts
@@ -34,6 +34,18 @@ function findUiFile(name: string): string | null {
   return null;
 }
 
+// Serve a resolved preview-UI file. We go through `sendFile` with an explicit
+// `root` instead of passing the absolute path directly: Express 5's `send`
+// defaults to `dotfiles: "ignore"`, which 404s any path containing a dot-segment.
+// argent is routinely installed under one (nvm's `~/.nvm/...`, fnm, volta,
+// asdf), so `sendFile(absolutePath)` silently fails there and the Lens preview
+// window can't load. Scoping to `{ root: dir }` makes the request path just the
+// basename — no dot-segment — so the file is served regardless of install path.
+export function serveUiFile(res: Response, filePath: string, contentType: string): void {
+  res.set("Cache-Control", "no-store, must-revalidate");
+  res.type(contentType).sendFile(path.basename(filePath), { root: path.dirname(filePath) });
+}
+
 function wsUrlFromHttp(httpUrl: string): string {
   const u = new URL(httpUrl);
   const scheme = u.protocol === "https:" ? "wss:" : "ws:";
@@ -340,8 +352,7 @@ export function createPreviewRouter(registry: Registry): Router {
       res.status(404).type("text/plain").send("theme.css not found");
       return;
     }
-    res.set("Cache-Control", "no-store, must-revalidate");
-    res.type("text/css").sendFile(p);
+    serveUiFile(res, p, "text/css");
   });
 
   router.get("/", (req: Request, res: Response) => {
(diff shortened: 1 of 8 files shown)


hubgan:
`propose_variant` now blocks on the full retry budget for absent/off-screen elements (Android severe). `captureElementFrame` is awaited synchronously inside `propose_variant.execute()`. When there's no exact hit, it runs all 8 attempts (8 × describe + 7 × 300ms). The 300ms estimate assumes a near-instant describe, which holds on iOS but not Android, where each `uiautomator` dump is seconds:

| Platform / case | Result |
|---|---|
| iPhone 17 Pro — **warm** exact hit | 36 ms (returns on attempt 1) |
| Pixel 7 — **warm** exact hit | 57 ms (returns on attempt 1) |
| iPhone 17 Pro — cold screen, eventual exact hit | 2208 ms |
| iPhone 17 Pro — absent / substring-only (full budget) | ~2.5–4 s |
| **Pixel 7 — absent element (full budget)** | **17,425 ms** (returned `null`) |
| Pixel 7 — wrong screen / absent | 19,873 ms |

hubgan (approved):

PR #396, opened 2026-06-23
feat(tool-server): implement await-ui-element tool

hubgan:
## Summary

Adds a `await-ui-element` tool that blocks until a UI condition is satisfied or a timeout elapses.

This is one of the most impactful missing tools. Without it, agents have to poll manually with `screenshot → describe → check` loops — slow, token-heavy, and unreliable (fixed `sleep` delays either over-wait or fire before the screen settles). `await-ui-element` moves the poll loop server-side: one call in, a definitive verdict out.

Works on **iOS, Android, and Chromium (CDP)**, polling the same accessibility / DOM tree as `describe`.

## Conditions

- `visible <selector>` — wait until an element appears on screen
- `hidden <selector>` — wait until an element disappears
- `exists <selector>` — wait until an element is in the tree
- `text <selector> <expectedText>` — wait until the matched element contains the text

**selector** is `{ text?, identifier?, role? }` — every field provided must match (case-insensitive substring). Polls internally; default `timeoutMs` **5000**, default `pollIntervalMs` **400**. Returns `{ success: boolean, elapsed: number }`, with a `note` explaining what was seen on timeout.

```json
{ "udid": "<UDID>", "condition": "visible", "selector": { "text": "Continue" } }
```

Also usable as a step inside `run-sequence`, so a tap → await-ui-element → tap chain runs in a single call.

## Notable details

- **Multi-match aware** — a substring selector can hit several nodes, so conditions evaluate the whole match set: `visible` holds if any match is on-screen, `hidden` only if none is (a zero-area container can't flip the verdict).
- **`hidden` guard** — a selector that matches nothing counts as already-hidden; the `note` flags when the selector never matched, so a typo doesn't read as a silent success.
- **Deadline-clamped poll** — a large `pollIntervalMs` can't overshoot `timeoutMs`.
- **Abort-aware** — both the `time` sleep and the poll loop stop promptly on cancellation.

## Changes

- `tools/await-ui-element/index.ts` — new tool (factory, mirrors `describe`'s service resolution)
- `utils/setup-registry.ts` — registers the tool
- `tools/run-sequence/index.ts` — allows `await-ui-element` as a step + arg-doc + example
- `test/await-ui-element.test.ts` — new test suite
- `skills/rules/argent.md` + `argent-device-interact/SKILL.md` — document `await-ui-element`, replace the "don't poll `screenshot`" guidance with it

## Test plan

- [x] `time` sleeps and succeeds without touching a device
- [x] `visible` succeeds once the element appears; times out with a diagnostic note when it doesn't
- [x] `exists` / `hidden` / `text` happy paths; `text` timeout note reports last-seen text
- [x] `hidden` instant success **and** the "selector never matched" flag
- [x] poll sleep clamped to the deadline (large `pollIntervalMs` can't overshoot `timeoutMs`)
- [x] prompt cancellation aborts both the poll loop and the `time` sleep
- [x] multi-match evaluation (`findAll` / `evaluateMatches`)
- [x] schema validation (required fields per condition, `durationMs` cap)


latekvo (commented):
This sounds like a nitpick but i think it might be quite imporant for the agent - let's name this tool one of these names whichever you believe to be the most fitting: `wait-for`, `wait-for-ui` or `await-ui-element`

The reason is that by default the only thing the agent sees is tool name.
The agent then has to manually call `ToolSearch` tool to view descriptions of the tools it wants to call.  
Given that context, `wait` alone may be too broad or misleading to the agent.  

Note that we already use `await` in our `await_user_selection` tool, so following that tool naming convention would make some sense.

packages/skills/skills/argent-device-interact/SKILL.md
```diff
@@ -176,6 +177,23 @@ Special keys: `enter`, `escape`, `backspace`, `tab`, `space`, `arrow-up`, `arrow
...
 
+### wait — Block until a UI condition holds
+
+Instead of polling `screenshot`/`describe` in a loop, use `wait` to block server-side until a condition is met (or `timeoutMs`, default 5000ms, elapses). It polls the same accessibility/DOM tree as `describe`.
+
+```json
+{ "udid": "<UDID>", "condition": "visible", "selector": { "text": "Continue" } }
+```
+
+- `condition`: `time` (just sleep `durationMs`, no device needed), `exists`, `visible`, `hidden`, or `text`.
```
latekvo: We don't want to be adding redundant capabilities. Claude Code and other harnesses already implement their own sleep, and also strongly discourage using it.

packages/tool-server/src/tools/run-sequence/index.ts
```diff
@@ -18,6 +18,7 @@ const ALLOWED_TOOLS = new Set([
   "button",
   "keyboard",
   "rotate",
+  "wait",
```
latekvo: Adding `wait` here makes run-sequence host a step that can block on a UI condition for a long time (up to its `timeoutMs` — default 5s, max 120s), but that sub-step never receives the request's cancellation signal. `run-sequence`'s `execute` takes no `ToolContext`, and it dispatches each step with `registry.invokeTool(step.tool, toolArgs)` and no options, so the `AbortSignal` the HTTP layer wires to the outer request (via `res.on("close")`) reaches neither the sequence loop nor the `wait` sub-step. When the client disconnects mid-sequence, the server keeps polling the device until the wait's own timeout elapses — `wait`'s own abort handling only takes effect when it is invoked directly. Multiple `wait` steps stack this delay.

packages/tool-server/src/tools/run-sequence/index.ts
```diff
@@ -111,6 +113,13 @@ Example — type text and submit:
     { "tool": "keyboard", "args": { "key": "enter" } }
   ]}
 
+Example — tap, wait for the next screen's element, then tap it:
+  { "udid": "<UDID>", "steps": [
+    { "tool": "gesture-tap", "args": { "x": 0.5, "y": 0.9 } },
+    { "tool": "wait", "args": { "condition": "visible", "selector": { "text": "Continue" } } },
```
latekvo: This `tap → wait → tap` example relies on the `wait` gating the second tap, but run-sequence only stops the sequence when a step throws. `wait` reports an unmet condition by returning `{ success: false }` rather than throwing, so a timed-out `wait` is recorded as a completed step and the following `gesture-tap` runs regardless — the blind, mistimed tap the `wait` was inserted to prevent. The failed result is present in `steps[]`, but `completed` counts it and the chain does not halt, so the documented use case does not actually guard.

packages/tool-server/src/tools/wait/index.ts
```diff
@@ -0,0 +1,372 @@
...
+// match several nodes — e.g. on Android a zero-area container kept only for its
+// visible children plus the visible leaf you actually meant — so conditions are
+// evaluated across the whole set rather than the first hit (see evaluateMatches).
+// Exported for unit tests.
+export function findAll(
+  node: DescribeNode,
+  selector: Selector,
+  acc: DescribeNode[] = []
+): DescribeNode[] {
+  if (matchNode(node, selector)) acc.push(node);
```
latekvo: `findAll` tests the passed node itself, which includes the synthetic full-screen root every adapter injects (iOS `AXGroup`, Android `Screen`, Chromium `html`; frame `0,0,1,1`). The `describe` tool renders that root only as a non-selectable `ROOT` header line — never as a matchable element — but here it is matched like any other node, and its `1×1` frame always satisfies `isVisible`. A `role` selector whose value is a case-insensitive substring of the root role (e.g. `role: "AXGroup"`, which is also the iOS default role for untyped elements) therefore matches it: `visible`/`exists` succeed instantly on any screen — including when the AX backend is down and `describeIos` has fallen back to an empty tree — and `hidden` can never succeed because the root never disappears.

packages/tool-server/src/tools/wait/index.ts
```diff
@@ -0,0 +1,372 @@
...
+export function evaluateMatches(params: Params, matches: DescribeNode[]): boolean {
+  switch (params.condition) {
+    case "exists":
+      return matches.length > 0;
+    case "visible":
+      return matches.some(isVisible);
+    case "hidden":
+      return !matches.some(isVisible);
+    case "text": {
+      const first = matches[0];
```
latekvo: The `text` condition (and the timeout note at line 233) inspect `matches[0]` in `findAll`'s raw pre-order/array order. On iOS the `describe` tool presents leaves sorted into reading order (`frame.y`, then `frame.x`) via `format-tree.ts` `renderFlat`, so when a substring selector matches more than one element the element `wait` keys on can differ from the one the agent saw listed first in `describe`. The success/failure verdict, and the element text quoted in the timeout note, then refer to a different element than the agent intended.

packages/tool-server/src/tools/wait/index.ts
```diff
@@ -0,0 +1,372 @@
...
+// through the tool's own services() declaration. Only the Chromium CDP session
+// flows in as a normal service.
+export function createWaitTool(registry: Registry): ToolDefinition<Params, WaitResult> {
+  async function fetchTree(
+    device: DeviceInfo,
+    params: Params,
+    services: Record<string, unknown>
+  ): Promise<DescribeNode> {
+    if (device.platform === "ios") {
+      return (await describeIos(registry, device, { bundleId: params.bundleId })).tree;
```
latekvo: `fetchTree` keeps only `.tree` and drops the `should_restart` and `hint` that `describeIos` returns. On a simulator not booted through argent (degraded AX) `describe` surfaces an actionable hint (“call boot-device with force=true …”), and when the foreground app needs a restart for native injection it returns `should_restart: true`; `wait` discards both and just times out with a generic `no element matched`, so the agent never learns the real cause. Relatedly, `describeIos` swallows AX/native failures and returns an empty tree instead of throwing, so on iOS a backend that is actually down is indistinguishable from “element absent”: `hidden` returns `success: true` while AX is down, and the `last tree fetch failed` branch of `timeoutNote` is unreachable on iOS (it can only fire for the Android/Chromium fetches, which do throw).

packages/tool-server/src/tools/wait/index.ts
```diff
@@ -0,0 +1,372 @@
...
+      let fetchError: string | undefined;
+      // For `hidden`: did the selector ever match across polls? Distinguishes
+      // "the element was there and disappeared" from "the selector never matched
+      // at all" — otherwise a typo'd selector is an instant false-positive.
+      let everMatched = false;
+
+      for (;;) {
+        if (signal?.aborted) return cancelled();
+        try {
+          lastTree = await fetchTree(device, params, services);
```
latekvo: Each tree fetch is awaited with no per-fetch deadline or abort signal — `timeoutMs` and `signal` are only checked between polls. A single slow or hung fetch therefore runs to the underlying `describe` timeout regardless of `timeoutMs`: the Android `uiautomator dump` fallback allows up to 20s, so a `wait` with `timeoutMs: 1000` can block ~20s, and an abort that arrives while a fetch is in flight is not observed until that fetch resolves.

packages/tool-server/test/wait.test.ts
```diff
@@ -0,0 +1,362 @@
...
+
+const FRAME = { x: 0.1, y: 0.4, width: 0.8, height: 0.05 };
+
+function makeMockRegistry(axService: AXServiceApi) {
+  return {
+    resolveService: vi.fn(async (urn: string) => {
+      if (urn.startsWith("AXService:")) return axService;
+      // wait never needs native-devtools when AX returns elements; an empty
+      // AX tree would fall through to here, so fail loudly to surface misuse.
+      throw new Error(`unexpected service: ${urn}`);
```
latekvo: Every test in this file drives the iOS AX path: `IOS_UDID` plus a mock registry whose `resolveService` throws for any non-`AXService` URN. As a result the Android (`describeAndroid`) and Chromium (`describeChromium`) branches of `fetchTree`, and the `last tree fetch failed` branch of `timeoutNote` (reachable only when a fetch throws, which it can on those two platforms), are not covered.

packages/skills/skills/argent-device-interact/SKILL.md
latekvo: Fixed in 03023d9e — the `time` / bare-sleep condition was dropped entirely. The tool now accepts only `exists` / `visible` / `hidden` / `text` (there is no `durationMs` anywhere), so it no longer duplicates a harness-level sleep. The skill text now points at the harness's own sleep for a plain pause and frames the tool purely as a UI-condition wait.

packages/tool-server/src/tools/run-sequence/index.ts
latekvo: Fixed in 03023d9e — `run-sequence.execute` now takes a `ToolContext` and forwards `ctx.signal` (plus the child-invocation recorder) into every step through `invokeSubTool` -> `registry.invokeTool`, which builds the sub-tool's `ctx` from it. A client disconnect now aborts an in-flight `await-ui-element` step and the sequence loop between steps, instead of polling on to the wait's own timeout. Covered by the new "forwards the request abort signal into each sub-tool" / "does not run any step when the signal is already aborted" tests, and I confirmed the exact signal instance reaches the sub-tool's `ctx.signal`.

packages/tool-server/src/tools/run-sequence/index.ts
latekvo: Fixed in 03023d9e — `run-sequence` and `flow-execute` now recognise an unmet wait via the exported `isUnmetUiWaitResult` (tool id is `await-ui-element` and `result.success === false`) and stop the sequence at that step, recording it as an error so the following step does not run. Verified live against a simulator: a tap -> await-ui-element(visible, missing element) -> tap sequence returns completed:1 / total:3 with the trailing gesture-tap never dispatched.

packages/tool-server/src/tools/wait/index.ts
latekvo: Fixed in 03023d9e — `findAll` now iterates `root.children` and never tests the synthetic root itself, mirroring `describe`, which renders that container only as the non-matchable ROOT header. A `role:"AXGroup"` / `html` selector no longer satisfies `visible`/`exists` on any screen, and `hidden` can resolve. Covered by the "synthetic root is never matched" tests (including the empty-tree `visible`/`hidden` cases).

packages/tool-server/src/tools/wait/index.ts
latekvo: Fixed in 03023d9e — the `text` verdict and the timeout note now key off `firstInReadingOrder` (smallest frame.y, then frame.x), which matches format-tree's `renderFlat` sort, so the element they reference is the one the agent saw first in `describe` rather than whichever node DFS reached first. Covered by the "`text` inspects the first match in READING order, not DFS order" test.

packages/tool-server/src/tools/wait/index.ts
latekvo: Fixed in 03023d9e — the fetch now retains the full `DescribeTreeData`. `should_restart` and `hint` are folded into the timeout note via `appendDiagnostics`, and `isBlindRead` stops `hidden` from resolving positively off a degraded/empty iOS tree — it returns success:false and surfaces the boot hint instead. Covered by "does NOT report `hidden` success while the AX backend is down" and "surfaces the degraded boot hint in a timeout note".

packages/tool-server/src/tools/wait/index.ts
latekvo: Fixed in 03023d9e — each fetch is now raced against the remaining wait budget and the abort signal via `settleWithin`, so a slow/hung describe (e.g. the up-to-20s Android `uiautomator dump` fallback) can no longer overshoot `timeoutMs`, and an abort that arrives mid-fetch is observed promptly. The abandoned fetch's late settle is consumed so it can't surface as an unhandled rejection. Covered by "bounds a single slow fetch to the wait budget" and "aborts promptly even while a fetch is in flight".

packages/tool-server/test/wait.test.ts
latekvo: Fixed in 03023d9e — the suite now also drives the Android devtools path and the Chromium CDP path, and exercises the "last tree fetch failed" branch via a Chromium fetch that throws (the branch reachable on the platforms whose fetch actually rejects), alongside the degraded-AX and per-fetch-deadline cases.

packages/tool-server/src/tools/await-ui-element/index.ts
```diff
@@ -0,0 +1,475 @@
...
+
+  return {
+    id: AWAIT_UI_ELEMENT_TOOL_ID,
+    description: `Block until a UI element reaches an expected state or a timeout elapses, so you don't have to poll screenshot/describe yourself.
+
+Conditions:
+  exists   — the selector matches an element anywhere in the tree.
+  visible  — the selector matches an element with a non-zero on-screen frame.
+  hidden   — the selector matches nothing, or only a zero-area element (e.g. a spinner that disappeared).
+  text     — the element matched by the selector contains expectedText (case-insensitive substring).
```
latekvo: This describes `text` as checking "the element matched by the selector," but the selector is a substring match that can hit several nodes and the condition only inspects the first match in reading order. When a broad selector matches more than one element and a lower one — not the topmost — is the one containing `expectedText`, the wait reports failure even though an on-screen matching element does contain the text. The single-element phrasing doesn't signal that the verdict is keyed to the topmost match only, so an agent using a loose selector can be surprised by a false negative.

packages/tool-server/src/tools/await-ui-element/index.ts
```diff
@@ -0,0 +1,475 @@
...
+}
+
+function collectMatches(node: DescribeNode, selector: Selector, acc: DescribeNode[]): void {
+  if (matchNode(node, selector)) acc.push(node);
+  for (const child of node.children) collectMatches(child, selector, acc);
+}
+
+// Every node matching the selector in the subtree, EXCLUDING `root` itself.
+// `root` is the synthetic full-screen container every describe adapter injects
+// (iOS `AXGroup`, Android `hierarchy`/`Screen`, Chromium `html`; frame
```
latekvo: This rationale states every adapter injects a synthetic full-screen root with frame `0,0,1,1`, but on Chromium the root is the real `<html>` element (`describeChromium` walks `document.documentElement`) and its frame comes from `getBoundingClientRect`, not a synthetic `0,0,1,1`. Excluding it is still correct, but the stated reasoning misdescribes the Chromium case. A side effect worth noting: because `<html>`'s own `id` / `aria-label` / author role sit on the excluded root, a selector targeting those attributes matches nothing on Chromium.

packages/tool-server/src/tools/await-ui-element/index.ts
```diff
@@ -0,0 +1,475 @@
...
+export function findAll(root: DescribeNode, selector: Selector): DescribeNode[] {
+  const acc: DescribeNode[] = [];
+  for (const child of root.children) collectMatches(child, selector, acc);
+  return acc;
+}
+
+// describe prunes off-screen / zero-size nodes on Chromium and the compressed
+// Android dump, and iOS AX only returns on-screen leaves — so a non-zero frame
+// area is a cheap, reliable proxy for "visible".
+function isVisible(node: DescribeNode): boolean {
```
latekvo, line 203: `findAll` matches against every node in the tree, while `describe`'s rendered body only emits nodes that pass its content/role filter. As a result a `role`- or `identifier`-only selector can match a structural container (e.g. an unlabeled `AXGroup`) that `describe` never lists, so `visible` / `exists` can hold for an element the agent never saw in `describe`'s output. The comment frames the match set as mirroring format-tree, but only the root exclusion is shared — the per-node visibility gate is not.

packages/tool-server/src/tools/await-ui-element/index.ts
```diff
@@ -0,0 +1,475 @@
...
+  }
+}
+
+// A degraded / blind read: the tree came back EMPTY *and* the adapter flagged it
+// as unreliable (iOS AX down or native injection pending → `describeIos` returns
+// an empty tree plus a hint / should_restart instead of throwing). Absence in
+// such a tree is not evidence the element is gone, so we must not let `hidden`
+// resolve positively off it — otherwise "AX is down" reads as "element hidden".
+function isBlindRead(data: DescribeTreeData): boolean {
+  return data.tree.children.length === 0 && Boolean(data.hint || data.should_restart);
```
latekvo: `isBlindRead` only treats an empty tree as unreliable when `hint` or `should_restart` is set, and only the iOS path ever sets those. On Android and Chromium an empty tree is always treated as a confirmed read, so a `hidden` (or text-absence) wait resolves immediately on any empty tree, including a momentary blank frame mid-navigation. When the element matched on an earlier poll (`everMatched` true) and a later poll lands on a transient empty tree, `hidden` returns success with no caveat — which can release a gated tap against a screen that only briefly went blank.

latekvo (approved):
Reviewed end-to-end, including running the built tool-server against a live iOS simulator over HTTP: `visible` / `text` succeed once the element appears, `hidden` correctly refuses to fire while the element is still on screen, an unmet wait inside `run-sequence` stops the sequence so the trailing tap never runs, and the schema rejects `text` without `expectedText`. Cross-platform tree handling, abort/cancellation propagation, and the per-fetch deadline all check out. All eight earlier review threads are addressed in this revision (replied + resolved individually).

The inline notes below are minor and don't block.

One item that can't be anchored to a line: the PR title and description still document the previous `wait` tool and its `time` / `durationMs` sleep condition, which were removed in the rename to `await-ui-element`. The test-plan checkboxes and the `Changes` file paths (`tools/wait/index.ts`, `test/wait.test.ts`) are stale relative to the merged code and would carry into the squash-commit message — worth refreshing before merge.

PR #397, opened 2026-06-23
fix(debugger): stop unmatched device_id silently collapsing onto one device

latekvo:
## Problem

Opening debugger sessions with **different** `device_id`s could all resolve to the **same** physical device.

`selectTarget` filters Metro's `/json/list` CDP targets by `reactNative.logicalDeviceId === device_id`, but when the provided `device_id` matched **no** target it kept *all* candidates and fell through to the first/priority target:

```ts
const filtered = candidates.filter((t) => t.reactNative?.logicalDeviceId === options.deviceId);
if (filtered.length) candidates = filtered; // unmatched id → silently ignored
```

This silently no-ops surprisingly often: Metro's `logicalDeviceId` is an internal hash, **not** the iOS simulator UDID. Verified live against a real device:

- simulator UDID: `6DBF83B4-F341-4F8D-B48D-CD8FF312CCFB`
- Metro `logicalDeviceId`: `a1956f6bd1e3a3ac3dbb505e24d79687a5ea60b1`

With one device the fallback is harmless (it picks the only device). With **several** devices on one Metro, every unmatched `device_id` collapses onto whichever device Metro lists first — so two debugger sessions opened with different `device_id`s drive the same device.

## Fix

In `selectTarget`, only fall back to the priority target when there is **at most one distinct device** to fall back to. When multiple distinct devices are connected and the `device_id` matches none, throw an error that lists the valid `logicalDeviceId`s so the caller can re-target instead of being silently misrouted. The single-device convenience path (pass any id — including a UDID — and get the one connected device) is preserved.

## Verification

- `selectTarget` unit tests: distinct ids → distinct devices; ambiguous unmatched id throws and surfaces the valid ids; single-device and no-`logicalDeviceId` fallbacks still work.
- Full-stack multi-device test driving `Registry → JsRuntimeDebugger → debugger-connect` against a two-device mock Metro: distinct `device_id`s route to distinct devices; unmatched id is rejected.
- Full `tool-server` suite: **1203 passed**.
- Confirmed the real contract on a live iOS sim: `debugger-connect` with the UDID still connects (single-device fallback) and returns `logicalDeviceId a1956f…`.

## Notes

The three RN-only debugger tools (`debugger-component-tree`, `debugger-inspect-element`, `debugger-reload-metro`) build their Metro service ref via a raw template literal instead of `debuggerServiceRef()`. For the Metro path the resulting URN is byte-identical, so it's not part of this bug; left untouched to keep the fix scoped.

### DIFF
--- a/packages/tool-server/src/utils/debugger/target-selection.ts
+++ b/packages/tool-server/src/utils/debugger/target-selection.ts
@@ -24,8 +24,35 @@ export function selectTarget(
   let candidates = targets;
 
   if (options?.deviceId) {
-    const filtered = candidates.filter((t) => t.reactNative?.logicalDeviceId === options.deviceId);
-    if (filtered.length) candidates = filtered;
+    const deviceId = options.deviceId;
+    const filtered = candidates.filter((t) => t.reactNative?.logicalDeviceId === deviceId);
+    if (filtered.length) {
+      candidates = filtered;
+    } else {
+      // No target matches the requested device. Silently falling back to the
+      // first/priority target here would route EVERY unmatched device_id to the
+      // same device — so two debugger sessions opened with different device_ids
+      // would both land on whichever device Metro happens to list first. Only
+      // fall back when there is a single device to fall back to; with multiple
+      // distinct devices connected, refuse to guess and report the valid ids so
+      // the caller can re-target (the logicalDeviceId is what debugger-connect
+      // returns and what subsequent debugger-* calls must pass).
+      const distinctDeviceIds = [
+        ...new Set(
+          targets
+            .map((t) => t.reactNative?.logicalDeviceId)
+            .filter((id): id is string => id !== undefined && id !== "")
+        ),
+      ];
+      if (distinctDeviceIds.length > 1) {
+        throw new Error(
+          `No debugger target matches device_id "${String(deviceId)}". ` +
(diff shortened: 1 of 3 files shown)


packages/tool-server/src/utils/debugger/target-selection.ts
```diff
@@ -23,9 +23,36 @@ export function selectTarget(
...
+      ];
+      if (distinctDeviceIds.length > 1) {
+        throw new Error(
+          `No debugger target matches device_id "${deviceId}". ` +
+            `${distinctDeviceIds.length} devices are connected to Metro on port ${port}: ` +
+            `${distinctDeviceIds.join(", ")}. Pass one of these as device_id ` +
+            `(the logicalDeviceId returned by debugger-connect).`
+        );
+      }
+    }
```
hubgan: The thrown message lists only opaque logicalDeviceId hashes. A user with 2+ devices who passes an unmatched id e.g. an iOS UDID, which the debugger-connect doc string explicitly tells them to pass â gets back hashes they can't map to a physical device. selectTarget already has t.deviceName; emitting deviceName (logicalDeviceId) pairs would make re-targeting actionable. Empirically, the iPad-UDID throw returned 9e7844f7, 8a44101d with no hint which is the iPad.

hubgan (approved):

PR #399, opened 2026-06-23
fix(screenshot): wait for first frame so multi-sim captures don't fail (#391)

### LINKED ISSUE #391: screenshot: "no image to export" with >1 booted simulator (regression in 0.12.x; works in 0.11.0)
## Summary

`screenshot` fails with `no image to export` when **more than one simulator is booted**, even when a specific `--udid` is passed. This is a regression: it works in `0.11.0` and breaks in `0.12.1`.

## Environment

- argent `0.12.1` (broken) vs `0.11.0` (works)
- macOS 26.x, Apple Silicon
- Xcode iOS Simulators (iOS 26.x)

## Repro

1. Boot **two** iOS simulators:
   ```
   xcrun simctl boot <UDID_A>
   xcrun simctl boot <UDID_B>
   ```
2. Start the tool-server and capture a screenshot of a specific device:
   ```
   npx -y @swmansion/argent@0.12.1 server start --detach --force
   npx -y @swmansion/argent@0.12.1 run screenshot --udid <UDID_A> --out /tmp/s.png
   ```

### Actual (0.12.1)
```
[Tool:screenshot] Screenshot failed: no image to export.
```
No file is written. Happens for either booted UDID.

### Expected
A screenshot of the device named by `--udid` (the flag should disambiguate when multiple devices are booted).

## Isolation

- **`0.11.0` works** with the exact same two-booted-sim state and the same command → confirmed regression in `0.12.x`.
- **`0.12.1` works** if only a *single* simulator is booted.
- **Direct `xcrun simctl io <UDID_A> screenshot /tmp/s.png` works** in all cases (single or multiple booted), so the simulator display is capturable — the regression is in argent's device/display selection when multiple devices are booted.

## Impact

Breaks automated screenshot / visual-verify flows on any machine with more than one booted simulator (common during development and CI), despite passing an explicit `--udid`.

latekvo:
After more investigation i've found a much better fix for #391. 
I'm merging this either way, because it blocks 0.13.0 release and we already have this fix tested and approved.
By the next release, I want to revert this patch in favour of fixing core of the issue in simserver.

Regression introduced by https://github.com/software-mansion/radon/pull/101 adding the streaming feature to the argent simulator-server build, which made `/api/screenshot` serve the last streamed frame async instead of capturing on demand. 

<details>

## Summary

Fixes #391 — `screenshot` failing with `no image to export` when more than one simulator is booted, even with an explicit `--udid`. Regression in 0.12.x.

## Root cause

The simulator-server captures screenshots from its **live frame stream**, so a capture only succeeds once the first frame has arrived. Immediately after the server starts streaming, `POST /api/screenshot` replies `HTTP 200 { error: "no image to export" }` until that first frame lands (~0.5–1s). The tool-server requested the capture right after the server became ready, **losing that race**.

- With a single, foreground simulator the first frame arrives fast enough that the race is usually won — hiding the bug.
- With more than one simulator booted, the target is often backgrounded and its first frame is delayed, so the race is lost **reliably** → `no image to export`.

This became reachable when streaming was enabled for the argent simulator-server build in 0.12.x (radon [#101](https://github.com/software-mansion/radon/pull/101)); the 0.11.0 non-streaming build captured frames on demand, so it was immune. Device selection itself is correct — each device's stream captures the right device (verified: Pro Max → 1320×2868, iPhone 15 → 1179×2556).

## Fix

`httpScreenshot` now polls past that specific transient (every 250ms, up to 6s) instead of failing immediately, while still surfacing every other error (non-2xx, Android `wrong data size`, missing url/path) right away.

## Verification

End-to-end against the **shipped** simulator-server binary with **three** booted simulators, forcing a fresh sim-server spawn before each capture:

| | result |
|---|---|
| **unfixed** (current `dist`) | `no image to export` ~2/3 of fresh-spawn attempts |
| **fixed** | 7/7 success (~800ms), returns a correct PNG (1320×2868) |

- `npm test` (tool-server): **1199 passed**, including new retry / give-up cases for `httpScreenshot`
- `npx tsc --build`, `prettier --check`, `typecheck:tests`: clean

</details>

### DIFF
--- a/packages/tool-server/src/utils/simulator-client.ts
+++ b/packages/tool-server/src/utils/simulator-client.ts
@@ -2,8 +2,20 @@ import WebSocket from "ws";
 import { FAILURE_CODES, FailureError } from "@argent/registry";
 import type { SimulatorServerApi } from "../blueprints/simulator-server";
 import { toSimulatorNetworkError } from "./format-error";
+import { sleep } from "./timing";
 
 const DEFAULT_SCREENSHOT_SCALE = 0.3;
+
+// A simulator-server captures screenshots from its live frame stream, so the
+// first frame must have arrived before a capture can succeed. Right after the
+// server starts streaming it replies HTTP 200 `{ error: "no image to export" }`
+// until that first frame lands — typically ~0.5-1s, and reliably so for a
+// backgrounded simulator when more than one is booted (the regression in
+// https://github.com/software-mansion/argent/issues/391). Poll past that
+// transient instead of surfacing it as a hard failure.
+const NO_IMAGE_ERROR = /no image to export/i;
+const FIRST_FRAME_WAIT_MS = 6_000;
+const FIRST_FRAME_POLL_MS = 250;
 
 const connections = new Map<string, WebSocket>();
 let cmdId = 0;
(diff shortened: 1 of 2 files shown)


packages/tool-server/src/utils/simulator-client.ts
```diff
@@ -106,33 +118,51 @@ export async function httpScreenshot(
...
+
+    // Other HTTP-200 capture failures carry an `error` field rather than a
+    // non-2xx status (e.g. Android full-resolution requests that exceed what
+    // the emulator framebuffer can stream: "wrong data size, expected X got
+    // Y"). Surface that message instead of the misleading generic hint so the
+    // real cause is visible rather than sending callers to restart a perfectly
+    // healthy server.
     if (resBody.error) {
       throw new Error(`Screenshot failed: ${resBody.error}.`);
     }
```
hubgan, line 168: Nit: Abort mid-poll surfaces the wrong message. When signal aborts between polls, the guard `!signal?.aborted` is false, so control falls through and throws Screenshot failed: no image to export. rather than an abort/timeout error.

packages/tool-server/src/utils/simulator-client.ts
```diff
@@ -106,33 +118,51 @@ export async function httpScreenshot(
...
+    }
+
+    // Other HTTP-200 capture failures carry an `error` field rather than a
+    // non-2xx status (e.g. Android full-resolution requests that exceed what
+    // the emulator framebuffer can stream: "wrong data size, expected X got
+    // Y"). Surface that message instead of the misleading generic hint so the
+    // real cause is visible rather than sending callers to restart a perfectly
+    // healthy server.
     if (resBody.error) {
       throw new Error(`Screenshot failed: ${resBody.error}.`);
```
hubgan, line 167: Nit: This could be a `FailureError` instead of `Error`.

hubgan (approved):

PR #400, opened 2026-06-23
fix(telemetry): orchestator sub-tools invocations was missing ai_client and platform attributes

hubgan:
## Summary

Fixes telemetry attribution for tools dispatched by orchestrators. `run-sequence`,
`flow-execute`, and `flow-add-step` run their steps via `registry.invokeTool()`
directly, which minted a fresh, unregistered invocation id per nested call — so the
`ai_client` / `platform` the HTTP layer captured for the outer request never reached
the nested gestures, and they were recorded as **anonymous** in telemetry.
Remove `opt_out` event that is a leftover.

## Changes

- Add an optional `recordChildInvocation` recorder to `InvokeToolOptions` /
  `ToolContext`, bound by the HTTP layer to the outer request's attribution.
- New `invokeSubTool` helper: registers each sub-tool under a minted id inheriting
  the request's AI client, with the platform re-derived from the sub-tool's **own**
  device arg (`udid` / `device_id` / `avdName`); forwards the recorder so propagation
  survives arbitrary nesting (e.g. `flow-execute → run-sequence → gesture-tap`).
- Route the three orchestrators' sub-dispatch through
- Refactor platform inference in the HTTP layer into a shared `platformFromArgs`
  helper (also hardened to not throw on a malformed device arg).
- Remove opt_out event leftover

Why re-derive platform per child: an orchestrator like `flow-execute` has no platform
of its own and a single flow can target several devices, so the child's device arg is
the only correct platform source; the parent's platform is the fallback.

### DIFF
--- a/packages/telemetry/src/index.ts
+++ b/packages/telemetry/src/index.ts
@@ -11,7 +11,7 @@ import {
 import { sanitize } from "./sanitize.js";
 import { getBaseProps, type Runtime } from "./base-props.js";
 import { readOrCreateAnonId, peekAnonId } from "./identity.js";
-import { isEnabled as consentIsEnabled, writeConsentFlag, getConsentState } from "./consent.js";
+import { isEnabled, writeConsentFlag, getConsentState } from "./consent.js";
 import { emitDebugError, emitDebugPayload, isDebugEnabled } from "./debug.js";
 import { forget as forgetImpl, type ForgetOptions, type ForgetResult } from "./erasure.js";
 import type { EventName, EventPropertyMap } from "./events.js";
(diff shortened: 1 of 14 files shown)


latekvo (approved):

packages/telemetry/src/index.ts
```diff
@@ -165,48 +165,19 @@ export async function shutdown(timeoutMs = SHORT_FLUSH_TIMEOUT_MS): Promise<void
...
-  return consentIsEnabled();
-}
-
 /** Persist `enabled=true`. */
 export function markEnabled(): void {
   writeConsentFlag(true);
 }
 
-// Disable records one final opt-out event, persists the flag, then drains.
+// Disable persists the opt-out flag, then drains any already-queued events and
```
latekvo, line 173: This commit drops the `telemetry:opt_out` event and rewrites the opt-out path in `markDisabled` — a change to telemetry consent handling that is independent of this PR's stated purpose (propagating `ai_client`/`platform` attribution to orchestrator sub-tools). It is also an observable behaviour change: opting out no longer emits a final event, so the opt-out signal the backend previously received on a live opt-out is gone. Bundling this unrelated refactor and behaviour change under the attribution-fix PR couples two independent changes in one review and one squash-merged commit message.

PR #401, opened 2026-06-24
Add iOS device set support across Argent tooling

kmagiera:
## Summary
- Add `--ios-device-set` support to the CLI and propagate it through server startup
- Persist and reuse the configured CoreSimulator device set in the tools server environment and state
- Update iOS simulator and simctl call sites to honor the active device set
- Add coverage for flag parsing, environment normalization, and iOS-specific behavior

## Testing
- Added unit tests for CLI flag parsing and device-set path normalization
- Added tool-server tests for iOS boot and simulator-server wiring
- Not run (not requested)


filip131311:
I will move this into a new configuration aproach: https://github.com/software-mansion/argent/pull/542

PR #402, opened 2026-06-24
fix(vega): stop a VVD the `vega` CLI lost track of via the ps probe

latekvo:
## Problem

`vega virtual-device stop` exits non-zero with **"virtual device not running"** for any VVD the `vega` CLI is not tracking — and that is **every VVD argent boots**. `boot-device` starts a VVD via `vega virtual-device start -t N`, which returns once boot completes (rather than staying foreground), so the CLI never registers the device. `vega virtual-device status` then reports `running:false` and `stop` is a no-op-that-throws.

That error threw straight out of `stopVvd`, so **`boot-device {force:true}` failed outright and left the VVD running** — even though argent's *detection* (`isVvdRunning` / `listRunningVvdConsolePorts`) already deliberately trusts the OS process table over the flaky CLI.

Repro (against a VVD argent booted):
```
boot-device {vvdImage:"tv", force:true}
→ [Tool:boot-device] vega virtual-device stop failed: virtual device not running
     at stopVvd (vega-vvd.ts) → bootVegaImpl (boot-device.ts)
```

## Fix

Make **stop** symmetric with **detection**:
- Tolerate the CLI's stale failure (`vega virtual-device stop` is now best-effort).
- Terminate any VVD emulator process the `ps` probe still finds — SIGTERM, then SIGKILL the stragglers, verifying via `isVvdRunning` between.
- Killing by pid demands a **positive** emulator identity — the VVD binary name **and** a `-ports`/`-qmp` signal — so a process that merely mentions a `…/vega-virtual-device` path in its argv (e.g. a git command on a branch of that name) is never signalled.

Also stopped pointing users at `vega virtual-device stop` for an argent-booted VVD (the boot-device error message + the `argent-vega` skill); the reliable restart is `boot-device {force:true}`.

## Verification

- `npx vitest run --root packages/tool-server` → **1430 passed**. New unit tests cover `parseVvdPids` (identity/safety incl. the git-ref guard) and `stopVvd` (no-kill on clean stop; kill-without-rethrow when the CLI refuses; SIGTERM→SIGKILL escalation).
- **Live on a VVD**: reproduced the orphan state (argent boot → `status:false` while pid alive), then `boot-device force:true` — old pid gone, a fresh single VVD running, `booted:true`. Previously it threw.

### DIFF
--- a/packages/skills/skills/argent-vega/SKILL.md
+++ b/packages/skills/skills/argent-vega/SKILL.md
@@ -27,7 +27,7 @@ Then `describe` again to confirm. On a miss, run the loop again.
 
 - `list-devices` → Vega devices appear with a `serial` (use as `udid`) and a `vvdImage`. Start here to get both.
 - `boot-device {vvdImage}` — starts the single SDK-managed VVD (e.g. `vvdImage:"tv"`) and returns its `serial`. Skip if `list-devices` already shows a running device.
-- **Stopping the VVD** — run `vega virtual-device stop` in your shell. Caveat: the CLI tracks only VVDs it started in the foreground, so it often reports "virtual device not running" for a VVD that `boot-device` booted and stops nothing. To restart in that case use `boot-device {vvdImage, force:true}`, which stops the running VVD by process before re-booting.
+- **Stopping the VVD** — run `vega virtual-device stop` in your shell. Caveat: the CLI only tracks VVDs it starts in the foreground. As a result, it may report "virtual device not running" for a VVD started with `boot-device`, and won't stop it. To restart that VVD, run `boot-device {vvdImage, force: true}`. This stops the running VVD process before booting it again.
 
 ### App lifecycle
 
--- a/packages/skills/skills/argent-vega/SKILL.md
+++ b/packages/skills/skills/argent-vega/SKILL.md
@@ -27,7 +27,7 @@ Then `describe` again to confirm. On a miss, run the loop again.
 
 - `list-devices` → Vega devices appear with a `serial` (use as `udid`) and a `vvdImage`. Start here to get both.
 - `boot-device {vvdImage}` — starts the single SDK-managed VVD (e.g. `vvdImage:"tv"`) and returns its `serial`. Skip if `list-devices` already shows a running device.
-- **Stopping the VVD** — run `vega virtual-device stop` in your shell.
+- **Stopping the VVD** — run `vega virtual-device stop` in your shell. Caveat: the CLI tracks only VVDs it started in the foreground, so it often reports "virtual device not running" for a VVD that `boot-device` booted and stops nothing. To restart in that case use `boot-device {vvdImage, force:true}`, which stops the running VVD by process before re-booting.
 
 ### App lifecycle
 
--- a/packages/tool-server/src/utils/vega-vvd.ts
+++ b/packages/tool-server/src/utils/vega-vvd.ts
@@ -108,7 +108,11 @@ export async function stopVvd(
   // in this state, so a throwing stop must not abort the caller (e.g. a force reboot
   // in boot-device, which would otherwise fail outright and leave the VVD running).
   await runVega(["virtual-device", "stop"], { timeoutMs: options.timeoutMs ?? 60_000 }).catch(
-    () => {}
+    (err) => {
+      // Tolerated (the ps probe below tears the device down regardless), but logged so a
+      // genuine stop failure for a VVD the CLI *was* tracking is diagnosable, not silent.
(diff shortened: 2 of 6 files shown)


packages/skills/skills/argent-vega/SKILL.md
```diff
@@ -27,7 +27,7 @@ Then `describe` again to confirm. On a miss, run the loop again.
 
 - `list-devices` → Vega devices appear with a `serial` (use as `udid`) and a `vvdImage`. Start here to get both.
 - `boot-device {vvdImage}` — starts the single SDK-managed VVD (e.g. `vvdImage:"tv"`) and returns its `serial`. Skip if `list-devices` already shows a running device.
-- **Stopping the VVD** — run `vega virtual-device stop` in your shell.
+- **Stopping the VVD** — run `vega virtual-device stop` in your shell. Caveat: the CLI tracks only VVDs it started in the foreground, so it often reports "virtual device not running" for a VVD that `boot-device` booted and stops nothing. To restart in that case use `boot-device {vvdImage, force:true}`, which stops the running VVD by process before re-booting.
```
hubgan: Small grammar fix:
```
Caveat: the CLI only tracks VVDs it starts in the foreground. As a result, it may report “virtual device not running” for a VVD started with `boot-device`, and won’t stop it. To restart that VVD, run `boot-device { vvdImage, force: true }`. This stops the running VVD process before booting it again.
```

packages/tool-server/src/utils/vega-vvd.ts
```diff
@@ -94,8 +94,53 @@ export async function startVvd(params: {
...
+  const pids = await listRunningVvdPids();
+  if (pids.length === 0) return;
+  for (const pid of pids) signalQuietly(pid, "SIGTERM");
+  // Give SIGTERM a chance to bring the emulator down cleanly, then escalate.
+  const deadline = Date.now() + graceMs;
+  while (Date.now() < deadline) {
+    if (!(await isVvdRunning())) return;
+    await new Promise((r) => setTimeout(r, pollMs));
+  }
+  for (const pid of await listRunningVvdPids()) signalQuietly(pid, "SIGKILL");
```
hubgan: The SIGKILL pass returns the moment it finishes signalling, with no confirmation the process is actually gone. The very next thing the force-reboot path does is `startVvd()` → `waitForVvdRunning()`/`resolveRunningVvdSerial()`; if a just-SIGKILL'd qemu is still listed in the process table for a beat, that reads as a *second* VVD and can trip `MultipleVegaDevicesError`. Low probability (orphaned qemu reparents to launchd and reaps fast, and boot takes seconds), but cheap to close — mirror the SIGTERM grace poll after the kill

packages/tool-server/src/utils/vega-vvd.ts
```diff
@@ -94,8 +94,53 @@ export async function startVvd(params: {
...
+  // Graceful first — ask the CLI to stop the VVD it tracks — but best-effort: when
+  // the CLI has lost track of a running VVD it exits non-zero with "virtual device
+  // not running" (the same staleness that makes `vega virtual-device status`
+  // misreport). An argent-booted VVD — started via `vega virtual-device start -t N`,
+  // which returns once boot completes rather than staying foreground — is routinely
+  // in this state, so a throwing stop must not abort the caller (e.g. a force reboot
+  // in boot-device, which would otherwise fail outright and leave the VVD running).
+  await runVega(["virtual-device", "stop"], { timeoutMs: options.timeoutMs ?? 60_000 }).catch(
+    () => {}
+  );
```
hubgan, line 116: Swallowing the CLI failure is the right call for the orphan case, but a fully silent `.catch(() => {})` also hides a *genuine* `vega virtual-device stop` failure (a real foreground VVD the CLI was tracking but couldn't stop). The ps probe still tears the device down, so behaviour is fine — but a one-line debug/stderr note would make a real failure diagnosable instead of invisible:

packages/tool-server/src/utils/vega-vvd.ts
```diff
@@ -94,8 +94,53 @@ export async function startVvd(params: {
...
+
+// `process.kill` throws ESRCH if the pid already exited and EPERM if it isn't ours;
+// either way there's nothing left to do for that pid, so swallow it.
+function signalQuietly(pid: number, signal: NodeJS.Signals): void {
+  try {
+    process.kill(pid, signal);
+  } catch {
+    /* already gone or not ours */
+  }
 }
```
hubgan, line 162: The comment scopes this to ESRCH/EPERM, but the bare `catch {}` swallows everything (EINVAL, a non-Error throw, etc.). Harmless given the inputs here (pid is always a parsed positive int, signal is a literal), but narrowing it keeps the catch honest to the comment:

hubgan (approved):

PR #407, opened 2026-06-24
fix(boot-device): support Xcode 27 Device Hub fallback and SimulatorKit symlink

### LINKED ISSUE #406: boot-device fails on Xcode 27 (replaces Simulator.app with Device Hub)
## Problem

`boot-device` hardcodes `open -a Simulator.app` and `defaults write com.apple.iphonesimulator`, both of which fail on Xcode 27 beta because:

1. Xcode 27 **removes Simulator.app** entirely — it is replaced by Device Hub.app (`com.apple.dt.Devices`). `open -a Simulator.app` errors out.
2. Xcode 27 no longer reads the `com.apple.iphonesimulator` preferences domain. `defaults write` there is a no-op.
3. **SimulatorKit.framework** moved from `/Applications/Xcode.app/Contents/Developer/Library/PrivateFrameworks/` to `/Applications/Xcode.app/Contents/SharedFrameworks/`. The prebuilt `simulator-server` binary crashes at launch because it looks for the framework at the old path.

## Steps to reproduce

1. Install Xcode 27 beta (`/Applications/Xcode-beta.app/`)
2. Run `boot-device` with any iOS simulator UDID
3. Observe: Simulator.app not found → tool falls through with no UI opened; `defaults write` silently fails; `simulator-server` crashes on SimulatorKit resolution

## Additional context

Only Xcode 27+ is affected. Xcode 26.x and earlier work fine. The `ax-service` binary also crashes on iOS 27.0 simulators with `-[SimDevice booted]: unrecognized selector` (separate issue, not covered here).

## Suggested fix direction

- Fall back from `open -a Simulator.app` to `open -b com.apple.dt.Devices` when Simulator.app is absent
- Wrap `defaults write com.apple.iphonesimulator` in a try/catch — it is cosmetic and non-essential
- Auto-create a symlink from the old SimulatorKit path to the new location

etnikbeqiri1:
## Problem

`boot-device` hardcodes `open -a Simulator.app` and `defaults write com.apple.iphonesimulator`, both of which fail on Xcode 27 beta because:

1. **Simulator.app is removed** — replaced by Device Hub.app (`com.apple.dt.Devices`). `open -a Simulator.app` errors out.
2. **Preferences domain changed** — Xcode 27 no longer reads `com.apple.iphonesimulator`. `defaults write` there is a no-op.
3. **SimulatorKit.framework moved** — from `Developer/Library/PrivateFrameworks` to `SharedFrameworks`. The prebuilt `simulator-server` binary crashes at launch.

## Fix

Three changes in `bootIos()`:

- **Device Hub fallback**: try `open -a Simulator.app` first (backward compat), fall back to `open -b com.apple.dt.Devices` for Xcode 27+
- **Graceful defaults write**: wrap `defaults write com.apple.iphonesimulator` in try/catch since Xcode 27 ignores that domain
- **SimulatorKit symlink**: new `ensureSimulatorKitPath()` helper auto-creates a symlink from the old PrivateFrameworks path to SharedFrameworks so the prebuilt `simulator-server` binary works unmodified

## Verification

- `boot-device` with an iOS 27.0 UDID succeeds on Xcode 27 beta (Mac mini M4): Device Hub opens, simulator boots, `describe` / `gesture-tap` / `screenshot` all work end-to-end
- `boot-device` still succeeds on Xcode 26.x (Simulator.app path unchanged)
- `tsc --noEmit` clean across the changed file

## Not addressed here

- `ax-service` (compiled ObjC binary) crashes with `-[SimDevice booted]: unrecognized selector` on iOS 27.0 SDK — needs recompile from source by upstream

Fixes #406

### DIFF
--- a/packages/tool-server/src/tools/devices/boot-device.ts
+++ b/packages/tool-server/src/tools/devices/boot-device.ts
@@ -567,13 +567,18 @@ async function bootIos(
     "com.apple.iphonesimulator",
     "CurrentDeviceUDID",
     udid,
-  ]);
+  ]).catch(() => {});
   // `simctl boot` above already booted the device CORE (headless). Opening
   // Simulator.app only attaches the GUI window — surfaces that stream the
   // device through simulator-server (e.g. Argent Lens) don't need it, so
   // `headless` skips it to avoid popping a window the user didn't ask for.
   if (!headless) {
-    await execFileAsync("open", ["-a", "Simulator.app"]);
+    // Xcode 27 drops Simulator.app in favour of Device Hub.app
+    // (com.apple.dt.Devices); fall back to it, and keep the GUI attach
+    // best-effort so a missing app never fails a boot whose core is already up.
+    await execFileAsync("open", ["-a", "Simulator.app"])
+      .catch(() => execFileAsync("open", ["-b", "com.apple.dt.Devices"]))
+      .catch(() => {});
   }
   return { platform: "ios", udid, booted: true };
 }


latekvo:
Nice catch, thanks for taking your time to fix this :)

latekvo:
Apologies for the force-push, i'm developing a little review orchestration and messed up the system prompt 😅
Installing Xcode 27 on my end and we'll be merging this as soon as I'm able to test it on my end
Again thanks for the effort

etnikbeqiri1:
No worries, glad it's fixed. I've had my fair share of issues getting newer macOS/Xcode setups working properly too

Thanks for the update and for taking the time to review it.

latekvo (changes requested):
Hey so tested this, and it appears to be broken on SIP-enabled Mac. Apologies but can't really accept this PR in the current state :/

I'm happy to merge this if we only keep `open -b com.apple.dt.Devices` fix + the `.catch()` you added, but the `ensureSimulatorKitPath` has to be removed.

<details>

**Test host:** macOS 26.5 (25F71), Apple Silicon (arm64), Xcode-beta 27.0 (27A5209h) at `/Applications/Xcode-beta.app`, SIP **enabled** (the default). iOS 27.0 simulator (iPhone 17 Pro).

### Baseline reproduced ✅
With no symlink present, `simulator-server` crashes at launch and the tap tool fails end-to-end:
```
simulator-server[…] Error: SimulatorKit could not be loaded (Xcode search path: /Applications/Xcode-beta.app)
simulator-server[…] Error: Failed to initialize SimulatorControl
Error: Failed to initialize iOS simulator bridge
```
```
POST /tools/gesture-tap →
{"error":"[Tool:gesture-tap] Service dependency failed: [SimulatorServer:…] simulator-server exited with code before becoming ready"}
```
So the diagnosis in the PR description is spot-on.

### The fix doesn't restore clicking on this machine ❌
`ensureSimulatorKitPath()` tries to `mkdirSync` + `symlinkSync` **inside the Xcode bundle** (`…/Xcode-beta.app/Contents/Developer/Library/PrivateFrameworks/`). That path is protected by **SIP**, so the write fails with `EPERM` **even as root** — and the argent tool-server runs as root by default:
```
$ sudo mkdir -p "/Applications/Xcode-beta.app/Contents/Developer/Library/PrivateFrameworks"
mkdir: …/PrivateFrameworks: Operation not permitted        # rc=1, running as root, SIP on
```
And driving it through argent (tool-server → boot-device):
```
POST /tools/boot-device →
{"error":"[Tool:boot-device] EPERM: operation not permitted, mkdir '/Applications/Xcode-beta.app/Contents/Developer/Library/PrivateFrameworks'"}
```
Because the `mkdirSync`/`symlinkSync` in `ensureSimulatorKitPath()` (and the `await ensureSimulatorKitPath()` call in `bootIos`) aren't wrapped in try/catch, this is also a **regression**: on SIP-enabled Xcode 27 `boot-device` now hard-fails with EPERM *before it boots*, whereas previously it at least booted the sim.

I also confirmed the runtime can't be worked around with env vars — `simulator-server` is signed with hardened runtime, so `DYLD_FRAMEWORK_PATH` / `DYLD_FALLBACK_FRAMEWORK_PATH` pointing at `SharedFrameworks` are ignored; it still prints "SimulatorKit could not be loaded". So the symlink is the *only* lever this prebuilt binary exposes, and SIP takes that lever away.

### Suggested direction
Since the write target is SIP-protected and hardened-runtime blocks DYLD injection, I don't think there's a viable host-side symlink/env approach. The robust fix is to have `simulator-server` itself locate `SimulatorKit.framework` in `Contents/SharedFrameworks` on Xcode 27+ (same "needs an upstream recompile" bucket you already noted for `ax-service`). Happy to help test a new binary once it exists.

The **Device Hub fallback** (`open -b com.apple.dt.Devices`) and the **graceful `defaults write`** changes look good and are worth keeping. It's specifically the SimulatorKit symlink I'd hold on.

(Separately, the PR currently shows a merge conflict against `main`.)

</details>

packages/tool-server/src/tools/devices/boot-device.ts
```diff
@@ -448,6 +472,7 @@ async function bootIos(
     );
   }
   await ensureDep("xcrun");
+  await ensureSimulatorKitPath();
```
latekvo: As per my prev comment, if you remove this we'll be merging this PR.

packages/tool-server/src/tools/devices/boot-device.ts
```diff
@@ -501,8 +526,10 @@ async function bootIos(
     "com.apple.iphonesimulator",
     "CurrentDeviceUDID",
     udid,
-  ]);
-  await execFileAsync("open", ["-a", "Simulator.app"]);
+  ]).catch(() => {});
+  await execFileAsync("open", ["-a", "Simulator.app"])
+    .catch(() => execFileAsync("open", ["-b", "com.apple.dt.Devices"]))
+    .catch(() => {});
```
latekvo: This is one of the fixes we have to make either way for XCode 27, so we'll be happy to merge this part.

etnikbeqiri1:
Pushed the requested change.

`ensureSimulatorKitPath` is gone. I rebased the branch onto current `main` and kept only the two parts you signed off on: the Device Hub fallback (`open -b com.apple.dt.Devices`) and the `.catch()` on the `defaults write`. So there's no SIP-protected write anymore, and no EPERM regression on the boot path. The rebase also cleared the conflict against `main` — it's a single commit / one-file diff now.

I ran the `tool-server` suite locally against the rebased `main` and it's green.

Should be ready for another look whenever you have a moment. Thanks for testing it on a SIP-enabled machine and writing up the repro in that much detail — that was a genuinely useful catch.

latekvo (approved):
Thx!

PR #409, opened 2026-06-24
fix(argent-lens): native window controls, comment cursor, card layout polish

latekvo:
Part 1 out of N reworks for Argent Lens

Polish pass on the Argent Lens preview window.

## Changes
- **Native window controls.** `titleBarStyle: "hidden"` (+ `setWindowButtonVisibility`) shows the real macOS traffic lights on the transparent window instead of the handcrafted close button. `trafficLightPosition` insets them to share the corner panels' 14px padding.
- **Comment-mode cursor.** Entering comment mode swaps the crosshair for a Figma-style speech-bubble cursor over the stream, so "you are placing a comment" is unmistakable. Defined as a `--cursor-comment` data-URI token in `theme.css`.
- **Bottom toolbelt.** The "Add comment" button moved out of the top-right cluster into a new bottom-centre floating toolbelt — the seed of a Figma-style bottom tool tab. Same button look, placement only.
- **Cards show all variants.** Two fixes so short cards no longer clip a variant: the left column is no longer height-limited by the bottom-right footbar (per-column bound), and tighter variant-row padding lets a 2-variant card fit its column.
- **Cards repel / never overlap.** A deterministic per-poll de-overlap of the spring **home** targets (a stable top-down sweep), not a per-frame force. The damped springs glide to non-overlapping rests: no micro-jitter, no rubber-banding, and a card wedged between two settles instead of wiggling forever.
- **App named "Argent Lens".** The window title was already set, but the macOS menu bar / Cmd-Tab / Dock showed "Electron" (the unpackaged `Electron.app` bundle's `CFBundleName`, which `app.setName` can't override). On macOS the tool-server now launches the window through a thin **wrapper `.app`** whose `Info.plist` `CFBundleName` is "Argent Lens" — it symlinks Electron's Frameworks/Resources (no ~270MB copy) and is launched with `--no-sandbox` (the modified/symlinked bundle no longer matches the signed Helper apps). Acceptable for this window: it only loads the tool-server's localhost UI, and the renderer keeps `contextIsolation` + `sandbox:true`. Falls back to plain Electron on non-macOS or any failure.

## Verification
Driven live against the iOS simulator via a dev tool-server + the Electron window over CDP:
- Native traffic lights render on the transparent window; handcrafted close button removed.
- Comment-mode `.stage img` computed cursor is the speech-bubble data-URI.
- "Add comment" is in `#float-toolbelt`, horizontally centred at the window mid-line, 14px above the bottom.
- All demo cards report `bodyScrollHeight === clientHeight` (zero clip); both variants visible.
- 5- and 9-card layouts: zero overlaps, positions identical across 2s (no jitter), and pinning a card between two settles with no overlap and no wiggle.
- Menu bar / Cmd-Tab read "Argent Lens"; `ps` confirms the window launches via `…/argent-lens-app/Argent Lens.app/Contents/MacOS/Electron --no-sandbox`.
- `tsc --build`, `eslint`, `prettier --check`, and the preview/variant test files all pass.

### DIFF
--- a/packages/tool-server/src/utils/preview-window.ts
+++ b/packages/tool-server/src/utils/preview-window.ts
@@ -1,6 +1,80 @@
-import { spawn, type ChildProcess } from "node:child_process";
+import { execFileSync, spawn, type ChildProcess } from "node:child_process";
 import * as fs from "node:fs";
+import * as os from "node:os";
 import * as path from "node:path";
+
+/**
+ * macOS only. Build (and cache) a thin `.app` wrapper around the installed
+ * Electron.app whose Info.plist `CFBundleName` is "Argent Lens", so the OS names
+ * the window — menu bar, Cmd-Tab, Dock — "Argent Lens" instead of "Electron"
+ * (the framework default, which `app.setName()` at runtime cannot override).
+ *
+ * The wrapper SYMLINKS Electron's heavy Frameworks/Resources — there is no
+ * ~270MB copy — and supplies only its own Info.plist + a symlinked executable.
+ * The catch: a symlinked bundle with a modified Info.plist no longer matches the
+ * signed Helper apps inside Frameworks, so the OS sandbox cannot initialise and
+ * the helper processes crash unless the app is launched with `--no-sandbox`
+ * (the caller adds it). That is an acceptable trade-off for THIS window: it only
+ * ever loads the tool-server's own localhost preview UI — never untrusted or
+ * remote content — and the renderer still runs with contextIsolation +
+ * sandbox:true at the Electron level. The full alternative (a renamed, deeply
+ * re-signed copy of Electron.app, à la @electron/packager) is what avoids
+ * `--no-sandbox`, at the cost of that ~270MB copy.
+ *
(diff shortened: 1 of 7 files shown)


packages/tool-server/src/utils/preview-window.ts
```diff
@@ -1,7 +1,81 @@
...
+      "Contents"
+    );
+    const wrapperExec = path.join(wrapperContents, "MacOS", "Electron");
+    // Reuse the cached wrapper unless it's missing or points at a different
+    // Electron install (e.g. an upgrade moved the binary).
+    try {
+      if (fs.realpathSync(wrapperExec) === fs.realpathSync(electronBin)) return wrapperExec;
+    } catch {
+      /* not built yet — fall through and build it */
+    }
```
hubgan, line 46: Low risk. The wrapper builds into a shared $TMPDIR/argent-lens-app. Concurrent first-builds from two tool-servers could race on rmSync+rebuild; the realpath reuse-check makes this rare and content is deterministic.

hubgan (approved):

PR #411, opened 2026-06-24
fix(tool-server): authenticate Chromium-control WebSocket upgrades

latekvo:
## Problem

The per-device Chromium control WebSocket (`/chromium-server/<id>/ws`) accepted **any** upgrade handshake. The `ws` library owns the HTTP `upgrade` event, so neither protection that guards every HTTP route ran on it — the **Host-header guard** (DNS-rebinding defense) and the **Bearer-token auth gate** are both registered as Express `request`-event middleware and never fire on an `upgrade`.

That channel dispatches `touch` / `key` / `button` / `wheel` / `rotate` / `clipboardSync` straight onto the live `ChromiumServer`. So while a Chromium/Electron session is warm (the normal state when an agent is driving one), **any web page the operator visits could connect cross-origin and inject synthetic input**:

```js
new WebSocket("ws://127.0.0.1:<port>/chromium-server/chromium-cdp-9222/ws")
  .onopen = e => e.target.send(JSON.stringify({ cmd: "key", direction: "Down", text: "…" }));
```

WebSocket handshakes aren't subject to the Same-Origin Policy and no `Origin` was checked — cross-site WebSocket hijacking (CSWSH). An attacker can't bootstrap a session (that still needs the authenticated REST path), only ride one already open.

## Fix

The browser-loaded preview UI is the legitimate consumer and **can't set an `Authorization` header on a WebSocket** — which is why the whole `/preview` surface is intentionally token-exempt. So this mirrors the existing **origin/host-based** defense rather than adding token auth, applied in the upgrade handler:

- **Host guard** — reuse the same loopback/bind-host allow-list the HTTP requests use, closing the DNS-rebinding bypass on the upgrade path.
- **Origin guard** — browsers always send `Origin` on a WS handshake; accept only our own loopback/bind-host origin. A non-browser client (Node `ws`) sends no `Origin` and is still constrained by the Host guard. Opaque / `null` origins are rejected.

A wildcard bind (`--host 0.0.0.0`) disables the guard, matching the existing HTTP Host-guard behavior. The check runs before `resolveServer`, so a cross-origin probe can't even learn whether a device exists. The legitimate same-origin preview UI and Node clients are unaffected.

## Tests

New `chromium-ws-upgrade-auth.test.ts`:
- unit-tests `isWebsocketUpgradeAllowed` across loopback / localhost / `::1`, no-Origin (Node client), cross-origin (rejected), `null` origin, DNS-rebinding Host, missing Host, explicit bind host, and wildcard-disabled;
- wiring-tests that `attachChromiumServerWebsocket` destroys the socket and never reaches `resolveServer` on a cross-origin upgrade, and passes an allowed one through.

Verified locally on `origin/main`: `tsc --build` clean, full tool-server suite **1573 passing**, `typecheck:tests` clean, ESLint clean, Prettier clean.

### DIFF
--- a/packages/tool-server/src/http.ts
+++ b/packages/tool-server/src/http.ts
@@ -232,6 +232,44 @@ function extractHostname(host: string): string {
   return colon === -1 ? host : host.slice(0, colon);
 }
 
+// Hostname of an `Origin` request header ("http://127.0.0.1:5173" → "127.0.0.1").
+// Opaque origins (the literal "null" from sandboxed iframes / file:// pages) and
+// malformed values return null so they fail the allow-list check.
+function hostnameFromOrigin(origin: string): string | null {
+  const schemeEnd = origin.indexOf("://");
+  if (schemeEnd === -1) return null;
+  return extractHostname(origin.slice(schemeEnd + 3));
+}
+
+// Authorize a WebSocket upgrade for the browser-facing control channels (e.g.
+// the Chromium-control WS). A WS handshake bypasses the Express Host/auth
+// middleware — the `ws` library owns the upgrade — so the same defenses are
+// re-applied here. The in-process preview UI is browser-loaded and can't carry
+// a Bearer token (like the rest of the /preview surface), so the guard is
+// origin/host-based rather than token-based:
+//   • Host guard — same loopback/bind-host allow-list as HTTP requests, closing
+//     the DNS-rebinding bypass.
+//   • Origin guard (anti-CSWSH) — browsers always send `Origin` on a WS
+//     handshake; accept only our own loopback/bind-host origin. A non-browser
+//     client (Node `ws`) sends no Origin and is already constrained by the Host
+//     guard above.
+// A wildcard bind disables the guard (operator opted into network exposure),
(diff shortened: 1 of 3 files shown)


packages/tool-server/src/http.ts
```diff
@@ -232,6 +232,44 @@ function extractHostname(host: string): string {
...
+  const host = headers.host;
+  if (!host || !policy.allowedHostnames.has(extractHostname(host))) return false;
+  const origin = headers.origin;
+  if (origin !== undefined) {
+    const originHost = hostnameFromOrigin(origin);
+    if (!originHost || !policy.allowedHostnames.has(originHost)) return false;
+  }
+  return true;
+}
+
```
hubgan, line 272: the Origin guard matches the loopback family (any scheme/port), so another local dev-server page would also pass — not the targeted threat (remote evil.com is blocked) and consistent with the existing HTTP Host guard, but the comment `"accept only our own…origin"` slightly understates it. Plus a harmless empty-Host strictness asymmetry vs the HTTP guard.

hubgan (approved):

PR #414, opened 2026-06-25
fix(describe): surface React Native Web / Expo-web content on Chromium

latekvo:
## Problem

`describe` on a Chromium device (the `cdp-dom` walker) returned an essentially **empty tree** for React Native Web / Expo-web apps — just `html > body > div > div` — even though the page was fully rendered and interactive.

## Root cause

The walker pruned any element whose `getBoundingClientRect()` is **zero-area**, cutting its **entire subtree**. That is wrong whenever a zero-area element still paints visible descendants. Verified against `react-native-gesture-handler`'s `expo-example` on web (488 elements, depth 28):

1. **`display:contents` wrappers.** RNW nests content under `<div style="display:contents">` → no box → `0×0` rect → pruned (first one ~16 levels deep, above all content). Not the depth cap alone (disproven) and not the node budget (`truncated:false`).
2. **`MAX_DEPTH = 24`** — RNW leaf text sits ~25 levels deep; `MAX_NODES` (5000), not depth, bounds the CDP payload.
3. **Other `0×0`-but-visible cases** — zero-height wrapper of an `position:absolute` dropdown/popover/portal, `width:0;height:0;overflow:visible` wrapper, float-collapsed wrapper. Discriminator vs. a *correctly* pruned box (collapsed `overflow:hidden`): **does it clip overflow**.
4. **Invisible box-less content** — a box-less element whose content paints nothing (e.g. a `transform: scale(0)` subtree) must still be dropped.
5. **Crash on DOM-clobbering pages** — a `<form>` control named `title`/`id` shadows `el.title`/`el.id` to a DOM node; `el.title.slice(...)` threw (Wikipedia) and a clobbered `el.id` would break `JSON.stringify`. Pre-existing; describe failed entirely on such pages.

## Fix

- Prune a node only when truly invisible (`display:none`/`visibility:hidden`/`opacity:0`) **or** its border box is zero-area **and it clips overflow**. Otherwise traverse and promote its content.
- A box-less element's frame is its child-spanning union; with no surviving child, fall back to the **painted extent of its own inline content via a `Range`**. Still zero-area ⇒ invisible ⇒ dropped (handles `scale(0)`; also gives box-less text leaves a real frame instead of `0×0`).
- `MAX_DEPTH` 24 → 60 (pure recursion bound; `MAX_NODES` is the payload guard).
- Read `title`/`id` via `getAttribute` (clobber-proof, always `string | null`).
- `getComputedStyle` read once per node.

All other behavior (hidden-subtree pruning, structural single-`div` collapse, interactivity, shadow-DOM / iframe piercing) unchanged.

## Verification

Driven through the **actual built tool-server's** `describe` over HTTP, plus the real injected script against live pages.

**RNGH `expo-example` (web):** `7 → 295` nodes; `html>body>div>div` → full gesture list with `[clickable]` buttons + labels (Fling/Tap/Pinch/…) + normalized frames.

**Adversarial fixture — 15/15 correct:**

| case | result |
|---|---|
| display:contents / nested contents / contents-link | ✓ shown (link gets a real frame) |
| abs-positioned child in `height:0` wrapper | ✓ shown |
| overflowing child of `0×0` box / float-collapsed wrapper | ✓ shown |
| `position:fixed` child / sr-only (`1×1`) / `display:contents` button | ✓ shown |
| `0×0; overflow:hidden` (clipped) / `overflow:clip` / `display:none` child / mixed-overflow | ✓ stays pruned |
| `transform: scale(0)` content | ✓ pruned (invisible) |

**Real pages (old vs new script):**

| page | old | new | regression |
|---|---|---|---|
| github.com | 150 nodes | 961 nodes, not truncated | none — 0 dropped texts |
| example.com | 7 | 7 (identical) | none |
| en.wikipedia.org | **crash** (`el.title.slice`) | 761-node tree | **fixed** |

**New automated guard:** `test/describe-chromium-script.test.ts` evals the real `DESCRIBE_DOM_SCRIPT` against a mock DOM, locking in: contents traversal, zero-height-dropdown, clipped-stays-pruned, `scale(0)` invisibility, box-less painted-text, and the clobbering crash. (The script runs in the renderer; the rest of the suite can only mock the CDP response, and the repo has no headless-browser test infra.)

`npm run build` ✓ · `typecheck:tests` ✓ · prettier ✓ · eslint ✓ · **full suite 1580 tests pass**.

### DIFF
--- a/packages/tool-server/src/tools/describe/platforms/chromium.ts
+++ b/packages/tool-server/src/tools/describe/platforms/chromium.ts
@@ -16,14 +16,16 @@ import type { DescribeNode, DescribeTreeData } from "../contract";
  *    tree stays small.
  *  - Treat anchors, buttons, inputs, [role=button], [onclick], [tabindex]≥0
  *    as `clickable: true` so the agent knows which nodes to tap.
- *  - Prune a node only when it is truly invisible (display:none, visibility:hidden,
- *    opacity:0) or its border box is zero-area AND it clips overflow. A zero-area
- *    box with the default overflow:visible still paints its descendants — so we
- *    traverse it and promote them instead of cutting the subtree. This covers
- *    display:contents wrappers (React Native Web nests content under them), the
- *    zero-height anchor of an absolutely-positioned dropdown/popover/portal, and
- *    float/overflow wrappers that collapse to a zero-height box. A collapsed
- *    overflow:hidden container genuinely hides its content, so it stays pruned.
+ *  - Prune a node only when it is truly invisible (display:none, opacity:0) or its
+ *    border box is zero-area AND it clips overflow. A zero-area box with the default
+ *    overflow:visible still paints its descendants — so we traverse it and promote them
+ *    instead of cutting the subtree. This covers display:contents wrappers (React Native
+ *    Web nests content under them), the zero-height anchor of an absolutely-positioned
+ *    dropdown/popover/portal, and float/overflow wrappers that collapse to a zero-height
+ *    box. A collapsed overflow:hidden container genuinely hides its content, so it stays
+ *    pruned. visibility:hidden is NOT hard-pruned (it inherits, but a descendant can
+ *    override it back to visible): we descend and suppress only the hidden element's own
+ *    paint, so a visibility:visible descendant still surfaces.
  *  - Cap node count at 5000 — that, not depth, bounds the payload a runaway SPA
  *    would otherwise serialize past CDP's single Runtime.evaluate reply limit
  *    (~50MB). Cap depth at 60 purely to bound recursion: modern React DOMs
@@ -49,15 +51,26 @@ export const DESCRIBE_DOM_SCRIPT = `(() => {
   // walk, the exact crash class this walker must avoid. A prototype accessor is never
   // shadowed by a form's named properties, so EVERY inherited member read on a possibly-
   // clobbered element (methods and getters alike) goes through these via .call(el).
-  const getChildNodes = Object.getOwnPropertyDescriptor(Node.prototype, "childNodes").get;
-  const getTextContent = Object.getOwnPropertyDescriptor(Node.prototype, "textContent").get;
-  const getTagName = Object.getOwnPropertyDescriptor(Element.prototype, "tagName").get;
-  const getChildrenEls = Object.getOwnPropertyDescriptor(Element.prototype, "children").get;
-  const getShadowRoot = Object.getOwnPropertyDescriptor(Element.prototype, "shadowRoot").get;
-  const getScrollHeight = Object.getOwnPropertyDescriptor(Element.prototype, "scrollHeight").get;
-  const getClientHeight = Object.getOwnPropertyDescriptor(Element.prototype, "clientHeight").get;
-  const getScrollWidth = Object.getOwnPropertyDescriptor(Element.prototype, "scrollWidth").get;
-  const getClientWidth = Object.getOwnPropertyDescriptor(Element.prototype, "clientWidth").get;
+  //
+  // protoGetter reads the accessor's getter defensively: if the descriptor is ever
+  // absent (a page can delete/redefine a DOM prototype member), it falls back to a
+  // direct property read so one missing accessor degrades a single node instead of a
+  // \`.get\` on \`undefined\` throwing at script top and aborting the entire describe.
+  function protoGetter(proto, prop) {
+    const d = Object.getOwnPropertyDescriptor(proto, prop);
+    return d && d.get ? d.get : function () { return this[prop]; };
+  }
+  const getChildNodes = protoGetter(Node.prototype, "childNodes");
+  const getNodeType = protoGetter(Node.prototype, "nodeType");
+  const getNodeValue = protoGetter(Node.prototype, "nodeValue");
+  const getTextContent = protoGetter(Node.prototype, "textContent");
+  const getTagName = protoGetter(Element.prototype, "tagName");
+  const getChildrenEls = protoGetter(Element.prototype, "children");
+  const getShadowRoot = protoGetter(Element.prototype, "shadowRoot");
+  const getScrollHeight = protoGetter(Element.prototype, "scrollHeight");
+  const getClientHeight = protoGetter(Element.prototype, "clientHeight");
+  const getScrollWidth = protoGetter(Element.prototype, "scrollWidth");
+  const getClientWidth = protoGetter(Element.prototype, "clientWidth");
   const getAttr = Element.prototype.getAttribute;
   const hasAttr = Element.prototype.hasAttribute;
   const getBCR = Element.prototype.getBoundingClientRect;
@@ -101,7 +114,12 @@ export const DESCRIBE_DOM_SCRIPT = `(() => {
   function ownText(el) {
     let s = "";
     for (const child of getChildNodes.call(el)) {
-      if (child.nodeType === 3) s += child.nodeValue;
+      // nodeType/nodeValue via the prototype getters, keeping the "every inherited
+      // read goes through a captured getter" invariant whole: a child can be a
+      // clobbering <form> whose named control shadows these to a control element.
+      // (=== 3 already made a clobbered nodeType safe; this removes the lone
+      // directly-read member so no read's safety rests on the comparison semantics.)
+      if (getNodeType.call(child) === 3) s += getNodeValue.call(child);
     }
     return s.replace(/\\s+/g, " ").trim();
   }
@@ -165,11 +183,14 @@ export const DESCRIBE_DOM_SCRIPT = `(() => {
     return normRect(getBCR.call(el));
   }
 
-  // Painted extent of an element's own inline content. A box-less element
-  // (display:contents, or a zero-width box whose text overflows) has a 0x0 border box
-  // yet its text still paints; a Range measures that. Crucially it returns 0x0 when an
-  // ancestor transform (e.g. scale(0)) collapses the paint, which is how we tell
-  // genuinely-invisible content apart from merely box-less content.
+  // Painted extent of an element's own inline TEXT. A box-less element (display:contents,
+  // or a zero-width box whose text overflows) has a 0x0 border box yet its text still
+  // paints; a Range measures that. It returns 0x0 when an ancestor transform (e.g.
+  // scale(0)) or display:none collapses the paint. It does NOT distinguish a
+  // visibility:hidden / opacity:0 *descendant* — those keep their layout box, so the
+  // Range is non-zero even though nothing paints — which is why walk() consults this only
+  // when the element has its own text, never to resurrect a wrapper whose element
+  // children were all pruned as invisible.
   function contentFrame(el) {
     try {
       const range = document.createRange();
@@ -196,15 +217,17 @@ export const DESCRIBE_DOM_SCRIPT = `(() => {
   }
 
   function hidden(el, style) {
-    if (style.visibility === "hidden" || style.display === "none") {
-      return true;
-    }
+    if (style.display === "none") return true;
     // display:contents has no box, so box-only properties don't apply — it lays
     // its children out normally. Never prune it for opacity:0 (opacity affects a
     // box, of which there is none, so descendants still paint) or for its 0x0
     // rect; walk() descends and promotes the visible content.
     if (style.display === "contents") return false;
     if (style.opacity === "0") return true;
+    // visibility:hidden is deliberately NOT hard-pruned here: visibility inherits but a
+    // descendant can override it back to visible, so cutting the subtree would drop
+    // painted content. walk() descends and suppresses only this element's own paint
+    // (see \`invisibleSelf\`), pruning it only if no visible descendant survives.
     const r = getBCR.call(el);
     if (r.width > 0 && r.height > 0) return false;
     // Zero-area box: prune it only when it clips its overflow (a collapsed
@@ -282,30 +305,50 @@ export const DESCRIBE_DOM_SCRIPT = `(() => {
       }
     }
 
-    const text = ownText(el);
-    const name = accessibleName(el);
-    const clickable = isInteractive(el);
+    // A visibility:hidden element paints nothing itself, but a descendant can override
+    // visibility back to visible, so walk() descended instead of pruning (see hidden()).
+    // Suppress this element's own paint — its text / name / interactivity are invisible —
+    // and treat it as box-less so it contributes no box of its own: it survives only
+    // through, and is framed by, whatever visible descendants it has.
+    const invisibleSelf = style.visibility === "hidden";
+    const text = invisibleSelf ? "" : ownText(el);
+    const name = invisibleSelf ? null : accessibleName(el);
+    const clickable = invisibleSelf ? false : isInteractive(el);
     const role = nodeRole(el);
-    const bl = boxless(el, style);
+    // getAttribute, not el.id: like .title, a named form control can clobber the .id
+    // property to a DOM node, which would then break JSON.stringify of the tree.
+    // Computed here (before promotion) because an id / data-testid is agent-visible
+    // targeting info: a wrapper that carries one is not a pure layer, so neither
+    // promotion path below may drop it.
+    const id =
+      getAttr.call(el, "id") ||
+      getAttr.call(el, "data-testid") ||
+      getAttr.call(el, "data-test-id");
+    const bl = invisibleSelf || boxless(el, style);
 
     // A box-less element has no rect of its own. Its visible extent is whatever its
-    // children span; with no surviving child, fall back to the painted extent of its
-    // own inline content. If that is still zero-area it paints nothing — e.g. a
-    // transform: scale(0) subtree, whose descendants all collapse — so it is invisible
-    // and dropped. A box-less wrapper with one child and nothing of its own is just a
-    // layer, so promote the child. (Clickable / named box-less nodes fall through and
-    // are emitted with this child- or content-spanning frame.)
+    // children span; with no surviving child, fall back to the painted extent of its own
+    // inline TEXT — but only when it has its own text. A Range over a wrapper whose
+    // element children were all pruned as invisible still measures their
+    // visibility:hidden / opacity:0 layout boxes (non-zero though nothing paints), which
+    // would resurrect an empty wrapper with a real frame; own text is the only inline
+    // content not already represented by childResults. If the frame is still zero-area it
+    // paints nothing — e.g. a transform: scale(0) subtree, whose descendants all collapse
+    // — so it is invisible and dropped. A box-less wrapper with one child and nothing of
+    // its own (no clickable / name / text / identifier) is just a layer, so promote the
+    // child. (Clickable / named / identified box-less nodes fall through and are emitted
+    // with this child- or content-spanning frame.)
     let selfFrame;
     if (bl) {
       selfFrame = unionFrame(childResults);
-      if (selfFrame.width <= 0 && selfFrame.height <= 0) {
+      if (text && selfFrame.width <= 0 && selfFrame.height <= 0) {
         const cf = contentFrame(el);
         if (cf) selfFrame = cf;
       }
       if (childResults.length === 0 && selfFrame.width <= 0 && selfFrame.height <= 0) {
         return null;
       }
-      if (!clickable && !name && !text && childResults.length === 1) {
+      if (!clickable && !name && !text && !id && childResults.length === 1) {
         return childResults[0];
       }
     } else {
@@ -313,13 +356,14 @@ export const DESCRIBE_DOM_SCRIPT = `(() => {
     }
 
     // Prune structural wrappers with no info that just add a layer.
-    // Keep them if they're roots/clickable/named/have text.
+    // Keep them if they're roots/clickable/named/identified/have text.
     if (
       depth > 0 &&
       childResults.length === 1 &&
       !clickable &&
       !name &&
       !text &&
+      !id &&
       role === "div"
     ) {
       return childResults[0];
@@ -331,12 +375,6 @@ export const DESCRIBE_DOM_SCRIPT = `(() => {
     };
     if (name) node.label = name;
     if (text && text !== name) node.value = text.slice(0, 200);
-    // getAttribute, not el.id: like .title, a named form control can clobber the .id
-    // property to a DOM node, which would then break JSON.stringify of the tree.
-    const id =
-      getAttr.call(el, "id") ||
-      getAttr.call(el, "data-testid") ||
-      getAttr.call(el, "data-test-id");
     if (id) node.identifier = id;
     if (clickable) node.clickable = true;
     if (isDisabled(el)) node.disabled = true;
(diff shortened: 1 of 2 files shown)


hubgan (commented):
Reviewed the Chromium `describe` walker changes and ran the injected script against real headless Chrome, both directly over CDP and through the tool-server's `describe`. Comments inline.

packages/tool-server/src/tools/describe/platforms/chromium.ts
```diff
@@ -179,6 +286,32 @@ const DESCRIBE_DOM_SCRIPT = `(() => {
...
+    let selfFrame;
+    if (bl) {
+      selfFrame = unionFrame(childResults);
+      if (selfFrame.width <= 0 && selfFrame.height <= 0) {
+        const cf = contentFrame(el);
+        if (cf) selfFrame = cf;
+      }
+      if (childResults.length === 0 && selfFrame.width <= 0 && selfFrame.height <= 0) {
+        return null;
+      }
```
hubgan, line 373: This box-less branch keeps a wrapper whenever `contentFrame(el)` returns a non-zero rect, but `Range.getBoundingClientRect()` measures layout geometry, not paint. A `visibility:hidden` or `opacity:0` descendant still occupies layout, so its Range rect is non-zero even though it paints nothing. When such a wrapper's element children have all been pruned as invisible, `childResults` is empty yet `selfFrame` is non-zero, so the `childResults.length === 0 && zero-area` drop below never fires and the wrapper is emitted as an empty node with a real frame.

Reproduced in headless Chrome: `<div style="display:contents"><span style="visibility:hidden">…</span></div>` and the `opacity:0` equivalent each surface in the describe output as a bare `div` with a non-zero frame and no children — i.e. a tappable-looking node over screen area where nothing is painted. A `display:none` child is correctly dropped (it collapses layout to 0×0). This is the same subtree the "still prunes a normal (boxed) opacity:0 subtree" test asserts is pruned; wrapping it in a box-less element makes it survive.

packages/tool-server/src/tools/describe/platforms/chromium.ts
```diff
@@ -108,43 +144,109 @@ const DESCRIBE_DOM_SCRIPT = `(() => {
...
-      return false;
+  function frame(el) {
+    return normRect(getBCR.call(el));
+  }
+
+  // Painted extent of an element's own inline content. A box-less element
+  // (display:contents, or a zero-width box whose text overflows) has a 0x0 border box
+  // yet its text still paints; a Range measures that. Crucially it returns 0x0 when an
+  // ancestor transform (e.g. scale(0)) collapses the paint, which is how we tell
+  // genuinely-invisible content apart from merely box-less content.
```
hubgan: This comment states the Range "returns 0x0 when an ancestor transform (e.g. scale(0)) collapses the paint, which is how we tell genuinely-invisible content apart from merely box-less content." That distinction only holds for ancestor transforms and `display:none`. A `visibility:hidden` or `opacity:0` descendant keeps its layout box, so `Range.getBoundingClientRect()` is non-zero even though nothing paints — so this measurement does not actually separate invisible content from box-less content in those two cases.

packages/tool-server/src/tools/describe/platforms/chromium.ts
```diff
@@ -16,39 +16,72 @@ import type { DescribeNode, DescribeTreeData } from "../contract";
...
+ *    traverse it and promote them instead of cutting the subtree. This covers
+ *    display:contents wrappers (React Native Web nests content under them), the
+ *    zero-height anchor of an absolutely-positioned dropdown/popover/portal, and
+ *    float/overflow wrappers that collapse to a zero-height box. A collapsed
+ *    overflow:hidden container genuinely hides its content, so it stays pruned.
+ *  - Cap node count at 5000 — that, not depth, bounds the payload a runaway SPA
+ *    would otherwise serialize past CDP's single Runtime.evaluate reply limit
+ *    (~50MB). Cap depth at 60 purely to bound recursion: modern React DOMs
+ *    (React Native Web, navigator/provider stacks) routinely nest 25+ levels
+ *    before reaching leaf text, so a shallower cap silently clips real content.
```
hubgan, line 33: A side effect of no longer pruning zero-area subtrees: the walker now descends into subtrees it previously cut (every `display:contents` element and every zero-area `overflow:visible` box), and `nodeBudget` is decremented on entry to `walk` even for wrappers later collapsed out of the output. On `display:contents`-heavy or portal-heavy pages the 5000-node budget is reached noticeably sooner, and when it is, the tree is truncated with the only signal written to `process.stderr` — the returned `DescribeTreeData` carries no truncation flag, so the agent receives a silently partial tree.

packages/tool-server/src/tools/describe/platforms/chromium.ts
```diff
@@ -108,43 +144,109 @@ const DESCRIBE_DOM_SCRIPT = `(() => {
...
+  function boxless(el, style) {
+    if (style.display === "contents") return true;
+    const r = getBCR.call(el);
+    return r.width <= 0 || r.height <= 0;
+  }
+
+  function hidden(el, style) {
+    if (style.visibility === "hidden" || style.display === "none") {
+      return true;
+    }
```
hubgan: `visibility` inherits, but a descendant can override it back to `visible`. Returning early here on `style.visibility === "hidden"` means `walk` never descends, so a `visibility:visible` element nested under a `visibility:hidden` ancestor is dropped even though it paints. This predates the change, but it becomes more reachable now that box-less wrappers are traversed rather than pruned.

packages/tool-server/src/tools/describe/platforms/chromium.ts
```diff
@@ -179,6 +286,32 @@ const DESCRIBE_DOM_SCRIPT = `(() => {
...
+      if (selfFrame.width <= 0 && selfFrame.height <= 0) {
+        const cf = contentFrame(el);
+        if (cf) selfFrame = cf;
+      }
+      if (childResults.length === 0 && selfFrame.width <= 0 && selfFrame.height <= 0) {
+        return null;
+      }
+      if (!clickable && !name && !text && childResults.length === 1) {
+        return childResults[0];
+      }
```
hubgan: This single-child promotion runs for any role and at any depth, unlike the structural-wrapper collapse just below it which is gated on `role === "div" && depth > 0`. A box-less wrapper that carries an `id` / `data-testid` (emitted as `identifier` further down) but has a single anonymous child is replaced by that child, so its identifier is dropped from the tree.

packages/tool-server/src/tools/describe/platforms/chromium.ts
```diff
@@ -16,39 +16,72 @@ import type { DescribeNode, DescribeTreeData } from "../contract";
...
+  const getTagName = Object.getOwnPropertyDescriptor(Element.prototype, "tagName").get;
+  const getChildrenEls = Object.getOwnPropertyDescriptor(Element.prototype, "children").get;
+  const getShadowRoot = Object.getOwnPropertyDescriptor(Element.prototype, "shadowRoot").get;
+  const getScrollHeight = Object.getOwnPropertyDescriptor(Element.prototype, "scrollHeight").get;
+  const getClientHeight = Object.getOwnPropertyDescriptor(Element.prototype, "clientHeight").get;
+  const getScrollWidth = Object.getOwnPropertyDescriptor(Element.prototype, "scrollWidth").get;
+  const getClientWidth = Object.getOwnPropertyDescriptor(Element.prototype, "clientWidth").get;
+  const getAttr = Element.prototype.getAttribute;
+  const hasAttr = Element.prototype.hasAttribute;
+  const getBCR = Element.prototype.getBoundingClientRect;
```
hubgan: These captures assume every descriptor is present: `Object.getOwnPropertyDescriptor(Node.prototype, "childNodes").get` throws a TypeError at the top of the script if any of these accessors is ever absent, which aborts the entire describe rather than degrading a single node. Not reachable through page content today, but it's a broader failure mode than the per-node crashes this hardening is removing.

packages/tool-server/src/tools/describe/platforms/chromium.ts
```diff
@@ -57,44 +90,47 @@ const DESCRIBE_DOM_SCRIPT = `(() => {
...
+    // aborting the whole describe. getAttribute always yields string | null.
+    const t = getAttr.call(el, "title");
     if (t) return t.slice(0, 200);
     return null;
   }
 
   function ownText(el) {
     let s = "";
-    for (const child of el.childNodes) {
+    for (const child of getChildNodes.call(el)) {
```
hubgan, line 116: `child.nodeType` on the next line is read directly on a child that can itself be a clobbering `<form>` (a named control `nodeType` shadows the inherited property to an element). It is harmless here only because the comparison is `=== 3`: a shadowed value is `!== 3`, so the form is skipped exactly as its true `nodeType === 1` would skip it, and `nodeValue` is only read for genuine text nodes. It is the one inherited read left unrouted through the captured getters, and its safety rests on the `=== 3` semantics rather than on being clobber-proof.

packages/tool-server/test/describe-chromium-script.test.ts
```diff
@@ -0,0 +1,432 @@
...
+            left: c.x,
+            top: c.y,
+            right: c.x + c.w,
+            bottom: c.y + c.h,
+            width: c.w,
+            height: c.h,
+          };
+        },
+      };
+    },
```
hubgan, line 247: The mock `Range` returns `target.__content` verbatim, so it can only model a rect explicitly attached to the box-less element itself. It cannot reproduce the real-browser behavior where `Range.getBoundingClientRect()` over a box-less wrapper picks up the layout geometry of a `visibility:hidden` / `opacity:0` child. Consequently the suite cannot observe the case where such a wrapper is kept with a non-zero content frame despite painting nothing.

packages/tool-server/test/describe-chromium-script.test.ts
```diff
@@ -0,0 +1,432 @@
...
+  it("prunes a box-less leaf whose painted content is also zero-area (transform: scale(0))", () => {
+    const { tree } = run([el({ text: "SCALE0", rect: ZERO, content: null })]);
+    expect(valuesOf(tree)).not.toContain("SCALE0");
+  });
+
+  it("surfaces a box-less leaf whose text actually paints (overflowing / contents text)", () => {
+    const { tree } = run([
+      el({ text: "OVERFLOWTEXT", rect: ZERO, content: { x: 0, y: 50, w: 80, h: 15 } }),
+    ]);
+    expect(valuesOf(tree)).toContain("OVERFLOWTEXT");
```
hubgan, line 389: These visibility tests assert only that a text value appears (`valuesOf`); none assert the `frame` of a box-less node, which is the behavior being introduced. And in every fixture the box-less node has a single child and is promoted (`return childResults[0]`), so `unionFrame`'s multi-child `Math.min`/`Math.max` is computed and then discarded — never reached by an assertion. The emitted-with-a-union-frame path (a box-less node that is clickable/named or has 2+ children) is not covered.

packages/tool-server/test/describe-chromium-script.test.ts
```diff
@@ -0,0 +1,432 @@
...
+  const node = new MockElement() as MockElement & Record<string, unknown>;
+  const rect = opts.rect ?? { x: 0, y: 0, w: 100, h: 20 };
+  node.tagName = (opts.tag ?? "div").toUpperCase();
+  // Backing fields read by the Element.prototype getAttribute/hasAttribute/
+  // getBoundingClientRect methods defined above (the script reads them via the prototype).
+  node.__attrs = opts.attrs ?? {};
+  node.__rect = rect;
+  node.children = opts.children ?? [];
+  node.childNodes = opts.text ? [{ nodeType: 3, nodeValue: opts.text }] : [];
+  node.shadowRoot = null;
```
hubgan: The mock always sets `shadowRoot = null`, and no fixture supplies a non-null shadow root or an iframe `contentDocument`, so the shadow-DOM and iframe piercing branches of the walker are not exercised by this suite (the clobber test only confirms a control named `shadowRoot` is not re-walked).

packages/tool-server/src/tools/describe/platforms/chromium.ts
latekvo: Fixed in 0c17bc80c8cc6ff885e689d3177e20645e8d32b1.

The `contentFrame()` fallback is now consulted only when the element has its own text. A Range over a wrapper whose element children were all pruned as invisible still measures their `visibility:hidden` / `opacity:0` layout boxes (non-zero even though nothing paints); own text is the only inline content not already represented by `childResults`, so gating on it removes the phantom while still keeping genuinely-painting box-less text. A wrapper with no own text and only invisible children now hits the `childResults.length === 0 && zero-area` drop and returns null.

Reproduced in headless Chrome first: `<div style="display:contents"><span style="visibility:hidden">…` and the `opacity:0` equivalent each surfaced as a bare `div` with a non-zero frame and no children before the change, and are dropped after it (the `display:none` case was already correctly dropped). Locked in by a mock test in b2a6258318909b17165b6e420945ff0c656f0585 whose Range now unions descendant geometry, so it fails on the old code and passes on the new.

packages/tool-server/src/tools/describe/platforms/chromium.ts
latekvo: Fixed in 0c17bc80c8cc6ff885e689d3177e20645e8d32b1.

The comment was overclaiming and is corrected. It now states that `contentFrame` measures the element's own inline text, returns 0x0 only when an ancestor transform (or `display:none`) collapses the paint, and does NOT distinguish a `visibility:hidden` / `opacity:0` descendant — those keep their layout box, so the Range is non-zero though nothing paints. It also records that `walk()` consults it only when the element has its own text, precisely so that limitation can't resurrect a wrapper whose element children were all pruned as invisible.

packages/tool-server/src/tools/describe/platforms/chromium.ts
latekvo: Fixed in 0c17bc80c8cc6ff885e689d3177e20645e8d32b1.

The substantive part — the agent receiving a silently partial tree — is addressed: a truncated tree is now surfaced through the existing `hint` field on `DescribeTreeData`, which `withDescription` copies into `DescribeResult` and which iOS/Vega already use, so truncation reaches the agent in the response rather than only `process.stderr`. The hint tells the agent the tree is partial and to scope the inspection.

On the budget itself: decrementing on entry is deliberate — it bounds total traversal work (the real cost a runaway `display:contents`/portal-heavy DOM imposes, and what actually protects the single `Runtime.evaluate` reply), not just the emitted-node count. Switching to decrement-on-emit would let an adversarial DOM be walked unbounded. So the accounting stays, but truncation is now explicit to the agent instead of silent.

packages/tool-server/src/tools/describe/platforms/chromium.ts
latekvo: Fixed in 0c17bc80c8cc6ff885e689d3177e20645e8d32b1.

`visibility:hidden` is no longer a hard prune in `hidden()`. Because visibility inherits but a descendant can override it back to visible, cutting the subtree dropped painted content. `walk()` now descends into a `visibility:hidden` element and suppresses only that element's own paint (its text / name / interactivity, and its own box — it's treated as box-less), so a `visibility:visible` descendant surfaces with its own frame; a subtree with no visible descendant is still pruned, with no phantom node left behind. `opacity:0` and `display:none` stay hard prunes (neither can be overridden).

Reproduced in headless Chrome — `<div style="visibility:hidden"><span style="visibility:visible">…` dropped the span before the change and surfaces it after — and covered by mock tests in b2a6258318909b17165b6e420945ff0c656f0585 (override surfaces, own text suppressed, fully-hidden subtree pruned).

packages/tool-server/src/tools/describe/platforms/chromium.ts
latekvo: Fixed in 0c17bc80c8cc6ff885e689d3177e20645e8d32b1.

The `id` / `data-testid` is now computed before the box-less branch and both promotion paths are gated on the element carrying none: the box-less single-child promotion and the structural `role === "div"` collapse just below it (that one dropped the identifier for this case too). An identified wrapper is therefore kept with its `identifier` instead of being replaced by its child.

Reproduced in headless Chrome (`<div id="wrap" style="display:contents"><div>…</div></div>` lost `wrap` before, keeps it now) and locked in by mock tests in b2a6258318909b17165b6e420945ff0c656f0585 for both the box-less and the boxed-structural case, plus one confirming an anonymous wrapper is still promoted.

packages/tool-server/src/tools/describe/platforms/chromium.ts
latekvo: Fixed in 0c17bc80c8cc6ff885e689d3177e20645e8d32b1.

The descriptor captures now go through a `protoGetter(proto, prop)` helper that returns the accessor's getter, or — if the descriptor is absent — a direct-read fallback, instead of `.get` on `undefined` throwing at script top. A missing accessor now degrades a single property rather than aborting the whole describe.

Reproduced in headless Chrome: with `Element.prototype.scrollHeight` removed, the old script threw exactly this `TypeError: Cannot read properties of undefined (reading 'get')` at the top and aborted the entire describe, while the new one completes with a full tree (the `scrollHeight` read just returns undefined for the affected node). A mock test in b2a6258318909b17165b6e420945ff0c656f0585 removes a captured accessor and asserts the walk degrades instead of throwing.

packages/tool-server/src/tools/describe/platforms/chromium.ts
latekvo: Fixed in 0c17bc80c8cc6ff885e689d3177e20645e8d32b1.

You're right that it was safe (a clobbered `nodeType` is `!== 3`, so the form was skipped anyway), but it was the one inherited read left unrouted, so I removed the exception rather than leaving its safety resting on the `=== 3` comparison: `nodeType` and `nodeValue` now route through captured `Node.prototype` getters like every other inherited read. Behaviour is unchanged (verified — the full suite still passes), the "every inherited read goes through a captured getter" invariant is now whole.

packages/tool-server/test/describe-chromium-script.test.ts
latekvo: Fixed in b2a6258318909b17165b6e420945ff0c656f0585.

The mock Range no longer returns `__content` verbatim: `getBoundingClientRect` now unions the target's own painted content with the still-laid-out boxes of its descendants (everything except `display:none`, since `visibility:hidden` / `opacity:0` keep their box). That reproduces exactly the real-browser case you flagged — a box-less wrapper whose Range is non-zero purely because an invisible child still lays out — and a new test uses it to assert that wrapper is dropped rather than kept with a non-zero content frame (the fix from 0c17bc80). It fails on the pre-fix walker and passes on the fixed one.

packages/tool-server/test/describe-chromium-script.test.ts
latekvo: Fixed in b2a6258318909b17165b6e420945ff0c656f0585.

Added frame assertions for box-less nodes on both paths: an emitted box-less wrapper (made clickable + named so it is not promoted) with two children at differing rects is now asserted to be framed by `unionFrame` across them — so the multi-child `Math.min`/`Math.max` is actually reached by an assertion — and a promoted single-child node is asserted to carry the child's own rect. The earlier tests only checked text presence; these check the emitted `frame`.

packages/tool-server/test/describe-chromium-script.test.ts
latekvo: Fixed in b2a6258318909b17165b6e420945ff0c656f0585.

The mock gained `shadow` and `iframeDoc` support (a non-null shadow root exposing `.children`, and an `<iframe>` whose `contentDocument.documentElement` is a mock subtree), and there are now two tests that exercise the previously-uncovered branches: one pierces an open shadow root and asserts its content surfaces, and one pierces a same-origin iframe's `contentDocument` and asserts the same. The existing clobber test still covers that a control named `shadowRoot` is not re-walked.

packages/tool-server/src/tools/describe/platforms/chromium.ts
```diff
@@ -175,30 +305,76 @@ const DESCRIBE_DOM_SCRIPT = `(() => {
...
+    if (bl) {
+      selfFrame = unionFrame(childResults);
+      if (text && selfFrame.width <= 0 && selfFrame.height <= 0) {
+        const cf = contentFrame(el);
+        if (cf) selfFrame = cf;
+      }
+      if (childResults.length === 0 && selfFrame.width <= 0 && selfFrame.height <= 0) {
+        return null;
+      }
+      if (!clickable && !name && !text && !id && childResults.length === 1) {
```
hubgan: Box-less single-child promotion drops non-div semantic roles (asymmetry). chromium.ts:351 promotes any box-less wrapper with one child and no clickable/name/text/id — but, unlike the boxed structural prune at :360-370, it does not gate on role === "div" (nor depth > 0). So a display:contents element with role="navigation"/"list"/aria-disabled and a single anonymous child gets promoted away, losing that role/state. Uncommon (role-bearing elements are usually boxed), but the two promotion paths should apply the same role === "div" guard for consistency.

packages/tool-server/src/tools/describe/platforms/chromium.ts
```diff
@@ -108,43 +162,114 @@ const DESCRIBE_DOM_SCRIPT = `(() => {
...
+    // its children out normally. Never prune it for opacity:0 (opacity affects a
+    // box, of which there is none, so descendants still paint) or for its 0x0
+    // rect; walk() descends and promotes the visible content.
+    if (style.display === "contents") return false;
+    if (style.opacity === "0") return true;
+    // visibility:hidden is deliberately NOT hard-pruned here: visibility inherits but a
+    // descendant can override it back to visible, so cutting the subtree would drop
+    // painted content. walk() descends and suppresses only this element's own paint
+    // (see \`invisibleSelf\`), pruning it only if no visible descendant survives.
+    const r = getBCR.call(el);
```
hubgan, line 245: Pure visibility:hidden subtrees now consume the node budget. hidden() (:227-231) intentionally doesn't prune visibility:hidden, so walk() decrements nodeBudget (:269) and traverses the whole subtree before dropping it when no visible descendant survives. Old code cut it at the root (0 budget). On a page with a large pure visibility:hidden region this can trigger premature MAX_NODES truncation and drop visible content. Narrow in practice â the common transition pattern visibility:hidden; opacity:0 still short-circuits on opacity:0 at :226 (pruned, 0 budget) â but worth a comment acknowledging the trade-off.

packages/tool-server/src/tools/describe/platforms/chromium.ts
```diff
@@ -108,43 +162,114 @@ const DESCRIBE_DOM_SCRIPT = `(() => {
...
+
+  // Painted extent of an element's own inline TEXT. A box-less element (display:contents,
+  // or a zero-width box whose text overflows) has a 0x0 border box yet its text still
+  // paints; a Range measures that. It returns 0x0 when an ancestor transform (e.g.
+  // scale(0)) or display:none collapses the paint. It does NOT distinguish a
+  // visibility:hidden / opacity:0 *descendant* — those keep their layout box, so the
+  // Range is non-zero even though nothing paints — which is why walk() consults this only
+  // when the element has its own text, never to resurrect a wrapper whose element
+  // children were all pruned as invisible.
+  function contentFrame(el) {
```
hubgan, line 194: contentFrame can over-size a box-less element's frame. contentFrame (:194) uses Range.selectNodeContents, which spans the element's own text and the still-laid-out boxes of visibility:hidden/opacity:0 descendants (the code comment acknowledges Range can't distinguish them). It's correctly gated behind "has own text" (:344), so it never resurrects an empty wrapper â but when it does fire on an element whose element-children were all pruned as invisible, the resulting frame can be inflated to span those invisible descendants. Frame inaccuracy only (tap-center may drift), never a crash or phantom node.

hubgan (approved):
The comments below are non-blocking edge cases

packages/tool-server/src/tools/describe/platforms/chromium.ts
latekvo: Fixed in 7c1c964c. The box-less single-child promotion is now gated on `role === "div"` (or a visibility:hidden element with no meaningful paint of its own), matching the boxed structural collapse just below it. A box-less element with a semantic role (list, nav, section, listitem, …) now keeps its role instead of being flattened into its child. Pinned by a new test.

packages/tool-server/src/tools/describe/platforms/chromium.ts
latekvo: Fixed in 7c1c964c. Only elements that can actually emit consume the node budget now — a visibility:hidden element no longer decrements MAX_NODES on entry, so a large fully-hidden subtree (a closed drawer/modal) can't exhaust the budget and truncate genuinely visible content elsewhere. Its visible descendants, if any, still consume the budget themselves. Pinned by a 5100-node hidden-subtree test.

packages/tool-server/src/tools/describe/platforms/chromium.ts
latekvo: Fixed in 7c1c964c. contentFrame now measures only the element's own direct text nodes instead of selectNodeContents over the whole element, so it no longer spans the still-laid-out boxes of visibility:hidden / opacity:0 descendants and over-sizes the frame (mis-placing the tap point). Pinned by a frame assertion on a box-less node with an invisible large-box child.

PR #419, opened 2026-06-26
fix(tool-server): stop list-devices from hanging - wedged-VVD stall and CLI socket leak

hubgan:
## Summary

`list-devices` (an `alwaysLoad` tool called at session start and frequently after)
could appear to **hang**. Investigation found **two independent causes**, both fixed here:

1. **Server-side (Vega):** a wedged Vega VVD turned `list-devices` into a ~40s stall
   and leaked `vega` CLI subprocesses.
2. **Client-side (CLI):** every `argent run …` lingered a constant **~6s after it had
   already printed its result**, because a health-check `fetch` never drained its
   response body and kept an undici keep-alive socket ref'd until the server closed it.

Net effect for `argent run list-devices`: **6.15s → 0.91s**, and the worst-case
wedged-VVD stall drops from ~40s to a bounded short call.

---

## Problem 1 — wedged Vega VVD stalls `list-devices` (~40s) and leaks processes

On a host with the Vega SDK installed and a **running-but-unresponsive VVD**,
`list-devices`' Vega discovery path made **two serial 20s CLI calls** that both hung:

- `vega device list` → SIGKILL at 20s
- then (recovery) `vega device info` → SIGKILL at another 20s → **~40s total**

Three compounding defects:

- `execFileAsync` resolves on stdio **close**, not child **exit**. The `vega` launcher
  forks a `python3 → node → vda` worker tree; an orphaned worker holds the stdout pipe
  open, so the call stayed pending for the full timeout even after the launcher was
  killed.
- The `device info` recovery fired even when `device list` had **hung**, doubling the
  cost and spawning a second hanging subprocess.
- On timeout we killed only the **direct child**, so the `python3 → node → vda` worker
  tree was **orphaned, not reaped** — leaking a process tree per call (a wedged host had
  accumulated ~269 of them).

### Fix

- **`runVega` (`vega-cli.ts`):** switched from `execFileAsync` to `spawn` with
  **`detached: true`**, making the launcher its own **process-group leader**, and on
  timeout SIGKILL the **whole process group** (`process.kill(-pid)`) — which reaps the
  entire `python3 → node → vda` tree, **not just the launcher**. We also destroy our
  stdio stream ends so a pipe-holding worker can't keep the call pending. (`execFile`
  silently drops `detached`, which is why the switch to `spawn` was required.)
  - **Success path doesn't depend on stdio EOF either:** we prefer `close` (fully
    drained), but if a grandchild holds the stdout pipe open after a **clean exit**, we
    fall back to `exit` + a short (~1s) drain grace and resolve with the captured
    output — so a finished call can't be stalled into the timeout and then rejected,
    discarding valid data. The output cap is measured in real UTF-8 bytes. For that
    guarantee to hold at the boundary, the `exit` handler **disarms the main timeout
    timer**: once the child has terminated the wall-clock deadline is moot, so a clean
    exit landing within the drain grace of the deadline resolves with its output instead
    of being mis-rejected as a `timeout` (which would also wrongly suppress the
    `listVegaDevices` recovery for a still-running VVD).
  - **Belt-and-suspenders for a worker that escapes the group:** an earlier
    investigation reported the worker tree `setsid()`-ing into its own session (which
    would defeat a group-only SIGKILL). To be safe regardless of which is true, on
    timeout we **snapshot the launcher's descendants from the process table *before*
    killing** (a setsid'd worker keeps its `ppid` pointing at the launcher until the
    launcher dies, so a `ppid` tree-walk still reaches it) and SIGKILL those pids
    individually after the group kill. The snapshot is **precise** — only this
    launcher's own descendants, by pid, so a concurrent `vega` call's workers are never
    touched — and bounded, and the group kill runs regardless of whether it succeeds.
  - A real-subprocess test confirms **0 orphans** after a timeout, including a worker
    deliberately spawned into its **own** process group (so the group kill alone cannot
    reach it — the test fails without the descendant sweep).
- **`listVegaDevices` (`vega-devices.ts`):** skip the `device info` recovery only when
  `device list` **timed out** (a wedged agent — keyed off the failure's
  `error_kind: "timeout"`), so a hung list no longer pays for a second timeout wait. A
  *fast* (non-timeout) list failure still recovers a genuinely-running VVD rather than
  mis-reporting it as stopped. Tightened discovery timeouts (20s → 6s/4s — a generous
  multiple of the ~1s healthy time so a cold-start `vega` launcher isn't spuriously
  timed out, with their sum (10s) kept under the 15s branch deadline with margin).
- **`list-devices.ts`:** added a 15s `withDeadline` backstop around the Android and Vega
  branches so no single branch can stall the whole fan-out (returns partial results +
  logs a note). Sized to sit ~5s above the Vega branch's ~10s worst case, so the
  backstop never truncates a slow-but-completing branch (which would drop a real
  device); an invariant test guards that ordering.
- **`readAvdName` (`adb.ts`):** run the two `getprop` probes concurrently so a wedged
  emulator costs 5s, not 10s.

---

## Problem 2 — `argent run` hangs ~6s after printing output (CLI socket leak)

Independent of Vega. The tool-server returns `list-devices` in <1s, but the CLI process
took a **constant ~6s** to exit. Any path that calls `process.exit()` was instant; any
path that returned naturally (`argent run …`, `argent run … --help`, `argent tools`)
lingered ~6s.

**Root cause:** `isToolsServerHealthy` (`launcher.ts`) did
`const res = await fetch('/tools'); return res.ok;` and **never drained the body**.
`/tools` is a large (~112KB) payload, so undici keeps the keep-alive socket **ref'd**
until the tool-server's idle `keepAliveTimeout` (~5s) closes it. `ensureToolsServer`
runs this health check on every invocation, and `run()`'s success path has no
`process.exit(0)`, so the ref'd socket holds the event loop open. (Confirmed
telemetry-independent — `DO_NOT_TRACK` made no difference.) The same undrained-body
pattern existed in `link.ts` (`argent link`).

### Fix

Drain the response body before returning in both health-check-style fetches:

```ts
await res.body?.cancel().catch(() => {});
return res.ok;
```

(`fetchTools`/`callTool` were already fine — they `await res.json()`.)

---

## Testing

**Empirical, end-to-end** (on a machine reproducing the wedged VVD):

| Command | Before | After |
| --- | --- | --- |
| `argent run list-devices` | 6.15s | **0.91s** |
| `argent run list-devices --help` | 6.14s | **0.16s** |
| `argent tools` | 6.15s | **0.16s** |
| handle probe after run | 2 sockets ref'd for ~6s | **none — immediate exit** |

- Raw `POST /tools/list-devices` was ~0.7s throughout — confirming the ~5.4s was pure
  CLI teardown, not tool work.
- **Process-group reap proven empirically:** the live `vega → bash → python3 → node`
  tree all shares the detached launcher's PGID; a group-SIGKILL left **0 survivors**.
  A real-subprocess test (unmocked `runVega` vs. a fake hanging `vega` on `PATH`) confirms
  it settles at its deadline **and** leaves **0 orphaned processes**.

**Automated:**

- `vega-cli-timeout.test.ts` (real-subprocess test) — `runVega` rejects on its deadline,
  **reaps the entire worker tree (0 orphans)** — the fake's hung worker is spawned into its
  **own** process group, so the test fails without the descendant sweep — and returns
  normally when the CLI responds. Also covers the boundary case: a **clean exit whose
  drain grace outlasts the deadline** resolves with its captured output rather than being
  rejected as a `timeout` (guards the `exit`-handler timer disarm above).
- `vega-adb-connect-recovery.test.ts` — a *timed-out* `device list` does not trigger a
  second `device info` call (no stacking); a *fast, non-timeout* failure **still** recovers
  a running VVD.
- `list-devices-deadline.test.ts` — `withDeadline` returns the branch value on the happy
  path and the fallback (with a logged note) when a branch exceeds the deadline.
- `list-devices.test.ts` — added a `spawn` mock (vega moved off `execFile`).
- `launcher-helpers.test.ts` — `isToolsServerHealthy` cancels (drains) its response body.
- Suites green: **tool-server (1610), argent-tools-client (109), argent-cli (84)**.
- eslint (`--max-warnings 0`), prettier, and `tsc` clean.

---

### DIFF
--- a/packages/telemetry/src/posthog.ts
+++ b/packages/telemetry/src/posthog.ts
@@ -4,7 +4,7 @@ import { PostHog } from "posthog-node";
 export const POSTHOG_HOST = "https://eu.i.posthog.com";
 
 /** Public write-only PostHog project token. */
-export const POSTHOG_PROJECT_TOKEN = "phc_mebZhuUmASbeeXkdv7E2wqTZdtstPxQqHBhZfLSeQ5YL";
+export const POSTHOG_PROJECT_TOKEN = "phc_tkPxaBJ8WVr2KQAuu7FoN2nAcJ7MhVNsHSpUSuNC9HGV";
 
 interface ResolvedConfig {
   key: string;
--- a/packages/telemetry/src/posthog.ts
+++ b/packages/telemetry/src/posthog.ts
@@ -4,7 +4,7 @@ import { PostHog } from "posthog-node";
 export const POSTHOG_HOST = "https://eu.i.posthog.com";
 
 /** Public write-only PostHog project token. */
-export const POSTHOG_PROJECT_TOKEN = "phc_tkPxaBJ8WVr2KQAuu7FoN2nAcJ7MhVNsHSpUSuNC9HGV";
+export const POSTHOG_PROJECT_TOKEN = "phc_mebZhuUmASbeeXkdv7E2wqTZdtstPxQqHBhZfLSeQ5YL";
 
 interface ResolvedConfig {
   key: string;
--- a/packages/tool-server/src/tools/devices/list-devices.ts
+++ b/packages/tool-server/src/tools/devices/list-devices.ts
@@ -78,6 +78,48 @@ async function resolveVvdShadowAdbSerials<T extends { serial: string }>(
   return shadows;
 }
 
+// Hard backstop so no single discovery branch can stall this `alwaysLoad` tool,
+// which runs at session start and frequently after. The per-call subprocess
+// timeouts (and the no-stacking fix in listVegaDevices) already bound each
+// branch; this is defence-in-depth against an unforeseen stall — a wedged
+// device, a hung `ps`, a future serial call — so the worst case is a partial
+// list with a logged note, never a 40s "hang". Mirrors the existing
+// `.catch(() => [])` degradation, just for slowness rather than errors.
+//
+// Critically this must sit comfortably ABOVE every branch's own worst case, or it
+// stops being a last-resort backstop and starts truncating branches that would
+// have completed — dropping a real device from the list. The long pole is Vega: a
+// fast (non-timeout) `device list` failure followed by the `device info` recovery
+// is ~10s (its 6s + 4s discovery timeouts; see vega-devices.ts), so 15s keeps a
+// ~5s margin. (Android self-bounds at ~5s, the rest at <1s.) Still far below the
+// ~40s stall this whole fix targets.
+//
+// Note: this is a *deadline*, not cancellation — on timeout it resolves the
+// fallback while the underlying branch keeps running to completion in the
+// background. That's fine here because the per-call subprocess timeouts bound the
+// branch, so it settles shortly after rather than leaking work indefinitely.
+export const BRANCH_DEADLINE_MS = 15_000;
+
+export async function withDeadline<T>(p: Promise<T>, fallback: T, label: string): Promise<T> {
+  let timer: ReturnType<typeof setTimeout> | undefined;
+  try {
+    return await Promise.race([
+      p,
+      new Promise<T>((resolve) => {
+        timer = setTimeout(() => {
+          process.stderr.write(
+            `[list-devices] ${label} discovery exceeded ${BRANCH_DEADLINE_MS}ms; ` +
+              `returning partial results (a wedged device or its CLI is unresponsive)\n`
+          );
+          resolve(fallback);
+        }, BRANCH_DEADLINE_MS);
+      }),
+    ]);
+  } finally {
+    if (timer) clearTimeout(timer);
+  }
+}
+
 const zodSchema = z.object({});
 
 export const listDevicesTool: ToolDefinition<Record<string, never>, ListDevicesResult> = {
@@ -94,12 +136,32 @@ Booted/ready devices are listed first. Platforms whose CLI is unavailable are si
   zodSchema,
   services: () => ({}),
   async execute(_services, _params) {
+    // Every branch gets the same hard deadline so no single one can stall this
+    // `alwaysLoad` tool. The Android/Vega branches are the ones that actually shell
+    // out to potentially-wedged devices; iOS / AVD-list / Chromium already self-bound
+    // with their own short subprocess/socket timeouts, but wrapping them too makes the
+    // "no branch can hang the fan-out" guarantee universal at near-zero cost (the
+    // timer is cleared on the fast happy path). The deadline only substitutes a
+    // fallback on *slowness*; a rejection still propagates exactly as before — so the
+    // `.catch(() => [])` wrappers (and the lack of one on iOS/AVDs) are unchanged.
     const [ios, android, avds, chromium, vega] = await Promise.all([
-      listIosSimulators(),
-      listAndroidDevices().catch(() => []),
-      listAvds(),
-      discoverChromiumDevices().catch(() => []),
-      listVegaDevices().catch(() => []),
+      withDeadline(listIosSimulators(), [], "ios"),
+      withDeadline(
+        listAndroidDevices().catch(() => []),
+        [],
+        "android"
+      ),
+      withDeadline(listAvds(), [], "avds"),
+      withDeadline(
+        discoverChromiumDevices().catch(() => []),
+        [],
+        "chromium"
+      ),
+      withDeadline(
+        listVegaDevices().catch(() => []),
+        [],
+        "vega"
+      ),
     ]);
     const iosTagged: IosDevice[] = ios.map((s) => ({ platform: "ios", ...s }));
     iosTagged.sort(sortIos);
--- a/packages/tool-server/src/tools/devices/list-devices.ts
+++ b/packages/tool-server/src/tools/devices/list-devices.ts
@@ -1,6 +1,11 @@
 import { z } from "zod";
 import type { ToolDefinition } from "@argent/registry";
-import { listAndroidDevices, listAvds, consolePortFromAdbSerial } from "../../utils/adb";
+import {
+  listAndroidDevices,
+  listAvds,
+  consolePortFromAdbSerial,
+  ADB_DEVICES_TIMEOUT_MS,
+} from "../../utils/adb";
 import { listRunningVvdConsolePorts } from "../../utils/vega-process";
 import { listIosSimulators, type IosSimulator } from "../../utils/ios-devices";
 import { discoverChromiumDevices, type ChromiumDevice } from "../../utils/chromium-discovery";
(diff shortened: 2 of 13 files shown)


packages/telemetry/src/posthog.ts
```diff
@@ -4,7 +4,7 @@ import { PostHog } from "posthog-node";
 export const POSTHOG_HOST = "https://eu.i.posthog.com";
 
 /** Public write-only PostHog project token. */
-export const POSTHOG_PROJECT_TOKEN = "phc_mebZhuUmASbeeXkdv7E2wqTZdtstPxQqHBhZfLSeQ5YL";
+export const POSTHOG_PROJECT_TOKEN = "phc_tkPxaBJ8WVr2KQAuu7FoN2nAcJ7MhVNsHSpUSuNC9HGV";
```
latekvo: This swaps the production PostHog project token (`phc_mebZ…` to `phc_tkPx…`). On `main` and at this branch's merge base the token is still `phc_mebZ…`; the new value currently exists only on the unmerged `feat/structured-failure-telemetry` branch, and this change was made inside this PR's main fix commit. Nothing about it relates to the list-devices hang or the CLI socket leak. Merging it would silently redirect all CLI, tool-server and MCP telemetry to a different PostHog project than the one `main` currently reports to.

packages/tool-server/test/list-devices-deadline.test.ts
```diff
@@ -0,0 +1,49 @@
...
+// Guards the timeout/backstop ordering: the backstop must sit ABOVE the slowest
+// branch's own worst case, or it stops being a last-resort and starts truncating a
+// branch that would have completed — dropping a real device from the list. The long
+// pole is Vega's slow recovery path: a fast (non-timeout) `device list` failure
+// followed by the `device info` recovery runs both discovery timeouts back-to-back.
+// If a future edit bumps either Vega timeout past the backstop, this fails loudly.
+describe("discovery timeout vs backstop invariant", () => {
+  it("the Vega branch worst case stays comfortably under the branch deadline", () => {
+    const vegaWorstCase = VEGA_DISCOVERY_LIST_TIMEOUT_MS + VEGA_DISCOVERY_INFO_TIMEOUT_MS;
+    expect(vegaWorstCase).toBeLessThan(BRANCH_DEADLINE_MS);
```
latekvo, line 29: This invariant only sums `VEGA_DISCOVERY_LIST_TIMEOUT_MS` (6s) and `VEGA_DISCOVERY_INFO_TIMEOUT_MS` (4s) = 10s against the 15s branch deadline. The Vega recovery path in `listVegaDevices` also runs two serial `listRunningVvdConsolePorts()` `ps` probes — one directly in the recovery condition before `readVegaInfo()`, and a second one inside `readVegaInfo()` → `runVegaDevice` → `singleVvdSelector` — each with its own 5s timeout. Counting those, the path's worst case is ~20s (6 + 5 + 5 + 4), which exceeds `BRANCH_DEADLINE_MS` (15s). In that case `withDeadline` fires on a recovery that was about to complete and returns `[]`, dropping a genuinely-running VVD from `list-devices` — the exact truncation this test and the accompanying comment claim to rule out. The asserted 3s margin doesn't reflect the true worst case because the two `ps` terms aren't included.

packages/tool-server/src/utils/vega-devices.ts
```diff
@@ -200,13 +233,27 @@ export async function listVegaDevices(): Promise<VegaDevice[]> {
  * Resolve the `amazon-` runtime serial of the currently-running VVD — e.g. right
  * after `boot-device` starts it. A freshly-started VVD can take a moment to
  * surface in `vega device list`, so retry a few times before giving up.
+ *
+ * Bounded by an overall wall-clock budget on top of the retry count: each
+ * `listVegaDevices()` is already short (its discovery timeouts no longer stack),
+ * but the budget is a hard cap so a wedged device can't drag this out attempt by
```
latekvo: This describes the budget as a 'hard cap' of 15s, but the deadline is only checked after each `listVegaDevices()` attempt completes, and `listVegaDevices()` is never cancelled. An attempt that starts just under the deadline runs to completion, and a single `listVegaDevices()` can take up to ~15-20s against a wedged-but-running VVD, so the real worst-case wall time here is ~21-30s rather than the stated 15s.

latekvo (changes requested):
I worked through this end to end and the core of the fix is solid and reproducible. Against the real built tool-server, the `/tools` body drain takes the natural-exit path of `isToolsServerHealthy` / `argent run` from ~6.0s down to ~1-50ms on Node 20, 22 and 24; the `runVega` detached-spawn plus process-group and descendant-sweep reaper settles on its own deadline and leaves zero orphaned processes (the real-subprocess test passed 15/15 runs, including a worker spawned into its own process group). All suites pass locally: tool-server 1610, tools-client 109, cli 84.

I'm requesting changes for one item that is unrelated to the hang/leak fix and would have real production impact (the PostHog project-token change, inline), plus two Vega-timing observations on the discovery path that are worth confirming. Details inline.

packages/telemetry/src/posthog.ts
latekvo: This changes `POSTHOG_PROJECT_TOKEN` from the value on `main` (`phc_mebZhuU…`) to a different project token (`phc_tkPxaBJ…`). It is the ingest key every telemetry event resolves through (`getClient()` → `track()` → `capture()`, used across the CLI, installer, MCP server and tool-server), so merging this redirects all product telemetry to a different PostHog project: the existing project/dashboards stop receiving events and they flow to the new one instead. The change is unrelated to the list-devices hang and isn't mentioned in the PR description or commit body — it reads like an accidentally-committed local override rather than an intended migration.

latekvo (commented):
A few things surfaced while reviewing the list-devices hang fix end-to-end (including running the rewritten `runVega` against real subprocesses). Details inline.

packages/tool-server/src/tools/devices/list-devices.ts
```diff
@@ -78,6 +78,48 @@ async function resolveVvdShadowAdbSerials<T extends { serial: string }>(
...
+// device, a hung `ps`, a future serial call — so the worst case is a partial
+// list with a logged note, never a 40s "hang". Mirrors the existing
+// `.catch(() => [])` degradation, just for slowness rather than errors.
+//
+// Critically this must sit comfortably ABOVE every branch's own worst case, or it
+// stops being a last-resort backstop and starts truncating branches that would
+// have completed — dropping a real device from the list. The long pole is Vega: a
+// fast (non-timeout) `device list` failure followed by the `device info` recovery
+// is ~10s (its 6s + 4s discovery timeouts; see vega-devices.ts), so 15s keeps a
+// ~5s margin. (Android self-bounds at ~5s, the rest at <1s.) Still far below the
```
latekvo: This line states the Android branch self-bounds at ~5s, which is the stated justification for the 15s `BRANCH_DEADLINE_MS`. The Android branch is `listAndroidDevices()` → `listAndroidSerials()` → `runAdb(["devices"])`, and that `runAdb` call passes no `timeoutMs`, so it uses the 30s default (`adb.ts:161`); the per-device getprop enrichment (~5s) runs on top, putting the true Android worst case around 35s, not 5s. Because that is above the 15s deadline, the invariant this block relies on — that the deadline sits comfortably above every branch's worst case so it never truncates a branch that would have completed — does not hold for Android. Against a slow/cold/wedged adb server (15–30s, the exact condition this fix targets) `withDeadline` substitutes `[]` and every Android device is dropped from the result, with only a stderr note.

packages/tool-server/src/utils/vega-cli.ts
```diff
@@ -144,25 +261,217 @@ function describeVegaFailure(args: string[], err: unknown): Error {
...
+      // discard valid output and mis-classify it `error_kind: "timeout"`, suppressing
+      // listVegaDevices' recovery. From here `exitTimer` (bounded) is the sole backstop.
+      clearTimeout(timer);
+      exitTimer = setTimeout(() => {
+        // `close` didn't follow the exit within the grace — a worker is holding our
+        // stdout pipe open. The child already wrote everything, so finish with what we
+        // have and destroy our read ends so the lingering worker can't keep us pending.
+        child.stdout?.destroy();
+        child.stderr?.destroy();
+        finish(code, signal);
```
latekvo, line 543: On the exit + drain-grace path, `finish()` resolves/rejects without reaping the process group. This path is only reached when `close` is delayed because a worker (e.g. a `python3 → node → vda` grandchild) is still holding the stdout pipe open — so by construction that worker is alive, and it is left orphaned here. I reproduced this against the built `runVega`: a fake `vega` that exits 0 while a detached child keeps stdout open resolves via this grace (~1.3s) with the output intact, but the pipe-holding worker is still running after the promise settles. This also contradicts the comment at lines 427-428 ("the launcher already exited, so its workers went with it"), which is not true on this path — it is exactly the path where a worker did not exit with the launcher.

packages/tool-server/test/vega-cli-timeout.test.ts
```diff
@@ -0,0 +1,260 @@
...
+    count = strayCount(sentinel);
+  }
+  return count;
+}
+
+describe("runVega timeout (real subprocess)", () => {
+  it("rejects on its own deadline when the CLI never returns", async () => {
+    const start = Date.now();
+    await expect(runVega(["hang", SENTINEL_DEADLINE], { timeoutMs: 400 })).rejects.toThrow(
+      /timed out/i
```
latekvo: This real-timeout test asserts only that the rejection message matches `/timed out/i`, but that message is built unconditionally in `rejectTimeout` independent of the classification, so matching it proves nothing about `error_kind`. The behaviour the fix depends on is that a genuine timeout yields `error_kind: "timeout"` — that is what `listVegaDevices` keys its recovery-skip off. The subprocess/overflow cases do assert `error_kind` (lines 226/242/257), but the timeout path never does, and the recovery test fabricates the kind rather than deriving it from a real `runVega`. A regression that made a real timeout classify as `subprocess` (e.g. dropping `killed`/`signal` from the rejection shape) would pass every assertion here yet re-introduce the stacked ~40s stall this PR targets.

latekvo (approved):

PR #420, opened 2026-06-26
fix(telemetry): add missing failure codes

hubgan:
## Summary

Adds the missing **failure codes** across the tool-server and converts raw
`throw new Error(...)` sites into classified `FailureError`s, so operational
failures land in named telemetry buckets instead of falling through to the
generic `REGISTRY_TOOL_FAILURE_UNCLASSIFIED` bucket.

- **39 new `FAILURE_CODES`** plus a new `"electron"` entry in `FAILURE_COMMANDS`.
- Throw sites migrated to `FailureError`, each carrying
  `error_code` + `failure_stage` + `failure_area` + `error_kind`, and where
  relevant `failure_command` / `network_failure` / subprocess metadata
  (`failure_exit_code` / `failure_signal` / `failure_spawn_code`).

## Addressing review feedback

The review found that several proposed codes were thrown only on paths that are
caught / reformatted before the telemetry boundary (a code only buckets if its
`FailureError` propagates out of a registered tool's `execute`), so they could
never bucket a real failure. Those are removed rather than left as dead codes:

- **Dropped `VEGA_PAGE_SOURCE_{FETCH_FAILED,HTTP_ERROR,RPC_ERROR,TIMEOUT,PARSE_FAILED}`**
  - `describeVega` deliberately swallows every toolkit failure (empty tree +
  relaunch hint) except `MultipleVegaDevicesError`, so these never reach
  `toolFailed`. The throw sites revert to plain `Error` (behaviour unchanged).
- **Dropped `CHROMIUM_CLIPBOARD_FAILED`** (the chromium-server Express route
  reformats the error to a 500 body; the image-copy path has no production
  caller) and **`CHROMIUM_BUTTON_UNSUPPORTED`** (the WS `button` handler
  reformats it; the registered `button` tool is apple+android only).
- **Fixed the `CHROMIUM_CDP_INVALID_RESPONSE` gap** - a reachable HTTP-200 with a
  non-JSON body made `res.json()` reject with a raw `SyntaxError` that escaped
  `fetchJson` unclassified; it is now classified `CHROMIUM_CDP_INVALID_RESPONSE`
  (matching the non-2xx and missing-`webSocketDebuggerUrl` paths). `AbortError`
  still surfaces untouched.
- **Unified two codes that carried two `error_kind`s** - both
  `REACT_PROFILER_NO_ACTIVE_SESSION` sites and both `PROFILER_NATIVE_TRACE_MISSING`
  sites now use `not_found` (a missing session / a trace file missing on disk),
  keeping telemetry keyed on `error_code` from splitting by platform.
- **Reverted an out-of-scope edit** - the in-flight Vega-boot coalescing Map-key
  separator is back to its original NUL byte.
- **Added `failure-classification.test.ts` coverage** - expanded from 7 to 17
  pinned codes, reading each back exactly how telemetry does
  (`getFailureSignal(err).error_code`) across the chromium CDP discovery family
  (incl. the malformed-200 regression guard), vega input, and native-profiler
  metadata / stack-query.

This branch also merges the latest `main`; the keyboard tool was refactored on
`main` into per-platform modules + a `createKeyboardTool` factory, so the
`KEYBOARD_KEY_UNSUPPORTED` / `KEYBOARD_CHARACTER_UNSUPPORTED` classifications were
re-homed onto its new throw sites (`platforms/chromium.ts`,
`simulator-server-keys.ts`; the vega path keeps its classification in
`vega-input.ts`).

## What changed

By area (the 39 codes):

- **Chromium / CDP (16)** - discovery, input, viewport, storage, tabs, describe,
  and Electron boot: `CHROMIUM_CDP_UNREACHABLE`, `CHROMIUM_CDP_NO_PAGE_TARGET`,
  `CHROMIUM_CDP_INVALID_RESPONSE`, `CHROMIUM_PARAM_INVALID`,
  `CHROMIUM_INPUT_INVALID`, `CHROMIUM_VIEWPORT_READ_FAILED`,
  `CHROMIUM_SCREENSHOT_FAILED`, `CHROMIUM_STORAGE_EVAL_FAILED`,
  `CHROMIUM_TAB_NOT_FOUND`, `CHROMIUM_TAB_OPEN_FAILED`, `CHROMIUM_DESCRIBE_FAILED`,
  `CHROMIUM_DEVICE_ID_INVALID`, `CHROMIUM_ELECTRON_APP_PATH_INVALID`,
  `CHROMIUM_ELECTRON_SPAWN_FAILED`, `CHROMIUM_ELECTRON_CDP_TIMEOUT`,
  `CHROMIUM_ELECTRON_EXITED_BEFORE_READY`.
  - Unreachable debug ports are distinguished from reached-but-malformed
    responses; the `undici`-wrapped `ECONN*` code is read off `err.cause` and
    mapped to a precise `network_failure` instead of always reporting
    `connection_refused`.
- **Vega (8)** - device discovery/registration, boot timeouts, image lookup,
  input, and screenshot: `VEGA_DEVICE_NOT_FOUND`, `VEGA_DEVICE_NOT_REGISTERED`,
  `VEGA_MULTIPLE_DEVICES`, `VEGA_BOOT_TIMEOUT`, `VEGA_IMAGE_NOT_FOUND`,
  `VEGA_ALREADY_RUNNING`, `VEGA_SCREENSHOT_FAILED`, `VEGA_TEXT_INVALID`.
  - `MultipleVegaDevicesError` extends `FailureError` so callers that re-throw it
    (e.g. `describe`) surface a classified signal.
- **Native profiler (8)** - perfetto start failures (mirroring the iOS xctrace
  set), app-process detection, the metadata sidecar, and the shared analyze
  guard: `NATIVE_PROFILER_PERFETTO_PROCESS_ERROR`,
  `NATIVE_PROFILER_PERFETTO_READY_TIMEOUT`,
  `NATIVE_PROFILER_PERFETTO_READY_EXITED`,
  `NATIVE_PROFILER_APP_PROCESS_NOT_FOUND`, `NATIVE_PROFILER_NO_EXPORTED_TRACE`,
  `NATIVE_PROFILER_TRACE_TEMPLATE_MISSING`, `PROFILER_NATIVE_METADATA_MISSING`,
  `PROFILER_NATIVE_METADATA_INVALID`.
  - The iOS `analyze` guard moves off `PROFILER_NATIVE_TRACE_MISSING` onto the
    shared `NATIVE_PROFILER_NO_EXPORTED_TRACE` so the same logical failure isn't
    split by platform; `PROFILER_NATIVE_TRACE_MISSING` now means strictly a trace
    file missing on disk (and both of its sites use `error_kind: not_found`).
- **Android / simulator / screenshot-diff (3)** -
  `ANDROID_UIAUTOMATOR_CAPTURE_FAILED`, `SIMULATOR_SCREENSHOT_FAILED`,
  `SCREENSHOT_DIFF_INPUT_INVALID`.
- **Flows / keyboard (4)** - `FLOW_PROJECT_ROOT_INVALID` (split from `_REQUIRED`,
  which now means strictly "no root set yet"), `FLOW_NAME_INVALID`,
  `KEYBOARD_KEY_UNSUPPORTED`, `KEYBOARD_CHARACTER_UNSUPPORTED`.

`error_kind` is reserved consistently: `validation` for checks against caller
input or a schema we own; `unknown` for malformed output from an uncontrolled
source (renderer main world, toolkit RPC); `network` with a `network_failure`
class for transport and reached-but-malformed responses; `not_found` for a
missing resource / session / on-disk artifact.

## Testing

- `tsc --build` clean - `error_kind` / `failure_area` / `failure_command` /
  `network_failure` and every `FAILURE_CODES.X` are typed, so the compiler
  enforces structural validity of every signal.
- Unit tests: registry **83/83**, tool-server green (the one red
  `await-ui-element` case is a pre-existing local-only timing flake unrelated to
  this change - it fails identically on `main` and passes in CI).
- `failure-classification.test.ts` pins **17** codes, reading each back via
  `getFailureSignal(err).error_code`; the `CHROMIUM_CDP_INVALID_RESPONSE`
  malformed-200 fix is reproduced against a local HTTP server.
- `prettier --check` and `eslint` clean on all changed files.

## Notes

- No new codes are orphaned - every added `FAILURE_CODES` entry has at least one
  throw site that can reach the telemetry boundary.
- A handful of internal-invariant throws (SQL-template/identifier guards,
  Tesseract-TSV parsing, unreachable `default:` branches) are intentionally left
  as plain `Error`s - they represent programmer/tooling errors, not
  user-reachable failure modes, so they don't warrant telemetry codes.


packages/tool-server/src/utils/vega-inspect.ts
```diff
@@ -94,17 +114,66 @@ function postJson(
...
           if (status < 200 || status >= 300) {
-            reject(new Error(`toolkit HTTP ${status}: ${data.slice(0, 200)}`));
+            // Reachable and answered, just a non-2xx status — a malformed
+            // response, not a down endpoint. Distinct code from the connect-
+            // failure path below so it mirrors the chromium CDP split
+            // (UNREACHABLE vs INVALID_RESPONSE): VEGA_PAGE_SOURCE_FETCH_FAILED
+            // stays exclusively for transport/connect failures.
+            reject(
+              new FailureError(`toolkit HTTP ${status}: ${data.slice(0, 200)}`, {
+                error_code: FAILURE_CODES.VEGA_PAGE_SOURCE_HTTP_ERROR,
```
latekvo: This `VEGA_PAGE_SOURCE_HTTP_ERROR` (and its siblings `_FETCH_FAILED`, `_RPC_ERROR`, `_TIMEOUT`, and the `_PARSE_FAILED` raised here and in source-parser.ts) is only ever thrown by `fetchVegaPageSource` / `parseVegaPageSource`, whose sole non-test caller is `describeVega` (tools/describe/platforms/vega.ts). `describeVega` catches everything except `MultipleVegaDevicesError` and returns an empty tree plus a relaunch hint, so it never throws to the tool boundary. Because the failure signal is only read on the `toolFailed` event (a tool's `execute` throwing), none of these five new codes can reach telemetry — including the genuine toolkit-malfunction cases (HTTP error, RPC error, malformed page source) that this classification looks intended to capture.

packages/tool-server/src/chromium-server/clipboard.ts
```diff
@@ -51,8 +52,14 @@ export async function setClipboardText(cdp: CDPClient, text: string): Promise<vo
   )) as { result?: { value?: { ok?: boolean; error?: string } }; exceptionDetails?: unknown };
   const result = out.result?.value;
   if (!result?.ok) {
-    throw new Error(
-      `Chromium clipboard set failed: ${result?.error ?? "renderer rejected the write"}`
+    throw new FailureError(
+      `Chromium clipboard set failed: ${result?.error ?? "renderer rejected the write"}`,
+      {
+        error_code: FAILURE_CODES.CHROMIUM_CLIPBOARD_FAILED,
```
latekvo: The only caller of `setClipboardText` is the chromium-server Express route at `chromium-server/http-api.ts:60`, which catches the error and responds `res.status(500).json({ error: message })` — only the message string crosses that boundary, so the structured `FailureError` signal is dropped. The other site that throws this same code, `copyScreenshotToClipboard` (screenshot.ts:203), is never invoked anywhere (it's only wired into the server object and declared in the type). As a result `CHROMIUM_CLIPBOARD_FAILED` cannot reach telemetry on any path.

packages/tool-server/src/chromium-server/input.ts
```diff
@@ -139,9 +145,15 @@ export async function sendButton(
     }
     return;
   }
-  throw new Error(
+  throw new FailureError(
     `Chromium does not support the "${button}" hardware button. ` +
-      `Use a keyboard shortcut via the keyboard tool, or invoke an app-level handler via the debugger.`
+      `Use a keyboard shortcut via the keyboard tool, or invoke an app-level handler via the debugger.`,
+    {
+      error_code: FAILURE_CODES.CHROMIUM_BUTTON_UNSUPPORTED,
```
latekvo: The only caller of `sendButton` is the WebSocket `button` handler at `chromium-server/http-api.ts:319`, which is wrapped in a catch that returns `{ status: 'error', message }` — the structured signal doesn't cross back. No registered tool routes a hardware-button press into `sendButton` (the `button` tool declares no chromium capability and uses the simulator-server path), so `CHROMIUM_BUTTON_UNSUPPORTED` never reaches telemetry.

packages/registry/src/failure-codes.ts
```diff
@@ -166,16 +183,58 @@ export const FAILURE_CODES = {
...
   FLOW_ENTRY_UNRECOGNIZED: "FLOW_ENTRY_UNRECOGNIZED",
   PROFILER_QUERY_MODE_INVALID: "PROFILER_QUERY_MODE_INVALID",
   PROFILER_QUERY_REQUIRED_PARAM_MISSING: "PROFILER_QUERY_REQUIRED_PARAM_MISSING",
+  PROFILER_SESSION_ID_INVALID: "PROFILER_SESSION_ID_INVALID",
   PROFILER_DATA_NOT_LOADED: "PROFILER_DATA_NOT_LOADED",
   PROFILER_NATIVE_TRACE_MISSING: "PROFILER_NATIVE_TRACE_MISSING",
+  PROFILER_NATIVE_METADATA_MISSING: "PROFILER_NATIVE_METADATA_MISSING",
+  PROFILER_NATIVE_METADATA_INVALID: "PROFILER_NATIVE_METADATA_INVALID",
+
+  CHROMIUM_CDP_UNREACHABLE: "CHROMIUM_CDP_UNREACHABLE",
```
latekvo, line 203: This PR adds ~53 failure codes and converts ~90 throw sites, but no test file is added or changed (the commit touches 39 files, all under `src/`) and none of the new codes is asserted in any existing test. The description states each throw site was exercised and its `error_code` read back via `getFailureSignal`, but that verification isn't captured in CI, so a future regression in the classification of any of these live codes would go undetected.

packages/tool-server/src/tools/devices/boot-device.ts
```diff
@@ -1198,7 +1210,7 @@ function bootVega(params: {
   // restart, so it must NOT join an in-flight non-force boot (which would skip
   // the restart and hand back the stale device). Two same-mode boots of the same
   // image still share one promise.
-  const key = `${params.vvdImage} ${params.force ? "force" : "normal"}`;
+  const key = `${params.vvdImage} ${params.force ? "force" : "normal"}`;
```
latekvo: This changes the in-flight Vega-boot coalescing Map key separator from a NUL byte to a space, which is unrelated to the failure-code work in this PR. It's harmless in practice (the mode suffix is always `force` or `normal`), but it reads as an incidental edit that doesn't belong to a telemetry-classification change.

latekvo (changes requested):
I reviewed all 39 files. The mechanical conversion is careful and behavior-preserving: `FailureError extends Error` and every message is kept verbatim, the `MultipleVegaDevicesError` to `FailureError` change keeps its `instanceof` and `.name`, the semantic splits (`PROFILER_NATIVE_TRACE_MISSING` to `NATIVE_PROFILER_NO_EXPORTED_TRACE`, the FLOW root split, the Android/iOS profiler param twins) are consistent, there are no duplicate or typo'd codes, and every `failure_stage` / `error_kind` / `failure_command` / `network_failure` value is a valid union member. Builds and suites pass locally (registry 83, tool-server 1596, typecheck clean).

I'm requesting changes for one substantive theme: several of the newly-added codes are thrown only on paths whose errors are swallowed before they reach the `toolFailed` boundary, so they never land in the telemetry bucket this PR is adding them for (the Vega page-source family and the Chromium clipboard/button codes — details inline). There's also no committed test asserting any of the new classifications, and one small out-of-scope edit. Separately, the PR currently shows a merge conflict with `main` in `boot-device.ts` and `vega-vvd.ts` that will need resolving before it can land.

packages/tool-server/src/tools/profiler/query/profiler-load.ts
```diff
@@ -326,17 +332,29 @@ async function loadNativeSession(
...
-      throw new Error(
+      throw new FailureError(
         `No native profiler .pftrace found for session "${sessionId}". ` +
-          `Expected file at ${pftrace}`
+          `Expected file at ${pftrace}`,
+        {
+          error_code: FAILURE_CODES.PROFILER_NATIVE_TRACE_MISSING,
+          failure_stage: "profiler_load_android_pftrace",
+          failure_area: "tool_server",
+          error_kind: "not_found",
```
latekvo, line 342: This newly-classified site emits `PROFILER_NATIVE_TRACE_MISSING` with `error_kind: "not_found"`, but the sibling site at line 416 emits the same code with `error_kind: "validation"` for the same logical condition (the session's persisted trace file is absent on disk). Telemetry grouped by `(error_code, error_kind)` will see this one code split into two kinds by platform, which works against the per-platform consolidation this PR is doing elsewhere (e.g. moving both analyze guards onto a single shared code).

packages/tool-server/src/tools/profiler/react/react-profiler-stop.ts
```diff
@@ -209,9 +209,15 @@ Fails if no active profiling session exists or the CDP connection was lost durin
...
-        throw new Error(
+        throw new FailureError(
           "No active profiling run to stop. The session exists but Hermes CPU sampling was never started — typically because react-profiler-start failed before reaching the sampler (often on release builds without React DevTools). " +
-            "Check the error react-profiler-start returned, address the underlying cause (rebuild in dev mode, reconnect the debugger, etc.), then call react-profiler-start again."
+            "Check the error react-profiler-start returned, address the underlying cause (rebuild in dev mode, reconnect the debugger, etc.), then call react-profiler-start again.",
+          {
+            error_code: FAILURE_CODES.REACT_PROFILER_NO_ACTIVE_SESSION,
+            failure_stage: "react_profiler_stop_inactive",
+            failure_area: "tool_server",
+            error_kind: "validation",
```
latekvo: This newly-classified site emits `REACT_PROFILER_NO_ACTIVE_SESSION` with `error_kind: "validation"`, while the other site for the same code at line 182 uses `error_kind: "not_found"` — so one code now carries two kinds. Separately, the condition here is `!api.profilingActive`, an internal session-state issue (start failed before reaching the sampler) rather than a check against caller-supplied input, which doesn't match how `validation` is used for caller/schema input elsewhere in this PR.

latekvo (commented):
Reviewed the classification changes across all 39 files and exercised a sample of the new throw sites at runtime (reading the code back via `getFailureSignal`, including the CDP network mapping) — the conversions are faithful and behaviour-preserving. Two consistency observations inline.

packages/tool-server/src/tools/profiler/query/profiler-load.ts
```diff
@@ -29,9 +29,15 @@ const SESSION_ID_PATTERN = /^[A-Za-z0-9_-]+$/;
 
 function assertSafeSessionId(sessionId: string): void {
   if (!SESSION_ID_PATTERN.test(sessionId)) {
-    throw new Error(
+    throw new FailureError(
       `Invalid session_id "${sessionId}". Allowed: letters, digits, '_' and '-' ` +
-        `(no path separators, no "..").`
+        `(no path separators, no "..").`,
+      {
+        error_code: FAILURE_CODES.PROFILER_SESSION_ID_INVALID,
```
latekvo: `session_id` is already validated by the Zod schema with this same `SESSION_ID_PATTERN` regex (`session_id: z.string()…regex(SESSION_ID_PATTERN, …)`), and Zod runs before `execute`. Any value that would fail `assertSafeSessionId` is therefore rejected earlier as an unclassified validation error, so this classified throw can only fire on a direct call that bypasses the schema — never on a registry-invoked (telemetry-bearing) path. As a result `PROFILER_SESSION_ID_INVALID` can never appear in telemetry, which is at odds with the goal that every added code can bucket a real failure.

packages/tool-server/src/tools/screenshot-diff/index.ts
```diff
@@ -207,37 +208,73 @@ async function resolveInputPaths(
...
+// simulatorServer is declared as an unconditional service dependency, so the
+// registry resolves it before execute() runs. Guard anyway: executeScreenshotDiffTool
+// is exported and a direct caller (e.g. a test) can pass a services map without it —
+// a classified failure beats a downstream TypeError on `.captureScreenshot`. Only
+// the live-capture branches call this, so non-capture diffs never require the service.
+function requireSimulatorServer(services: Record<string, unknown>): SimulatorServerApi {
+  const api = services.simulatorServer as SimulatorServerApi | undefined;
+  if (!api) {
+    throw new FailureError("Live screenshot capture requires a simulatorServer service.", {
+      error_code: FAILURE_CODES.SCREENSHOT_DIFF_CAPTURE_UNAVAILABLE,
```
latekvo: `simulator-server` is declared as an unconditional service dependency of `screenshot-diff`, and the registry resolves declared services before `execute` runs, so `services.simulatorServer` is always present on any registry-invoked path. As the comment just above this guard notes, it can only trip for a direct/test caller — which never reaches the telemetry boundary — so `SCREENSHOT_DIFF_CAPTURE_UNAVAILABLE` cannot be emitted in telemetry.

packages/tool-server/src/utils/vega-cli.ts
```diff
@@ -212,6 +219,12 @@ export async function vegaDevice(
   subcommand: string[],
   options: { timeoutMs?: number } = {}
 ): Promise<VegaRunResult> {
-  if (!serial) throw new Error("vegaDevice requires a non-empty device serial");
+  if (!serial)
+    throw new FailureError("vegaDevice requires a non-empty device serial", {
+      error_code: FAILURE_CODES.VEGA_DEVICE_ID_INVALID,
```
latekvo: `vegaDevice` is only reached for a device that `classifyDevice` routed to Vega, which requires the non-empty `amazon-` serial prefix, and every caller forwards that serial straight through. `serial` is therefore always non-empty on any real path, so the `!serial` guard (and `VEGA_DEVICE_ID_INVALID`) can only fire from a direct call, never through the telemetry boundary.

packages/tool-server/src/tools/profiler/native-profiler/platforms/android.ts
```diff
@@ -100,17 +109,50 @@ export async function stopNativeProfilerAndroid(
...
 
   if (!api.traceFile || !api.androidOnDeviceTracePath || !api.capturePid) {
     if (recoveringPartialTrace) {
-      throw new Error(
+      throw new FailureError(
         "Native profiling recording exited unexpectedly and no trace file is available. " +
-          "Call native-profiler-start again."
+          "Call native-profiler-start again.",
+        {
+          error_code: FAILURE_CODES.NATIVE_PROFILER_RECORDING_CRASHED,
```
latekvo: This throw requires `recoveringPartialTrace` together with a missing trace handle, but on Android that combination can't occur: `recordingExitedUnexpectedly` is only ever set true in the iOS path, and the Android recording-timeout callback sets `recordingTimedOut = true` while leaving `traceFile` / `androidOnDeviceTracePath` / `capturePid` intact (and the stop `finally` resets `recordingTimedOut`). So `NATIVE_PROFILER_RECORDING_CRASHED` is unreachable here, and an actual perfetto crash mid-recording on Android ends up unclassified by this code.

packages/tool-server/src/blueprints/chromium-cdp.ts
```diff
@@ -154,21 +156,39 @@ export const chromiumCdpBlueprint: ServiceBlueprint<ChromiumCdpApi, DeviceInfo>
     const deviceFromOpts = opts?.device;
     const payloadStr = typeof payload === "string" ? payload : (payload as DeviceInfo)?.id;
     if (deviceFromOpts && payloadStr && deviceFromOpts.id !== payloadStr) {
-      throw new Error(
-        `${CHROMIUM_CDP_NAMESPACE}.factory: options.device.id "${deviceFromOpts.id}" disagrees with URN payload "${payloadStr}".`
+      throw new FailureError(
+        `${CHROMIUM_CDP_NAMESPACE}.factory: options.device.id "${deviceFromOpts.id}" disagrees with URN payload "${payloadStr}".`,
+        {
+          error_code: FAILURE_CODES.CHROMIUM_FACTORY_OPTIONS_INVALID,
```
latekvo: Both throw sites for this code are unreachable on a registry path. Tools build their Chromium service ref with `chromiumCdpRef(device)`, which sets the URN to `ChromiumCdp:${device.id}` and `options.device` to that same `device` — so the mismatch guard (`deviceFromOpts.id !== payloadStr`) compares a value against itself and can't fire, and the transitive-dep path passes no options so the guard is skipped. The `!device` guard just below is likewise dead (`resolveDevice` never returns null and the URN payload is always present). So `CHROMIUM_FACTORY_OPTIONS_INVALID` can only be produced by a direct `factory()` call with a hand-crafted mismatch (the unit test), never through telemetry. (The adjacent `CHROMIUM_DEVICE_ID_INVALID` at the port-parse guard is genuinely reachable.)

latekvo (commented):
Reviewed the classification changes end-to-end: built the workspace, ran the registry/tool-server suites, and traced each new code's throw site to the telemetry boundary (`getFailureSignal` via `registry.invokeTool`). The mechanical conversion is faithful and the semantic splits are consistent. A few inline notes on codes whose throw sites can't actually fire on a registry/telemetry path, so they won't bucket a real failure.

packages/tool-server/src/utils/vega-inspect.ts
```diff
@@ -94,17 +114,66 @@ function postJson(
...
+        code === "ECONNREFUSED"
+          ? "connection_refused"
+          : code === "ECONNRESET"
+            ? "connection_reset"
+            : "other";
+      reject(
+        new FailureError(
+          `toolkit request failed: ${err instanceof Error ? err.message : String(err)}`,
+          {
+            error_code: FAILURE_CODES.VEGA_PAGE_SOURCE_FETCH_FAILED,
```
latekvo: This `VEGA_PAGE_SOURCE_FETCH_FAILED` and its siblings (`_HTTP_ERROR` :124, `_RPC_ERROR` :65, `_TIMEOUT` :169, and `_PARSE_FAILED` here and in source-parser.ts) are only ever thrown by `fetchVegaPageSource` / `parseVegaPageSource`, whose sole production consumer is `describeVega` (describe/platforms/vega.ts). That function catches every error except `MultipleVegaDevicesError` and returns `{ tree: EMPTY_TREE, hint }`, so the `describe` tool's `execute()` resolves normally and the registry emits `toolCompleted`, never `toolFailed`. `getFailureSignal` is never consulted, so none of these five codes can reach the telemetry bucket they were added for — and `vega-describe.test.ts` pins that swallow (ECONNREFUSED -> empty tree, not a throw).

packages/tool-server/src/chromium-server/clipboard.ts
latekvo: `CHROMIUM_CLIPBOARD_FAILED`: the only production caller of `setClipboardText` is the chromium-server Express route `POST /api/clipboard/text` (http-api.ts:60), whose catch responds `res.status(500).json({ error: message })` — the structured signal is discarded there and this router is mounted on the standalone HTTP app, not invoked through any registered tool. The other throw site (`copyScreenshotToClipboard`, screenshot.ts) has no production caller at all. So this code never reaches `getFailureSignal`.

packages/tool-server/src/chromium-server/input.ts
latekvo: `CHROMIUM_BUTTON_UNSUPPORTED`: the only caller of `sendButton` is the WebSocket `button` handler (http-api.ts:319), wrapped in a catch that replies `{ status: "error", message }`. The registered `button` tool is apple+android only (its chromium button set is empty) and routes to the simulator-server, so it can't reach this path. Like the clipboard code, the classified signal is caught and reformatted before any registry boundary, so it never buckets.

packages/tool-server/src/chromium-server/cdp-session.ts
```diff
@@ -73,17 +88,76 @@ export async function browserWebSocketUrl(port: number, signal?: AbortSignal): P
...
+    throw new FailureError(`Chromium CDP discovery: GET ${url} failed (HTTP ${res.status})`, {
+      error_code: FAILURE_CODES.CHROMIUM_CDP_INVALID_RESPONSE,
+      failure_stage: "chromium_cdp_discovery_fetch",
+      failure_area: "tool_server",
+      error_kind: "network",
+      failure_command: "cdp",
+      network_failure: "invalid_response",
+    });
   }
   return (await res.json()) as T;
```
latekvo: `return (await res.json())` sits outside the try/catch that classifies fetch failures, and the `!res.ok` branch above only handles non-2xx. A reachable-but-malformed HTTP 200 body (a non-CDP service squatting the debug port, or a truncated/reset body) makes `res.json()` reject with a raw `SyntaxError` that escapes `fetchJson` unclassified and falls back to the generic bucket — which is exactly the "reached but malformed" case `CHROMIUM_CDP_INVALID_RESPONSE` was added to cover for the non-2xx and missing-`webSocketDebuggerUrl` paths. Reproduced with a local server returning 200 + non-JSON.

packages/tool-server/src/tools/profiler/react/react-profiler-stop.ts
latekvo: This `REACT_PROFILER_NO_ACTIVE_SESSION` is emitted with `error_kind: "validation"`, while the sibling throw for the same code at line 182 uses `error_kind: "not_found"`. The iOS/Android profiler sites you changed to `not_found` in this PR cite this react twin as the canonical `not_found` for the "no active session" family (see the comments there), so the same code now carries two different kinds and the stated unification is broken. Not caught by tsc since both are valid `FailureKind` members.

packages/tool-server/src/tools/devices/boot-device.ts
```diff
@@ -1199,7 +1211,7 @@ function bootVega(params: {
...
```
latekvo: This changes the in-flight Vega-boot coalescing Map-key separator from a NUL byte to a space (confirmed at the byte level against the merge-base), which is unrelated to the failure-code work in this PR. It's harmless in practice — the key is only ever compared to keys built the same way in the same function — but it's out of scope for a telemetry-classification change.

packages/tool-server/test/failure-classification.test.ts
```diff
@@ -0,0 +1,119 @@
...
+  try {
+    fn();
+  } catch (e) {
+    return e;
+  }
+  throw new Error("expected the call to throw, but it returned");
+}
+
+function expectCode(err: unknown, code: FailureCode): void {
+  expect(getFailureSignal(err)?.error_code).toBe(code);
```
latekvo, line 65: Good to have a test that reads codes back exactly how telemetry does (`getFailureSignal(err).error_code`), but it only pins ~7 of the 46 added codes (flows, keyboard, the chromium factory). The runtime families that are the bulk of this PR — the CDP/electron/input/viewport/tabs/storage/describe Chromium codes, the new Vega codes, and the native-profiler perfetto/metadata codes — have no assertion tying a throw site to its code, so a future regression in any of those classifications wouldn't be caught here.

latekvo (changes requested):
Re-reviewed the full diff independently: traced each new/changed throw site to the registry telemetry boundary (`registry.invokeTool` -> `toolFailed` -> `getFailureSignal`), reproduced the CDP/undici network-mapping paths, and re-ran the registry + tool-server suites (green). The mechanical `Error` -> `FailureError` conversion is careful and behaviour-preserving — `FailureError extends Error`, messages are verbatim, `MultipleVegaDevicesError` keeps its `instanceof`/`.name`, the semantic splits (`NATIVE_PROFILER_NO_EXPORTED_TRACE`, the FLOW root split, the profiler param twins) are consistent, and none of the 46 added codes is orphaned or typo'd.

Requesting changes for the same substantive theme as the earlier review, which is still present at this tip: several of the newly-added codes are thrown only on paths that are caught and reformatted before they can reach the telemetry boundary this PR is adding them for, so they can never bucket a real failure — the `VEGA_PAGE_SOURCE_*` family and the two Chromium `CHROMIUM_CLIPBOARD_FAILED` / `CHROMIUM_BUTTON_UNSUPPORTED` codes (details inline). Alongside that: a reached-but-malformed CDP response that escapes the classification this PR set out to add, and one code emitted with two different `error_kind`s.

Two smaller points not requiring a re-review but worth fixing: the PR description's per-area code lists and the "53 new codes" count are stale relative to the 46 actually added (7 names in the description — e.g. `VEGA_CLI_NOT_FOUND`, `SCREENSHOT_DIFF_CAPTURE_UNAVAILABLE`, `PROFILER_SESSION_ID_INVALID` — don't exist in the code; those paths were deliberately left as plain `Error`s), and one unrelated edit is bundled in (inline). The added `failure-classification.test.ts` is a good start but only pins ~7 of the 46 codes; the runtime Chromium/Vega/native-profiler families that make up most of the PR are unasserted.

packages/tool-server/src/tools/profiler/query/profiler-load.ts
latekvo, line 342: Same pattern as the react-profiler note: this newly-added Android `.pftrace`-missing site emits `PROFILER_NATIVE_TRACE_MISSING` with `error_kind: "not_found"`, while the pre-existing iOS site at line 416 emits the same code with `error_kind: "validation"` for the analogous "trace file missing on disk" case. So this PR introduces a second `error_kind` for one code — telemetry keyed on `error_code` for this failure now sees a mix depending on platform.

packages/tool-server/src/tools/devices/boot-device.ts
latekvo: Following up at the current tip: this separator was restored to the NUL byte in an earlier commit on the branch, but the final `fix: remove artifact` commit changed it back to a space, so `HEAD` still differs from `main` here (space vs NUL) even though the PR description states the NUL byte was reverted. It is functionally harmless — no realistic `vvdImage` combined with the ` force`/` normal` suffix can produce a colliding coalescing key under a space — so this is just a heads-up that the branch tip and the description are out of sync on this line.

packages/tool-server/src/chromium-server/http-api.ts
```diff
@@ -331,6 +332,11 @@ async function handleWsCommand(
       return {};
     }
     default:
-      throw new Error(`Unknown ws cmd: ${msg.cmd}`);
+      throw new FailureError(`Unknown ws cmd: ${msg.cmd}`, {
+        error_code: FAILURE_CODES.CHROMIUM_PARAM_INVALID,
```
latekvo: This `default` branch throws a classified `CHROMIUM_PARAM_INVALID`, but its only caller (`handleWsCommand`, invoked in the `ws.on("message")` handler just above) catches the throw and reformats it to `{ id, status: "error", message }`, keeping only the message string. The structured signal is dropped here before it can reach the `toolFailed` boundary, and this WebSocket server isn't driven through any registered tool. The `CHROMIUM_PARAM_INVALID` code still buckets via its reachable tool sites (chromium-cookies / chromium-storage / chromium-tabs), so no telemetry coverage is lost — but the classification added at this specific site is inert, the same swallowed-path situation the adjacent `sendButton` was deliberately left a plain `Error` for.

packages/tool-server/src/chromium-server/tabs.ts
```diff
@@ -154,7 +167,18 @@ export function createTabsManager(deps: TabsManagerDeps): TabsManager {
...
-      if (!out.targetId) throw new Error("Target.createTarget returned no targetId.");
+      if (!out.targetId)
+        throw new FailureError("Target.createTarget returned no targetId.", {
+          error_code: FAILURE_CODES.CHROMIUM_TAB_OPEN_FAILED,
+          failure_stage: "chromium_tab_open",
+          failure_area: "tool_server",
+          // The CDP call round-tripped but its response was missing the expected
+          // targetId — a "reached but malformed" response, classified the same
+          // way as CHROMIUM_CDP_INVALID_RESPONSE rather than as a generic unknown.
+          error_kind: "network",
```
latekvo: `Target.createTarget` is a CDP WebSocket command that round-tripped successfully and returned a well-formed object merely missing `targetId`, yet this is classified `error_kind: "network"` with `network_failure: "invalid_response"`. The structurally identical cases in the same server — `Page.captureScreenshot` returning no `data` (`CHROMIUM_SCREENSHOT_FAILED`) and the renderer-eval failures in `viewport.ts` / `storage.ts` / the chromium `describe` path — are classified `error_kind: "unknown"`, matching the convention documented at the top of `viewport.ts`. The `network` / `invalid_response` bucket is otherwise used only for the HTTP `/json` discovery layer, where there is a genuine fetch transport, so a failed tab-open would group in telemetry with port-unreachable / HTTP-error transport failures rather than a CDP-command / browser-state one.

latekvo (approved):
Re-reviewed the full diff independently at this tip, and the earlier requested changes are addressed. I rebuilt the workspace (`tsc --build` clean), ran the registry (83) and tool-server suites green — the single red `lens-tools-platform-gate` case is a pre-existing full-suite import-isolation flake that passes in isolation and isn't touched here — the new `failure-classification.test.ts` passes 17/17, and I traced every one of the 39 added codes from its throw site to the registry telemetry boundary (`invokeTool` -> `toolFailed` -> `getFailureSignal`). I also drove the CDP-discovery family end-to-end through a real `invokeTool` -> `toolFailed` -> `getFailureSignal` to confirm `CHROMIUM_CDP_INVALID_RESPONSE` (a malformed HTTP-200 body), `CHROMIUM_CDP_UNREACHABLE` (`connection_refused`), and `CHROMIUM_CDP_NO_PAGE_TARGET` actually bucket.

Everything from the prior reviews is resolved: the `VEGA_PAGE_SOURCE_*` / `CHROMIUM_CLIPBOARD_FAILED` / `CHROMIUM_BUTTON_UNSUPPORTED` codes and the earlier unreachable ones (`PROFILER_SESSION_ID_INVALID`, `SCREENSHOT_DIFF_CAPTURE_UNAVAILABLE`, `VEGA_DEVICE_ID_INVALID`, `NATIVE_PROFILER_RECORDING_CRASHED`, `CHROMIUM_FACTORY_OPTIONS_INVALID`) are back to plain `Error`s; the reached-but-malformed CDP 200 body is now classified (with `AbortError` still surfaced untouched); `REACT_PROFILER_NO_ACTIVE_SESSION` and `PROFILER_NATIVE_TRACE_MISSING` each carry a single consistent `error_kind` again; `MultipleVegaDevicesError` -> `FailureError` preserves `instanceof` / `.name` and its `describe` re-throw; every `error_kind` / `failure_command` / `network_failure` literal is a valid union member; and all 39 codes are reachable on at least one un-swallowed registered-tool path.

Approving. Two small, non-blocking classification points inline.

PR #421, opened 2026-06-26
Windows support: Android + Chromium host control plane

latekvo:
## What

Makes Argent run on Windows. iOS Simulator is macOS-only (Apple), so on Windows that means the **Android + Chromium** host control plane. The blocker was twofold: (1) no Windows `simulator-server` binary, and (2) POSIX-only assumptions in the tool-server.

Companion radon PR builds the binary and runs the Android emulator E2E on Windows: software-mansion/radon#134.

## Changes

**Windows-aware tool-server / resolver / bundling**
- New shared `commandOnPath()` helper — `where` on Windows, `command -v` via `/bin/sh` on POSIX. `android-binary`, `check-deps`, and `vega-cli` previously hardcoded `/bin/sh`, which never matches on Windows.
- `android-binary`: append `.exe` to `adb`/`emulator` in the `$ANDROID_HOME` fallback; probe `%LOCALAPPDATA%\Android\Sdk` (Android Studio's Windows default).
- simulator-server resolver + dispatcher + `bundle-tools` + `download-simulator-server.sh`: ship and resolve `simulator-server.exe` on `win32` (new `simulatorServerBinaryName()`), add `win32` to the bundled host keys, add the `simulator-server-argent-windows.exe` download target with PE arch validation.
- `stop-metro`: resolve listening PIDs via `netstat -ano` on Windows (`lsof` is POSIX-only and threw ENOENT).
- `adb.ts`: AVD root resolution used `$HOME` + `startsWith("/")` (rejects `C:\…`); now `os.homedir()` + `path.isAbsolute()`.
- **`boot-device`**: the emulator was spawned `detached` with `stdio:"ignore"`; on Windows a detached emulator with NUL stdout/stderr never reaches `sys.boot_completed` (verified on `windows-latest`: the identical flag set boots fine with real handles, hangs with NUL). Redirect the detached emulator's output to a temp log file on win32. POSIX keeps `"ignore"`.

**Tests** — cross-platform unit coverage for `commandOnPath` (both branches), win32 `.exe` Android resolution, win32 `.exe` simulator-server resolution, and a win32 AVD drive-path case; dep-gate tests now mock at the `commandOnPath` boundary so they're platform-agnostic.

**CI** — `windows-e2e.yml` (`windows-chromium` on `windows-latest`): runs the resolver/dep unit tests natively on Windows, confirms `adb.exe` resolution via the SDK root, then drives a **real headless Chrome over CDP** through the tool-server (discover → screenshot → describe → tap → observe the DOM mutation). Flow in `scripts/ci/windows-chromium-e2e.mjs` (also runnable locally / in Docker). The `simulator-server.exe` build + the full **Android-emulator E2E on Windows** live in radon CI (the binary's source is the private radon repo).

## Verification (all green)
- **Chromium control plane on real Windows** (`windows-latest`): `✓ discovered chromium-cdp-9222 → ✓ screenshot → ✓ describe → ✓ gesture-tap → ✓ DOM mutation observed`. Resolver/dep unit tests run natively; `adb.exe` resolves via the SDK root (it's *not* on PATH on the runner — the gap the resolver closes).
- **Android emulator on real Windows** (radon#134, `windows-latest` / WHPX): the real Argent path — `boot-device` → `{ booted: true }` → screenshot via `simulator-server.exe` → gesture-tap — all green. (The boot-device fix above is what makes the detached emulator reach `boot_completed`.)
- **Local clean-room E2E**: the Chromium flow runs green in a headless Linux Docker container (fresh `npm ci`, no host pollution).
- **No regressions**: full tool-server suite (1624) + native-devtools-ios (16); ESLint/Prettier/typecheck clean; **Wayland E2E** (Linux AVD boot) and **Vega VVD E2E** still pass (they exercise the changed `adb.ts` / `boot-device.ts` / `vega-cli.ts` paths).

## Notes
Android-emulator E2E needs WHPX nested virtualization — confirmed available on the current hosted `windows-latest` image (Server 2025); the emulator boots hardware-accelerated in ~3 min.


packages/tool-server/src/utils/command-on-path.ts
```diff
@@ -0,0 +1,56 @@
...
+      const match = stdout
+        .split(/\r?\n/)
+        .map((line) => line.trim())
+        .filter(Boolean)
+        // Skip the CWD entry `where` lists ahead of PATH — Windows paths are
+        // case-insensitive, so compare normalized + lowercased directories.
+        .find((candidate) => pathWin32.resolve(pathWin32.dirname(candidate)).toLowerCase() !== cwd);
+      return match ?? null;
+    }
+    const { stdout } = await execFileAsync("/bin/sh", ["-c", `command -v ${name}`], {
```
filip131311, line 54: **Hardening suggestion: validate `name` before the `/bin/sh -c` interpolation.**

The POSIX branch builds a shell string — `command -v ${name}` — and hands it to `/bin/sh -c`. If `name` ever contained shell metacharacters (`;`, `$(...)`, backticks), the shell would evaluate them. All current callers pass fixed literals (`adb`, `emulator`, `xcrun`, the vega CLI names), and this interpolation is pre-existing behavior inherited from the three private copies — so nothing is exploitable today.

But this PR promotes it into one exported, general-purpose utility, which invites future callers with less-trusted input. Cheap hardening while it's being centralized:

```ts
export async function commandOnPath(name: string): Promise<string | null> {
  // Command names are bare binary names. Rejecting anything else keeps the
  // POSIX branch's `/bin/sh -c` interpolation safe if a future caller passes
  // untrusted input, and keeps `where`'s glob matching (`adb*`) from returning
  // something the caller didn't ask for.
  if (!/^[A-Za-z0-9_.-]+$/.test(name)) return null;
  ...
```

Notes:
- The risk surface is POSIX-only — the Windows branch uses `execFile("where", [name])` with no shell, so the name is inert there. The guard is still worth applying to both branches: `where` does glob/pattern matching, so a stray `adb*` or a path-shaped string could resolve to something unexpected.
- Behavior-neutral for every existing caller on every platform; it only turns a hypothetical hostile/malformed string into a clean `null` instead of a shell evaluation.

Fine as a follow-up if you'd rather not touch this PR — flagging it here since the centralization is what widens the future caller surface.

.github/workflows/windows-e2e.yml
```diff
@@ -0,0 +1,186 @@
...
+      - "packages/tool-server/src/utils/chromium-discovery.ts"
+      - "packages/tool-server/src/tools/simulator/stop-metro.ts"
+      - "packages/tool-server/src/tools/devices/boot-device.ts"
+      # The unit tests this job runs — editing one must re-run it here.
+      - "packages/tool-server/test/command-on-path.test.ts"
+      - "packages/tool-server/test/android-binary-windows.test.ts"
+      - "packages/tool-server/test/adb-resolve-avd-path.test.ts"
+      - "packages/tool-server/test/check-deps.test.ts"
+      - "packages/tool-server/test/http-dep-gate.test.ts"
+      - "packages/tool-server/test/stop-metro-netstat.test.ts"
```
filip131311, line 43: **Path filter misses `device-shutdown.ts` (changed in this PR) and its test.**

`packages/tool-server/src/utils/device-shutdown.ts` gained a win32-relevant change in this PR (adb now resolved via `resolveAndroidBinary` instead of relying on a bare `adb` on PATH — which on Windows usually isn't there), but it's not in this `paths:` list, so a future edit to it won't re-run the Windows job. Same for its test, `packages/tool-server/test/device-shutdown.test.ts` (also touched here).

Suggest adding both:

```yaml
      - "packages/tool-server/src/utils/device-shutdown.ts"
      - "packages/tool-server/test/device-shutdown.test.ts"
```

Since that test is now platform-agnostic (it mocks `resolveAndroidBinary`), it could also join the curated vitest list in the "Resolver unit tests on Windows" step so the fallback-to-bare-`adb` behavior is exercised natively on a Windows host — optional, but the path-filter entries are the part that matters.

scripts/dev.cjs
```diff
@@ -146,8 +146,12 @@ console.log("✓ Dispatcher TypeScript built\n");
 const HOST_PLATFORM_KEY =
   process.platform === "linux" && process.arch === "arm64" ? "linux-arm64" : process.platform;
 const BIN_DIR = path.join(ARGENT_PKG, "bin", HOST_PLATFORM_KEY);
-const BIN_SRC = path.join(NATIVE_DEVTOOLS_PKG, "bin", HOST_PLATFORM_KEY, "simulator-server");
-const BIN_DEST = path.join(BIN_DIR, "simulator-server");
+// win32 ships a PE `.exe` (simulator-server.exe); every other host an
+// extensionless binary. Mirrors serverBinaryBasename() in
+// @argent/native-devtools-ios and bundle-tools.cjs.
```
filip131311, line 151: **Stale cross-reference: `serverBinaryBasename()` doesn't exist — the exported function is `simulatorServerBinaryName()`.**

Worth fixing because these comment cross-references are doing real work: the `simulator-server.exe` name literal is now inlined in five places (the resolver in `@argent/native-devtools-ios`, the dispatcher shim `argent-simulator-server.cjs`, `bundle-tools.cjs`, this file, and `download-simulator-server.sh`). The shim and the shell script legitimately can't import the helper, so the comments pointing back at the canonical function are the only thing keeping the five copies in sync — a pointer at a name that doesn't exist sends the next editor grepping for nothing.

(The two win32 resolver tests do pin the canonical name — `expect(r.simulatorServerBinaryName()).toBe("simulator-server.exe")` — which helps, but only for the resolver copy.)

```suggestion
// win32 ships a PE `.exe` (simulator-server.exe); every other host an
// extensionless binary. Mirrors simulatorServerBinaryName() in
// @argent/native-devtools-ios and bundle-tools.cjs.
```

packages/tool-server/src/utils/command-on-path.ts
latekvo: Fixed in 91ec5978. Added the guard at the top of `commandOnPath` — `if (!/^[A-Za-z0-9_.-]+$/.test(name)) return null;` — so a name with shell/glob metacharacters short-circuits to `null` before either the POSIX `/bin/sh -c` interpolation or the win32 `where` glob match. Behaviour-neutral for the fixed-literal callers; a new test asserts `execFile` is never spawned for a metacharacter name (verified it fails without the guard).

.github/workflows/windows-e2e.yml
latekvo: Fixed in 91ec5978. Added both `packages/tool-server/src/utils/device-shutdown.ts` and `packages/tool-server/test/device-shutdown.test.ts` to the `paths:` filter, and — since that test is now platform-agnostic (mocks `resolveAndroidBinary`) — also added it to the curated vitest list so the fallback-to-bare-`adb` behaviour runs natively on the Windows runner.

scripts/dev.cjs
latekvo: Fixed in 91ec5978. Corrected the comment to point at the real exported helper, `simulatorServerBinaryName()` — `serverBinaryBasename()` existed nowhere in the tree.

filip131311 (approved):
Reviewed with a focus on regressions to the existing macOS/Linux paths — found none. Every shared-code change is behavior-preserving or a strict superset:

- **`commandOnPath()` extraction** (`android-binary`, `check-deps`, `vega-cli`): POSIX branch is the old `/bin/sh -c "command -v"` logic verbatim, same timeout. The new name-validation regex admits every existing caller (`adb`, `xcrun`, `emulator`, `sim-remote`, `vega`, `kepler`). The tightened non-empty-stdout requirement is harmless — POSIX `command -v` always prints on success.
- **`adb.ts`**: `os.homedir()` ≡ `$HOME` when set (superset when unset); `isAbsolute()` ≡ `startsWith("/")` on POSIX.
- **`device-shutdown`**: resolver-first with bare-`adb` fallback — PATH still wins inside the resolver, so anyone who worked before resolves identically.
- **`stop-metro`**: POSIX `lsof -sTCP:LISTEN` path is a pure refactor.
- **`boot-device`**: stdio change fully fenced behind `win32`.
- **Packaging/publish**: win32 additive everywhere — copy-if-exists in bundle-tools, `ARGENT_REQUIRE_WINDOWS_BINARY: "0"` gate, download script `continue`s on the missing asset — so releases keep working before radon#134 publishes the `.exe`.

CI corroborates: all green, including the suites exercising the changed shared paths on existing platforms (Wayland AVD, Vega VVD, macOS iOS E2E, Linux/macOS Chromium E2E, full unit suite).

The Windows-specific work is unusually careful — locale-independent netstat parsing (wildcard-foreign-endpoint instead of localized "LISTENING", with German/French fixtures), the `:18081`-vs-`:8081` guard, the maxBuffer bump, and the CWD-skip in `where` resolution closing a planted-binary hazard.

Minor, non-blocking:
- `boot-device` on win32 writes one `argent-emulator-*.log` per boot into temp and never cleans up — a fixed per-AVD name (truncate on open) would avoid accumulation.
- `windows-e2e.yml` triggers on PRs touching widely-shared files and depends on Chrome being preinstalled on `windows-latest` — new flake exposure for unrelated PRs; acceptable, just noting it.
- Stale docstring: `listeningPids()` says Windows rows are filtered "in the LISTENING state", but the implementation deliberately avoids the State column (the parser's own docstring explains why).

latekvo:
Thanks for the careful pass @filip131311. Addressed the two concrete nits in b5035758:

- **`boot-device` temp-log accumulation** — the win32 emulator redirect named the log \`argent-emulator-<avd>-<Date.now()>.log\` and opened it append-mode, so every boot dropped a fresh never-cleaned log into temp. Now a fixed per-AVD name (\`argent-emulator-<avd>.log\`) opened with \`"w"\` (truncate). Only one emulator per AVD can run at once (the AVD is lock-guarded), so there is no concurrent writer to clobber — repeated boots just reuse and overwrite the one file. Comment updated to explain the naming/truncate choice.
- **Stale `listeningPids()` docstring** — corrected: it now describes the actual win32 logic (keep TCP rows whose local address ends in `:<port>` and whose foreign endpoint is a wildcard, the locale-independent listener signature) instead of the wrong "LISTENING state" filter, and points at `parseNetstatListeningPids` for the why-not-the-State-column rationale.

- **`windows-e2e.yml` trigger breadth + preinstalled Chrome** — leaving as-is intentionally. The broad `paths:` filter is deliberate: the shared resolver/dep files it lists (`command-on-path`, the Android/simulator-server resolvers, `stop-metro`, `adb.ts`, `boot-device.ts`) are exactly the surfaces where a Windows regression would hide, and the trigger was just broadened to `native-devtools-ios` tests for the same reason — narrowing it to buy fewer runs would give back the coverage this PR exists to add. Chrome is part of the GitHub-hosted `windows-latest` image, so the preinstall dependency is a runner guarantee rather than a per-PR setup step. Noting the tradeoff is fair; happy to revisit if it turns out to actually flake.

PR #426, opened 2026-06-30
feat(ios): pass ax identifier to the `DescribeNode`

j-piasecki:
Requires https://github.com/software-mansion/argent-private/pull/21

Passes the identifiers read from the ax-tree to `DescribeNodes`


filip131311 (approved):
👍

PR #427, opened 2026-06-30
feat(tool-server): add find tool to locate an element and optionally act on it

hubgan:
## Summary

Adds a **`find`** tool that locates a UI element by a locator and optionally performs an action on the match, collapsing the usual `describe` -> read coordinates -> `gesture-tap` cycle into a single call. Works on iOS, Android, and Chromium (not Vega, which has no touch model).

```
find { query: "Sign In", by: "text", action: "tap" }
```

- **Locators** (`by`): `any` (label/value/id, the default), `text` (label or value), `label`, `value`, `role`, `id`. Case-insensitive substring.
- **Actions** (`action`): `tap` (default, centre), `focus`, `type` (focus + type), `fill` (focus + clear + type), `exists` (single presence check), `wait` (block until visible), `get-text`, `get-attrs`.
- Returns `{ found, matchCount, match?, actionResult?, elapsed, note? }`. `matchCount` flags ambiguous queries; `index` disambiguates. A single tree snapshot is taken by default; `timeoutMs` opts into polling until the element appears before acting.

## How it's built

- Delegates the device effect to the existing `gesture-tap` / `keyboard` tools via `invokeSubTool` (free telemetry attribution; no duplicated transport; read-only actions never spawn the simulator-server).
- Extracts the shared, hardened tree-matching primitives into `describe/match.ts` (root-excluding walk, reading-order sort, visibility proxy) and the tree fetch + `settleWithin` into `describe/fetch-tree.ts` / `utils/timing.ts`, reused by both `find` and `await-ui-element` (the latter shrinks by ~170 lines, behaviour-preserving).
- Per-platform focus settle so the first keystroke isn't dropped while a soft keyboard raises; a centre-tap clear for `fill` (verified to fully clear fields on iOS and Android); fetch bounded by the deadline + abort signal so `wait` can't overshoot `timeoutMs`.
- `find` is intentionally kept out of `run-sequence` for v1: unlike `await-ui-element`, a missed `find` returns `{ found: false }` without throwing, which the sequence stop-guard wouldn't catch (documented inline). The skill/flow docs are updated to match.

## Verification

- `vitest`: new `find.test.ts` (38 cases) plus `await-ui-element` / `run-sequence` / `sub-invoke` / flows all green; `tsc --noEmit` and `eslint` clean.
- End-to-end on **iPhone 16 Pro** and **Pixel 3a** against a real app: `exists`, `get-attrs`, `tap` (verified tab switch via before/after screenshots), `type`, `fill` (clean clear + replace), and `wait` (hit + timeout). The device run caught and fixed an Android first-keystroke-drop race.

Draft for review.


latekvo (commented):
Reviewed end-to-end: built and ran the tool-server suite, drove the compiled `find` against a real device's accessibility tree (it selected the correct element, flagged the ambiguity, and the computed tap landed on the intended row), and exercised the edge paths directly. The shared-primitive extraction and the await-ui-element refactor are behaviour-preserving, and the locator/selection logic is sound. Two inline notes.

packages/tool-server/src/tools/find/index.ts
```diff
@@ -0,0 +1,555 @@
...
+            params.udid,
+            match.tapPoint,
+            focusSettleMs(device.platform)
+          );
+          if (!focused) return cancelled();
+          const clearedChars = await clearField(
+            registry,
+            ctx,
+            params.udid,
+            chosen.value?.length ?? 0
```
latekvo: `fill` derives the number of backspaces to clear the field from `chosen.value?.length`, but on several platforms a populated field's current editable text is not in `node.value`. On Chromium an `<input>`'s text is exposed as `node.label` (accessibleName → `el.value`/placeholder), while `node.value` is `ownText()`, which is empty for an input; on Android an EditText without a `content-desc` likewise reports its text as `label` with `value` unset (`makeUiNode` only sets `value` when `text !== label`). In those cases `chosen.value?.length` is 0, so `clearField` sends only the 2-char `CLEAR_BUFFER` no matter how much text the field holds, and `fill` types the new text on top of the leftover — e.g. filling a field containing `old@example.com` with `NEW` yields `old@example.cNEW`. The result still reports `found: true`, `clearedChars: 2`, and no `note` (the only incomplete-clear caveat is the masked-password branch below). Confirmed against the compiled tool with a Chromium-shaped input node (15 chars of text in `label`): 2 backspaces sent, 13 characters left behind, success reported. (Separately, even when `value` is correctly populated, the clear is capped at `MAX_CLEAR_CHARS` = 64 with the same silent-success behavior for longer fields.)

packages/tool-server/src/tools/flows/flow-step-results.ts
```diff
@@ -0,0 +1,21 @@
+const FIND_TOOL_ID = "find";
```
latekvo: `FIND_TOOL_ID` is re-declared here as a local `"find"` literal, while the tool already exports `FIND_TOOL_ID` from `tools/find/index.ts` — the parallel `await-ui-element` guard imports its shared `AWAIT_UI_ELEMENT_TOOL_ID` constant for exactly this purpose. If the tool id ever changes, this private copy silently stops matching and missed-`find` detection in both flow-add-step and flow-run regresses with no compile error.

packages/tool-server/src/tools/find/index.ts
```diff
@@ -0,0 +1,607 @@
...
+            focusSettleMs(device.platform)
+          );
+          if (!focused) return cancelled();
+          // Size the clear. On Chromium the DOM a11y snapshot masks the live
+          // `el.value` behind a static aria-label / placeholder (and never sets
+          // `value` for a form control), so the field's current length is
+          // unknowable — clear up to the cap so a populated field is emptied
+          // regardless, and flag it. Elsewhere max(value,label) is the true
+          // length and never shorter than the real text.
+          const lengthHidden = device.platform === "chromium" && !chosen.value;
```
latekvo: On Chromium, `fill` sizes its backspace-clear from the described `value` (via `editableTextLength`), and the cap-and-warn path is only taken when `value` is empty (`lengthHidden`). But for a `contenteditable` element the DOM accessibility snapshot sets `value` to the element's *own direct text nodes only* — it excludes any text inside nested inline children (a `<b>`, a mention/emoji span, a link). A populated contenteditable therefore reports a non-empty but undercounted `value`, so `lengthHidden` stays false and the safeguard just below is skipped: the clear is sized to the undercount, only the leading characters are deleted, the remainder stays in the field, and the new text is inserted in front of it — reported as a successful fill with no caveat.

Reproduced on Chrome: `find` `fill` on `<div contenteditable>abc<b>defghijklmnopqrst</b></div>` (real content `abcdefghijklmnopqrst`) with text `ZZZ` sent 5 backspaces (`clearedChars: 5`), returned `note` unset, and the element's `textContent` ended up as `abcdefghijklmnoZZZ` rather than `ZZZ`.

packages/tool-server/src/tools/find/index.ts
```diff
@@ -0,0 +1,607 @@
...
+        action,
+        by,
+        query,
+        elapsed: Date.now() - start,
+      });
+      const cancelled = (): FindResult => ({
+        ...baseResult(),
+        found: false,
+        matchCount: 0,
+        note: "find was cancelled before the element was located",
```
latekvo, line 535: This message is also returned from the `type` and `fill` abort paths — when the signal fires during the focus settle (the `focusAndSettle` false-return) or mid-clear just below. On those paths the element has already been located and a device effect has already been dispatched: the focus tap always fires, and for `fill` one or more backspaces may have already deleted field content. So on a mid-action cancel the field can end up focused and partially cleared, yet the result reports `found: false, matchCount: 0` with "cancelled before the element was located" — telling a caller nothing happened when the field was in fact mutated, which is misleading for recovery.

latekvo (approved):
Reviewed the full diff and ran it end-to-end against a live iOS 18.5 simulator and a live Chromium (CDP) target, plus the tool-server suite (1665 tests) and argent-mcp suite — all green — and tsc / eslint / prettier clean.

The tool behaves as intended on device: `exists`, `get-attrs`, the not-found diagnostics, the ambiguity `matchCount`/note, `tap` (verified a real navigation), `type`, `wait` (hit and bounded timeout), and `fill` (clean replacement of a long overflowing single-line field) all worked. The `await-ui-element` refactor is behaviour-preserving, the flow record/replay integration and the run-sequence exclusion are sound, and the skill/docs changes match the implementation. No regressions in the affected packages.

Two things worth a look, both inline below and neither blocking the core behaviour: a Chromium `contenteditable` case where `fill` can silently under-clear, and a cancellation note that misdescribes a field that was already touched.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -73,6 +74,8 @@ Use when you want to replay a recorded flow or run a scripted sequence of device
 Fails if the flow file does not exist or a step tool raises an error (execution stops at that step).
 An \`await-ui-element\` step whose condition is not met before its timeout also stops the flow there
 (its later steps were recorded assuming the condition held), so use one to gate a step on a screen transition.
+A \`find\` step that returns \`{ found: false }\` also stops the flow there, so replay never silently
```
latekvo, line 77: This says a `find` step returning `{ found: false }` stops the flow, but `isMissedFindResult` exempts `action: "exists"` — an `exists` step reporting `found: false` is a successful "not present" answer and neither stops the flow here nor blocks recording in flow-add-step. As written, the description (and the matching one in flow-add-step) reads as if every `found: false` halts replay, which would mislead an agent that uses `find … action:"exists"` as a presence check inside a flow.

latekvo (approved):
Re-reviewed the full diff independently, drove the compiled `find` tool end-to-end against a live iOS 18.5 device, and re-ran the tool-server + argent-mcp suites (all green). The shared-primitive extraction (`describe/match.ts`, `describe/fetch-tree.ts`, `utils/timing.ts`) and the `await-ui-element` refactor are behaviour-preserving; the locator/selection, poll/deadline, and abort logic are sound; the flow record/replay integration and the run-sequence exclusion are correct; and the skill/docs changes match the implementation.

On device I confirmed `exists`, `get-attrs`, the not-found diagnostics, the `matchCount` ambiguity note, `tap` (verified a real navigation), `type`, and `fill` all behave as intended. I specifically stress-tested `fill` on a wide field holding content that extends past the tap centre on native iOS fields (both a search field and a plain text field, each not pre-focused): it cleared the field completely in every case — the focusing tap parks the caret at the end of the existing text, so the backspace clear is not corrupted by tap position.

The two earlier inline notes that were substantive have been resolved (the `fill` clear now sizes from the longer of value/label via `editableTextLength`, and `flow-step-results.ts` now imports `FIND_TOOL_ID` rather than re-declaring it). The Chromium-`fill` under-clear and the cancellation-note wording remain as previously noted and are non-blocking. One small documentation note inline.

Approving.

PR #428, opened 2026-06-30
feat(installer): committable repo-local (`--local`) install mode

filip131311:
## Summary

Adds an opt-in **committable / repo-local** install mode so a team can version Argent _with the repo_. `argent init --local` installs `@swmansion/argent` as a project **devDependency** and writes MCP configs that launch the project-local copy — so after `git clone && npm install`, every teammate has the same Argent version and MCP wiring with **no global install and no per-developer `argent init`**.

**Global mode stays the default** and is unchanged; everything here is additive and backward-compatible.

Motivation: the workflow some users already do by hand (#207). Supersedes the earlier, since-closed #232 — this version is Windows-safe (`node`, not the `.bin` shim), handles Yarn PnP, and closes the cross-install tool-server hazards rather than leaving them.

## What changed

**`init`**
- `--local` / `--global` flags + an interactive "How should argent be installed?" prompt. Mutually exclusive; non-interactive (`--yes`) defaults to **global**.
- Local mode installs the devDependency with the package manager detected from the project's **lockfile** (run with `cwd = projectRoot`).
- MCP entry points at the local copy: `node node_modules/@swmansion/argent/dist/cli.js mcp` (the bin path is derived from the installed `package.json` `bin` and existence-checked — forward slashes, committable, Windows-safe). Yarn PnP → `yarn argent mcp`; an `npx --no-install argent mcp` fallback is used only when the bin can't be resolved.
- Records the mode in a committed `<root>/.argent/install.json`.
- Drops global-only adapters (Windsurf/Hermes) in local mode; prints team-setup / commit guidance.

**`update`** — detects the project's mode; in local mode reads the installed version from `node_modules` (not the global binary / npx cache) and bumps the **devDependency + lockfile**, then re-emits the local MCP command.

**`uninstall`** — scoped to the project's mode: a local-mode uninstall removes the **local devDependency only and never the shared global install**, and prunes `.argent/install.json`. Global-mode behavior is unchanged.

**Cross-install hardening (the gate that makes per-project pinning safe)**
- `launcher`: a tool-server is reused only when its `bundlePath` matches the caller's — a repo-local pin never silently runs against (or kills) a different version's server.
- `postinstall`: only stops a tool-server spawned from *this package's own* bundle, so a devDependency install can't tear down an unrelated editor session.

**Telemetry** — adds an `install_mode` dimension to init/update/uninstall events.

## Testing

- New unit tests: install-mode helpers + record (`utils`), mode-aware `getMcpEntry`/`resolveLocalCommandMode` incl. the opencode array form (`mcp-configs`), local-only uninstall guard (`uninstall`), and bundlePath-gated tool-server reuse (`launcher`).
- All existing suites green (installer 346, tools-client 115, telemetry 191, registry 83, argent 13, tool-server 1621). Build + lint clean.
- **Live end-to-end** in an isolated sandbox: `init --local` → real MCP stdio handshake serves **69 tools from `node_modules`**; `update --local` reads the local version; `uninstall --local` removes the local install and **leaves the global install intact**; `init --global` unchanged; flag conflict rejected.

## Notes / known limitations

- Recommended on mainstream platforms (macOS, Linux x64, Windows x64) with prebuilt native binaries; Linux arm64 / Windows arm need a C/C++ toolchain to build `tree-sitter` on `npm install` (documented in the README).
- Concurrent sessions of two *different* Argent versions on one machine each spawn their own tool-server; an orphaned one idle-times-out (no session is ever broken).


filip131311:
### Follow-up: the feature does not depend on the `postinstall` script

postinstall is frequently disabled (`--ignore-scripts`, pnpm `onlyBuiltDependencies`, Yarn PnP, locked-down CI), so the committable flow is designed to work without it:

- After `npm install`, the committed MCP config + the binary in `node_modules` are all that's needed — the editor spawns `node node_modules/@swmansion/argent/dist/cli.js mcp` directly. The `postinstall` change here is purely a **defensive guard** on the pre-existing tool-server kill (don't tear down a server from a *different* bundle), not new reliance.
- The one thing postinstall used to provide — killing a **stale** tool-server after an in-place version bump so the new binary takes effect — is now handled in the **launcher** instead (commit `e217e6e`): the tool-server state records the installed `version`, and a server tracked under a different version is not reused. So a teammate who pulls a lockfile bump and runs plain `npm install` self-heals on their next MCP call, with `postinstall` disabled or not.

Net: `postinstall` is now a best-effort optimization (faster cleanup when it runs), never a correctness dependency.

filip131311:
### Adopted #242's clean module architecture (commit `c2f83ae`)

Per discussion, rather than carry forward the stale #242 branch (112 commits behind `main`, on the closed #232 base), I brought **#242's focused-module architecture onto this branch** — so #428 now has both the robustness *and* the clean structure, on live `main`.

`init.ts` went **1094 → 316 lines** (a thin orchestrator that reads like a recipe). New single-responsibility modules mirror #242: `package-manager.ts`, `preflight.ts`, `shell.ts`, `topology.ts` (GLOBAL/LOCAL abstraction), `install-record.ts`, `init-args.ts`, `init-telemetry.ts`, `init-mode-prompt.ts`, `install-runner.ts`, `init-adapters.ts`, `init-scope.ts`, `init-mcp-write.ts`, `init-allowlist.ts`, `init-skills.ts`. `utils.ts` re-exports every moved symbol so all import sites/tests resolve unchanged (zero test files edited).

Behaviour-preserving: build + lint clean, all suites green (installer 346 / tools-client 116 / telemetry 191 / registry 83 / argent 13), and live-verified end-to-end (init --local serves 69 tools from `node_modules`; init --global unchanged; uninstall --local leaves the global install intact). Credit to #242 for the structure.

packages/argent-tools-client/src/launcher.ts
```diff
@@ -560,49 +722,94 @@ async function acquireSpawnLock(): Promise<SpawnLock | null> {
...
+  // running the bumped bundle: reuse it instead of killing and downgrading.
+  // Killing on any mismatch is what makes two long-lived sessions of different
+  // versions ping-pong SIGTERMs — each frozen at its own startup version, each
+  // tearing down the other's healthy server on every call. Only compare when
+  // BOTH versions are known; a legacy server with no recorded version is reused.
+  if (
+    wantVersion !== undefined &&
+    state.version !== undefined &&
+    state.version !== wantVersion &&
+    isVersionNewer(wantVersion, state.version)
```
latekvo: `isVersionNewer` compares only the numeric major.minor.patch and ignores any prerelease suffix, so this gate treats two versions that share a base but differ only in a prerelease identifier as equal. The `next` canary channel publishes `<patch>-next.<counter>` on every push to main, so consecutive canaries — and the final `-next.N` build vs the stable release at the same base — all compare as not-newer here. When a project pins such a version through a committed lockfile and it is bumped in place at a stable bundlePath (npm-local, or npm/yarn-classic global; pnpm and Yarn PnP encode the version in the path and are unaffected) while a server from the previous build is still alive and install scripts are disabled (`--ignore-scripts` / locked-down CI — the case this `version` field exists to cover), the launcher keeps reusing the old server and serves stale tool-server code until the idle timeout. The version-aware self-heal silently does not fire for canary-channel installs.

packages/argent-tools-client/src/launcher.ts
```diff
@@ -76,6 +87,12 @@ export interface ToolsServerState {
   pid: number;
   startedAt: string;
   bundlePath: string;
+  /**
+   * Version of the package that spawned this server. Optional for backward-compat
+   * with state files written by older versions (treated as "unknown" — reused
+   * rather than forcing a respawn). See reusableHandle.
+   */
+  version?: string;
```
latekvo, line 110: This field is recorded only by the auto-spawn path; `argent server start` (runDetached and runForeground in `packages/argent-cli/src/server.ts`) writes its state with `bundlePath` and `managed: "cli"` but no `version`, though `paths.version` is available there. Because the reuse gate only compares versions when `state.version !== undefined`, a CLI-started server always falls into the version-unknown branch and is reused regardless of the caller's version. So if a long-lived CLI server is running when the install is bumped in place (without `argent update` or the postinstall kill), the MCP auto-spawn path reuses that stale server instead of spawning a fresh one — and this affects any version bump, not only the prerelease case.

packages/argent-installer/src/update.ts
```diff
@@ -189,10 +227,36 @@ export async function update(args: string[]): Promise<void> {
...
     }
 
+    // Local mode, but nothing is in node_modules — a fresh clone before the
+    // package manager ran, or the marker sits at a root without a manifest.
+    // Running the package-manager add here is wrong either way: it rewrites the
+    // team's committed version pin to @latest, and from a dir with no
+    // package.json the package manager walks up and mutates an unrelated
+    // ancestor project. Never auto-mutate; tell the user to materialize the
+    // install themselves.
+    if (installMode === "local" && localProbe && !localProbe.installed) {
```
latekvo: The local-mode branches added to `update()` — this "declared but not materialized → stop rather than run `<pm> install @latest`" guard, and the PnP `versionUnknown` / `needsInstall` logic below it — are only exercised by calling `update(args)`, but `update.test.ts` covers only the pure helpers (`getUpdateTriggerFromEnv`, `resolveUpdatePackageAction`); nothing drives `update()` itself. The sibling `uninstall()` has a direct test for its analogous local-only guard, so this is the one new high-risk flow (it can rewrite a team's committed version pin) left without behavioral coverage.

packages/argent-installer/src/topology.ts
```diff
@@ -0,0 +1,278 @@
...
+  const stableRel = path.join("node_modules", PACKAGE_NAME, binSub);
+  if (fs.existsSync(path.join(root, stableRel))) {
+    return stableRel.split(path.sep).join("/");
+  }
+  // Fallback for layouts where the package isn't under the project's own
+  // node_modules (hoisted workspace root): use the resolved path. Still
+  // committable, just less resilient to an in-place version bump.
+  const abs = path.join(pkgDir, binSub);
+  if (!fs.existsSync(abs)) return null;
+  return path.relative(root, abs).split(path.sep).join("/");
```
latekvo, line 265: This fallback runs when the package isn't under the project's own `node_modules` (a hoisted workspace). On Windows, `path.relative(root, abs)` returns an absolute path when `root` and the resolved package dir are on different drives, and that absolute path is then committed into the MCP config — machine-specific, so it breaks for any teammate who clones the repo. Narrow (node_modules on a separate volume from the project), but the committed command would be silently non-portable.

latekvo (approved):
Reviewed the committable repo-local install mode closely, including a full end-to-end run on the built binaries (v0.14.0 packed and installed as a local devDependency):

- `init --local` commits `node node_modules/@swmansion/argent/dist/cli.js mcp` plus an `.argent/install.json`, and that command serves the full tool set (71 tools) over a real MCP stdio handshake, spawning a tool-server whose per-bundle state is correctly scoped to the local install (never touching a coexisting global one).
- `update --local` reads the project-local version rather than the global/npx one; `uninstall --local` removes only the local devDependency and its marker and leaves a global install intact.
- Build, all affected suites (installer / tools-client / telemetry / registry / tool-server), and both `typecheck:tests` gates are green; the default global install/update/uninstall flow is preserved.

The cross-install hardening (per-bundle state registry, bundlePath-gated reuse, scoped teardown, token-aware health probe, PnP/workspace-aware probes) is coherent and well reasoned.

The inline notes below are all minor and non-blocking — a couple of edges in the new version-aware server self-heal, one high-risk flow (`update()`) without a driving test, and a rare Windows path case. None of them affect the common paths, which work correctly. Approving; worth a glance at the inline points before merge.

packages/argent-installer/src/install-runner.ts
```diff
@@ -0,0 +1,351 @@
...
+  } catch (err) {
+    installError = err instanceof Error ? err : new Error(String(err));
+  }
+
+  // isLocallyInstalled is PnP-aware (a Yarn PnP project has no node_modules but
+  // declares the dep in-manifest); the extra isYarnPnp keeps that leniency. A
+  // non-PnP layout with no node_modules entry means the add really failed (a bad
+  // spec, ERR_PNPM_ADDING_TO_ROOT, a network error) — don't write a config that
+  // runs a missing binary.
+  if (!isLocallyInstalled(projectRoot) && !isYarnPnp(projectRoot)) {
```
latekvo: This post-install check treats mere resolvability as proof the add succeeded. When the project's own `package.json` does not declare `@swmansion/argent` (so the reuse guard at line 90 is skipped and an add is attempted) but a resolvable copy already exists in `node_modules` — a hoisted sibling/transitive dependency in a monorepo, or a stale copy left from a prior run — and the `npm`/`yarn`/`pnpm add` actually fails (network error, `ERR_PNPM_ADDING_TO_ROOT`, a bad spec), `isLocallyInstalled(projectRoot)` still returns true, so this fatal branch is skipped and the failure is swallowed.

init then prints "Added @swmansion/argent to devDependencies", attributes the non-zero exit to pnpm's blocked build scripts, writes `.argent/install.json` (with `writtenBy` set to the stale version), writes the MCP config, and exits 0 with `is_success: true` telemetry — while `package.json` was never modified. The committed "team-share" setup is therefore silently unbacked: a teammate's or CI's `npm install` won't install argent, and if the stale copy is ever pruned the committed `node node_modules/…/cli.js mcp` command points at a missing file. This is the exact hazard the reuse-guard comment at lines 84-89 was written to avoid, reintroduced through the disk-resolvability check.

Reproduced: a project `package.json` without argent + a stale `node_modules/@swmansion/argent` (v0.9.9-STALE) + a failing add (dead registry) → "Added…" printed, `.argent/install.json` written with `writtenBy: "0.9.9-STALE"`, `package.json` unchanged, exit 0.

packages/argent-tools-client/src/launcher.ts
```diff
@@ -133,6 +161,24 @@ export function findFreePort(): Promise<number> {
   });
 }
 
+/**
+ * True iff semver-ish `a` is strictly newer than `b` (compares the numeric
+ * major.minor.patch; ignores prerelease tags). Anything unparseable compares as
+ * "not newer" so the caller reuses rather than kills — the conservative choice.
+ */
+function isVersionNewer(a: string, b: string): boolean {
```
latekvo: `isVersionNewer` compares only the numeric `major.minor.patch` and ignores the prerelease tag, so the version-based reuse gate never refuses reuse when the numeric triple is unchanged. That gate is the sole in-place-bump self-heal for a tracked server whose `bundlePath` does not change across the bump — i.e. same-path layouts such as the global npm install or a flat (non-pnpm) local `node_modules`, where `tool-server.cjs` is rewritten in place. For the `next` prerelease channel this repo publishes on every push to main, in-place bumps are `…-next.N → …-next.N+1` (same triple), and a `…-next.N → stable` bump lands on the same triple too. In all of those, `reusableHandle` treats the running server as current.

So after such a bump — with the postinstall kill disabled (`--ignore-scripts` / pnpm `onlyBuiltDependencies`) and without an explicit `argent update` — a long-lived autospawn tool-server keeps executing the previous bundle's code until it idle-times-out (and a `managed:"cli"` server, indefinitely), even though the on-disk bundle was replaced. The `ToolsServerPaths.version` / `bundled-paths.ts` comments frame this same-path in-place rewrite as exactly the case the version field is meant to self-heal, which is where the prerelease gap bites. (pnpm is not affected: its versioned store path changes the `bundlePath`, so the `state.bundlePath !== wantBundlePath` branch respawns independently of this gate.)

Verified against the built algorithm: `isVersionNewer("0.14.0-next.6","0.14.0-next.5")` and `isVersionNewer("0.14.0","0.14.0-next.9")` both return false (while `isVersionNewer("0.14.1","0.14.0")` is true), so reuse is not refused across a same-triple prerelease bump.

latekvo (commented):
Two inline notes from going through the local-install flow — both on specific lines below.

PR #431, opened 2026-06-30
test(restart-app): cover dot-prefixing of bare Android activity class names

latekvo:
## What

Adds regression coverage for the Android `restart-app` / `launch-app` activity-component builder.

Both handlers route the `activity` param through the shared `normalizeActivityComponent(bundleId, activity)` helper in `launch-app/platforms/android.ts`, which dot-prefixes a **bare class name** like `MainActivity` into `com.example.app/.MainActivity`. Without the leading dot, `am start -W -n` treats the class as default-package, matches no activity, and `restart-app` throws a spurious `ANDROID_RESTART_FAILED`.

That helper (and `restart-app`'s use of it) shipped in #374; this PR locks the behavior in with a dedicated test so a future refactor can't silently regress it.

## Test

Adds `packages/tool-server/test/restart-app-android-activity.test.ts`, which drives the real `restart-app` and `launch-app` Android handlers (mocking only `adbShell`), captures the emitted `am start -W -n '<component>'`, and asserts the component for bare / relative / FQCN / full-component shapes plus parity between the two tools. Verified it fails when `normalizeActivityComponent` is reverted to the pre-#374 three-case logic (bare name emitted un-dotted) and passes against current source.


hubgan (approved):
Note 1 — overlaps with existing unit coverage

test/android-tv-launch.test.ts:43-68 already unit-tests normalizeActivityComponent directly with the same four input shapes, under a describe block titled "shared by launch-app + restart-app". So the helper's behavior was already locked in; what this PR uniquely adds is the handler-level wiring — proof that both restart-app and launch-app actually route activity through the helper and emit it into am start -W -n. That's real new coverage (a refactor that stopped calling the helper would slip past the unit test), but the four-case table is duplicated. If you want it tighter, one bare-name case per handler plus the parity check would carry the same signal. Not blocking.

One related observation: during the mutation check, the four "restart-app matches launch-app" parity tests still passed — expected, since both tools share the helper and were wrong identically. Those tests guard drift between the tools only; correctness rests entirely on the four direct-component assertions. That's consistent with the file's intent, just worth knowing when reading the test names.

Note 2 — shellCalls should use vi.hoisted

The vi.mock factory closes over the top-level const shellCalls. This works today because vitest executes imports in source order and the const precedes the handler imports — but it's order-fragile: an import-sorting autofix that moves the two handler imports above the const would turn this into a ReferenceError: Cannot access 'shellCalls' before initialization at collection time. The repo's own convention handles this properly (vega-sdk.test.ts, update-checker.test.ts, artifacts.test.ts all use vi.hoisted):

PR #432, opened 2026-06-30
fix(cli): reject empty numeric flags and array/-json interleave crash

latekvo:
Fixes two related defects in `packages/argent-cli/src/flag-parser.ts`, the argv-to-JSON flag parser used by `argent run`. The package previously had no dedicated test file for it; this PR adds one covering both fixes.

## Defect A: empty/whitespace numeric values coerced to 0

`coerceScalar` validates number/integer fields with `Number.isNaN` / `Number.isInteger`, but `Number("")` and `Number("   ")` both evaluate to `0`, so those guards pass. An empty or whitespace-only value was therefore silently accepted as `0` instead of being rejected.

Reachable via `--x=` (empty inline), `--x ""`, and `--x "   "`.

- Observed: `parseFlags(["--x="], { type: "object", properties: { x: { type: "number" } } })` returned `{ x: 0 }`.
- Expected: rejected, the same way a non-numeric string like `"abc"` already is.

Fix: reject an empty or whitespace-only value up front for number/integer types, throwing the same `FlagParseException` the function already throws for `"abc"` (`--<flag> expected a number/integer, got "<raw>"`). Valid numbers are unaffected.

## Defect B: array + `-json` interleave throws a raw TypeError

`seenArrayFields` is only populated by the array branch, but the `--field-json` escape hatch writes `args[fieldName]` directly and can set it to a non-array. If a field was already seen as an array, a later array-form occurrence ran `(args[flag] as unknown[]).push(...)` on a non-array value and threw a raw `TypeError`. `run.ts` only catches `FlagParseException` and re-throws anything else, so the user got an uncaught stack trace instead of a clean error.

Repro:

```
parseFlags(
  ["--tags", "a", "--tags-json", '"b"', "--tags", "c"],
  { type: "object", properties: { tags: { type: "array", items: { type: "string" } } } }
)
```

- Observed: throws `TypeError: args[flag].push is not a function` (uncaught, surfaces as a stack trace through `run.ts`).
- Expected: a clean validation error.

Fix: in the array branch, when `args[flag]` is not actually an array (because `--field-json` replaced it), throw a `FlagParseException` explaining the `--field` and `--field-json` forms cannot be mixed for the same field, never a raw `TypeError`.

## Tests

Adds `packages/argent-cli/test/flag-parser.test.ts`. The new tests fail before the fix (4 failures) and pass after; the full `argent-cli` suite stays green (90 passed). Behavior for valid inputs is unchanged.

### DIFF
--- a/packages/argent-cli/src/flag-parser.ts
+++ b/packages/argent-cli/src/flag-parser.ts
@@ -204,15 +204,20 @@ export function parseFlags(argv: string[], schema: JsonSchema | undefined): Flag
       }
       const { value, nextIndex } =
         inlineValue !== undefined ? { value: inlineValue, nextIndex: i } : takeNext(i, flag);
-      const coerced = coerceScalar(value, itemType, flag);
       if (jsonFields.has(flag)) {
         // --${flag}-json already set this field (in either order relative to
         // this occurrence); mixing the two forms is ambiguous and one would
-        // silently clobber or corrupt the other's value.
+        // silently clobber or corrupt the other's value. Checked BEFORE
+        // coerceScalar so a mixed --field/--field-json with an also-invalid
+        // plain value surfaces this (more actionable) mixing error rather than
+        // a scalar-coercion error — matching the --field-json branch above,
+        // which checks mixing before parsing the JSON.
         throw new FlagParseException(
           `--${flag} and --${flag}-json cannot be mixed for the same field; pass it entirely as --${flag}-json '<json>' or --args '<json>'`
         );
-      } else if (!seenArrayFields.has(flag)) {
+      }
+      const coerced = coerceScalar(value, itemType, flag);
+      if (!seenArrayFields.has(flag)) {
         args[flag] = [coerced];
         seenArrayFields.add(flag);
       } else {
--- a/packages/argent-cli/src/flag-parser.ts
+++ b/packages/argent-cli/src/flag-parser.ts
@@ -94,6 +94,11 @@ export function parseFlags(argv: string[], schema: JsonSchema | undefined): Flag
   // an array field appends; a second value for a scalar field overwrites
   // (with a warning would be nice but we keep it silent to avoid stderr noise).
   const seenArrayFields = new Set<string>();
+  // Track which fields were set via `--field-json`, independent of order: a
+  // field touched by BOTH `--field` (scalar-array form) and `--field-json` is
+  // ambiguous no matter which came first, so both directions must throw rather
+  // than one silently overwriting/discarding the other's value.
+  const jsonFields = new Set<string>();
 
   function takeNext(i: number, flag: string): { value: string; nextIndex: number } {
     if (i + 1 >= argv.length) {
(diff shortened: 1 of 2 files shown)


packages/argent-cli/src/flag-parser.ts
```diff
@@ -186,7 +205,14 @@ export function parseFlags(argv: string[], schema: JsonSchema | undefined): Flag
       const { value, nextIndex } =
         inlineValue !== undefined ? { value: inlineValue, nextIndex: i } : takeNext(i, flag);
       const coerced = coerceScalar(value, itemType, flag);
-      if (!seenArrayFields.has(flag)) {
+      if (jsonFields.has(flag)) {
```
hubgan, line 207: When the two forms are mixed for a numeric/boolean array field and the plain value is itself invalid, the `coerceScalar` call on the line above throws first, so the user sees the scalar-coercion error rather than the mixing error this branch is meant to report. Confirmed with `parseFlags(["--nums-json","[1]","--nums","abc"], …)`, which surfaces `--nums expected a number, got "abc"`. Both are FlagParseExceptions so there is no crash, but for this class of input the message points at the wrong problem.

packages/argent-cli/test/flag-parser.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+describe("flag-parser array + -json interleave never throws a raw error", () => {
+  it("throws FlagParseException (not TypeError) on interleave", () => {
+    let err: unknown;
+    try {
+      parseFlags(["--tags", "a", "--tags-json", '"b"', "--tags", "c"], arrSchema);
+    } catch (e) {
+      err = e;
+    }
+    expect(err).toBeInstanceOf(FlagParseException);
```
hubgan, line 68: These mixing cases assert only the exception type, not its message. The message embeds the field and flag names plus the guidance text that users act on, so a regression in the wording, or an incorrect flag interpolation, would still pass here unnoticed.

packages/argent-cli/test/flag-parser.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    expect(() => parseFlags(["--x="], numSchema)).toThrow(FlagParseException);
+  });
+  it('rejects --x "   " (whitespace)', () => {
+    expect(() => parseFlags(["--x", "   "], numSchema)).toThrow(FlagParseException);
+  });
+  it("rejects --n= (empty integer)", () => {
+    expect(() => parseFlags(["--n="], numSchema)).toThrow(FlagParseException);
+  });
+  it("still accepts a valid number", () => {
+    expect(parseFlags(["--x", "12"], numSchema).args.x).toBe(12);
```
hubgan, line 28: There is no positive case asserting that a numeric value with surrounding whitespace (e.g. " 12 ") is still accepted. The new guard keys off `raw.trim() === ""`, so a future tightening of that check could begin over-rejecting otherwise-valid padded numbers with no test here to catch the regression.

packages/argent-cli/src/flag-parser.ts
latekvo: Fixed in a12a53b3. The mixing check now runs BEFORE coerceScalar, so a mixed --field/--field-json whose plain value is also invalid surfaces the (more actionable) mixing error rather than the scalar-coercion error — matching the --field-json branch, which checks mixing before parsing its JSON. Pinned by a numeric-array test.

packages/argent-cli/test/flag-parser.test.ts
latekvo: Fixed in a12a53b3. The mixing tests now assert the message (`--tags and --tags-json cannot be mixed`), not just the exception type, so a wording or flag-interpolation regression is caught.

packages/argent-cli/test/flag-parser.test.ts
latekvo: Fixed in a12a53b3. Added a positive test that a whitespace-padded number (` 12 `) is still accepted, guarding the raw.trim() guard against a future over-tightening.

hubgan (approved):
Just a note that Defect B is not reachable through any shipped tool

PR #433, opened 2026-06-30
fix(react-profiler): stop ANIMATED_PATTERN matching 'motion' inside ordinary names

latekvo:
## Defect

In `packages/tool-server/src/utils/react-profiler/pipeline/03-tag.ts`, the animation-detection regex was both unanchored and case-insensitive:

```ts
const ANIMATED_PATTERN = /(Animated|Animation|Transition|Motion)/i;
```

Because of the `/i` flag, the token `Motion` matches the lowercase substring "motion" that appears inside perfectly ordinary PascalCase component names: `PromotionCard`, `EmotionThemeCard`, `CommotionList`, `LocomotionView`, and so on. Those components get tagged `isAnimated = true` in stage 3, and stage 4 (`04-rank.ts`) then unconditionally drops every animated component from the findings (`if (comp.isAnimated) continue;`).

The result: a genuinely slow or frequently re-rendering component such as `PromotionCard` silently disappears from the report's Top Components / Suggested Improvements, even though it is exactly the kind of component the profiler exists to surface. `ProductCard` (no "motion" substring) is handled correctly, which is what makes the bug easy to miss.

## Fix

Dropping only the `/i` flag removes the lowercase-embedded false positives, but still mis-tags acronym/number-glued names such as `SVGAnimatedPath`, `RNAnimatedView`, and `MotionlessIndicator`. So the committed pattern anchors the token to a genuine PascalCase segment boundary:

```ts
const ANIMATED_PATTERN = /(?<![A-Z0-9])(Animated|Animation|Transition|Motion)(?=[A-Z0-9_(.]|$)/;
```

- **Case-sensitive**, so a lowercase-embedded "motion" (`PromotionCard`, `Emotion`, `Locomotion`) never contains the capitalized token.
- The leading `(?<![A-Z0-9])` rejects acronym/number-glued prefixes (`SVGAnimatedPath`, `RNAnimatedView`, `CMMotionManager`, `G2MotionSensor`) where the token is not a real PascalCase word-start.
- The trailing `(?=[A-Z0-9_(.]|$)` requires the token to end a segment — another capitalized word, a digit/underscore, `(` or `.` (HOC/dotted display names like `Animated(View)`, `Animated.View`), or end-of-string — so it still matches real names while rejecting tokens that bleed into more lowercase (`MotionlessIndicator`).

**Known limitation:** a bare `Motion` starting a PascalCase word (`MotionSensor`, `MotionManager`) still matches — it is structurally indistinguishable from a real animation name of the same shape (`MotionView`, and `@legendapp/motion`'s `Motion.View`) that we intend to tag. This over-tag is accepted collateral (it only excludes a component from perf findings); a hardcoded device-motion denylist would be more fragile than the rare false positive it removes. The behavior is documented in the code comment and pinned by a test.

## Tests

`packages/tool-server/test/react-profiler-animated-pattern.test.ts` covers:
- lowercase-embedded false positives (`PromotionCard`, `EmotionThemeCard`, `CommotionList`, `ProductCard`) → `isAnimated = false`, and `PromotionCard` reappears in `rank()` findings;
- real animation components (`Animated`, `AnimatedView`, `MotionView`, `FadeTransition`, dotted/HOC forms) → `isAnimated = true`;
- acronym-prefixed and digit-suffixed device-motion names (`CMMotionManager`, `G2MotionSensor`, …) → `isAnimated = false`;
- the accepted bare-`Motion` device-name over-tag (`MotionSensor`, `MotionManager`, `MotionDetector`, `MotionTracker` → `isAnimated = true`), pinned so any future change is deliberate.

The suite fails against the old `/i` regex and passes with the fix. `tsc --build` and ESLint on the changed files are clean.

### DIFF
--- a/packages/tool-server/src/utils/react-profiler/pipeline/03-tag.ts
+++ b/packages/tool-server/src/utils/react-profiler/pipeline/03-tag.ts
@@ -14,18 +14,27 @@ import type { EnrichOutput, TagOutput, TaggedComponent } from "../types/pipeline
 // lowercase substrings. Real RN animation component names are PascalCase and
 // contain a capitalized token (`Animated`, `AnimatedView`, `MotionView`,
 // `FadeTransition`), whereas ordinary names like `PromotionCard`,
-// `EmotionThemeCard`, or `CommotionList` only contain a lowercase "motion".
-// Case-sensitivity alone isn't enough: without a boundary, a capitalized token
-// that merely trails off into lowercase letters (`MotionlessIndicator`,
-// `AnimationsDisabledBanner`) still matches. Anchor the token to a PascalCase
-// segment boundary — start-of-string or preceded by a lowercase/digit (i.e. a
-// new capitalized word starting), and followed by another capitalized word or
-// end-of-string. This still can't disambiguate a token that legitimately
-// starts its own segment in a non-animation compound name (e.g. a sensor
-// component like `DeviceMotionListener`) — that would need a semantic
-// allow/deny list, not a regex — but it closes the trailing-continuation gap
-// the case-only fix left open.
-const ANIMATED_PATTERN = /(?:^|(?<=[a-z0-9]))(Animated|Animation|Transition|Motion)(?=[A-Z]|$)/;
+// `EmotionThemeCard`, or `CommotionList` only contain a lowercase "motion" —
+// case-sensitivity alone (no boundary needed) already rejects those, since the
+// capitalized token literally doesn't appear as a substring.
+// Case-sensitivity alone isn't enough on its own, though: without a trailing
+// boundary, a capitalized token that merely trails off into more lowercase
+// letters of the SAME word (`MotionlessIndicator`, `AnimationsDisabledBanner`)
+// still matches as a bare substring. The lookahead below requires the token be
+// followed by another capitalized word, a digit/underscore (`Animated2`,
+// `Animated_View`), or end-of-string — closing that gap.
+// Deliberately NO leading-boundary requirement: real animation wrapper names
+// are routinely acronym-prefixed (`SVGAnimatedPath`, `HTTPTransitionHandler`,
+// `IOSMotionView`) where the token is preceded by another capital letter, not
+// a lowercase-to-uppercase transition — requiring one caused exactly these to
+// stop matching (a false-negative regression caught in review) with no
+// corresponding false-positive to justify it.
+// This still can't disambiguate a token that legitimately starts its own
+// segment in a non-animation compound name (e.g. a sensor component like
+// `DeviceMotionListener`) — that would need a semantic allow/deny list, not a
+// regex — so that class of false positive remains a known, documented
+// limitation.
+const ANIMATED_PATTERN = /(Animated|Animation|Transition|Motion)(?=[A-Z0-9_]|$)/;
 const RECYCLER_CHILD_PATTERN = /(ListItem|CellItem|Cell|Row|Item)$/i;
 const RECYCLER_PARENT_PATTERN =
   /^(FlatList|SectionList|VirtualizedList|FlashList|RecyclerListView)/i;
--- a/packages/tool-server/src/utils/react-profiler/pipeline/03-tag.ts
+++ b/packages/tool-server/src/utils/react-profiler/pipeline/03-tag.ts
@@ -5,8 +5,8 @@
  *   isAnimated     — component is part of an animation subtree
  *   isRecyclerChild — component is a FlatList/VirtualizedList item
  *
- * These flags drive filtering in Stage 4. Components tagged here are
- * excluded from findings and recorded in IssueReport.excluded.
+ * These flags drive filtering in Stage 4: components tagged here are
+ * excluded from the ranked findings.
  */
 import type { EnrichOutput, TagOutput, TaggedComponent } from "../types/pipeline";
 
@@ -23,14 +23,19 @@ import type { EnrichOutput, TagOutput, TaggedComponent } from "../types/pipeline
 //    `Memo(AnimatedComponent(View))`), or end-of-string — so it still matches
 //    real names while rejecting tokens that bleed into more lowercase
 //    (`MotionlessIndicator`).
-//  - KNOWN LIMITATION: only acronym/digit-GLUED device-motion prefixes
-//    (`CMMotionManager`, `G2MotionSensor`) are rejected. A BARE `Motion` starting
-//    a PascalCase word (`MotionSensor`, `MotionManager`) still matches — it is
-//    structurally indistinguishable from a real animation name of the same shape
-//    (`MotionView`, and `@legendapp/motion`'s `Motion.View`), which we DO want to
-//    tag. This over-tag is accepted collateral: it only excludes a component from
-//    perf findings, and a hardcoded device-motion denylist would be more fragile
-//    than the rare false positive it removes.
+//  - KNOWN LIMITATION: only acronym/digit-GLUED prefixes (`CMMotionManager`,
+//    `G2MotionSensor`) are rejected. A BARE token that starts a PascalCase word
+//    still matches even in a non-animation name: `MotionSensor`/`MotionManager`
+//    (device motion), and — since `Transition` is also an ordinary word (state
+//    machines, phase changes) — `StateTransitionDiagram`, `TransitionMatrix`.
+//    The match is unanchored on purpose so real mid-name animation forms are
+//    caught (`FadeTransition`, `FadeInAnimation`, `SharedElementTransition`),
+//    which makes each over-tag structurally
+//    indistinguishable from a genuine animation name of the same shape
+//    (`MotionView`, `@legendapp/motion`'s `Motion.View`). This over-tag is
+//    accepted collateral: it only excludes a component from perf findings, and a
+//    hardcoded denylist would be more fragile than the rare false positive it
+//    removes.
 const ANIMATED_PATTERN = /(?<![A-Z0-9])(Animated|Animation|Transition|Motion)(?=[A-Z0-9_(.]|$)/;
 const RECYCLER_CHILD_PATTERN = /(ListItem|CellItem|Cell|Row|Item)$/i;
 const RECYCLER_PARENT_PATTERN =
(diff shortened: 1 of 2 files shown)


packages/tool-server/src/utils/react-profiler/pipeline/03-tag.ts
```diff
@@ -10,7 +10,20 @@
...
+//  - Leading `(?<![A-Z0-9])` rejects acronym/number-glued prefixes
+//    (`SVGAnimatedPath`, `RNAnimatedView`, `G2MotionSensor`) where the token
+//    is not a real PascalCase word-start.
+//  - Trailing `(?=[A-Z0-9_(.]|$)` requires the token to end a segment: another
+//    capitalized word, a digit/underscore, `(` or `.` (paren/dot/HOC-wrapped
+//    display names like `Animated(View)`, `Animated.View`,
+//    `Memo(AnimatedComponent(View))`), or end-of-string — so it still matches
+//    real names while rejecting tokens that bleed into more lowercase
+//    (`MotionlessIndicator`).
+const ANIMATED_PATTERN = /(?<![A-Z0-9])(Animated|Animation|Transition|Motion)(?=[A-Z0-9_(.]|$)/;
```
hubgan, line 39: The case-sensitive segment boundary correctly rejects the lowercase-`motion` names this change targets, but it still matches un-prefixed device-motion names — `MotionSensor`, `MotionManager`, `MotionDetector`, `MotionTracker` — as `Motion` followed by a PascalCase boundary. Those are CoreMotion/sensor components rather than animations, so they get tagged `isAnimated` and dropped from `rank()` findings: the same hidden-slow-component outcome this change sets out to prevent. The new tests deliberately protect the acronym-prefixed spellings (`CMMotionManager`, `RNMotionDetector`, `GPSMotionTracker`) as must-not-exclude, yet the more natural un-prefixed `MotionManager`/`MotionSensor` stay excluded. This isn't introduced here — the previous `/i` regex dropped them too — but it's a gap the boundary work leaves open.

hubgan, line 39: The committed pattern here is the boundary-anchored lookbehind/lookahead form, but the PR description presents the fix as simply dropping the `/i` flag (`/(Animated|Animation|Transition|Motion)/`), and its Fix/Tests sections don't mention the leading/trailing boundaries at all. Anyone reviewing from the description alone would be evaluating a simpler regex than what's shipped, and would miss cases the simple form doesn't handle — for example `MotionlessIndicator`, which the plain `/i`-drop still tags as animated.

hubgan (commented):
A few inline notes below on the animation-tagging change. The core fix verifies out — the `motion`-substring names now surface through `rank()` as intended.

packages/tool-server/test/react-profiler-animated-pattern.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    }
+  });
+  it("does NOT tag a real, non-animation acronym-prefixed name", () => {
+    // Without a leading-boundary requirement, an acronym immediately followed
+    // by the token (no lowercase-to-uppercase transition) reads as a match
+    // with nothing to reject it. CMMotionManager/CMMotionActivity are real
+    // Apple CoreMotion SDK class names with nothing to do with animation —
+    // a React Native wrapper/bridge component plausibly named after them must
+    // not be silently excluded from profiler findings.
+    const names = [
```
hubgan, line 101: This case pins the acronym-prefixed CoreMotion spellings as not-animated, but nothing in the suite covers the un-prefixed forms (`MotionManager`, `MotionSensor`, `MotionDetector`). Those currently match the pattern and get excluded from findings, so that behavior sits unpinned and could shift without any test failing.

packages/tool-server/src/utils/react-profiler/pipeline/03-tag.ts
latekvo: Addressed in 421043bd. You're right that bare `Motion`-prefixed device names (MotionSensor, MotionManager, MotionDetector, MotionTracker) still match. They are structurally indistinguishable from a real animation name of the same shape — `MotionView`, and `@legendapp/motion`'s `Motion.View` — which we do want to tag, so there is no regex that separates them without a device-name denylist. I've decided against the denylist (more fragile than the rare false positive, which only excludes a component from perf findings) and instead documented this as accepted collateral in the code comment and pinned it with a test, so the behavior is explicit and any future change is deliberate.

latekvo: Fixed in 421043bd (plus a PR-description edit). The description now shows the actual committed boundary-anchored pattern (not the simplified `/i`-drop), explains the leading/trailing boundaries, and calls out the accepted bare-Motion limitation; the code comment was expanded to match.

packages/tool-server/test/react-profiler-animated-pattern.test.ts
latekvo: Fixed in 421043bd — added a test pinning the un-prefixed forms (MotionSensor, MotionManager, MotionDetector, MotionTracker → isAnimated = true), documented as accepted collateral of matching MotionView, so the behavior is no longer unpinned.

packages/tool-server/test/react-profiler-animated-pattern.test.ts
```diff
@@ -0,0 +1,187 @@
...
+    const ranked = rank(tagged).map((f) => f.component);
+    expect(ranked).toContain("PromotionCard");
+  });
+  it("DOES still tag real animation components", () => {
+    const tagged = tag(enrich(["Animated", "AnimatedView", "MotionView", "FadeTransition"]));
+    for (const n of ["Animated", "AnimatedView", "MotionView", "FadeTransition"]) {
+      expect(tagged.components.get(n)!.isAnimated).toBe(true);
+    }
+  });
+  it("does NOT tag a capitalized token that merely trails into lowercase letters", () => {
```
hubgan, line 65: This pins the trailing-lowercase rejection for the Motion token (MotionlessIndicator) and the Animation token (AnimationsDisabledBanner), but there is no equivalent case for Transition. A name like TransitionalScreen correctly does not match today, yet nothing guards it. If the trailing boundary regressed for that one token, a real component would be silently tagged animated and dropped from findings — the more damaging direction of failure — and the suite would stay green. Transition is the only one of the four tokens without a trailing-continuation case.

packages/tool-server/test/react-profiler-animated-pattern.test.ts
```diff
@@ -0,0 +1,187 @@
...
+      expect(tagged.components.get(n)!.isAnimated).toBe(false);
+    }
+  });
+
+  // Full match/no-match matrix. The paren/dot/HOC-wrapped and dotted forms are
+  // the real regression this fix targets: they are the display names React DevTools
+  // actually produces for animated subtrees (and the exact prefixes the codebase
+  // hard-codes in skip-rules.ts / component-tree.ts), yet the over-tightened
+  // pre-fix regex failed to match them, leaking them back into findings.
+  const MUST_MATCH = [
```
hubgan, line 149: Every paren/dot/HOC-wrapped entry in this matrix is an Animated* form. The Motion.View case explicitly called out as motivating in the 03-tag.ts comment, and the wrapped forms of the other tokens (Transition(View), Animation.View), all match today but are asserted nowhere. A token-specific regression in the boundary handling for Motion, Transition, or Animation would therefore pass unnoticed. The wrapped-name coverage is asymmetric across the four tokens.

packages/tool-server/test/react-profiler-animated-pattern.test.ts
```diff
@@ -0,0 +1,187 @@
...
+    const tagged = tag(enrich(names));
+    for (const n of names) {
+      expect(tagged.components.get(n)!.isAnimated).toBe(false);
+    }
+  });
+
+  // Full match/no-match matrix. The paren/dot/HOC-wrapped and dotted forms are
+  // the real regression this fix targets: they are the display names React DevTools
+  // actually produces for animated subtrees (and the exact prefixes the codebase
+  // hard-codes in skip-rules.ts / component-tree.ts), yet the over-tightened
```
hubgan: skip-rules.ts and component-tree.ts hardcode only the two parenthesized prefixes (Animated( and AnimatedComponent(). They do not cover the dotted (Animated.View) or nested-HOC (Memo(AnimatedComponent(View))) forms this matrix lists, so calling them the exact prefixes the codebase hard-codes slightly overstates the overlap — the pattern here is a superset of what those files match.

hubgan (approved):

packages/tool-server/test/react-profiler-animated-pattern.test.ts
latekvo: Fixed in b95f3b6f. The trailing-lowercase test now pins one case per token — added `TransitionalScreen`/`TransitionerRow` for `Transition` (and `AnimatedlyBanner` for `Animated`) alongside the existing `Motion`/`Animation` cases — so a single-token boundary regression can no longer stay green behind the other tokens that share the regex. Re-ran the exact repro (dropping the trailing lookahead for `Transition` only): the suite was green before and now fails on this test.

packages/tool-server/test/react-profiler-animated-pattern.test.ts
latekvo: Fixed in b95f3b6f. `MUST_MATCH` now covers the wrapped/dotted forms of the other three tokens (`Motion.View`, `Motion(View)`, `Transition(View)`, `Transition.View`, `Animation.View`, `Animation(View)`), and the rank()-exclusion test now also covers `Motion.View` and `Transition(View)`. Verified with a mutation that narrows the trailing class for the non-Animated tokens only: previously green, now red on the matrix and rank tests.

packages/tool-server/test/react-profiler-animated-pattern.test.ts
latekvo: Fixed in b95f3b6f. Reworded the matrix comment so it no longer claims all wrapped forms are what the codebase hard-codes. It now states that only the parenthesized `Animated(...)` / `AnimatedComponent(...)` forms are hard-coded in skip-rules.ts / component-tree.ts, and that the dotted (`Animated.View`), other-token, and nested-HOC (`Memo(AnimatedComponent(View))`) forms go beyond those two prefixes.

PR #434, opened 2026-06-30
fix(tools-client): key downloaded-artifact cache by unique id, not filename

latekvo:
## Defect

In `packages/argent-tools-client/src/artifacts.ts`, the `materializeArtifacts()` download path cached the downloaded bytes at a path keyed by `value.filename`:

```ts
const downloadedPath = join(dir, sanitizeSegment(value.filename));
```

`filename` carries no uniqueness guarantee — server-side it defaults to `basename(hostPath)`, while the wire `id` field is the unique identifier. So two distinct `ArtifactHandle`s in one result that share a `filename` mapped to the **same** local file: the second download overwrote the first, and both handles were rewritten to that one path. The first artifact's path then contained the second's bytes (and for an array of same-named handles, the concurrent `Promise.all` writes could even race).

Only the download branch (remote / `argent link` mode, or when the local integrity check misses) was affected. The co-located `hostPath` fast-path is correct and is left unchanged.

### Repro

`materializeArtifacts` on `{ a: {id:"1", filename:"shot.png", …}, b: {id:"2", filename:"shot.png", …} }` with a fetch returning `"AAA"` for id 1 and `"BBBBBB"` for id 2 → reading the rewritten path for `a` yields `"BBBBBB"` (should be `"AAA"`), because both handles resolve to the same `shot.png` cache path.

## Fix

Key the download cache path by the unique `id` while keeping the human-readable filename:

```ts
const downloadedPath = join(dir, sanitizeSegment(`${value.id}-${value.filename}`));
```

Distinct artifacts can no longer collide, the filename stays legible in the path, and the existing flat cache dir + `sanitizeSegment` usage is preserved (no new directory to create). The local `hostPath` fast-path is untouched.

## Impact

Today's first-party tools happen to emit distinct basenames, so current real-world impact is low. But this is a genuine correctness violation of a public API for inputs the wire format explicitly allows — same-`filename`, distinct-`id` handles — so a remote/linked tool-server (or any future tool emitting same-named artifacts) would silently serve one artifact's bytes under another's path.

## Test

Adds `packages/argent-tools-client/test/artifacts-collision.test.ts`, which materializes two same-`filename`, distinct-`id` image handles through the injected `fetchImpl` download branch and asserts each rewritten path holds its own content. Verified it **fails** before the fix (both paths identical) and **passes** after. Full package suite: 114 tests pass; `tsc --build` and `eslint` clean.

### DIFF
--- a/packages/argent-tools-client/src/artifacts.ts
+++ b/packages/argent-tools-client/src/artifacts.ts
@@ -248,7 +248,14 @@ export async function materializeArtifacts(
         // whole. Mirrors the gate's size check on the local path; skipped when
         // size is unknown (0, e.g. a lazily-registered file).
         if (value.size > 0 && data.length !== value.size) return null;
-        const downloadedPath = join(dir, sanitizeSegment(value.filename));
+        // Key the cache path by the unique `id`, not `filename`: `filename`
+        // defaults to `basename(hostPath)` server-side and carries no uniqueness
+        // guarantee, so two distinct handles in one result can share it. Keying
+        // by filename alone would collide them onto one path — the second
+        // download overwriting the first, and both handles rewritten to those
+        // bytes. The id prefix keeps the human-readable filename while making
+        // the path collision-free.
+        const downloadedPath = join(dir, sanitizeSegment(`${value.id}-${value.filename}`));
         await writeFile(downloadedPath, data);
         if (value.mimeType.startsWith("image/")) {
           images.push({ localPath: downloadedPath, data, mimeType: value.mimeType });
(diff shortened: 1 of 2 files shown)


latekvo:
Closing after a reachability review: the filename-collision cannot occur with any current tool. Every artifact-producing tool returns distinct basenames (screenshot-diff uses *-diff.png / *-context-diff.png; profiler exports are distinct-keyed), and the affected download branch only runs in remote/link mode — local routing resolves via the hostPath fast-path and never hits it. It is a real latent correctness gap in a public API but is not reachable today, so it does not meet the bar for a fix here. Worth revisiting if a tool starts returning same-basename artifacts.

PR #435, opened 2026-06-30
fix(android-profiler): coerce bigint trace timestamps before arithmetic

latekvo:
## Problem

The Perfetto WASM decoder in `packages/native-devtools-android` returns integer columns as `bigint` and only down-converts to `Number` when the value is safe-integer-sized. From `readCell` in `wasm-trace-processor.ts`:

```ts
if (typeof v === "bigint") {
  // ... keep the bigint only for the rare unsafe-range value.
  const n = Number(v);
  return Number.isSafeInteger(n) ? n : v;
}
```

Native `CLOCK_MONOTONIC` timestamps exceed `2^53` ns after roughly 104 days of device/emulator uptime, so the absolute-time columns (`ts_ns`, `first_ts_ns`, `last_ts_ns`) then arrive as `bigint` instead of `number`. Meanwhile `getTraceStartNs()` already coerces `trace_bounds.start_ts` to a plain `Number`.

The Android profiler pipeline then mixes the two — e.g. `row.first_ts_ns - traceStartNs` and `Math.round((row.first_ts_ns - traceStartNs) / 1_000_000)` — which throws:

```
TypeError: Cannot mix BigInt and other types, use explicit conversions
```

The row TypeScript types annotate these columns as `number`, so the compiler can't catch it, and `analyzeNativeProfilerAndroid`'s catch only handles `TraceProcessorUnavailableError`. The result: a freshly captured trace on a long-uptime device fails with a cryptic error instead of producing a report.

## Repro

Feeding `start_ts` (Number-coerced) and `ts_ns`/`first_ts_ns`/`last_ts_ns` as `bigint` greater than `2^53` through `runAndroidProfilerPipeline` throws at `pipeline/index.ts:535` (`cpuRowsToAggregatorRows`) / `:546` (`hangRowsToBottlenecks`). See the new regression test, which fails before this change with the `TypeError` above.

## Fix

Coerce the decoded absolute native-ns row fields with `Number(...)` before any arithmetic that involves the already-Number `traceStartNs` (matching what `getTraceStartNs` does). The downstream `Math.round`/format logic is unchanged, and safe-range traces behave identically. Sites fixed in `packages/tool-server/src/utils/android-profiler/pipeline/index.ts`:

- `cpuRowsToAggregatorRows` — `first_ts_ns` / `last_ts_ns` offset against `traceStartNs`.
- `hangRowsToBottlenecks` — `ts_ns` offset against `traceStartNs`, plus `dur_ns` in `durationMs` and `endNs`.
- `renderHangStacksAndroid` (hang drill-down) — `ts_ns + dur_ns` and the `durationMs` round.

Duration-only columns that never approach `2^53` ns (state/GC durations) are left as-is; the GC fold path already normalises through `toFiniteNumber`.

## Test

New `packages/tool-server/test/android-profiler-bigint.test.ts` drives the real pipeline with `bigint` `ts_ns`/`first_ts_ns`/`last_ts_ns` values greater than `2^53` (mocking only the WASM query layer, same pattern as `pipeline-offset.test.ts`). It fails before the fix (the documented `TypeError`) and passes after. Full tool-server suite: 1622 tests pass; `tsc --build` and `eslint` clean.

### DIFF
--- a/packages/tool-server/src/utils/android-profiler/pipeline/index.ts
+++ b/packages/tool-server/src/utils/android-profiler/pipeline/index.ts
@@ -253,8 +253,11 @@ async function renderHangStacksAndroid(
     return `_Invalid hang_index ${opts.hangIndex}. There are ${hangRows.length} hangs (0-indexed)._`;
   }
   const hang = hangRows[opts.hangIndex]!;
-  const startNs = hang.ts_ns;
-  const endNs = hang.ts_ns + hang.dur_ns;
+  // ts_ns is absolute CLOCK_MONOTONIC ns and can decode as bigint on long-uptime
+  // devices (> 2^53; see readCell). Coerce before arithmetic so adding dur_ns
+  // (a Number) doesn't throw "Cannot mix BigInt and other types".
+  const startNs = Number(hang.ts_ns);
+  const endNs = startNs + Number(hang.dur_ns);
 
   const [stateRows, sampleRows] = await Promise.all([
     runTpQuery<AndroidHangStateRow>({
@@ -277,7 +280,7 @@ async function renderHangStacksAndroid(
     }).catch(() => [] as AndroidHangMainThreadSampleRow[]),
   ]);
 
-  const durationMs = Math.round(hang.dur_ns / 1_000_000);
+  const durationMs = Math.round(Number(hang.dur_ns) / 1_000_000);
   const lines: string[] = [
     `## Hang #${opts.hangIndex} — ${hang.kind} (${durationMs}ms)` +
       (hang.reason ? ` — reason: \`${hang.reason}\`` : ""),
@@ -532,8 +535,12 @@ function cpuRowsToAggregatorRows(
       timestampsNs: [],
       callChains: [{ chain: [dominant], count: row.sample_count }],
       precomputedBursts: parseBurstWindows(row.burst_windows, traceStartMs),
-      firstMs: Math.round((row.first_ts_ns - traceStartNs) / 1_000_000),
-      lastMs: Math.round((row.last_ts_ns - traceStartNs) / 1_000_000),
+      // first/last_ts_ns are absolute CLOCK_MONOTONIC ns: they exceed 2^53 after
+      // ~104 days of device uptime, so the WASM decoder hands them back as bigint
+      // (see readCell). traceStartNs is already a plain Number, so coerce here to
+      // avoid "Cannot mix BigInt and other types" — values stay integral.
+      firstMs: Math.round((Number(row.first_ts_ns) - traceStartNs) / 1_000_000),
+      lastMs: Math.round((Number(row.last_ts_ns) - traceStartNs) / 1_000_000),
       sampleCount: row.sample_count,
     });
   }
@@ -542,8 +549,11 @@ function cpuRowsToAggregatorRows(
 
 function hangRowsToBottlenecks(rows: AndroidJankRow[], traceStartNs: number): UiHang[] {
   return rows.map((row) => {
-    const durationMs = Math.round(row.dur_ns / 1_000_000);
-    const startNs = row.ts_ns - traceStartNs;
+    // ts_ns is an absolute CLOCK_MONOTONIC ns value that can arrive as bigint on a
+    // long-uptime device (> 2^53 ns ≈ 104 days; see readCell). traceStartNs is a
+    // plain Number, so coerce both ts_ns and dur_ns before the arithmetic.
+    const durationMs = Math.round(Number(row.dur_ns) / 1_000_000);
+    const startNs = Number(row.ts_ns) - traceStartNs;
     return {
       type: "ui_hang",
       platform: "android",
@@ -551,7 +561,7 @@ function hangRowsToBottlenecks(rows: AndroidJankRow[], traceStartNs: number): Ui
       durationMs,
       startTimeFormatted: formatTraceTime(startNs),
       startNs,
-      endNs: startNs + row.dur_ns,
+      endNs: startNs + Number(row.dur_ns),
       suspectedFunctions: [],
       appCallChains: [],
       severity: classifyAndroidHangSeverity(row),
--- a/packages/tool-server/src/utils/android-profiler/pipeline/index.ts
+++ b/packages/tool-server/src/utils/android-profiler/pipeline/index.ts
@@ -291,7 +291,12 @@ async function renderHangStacksAndroid(
     lines.push("### Main-thread State Breakdown", "");
     lines.push("| State | Blocked on | Duration |", "|---|---|---|");
     for (const row of stateRows) {
-      const ms = Math.round(row.total_dur_ns / 1_000_000);
+      // Coerce like every other trace-processor duration column: readCell keeps
+      // a value as bigint above 2^53, and mixing bigint with the Number divisor
+      // throws. total_dur_ns is a clipped per-hang SUM so it stays well under
+      // 2^53 in practice, but coerce anyway to keep every duration read here
+      // uniform and robust to a future unclipped/aggregate query.
+      const ms = Math.round(Number(row.total_dur_ns) / 1_000_000);
       lines.push(
         `| ${row.state} | ${row.blocked_function ? `\`${row.blocked_function}\`` : "—"} | ${ms}ms |`
       );
@@ -328,7 +333,7 @@ async function renderHangStacksAndroid(
       stateRows.map((r) => ({
         state: r.state,
         blockedFunction: r.blocked_function,
-        durationMs: Math.round(r.total_dur_ns / 1_000_000),
+        durationMs: Math.round(Number(r.total_dur_ns) / 1_000_000),
       }))
     );
     lines.push("### Main-thread Samples During Hang", "");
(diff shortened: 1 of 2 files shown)


packages/tool-server/src/utils/android-profiler/pipeline/index.ts
```diff
@@ -630,7 +640,9 @@ function classifyAndroidHangSeverity(row: AndroidJankRow): "RED" | "YELLOW" {
   // enough to be user-perceptible (duration check below). Exact === is
   // deliberate — it isolates the standalone case from the combined ones.
   if (row.reason === "App Deadline Missed") return "RED";
-  const durationMs = row.dur_ns / 1_000_000;
+  // Same bigint risk as the other dur_ns/ts_ns usages in this file (see
+  // readCell) — coerce before dividing.
+  const durationMs = Number(row.dur_ns) / 1_000_000;
```
hubgan, line 653: This coercion was absent from the first commit and only added in a follow-up, even though the same `row` reaches this function from `hangRowsToBottlenecks`, where the other `dur_ns` uses were already coerced. Because the row type still declares these columns as `number`, the compiler gives no signal when a `dur_ns`/`ts_ns` arithmetic site is added or left uncoerced. On a long-uptime device the bigint value then reaches the division here (or at any future site added the same way) and throws `Cannot mix BigInt and other types` — the exact failure this PR addresses. The safety of every current and future site rests entirely on a reviewer spotting the missing coercion by hand, which already didn't happen once within this PR.

packages/tool-server/src/utils/android-profiler/pipeline/index.ts
```diff
@@ -277,7 +280,7 @@ async function renderHangStacksAndroid(
     }).catch(() => [] as AndroidHangMainThreadSampleRow[]),
   ]);
 
-  const durationMs = Math.round(hang.dur_ns / 1_000_000);
+  const durationMs = Math.round(Number(hang.dur_ns) / 1_000_000);
```
hubgan: The hang's own `dur_ns` is coerced here, but the state rows' `total_dur_ns` used later in this same function (`Math.round(row.total_dur_ns / 1_000_000)`) is read as a raw field with no coercion. Both are duration columns coming back from the same decoder, so treating one defensively and the other not leaves it ambiguous which column's magnitude is actually being relied on to stay in safe-integer range — a reader can't tell from the code whether the uncoerced site is safe by design or simply overlooked.

packages/tool-server/src/utils/android-profiler/pipeline/index.ts
latekvo: Fixed in e38cf34c. Both `total_dur_ns` reads are now Number()-coerced, so every trace-processor duration read in this file is uniform and the asymmetry you flagged is gone. In practice the value stays a plain Number — it is a hang-window-clipped SUM (hang-state-breakdown.sql) well under 2^53, so readCell never keeps it as bigint — but coercing it anyway is robust to a future unclipped/aggregate query.

packages/tool-server/src/utils/android-profiler/pipeline/index.ts
latekvo: Fixed in e38cf34c. The state rows' `total_dur_ns` is now coerced like the hang's own `dur_ns`, at both use sites, with a defensive test row that forces a bigint through the state-breakdown render to prove it no longer throws.

hubgan (commented):
A few minor, non-blocking notes inline. The fix itself looks correct and complete — verified end-to-end against a real Perfetto trace and covered by the new regression test.

packages/tool-server/src/utils/android-profiler/pipeline/index.ts
```diff
@@ -325,7 +333,7 @@ async function renderHangStacksAndroid(
       stateRows.map((r) => ({
         state: r.state,
         blockedFunction: r.blocked_function,
-        durationMs: Math.round(r.total_dur_ns / 1_000_000),
+        durationMs: Math.round(Number(r.total_dur_ns) / 1_000_000),
```
hubgan: This coercion lives in the `summarizeHangBlocking` fallback that only runs when the hang window has no on-CPU samples (`uniqueStacks.size === 0`). The new drill-down test always supplies a main-thread sample, so this branch never executes under the suite — dropping the `Number(...)` here fails no test. It is the only one of the coerced sites in this change with no regression guard, so a future edit that reintroduced the raw read at this line would go unnoticed by CI.

packages/tool-server/src/utils/android-profiler/pipeline/index.ts
```diff
@@ -253,8 +253,11 @@ async function renderHangStacksAndroid(
     return `_Invalid hang_index ${opts.hangIndex}. There are ${hangRows.length} hangs (0-indexed)._`;
   }
   const hang = hangRows[opts.hangIndex]!;
-  const startNs = hang.ts_ns;
-  const endNs = hang.ts_ns + hang.dur_ns;
+  // ts_ns is absolute CLOCK_MONOTONIC ns and can decode as bigint on long-uptime
+  // devices (> 2^53; see readCell). Coerce before arithmetic so adding dur_ns
+  // (a Number) doesn't throw "Cannot mix BigInt and other types".
+  const startNs = Number(hang.ts_ns);
+  const endNs = startNs + Number(hang.dur_ns);
```
hubgan: `Number(hang.dur_ns)` is evaluated here and again at line 283 in the same function, coercing the same field twice.

packages/tool-server/src/utils/android-profiler/pipeline/index.ts
```diff
@@ -542,16 +554,19 @@ function cpuRowsToAggregatorRows(
...
+    const startNs = Number(row.ts_ns) - traceStartNs;
     return {
       type: "ui_hang",
       platform: "android",
       hangType: row.kind,
       durationMs,
       startTimeFormatted: formatTraceTime(startNs),
       startNs,
-      endNs: startNs + row.dur_ns,
+      endNs: startNs + Number(row.dur_ns),
```
hubgan: `Number(row.dur_ns)` is evaluated here and also at line 560 (`durationMs`) within the same `map` callback — the same field is coerced twice per row.

packages/tool-server/test/android-profiler-bigint.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    queryResponses.push(
+      { name: "trace-bounds.sql", rows: [{ start_ts: TRACE_START_NS }] },
+      { name: "cpu-hotspots.sql", rows: [] },
+      {
+        name: "ui-hangs.sql",
+        rows: [
+          {
+            kind: "jank",
+            ts_ns: TRACE_START_NS + 1_000_000_000n,
+            dur_ns: TRACE_START_NS, // bigint — same magnitude source as ts_ns
```
hubgan: This comment — and the drill-down block at lines 108-117 — describes `dur_ns` as an absolute, uptime-scaled value ("same magnitude source as ts_ns"). Unlike `ts_ns`, `dur_ns` is a bounded duration: it only crosses 2^53 if a single hang lasts ~110 days, not as a function of device uptime. The coercion is a valid mechanical guard regardless, but the rationale here reads differently from the accurate framing used for `total_dur_ns` at lines 148-150.

hubgan (approved):

packages/tool-server/src/utils/android-profiler/pipeline/index.ts
latekvo: Fixed in c25a981b. The state `total_dur_ns` coercion is now computed once into a shared `stateBreakdown` (reused by both the rendered table and the `summarizeHangBlocking` fallback), so there is a single coercion site instead of a second, separately-maintained one. That single site is guarded by the existing drill-down test (which renders the table), and I added a dedicated regression test that drives the no-samples `uniqueStacks.size === 0` fallback with a bigint state `total_dur_ns`. Verified: removing the `Number()` now fails 2 tests, where before it failed none.

packages/tool-server/src/utils/android-profiler/pipeline/index.ts
latekvo: Fixed in c25a981b. `hang.dur_ns` is now coerced once into a local `durNs` and reused for both `endNs` and the `durationMs` header, so the field is coerced a single time in this function.

packages/tool-server/src/utils/android-profiler/pipeline/index.ts
latekvo: Fixed in c25a981b. `row.dur_ns` is now coerced once into a local `durNs` and reused for both `durationMs` and `endNs`, so it is coerced a single time per row in the `map` callback.

packages/tool-server/test/android-profiler-bigint.test.ts
latekvo: Fixed in c25a981b. Reworded the `dur_ns` comments (this line and the drill-down block) to state that `dur_ns` is a bounded duration — the ANR length or longest janky-frame slice — not an uptime-scaled absolute value like `ts_ns`. It would only exceed 2^53 if a single hang lasted ~104 days, so the forced `>2^53` value here is a mechanical guard, not a realistic one. This brings the framing in line with the already-accurate `total_dur_ns` comments.

PR #436, opened 2026-06-30
fix(scripts): stop extract-tools dropping tools with long descriptions

latekvo:
## Defect

`scripts/extract-tools.mjs` extracts MCP tool definitions by regex to feed the downstream `spidershield scan --tools-json` security scan. For each `id:` match it scanned only the next 2000 chars (`const descWindow = afterId.slice(0, 2000)`) for the `description:` value, and the description regexes require the string's closing delimiter to fall inside that window.

Any tool whose description's closing delimiter sits past ~2000 chars is therefore matched as `null` and **silently dropped** - never emitted, never security-scanned. Two shipping tools hit this:

- `run-sequence` (`packages/tool-server/src/tools/run-sequence/index.ts`) - a ~4285-char multi-line template-literal description whose closing backtick sits ~4324 chars past its `id:`.
- `describe` (`packages/tool-server/src/tools/describe/index.ts`) - a ~2195-char description whose closing backtick sits ~2252 chars past its `id:`.

**Security impact:** `describe` is the accessibility / DOM tree tool, exactly the kind of tool a `spidershield` scan should cover, and it was being silently excluded alongside `run-sequence`. A dropped tool is not a scan finding, so the gap was invisible. (Character counts as measured at approval time; both descriptions have since grown on `main` - `run-sequence`'s is ~4478 chars at HEAD.)

### Repro (before)

```
$ node scripts/extract-tools.mjs | grep -c '"run-sequence"'
0
$ node scripts/extract-tools.mjs | grep -c '"describe"'
0
$ node scripts/extract-tools.mjs | grep -c '"gesture-tap"'
1
# total tools emitted: 70  (run-sequence AND describe both missing)
```

### After

```
$ node scripts/extract-tools.mjs | grep -c '"run-sequence"'
1
$ node scripts/extract-tools.mjs | grep -c '"describe"'
1
# total tools emitted: 72  (both captured, run-sequence's full 4285-char description intact)
```

(Counts as of the approval-time tree. At the branch's original base only `run-sequence` overflowed the window - 71 -> 72, with `describe` still under 2000 chars; the branch's mid-review merge of `main` grew `describe`'s description to ~2195 chars, past the window, which is the 70-with-both-missing state shown above. After the final merge of `main` the tree emits 73: the delta is main's new `settings-permissions` tool. `await-screen-idle` also arrived with that merge but registers via a const-ref id, out of extractor scope - see the note below.)

## Fix

- Drop the fixed 2000-char lookahead. Each tool's `description:` is resolved at the **same object-literal brace level** as its `id:` by a forward scan that tracks brace depth over real code only: comments, string literals, template literals (with `${...}` interpolations skipped opaquely, recursively), and regex literals (prev-token heuristic incl. regex-permitting keywords like `return`) are skipped whole, so nothing inside them can open a fake string, shift the depth, or match as a `description:`/`id:` token. The value is then parsed to its actual closing delimiter, so a description of any length is captured in full.
- All three scanners (interpolation contents, the sibling-description scan, and id discovery) route every candidate character through one shared `skipNonCode` dispatcher, so their notion of "code context" cannot drift apart - deleting the dispatcher's regex branch fails 5 fixtures spanning all three scanners.
- Because matching is by brace scope rather than raw position, a stray `id:`/`description:` token can't corrupt extraction: a description-less tool can't borrow the next tool's description, an `id:` appearing inside description text or a comment can't truncate or overwrite anything, and an `id:` key nested inside a value object (e.g. `defaultPayload: { id: "example" }`) is neither emitted as a spurious tool nor allowed to drop the real one.
- Captured text follows full JS cooked-string semantics: named escapes, `\xNN`, `\uNNNN`, `\u{...}`, identity escapes, line continuations, and template `\r\n`/`\r` -> `\n` normalization - the emitted (and security-scanned) text is the string the runtime renders.
- A description value that is not a single plain literal - a concatenation, a `${...}` interpolation, a method/operator suffix (`"...".slice(0, 3)`), `as const` - is never emitted as a truncated/wrong literal: it warn-skips on **stderr** (stdout stays valid JSON for the scanner). Non-interpolated template-literal ids (`` id: `x` ``) are extracted; interpolated ids stay out of scope like const-reference ids.
- Duplicate ids keep the first occurrence and warn (exported as a pure, unit-tested function).
- The extractor stays dependency-free (node: builtins only): the tool-description-quality workflow runs it on a bare checkout without `npm ci`.
- Output JSON shape is unchanged.

## CI guard

Adds `scripts/extract-tools.test.mjs` (node's built-in test runner, wired into `npm run test:scripts`; 46 tests). Ground truth: the real-tree tests parse every tool source file with the actual **TypeScript parser** (a root devDependency, test-only) and require the extractor's output to equal the AST **exactly - names and descriptions byte-for-byte**. Any drop, corruption, or fabrication on a shape that enters the tree fails CI loudly, with no reliance on stderr warning formats and no false red from `id:`-like text in comments or strings.

Two policy gates keep un-extractable shapes out of the tree instead of letting them silently leave the scan: a `description` that cannot be read statically (non-literal value, shorthand/getter/method/computed member, or a spread sibling that may carry/override it) and a plain-literal description written **before** its `id:` (the scan is forward-only) each fail with an actionable, file-naming message.

Fixture tests pin the parsing rules independent of any real tool's wording - long descriptions (>4000-char value, >2000 chars of siblings between `id:` and `description:`), embedded `id:`/`description:` tokens, comments, keyword-position and sibling regex literals, a quote-bearing regex inside a template interpolation, division, nested templates and interpolation content, template/`as const` ids, the full escape suite, CRLF cooking, single-quoted forms, trim, warn-skip classes, and the duplicate-id warning. Each new fixture was verified to fail under a targeted mutation of the extractor.

Note: tools registered with a const-reference / dynamic id (e.g. `id: AWAIT_UI_ELEMENT_TOOL_ID`) rather than a static literal stay out of scope for this extractor - a separate, pre-existing limitation unrelated to the description window, and one this static/lexical extractor cannot close soundly (it has no tool-scope knowledge, so a blanket warning on every `id: <identifier>` would also fire on legitimate nested non-tool keys like `{ id: someVar }` in a payload/example object). The full set currently taken out of the scan this way is `await-screen-idle` and `await-ui-element`. (Earlier revisions of this note and one review thread also listed four `app-map-*` tools - those names match no tool anywhere in the repo's history; the list was wrong, not the tree.) A proper fix is a follow-up covering const-ref id resolution plus a real-tree emit-count and empty-stderr assertion so this silent-omission class trips CI if it ever regresses.

## Post-approval updates

- **Merged `main`.** The merge tripped this PR's own non-static-description policy gate on CI: #459 had interpolated `${NON_INJECTABLE_RECOVERY}` into the `native-devtools-status` description, which a static extractor cannot read, so that tool would have silently left the spidershield scan. Fixed by inlining the recovery text verbatim as a plain template literal (escaped backticks cook to the identical runtime string) - the one product-code change in this PR. The existing `native-devtools-status.test.ts` assertion (description contains `NON_INJECTABLE_RECOVERY`) pins the inlined text byte-for-byte against the runtime constant, so the shared-verbatim guarantee #459 built survives as a test-time invariant, and a doc comment on the constant points editors at both sites.
- **Scanner-dispatch unification** (see Fix above): the three previously verbatim-duplicated non-code skip prologues became the single `skipNonCode` dispatcher; real-tree output verified byte-identical before/after. The previously-unpinned interpolation-regex path gained its fixture, and dead surface was removed (shadowed `REGEX_PREV_CHARS` entries, the importerless `extractAllTools` export, an unread `stderr` field in the test helper).

## Verification

- `node scripts/extract-tools.mjs` -> 73 tools at HEAD (71 -> 72 at the original base, 70 -> 72 as of the approval-time tree, +`settings-permissions` from the final `main` merge = 73), `run-sequence` and `describe` both present, byte-identical to the TypeScript-AST parse of the tree, no stderr warnings
- `npm run test:scripts` -> all pass (`extract-tools.test.mjs`: 46 tests; 59 total across the three script suites)
- mutation checks: window-bounding, depth-anchor removal, trim removal, dedup neutralization, single-quote-branch neutralization, and shared-dispatcher regex-branch deletion (5 fixtures across all three scanners) each fail the suite
- `npm run typecheck:scripts` -> clean
- tool-server after the merge + description inline: full vitest suite green, `tsc --noEmit` (src and `tsconfig.test.json`) clean
- `prettier` + `eslint` on changed files -> 0 problems
- lockfile: `npm install --package-lock-only --ignore-scripts` -> no diff (idempotent)


hubgan (commented):
A few inline notes below.

scripts/extract-tools.test.mjs
```diff
@@ -0,0 +1,297 @@
...
+function discoverIdLiterals() {
+  const ids = new Set();
+  const walk = (dir) => {
+    for (const name of readdirSync(dir)) {
+      const full = join(dir, name);
+      if (statSync(full).isDirectory()) {
+        walk(full);
+      } else if (extname(name) === ".ts" && !name.endsWith(".d.ts")) {
+        const src = readFileSync(full, "utf8");
+        const re = /\bid:\s*["']([^"']+)["']/g;
```
hubgan: This oracle counts every `id: "..."` string literal in the source as a tool definition, but the extractor deliberately does not: a nested `id:` key (for example `defaultPayload: { id: "..." }`) or an `id: "..."` token appearing inside a description is intentionally skipped, and this PR adds dedicated handling for exactly those shapes. On the current tree the two views happen to coincide only because no such nested or embedded `id:` exists.

I confirmed that adding `defaultPayload: { id: "example" }` to a real tool file makes this test fail with `tool definitions dropped by the extractor: example` and also trips the no-warnings assertion below — even though the extractor handled that input correctly and emitted the expected warning. So the moment a tool file uses one of the nested-id shapes the extractor is built to tolerate, this guard turns CI red with a misleading "dropped" message.

scripts/extract-tools.mjs
```diff
@@ -31,62 +37,186 @@ function walk(dir) {
...
+    }
+    if (ch === "/" && afterId[i + 1] === "*") {
+      const end = afterId.indexOf("*/", i + 2);
+      if (end === -1) return null; // unterminated block comment; nothing parseable after
+      i = end + 1; // land on the '/'; loop's i++ steps past it
+      continue;
+    }
+    if (ch === '"' || ch === "'" || ch === "`") {
+      quote = ch;
+    } else if (ch === "{") {
```
hubgan: The forward scan tracks `{`/`}` but doesn't recognize regex literals. A `{` or `}` inside a regex literal that appears between a tool's `id:` and its `description:` (for example a field like `match: /[{]/`) shifts the tracked depth, so the real `description:` is never seen at depth 0 and the tool is skipped (with a warning). No shipping tool has this shape today, but the previous fixed-window matcher tolerated it, so this is a narrowing of what the extractor accepts.

scripts/extract-tools.mjs
```diff
@@ -31,62 +37,186 @@ function walk(dir) {
...
+    //     template literals are captured in full.
+    // A null result means this `id:` is a nested object key, not a tool.
+    const value = findOwnDescriptionValue(src.slice(idMatch.index + idMatch[0].length));
+
+    let description = null;
+    if (value !== null) {
+      const delim = value[0];
+      // Template literal allows escaped backticks (\`) so inline code like
+      // `adb pull` doesn't truncate; the quoted forms allow the standard escapes.
+      const valueMatch =
```
hubgan, line 484: When the `description:` value is not a single string or template literal, capture is lossy and silent. For a concatenated value such as `description: "a" + "b"`, this delimiter parse matches only the leading `"a"`, so the tool is emitted with a truncated description and no warning that the remainder was dropped. A template literal containing `${...}` interpolation is likewise captured as its literal source text. By contrast, a bare const reference like `description: SOME_CONST` is handled loudly — it warns and skips. No shipping tool currently uses these forms.

scripts/extract-tools.mjs
```diff
@@ -31,62 +37,186 @@ function walk(dir) {
...
+      depth++;
+    } else if (ch === "}") {
+      if (--depth < 0) return null; // id's own object closed - not a tool
+    } else if (
+      depth === 0 &&
+      ch === "d" &&
+      // word boundary before the token so `...Xdescription:` doesn't match
+      !/[A-Za-z0-9_$]/.test(afterId[i - 1] ?? "")
+    ) {
+      const kw = afterId.slice(i).match(/^description:\s*/);
```
hubgan: This allocates `afterId.slice(i)` and runs a fresh regex match for every `d` character reached at depth 0 — i.e. a substring of the remaining file per candidate. Harmless at the current file sizes; noting the per-candidate allocation.

scripts/extract-tools.test.mjs
```diff
@@ -0,0 +1,297 @@
...
+  `;
+  const tools = extractToolsFromSource(src, "fixture.ts");
+  const byName = Object.fromEntries(tools.map((t) => [t.name, t.description]));
+  assert.equal(byName["create-thing"], "Creates a thing.", "real tool dropped by a nested id: key");
+  assert.ok(!("example-id" in byName), "a nested object's id: key was emitted as a spurious tool");
+  assert.equal(tools.length, 1, "exactly the real tool should be emitted");
+});
+
+test("a balanced nested object (with a deep inner id:) between id: and description: keeps the tool", () => {
+  // Mirrors the real `capability: { apple: { ... } }` shape 12 tools already
```
hubgan: This comment says the `capability: { apple: { ... } }` shape is used by 12 tools, but the corresponding comment in extract-tools.mjs says 7, and counting tools that actually have a `capability: {` between `id:` and `description:` on the current tree gives 7. The two comments disagree on the count.

scripts/extract-tools.test.mjs
latekvo: Fixed in 972dff08. You're right — the whole-tree oracle counted every `id:` literal as a tool, so any legitimately-skipped nested id turned CI red with a misleading "dropped" message. I reproduced exactly that by adding `defaultPayload: { id: "example" }` to a real tool file: the extractor handled it correctly (kept the real tool, skipped `example` with a warning) but the test failed with "tool definitions dropped by the extractor: example".

The guard now asserts the real invariant instead: every discovered id must be either emitted or named in a stderr WARNING (skipped loudly), and no emitted name may be fabricated. It stays green when a tool uses a nested-id / non-static-description shape, and fails only on a genuinely silent drop (verified both directions — including by forcing a silent drop of run-sequence, which the guard now catches).

scripts/extract-tools.mjs
latekvo: Kept as-is. Neither scanner lexes regex literals — doing that correctly needs the full regex-vs-division lexical disambiguation, which is disproportionate for a build script and a shape no tool uses (no tool definition has a regex-literal field between its `id:` and `description:`). The specific `{`/`}` case you show is dropped with a stderr WARNING (loud, not silent), and the rewritten "no silent drops" test turns red if such a shape ever reached the tree. Leaving it as a documented limitation; happy to revisit if a real tool ever needs a regex-literal sibling field.

scripts/extract-tools.mjs
latekvo: Fixed in efdc5fb9. A description value that isn't a single string/template literal — a `"a" + "b"` concatenation, or a template with `${...}` interpolation — is no longer emitted as its truncated leading literal. It now warns on stderr and is skipped, so a partial/placeholder description can't reach the security scan unnoticed (consistent with how a bare const reference was already handled). Added fixtures for the concatenation and interpolation cases, plus one confirming an escaped `\${` stays a literal and is still captured in full.

scripts/extract-tools.mjs
latekvo: Fixed in efdc5fb9. The `description:` lookup now uses a module-level sticky regex anchored at the candidate index (`DESCRIPTION_KEY.lastIndex = i`), so there's no per-candidate `slice`/substring allocation while scanning.

scripts/extract-tools.test.mjs
latekvo: Fixed in 972dff08. The count was wrong — I measured it: 7 tools have a `capability: { ... }` between their `id:` and `description:` (the `native-*` ones). Both this comment and the corresponding one in extract-tools.mjs now say 7.

scripts/extract-tools.mjs
```diff
@@ -31,62 +45,257 @@ function walk(dir) {
...
+      if (end === -1) return null; // unterminated block comment; nothing parseable after
+      i = end + 1; // land on the '/'; loop's i++ steps past it
+      continue;
+    }
+    if (ch === '"' || ch === "'" || ch === "`") {
+      quote = ch;
+    } else if (ch === "{") {
+      depth++;
+    } else if (ch === "}") {
+      if (--depth < 0) return null; // id's own object closed - not a tool
```
hubgan: The forward scan here tracks string, template, and comment state but has no state for regex literals. A regex whose body contains a quote or an unbalanced brace — e.g. `z.regex(/["']/)` or `s.replace(/"/g, ...)` — sitting between an `id:` and its `description:` opens a fake string or shifts brace depth, so the sibling `description:` is never matched at depth 0 and the tool is skipped with a "no sibling description" warning. Because the no-silent-drop test accepts a warned drop, such a tool would silently leave the spidershield scan while CI stays green. (A quote-regex placed before the `id:` instead hides the id itself; that variant is caught by the test's raw-regex discovery and turns CI red.) Nothing in the current tool tree triggers this — verified 72 tools, zero warnings, all descriptions byte-exact — so it is latent, not an active break. It is a narrow behavioral change from the old windowed matcher, which tolerated these regexes; a separate hardening PR is the right place for it rather than blocking this one.

scripts/extract-tools.mjs
```diff
@@ -31,62 +45,257 @@ function walk(dir) {
...
+        // The literal is only the whole description if nothing extends it. A
+        // trailing `+` (string concatenation) or an unescaped `${...}` in a
+        // template (runtime interpolation) means the rendered text isn't the
+        // captured literal alone - leave description null so it falls through to
+        // the loud warn-and-skip below rather than emitting a truncated string
+        // silently into the security scan.
+        const rest = value.slice(valueMatch[0].length).replace(/^\s+/, "");
+        const isConcatenation = rest.startsWith("+");
+        const hasInterpolation = delim === "`" && /\$\{/.test(valueMatch[1].replace(/\\./g, ""));
+        if (!isConcatenation && !hasInterpolation) {
```
hubgan: This guard treats the captured literal as the whole description unless it is followed by `+` (concatenation) or a template `${...}` (interpolation). A method or operator suffix on the literal is not covered: `"abcdefghij".slice(0, 3)` and `"hello WORLD".replace("WORLD", "there")` both start with a literal, so the raw literal is emitted (`"abcdefghij"` / `"hello WORLD"`) — text that differs from the runtime string — with no warning on either invariant. That is the one "emitted but wrong" class that is fully silent to both the extractor and the tests. It is pre-existing (the old script did the same, verified) and low-likelihood, since no current tool writes a method-suffixed description; flagging it only because it sits right next to the silent-partial class this script exists to catch.

scripts/extract-tools.mjs
```diff
@@ -31,62 +45,257 @@ function walk(dir) {
...
+  return tools.filter((t) => {
+    if (seen.has(t.name)) {
+      console.error(
+        `extract-tools: WARNING: duplicate tool id "${t.name}"; keeping the first occurrence and skipping the rest.`
+      );
+      return false;
+    }
+    seen.add(t.name);
+    return true;
+  });
```
hubgan, line 577: The cross-file dedup and its duplicate-id warning here, and the single-quoted `id:` / `description:` branches elsewhere, are not exercised by any test: the real tool tree has no duplicate ids and uses only double quotes or backticks, so these paths never run under the suite. Noting it as a coverage gap, not a defect.

scripts/extract-tools.test.mjs
```diff
@@ -0,0 +1,406 @@
...
+function discoverIdLiterals() {
+  const ids = new Set();
+  const walk = (dir) => {
+    for (const name of readdirSync(dir)) {
+      const full = join(dir, name);
+      if (statSync(full).isDirectory()) {
+        walk(full);
+      } else if (extname(name) === ".ts" && !name.endsWith(".d.ts")) {
+        const src = readFileSync(full, "utf8");
+        const re = /\bid:\s*["']([^"']+)["']/g;
```
hubgan: This discovery regex matches `id: "..."` anywhere in the source, including inside comments and description strings, whereas the extractor's lexer deliberately skips those without warning. So the "no silent drop" assertion holds only while every commented or embedded `id:` also happens to be a real emitted tool — true today (72 discovered equals 72 emitted). The first time a tool file adds a cross-reference comment like `// see id: "other-tool"`, or an `{ id: "example" }` inside a description's structured example, `discovered` gains a name the extractor correctly omits and does not warn about, and this test fails against a correct extractor. That also contradicts the fixture at line 125, which endorses an `id:` inside description text as a supported shape.

scripts/extract-tools.test.mjs
```diff
@@ -0,0 +1,406 @@
...
+  // chars; assert it runs well past the old window AND reaches its closing
+  // backtick (the final sentence), so the capture is provably complete.
+  assert.ok(
+    runSequence.description.length > 3000,
+    `run-sequence description truncated to ${runSequence.description.length} chars (past the old 2000-char window expected)`
+  );
+  assert.ok(
+    runSequence.description.includes("returns partial results"),
+    "run-sequence description does not reach its final sentence — captured value is truncated"
+  );
```
hubgan: These assertions are coupled to the current wording of run-sequence's description: a length over 3000 and the literal final phrase "returns partial results". A legitimate rewrite that shortens the description under the threshold, or rewords its final sentence, would fail here even though capture is complete — which reads as a capture regression when nothing regressed.

scripts/extract-tools.test.mjs
```diff
@@ -0,0 +1,406 @@
...
+  // emit — a nested object key (e.g. `defaultPayload: { id: "x" }`), a tool
+  // whose description isn't a static string/template literal, or a
+  // description-less tool — must be named in a stderr WARNING, i.e. skipped
+  // LOUDLY. This is the exact invariant the PR restores (the old fixed-window
+  // matcher dropped long-description tools SILENTLY): the check stays green when
+  // a future tool uses one of those shapes (the extractor warns and the count
+  // shifts) and fails only on a genuinely SILENT drop.
+  const silentlyDropped = [...discovered].filter(
+    (id) => !extracted.has(id) && !stderr.includes(`"${id}"`)
+  );
```
hubgan: Two properties of this predicate limit the safety net. First, it accepts a warned drop as acceptable, and nothing pins the emitted tool count or asserts empty stderr against the real tree — so if any tool other than run-sequence or gesture-tap regresses to a warn-skip (description becomes a concatenation, interpolation, method-suffixed literal, or is removed), it drops out of the spidershield scan while this suite stays green. Second, the check relies on every warning wrapping the id in double quotes; the fixture warning tests assert only bare-substring presence, so rewording a warning to change that quoting would break this invariant while the fixtures stay green.

hubgan (commented):
Reviewed the extractor rewrite end-to-end against the real tool tree, not only the suite: it does what it claims — recovers run-sequence and describe (70 to 72) with byte-exact descriptions, drops nothing, and correctly renders flow-add-step's embedded newline. Tests, typecheck, and lint all pass locally. None of the inline comments are blocking; most are latent or pre-existing.

Two items are out of scope for this PR and better handled as follow-ups: (1) tool ids defined via a const reference — e.g. await-ui-element's `id: AWAIT_UI_ELEMENT_TOOL_ID` — stay invisible to the scan (already acknowledged in the description); a follow-up could cover const-ref ids and add a real-tree emit-count and empty-stderr check so a future silent regression trips CI. (2) The extractor also emits two defs that are not registered as live MCP tools, `paste` and `simulator-server` — pre-existing and harmless to the scan, not introduced here.

scripts/extract-tools.mjs
latekvo: Reconsidered after your follow-up below — fixed in a63e99c4 rather than kept as a documented limitation. Both scanners now lex regex literals whole, using the standard prev-token heuristic to keep division as division. Fixtures cover `match: /[{]/`, `z.string().regex(/["']/)` and `s.replace(/"/g, ...)` between `id:` and `description:`, plus a division control case so the heuristic's other side is pinned too.

scripts/extract-tools.mjs
latekvo: Fixed in a63e99c4, and the blind spot it exploited is closed in 105908fa. The lexer now skips regex literals whole — including the fully-silent variant where a quote-bearing regex before the `id:` swallowed the tool with no warning at all. Separately, the suite no longer treats a warned drop as green: the extractor's output must equal a TypeScript-AST parse of the tool tree exactly, so any residual lexer gap that loses a real tool now fails CI instead of silently leaving the spidershield scan. Reproduced your exact scenario before the fix (a real tool file with `z.string().regex(/["']/)` between id and description: extractor emitted 0 occurrences of it while the whole suite passed) and re-ran it after: the tool is captured byte-identical to the AST and the suite still passes.

scripts/extract-tools.mjs
latekvo: Fixed in a63e99c4. The whole-value check no longer special-cases `+`: after the captured literal the property must simply end (`,` or `}`, allowing trailing comments), so `.slice(0, 3)`, `.replace(...)`, `as const`, or any other suffix now warns and skips instead of emitting the raw literal. Reproduced both of your examples emitting wrong text with zero warnings before the fix; both now warn-skip, with fixtures added. On top of that, 105908fa makes this class un-shippable: the AST oracle classifies a non-literal description value as non-static and the suite fails if the real tree ever contains one.

scripts/extract-tools.mjs
latekvo: Fixed in 105908fa (with the dedup extracted to an exported pure function in a63e99c4). The duplicate-id path now has a fixture (first occurrence kept, warning names the duplicate) and single-quoted `id:`/`description:` forms have a dedicated fixture including an escaped quote. Verified by mutation: neutralizing the dedup body or the single-quote branch previously left the suite fully green; each now fails its test.

scripts/extract-tools.test.mjs
latekvo: Fixed in 105908fa. The discovery regex is gone entirely — the oracle is now a real TypeScript-AST parse of each tool file, which by construction ignores `id:` tokens inside comments and description strings, so it can never over-discover. Reproduced the false red before the fix: adding a benign `// see id: "phantom-tool"` cross-reference to a tool file failed the suite with a misleading 'dropped with no stderr warning' message even though the extractor handled it correctly. The same file passes under the AST oracle, while genuinely dropping a real tool still fails.

scripts/extract-tools.test.mjs
latekvo: Fixed in 105908fa. The length threshold and the final-phrase literal are gone: run-sequence (plus describe and gesture-tap as control) is asserted present-only, fullness is covered byte-for-byte by the AST-equality test — which follows any legitimate rewrite automatically — and the fixed-window regression class is pinned by a synthetic >4000-char fixture that is independent of any real tool's wording. Reproduced first: rewording run-sequence's final sentence failed the old test with capture complete; after the change the same rewording passes.

scripts/extract-tools.test.mjs
latekvo: Fixed in 105908fa. Both properties are gone. (1) A warned drop is no longer acceptable: the extractor's output must exactly equal the TypeScript-AST ground truth — names and descriptions — and a separate test requires every real tool description to remain a plain literal, so a regression to a concatenation, interpolation, or method-suffixed value turns CI red with an actionable message. Reproduced your scenario: a real tool regressed to `"part A " + "part B"` kept the old suite green while the tool left the scan; the new suite fails on exactly that tree. (2) The real-tree tests no longer parse stderr at all — the AST equality is the drop detector — so warning wording is free to change without weakening the net.

scripts/extract-tools.mjs
```diff
@@ -4,19 +4,67 @@
...
+// Sticky matcher for an `id: "..."` string/template literal with a matching
+// closing delimiter, anchored at a known code-context position (see
+// findIdLiteralsInCode). A template id is only static without interpolation;
+// the caller rejects a captured `${`.
+const ID_LITERAL = /id:\s*(["'`])([^"'`]+)\1/y;
+
+// Characters after which a `/` starts a regex literal rather than division:
+// an operator, an opening bracket, or a separator — never an identifier, a
+// number, a closing paren/bracket, or a string (where `/` means division).
+const REGEX_PREV_CHARS = new Set("(,=:[!&|?{};+-*%<>~^");
```
hubgan: `!` is (correctly) in this set so a prefix logical-NOT can precede a regex, e.g. `if (!/Success/i.test(x))`. The heuristic can’t distinguish that prefix `!` from a TS postfix non-null assertion, though, so `x! / <divisor>` reads the `/` as a regex opener. If a `/` or a quote then appears in the divisor on the same line, string/regex lexing desyncs, the enclosing object’s `description:` is hidden, and the tool - plus every tool after it in the file - drops out of the scan (e.g. `w: a! / "x/y".length` sitting between an `id:` and its `description:`). Not triggerable by any current tool object, and the AST-equality test would catch it if it ever appeared; flagging it only as the one value-position where the regex-vs-division heuristic is unsound. The common `count! / total` (no same-line slash) self-corrects and is fine. Non-blocking.

scripts/extract-tools.mjs
```diff
@@ -31,62 +79,534 @@ function walk(dir) {
...
+    }
+    if (ch === "{") {
+      depth++;
+    } else if (ch === "}") {
+      if (--depth < 0) return null; // id's own object closed - not a tool
+    } else if (
+      depth === 0 &&
+      ch === "d" &&
+      // word boundary before the token so `...Xdescription:` doesn't match
+      !/[A-Za-z0-9_$]/.test(afterId[i - 1] ?? "")
```
hubgan, line 351: This word-boundary check is load-bearing - it stops a lowercase `*description:` sibling (e.g. `subdescription:`) from being captured as the tool’s own description. If it regressed, `{ id, subdescription: "X", description: "Y" }` would emit `X` (wrong text into the security scan) and the suite would stay green: no fixture exercises a `*description:` sibling appearing before the real one, so only the real-tree AST-equality test would catch it, and only once such a sibling actually lands in the tree. The guard is correct as written; noting that this particular path isn’t pinned by a test. Non-blocking.

scripts/extract-tools.mjs
```diff
@@ -31,62 +79,534 @@ function walk(dir) {
...
+    }
+    if (ch === "`") {
+      i = skipTemplateLiteral(src, i) - 1;
+      prevSig = "`";
+      prevSigIdx = i;
+      continue;
+    }
+    // A code-context `id:` token. The word-boundary check before it means
+    // `grid:`, `$id:`, `androidId:`, etc. are not mistaken for a tool id.
+    if (ch === "i" && !/[A-Za-z0-9_$]/.test(src[i - 1] ?? "")) {
```
hubgan, line 397: Same situation as the `description:` word-boundary check above: this guard keeps a lowercase `*id:` sibling (`uuid:`, `grid:`) from being read as a tool id, and it isn’t pinned by any fixture either - removing it fabricates spurious tools from those keys while the suite stays green. Lower impact than the description case (fabrication rather than corruption); just flagging the unpinned guard. Non-blocking.

scripts/extract-tools.mjs
```diff
@@ -31,62 +79,534 @@ function walk(dir) {
...
+ * appearing in a comment (e.g. a `// see id: "screenshot"` cross-reference) or in
+ * another tool's description text became a tool candidate. `findOwnDescriptionValue`
+ * then read the enclosing object's real `description:` and, via first-wins dedup,
+ * that spurious entry could silently overwrite a real tool's description in the
+ * downstream security scan. Lexing to code context first removes that whole class,
+ * and also skips non-id keys like `$id:` (the `$` fails the word-boundary check).
+ *
+ * A template-literal id (`` id: `x` ``) counts only without `${...}`
+ * interpolation; an interpolated id is dynamic, like a const-reference id, and
+ * is out of scope for this static extractor.
```
hubgan, line 379: Beyond `await-ui-element`, const-reference / dynamic ids currently take six shipping tools out of the scan - `app-map-crawl/export/navigate/query`, `await-screen-idle`, `await-ui-element` - and unlike the missing-description and duplicate-id cases, this one is fully silent (no stderr warning). It’s the same invisible scan-bypass class this PR closes, reached by a different cause (id resolution rather than the description window). Pre-existing and reasonable to leave out of this fix; noting it so the silent omission is on record. Non-blocking.

hubgan (commented):
Reviewed the extractor and its test suite - these comments are all non-blocking.

The change does what it says: I reproduced the before/after (70 -> 72 extracted tools) and ran the real `spidershield scan --tools-json` end to end, confirming `describe` and `run-sequence` are now scored by the scan (both also verified working on-device). Tests, typecheck, prettier, eslint and lockfile idempotency all pass. The notes below are minor hardening and test-coverage observations.

hubgan (approved):

scripts/extract-tools.mjs
latekvo: Fixed in 428e4304. You're right — `!` sat in REGEX_PREV_CHARS, so a postfix non-null assertion `x!` followed by division read the `/` as a regex opener. Reproduced your exact `w: a! / "x/y".length` shape between an `id:` and its `description:` before the fix: the tool's own description was hidden **and** the next tool in the file dropped with no warning at all (0 emitted). `isRegexPosition` now distinguishes prefix logical-NOT (`!/re/`, still a regex position) from a postfix non-null assertion by the character preceding the `!`, so `x! / …` is division while `!/re/` stays a regex. Added a fixture covering both directions, verified to fail if the disambiguation is reverted. Real-tree output unchanged (72 tools, no warnings).

scripts/extract-tools.mjs
latekvo: Fixed in 428e4304 — added a fixture that pins this guard. A `subdescription:` sibling written before the real `description:` must not be captured as the tool's description; the test asserts the real description wins, and I verified it fails (captures the wrong `subdescription:` text into the scan) when the word-boundary check is removed.

scripts/extract-tools.mjs
latekvo: Fixed in 428e4304 — added a fixture that pins this guard. `uuid:` and `grid:` siblings must not be discovered as tool ids; the test asserts only the real `id:` is emitted, and I verified it fabricates spurious `uuid`/`grid` tools when the word-boundary check is removed.

scripts/extract-tools.mjs
latekvo: Agreed, and leaving it out of scope for this PR (as you and the earlier review both noted). Two reasons it belongs in a follow-up rather than here: (1) making const-reference / dynamic ids visible needs actual id *resolution* — following `id: SOME_CONST` to its definition — which is a different capability from the fixed-description-window fix this PR is about; (2) a purely lexical warning on every `id: <identifier>` would also fire on legitimate nested non-tool keys (e.g. `{ id: someVar }` inside a payload/example object), because this extractor has no tool-scope knowledge to tell a real tool from data. So it can't be closed soundly with the lexical approach here.

For the record, the full set currently taken out of the scan by a const-ref/dynamic id is: `app-map-crawl`, `app-map-export`, `app-map-navigate`, `app-map-query`, `await-screen-idle`, and `await-ui-element`. The right follow-up covers const-ref id resolution plus a real-tree emit-count and empty-stderr assertion, so this silent-omission class trips CI if it ever regresses further.

scripts/extract-tools.mjs
latekvo: Follow-up: refined in 627d6c27. Reviewing my own first cut, the single-char check misclassified a prefix `!` written after a keyword — `return !/["']/.test(x)` in a helper — as postfix (the char before the `!` is an identifier char, the keyword's last letter), reintroducing a silent drop. The sound formulation: whether a `/` after `!` is a regex equals whether it would be a regex at the position just *before* the `!` (a prefix `!` leaves the value-expected context unchanged; a postfix `!` inherits the value context of what it follows), so it now recurses into `isRegexPosition` on the preceding significant char — reusing the existing keyword walk-back. Handles `= !/re/`, `return !/re/` (regex) and `x! / y`, `foo()! / y` (division) alike. The fixture now covers the keyword-preceded form too; verified it fails under both the full revert and the prior single-char heuristic. Real-tree output still 72 tools, no warnings.

scripts/extract-tools.mjs
latekvo: Correction for the record: the six-tool list above (mine and the review comment it answered) is wrong. Measured over the whole repo history (`git log --all -S 'app-map'` is empty in both spellings): no `app-map-*` tool has ever existed in this repo. The actual const-ref/dynamic-id set taken out of the scan is exactly two tools - `await-screen-idle` (`AWAIT_SCREEN_IDLE_TOOL_ID`, arrived with the final main merge) and `await-ui-element` (`AWAIT_UI_ELEMENT_TOOL_ID`). The PR body has been corrected accordingly; the follow-up scope (const-ref id resolution + real-tree emit-count assertion) is unchanged, just smaller than stated.

PR #438, opened 2026-06-30
fix(debugger): deliver component-tree errors via binding + stop truncating log levels

latekvo:
Fixes two confirmed defects in the JS-runtime debugger subsystem.

## Defect A — `debugger-component-tree` hangs ~15s and throws on every error condition

**File:** `packages/tool-server/src/utils/debugger/scripts/component-tree.ts`

The injected fiber-walk script delivers its success result through the `__argent_callback` CDP binding, but every error path instead did `return JSON.stringify({ error: ... })` from the async IIFE — `'No DevTools hook'`, `'No fiber roots'`, `'Could not find UIManager'`.

The consumer (`evaluateWithBinding`, `cdp-client.ts`) runs `Runtime.evaluate` with `returnByValue:false`/`awaitPromise:false`, so the IIFE's return value is discarded. The promise only settles when the binding fires or the 15s timeout elapses. So on a release/production build, before the first render, or on a non-RN runtime, `debugger-component-tree` blocked for ~15s and then threw a generic `DEBUGGER_CDP_BINDING_TIMEOUT` instead of returning the intended diagnostic immediately.

**Observed:** callback never invoked (0 calls) on the error paths; the tool hangs to the binding timeout.
**Expected:** the error is delivered through the binding right away and surfaces as `Error: <message>` via the consumer's `JSON.parse(response.result).error` path.

**Fix:** mirror the sibling `inspect-at-point.ts`. Added an `__argent_fail(msg)` helper that routes the error through `__argent_callback` in the exact same wire shape as the success path (`{ requestId, result: JSON.stringify({ error: msg }) }`), and converted all three error paths to call it and `return`.

## Defect B — `LogFileWriter` truncates console levels longer than 5 chars

**File:** `packages/tool-server/src/utils/debugger/log-file-writer.ts`

The flat-line level was rendered as `LEVEL_DISPLAY[entry.level] ?? entry.level.toUpperCase().padEnd(5).slice(0, 5)`. The `.slice(0,5)` truncated any level not in the 5-entry `LEVEL_DISPLAY` map. CDP's `Runtime.consoleAPICalled.type` includes `"warning"` (`console.warn`) and `"assert"` (`console.assert`) — neither is in the map — so `readAll()`/`readFiltered()` returned `"warni"`/`"asser"`, and `readFiltered({ level: "warning" })` matched nothing.

**Observed:** writing level `"warning"` → `readAll()[0].level === "warni"`; `readFiltered({ level: "assert" }).total === 0`.
**Expected:** the canonical level round-trips exactly for levels of any length.

**Fix:** drop `.slice(0,5)` — keep `padEnd(5)` for column alignment but never cut. `LINE_RE` captures the level as `\S+`, so a non-truncated, whitespace-free level round-trips through `parseFlatLine` unchanged. Short levels (log/info/warn/error/debug) are unaffected.

## Tests

- `packages/tool-server/test/debugger/comptree-error-channel.test.ts` — asserts `__argent_callback` is invoked on the error paths and the error is encoded as `parsed.error`.
- `packages/tool-server/test/debugger/log-file-writer-level.test.ts` — asserts `"warning"`/`"assert"` round-trip and filter correctly, and short levels still work.

Both new files fail before the fix and pass after. Full `tool-server` suite green (1627 tests); repo-root `tsc --build` clean; eslint clean on changed files.

### DIFF
--- a/packages/tool-server/src/utils/debugger/log-file-writer.ts
+++ b/packages/tool-server/src/utils/debugger/log-file-writer.ts
@@ -119,8 +119,13 @@ export class LogFileWriter {
 
     // Collapse newlines in message for flat format
     const flatMessage = entry.message.replace(/\n/g, " ");
-    const levelDisplay =
-      LEVEL_DISPLAY[entry.level] ?? entry.level.toUpperCase().padEnd(5).slice(0, 5);
+    // Pad to 5 for column alignment but NEVER truncate: the level must round-trip
+    // exactly through parseFlatLine for levels of any length. CDP emits levels
+    // longer than 5 chars (e.g. "warning" from console.warn, "assert" from
+    // console.assert); slicing to 5 would persist "warni"/"asser" and break
+    // readFiltered({ level: "warning" }). LINE_RE captures the level as \S+, so a
+    // non-truncated, whitespace-free level survives the write→read round-trip.
+    const levelDisplay = LEVEL_DISPLAY[entry.level] ?? entry.level.toUpperCase().padEnd(5);
     const line = `[L:${entry.id}] ${entry.timestamp} ${levelDisplay} ${source} | ${flatMessage}\n`;
 
     if (this.ready && this.fd !== null) {
--- a/packages/tool-server/test/debugger/comptree-error-channel.test.ts
+++ b/packages/tool-server/test/debugger/comptree-error-channel.test.ts
@@ -70,4 +70,24 @@ describe("component-tree delivers errors via the binding channel", () => {
     };
     expect(inner.error).toBe("No fiber roots");
   });
+
+  it("routes an UNEXPECTED throw (outside the named guards) through the binding", async () => {
+    // getFiberRoots throws on the legacy-roots path, which runs outside any inner
+    // try/catch. Without the top-level try/catch the async IIFE would reject and
+    // __argent_callback would never fire — the tool would hang to the 15s binding
+    // timeout instead of surfacing the crash.
+    const hook = {
+      getFiberRoots: () => {
+        throw new Error("boom");
+      },
+    };
+    const payloads = run({ hook });
+    await new Promise((r) => setTimeout(r, 10));
+    expect(payloads.length).toBeGreaterThan(0);
+    const inner = JSON.parse((JSON.parse(payloads[0]!) as { result: string }).result) as {
+      error?: string;
+    };
+    expect(inner.error).toContain("Component-tree script crashed");
+    expect(inner.error).toContain("boom");
+  });
 });
--- a/packages/tool-server/test/debugger/comptree-error-channel.test.ts
+++ b/packages/tool-server/test/debugger/comptree-error-channel.test.ts
@@ -190,4 +190,51 @@ describe("component-tree delivers errors via the binding channel", () => {
     expect(inner.error).toContain("Component-tree script crashed");
     expect(inner.error).toContain("boom");
   });
+
+  it("delivers a crash that happens AFTER the measurement await through the still-installed binding", async () => {
+    // Distinct from the synchronous crash above (getFiberRoots throws before any
+    // await): here every guard passes and the throw happens in the post-await
+    // continuation, after `await Promise.all(...)` on the host measurements. That
+    // is the path closest to the production hang this change fixes — the failure
+    // has to travel through the binding that is still installed across the await,
+    // not be `return`ed. The fiber/module setup is the known-good success case,
+    // except the kept component carries a BigInt accessibilityLabel (read without
+    // a typeof guard), which the post-await `JSON.stringify(result)` cannot
+    // serialize, forcing a genuine async throw into the top-level catch.
+    //
+    // Mutation-checked: ending the try before the await, or dropping the async
+    // delivery, leaves the async IIFE to reject with the binding never firing —
+    // `runToSuccess` then times out with zero payloads and this test fails, while
+    // the synchronous crash test above (captured by the sync `run()` helper) stays
+    // green. That is exactly the ~15s binding-timeout regression this guards.
+    const hostFiber = {
+      type: "RCTView",
+      stateNode: { _nativeTag: 42 },
+      child: null,
+      sibling: null,
+    };
+    const buttonFiber = {
+      type: { displayName: "MyButton" },
+      memoizedProps: { testID: "btn", accessibilityLabel: BigInt(1) },
+      child: hostFiber,
+      sibling: null,
+    };
+    const root = { current: { child: buttonFiber, sibling: null } };
+    const hook = { renderers: new Map([[1, {}]]), getFiberRoots: () => new Set([root]) };
+    const moduleObj = {
+      UIManager: {
+        measureInWindow: (_tag: number, cb: (x: number, y: number, w: number, h: number) => void) =>
+          cb(10, 20, 100, 50),
+      },
+      Dimensions: { get: () => ({ width: 390, height: 844 }) },
+    };
+    const r = (i: number) => (i === 0 ? moduleObj : undefined);
+
+    const payloads = await runToSuccess({ hook, r });
+    expect(payloads.length).toBeGreaterThan(0);
+    const outer = JSON.parse(payloads[0]!) as { requestId: string; result: string };
+    expect(outer.requestId).toBe("t");
+    const inner = JSON.parse(outer.result) as { error?: string };
+    expect(inner.error).toContain("Component-tree script crashed");
+  });
 });
--- a/packages/tool-server/test/debugger/comptree-error-channel.test.ts
+++ b/packages/tool-server/test/debugger/comptree-error-channel.test.ts
@@ -1,0 +1,73 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { makeComponentTreeScript } from "../../src/utils/debugger/scripts/component-tree";
+
+/**
+ * Regression for the component-tree error channel: every error path must deliver
+ * its result through the `__argent_callback` binding, exactly like the success
+ * path. The consumer (`evaluateWithBinding`) runs Runtime.evaluate WITHOUT
+ * returnByValue/awaitPromise, so the IIFE's return value is discarded — an error
+ * that is merely `return`ed never settles the promise and the tool hangs until
+ * the 15s binding timeout. This test asserts the binding IS invoked on error.
+ */
+function run(opts: { hook: unknown }): string[] {
+  const g = globalThis as Record<string, unknown>;
+  const saved = {
+    window: g.window,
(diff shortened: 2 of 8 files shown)


packages/tool-server/src/utils/debugger/log-file-writer.ts
```diff
@@ -119,8 +119,13 @@ export class LogFileWriter {
...
     const flatMessage = entry.message.replace(/\n/g, " ");
-    const levelDisplay =
-      LEVEL_DISPLAY[entry.level] ?? entry.level.toUpperCase().padEnd(5).slice(0, 5);
+    // Pad to 5 for column alignment but NEVER truncate: the level must round-trip
+    // exactly through parseFlatLine for levels of any length. CDP emits levels
+    // longer than 5 chars (e.g. "warning" from console.warn, "assert" from
+    // console.assert); slicing to 5 would persist "warni"/"asser" and break
+    // readFiltered({ level: "warning" }). LINE_RE captures the level as \S+, so a
+    // non-truncated, whitespace-free level survives the write→read round-trip.
+    const levelDisplay = LEVEL_DISPLAY[entry.level] ?? entry.level.toUpperCase().padEnd(5);
```
hubgan, line 128: The value corrected here — the level written to the flat file and reconstructed by parseFlatLine — only reaches a consumer through the on-disk `.log` file that debugger-log-registry returns as `file`. `readFiltered()`, the method the description's `readFiltered({ level: "warning" })` example calls, has no caller outside tests, and the only `readAll()` consumer (the console-log WS replay in createConsoleLogServer) is not itself connected to anything in the repo. `getStats().byLevel` and the clusters were already correct because they key on the raw level. Worth confirming the intended surface is the log file itself, since that's the only place the corrected level is currently observable — the description points at a filter path that nothing invokes today.

packages/tool-server/test/debugger/comptree-error-channel.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+afterEach(() => {
+  const g = globalThis as Record<string, unknown>;
+  delete g.window;
+  delete g.__REACT_DEVTOOLS_GLOBAL_HOOK__;
+  delete g.__argent_callback;
+  delete g.__r;
+});
+
+describe("component-tree delivers errors via the binding channel", () => {
```
hubgan, line 81: This suite covers the 'No DevTools hook' and 'No fiber roots' guards plus an unexpected throw, but not the third guarded path ('Could not find UIManager'), and it never asserts that the success path still delivers its result through the binding now that the whole body is wrapped in the top-level try/catch. A regression in either — the UIManager guard reverting to a bare return, or the try/catch redirecting/suppressing a successful callback — would pass this suite unnoticed.

packages/tool-server/test/debugger/log-file-writer-level.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    w.write({
+      id: 0,
+      timestamp: new Date(1710000000000).toISOString(),
+      level: "warning",
+      message: "x",
+    });
+    expect(w.readAll()[0]!.level).toBe("warning");
+  });
+
+  it("filters by 'assert'", () => {
```
hubgan, line 46: Level-based filtering is asserted only for 'assert'; the 'warning' case checks readAll() but not readFiltered({ level: 'warning' }). Filtering by the canonical multi-character level is the exact behavior this change restores, so that path isn't directly exercised here.

hubgan (commented):
A few inline notes on test coverage and where the log-level change is actually observable. Nothing blocking.

hubgan:
Log timestamps are decoded as year 58470. Every flat-file line this PR's level-fix formats shows +058470-09-…Z on both iOS and Android. Root cause: packages/tool-server/src/blueprints/js-runtime-debugger.ts:221 (and the chromium blueprint) do new Date(entry.timestamp * 1000), but CDP Runtime.consoleAPICalled.timestamp is already milliseconds. Not introduced by #438, but it lives in the same log pipeline and is worth a follow-up. (Doesn't break parsing since the token is space-free.)

packages/tool-server/test/debugger/comptree-error-channel.test.ts
latekvo: Fixed in 407f71f3. Added two cases to `comptree-error-channel.test.ts`:

1. A test for the third named guard (`'Could not find UIManager'`) — hook + a mounted root get past the earlier guards, Fabric is absent, and the stub `__r` yields no module with a `UIManager`, so it hits that guard and fails through the binding.
2. A test asserting the SUCCESS path still delivers its result through `__argent_callback` (`parsed.error === undefined`, correct `screenW/H`, and a components array) now that the whole body is wrapped in the top-level try/catch. The success callback fires asynchronously (after the measurement `Promise.all`), so it uses a helper that keeps the binding installed across the await; the synchronous `run()` helper would drop it.

Both are mutation-checked: reverting the UIManager guard to a bare `return`, or suppressing the final success `__argent_callback`, each makes one of the new tests fail (zero payloads).

packages/tool-server/test/debugger/log-file-writer-level.test.ts
latekvo: Fixed in 407f71f3. Added a `readFiltered({ level: "warning" })` case that asserts `total === 1`, one returned entry, and that the entry's level round-trips as `"warning"` — the exact filter path this change restores. With the old `.slice(0, 5)` the level persists as `"warni"`, so that filter returns `0`; I mutation-checked that direction too.

packages/tool-server/src/utils/debugger/log-file-writer.ts
latekvo: Confirmed — the intended observable surface is the on-disk `.log` file, not the tool's structured body. That file's path is returned as `file` by `debugger-log-registry`, whose description explicitly directs the agent to read it, and its level column now shows the full `WARNING`/`ASSERT` instead of the truncated `WARNI`/`ASSER`.

You're right that the structured body was already correct: `getStats().byLevel` and `getClusters()` key off the in-memory raw `entry.level`, never the flat-file column, so they were never affected. And `readFiltered()`/`readAll()` have no live tool caller today (`readAll()`'s only consumer is the console-replay WS server, whose URL is set on the API object but never wired up). So the round-trip fix is what keeps the persisted file accurate for the agent reading it, and keeps those methods correct for any future / WS consumer.

Leaving this open in case you want to weigh in on whether the file is the surface you had in mind.

hubgan, line 128: This stops the level column of the flat log file from being truncated, which is correct. Flagging that the timestamp column of the same file is separately corrupted: the producer that feeds write() in js-runtime-debugger.ts does new Date(entry.timestamp * 1000), but the CDP consoleAPICalled.timestamp is already in milliseconds, so every line is stamped with a far-future date. Running this build against a device, every captured log line read +058473-..., i.e. year 58473. The Chromium runtime-debugger path does not apply that multiplication, so the RN and Chromium log files currently disagree. Since this change is about the fidelity of exactly this file, the timestamp column is the other half of the same corruption.

packages/tool-server/test/debugger/comptree-error-channel.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    expect(inner.error).toBeUndefined();
+    expect(inner.screenW).toBe(390);
+    expect(inner.screenH).toBe(844);
+    expect(inner.components.length).toBeGreaterThan(0);
+    expect(inner.components[0]!.name).toBe("MyButton");
+    expect(inner.components[0]!.testID).toBe("btn");
+    expect(inner.components[0]!.rect).toEqual({ x: 10, y: 20, w: 100, h: 50 });
+  });
+
+  it("routes an UNEXPECTED throw (outside the named guards) through the binding", async () => {
```
hubgan, line 174: This drives the new top-level catch on its synchronous path only: getFiberRoots throws before the await Promise.all(...) in the script, so the failure is captured synchronously. A throw that reaches the same catch after that await is a distinct case: it has to be delivered asynchronously through the still-installed binding, and it is the path closest to the production hang this change fixes. Nothing in the suite covers it, so a later change that ended the try before the await, or dropped the async delivery, would bring back the ~15s binding timeout with the tests still green.

hubgan (approved):

packages/tool-server/src/utils/debugger/log-file-writer.ts
latekvo: Fixed in 61feff86.

Confirmed from the RN source, not just the arithmetic: RN 0.81.5's `jsinspector-modern` stamps every console message with `getTimestampMs()` (`std::chrono::duration<double, std::milli>`, `ReactCommon/jsinspector-modern/RuntimeTargetConsole.cpp`) and forwards that value onto the CDP `timestamp` field unchanged — so `consoleAPICalled.timestamp` is milliseconds on Hermes/RN exactly as on Chrome. The `* 1000` was multiplying an already-ms value, which lands ~56000 years out (year 58473 matches your on-device read precisely).

The fix pulls the conversion into one shared `consoleTimestampToIso` helper that both blueprints call, so the RN and Chromium files can no longer diverge. It drops the multiply and keeps the non-finite guard (a bare `new Date(NaN).toISOString()` throws RangeError inside the typed-emitter listener and would silently drop the entry). I also corrected the chromium comment that claimed Hermes emits seconds — that assumption was what produced the bug. New `console-timestamp.test.ts` feeds a realistic ms timestamp and asserts the current-millennium year, and pins that the old `* 1000` produced the far-future date.

packages/tool-server/test/debugger/comptree-error-channel.test.ts
latekvo: Fixed in 8f49d73a.

You're right — the existing crash test only drove the synchronous path (`getFiberRoots` throws before the `await`, so the sync `run()` helper captures it). Added a case that forces a genuine post-await throw: the fiber/module setup is the known-good success case except the kept component carries a BigInt `accessibilityLabel`, which the post-await `JSON.stringify(result)` can't serialize — so the failure has to travel through the binding that is still installed across the await, delivered via `runToSuccess`.

Mutation-checked exactly as you framed it: moving the top-level catch to end before the await makes the new test time out with zero payloads (the ~15s hang), while the synchronous crash test stays green.

latekvo:
Fixed in 61feff86 (details on the inline thread). The `* 1000` on the RN path was the culprit — CDP `consoleAPICalled.timestamp` is milliseconds on Hermes/RN too (jsinspector-modern's `getTimestampMs`), so multiplying pushed every line to year 58473. Both blueprints now share one `consoleTimestampToIso` helper so the RN and Chromium files agree, with a regression test.

PR #439, opened 2026-06-30
fix(variants): roll a fresh round when proposing after a completed selection

latekvo:
## Defect

`VariantProposalStore` (packages/tool-server/src/utils/variant-proposals.ts) could leave a round in the `completed && !consumed` state, and a variant proposed in that state was silently dropped from the next selection outcome.

`submitSelection()` sets `completed = true, consumed = false`, and only sets `consumed = true` when there is a parked waiter to settle (`toSettle.length > 0`). When the preview window POSTs a selection while no `await_user_selection` is currently parked (the window lingers open after a submit, so the user can complete a selection during the next proposing phase, or after a pending-timeout before the agent re-awaits), the round stays `completed && !consumed`.

The next `proposeVariant()` for a NEW element called `autoRollIfConsumed()`, which only reset the store when `completed && consumed`. So it did not roll: the new element was appended to the already-completed round. The subsequent `awaitSelection()` then hit the fast path `if (this.completed && !this.consumed)` and returned the frozen `lastOutcome` captured at submit time — which predates the new element. `proposeVariant` reported success (`totalElements` incremented, "staged"), but the element never appeared in any outcome and was never presented for selection.

### Repro (dropped element)

- `proposeVariant("Alpha")` -> `submitSelection(Alpha)` with no awaiter parked -> `proposeVariant("Beta")` (returns `totalElements: 2`) -> `await awaitSelection()` resolves a completed outcome whose `selections`/`unselected` contain "Alpha" but NOT "Beta". Beta is silently dropped.

## Fix

Roll a fresh round whenever a proposal lands on an already-`completed` round, not only when `completed && consumed`. The helper (renamed `autoRollIfCompleted`) now calls `reset()` on `this.completed`, so a post-completed proposal begins a new round and the new element is staged into it. The legitimate submit-then-await fast path (submit with no waiter, then await with no new proposal) is unchanged — `completed` stays true with no intervening propose, so the frozen outcome is still handed back and the round consumed.

With the fix, the repro above starts round 2 with "Beta" as a live proposal; the next await presents Beta (as a pending proposal) instead of dropping it.

## Tests

- New regression test `packages/tool-server/test/variant-store-propose-after-submit.test.ts` — fails before the fix (Beta dropped from the completed outcome), passes after (Beta presented in the fresh round).
- All existing variant-proposals tests pass, including the submit-then-await fast path and the await-then-submit consumed path.
- Full tool-server suite: 1622 passed. `tsc --build` clean; eslint 0 problems on changed files.


packages/tool-server/src/utils/variant-proposals.ts
```diff
@@ -246,6 +246,18 @@ export class VariantProposalStore {
...
+   * receive them directly (submitSelection's "no waiter" branch), queued in
+   * completion order. `autoRollIfCompleted` rolls into a fresh round on the
+   * very next `propose_variant` regardless of whether the outcome was ever
+   * retrieved — so this queue, unlike `lastOutcome`/`completed`/`consumed`
+   * (which describe only the CURRENT round), survives `reset()` and is
+   * drained first by `awaitSelection`, oldest first. Without it, a human's
+   * already-submitted selection is destroyed the moment the next
+   * `propose_variant` call rolls the round out from under it.
+   */
+  private pendingOutcomes: Array<Extract<AwaitOutcome, { status: "completed" }>> = [];
```
hubgan, line 295: This queue intentionally survives `reset()`, but `reset()` is also how `setCliSession()` starts a "clean round," and `await_user_selection` is hidden for the whole duration of a CLI session. So an outcome queued here by an earlier non-CLI submit that had no waiter parked is neither cleared by that clean-start reset nor drained by an await during the session. Once the session ends, the first `await_user_selection` of the next, unrelated round returns this stale outcome as that round's `completed` result, so the agent applies a variant choice belonging to a discarded round. Reproduced end-to-end: `propose(Alpha)` → submit with no await parked → begin+end an `argent lens` session → `propose(Zeta)` → the next await hands back Alpha's outcome instead of parking on Zeta.

packages/tool-server/src/utils/variant-proposals.ts
```diff
@@ -540,17 +579,32 @@ export class VariantProposalStore {
...
-    // the round number, and getLastOutcome stops returning a stale outcome)
-    // rather than appending to this already-submitted one.
-    if (this.cliSession) this.consumed = true;
-    for (const w of toSettle) {
-      w.settled = true;
-      w.settle(this.lastOutcome);
+    if (toSettle.length > 0) {
+      this.consumed = true;
+      for (const w of toSettle) {
+        w.settled = true;
```
hubgan, line 659: Every waiter-less non-CLI submit appends here with no upper bound. If the agent proposes and the human submits repeatedly without any `await_user_selection` running in between, this queue grows without limit, and each entry retains the full `selections`/`annotations` arrays. It sits directly below the `MAX_COMMENT_LENGTH` cap whose comment states user text is bounded "to bound memory regardless of caller" — this queue is the one ingestion path that escapes that guarantee. Confirmed that 50 such submits accumulate 50 undrained entries.

packages/tool-server/src/utils/variant-proposals.ts
```diff
@@ -316,10 +328,19 @@ export class VariantProposalStore {
     }
   }
 
-  private autoRollIfConsumed(): void {
-    // A completed round that has already been handed to the agent is closed —
-    // the next proposal starts a clean round automatically.
-    if (this.completed && this.consumed) this.reset();
+  private autoRollIfCompleted(): void {
```
hubgan, line 366: With this rename and behavior change the roll now fires on any `completed` round, consumed or not. The class-level "Rounds" summary at the top of the file (the paragraph beginning "Rounds: a round is one propose→await→submit cycle") still states that a fresh round opens only after a completed selection is *consumed* by `await_user_selection`, which no longer matches what this method does.

packages/tool-server/test/variant-store-propose-after-submit.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    const s = new VariantProposalStore();
+    s.proposeVariant({ element: "Alpha", variant: variant("A1") });
+    const a = s.snapshot().proposals.find((p) => p.element === "Alpha")!;
+    // Submit with NO await parked → round becomes completed && !consumed.
+    s.submitSelection({ selections: [{ elementId: a.id, variantId: a.variants[0]!.id }] });
+
+    // A new element proposed during this lingering completed round must begin a
+    // fresh round, not vanish behind the frozen pre-submit outcome.
+    const b = s.proposeVariant({ element: "Beta", variant: variant("B1") });
+    expect(b.totalElements).toBeGreaterThanOrEqual(1); // Beta is accepted
```
hubgan: This assertion can never fail: after the fresh round rolls, `totalElements` is 1, and even the pre-fix behavior (appending Beta onto the already-completed round) returned 2 — both satisfy `>= 1`. So it does not actually guard against the dropped-element regression; the meaningful coverage is the `mentioned`-set check further down. As written it reads like a guard but exercises nothing.

hubgan (commented):
A few inline notes on the round and pending-outcome lifecycle.

packages/tool-server/src/utils/variant-proposals.ts
latekvo: Fixed in a0f53d12. `setCliSession(true)` clears `pendingOutcomes` at the session boundary (unconditionally within the begin transition), so a queued outcome from a prior non-CLI waiter-less submit can't survive into the session.

I re-ran your exact repro (`propose(Alpha)` → waiter-less submit → begin+end an `argent lens` session → `propose(Zeta)` → await): with that boundary clear removed the await returns Alpha's outcome (the leak); with it in place the await parks on Zeta (`status: "pending"`). Guarded by the regression test `starting a CLI session does not leak a prior waiter-less outcome into its first await`.

packages/tool-server/src/utils/variant-proposals.ts
latekvo: Fixed in 9e0f9a62. Added a `MAX_PENDING_OUTCOMES` (32) cap; on overflow the oldest entries are dropped (the caller is furthest behind on those), restoring the `MAX_COMMENT_LENGTH` "bound memory regardless of caller" guarantee for this path.

Verified with your 50-submit repro (the queue now settles at 32 instead of 50) and a regression test `bounds the undelivered-outcome queue against a runaway waiter-less submitter`.

packages/tool-server/src/utils/variant-proposals.ts
latekvo: Fixed in 9e0f9a62. Rewrote the class-level "Rounds" summary: a fresh round now opens on the next `propose_variant` after any completed round — whether or not the outcome was already consumed — and a selection submitted with no await parked is preserved across that roll and delivered on the next await.

packages/tool-server/test/variant-store-propose-after-submit.test.ts
latekvo: Fixed in 9e0f9a62. Replaced `expect(b.totalElements).toBeGreaterThanOrEqual(1)` (satisfied by both the pre-fix append and the post-fix roll) with `expect(b.round).toBe(2)` + `expect(b.totalElements).toBe(1)`, which fails under the pre-fix behavior (round 1, totalElements 2) and so actually guards the dropped-element regression.

packages/tool-server/src/utils/variant-proposals.ts
```diff
@@ -617,6 +691,33 @@ export class VariantProposalStore {
...
    */
   awaitSelection(opts: { signal?: AbortSignal; timeoutMs: number }): Promise<AwaitOutcome> {
+    // Honor an already-aborted signal before touching any state. The caller has
+    // disconnected, so there is no one to receive a resolved value — and, most
+    // importantly, the `pendingOutcomes` fast-path below MUTATES state (it
+    // `shift()`s a queued outcome). Draining it for a dead caller would destroy
+    // a human's already-submitted selection, the exact loss `pendingOutcomes`
+    // exists to prevent. Reject with the same AbortError the parked path uses so
+    // callers handle both the pre-aborted and mid-park cases identically.
+    if (opts.signal?.aborted) {
```
hubgan, line 764: This early return handles an already-aborted signal before any state is touched. It also makes the later `if (opts.signal.aborted) return onAbort();` inside the Promise executor unreachable: every statement between here and that executor runs synchronously with no await, so the signal cannot transition to aborted in between. The two abort checks now encode the same condition in two places and can drift apart.

packages/tool-server/src/utils/variant-proposals.ts
```diff
@@ -540,17 +592,39 @@ export class VariantProposalStore {
...
+    } else if (!this.consumed) {
+      // No waiter was parked to receive this outcome directly. Queue it so a
+      // later await_user_selection still gets it, even across an intervening
+      // propose_variant that rolls the round (see `pendingOutcomes`' doc).
+      // Dedup by round: a repeat submit of this same round REPLACES its queued
+      // copy rather than enqueuing a phantom duplicate a second await would
+      // deliver. (`consumed` is true here only for an already-delivered round,
+      // whose outcome must not be re-queued.)
+      this.pendingOutcomes = this.pendingOutcomes.filter((o) => o.round !== round);
+      this.pendingOutcomes.push(this.lastOutcome);
```
hubgan, line 681: Retaining the frozen outcome here is what preserves a waiter-less submission across a round roll. One note on the memory-bound intent: MAX_PENDING_OUTCOMES caps the number of queued outcomes but not their size. Selection, annotation and global comments are length-capped, but an annotation's `match.value` is ingested uncapped and the selection route accepts an unbounded number of annotations, so each queued outcome can be arbitrarily large. Previously that data lived only in a single `lastOutcome`; it is now retained across up to 32 entries, so the effective retained size for a caller that submits without ever awaiting is roughly 32x larger.

hubgan (commented):
Reviewed the store state machine and ran the change end-to-end. The fix looks correct: rolling a fresh round on any completed round, with the undelivered outcome preserved in the bounded pendingOutcomes queue and drained oldest-first, closes the dropped-element path without discarding an already-made selection. I traced the round / completed / consumed / pendingOutcomes transitions across the propose, submit (with and without a parked waiter), await, reset and CLI-session orderings and the invariants hold up (round is strictly monotonic; consumed implies completed), so the removed completed-and-not-consumed await fast-path is genuinely covered by the queue drain.

I also drove it against real devices: propose then submit-with-no-awaiter then propose(new) then await, through the real tools and the real selection route, on both an iOS and an Android simulator. The new element is presented in a fresh round and the earlier submission is delivered first, exactly as intended; an idempotent double-submit delivers a single completion. The suite passes, and the new regression tests fail on main as expected.

Comments below are all minor: a couple of cleanups, one memory-bound observation, a tool-description note, and a few test-coverage gaps. None are blocking.

packages/tool-server/src/utils/variant-proposals.ts
```diff
@@ -540,17 +592,39 @@ export class VariantProposalStore {
...
+      // Bound the queue so a caller repeatedly submitting without ever awaiting
+      // cannot grow it without limit (the route is unauthenticated). Drop the
+      // oldest entries — the agent is furthest behind on those — mirroring the
+      // MAX_COMMENT_LENGTH "bound memory regardless of caller" guarantee.
+      if (this.pendingOutcomes.length > MAX_PENDING_OUTCOMES) {
+        this.pendingOutcomes.splice(0, this.pendingOutcomes.length - MAX_PENDING_OUTCOMES);
```
hubgan: This overflow trim only ever removes exactly one element: the queue is bound-checked immediately after a single push (and the dedup filter above removes at most one entry first), so its length here is at most MAX_PENDING_OUTCOMES + 1. The general `splice(0, length - MAX)` form reads as if it handles a multi-element overflow that cannot occur on this path.

packages/tool-server/src/utils/variant-proposals.ts
```diff
@@ -617,6 +691,33 @@ export class VariantProposalStore {
...
+      const err = new Error("await_user_selection aborted (client disconnected)");
+      err.name = "AbortError";
+      return Promise.reject(err);
+    }
+
+    // Deliver the OLDEST undelivered outcome first, even if propose_variant
+    // has since rolled the store into a fresh round for new elements — this
+    // outcome is real, already-decided-by-the-human data and must never be
+    // silently lost to a later roll. See `pendingOutcomes`' doc comment.
+    if (this.pendingOutcomes.length > 0) {
```
hubgan, line 774: When a proposal follows a waiter-less submit, this drain returns the previous round's `completed` outcome on the next await_user_selection, and a second await is then needed to surface the freshly-rolled round. The await_user_selection description tells the agent to re-await on a `pending` result but not on a `completed` one, so an agent may apply the returned selections and end its turn, leaving the new round's proposals unpresented until it happens to await again.

packages/tool-server/test/variant-store-propose-after-submit.test.ts
```diff
@@ -0,0 +1,179 @@
...
+        for (const x of out.proposedElements) mentioned.add(x.element);
+      }
+    }
+    // Beta must be accounted for somewhere (presented for selection or in the
+    // outcome), never silently dropped.
+    expect(mentioned.has("Beta")).toBe(true);
+    expect(mentioned.has("Alpha")).toBe(true);
+  });
+
+  it("delivers the already-submitted-but-undelivered outcome before the rolled round's pending state", async () => {
```
hubgan, line 43: This is the one test that delivers a waiter-less outcome through the queue-drain path, but it only checks the element name. The comment / globalComment / annotation fields are asserted only on the parked-waiter settle path elsewhere, never through this drain path, yet the drain depends on buildOutcome() having frozen a separate object before reset() clears the live comment and annotation state. That retention across the roll is currently unpinned by any assertion.

packages/tool-server/test/variant-store-propose-after-submit.test.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    // The decision survived: a fresh live await still delivers Alpha.
+    const retry = await s.awaitSelection({ timeoutMs: 100 });
+    expect(retry.status).toBe("completed");
+    if (retry.status === "completed") {
+      expect(retry.selections.map((x) => x.element)).toContain("Alpha");
+    }
+  });
+
+  it("starting a CLI session does not leak a prior waiter-less outcome into its first await", async () => {
```
hubgan, line 168: This covers that a prior outcome does not leak into a CLI session's first await, but no test awaits after a submit made during a CLI session. The submit path that sets consumed = true and deliberately queues nothing (CLI session, no parked waiter) is never observed through awaitSelection, so a regression there (queuing a phantom completion, or failing to mark consumed and stranding the next await) would not be caught.

packages/tool-server/test/variant-store-propose-after-submit.test.ts
```diff
@@ -0,0 +1,179 @@
...
+    // A queued-but-undelivered outcome from a prior (possibly non-CLI) flow
+    // must not surface as the CLI session's first await result.
+    const s = new VariantProposalStore();
+    s.proposeVariant({ element: "Alpha", variant: variant("A1") });
+    const a = s.snapshot().proposals.find((p) => p.element === "Alpha")!;
+    s.submitSelection({ selections: [{ elementId: a.id, variantId: a.variants[0]!.id }] });
+    s.setCliSession(true, [{ id: "x", name: "X" }]);
+    const out = await s.awaitSelection({ timeoutMs: 100 });
+    expect(out.status).not.toBe("completed");
```
hubgan: not.toBe("completed") also passes for a `pending` result, which would indicate a stranded await rather than the intended clean start. The precise expectation for a fresh CLI session with nothing queued is `no_proposals`, so this assertion is looser than the property it guards.

hubgan (approved):

packages/tool-server/src/utils/variant-proposals.ts
latekvo: Fixed in 9ea79ba2. Removed the now-unreachable `if (opts.signal.aborted) return onAbort();` inside the executor. The top-of-function early return is the single abort guard, and since nothing between it and the executor awaits, the signal cannot flip to aborted in between — so only the future-abort listener is registered now and the two checks can no longer drift apart.

packages/tool-server/src/utils/variant-proposals.ts
latekvo: Fixed in 9ea79ba2. Bounded each queued outcome's size, not just the count: on ingestion in `submitSelection`, an annotation's `match.value` is clamped to `MAX_MATCH_VALUE_LENGTH` (200) and the annotation count to `MAX_ANNOTATIONS` (200), restoring the "bound memory regardless of caller" guarantee on the unauthenticated route. The new regression test `bounds a queued outcome's size` asserts both are clamped in the frozen, queued outcome (it fails pre-fix, with 1000 annotations / a 5000-char matcher value retained).

packages/tool-server/src/utils/variant-proposals.ts
latekvo: Fixed in 9ea79ba2. Replaced `splice(0, length - MAX)` with a single `shift()` and spelled out the invariant in the comment: this path adds at most one entry per submit (the dedup filter above removes at most one first), so the queue can overshoot the cap by at most one and a single oldest-drop restores it.

packages/tool-server/src/utils/variant-proposals.ts
latekvo: Fixed in 9ea79ba2. The drained completed outcome now carries `morePending: true` whenever the drain leaves more to see — a further queued outcome, or a freshly-rolled live round that superseded this one. The `await_user_selection` description now tells the agent to re-await while `morePending` is set (and documents the `round` field), so it no longer applies-and-stops on the drained result and strands the new round. Guarded by `signals morePending when a completed outcome is drained while a fresh round is live` (fails pre-fix).

packages/tool-server/test/variant-store-propose-after-submit.test.ts
latekvo: Fixed in 9ea79ba2. Added `preserves selection/global/annotation comments across the roll on the drain path`: it submits a selection comment, a globalComment and an annotation, rolls the round with a `propose_variant`, then drains and asserts all three survive — pinning that `buildOutcome()` froze a separate object before `reset()` cleared the live comment/annotation state.

packages/tool-server/test/variant-store-propose-after-submit.test.ts
latekvo: Fixed in 9ea79ba2. Added `delivers a submit made DURING a CLI session (consumed, nothing queued) as a clean no_proposals`, which drives the CLI-session submit branch (sets `consumed = true`, queues nothing) through `awaitSelection` and asserts it returns `no_proposals` — no phantom completion and no strand.

packages/tool-server/test/variant-store-propose-after-submit.test.ts
latekvo: Fixed in 9ea79ba2. Tightened the assertion from `not.toBe("completed")` to `toBe("no_proposals")`, the precise clean-start expectation (a stranded `pending` would now fail).

PR #440, opened 2026-06-30
fix(describe): preserve accessibility label/id when collapsing Android wrappers

latekvo:
## Summary

The Android uiautomator describe parser collapses redundant wrapper nodes in two places. Both sites kept the inner/child node and silently discarded the wrapper's own accessibility `content-desc` label and `resource-id`, even when the surviving node had neither. The result is that a labelled tap target becomes anonymous in the describe tree, so an agent can no longer identify or target it.

This change makes both collapse sites merge rather than drop: the surviving node still prefers its own label/identifier (the existing "inner is more specific" intent), but now falls back to the wrapper's values instead of throwing them away.

## Defect A — duplicate-wrapper collapse (`computeNodeOutput`)

When a clickable parent and its single clickable child have identical bounds, the duplicate-wrapper collapse keeps only the inner child and discards the parent's `content-desc`/`resource-id` — even when the inner child has neither. A React Native `Pressable` with an `accessibilityLabel` wrapping a bare clickable native view (label only on the outer node) collapses to an anonymous node.

Repro:
```
<node clickable content-desc="Like, 5 likes">
  <node clickable content-desc="" />
</node>
```
- Observed: collapses to a single node with NO label.
- Expected: the collapsed node keeps `label="Like, 5 likes"`.

Fix: when collapsing, carry the wrapper's label/`resource-id` onto the surviving child only when the child lacks its own. The existing test that asserts an inner `inner` label wins over an outer `outer` label is preserved (child value still takes precedence when present).

## Defect B — bounds-less single-child pass-through (`finalizeUiNode`)

A node with no parseable `bounds`, exactly one child, and its OWN `content-desc` was replaced wholesale by its child (`if (children.length === 1) return children[0]`), discarding the wrapper's label/identifier. This is reachable for the bounds-less Compose group containers the parser already supports when such a container carries a `content-desc`.

Repro:
```
<node content-desc="Profile photo">
  <node bounds=... content-desc="avatar" />
</node>
```
- Observed: keeps only `avatar`; `Profile photo` is lost.
- Expected: both labels survive.

Fix: only pass a single child straight through when the wrapper has no label/identifier of its own. When the wrapper carries its own identity, keep it as a node whose frame is the union of its (sole) child — so neither the wrapper's nor the child's label is dropped. The bounds-less multi-child path (Compose containers with 2+ children) is unchanged.

## Tests

New regression test `packages/tool-server/test/describe-android-label-preserve.test.ts` covers both defects. It fails before the fix (`[]` for A, `['avatar']` for B) and passes after. The full existing uiautomator/describe android suites remain green.

### DIFF
--- a/packages/tool-server/src/tools/describe/platforms/android/uiautomator-parser.ts
+++ b/packages/tool-server/src/tools/describe/platforms/android/uiautomator-parser.ts
@@ -313,6 +313,17 @@ function descendantText(parsed: ParsedXmlNode, maxChars = 120): string {
   const stack: ParsedXmlNode[] = [parsed];
   while (stack.length > 0) {
     const x = stack.pop()!;
+    // A password field's `text` is the secret. Never mine it — or its subtree —
+    // into a borrowed container label; this walk reads the raw XML and would
+    // otherwise bypass the "[password]" redaction the rest of the parser
+    // applies to a password node's own label. Substitute the redacted marker.
+    if (attrIsTrue(x.attrs, "password")) {
+      if (!seen.has("[password]")) {
+        seen.add("[password]");
+        parts.push("[password]");
+      }
+      continue;
+    }
     for (const k of ["text", "content-desc"] as const) {
       const v = (x.attrs[k] ?? "").trim();
       if (v && !seen.has(v)) {
@@ -451,6 +462,15 @@ function computeNodeOutput(
   const interactive = isInteractive(attrs);
   let label = labelOf(attrs);
 
+  // Password fields: keep the ref but never leak the value. Redact HERE, at the
+  // point the label is derived, not just before `makeUiNode` — the collapse
+  // sites below copy this node's `label` onto a surviving child and return
+  // early, so a late redaction would be bypassed and the secret would ride out
+  // on the collapsed tap target. Redacting at the source covers every path.
+  if (attrIsTrue(attrs, "password")) {
+    label = "[password]";
+  }
+
   // A resource-id is still worth keeping even with no label — dropping it
   // silently here is the same class of bug as the collapse cases below fixed
   // (a decorative-looking wrapper can still be a real, agent-addressable
@@ -530,11 +550,6 @@ function computeNodeOutput(
           !c.clickable
         )
     );
-  }
-
-  // Password fields: keep the ref but never leak the value.
-  if (attrIsTrue(attrs, "password")) {
-    label = "[password]";
   }
 
   const node = makeUiNode(attrs, deriveUiAutomatorRole(cls), bounds, label, keptChildren);
--- a/packages/tool-server/src/tools/describe/platforms/android/uiautomator-parser.ts
+++ b/packages/tool-server/src/tools/describe/platforms/android/uiautomator-parser.ts
@@ -467,8 +467,19 @@ function computeNodeOutput(
 
   // Layout container with no own info — pass children through. With
   // --compressed dumps this is what flattens FrameLayout > LinearLayout >
-  // ConstraintLayout chains down to their actual content.
-  if (LAYOUT_CONTAINERS.has(cls) && !interactive && !label && !identifier) {
+  // ConstraintLayout chains down to their actual content. Keeping such a
+  // container just because it carries a resource-id would re-inflate the tree
+  // the v2 trim exists to shrink AND break the duplicate-wrapper collapse
+  // below (an id-bearing, non-clickable middle container stops a clickable
+  // parent from collapsing onto its clickable child). So still pass through —
+  // but don't silently lose the id: propagate it onto a lone surviving child
+  // that has none, so a container-only identifier stays addressable. With 0 or
+  // 2+ kept children there is no single target, so the scaffolding id is
+  // dropped, exactly as it was before id-preservation was added.
+  if (LAYOUT_CONTAINERS.has(cls) && !interactive && !label) {
+    if (identifier && keptChildren.length === 1 && !keptChildren[0]!.identifier) {
+      keptChildren[0]!.identifier = identifier;
+    }
     return keptChildren;
   }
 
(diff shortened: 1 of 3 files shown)


packages/tool-server/src/tools/describe/platforms/android/uiautomator-parser.ts
```diff
@@ -486,6 +493,14 @@ function computeNodeOutput(
   if (interactive && bounds && keptChildren.length === 1) {
     const c = keptChildren[0]!;
     if (c.clickable && c.pixelBounds && rectsEqual(c.pixelBounds, bounds)) {
+      // The inner node is usually the more specific one, so prefer its own
+      // label/identifier. But the accessibility label can live ONLY on the
+      // outer wrapper (an RN `Pressable accessibilityLabel` wrapping a bare
+      // clickable native view) — fall back to the wrapper's values so the
+      // collapsed tap target isn't left anonymous instead of dropping them.
+      if (!c.label && label) c.label = label;
```
hubgan, line 514: duplicate-wrapper collapse can now leak password text. The PR copies the outer wrapper’s label onto the child and returns before the existing password redaction at packages/tool-server/src/tools/ describe/platforms/android/uiautomator-parser.ts:525.

latekvo: Fixed in e5daac8d.

Moved the password redaction to the point the label is derived, so the duplicate-wrapper collapse's early `return [c]` can no longer bypass it — an outer `password="true"` node's value is `[password]` before it can be copied onto the collapsed child.

While verifying, I found the same redaction was also bypassed one layer over: `descendantText()` (the compound-clickable label borrow) reads the raw XML and mined a password descendant's `text` verbatim into a non-password container's label. That path now substitutes `[password]` too. Regression tests added for both.

packages/tool-server/src/tools/describe/platforms/android/uiautomator-parser.ts
```diff
@@ -486,6 +506,14 @@ function computeNodeOutput(
...
```
hubgan, line 514: The collapse carries the wrapper's `label` and `resource-id` onto the surviving child, but not its `text`. When a wrapper's `content-desc` and `text` diverge (e.g. `content-desc="Cart" text="3 items"`), the `3 items` value is dropped while `Cart` is preserved. This matches the pre-fix behavior (the whole wrapper was discarded before), so it is not a regression, but the label/id-versus-value asymmetry is currently untested and easy to overlook.

packages/tool-server/src/tools/describe/platforms/android/uiautomator-parser.ts
```diff
@@ -568,12 +591,17 @@ function finalizeUiNode(
...
+    // whole subtree with `null` silently dropped every child on Compose
+    // hierarchies that emit bounds-less group containers — preserved here for
+    // that case. A single-child wrapper that carries its OWN label/identifier
+    // (e.g. a bounds-less Compose group container with a content-desc) must NOT
+    // be flattened into its child: doing so discards that accessibility label.
+    // Keep it as a node whose frame is the union of its (here, sole) child so
+    // neither the wrapper's nor the child's label is dropped.
     if (children.length === 0) return null;
-    if (children.length === 1) return children[0]!;
+    if (children.length === 1 && !n.label && !n.identifier) return children[0]!;
```
hubgan, line 604: The added `!n.identifier` clause retains a bounds-less single-child wrapper whenever it carries an identifier. In practice an id-only bounds-less layout container (e.g. `android.view.View` with a `resource-id` and no label) is already flattened one stage earlier by the `LAYOUT_CONTAINERS` passthrough, which drops its id — the same id-drop that was deliberately reverted for the passthrough sites earlier in this PR. So this clause only takes effect for a bounds-less, non-container, id-bearing, label-less node, which is a narrow and currently untested path. Worth confirming that retaining the id here is intended and consistent with that earlier revert, and pinning it with a test.

packages/tool-server/src/tools/describe/platforms/android/uiautomator-parser.ts
```diff
@@ -568,12 +591,17 @@ function finalizeUiNode(
...
     const x1 = Math.min(...children.map((c) => c.frame.x));
```
hubgan, line 605: When a kept bounds-less wrapper and its sole child carry the same label, both now surface as separate nodes at the same frame — e.g. a `content-desc="Save"` wrapper over a `text="Save"` child yields two `Save` entries. The substring-dedup filter in `computeNodeOutput` does not cover this because a non-interactive wrapper is not run through it. It is only a readability redundancy in the tree, not a wrong tap target, but flagging it as a visible side effect of keeping the wrapper.

packages/tool-server/test/describe-android-label-preserve.test.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+  return out;
+}
+describe("describe android collapse preserves label/identifier", () => {
+  it("duplicate-wrapper collapse keeps the outer-only label", () => {
+    const xml = `<?xml version='1.0' encoding='UTF-8'?>\n<hierarchy>\n  <node class="android.view.ViewGroup" bounds="[0,0][100,100]" clickable="true" content-desc="Like, 5 likes">\n    <node class="android.widget.Button" bounds="[0,0][100,100]" clickable="true" content-desc="" text=""/>\n  </node>\n</hierarchy>`;
+    const labels = flatten(parseUiAutomatorDump(xml, 100, 100))
+      .map((n) => n.label)
+      .filter(Boolean);
+    expect(labels).toContain("Like, 5 likes");
```
hubgan, line 20: This asserts the label appears somewhere in the flattened tree but does not assert the node count, so it would still pass if the collapse regressed to two stacked nodes with one of them carrying the label. The analogous duplicate-wrapper test in `uiautomator-parser-v2-trim.test.ts` pins the result with a `toHaveLength(1)` check; mirroring that here would let this guard against a de-collapse regression as well as against label loss.

hubgan (commented):
A few non-blocking notes on the collapse edge cases and test coverage, all inline.

hubgan (approved):

PR #441, opened 2026-06-30
fix(installer): stop clobbering foreign and JSONC MCP server config

latekvo:
## Summary

The MCP config installer (`packages/argent-installer/src/mcp-configs.ts`) is supposed to only ever touch the `argent` entry in a user's editor MCP config. Two confirmed defects broke that contract and silently corrupted unrelated configuration. This PR fixes both across every backend (JSON, JSONC, and TOML) and adds regression coverage.

## Defect A — uninstall over-prunes unrelated config

`writeJsonOrRemove(value)` ran the whole config through `pruneEmptyConfig`, which recursively drops **any** empty object/array anywhere in the tree, and then wrote the pruned result. So when `remove()` deleted the argent entry and wrote back, other MCP servers lost their `args: []` / `env: {}` keys, and any sibling user key containing an empty object/array was deleted. The TOML-backed Codex config had the identical bug in `writeTomlOrRemove`.

**Repro:** a Cursor `.cursor/mcp.json` with
`mcpServers: { argent: {...}, other: { command: "other-bin", args: [], env: {} } }, userSettings: { theme: "dark", overrides: {} }`

- Observed after `Cursor.remove(path)`: `other` becomes just `{ command: "other-bin" }` (its `args`/`env` gone) and `userSettings.overrides` is stripped.
- Expected: `other` and `userSettings` are preserved byte-for-byte; only `argent` is removed.

## Defect B — clobbering a JSONC config on write

`.vscode/mcp.json` is JSONC (VS Code allows comments and trailing commas), but the VS Code adapter read it with `readJson`, whose `catch { return {} }` turns any non-strict-JSON file into `{}`. `write` then persisted only `{ servers: { argent } }`, destroying every pre-existing user server (and comments). The same flaw affected the other `readJson → writeJson` adapters — most importantly **Cursor** and **Kiro**, which are comment-tolerant VS Code forks whose `mcp.json` legitimately contains comments, plus **Claude Code**, **Windsurf**, and **Gemini** (strict JSON, but a hand-added comment would have wiped the file). The Zed and opencode adapters were already hardened with `readJsonc` + `editJsoncFile`.

**Repro:** a Cursor/Kiro `mcp.json` containing a `// comment` and `mcpServers: { myserver: {...} }`

- Observed after the adapter's `write(path, getMcpEntry())`: `myserver` and the comment are gone.
- Expected: `myserver` and the comment survive; `argent` is added alongside.

## Fixes

- **A:** `writeJsonOrRemove` no longer deep-prunes. It writes `data` unchanged, and deletes the file only when the top-level object has no own keys. Each remover collapses just its own now-empty argent container along the argent path (new `deleteIfEmpty` helper) before writing, so a config that held only argent still results in file deletion. Foreign servers and user keys are preserved exactly. `writeTomlOrRemove` received the identical non-deep-pruning treatment so the Codex TOML config is protected the same way (`pruneEmptyConfig` is removed entirely). Applied to the Cursor / Claude Code / Windsurf / Gemini / Kiro / Codex removers, the Cursor allowlist remover, and `removeClaudePermission`.
- **B:** the JSONC-safe write path (read with `readJsonc`, mutate via `editJsoncFile` — path-targeted edits that preserve comments and foreign servers) now covers **VS Code** (`servers` key) and the `mcpServers`-keyed adapters **Cursor, Kiro, Claude Code, Windsurf, and Gemini**, matching the Zed/opencode adapters. `write`, `remove`, and `hasArgentEntry` all go through the JSONC path per adapter so a commented file round-trips cleanly on install *and* uninstall. JSONC is a superset of JSON, so this is safe for the strict-JSON adapters and unifies the write path; each adapter keeps its own key structure, value shape (e.g. Claude Code's `type: "stdio"`), and config location. The allowlist helpers (Cursor's `permissions.json`, and the `autoApprove`/`alwaysAllow`/`trust` toggles) are left on `readJson`/`writeJson` — they are a separate concern and never clobber (they early-return when the argent entry is absent), so they stay out of scope.

## Tests

Added to the `describe("installer preserves foreign MCP config")` block in `packages/argent-installer/test/mcp-configs.test.ts`. Each new test **fails on the original code** and **passes with the fix**:

- Defect A (JSON): `Cursor.remove` keeps a foreign server's empty `env`/`args` and a sibling user key.
- Defect A (TOML): `Codex.remove` keeps a foreign server's `args = []` and a sibling empty table, and deletes the file when argent was the only content.
- Defect B (JSONC): parametrized `write` test over **Cursor, Claude Code, Windsurf, Gemini, Kiro** — a hand-authored comment and a pre-existing foreign server both survive an argent `write` (alongside the existing VS Code case).

Full installer suite: 309 passed. `tsc --noEmit` (src + tests), `eslint`, and `prettier` all clean.


packages/argent-installer/src/mcp-configs.ts
```diff
@@ -213,6 +223,7 @@ const cursorAdapter: McpConfigAdapter = {
     if (idx === -1) return;
     list.splice(idx, 1);
     config.mcpAllowlist = list;
+    deleteIfEmpty(config, "mcpAllowlist");
```
hubgan: Cursor's `removeAllowlist` is safe here, but the sibling `addAllowlist` just above (the `~/.cursor/permissions.json` writer) still goes through `readJson`, whose `catch { return {} }` discards the entire file on any comment or trailing comma. Cursor is a comment-tolerant VS Code fork, so a user-commented or multi-rule `permissions.json` gets flattened to just `{ "mcpAllowlist": [...] }` on the next `addAllowlist`, dropping every other rule and comment — the same clobber this PR fixes for `mcp.json`, one file over.

It's also newly reachable through `update`: now that `hasArgentEntry` reads with `readJsonc`, a commented `.cursor/mcp.json` that `update` used to skip is detected as configured, so `update` will call this adapter's `addAllowlist` and touch `permissions.json` for those users. The "allowlist helpers never clobber" rationale in the description holds for the removers (they early-return on a missing entry) but not for `addAllowlist`, which reads-and-writes unconditionally.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -1005,11 +1020,14 @@ export function removeClaudePermission(root: string, scope: "local" | "global"):
...
-  const allow = (config?.permissions as Record<string, unknown>)?.allow as string[];
-  if (!Array.isArray(allow)) return;
+  const permissions = config?.permissions as Record<string, unknown> | undefined;
+  const allow = permissions?.allow as string[] | undefined;
+  if (!permissions || !Array.isArray(allow)) return;
   const idx = allow.indexOf(PERMISSION_RULE);
   if (idx === -1) return;
   allow.splice(idx, 1);
+  deleteIfEmpty(permissions, "allow");
+  deleteIfEmpty(config, "permissions");
```
hubgan: `removeClaudePermission` is hardened here, but its counterpart `addClaudePermission` still reads `.claude/settings.json` with `readJson`, appends to `permissions.allow`, and writes back. If that file ever contains a comment (non-strict JSON), the read returns `{}` and the write drops the user's other permissions — the same unconditional read-write shape as the Cursor allowlist above. Lower likelihood since Claude's settings.json is normally strict JSON, but it's the same class left unmigrated by this PR.

packages/argent-installer/src/utils.ts
```diff
@@ -317,7 +318,12 @@ export function editJsoncFile(filePath: string, jsonPath: JSONPath, value: unkno
   }
 
   fs.mkdirSync(path.dirname(filePath), { recursive: true });
-  fs.writeFileSync(filePath, (hadBom ? "﻿" : "") + text);
+  // A freshly-created config gets a trailing newline, matching writeJson/TOML
+  // (POSIX text-file convention, avoids a git "\ No newline at end of file").
+  // An existing file's own EOL/formatting is preserved untouched — setJsoncIn
+  // edits in place — so only append when the file did not exist before.
+  const out = !existedBefore && !text.endsWith("\n") ? text + "\n" : text;
```
hubgan: This `!existedBefore` branch is the "preserve an existing file's formatting, only add a trailing newline on fresh create" behavior. The fresh-create half is covered by a test, but I don't see one that seeds an existing file *without* a trailing newline and asserts `write()` leaves it that way — regressing this to an unconditional append would still pass the suite. The behavior is correct as written; flagging the branch as unpinned.

packages/argent-installer/test/mcp-configs.test.ts
```diff
@@ -1602,3 +1705,167 @@ describe("generated configs are portable across machines (issue #238)", () => {
...
+  });
+
+  // Same Defect B, but for the { mcpServers } adapters: Cursor and Kiro are
+  // comment-tolerant VS Code forks, and Claude / Windsurf / Gemini configs are
+  // strict JSON that JSONC is a superset of. All five now write through
+  // editJsoncFile, so a hand-authored comment and a pre-existing foreign server
+  // survive an argent `write` — the old readJson → writeJson path reduced the
+  // commented file to {} and rewrote it with only the argent entry.
+  for (const name of ["Cursor", "Claude Code", "Windsurf", "Gemini", "Kiro"]) {
+    it(`${name} write preserves a comment and a foreign server in a JSONC file`, () => {
```
hubgan, line 2056: Two behaviors this suite covers indirectly but doesn't pin directly:

1. Re-`write()` replacement — no test writes an entry carrying `env`, then re-writes the env-less `getMcpEntry()` to prove a stale `env` key is dropped rather than merged. The `update` upgrade path over a pre-#238 entry that still had `env` depends on this being a replace.
2. The Defect A JSONC-remove case where `argent` is the only server beside a top-level empty sibling, e.g. `{ mcpServers: { argent }, userThing: {} }` — the Cursor Defect A test keeps a second server, so the ancestor-prune-then-keep-file path never fires while a top-level empty sibling has to survive.

Both are correct in practice; noting as coverage rather than defects.

hubgan (commented):
Reviewed the installer changes and drove the write/remove paths end-to-end across all backends — both defects are genuinely fixed: foreign servers and comments are preserved on install *and* uninstall for JSON, JSONC, and TOML, and typecheck/lint/the full suite are clean. A few inline notes below — one scope gap in the allowlist helpers that the "only touch argent" contract doesn't yet cover (pre-existing, but newly reachable via `update`), plus a couple of behaviors that are correct but not pinned by a test.

packages/argent-installer/src/mcp-configs.ts
latekvo: Fixed in 837c44e7. `cursorAdapter.addAllowlist` (and `removeAllowlist`) now read `~/.cursor/permissions.json` with `readJsonc` and write through `editJsoncFile`, so a hand-commented or multi-rule file keeps its comments and foreign rules: the `argent:*` pattern is merged into the existing `mcpAllowlist` array rather than the whole file being flattened to `{ mcpAllowlist: [...] }`. Since `addAllowlist` is now reachable via `update` (as you noted), this closes the last read-and-write-unconditionally surface. Added a regression test (a comment, a foreign `mcpAllowlist` rule, and a foreign top-level key all survive `addAllowlist`) plus an install->uninstall round-trip; both fail on the pre-fix code.

packages/argent-installer/src/mcp-configs.ts
latekvo: Fixed in 837c44e7. `addClaudePermission` (and `removeClaudePermission`) now go through `readJsonc` / `editJsoncFile`, so a comment in `.claude/settings.json` no longer makes the read return `{}` and drop the user's other permissions on write. `editJsoncFile` targets `permissions.allow`, appending the rule while preserving comments and foreign keys. Added a regression test (comment + a foreign `Bash(ls:*)` permission + a foreign top-level key survive `addClaudePermission`) plus an install->uninstall round-trip; both fail on the pre-fix code. With these two migrated, every JSON write now runs through `editJsoncFile` and `writeJsonOrRemove` is removed.

packages/argent-installer/src/utils.ts
latekvo: Fixed in 837c44e7. Added a test that seeds an existing `.vscode/mcp.json` *without* a trailing newline, calls `write()`, and asserts the file stays newline-free (with argent and the foreign server both present). It pins the `!existedBefore` branch: making the append unconditional would flip it to fail.

packages/argent-installer/test/mcp-configs.test.ts
latekvo: Fixed in 837c44e7. Both are now pinned directly:

1. A test writes an entry carrying `env`, re-writes the env-less `getMcpEntry()`, and asserts the stale `env` key is gone, proving the write is a replace and not a merge (the `update` upgrade path over a pre-#238 entry depends on this).
2. A test removes argent when it is the only server beside a top-level empty sibling (`{ mcpServers: { argent }, userThing: {} }`) and asserts the emptied `mcpServers` is pruned, the empty top-level sibling survives, and the file is kept, exercising the ancestor-prune-then-keep-file path the existing Cursor case didn't reach.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -9,9 +9,7 @@ import {
   CURSOR_ALLOWLIST_PATTERN,
 } from "./constants.js";
 import {
-  readJson,
```
hubgan, line 13: Removing these imports leaves `readJson`/`writeJson` (utils.ts) with no remaining call site in the installer source — every adapter and allowlist helper now routes through `readJsonc`/`editJsoncFile`. They're effectively dead exports now (only their definitions and doc-comments remain). Relatedly, the `editJsoncFile` JSDoc still steers pure-JSON callers to "go through `writeJson` instead," which no longer matches how the strict-JSON adapters (Claude, Windsurf, Gemini) actually use it.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -163,57 +160,62 @@ const cursorAdapter: McpConfigAdapter = {
...
     const servers = config.mcpServers as Record<string, unknown> | undefined;
     return Boolean(servers?.[MCP_SERVER_KEY]);
   },
 
+  // Cursor's allowlist lives in a separate ~/.cursor/permissions.json, which —
+  // like every Cursor config — is JSONC (comments, trailing commas) and may
+  // carry the user's own unrelated rules. Route through readJsonc / editJsoncFile
+  // so a hand-commented or multi-rule permissions.json survives: the old readJson
+  // path ran it through `catch { return {} }` and rewrote the whole file with
+  // only { mcpAllowlist }, dropping every foreign rule and comment. Newly matters
```
hubgan, line 350: This is the behavior transition the fix is built around, and it isn't pinned by a test. Because `hasArgentEntry` now parses a commented config as JSONC, `update` / `findConfiguredAdapterScopes` will treat a config that previously parsed to `{}` (and was skipped) as configured, and then call `addAllowlist`. Every `hasArgentEntry` / `findConfiguredAdapterScopes` fixture uses strict `JSON.stringify` content, so a regression of exactly this commented-config-is-now-detected behavior would pass CI.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -99,47 +97,46 @@ function removeDirIfEmpty(dirPath: string): void {
...
+// After the Codex remover deletes the argent entry, collapse the now-empty
+// `mcp_servers` table it lived in so a config that held only argent reduces to {}
+// and writeTomlOrRemove can delete the file. Only this one key is touched —
+// foreign sibling keys and empty values elsewhere in the tree are preserved
+// byte-for-byte (the contract is "only touch argent"). The JSON adapters get the
+// same ancestor collapse for free from editJsoncFile.
+function deleteIfEmpty(parent: Record<string, unknown>, key: string): void {
+  const value = parent[key];
+  if (
+    (Array.isArray(value) && value.length === 0) ||
```
hubgan: This `Array.isArray(value) && value.length === 0` branch is unreachable. `deleteIfEmpty`'s only caller is `deleteIfEmpty(config, "mcp_servers")` (line 614), where `mcp_servers` is always a TOML table (object), never an array — so the array case never runs.

packages/argent-installer/src/utils.ts
```diff
@@ -317,7 +320,14 @@ export function editJsoncFile(filePath: string, jsonPath: JSONPath, value: unkno
...
   fs.mkdirSync(path.dirname(filePath), { recursive: true });
-  fs.writeFileSync(filePath, (hadBom ? "﻿" : "") + text);
+  // A synthesized document — a fresh file, or an existing file that was empty /
+  // whitespace-only and so had no real content to preserve — gets a trailing
+  // newline, matching writeJson/TOML (POSIX text-file convention, avoids a git
+  // "\ No newline at end of file"). An existing file with real content keeps its
+  // own EOL/formatting untouched — setJsoncIn edits in place — so we never
+  // rewrite its trailing byte.
+  const out = wasEmpty && !text.endsWith("\n") ? text + "\n" : text;
+  fs.writeFileSync(filePath, (hadBom ? "﻿" : "") + out);
```
hubgan, line 362: This write isn't atomic (no temp-file + rename). If the process is killed or the disk fills mid-write, the user's config is left truncated — the same "foreign config lost" outcome this change is meant to prevent, reached by a different route. It's pre-existing (the old `writeJson` had the same shape and this PR doesn't widen it), but worth flagging given the PR's explicit goal of never clobbering foreign config.

hubgan (commented):
A few inline notes below: one test-coverage gap on the update/`hasArgentEntry` path plus a couple of cleanups. No blocking correctness issues — the foreign-server and JSONC/comment preservation itself holds up.

hubgan (approved):

packages/argent-installer/src/mcp-configs.ts
latekvo: After merging `main` (which landed #428, the committable `--local` install mode), `readJson`/`writeJson` are no longer dead — #428's Claude shadowing-config code gave them real call sites: `readJson` reads `~/.claude.json` project scopes and `disabledMcpjsonServers` in `findShadowingConfigs`/`claudeDisabledListFinding`, and `writeJson` writes those back plus the global `~/.claude.json` remove (which must never be deleted or deep-pruned). So both keep call sites and stay exported.

The stale JSDoc is fixed in 4671b869 (+ 8bde13e9 for the section banner): `editJsoncFile`/`readJsonc` no longer reference the removed `writeJsonOrRemove` or read as Zed-only. `editJsoncFile` is now documented as the shared write path for every MCP-config adapter (strict-JSON ones included, since JSONC is a superset), with `writeJson` scoped to whole-document rewrites that must never delete the file (`~/.claude.json`).

packages/argent-installer/src/mcp-configs.ts
latekvo: Fixed in 4671b869. Added a `findConfiguredAdapterScopes` test that seeds a Cursor config carrying a line comment AND a trailing comma (invalid strict JSON) alongside the argent entry, and asserts both `hasArgentEntry` and `findConfiguredAdapterScopes` detect it as configured. It pins exactly the transition you flagged — I verified it FAILS when `getArgentEntry` is reverted to strict `readJson` (the commented file parses to `{}`, so `hasArgentEntry` returns false) and passes on the `readJsonc` path. The test also self-guards with an `expect(() => JSON.parse(...)).toThrow()` on the fixture so it can't silently become valid JSON.

packages/argent-installer/src/mcp-configs.ts
latekvo: Fixed in 81a00049 (as part of the main merge resolution). Dropped the `Array.isArray(value) && value.length === 0` branch — the sole caller is `deleteIfEmpty(config, "mcp_servers")`, and the preceding `if (!servers?.[MCP_SERVER_KEY]) return false` guard means `mcp_servers` is provably a table (an array has no `"argent"` key), so only the empty-object case can fire. `deleteIfEmpty` is now object-only, with a comment noting why.

packages/argent-installer/src/utils.ts
latekvo: Confirmed real, but leaving it out of scope here. The non-atomic write is pre-existing and shared identically by all four writers — `writeJson`, `writeToml`, `writeYaml`, and `editJsoncFile` each do a bare `fs.writeFileSync`; this PR doesn't widen it. A proper fix is a cross-cutting durability change (temp-file + rename across every writer, with the Windows rename-over-existing-target handling atomic replace needs) and is orthogonal to the logical clobbering this PR targets (deep-prune and `readJson`→`{}`). Better as a dedicated follow-up than a half-measure bolted onto one of the four writers.

PR #443, opened 2026-07-01
feat(file-inputs): add tar-upload kind for streaming remote file/directory uploads

gwdp:
## What

A new `tar-upload` file-input kind so a tool that reads a local file or directory (e.g. `reinstall-app`'s app bundle) works when the tool-server runs on a different machine than the caller.

## Why

`reinstall-app` assumes the bundle is on the tool-server's filesystem. That breaks with a cloud-hosted simulator — the bundle is on your machine, not the server's. The existing `file` kind inlines base64 and is capped at 32 MB, too small for a real `.app`/`.apk`; `tar-upload` streams instead.

## How

When you're not linked to a remote server, an existing local path is used in place.

When you are (e.g. `argent link`), the client tars the path, streams it to `POST /upload`, and the tool call carries `uploadId` plus a SHA-256 `contentHash`. The server always extracts from that upload — even if the same path already exists on disk — verifies the hash, and vets tar members before extracting.

The client only uploads when the path is on its own machine (a line on stderr so you know it's working). Same command local or remote:

```
argent run reinstall-app --udid <udid> --bundleId <id> --appPath ./MyApp.app
```

## Shared extraction

Per review feedback, the tar pack/unpack logic is unified in a small `@argent/archive` package used by **both** directions — this upload path and the pre-existing artifact download (`streamDirectoryAsTarGz` / `downloadAndExtractArchive`). The shared extractor rejects path-traversal entries, symlinks whose target escapes the extract dir, and non-file/dir members (hardlinks, devices) — while still allowing a bundle's internal symlinks. This also hardens the artifact-download path, which previously extracted without those checks.

## Tests

Client + server unit and HTTP round-trip tests: dir and file upload, skip-when-not-local, co-located no-op, hash required, path-traversal / escaping-symlink / hardlink rejection, sidecar handling, single-member resolution, `/upload` auth + size caps, and artifact download extraction. All suites pass.


packages/tool-server/src/file-inputs.ts
```diff
@@ -139,6 +153,37 @@ async function resolveOne(
...
+    const entry = lookupUpload?.(wire.uploadId);
+    if (!entry) {
+      throw new FileInputError(
+        `Upload "${wire.uploadId}" was not found on the tool-server — it may have expired. ` +
+          `Re-run the tool to upload the path again.`
+      );
+    }
+    const extractDir = await mkdtemp(join(tmpdir(), "argent-tar-upload-"));
+    tempDirs.push(extractDir);
+    await execFileAsync("tar", ["-xzf", entry.tarPath, "-C", extractDir]);
```
szdziedzic: Could we make the uploaded tar cleanup unconditional here? `/tools/:name` deletes the upload entry from the `uploads` map before calling `resolveFileInputs`, so if `tar -xzf` fails on a corrupt/truncated archive, the sweeper no longer owns `entry.tarPath` and this file can leak until process exit. Since uploads can be large, I think this should be a `try/finally` around extraction with `rm(entry.tarPath, { force: true })` in the finally, plus a test for an invalid tar upload.

packages/tool-server/src/http.ts
```diff
@@ -277,13 +277,36 @@ export function isWebsocketUpgradeAllowed(
...
   // request; the decoded per-file ceiling is enforced in file-inputs.ts.
   app.use(express.json({ limit: "48mb" }));
 
+  const maxUploadBytes = options?.maxUploadBytes ?? MAX_UPLOAD_STREAM_BYTES;
+
+  // Pending tar-upload archives, keyed by uploadId. Consumed by the first tool
+  // call that references them; the TTL sweeper clears orphans from aborted or
+  // failed calls.
+  const uploads = new Map<string, UploadEntry & { expireAt: number }>();
+  const uploadSweeper = setInterval(() => {
```
szdziedzic, line 360: Since this introduces a new app-lifetime resource, can we tear it down from `createHttpApp().dispose()` too? `uploadSweeper.unref()` means it won't keep Node alive, but disposed app instances can still leave the timer, the `uploads` map, and unconsumed temp tar files around until TTL/process exit. `dispose()` should probably clear the interval and remove pending upload tar files.

packages/tool-server/src/http.ts
```diff
@@ -277,13 +277,36 @@ export function isWebsocketUpgradeAllowed(
   return true;
 }
 
+// The tool call that consumes an upload arrives right after it (same client
+// invocation), so the TTL only has to outlive that gap — generously.
+const UPLOAD_TTL_MS = 15 * 60 * 1000; // 15 minutes
+// Bounds the tar-upload so a bad client can't fill the host disk.
+const MAX_UPLOAD_STREAM_BYTES = 2 * 1024 * 1024 * 1024; // 2 GiB
```
szdziedzic, line 336: This bounds each individual upload, but there does not seem to be a total cap on pending upload storage. A client can POST multiple uploads under the 2 GiB per-upload limit and never consume the returned `uploadId`s, leaving all of them on disk until the 15-minute TTL sweeper runs. Could we track pending upload bytes/count and reject new uploads once the store is over a configured cap?

packages/tool-server/src/http.ts
gwdp: Totally, I set up the sweeper but never made it die with the app. `dispose()` now kills the interval, removes the pending tars, and clears the map. Thank you for catching that 🙏

packages/tool-server/src/file-inputs.ts
gwdp: Oh yeah, 100%, that's a real leak. The entry's already out of the map by the time extraction runs, so if the tar's corrupt, nothing was reclaiming the space. Updated, thank you!

packages/tool-server/src/http.ts
gwdp: Ha, excellent catch, per-upload alone doesn't stop someone stacking a bunch of them; I was only thinking about the single op there (too narrow of a vision). Added a `maxPendingUploadBytes` (8gb default) that 507s new uploads while the store's over the cap

filip131311 (commented):
This change makes sense I don't know why I didn't implement compressing large files from client to server but I think you are duplicating a lot of logic that already exists, previously we had `downloadAndExtractArchive` utility on the client side and `streamDirectoryAsTarGz` on the server side, I think your implementation might even have advantages elements but i feel that we should unify the server -> client and cilent -> server approaches before merging into some kind of core package. What do you think?

gwdp:
> This change makes sense I don't know why I didn't implement compressing large files from client to server but I think you are duplicating a lot of logic that already exists, previously we had `downloadAndExtractArchive` utility on the client side and `streamDirectoryAsTarGz` on the server side, I think your implementation might even have advantages elements but i feel that we should unify the server -> client and cilent -> server approaches before merging into some kind of core package. What do you think?

Good call, appreciate the suggestion! Pulled the tar pack/unpack into a small `@argent/archive` package, so the artifact side (`streamDirectoryAsTarGz / downloadAndExtractArchive`) and the new upload path share one set of helpers now instead of two drifting copies.

Kept it tight, just the sharing helpers, not a reshape of the endpoints/handles to avoid changing more than needed on this PR. 
Nice side effect though: the shared extractor does the path-slip + symlink/hardlink checks, so the download path picked up hardening it lacked before (was a plain tar -xzf).

Let me know how you feel about the direction, happy to continue iterating if opposed thoughts 🙇‍♂️

filip131311:
The new approach seems good, but there are some fails on the CI please take a look at them. 

Also an agent found a problem with one of the paths, I don't think it is very dengerous as uploads leave in temp anyway, but it's still worth patching: 

```
Leaked upload tar on the missing-hash / hash-mismatch paths.

The finally that removes entry.tarPath only covers the extraction try block — but the !wire.contentHash and entry.sha256 !== wire.contentHash throws exit the function before reaching it. By that point the HTTP layer's lookup callback has already done uploads.delete(id), so the TTL sweeper and dispose() can no longer see the file either. Nothing else references it: it isn't in tempDirs (only the extract dir goes there, and it isn't created yet), so resolveFileInputs' catch-side cleanup() is a no-op here.

Result: the tarball is stranded in tmpdir forever, and it also silently leaves the maxPendingUploadBytes accounting. A client with a broken hash would hit this on every call — and the error message tells it to re-upload, stranding another copy each retry (up to 2 GiB each) until the disk fills.

Fix: start the try/finally right after the successful lookup so it covers all three failure paths, e.g.:

const entry = lookupUpload?.(uploadId);
if (!entry) throw new FileInputError(/* not found */);
try {
  if (!wire.contentHash) throw new FileInputError(/* missing hash */);
  if (entry.sha256 !== wire.contentHash) throw new FileInputError(/* mismatch */);
  const extractDir = await mkdtemp(join(tmpdir(), `argent-tar-upload-${entry.sha256.slice(0, 16)}-`));
  tempDirs.push(extractDir);
  const uploaded = await safeExtractTarGz(entry.tarPath, extractDir, basename(wire.path));
  return { value: uploaded, meta: { ...meta, viaUpload: true } };
} catch (err) {
  if (err instanceof FileInputError) throw err;
  throw new FileInputError(`Could not extract the uploaded archive for "${wire.path}": ...`);
} finally {
  await rm(entry.tarPath, { force: true }).catch(() => {});
}
```

gwdp:
> The new approach seems good, but there are some fails on the CI please take a look at them.
> 
> Also an agent found a problem with one of the paths, I don't think it is very dengerous as uploads leave in temp anyway, but it's still worth patching:
> 
> ```
> Leaked upload tar on the missing-hash / hash-mismatch paths.
> 
> The finally that removes entry.tarPath only covers the extraction try block — but the !wire.contentHash and entry.sha256 !== wire.contentHash throws exit the function before reaching it. By that point the HTTP layer's lookup callback has already done uploads.delete(id), so the TTL sweeper and dispose() can no longer see the file either. Nothing else references it: it isn't in tempDirs (only the extract dir goes there, and it isn't created yet), so resolveFileInputs' catch-side cleanup() is a no-op here.
> 
> Result: the tarball is stranded in tmpdir forever, and it also silently leaves the maxPendingUploadBytes accounting. A client with a broken hash would hit this on every call — and the error message tells it to re-upload, stranding another copy each retry (up to 2 GiB each) until the disk fills.
> 
> Fix: start the try/finally right after the successful lookup so it covers all three failure paths, e.g.:
> 
> const entry = lookupUpload?.(uploadId);
> if (!entry) throw new FileInputError(/* not found */);
> try {
>   if (!wire.contentHash) throw new FileInputError(/* missing hash */);
>   if (entry.sha256 !== wire.contentHash) throw new FileInputError(/* mismatch */);
>   const extractDir = await mkdtemp(join(tmpdir(), `argent-tar-upload-${entry.sha256.slice(0, 16)}-`));
>   tempDirs.push(extractDir);
>   const uploaded = await safeExtractTarGz(entry.tarPath, extractDir, basename(wire.path));
>   return { value: uploaded, meta: { ...meta, viaUpload: true } };
> } catch (err) {
>   if (err instanceof FileInputError) throw err;
>   throw new FileInputError(`Could not extract the uploaded archive for "${wire.path}": ...`);
> } finally {
>   await rm(entry.tarPath, { force: true }).catch(() => {});
> }
> ```

Thanks for flagging both!

Indeed, the agent caught a real edge case 😅  patched the leak as suggested.
CI's sorted too: a prettier miss and a tar-slip test that only worked on BSD tar (not the linux runners)

filip131311 (approved):
thank you i pushed a simple package.lock update and will merge it now, this should land on Wednesday

PR #444, opened 2026-07-01
feat(telemetry): derive distinct_id from a stable host fingerprint

hubgan:
## Summary

Replaces the random-UUID telemetry `distinct_id` with a stable per-machine id derived from the host fingerprint, so PostHog's native "Unique users" counts machines (funnels / retention / breakdowns work with no HogQL). Every Argent entry point - installer, CLI, tool-server, MCP - derives the id from one shared code path, so a given machine reports a single identity across every event and across install / uninstall / reinstall.

## What changed

- `distinct_id = uuidv5(hostFingerprint, NAMESPACE)`, where the fingerprint comes from `simulator-server fingerprint`. Deterministic, keeps the UUID shape, and never stores the raw hardware value in PostHog (only the SHA-1-derived UUID leaves the process).
- The fingerprint is resolved in `buildPayload` - the single `capture()` path shared by every entry point - so installer / CLI / tool-server / MCP all produce the same id. Resolved lazily, once per process, memoized.
- One-time, local-only migration: the first time a fingerprint resolves, `~/.argent/telemetry-id` is rewritten from the legacy random UUID to the derived id. No PostHog `alias` / `$identify` event is emitted.
- Fallback: if the fingerprint can't be resolved (unsupported platform, binary absent), the stored or a fresh random UUID is used - identically for every entry point on that machine, self-healing to the derived id on the next successful resolve.
- Hardening: never throws; the fingerprint is validated (exactly 64 hex, case-insensitive) before hashing; corrupt / empty / unreadable id files self-heal; the binary is spawned at most once per process.

## Why the id is stable across install -> uninstall -> reinstall

Uninstall deletes `~/.argent/telemetry-id`. Because the id is `uuidv5(fingerprint)` and the fingerprint is deterministic per machine, a fresh install recomputes the identical id. Previously only the tool-server resolved the fingerprint while the installer emitted under a random UUID, so this cycle produced a different random id each time.

## Bundling note for reviewers

`@argent/telemetry` now depends on `@argent/native-devtools-ios` to locate the binary. That package computes its bin dir from `__dirname` at module load, which is undefined in an ES module. The installer / CLI / MCP ship as ESM bundles, so `bundle-tools.cjs`'s ESM banner is extended to shim `__dirname` / `__filename` (the CJS tool-server bundle already has them natively). Without this shim those bundles would crash on load, so the banner change is load-bearing.

## Testing

- Unit: 219 telemetry tests - RFC 4122 v5 vectors, migration, idempotency, case/length validation, corrupt-file self-heal, spawn-at-most-once, and all fallbacks.
- E2E on a real machine + booted iPhone 16 Pro: installer, CLI, and tool-server runtimes all emit the same derived id; the id is stable across a delete/recreate (uninstall/reinstall) cycle; a legacy v4 id migrates locally with no alias/identify emitted; verified from an ESM bundle placed in the real install's `dist/` so the binary resolves via the bundled `__dirname` with no env override; tool calls against the device carry the same id.


latekvo (commented):
Review requested in person. A few notes on the fingerprint-derived `distinct_id` path, left inline.

packages/telemetry/src/fingerprint.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Used to derive a stable per-machine `distinct_id`. Best-effort: returns null
+ * — never throws — when the binary is absent or the command fails, so identity
+ * falls back to a random id. Resolved once per process, lazily, on the first
+ * tracked event, for every Argent entry point (installer / CLI / tool-server /
+ * MCP) — so the id is stable across uninstall + reinstall, not only for the
+ * tool-server.
+ */
+export function resolveHostFingerprint(): string | null {
+  try {
+    const out = execFileSync(simulatorServerBinaryPath(), ["fingerprint"], {
```
latekvo, line 42: `resolveHostFingerprint` runs `simulator-server fingerprint` with a synchronous `execFileSync`, so it blocks the Node event loop for the whole spawn. It is reached from `buildPayload()` inside `track()`, and callers invoke `track()` synchronously: `toolserver:start` is tracked inside the tool-server's `listen()` callback right after the socket binds, and the installer/CLI track on their first event. On `main` this path was pure file I/O.

So the first tracked event of each process blocks the loop for the duration of the spawn — I measured ~230-310ms warm (the comment above estimates <100ms), bounded by the 5s timeout that, as the comment notes, a freshly-installed binary under Gatekeeper can approach. In the tool-server the parent process keys off the "listening" line, so a request sent on readiness sits unprocessed in the accept backlog for that window.

Because the fast path only skips the spawn once a 64-hex id is persisted, a binary that is present but never emits a valid 64-hex value (empty or non-hex output) causes a random UUID to be stored instead, and every subsequent process then re-spawns and blocks again rather than paying this cost only once.

packages/telemetry/src/identity.ts
```diff
@@ -3,27 +3,190 @@ import * as fs from "node:fs";
...
+  try {
+    raw = resolveFingerprint();
+  } catch {
+    raw = null;
+  }
+  if (raw) {
+    const normalized = raw.trim().toLowerCase();
+    if (FINGERPRINT_PATTERN.test(normalized)) value = normalized;
+  }
+  fingerprintResolved = { value };
```
latekvo, line 343: `resolveFingerprintOnce` memoizes a `null` result here for the entire process, and `readOrCreateAnonId` returns the module-level `cached` id at the top (line 51). If a long-lived process — the tool-server, which emits on every tracked event — hits a transient resolve failure on its first event (the cold-binary timeout described in fingerprint.ts, or any spawn error), it caches the random/legacy fallback id and memoizes `null`, then never re-reads disk or re-spawns for the rest of its lifetime.

When a later short-lived process resolves the real fingerprint and migrates the on-disk id, this process keeps emitting the divergent fallback id, so the machine reports under two `distinct_id`s at once: the high-volume tool-server stream under the fallback and everything else under the fingerprint. That contradicts the "unique users == unique machines" goal in the docstring above — the "until a later run migrates" note only rescues short-lived processes, not a long-lived one that has already memoized `null`.

Reproduced: a first call with a throwing resolver returns a v4 id; writing the fingerprint to the id file out-of-band; a second call with a working resolver still returns the v4 while the on-disk id is the fingerprint.

packages/telemetry/src/identity.ts
```diff
@@ -3,27 +3,190 @@ import * as fs from "node:fs";
...
+// process even if id-file persistence keeps failing (a broken ~/.argent would
+// otherwise re-enter readOrCreateAnonId — and re-spawn — on every event).
+let fingerprintResolved: { value: string | null } | null = null;
+
+/**
+ * Resolve the anonymous telemetry id.
+ *
+ * When a host fingerprint is available, the id IS that fingerprint — the 64-hex
+ * one-way hash emitted by `simulator-server fingerprint`, used verbatim so
+ * PostHog's native unique users == unique machines, stable across reinstalls.
```
latekvo, line 52: Using the fingerprint verbatim makes the `distinct_id` a stable, deterministic, hardware-derived value that persists across uninstall/reinstall by design — a persistent identifier that singles out and re-identifies the same machine over time, which the previous per-install random UUID did not.

Two descriptions elsewhere no longer match this new nature: the first-run consent notice shown to users (`FIRST_RUN_NOTICE_BODY_LINES` in `notice.ts`, rendered in the installer TUI and written to stderr by the MCP server) still describes the collected data as "anonymous"; and the identity-file JSDoc at `paths.ts:19` still describes the file as "UUID v4", which is now only the fallback shape rather than the steady-state contents.

packages/telemetry/src/fingerprint.ts
```diff
@@ -0,0 +1,137 @@
...
+ * (no id persisted yet), so that first event already carries the stable id
+ * rather than a random fallback that later migrates. Every other resolution —
+ * upgrading a fallback id, re-probing after a transient failure, the
+ * tool-server's off-the-accept-path warm-up — uses the async variant below so
+ * it never stalls the loop. Best-effort: returns null (never throws) when the
+ * binary is absent or the command fails.
+ */
+export function resolveHostFingerprint(): string | null {
+  try {
+    const out = execFileSync(simulatorServerBinaryPath(), ["fingerprint"], {
```
latekvo, line 42: This synchronous resolve is reached from `track()` -> `buildPayload` -> `readOrCreateAnonId(resolveHostFingerprint)` on the truly-fresh path (a fresh install's first event, before any id is on disk) for the short-lived entry points that don't warm the identity - installer / CLI / MCP. `execFileSync`'s `timeout` only sends SIGTERM once; it does not escalate to SIGKILL. So a `fingerprint` process that ignores/handles SIGTERM and stays alive blocks the event loop past the 5s cap indefinitely, freezing the whole command. The async sibling `resolveHostFingerprintAsync` is hardened against exactly this - its own comment notes "spawn's timeout (like execFile's) only SENDS SIGTERM once and never escalates ... would leave the promise pending forever" - with an independent watchdog + SIGKILL; the sync path has no equivalent bound. Reproduced: a fresh-HOME `track()` pointed at a SIGTERM-ignoring `simulator-server` never returns (still blocked well past the 5s timeout; killed at 12s), while the real binary returns in ~270ms. On `main` this path was pure file I/O and could not hang.

packages/telemetry/src/notice.ts
```diff
@@ -20,8 +20,9 @@ export const TELEMETRY_DETAILS_URL = "https://swmansion.com/legal/argent/privacy
  * match its own conventions (e.g. cyan in the installer TUI).
  */
 export const FIRST_RUN_NOTICE_BODY_LINES: readonly string[] = [
-  "Argent collects anonymous usage data to help us improve the tool. We never",
-  "collect your source code, file paths, tool inputs, or error contents.",
+  "Argent collects anonymous usage data, linked to a per-device identifier, to",
```
latekvo: This line was updated to call the id a per-device identifier, but the privacy notice it links to (via `TELEMETRY_DETAILS_URL`, line 15 above - repo source `Telemetry.md`) still describes the same id the opposite way. After this change the `distinct_id` is the 64-hex host fingerprint used verbatim: a deterministic one-way hash of stable hardware identifiers, stable across uninstall/reinstall (identity.ts:6-8, 48-49; paths.ts:11-14). `Telemetry.md:63` still tells users it is "a randomly generated identifier that persists for an installation ... [that] does not contain, and is not derived from ... any device identifier" - both "randomly generated" and "not derived from ... any device identifier" are now inaccurate. (Its lines 7 and 74, "no device identifiers / hardware IDs collected", are more arguable, since the value is a one-way hash rather than the raw id.) `README.md:135` and the `argent telemetry status` label also still call the id "anonymous." The authoritative notice (Version 1.0) is not updated in this PR to match what is now collected and how long it persists.

packages/telemetry/src/identity.ts
```diff
@@ -3,27 +3,345 @@ import * as fs from "node:fs";
...
+  );
+  const fd = fs.openSync(tmpPath, "wx", 0o600);
+  try {
+    try {
+      fs.writeSync(fd, id);
+      fs.fsyncSync(fd);
+    } finally {
+      fs.closeSync(fd);
+    }
+    fs.renameSync(tmpPath, finalPath);
```
latekvo, line 400: This rename replaces whatever occupies the identity path - including a symlink - with a regular file, so `writeIdFileAtomic` does not apply the symlink-hardening the rest of the module does. `tryReadId` / `isCorruptIdFile` / `mintRandomId` all lstat-guard and treat a symlink as absent or leave it untouched, so the no-fingerprint mint path fails closed on a symlink and preserves it (the "symlink" test asserts this). But the fingerprint-resolving paths that reach here - the truly-fresh write in `readOrCreateAnonId` and `adoptFingerprint` - silently replace a symlink at `~/.argent/telemetry-id` with a regular file, so symlink handling diverges between the two write paths. (`rename` does not follow the link, so the pointed-to file is not written through - the effect is the inconsistency, not an overwrite of the target.)

packages/telemetry/src/index.ts
```diff
@@ -74,22 +80,54 @@ function activeRuntime(): Runtime {
...
+ *
+ * Resolves the fingerprint asynchronously and persists it (or a fallback) before
+ * the caller advertises readiness, so the first tracked event and all inbound
+ * requests find the id already on disk — never triggering a synchronous spawn in
+ * `track()`'s accept-path callback. Respects consent: a disabled machine mints
+ * no identity. Best-effort — never throws.
+ */
+export async function warmTelemetryIdentity(): Promise<void> {
+  try {
+    if (!isEnabled()) return;
```
latekvo, line 97: `warmTelemetryIdentity` gates only on consent here, then resolves the host fingerprint (spawning `simulator-server`) and persists an id via `warmIdentity`. `track()` / `buildPayload` instead check `getClient()` first and return before creating the id file when the PostHog key is unusable - with the comment "no reason to provision a persistent identifier on disk for an event that can never be transmitted," and a test asserting it. So on a machine that is consent-enabled but has an unusable key, warm-up still spawns the binary and writes a durable per-machine id that no event can use - the case `track()` avoids. Not reachable in the shipped build (the bundled token is usable), but reachable in the emergency-local / token-stripped builds that `resolveConfig()` anticipates ("" / "phc_disabled").

latekvo (commented):
A few notes inline from a closer pass over the fingerprint-derived `distinct_id` - the identity resolution path, its lifecycle/edge cases, and how the id is described to users.

packages/telemetry/src/identity.ts
```diff
@@ -43,26 +361,44 @@ export function readOrCreateAnonId(): string {
...
+      cached = { path: finalPath, id: value };
+      return value;
     } catch (err) {
       // openSync's EEXIST is handled above; the only EEXIST reaching here is
-      // from linkSync, i.e. another process published first.
+      // from linkSync, i.e. something already occupies the final path.
       if ((err as NodeJS.ErrnoException).code === "EEXIST") {
         const beatUs = tryReadId(finalPath);
         if (beatUs) {
+          // A racer published a VALID id first — adopt it, never clobber. This
```
latekvo, line 457: This adopt-the-winner branch - a racer publishes a valid id at the final path with no corrupt occupant, so `linkSync` throws EEXIST and we adopt it - is the concurrency guarantee, but nothing exercises it: disabling the `if (beatUs)` return leaves the whole identity suite green. The concurrency tests only reach the corrupt-occupant variant, where the racer's id is injected later during `claimCorruptOccupant`'s rename, so `beatUs` is null at this point. A regression that dropped this branch would ship undetected. (Confirmed with a repro where a racer writes a valid id just before this `linkSync`, with no corrupt file present: it passes on the current code and throws "failed to create anonymous identity after retries" once the branch is removed.)

packages/telemetry/src/identity.ts
```diff
@@ -3,27 +3,345 @@ import * as fs from "node:fs";
...
+let upgradeLastAttemptMs = 0;
+const UPGRADE_COOLDOWN_MS = 60_000;
+const UPGRADE_MAX_ATTEMPTS = 3;
+
+/**
+ * Resolve the anonymous telemetry id.
+ *
+ * When a host fingerprint is available, the id IS that fingerprint — the 64-hex
+ * one-way hash emitted by `simulator-server fingerprint`, used verbatim so
+ * PostHog's native unique users == unique machines, stable across reinstalls.
```
latekvo, line 52: Because the id is now deterministic and stable across reinstalls, deleting the identity file no longer yields a fresh identity: `deleteAnonId` (uninstall) and `forget()` remove the file, but the next tracked event re-derives the identical id and re-links to the machine's retained history (72 months per the privacy notice). This is the intended stability, but it changes what "forget" / "erase" mean for the user - `erasure.ts`'s "delete the anonymous id" framing and the `index.test.ts` "forget can erase telemetry identity" test read as a true erasure, whereas the id comes back on the next event. A genuine reset is now only reachable via opt-out (consent = false), not deletion; the erasure/uninstall wording and that test name no longer match the behavior.

latekvo (commented):
Two more, following up the earlier inline notes to make the pass complete - a concurrency-branch coverage gap and a forget/uninstall semantics point. (I left out one micro-perf nit.)

packages/telemetry/src/identity.ts
hubgan: This branch is already covered — the test is just in the older `identity.test.ts` suite rather than the new `identity-fingerprint.test.ts` one, which is likely why it looked uncovered.

`identity.test.ts > "does not overwrite an id published between temp write and final publish"` plants a valid id at the final path and throws `EEXIST` from `linkSync` with no corrupt occupant, which drives exactly this `beatUs` adopt-the-winner path. I mutation-tested it: disabling the `if (beatUs) return beatUs` makes that test fail with `telemetry: failed to create anonymous identity after retries` (suite goes 247 → 246). So a regression that dropped this branch would be caught.

packages/telemetry/src/identity.ts
hubgan: Fixed in a9afc0fa.

Clarified on `forget()` that deleting the id file is a local removal, not a permanent erasure: with consent enabled the deterministic fingerprint id re-derives identically on the next tracked event, so a genuine reset is the opt-out (the default `disableConsent`), not deletion. Also renamed the misleading "forget can erase telemetry identity" test and added an assertion that the id re-derives on the next event, so the test name and body now match the behavior.

Telemetry.md
```diff
@@ -1,10 +1,10 @@
 # Argent Privacy Notice (Telemetry)
 
-Effective date: 16 June 2026 · Version: 1.0
+Effective date: 2 July 2026 · Version: 1.1
```
latekvo: The in-repo notice is bumped to v1.1 here to describe the new hardware-derived identifier. The copy served at the public URL this notice is published to — the same `TELEMETRY_DETAILS_URL` the first-run consent notice, README, and installer link users to — is still Version 1.0 (16 June 2026) at review time. That live page still tells users the identifier is “a randomly generated identifier that persists for an installation … [that] does not contain, and is not derived from … any device identifier,” and that hardware IDs / MAC addresses / serial numbers are never collected. After this change the distinct_id is a one-way hash derived from stable hardware characteristics and stable across reinstall, so a user who follows the Details link reads a privacy description that directly contradicts what is now collected.

packages/telemetry/src/identity.ts
```diff
@@ -3,27 +3,366 @@ import * as fs from "node:fs";
...
+  if (upgradeInFlight) return;
+
+  const finalPath = identityFilePath();
+
+  // Already have the fingerprint (cached or persisted) → nothing to upgrade.
+  if (cached && cached.path === finalPath && FINGERPRINT_PATTERN.test(cached.id)) return;
+  const stored = tryReadId(finalPath);
+  if (stored && FINGERPRINT_PATTERN.test(stored)) return;
+
+  if (upgradeAttempts >= UPGRADE_MAX_ATTEMPTS) return;
```
latekvo, line 166: `upgradeAttempts` is process-global and only ever reset by the test-only `_resetIdentityCacheForTest`; neither `adoptFingerprint` nor a successful probe clears it. Once three background probes have failed (spaced past the 60s cooldown), this early return fires before the resolver is invoked again, so a resolver that only starts succeeding later is never retried by this process. For a long-lived tool-server that is the only Argent process on a machine — its startup `warmIdentity` resolve failed and its first three background probes failed before the binary became resolvable (e.g. a slow cold start) — it then emits under the random fallback id for the rest of its lifetime, converging to the stable fingerprint only if some other process happens to migrate the on-disk id. In the single-process case it never reaches the stable per-machine id the feature is meant to produce.

packages/telemetry/src/identity.ts
```diff
@@ -3,27 +3,366 @@ import * as fs from "node:fs";
...
+// chars — a one-way hash of stable hardware identifiers, so no raw hardware id
+// is ever exposed. That validated hash is used verbatim as the telemetry
+// distinct_id. Require the exact shape (after lower-casing) so a truncated read,
+// a bare hex token (git SHA, all-zeros), or an error banner is rejected and we
+// fall back to a random id rather than persisting garbage as a "stable" id. We
+// lower-case before matching so a future binary emitting upper-case hex still
+// maps to the same id — and it doubles as the self-describing marker for the
+// fast path below (a fingerprint id is 64-hex; the random fallback is a dashed
+// v4 UUID), letting us skip re-spawning the binary once one is on disk.
+const FINGERPRINT_PATTERN = /^[0-9a-f]{64}$/;
```
latekvo, line 19: The header comment above (line 10) states that a “bare hex token (git SHA, all-zeros) … is rejected … rather than persisting garbage as a stable id,” but `/^[0-9a-f]{64}$/` matches a 64-character all-zeros string (and any other low-entropy 64-hex constant); `normalizeFingerprint` then accepts it verbatim, persists and serves it as the fingerprint, and never re-probes. A 40-hex git SHA is excluded by the length, but the all-zeros case the comment explicitly calls out is not. If `simulator-server fingerprint` ever emitted a degenerate constant on some host (a machine with no readable hardware ids, or a zero-filled buffer on an error path), every affected machine would collapse onto the same distinct_id — the exact “garbage as a stable id” the comment intends to prevent. The shipped binary emits a proper hash today, so the practical exposure is small, but the comment and the pattern disagree about what is actually rejected.

Telemetry.md
```diff
@@ -60,7 +60,7 @@ We do not use telemetry for advertising, marketing, profiling, automated decisio
 
 ### Identifiers
 
-- a randomly generated identifier that persists for an installation, used only to distinguish unique installations and to de-duplicate events. It does not contain, and is not derived from, your name, username, account, or any device identifier;
+- a per-device identifier, used only to distinguish unique devices and to de-duplicate events. It is a one-way hash derived from stable characteristics of your machine, so it stays the same across reinstalls of Argent on the same device. We store only the resulting hash - it does not contain your name, username, or account, and we neither collect nor attempt to recover from it any raw device identifier (such as a serial number or MAC address). If the fingerprint cannot be computed, a randomly generated identifier is used instead;
```
latekvo: This section now correctly describes a persistent, cross-reinstall, hardware-derived identifier. Two downstream statements in the same notice still describe the previous resettable-per-install identifier and don’t appear to have been revisited: retention is stated as up to 72 months (line 89), and the Data Subject Rights paragraph says “we may be unable to identify your specific records … (Article 11 GDPR)” (line 97). A device-stable hash that is recomputable on the same machine and survives reinstall has a different re-identifiability profile than a random per-install token, so it seems worth confirming with whoever owns this notice that the retention window and the Article 11 characterization still hold. Flagging factually, not as legal advice.

packages/telemetry/src/identity.ts
```diff
@@ -3,27 +3,366 @@ import * as fs from "node:fs";
...
+ * safe (all write the same value). Best-effort persistence — an unwritable home
+ * still yields a consistent in-memory id, and a later run retries the write.
+ */
+function adoptFingerprint(fp: string): void {
   const finalPath = identityFilePath();
-  if (cached && cached.path === finalPath) return cached.id;
+  const stored = tryReadId(finalPath);
+  if (stored !== fp) {
+    try {
+      writeIdFileAtomic(finalPath, fp);
```
latekvo, line 313: `adoptFingerprint` is the one id-provisioning path that writes the identity file without a consent/client check — `warmTelemetryIdentity` and `track`/`buildPayload` both gate on `isEnabled()`/`getClient()` before provisioning an id, but a background upgrade scheduled by a prior `track()` reaches this write with no such gate and no coordination with a concurrent or preceding `forget()`. If an upgrade is in flight when the id file is deleted, this re-writes the durable fingerprint back to disk afterward. It does not manifest in the current uninstall flow — the only `forget()` caller (`uninstall.ts:564`) awaits `shutdown()` before deleting (so the upgrade settles first) and then exits promptly, abandoning the unref'd probe (I confirmed 0 resurrections across realistic runs; it only reappears when an artificial keepalive holds the loop open past the delete). So this is a latent gap rather than a live defect: a future `forget()` caller that lingers, or dropping the shutdown-before-delete ordering, would let an erased id reappear on disk, which is the erasure contract this PR documents on `forget()`.

latekvo (approved):
Reviewed the fingerprint-derived `distinct_id` path end to end at 0cb9efa2, focused on the identity state machine, the sync/async resolvers and their bounds, the tool-server warm-up, the ESM `__dirname` banner, and the user-facing wording. The earlier inline rounds are addressed, and the change holds up under exercise: driven against the real `simulator-server fingerprint` binary a fresh install, an uninstall/reinstall (delete-and-re-derive), and a legacy-id migration all produce the stable per-machine id; a 240-process concurrent stress of the id-file protocol converged to a single id every time; and the async/sync resolvers always settle and reap their child. The test suite is thorough and its load-bearing branches survive mutation.

The inline notes are all non-blocking. The one worth prioritising is the hosted privacy notice: the live page users are linked to still carries the pre-change wording. Approving.

PR #445, opened 2026-07-01
ci(vega-vvd-e2e): build host image on runner, drop prebuilt GHCR image

finloop:
## Summary

[`finloop/vega-virtual-device-action`](https://github.com/finloop/vega-virtual-device-action) shipped a new version that **builds the `vega-virtual-device-host` image on the consumer runner** from its bundled Dockerfile (cached via the GitHub Actions cache), like `reactivecircus/android-emulator-runner` sets up the Android SDK — instead of pulling a published GHCR image. This updates argent's Vega VVD E2E workflow to the new model.

## Changes (`.github/workflows/vega-vvd-e2e.yml`)

- **Re-pin the action** `fe151d9 → a0f1776` ([latest `main`](https://github.com/finloop/vega-virtual-device-action/commit/a0f17767abafb27ec9b58974d94643fc46a4a35d)). That commit's bundled Dockerfile carries the latest Vega SDK (`.sdk-version` = `0.23.8358`), so the image the action builds uses it.
- **Drop the prebuilt SDK image.** Removed the `image:`/`pull:` inputs and the pinned `HOST_IMAGE` GHCR digest (`ghcr.io/finloop/vega-virtual-device-host@sha256:2110746…`). With no `image:`, the action builds the host image locally with the pinned SDK. Also removed the now-unused `host-image` `workflow_dispatch` input.
- **Comment refresh** for the pinning rationale and `permissions` block — no registry image is pulled anymore.

## Notes

- First run on a fresh cache pays the full host-image build (Vega SDK install); subsequent runs restore the cached layers. `timeout-minutes: 60` and the `Free disk space` step already accommodate this.
- No secrets / registry auth needed — the image is built on the runner. `permissions: contents: read` is unchanged.
- The Kepler fixture download + checksum are untouched.

### DIFF
--- a/.github/workflows/vega-vvd-e2e.yml
+++ b/.github/workflows/vega-vvd-e2e.yml
@@ -15,11 +15,6 @@ name: Vega VVD E2E (remote + screenshot)
 
 on:
   workflow_dispatch:
-    inputs:
-      host-image:
-        description: "vega-virtual-device-host image to boot the VVD from (override the pinned digest for manual runs)"
-        required: false
-        default: "ghcr.io/finloop/vega-virtual-device-host@sha256:2110746fa9ba579635c6bdeac00c742749776fd135483d72f056cd1f2092d486"
   pull_request:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


latekvo (approved):
Approving — verified end-to-end, not just that it parses.

Confirmed against the passing E2E run on this branch (run 28514656119): the action was fetched at the pinned SHA a0f1776, the build-on-runner path actually executed (setup-buildx + build-push ran, no registry pull), the job ran with only `contents: read` / no secrets, and it reached `boot-status=ready` and produced `artifacts/vvd-ready.png` — the full boot → drive-remote → screenshot flow passed under `set -ex`.

Also checked:
- The action interface at a0f1776 matches this call: empty `image` builds the host image locally from the bundled Dockerfile, and `pull` is documented as ignored in that mode, so dropping `image:`/`pull:` is correct.
- SDK version is deterministic — `.sdk-version` = 0.23.8358 is fixed at the pinned action SHA.
- No dangling references to the removed `HOST_IMAGE` / `host-image` input / GHCR digest remain anywhere in the repo; the comment refresh lines up with the new model.

Clean drop of the prebuilt GHCR image. Good to go.

PR #455, opened 2026-07-01
fix(tool-server): inject Android keyboard/button over adb input (#449)

### LINKED ISSUE #449: keyboard/button silently succeed (no-op) on Android AVDs with hw.keyboard=no
## Summary
On Android AVDs configured with `hw.keyboard = no`, the `keyboard` and `button` tools **return a success payload while injecting nothing**. Text/keys/hardware-buttons never reach the app, but callers get `{ "typed": "...", "keys": N }` / `{ "pressed": "back" }` and no error — so an agent has no signal that input was silently dropped.

## Environment
- argent 0.13.0 (tool-server bundle `3152128c…`)
- Android emulator, AVD `argent-test-ci-repro` (`hw.keyboard = no`, `hw.mainKeys = no`, `hw.gpu.enabled = no`, 1.5 GB) — API 34 google_apis
- Host: Linux

## Reproduction
1. Boot an AVD whose `config.ini` has `hw.keyboard = no`.
2. Open any text field (e.g. Settings → search), focus it.
3. Call `keyboard { text: "battery" }` → returns `{ "typed": "battery", "keys": 7 }`, **field stays empty**.
4. Call `keyboard { key: "backspace" }` and `button { button: "back" }` → both return success, **nothing happens** on screen.
5. Flip the AVD to `hw.keyboard = yes` and reboot → the exact same calls now work.

Cross-check that isolates the tool (not the device): raw `adb -s <serial> shell input text "hello"` and `adb ... input keyevent 67/4` **do** land on the same focused field while argent's calls do not.

## Root cause
Android key/button injection goes through the emulator gRPC `sendKey` (hardware key events), which requires a guest keyboard device. With `hw.keyboard = no` there is no target, so the events are dropped. Touch injection (tap/swipe) is unaffected. The tool does not fall back to `adb shell input text` / `input keyevent`, and — more importantly — does not detect or report the drop.

## Impact
- Not universal: standard AVDs default `hw.keyboard = yes`, where both tools work.
- The real defect is the **silent false-success**: callers/agents believe input succeeded when it did not.
- Note: this reproduces on the `argent-test-ci-repro` fixture (`hw.keyboard = no`). If CI E2E AVDs share this setting, keyboard/button assertions there may be passing against no-ops — worth verifying.

## Suggested direction
Detect when key injection has no target (e.g. `hw.keyboard = no`) and either (a) fall back to `adb input text`/`input keyevent`, or (b) fail loudly instead of returning a success payload.

latekvo:
## What & why

`keyboard` and `button` drive Android through the simulator-server's USB-HID transport (`pressKey` / `cmd:"button"`). The guest only receives those HID events when the AVD exposes a hardware keyboard (`hw.keyboard = yes`) — the default, but CI / headless AVDs are routinely created with `hw.keyboard = no` (and `hw.mainKeys = no`), where the guest **silently drops** them. Because that transport is fire-and-forget, both tools returned `{ typed }` / `{ pressed }` **success while injecting nothing**.

This still affects current `main`: after the keyboard refactor, `makeAndroidImpl` routes Android **TV** to `adb input` but Android **phones/tablets** still to `typeSimulatorServer` (the HID path).

Fixes #449 (found during 0.13.0 RC E2E testing on a `hw.keyboard=no` AVD — `argent-test-ci-repro`, the config CI uses).

## Change

Route Android **phone/tablet** key / text / button injection through `adb shell input text` and `input keyevent` (new `utils/android-input.ts`), the same on-device transport Android TV and Vega already use:

- **`keyboard`** — `platforms/android.ts` now sends the non-TV branch to a `typeAndroidPhone` helper (adb `input`); the Android-TV and iOS branches are untouched.
- **`button`** — the Android branch presses via `injectAndroidKeycode`; iOS keeps the simulator-server HID path.

`adb input` lands regardless of `hw.keyboard`, on emulators (any config) and physical devices alike, and a non-zero exit surfaces as a **thrown error** instead of a silent no-op — directly addressing the fire-and-forget concern already called out in the `button` tool. Touch gestures are unchanged.

## Verification

- **Unit tests — 11/11 pass** (rebased on current `main`): new `keyboard-android.test.ts` (vocabulary parity vs `NAMED_KEYS`, `input text` shell-quoting incl. spaces, `input keyevent` mapping, empty-text no-op, newline rejection) + updated `button.test.ts` (Android `back` now asserts `KEYCODE_BACK` over adb).
- **eslint** on the changed files: clean.
- **Live tool-server** on a `hw.keyboard=no` AVD: the `keyboard` tool now issues `adb … shell input text '…'` (vs the RC's HID `pressKey`), and a failed inject **raises an error** rather than a fake success. The `adb input` path was independently observed typing/deleting on this exact AVD during RC testing (where the HID path silently no-op'd).
- Note: a full `tsc --build` / whole-suite run couldn't be completed in my environment (the `argent-private` submodule isn't checked out, which breaks the npm workspace install, and `main` added upstream deps my base lacked). `platforms/android.ts` mirrors the existing `platforms/ios.ts` branch structure one-for-one, and the pre-rebase revision of this change passed `tsc --build` + eslint. CI here is authoritative for the full typecheck.

## Reviewer notes

- `input text` receives the string as one shell-quoted argv token, so spaces are preserved verbatim; newlines are rejected (they'd truncate the `input` line) — callers use `key: "enter"`.
- `keys` counts by codepoint, matching the tv/vega/simulator-server backends.


packages/tool-server/src/tools/button/index.ts
```diff
@@ -71,6 +79,15 @@ Fails if the simulator-server / emulator backend is not reachable for the given
...
     }
+    if (device.platform === "android") {
+      // Android presses go over `adb shell input keyevent`, not the
+      // simulator-server's HID transport, which the guest silently drops on AVDs
+      // created with `hw.keyboard = no` / `hw.mainKeys = no`. adb lands
+      // regardless and surfaces a failure as a throw. The BUTTONS_BY_PLATFORM
+      // guard above guarantees a keycode exists for every accepted button.
+      await injectAndroidKeycode(params.udid, ANDROID_BUTTON_KEYCODES[params.button]!);
+      return { pressed: params.button };
+    }
```
hubgan, line 97: Android button now calls injectAndroidKeycode, but the tool has no Android adb dependency preflight. If adb is missing, this falls through resolveAdbOrThrow and the HTTP layer’s generic 500 path instead of the expected 424 dependency response. keyboard handles this correctly with requires: ["adb"].

packages/tool-server/src/utils/android-input.ts
```diff
@@ -0,0 +1,135 @@
...
+    if (cp < 0x20 || cp > 0x7e) {
+      const hex = cp.toString(16).toUpperCase().padStart(4, "0");
+      throw new Error(
+        `keyboard text can only contain printable ASCII on Android; character "${char}" ` +
+          `(U+${hex}) can't be typed via \`adb input text\` — emoji crash it and other ` +
+          `non-ASCII (accented, CJK) is silently dropped. Remove it.`
+      );
+    }
+  }
+}
```
hubgan, line 108: rejected Android text input throws a plain Error, so POST /tools/keyboard with café returned HTTP 500 in E2E. This is user input validation, not an internal server error. Map this to a 400-style tool/input error.

packages/tool-server/src/tools/button/index.ts
```diff
@@ -59,9 +60,16 @@ Returns { pressed: buttonName }.
...
   capability,
-  services: (params) => ({
-    simulatorServer: simulatorServerRef(resolveDevice(params.udid)),
-  }),
+  // Android presses go over `adb shell input keyevent` (see execute), not the
+  // simulator-server's HID transport, so declaring the service for an Android
+  // target would needlessly resolve + spawn a sim-server the tool never uses (up
+  // to a 30s ready-wait) and could throw ServiceInitializationError before the
+  // adb path even runs. Declare it only for the iOS / ios-remote path that
+  // actually consumes it (mirrors the sibling `keyboard` tool's lazy services).
```
hubgan, line 69: button docs still describe Android as “emulator backend” / simulator-server-style failure, but Android now uses adb shell input keyevent and also supports physical Android devices.

packages/tool-server/src/tools/button/index.ts
latekvo: Fixed in 4753be71. The Android button path now preflights adb via `ensureDep("adb")` before injecting, so a missing binary returns the same 424 install hint as the sibling `keyboard` tool. A tool-level `requires: ["adb"]` would also 424 the iOS path (which does not use adb), so the preflight lives in the Android branch of `execute`, using the `ensureDep` convenience documented for exactly this "classify device, then check the dep" pattern. Added a button unit test asserting the preflight fails closed (no keyevent injected when adb is missing) and an http-dep-gate case for the 424 mapping; verified end-to-end against an emulator.

packages/tool-server/src/utils/android-input.ts
latekvo: Fixed in 4753be71. Rejected Android text (a newline, or non-ASCII such as `café`) and unknown named keys now throw `InvalidToolInputError` — a new client-input error class the HTTP dispatcher maps to 400 (matching the zod-validation path), walking the cause chain so the `ToolExecutionError` wrap does not mask it. Added an http-dep-gate test pinning the 400 mapping and tightened the android-input tests to assert the error type. Verified end-to-end: `POST /tools/keyboard` with `café`, a newline, or `key:"nope"` on an emulator now returns 400 with the descriptive message instead of 500.

packages/tool-server/src/tools/button/index.ts
latekvo: Fixed in 4753be71 (building on 42a254f4). The "Fails if…" line now names the per-platform backend — the simulator-server for iOS, `adb` for Android — and the summary line reads "iOS simulator or device, Android emulator or device", so the physical-Android support the capability already declares is reflected in the description.

packages/tool-server/src/tools/keyboard/platforms/android.ts
```diff
@@ -1,27 +1,54 @@
...
+    await injectAndroidNamedKey(device.id, params.key);
+    keysPressed++;
+  }
+  if (params.text) {
+    await injectAndroidText(device.id, params.text);
+    // Count by codepoint (not UTF-16 units) to match the tv / simulator-server
+    // backends. `injectAndroidText` has already rejected any non-ASCII (so this
+    // equals `params.text.length` here), but keep the codepoint-safe count so the
+    // result stays correct if the printable-ASCII guard is ever relaxed.
+    keysPressed += [...params.text].length;
```
hubgan: `[...params.text].length` spreads to count codepoints, but the `text` branch reaches here only after `injectAndroidText` has already run `assertTypeableAndroidText`, which rejects every non-ASCII character. So at this point `params.text` is guaranteed printable ASCII and this is always equal to `params.text.length` — the codepoint distinction can never be observed. The accompanying comment acknowledges it only matters "if the guard is ever relaxed", so it's dead defensiveness today.

packages/tool-server/src/tools/keyboard/platforms/android.ts
```diff
@@ -1,27 +1,54 @@
...
+  }
+  return { typed: params.text ?? params.key ?? "", keys: keysPressed };
```
hubgan, line 39: When a caller passes neither `key` nor `text` (the schema leaves both optional with no refinement), this returns `{ typed: "", keys: 0 }` and issues zero adb calls — a silent success for an empty request. That contract isn't exercised by any test, so a future change to it would go unnoticed.

packages/tool-server/src/http.ts
```diff
@@ -717,6 +718,15 @@ export function createHttpApp(registry: Registry, options?: HttpAppOptions): Htt
           res.status(400).json({ error: unsupportedErr.message });
           return;
         }
+        // A tool rejecting its arguments (e.g. a newline in Android keyboard
+        // text, an unknown named key) is a client input error, not an internal
+        // fault — surface it as 400, matching the zod-validation path, instead
+        // of a misleading 500.
+        const invalidInputErr = findErrorInCauseChain(err, InvalidToolInputError);
```
hubgan, line 736: This `InvalidToolInputError` -> 400 check runs after `findDependencyMissing` (line 708), which scans the *entire* cause chain. It's correct only because the input-error throw sites construct `InvalidToolInputError` with no `.cause`. If a future throw site ever nested a `DependencyMissingError` as the cause of an `InvalidToolInputError`, the whole-chain scan would surface the dependency error first and map the response to 424 instead of 400 — an unstated invariant worth being aware of.

packages/tool-server/src/tools/button/index.ts
```diff
@@ -52,16 +54,23 @@ const capability: ToolCapability = {
 
 export const buttonTool: ToolDefinition<Params, Result> = {
   id: "button",
-  description: `Press a device hardware button (iOS simulator or Android emulator). Sends Down then Up events automatically.
+  description: `Press a device hardware button (iOS simulator or device, Android emulator or device). Sends Down then Up events automatically.
```
hubgan: "Sends Down then Up events automatically" now describes only the iOS branch. The Android path issues a single `adb input keyevent` (one atomic press, no separate Down/Up), so this line presents the iOS transport as if it were universal.

hubgan (commented):
Tested this end-to-end by building the branch and driving the `keyboard` and `button` tools over HTTP against a real Android emulator (Pixel 7) and an iOS simulator (iPhone 17 Pro):

- Android text / named-key / button injection lands via `adb input` and reads back verbatim on-device, including the `%s` no-corruption case (`100%safe`), preserved spaces, and shell-metacharacter safety (`&`, `|`, `$(...)`, backticks stay inert).
- The new rejection paths return the right HTTP status: newline / emoji / non-ASCII text and unknown key names each 400, unsupported buttons 400.
- iOS keyboard/button behave exactly as before (no regression from the shared refactor).

Unit tests (42) and eslint/prettier pass locally. Nice fix. A few minor notes inline.

packages/tool-server/test/button.test.ts
```diff
@@ -27,10 +45,29 @@ describe("button tool — per-platform validation", () => {
...
+      new DependencyMissingError(["adb"], "install android-platform-tools")
+    );
+    await expect(
+      buttonTool.execute(services, { udid: androidUdid, button: "home" })
+    ).rejects.toBeInstanceOf(DependencyMissingError);
+    // Preflight fails closed: no keyevent is injected when adb is missing.
+    expect(injectAndroidKeycode).not.toHaveBeenCalled();
   });
 
   it("accepts every iOS-valid button", async () => {
```
hubgan: This case asserts only the `{ pressed }` result; `sendCommand` is mocked but never inspected here or elsewhere in the file. Since this PR restructured `execute()` with the Android early-return, a regression that reordered or dropped the iOS `Down -> sleep -> Up` pair, or misrouted an iOS press into the adb branch, would still pass this test.

packages/tool-server/src/tools/keyboard/platforms/android.ts
latekvo: Fixed in d8517bf9. Dropped the spread — the count is now `keysPressed += params.text.length`. As you note, `assertTypeableAndroidText` guarantees the text here is printable ASCII (1 char = 1 codepoint = 1 UTF-16 unit), so `.length` *is* the codepoint count; the spread was defending a case the guard makes unreachable. Reworded the comment to state that plainly instead of the speculative "if the guard is ever relaxed". The routing tests still pin the count (`keys: 8` for `hi there`, `keys: 4` for key + `abc`), so the value stays covered.

packages/tool-server/src/tools/keyboard/platforms/android.ts
latekvo: Fixed in d8517bf9 — added a test pinning the empty-request no-op on the android impl (`{ typed: "", keys: 0 }`, zero adb calls).

On the contract itself: this silent no-op isn't android-specific. Every keyboard backend returns `{ typed: params.text ?? params.key ?? "", keys: 0 }` and issues nothing for an empty request — `simulator-server-keys.ts`, `platforms/tv.ts`, and `platforms/vega.ts` all do the same — so it's the shared cross-platform behaviour, not something this PR introduced. I deliberately did **not** add a zod refinement to reject empty input: the schema is shared across all platforms, so that would be a cross-platform contract change out of scope for this fix. The new test locks the current behaviour so any future change to it is a visible, deliberate edit rather than unnoticed drift.

packages/tool-server/src/http.ts
latekvo: Documented in d8517bf9. You're right that the 400 mapping is correct today only because every `InvalidToolInputError` throw site constructs it as a leaf (no `.cause`) — I checked the two sites in `android-input.ts` (`assertTypeableAndroidText` and `injectAndroidNamedKey`) and both are `new InvalidToolInputError(msg)` with no cause.

There's no reproducible bug today, and reordering isn't obviously right either (there's no principled precedence for an error that is genuinely *both* invalid-input and dependency-missing). So rather than change behaviour, I made the invariant explicit at the check site: a comment stating that this must stay after `findDependencyMissing`, and that it holds only while `InvalidToolInputError` is thrown causeless — keep it causeless, or reorder the two checks. That turns the unstated assumption into one a future throw site will actually see.

packages/tool-server/src/tools/button/index.ts
latekvo: Fixed in d8517bf9. The summary line now reads: "iOS sends a Down then Up event automatically; Android injects a single `adb` key event." The Down→Up phrasing is scoped to the iOS sim-server transport, and the single atomic `adb input keyevent` on Android is described accurately.

packages/tool-server/test/button.test.ts
latekvo: Fixed in d8517bf9. The iOS case now asserts the exact ordered `sendCommand` pair — `{ direction: "Down" }` then `{ direction: "Up" }` — and that `injectAndroidKeycode` is never called, so a dropped/reordered event or a misrouted iOS press fails instead of passing on `{ pressed }` alone.

Proved the gap was real before relying on the new assertion: dropping the `Up` `sendCommand` in `execute()` leaves the old result-only assertion green but turns the new one red (`AssertionError: expected [ { direction: "Down" } ] to deeply equal [ { direction: "Down" }, { direction: "Up" } ]`); restoring it goes green again.

packages/tool-server/src/utils/android-input.ts
```diff
@@ -0,0 +1,139 @@
...
+  await adbShell(serial, `input keyevent ${keycode}`, { timeoutMs: ADB_INPUT_TIMEOUT_MS });
+}
+
+/** Press a named key (keyboard tool `key` vocabulary) on Android. */
+export async function injectAndroidNamedKey(serial: string, name: string): Promise<void> {
+  const keycode = ANDROID_NAMED_KEYCODES[name.toLowerCase()];
+  if (keycode == null) {
+    // Unknown key name is a caller input error (HTTP 400), not a 500.
+    throw new InvalidToolInputError(
+      `Unknown key "${name}". Supported: ${Object.keys(ANDROID_NAMED_KEYCODES).join(", ")}`
```
hubgan: This routes an unknown key name to a 400 (and, via `assertTypeableAndroidText`, a newline or non-ASCII text as well). The other keyboard backends still throw a plain `Error` for the same class of user mistake and fall through to the 500 handler: `simulator-server-keys.ts` for an unknown key, and `vega-input.ts` for an unknown key and for a newline. Confirmed against a running server: `key: "pageup"` returns 400 on an Android target but 500 (with the internal `[Tool:keyboard]` prefix in the body) on an iOS target. So after this change the same well-typed-but-unusable input maps to different status codes depending on the platform.

packages/tool-server/src/tools/keyboard/platforms/android.ts
```diff
@@ -1,27 +1,54 @@
...
-    // either `adb input text` (TV) or the simulator-server (phone, adb-backed).
-    // Declare it so `dispatchByPlatform` preflights adb and a missing binary
-    // fails with the clean 424 install hint rather than surfacing from deeper in
-    // the probe. Matches the android branch of `describe` and `tv-remote`.
+    // `adb input` either way (TV via the focus daemon, phone via `input text` /
+    // `input keyevent`). Declare it so `dispatchByPlatform` preflights adb and a
+    // missing binary fails with the clean 424 install hint rather than surfacing
+    // from deeper in the probe. Matches the android branch of `describe` and
+    // `tv-remote`.
     requires: ["adb"],
```
hubgan, line 56: Nothing in the suite pins this `requires: ["adb"]` preflight. The keyboard-android tests drive `makeAndroidImpl().handler(...)` directly, which bypasses `dispatchByPlatform`'s preflight, so removing this line leaves every test green while an adb-less host would get a deep ENOENT instead of the clean 424 install hint. The sibling `button` tool has explicit coverage for its adb preflight; the keyboard path has none.

packages/tool-server/src/tools/button/index.ts
```diff
@@ -71,6 +80,21 @@ Fails if the simulator-server / emulator backend is not reachable for the given
...
+      // regardless and surfaces a failure as a throw. The BUTTONS_BY_PLATFORM
+      // guard above guarantees a keycode exists for every accepted button.
+      //
+      // Preflight adb here (the tool declares no global `requires` because the
+      // iOS path doesn't need it, and `services` skips the sim-server for
+      // Android) so a missing binary fails with the clean 424 install hint,
+      // mirroring the sibling `keyboard` tool's per-platform `requires: ["adb"]`.
+      await ensureDep("adb");
+      await injectAndroidKeycode(params.udid, ANDROID_BUTTON_KEYCODES[params.button]!);
+      return { pressed: params.button };
```
hubgan, line 96: No Android `button` test asserts that the simulator-server transport is not used here. If this early return were removed, control would fall through to the iOS `sendCommand(...)` Down/Up path with an undefined service (a mocked no-op under test), and the Android cases would still resolve to `{ pressed }` and stay green while an Android press silently went back over the HID transport this PR is moving off. The iOS cases pin the mirror direction (adb not called); the Android side has no equivalent guard.

packages/tool-server/src/tools/keyboard/platforms/android.ts
```diff
@@ -1,27 +1,54 @@
...
+    await injectAndroidNamedKey(device.id, params.key);
+    keysPressed++;
+  }
+  if (params.text) {
+    await injectAndroidText(device.id, params.text);
+    // `injectAndroidText` (via `assertTypeableAndroidText`) has already rejected
+    // any non-ASCII, so every character here is a single codepoint and a single
+    // UTF-16 unit — `.length` is the codepoint count (matching the tv /
+    // simulator-server backends) without a spread.
+    keysPressed += params.text.length;
```
hubgan, line 37: This counts UTF-16 code units, whereas the tv / vega / simulator-server backends count code points. It matches them only because `assertTypeableAndroidText` has already rejected every non-ASCII character before this line runs, so each char is one unit. Worth flagging that the correctness of the `keys` count is silently coupled to that ASCII guard staying in front of it.

packages/tool-server/src/utils/android-input.ts
```diff
@@ -0,0 +1,139 @@
...
+
+// `input` opens the app-process VM per call, so it is not instant; 15s comfortably
+// covers a single text/keyevent injection on a slow CI emulator while still
+// bounding a hung adb child.
+const ADB_INPUT_TIMEOUT_MS = 15_000;
+
+/** Type text into the focused field via `adb shell input text`. No-op for "". */
+export async function injectAndroidText(serial: string, text: string): Promise<void> {
+  assertTypeableAndroidText(text);
+  if (text.length === 0) return;
```
hubgan: This early return is unreachable as behavior: for `""`, `splitForVerbatimPercent` already returns `[]` (the regex match is null, `?? []`), so the loop below no-ops and no `input text` is issued. Harmless, just noting it's dead relative to the split logic.

hubgan (commented):
Tested this end-to-end against a branch tool-server driving iPhone 16 Pro and a Pixel 3a, plus a purpose-built `hw.keyboard=no` AVD to reproduce #449. On that stripped AVD the new `adb input` path types into a focused field where the old HID transport silently dropped the keys. Verified on-device: text (spaces, uppercase, digits), the `%`/`%s` verbatim-typing fix, shell-metachar safety, named keys, the 400 rejections (newline / emoji / non-ASCII / unknown key), and every Android button; the iOS keyboard/button path is unchanged and still works. A few follow-up notes inline, none blocking.

packages/tool-server/src/utils/android-input.ts
latekvo: Fixed in 8d44edcd. You're right that this made the same well-typed-but-unusable input map to different status codes by platform. Rather than special-case Android, I routed all four keyboard backends' input rejections through `InvalidToolInputError` (→400): `simulator-server-keys.ts` (unknown key + un-typeable char), `platforms/chromium.ts` (unknown key + no CDP descriptor), and `vega-input.ts` (unknown key + newline). So `key: "pageup"` is now 400 on iOS/chromium/vega too, matching Android.

Proved the divergence first (a focused repro showed the iOS and vega backends threw a plain `Error`, i.e. not `InvalidToolInputError` → 500), then re-ran after the change to confirm it flipped. Added `keyboard-error-taxonomy.test.ts` pinning the four backends to `InvalidToolInputError` for unknown key / newline / un-typeable char in one place. `vega-injection.test.ts`'s `/Unknown Vega key/` assertion still passes (message unchanged; `InvalidToolInputError` extends `Error`).

packages/tool-server/src/tools/keyboard/platforms/android.ts
latekvo: Fixed in 8d44edcd. Added a preflight test that drives the tool's real `execute` (i.e. `dispatchByPlatform`), not `makeAndroidImpl().handler` directly, so the android branch's `requires: ["adb"]` is actually exercised: with `ensureDeps` stubbed to reject, `keyboard` on an android target rejects with `DependencyMissingError` and issues no `adb input`; a companion happy-path test asserts `ensureDeps(["adb"])` runs before injection.

Proved the gap was real by mutation: deleting `requires: ["adb"]` makes `execute` resolve `{ typed: "hi", keys: 2 }` instead of rejecting, turning both new assertions red (`promise resolved … instead of rejecting`, `ensureDeps` not called with `["adb"]`); restoring it goes green.

packages/tool-server/src/tools/button/index.ts
latekvo: Fixed in 8d44edcd. The Android button cases now assert `sendCommand` (the sim-server HID transport) is never called — the mirror of the iOS cases' `injectAndroidKeycode` not-called guard, on both the `back` case and the every-button loop.

Proved by mutation: dropping the android branch's `return` (so control falls through to the sim-server `Down → Up` path) keeps `injectAndroidKeycode` / `ensureDep` / `{ pressed }` green, but the new `expect(sendCommand).not.toHaveBeenCalled()` fails with `expected "vi.fn()" to not be called at all, but actually been called 2 times`; restoring the `return` goes green.

packages/tool-server/src/tools/keyboard/platforms/android.ts
latekvo: This is the deliberate, documented coupling rather than a latent bug, so I'm leaving it as-is. The comment right above the line (`android.ts:25-28`) states it explicitly: `injectAndroidText` runs `assertTypeableAndroidText` on its first line, which rejects every non-ASCII char, so each char here is one codepoint == one UTF-16 unit and `.length` is the codepoint count. The guard runs unconditionally in the same call immediately before the count, and the routing tests pin the value (`keys: 8` for `hi there`, `keys: 4` for key + `abc`).

Re-introducing `[...text].length` is exactly the dead defensiveness the previous review round (d8517bf9) asked to remove, so the documented `.length` + comment is the intended end state.

packages/tool-server/src/utils/android-input.ts
latekvo: Fixed in 8d44edcd. Removed the dead `if (text.length === 0) return` — as you note, `splitForVerbatimPercent("")` returns `[]`, so the loop no-ops for `""` without a separate guard (and the sole caller, `typeAndroidPhone`, already gates on `if (params.text)`). Reworded the loop comment to state the empty-string no-op comes from the split. The existing `does not shell out for empty text` test still pins `injectAndroidText(serial, "")` → zero adb calls, so the behavior is unchanged and covered.

packages/tool-server/src/utils/android-input.ts
latekvo: Follow-up after rebasing onto current `main`: #420 (`fix(telemetry): add missing failure codes`, merged into main after this branch was cut) had independently reclassified these same iOS/chromium/vega throw sites from a plain `Error` to a `FailureError` with granular codes — `KEYBOARD_KEY_UNSUPPORTED` / `KEYBOARD_CHARACTER_UNSUPPORTED` / `VEGA_TEXT_INVALID` — but it left them mapping to HTTP 500 (the dispatcher only special-cases 404/424/400-unsupported/501; a bare `FailureError` still falls through to 500). So the divergence you flagged is real against main too, just via `FailureError` rather than plain `Error`.

Reconciled in the merge (7fa2cd81) so both concerns are satisfied at once: every keyboard input-rejection now throws `InvalidToolInputError` (→ 400, uniform across Android/iOS/chromium/vega) **carrying #420's telemetry code**. `InvalidToolInputError` gained an optional signal override for this — the 400 mapping keys off the error *class*, not the code, so the granular failure bucket is preserved. android's unknown-key path adopts `KEYBOARD_KEY_UNSUPPORTED` too for parity.

Verified both invariants hold together: main's `failure-classification.test.ts` (the #420 code pins) and this PR's `keyboard-error-taxonomy.test.ts` (400 + code pins for all four backends) are green.

packages/tool-server/src/utils/android-input.ts
```diff
@@ -0,0 +1,148 @@
...
+    // InvalidToolInputError), not an internal server fault (500).
+    throw new InvalidToolInputError(
+      'keyboard text must not contain a newline on Android; press it with key: "enter" instead'
+    );
+  }
+  for (const char of text) {
+    const cp = char.codePointAt(0)!;
+    if (cp < 0x20 || cp > 0x7e) {
+      const hex = cp.toString(16).toUpperCase().padStart(4, "0");
+      throw new InvalidToolInputError(
```
hubgan, line 96: This rejection (a character that can't be typed via `adb input text`) throws `InvalidToolInputError` without a signal override, so it buckets under the generic `TOOL_INPUT_INVALID` telemetry code. The semantically identical "this character can't be typed on this backend" rejection carries the granular `KEYBOARD_CHARACTER_UNSUPPORTED` code on the iOS (`simulator-server-keys.ts`) and chromium backends, and the named-key rejection just below in this file (`injectAndroidNamedKey`) carries `KEYBOARD_KEY_UNSUPPORTED`. As written, an un-typeable character on Android falls out of the `KEYBOARD_CHARACTER_UNSUPPORTED` bucket the other backends populate, so telemetry for the same failure diverges by platform. The HTTP status is unaffected (still 400).

packages/tool-server/src/utils/android-input.ts
```diff
@@ -0,0 +1,148 @@
...
+// `input text` receives the string as a single argv token (we `shellQuote` it, so
+// the device shell doesn't split on spaces). It reliably types only printable
+// ASCII: spaces and punctuation work, but a newline can't be represented, emoji
+// crash `InputShellCommand.sendText` with a NullPointerException, and other
+// non-ASCII (accented letters, CJK) is silently dropped by the virtual
+// KeyCharacterMap. Reject anything outside printable ASCII up front, naming the
+// offending character, so the caller gets a clear error instead of a cryptic
+// crash or a silently-wrong field. (`%` is handled separately — see
+// `splitForVerbatimPercent` — because it is typeable but needs escaping.)
+export function assertTypeableAndroidText(text: string): void {
```
hubgan, line 67: The Android-TV keyboard path (`blueprints/android-tv-control.ts` `type()`) injects through the same `adb shell input text` transport as this new phone path and inherits the same limitations (newline unrepresentable, emoji crashes `sendText`, non-ASCII silently dropped), but it does not run this guard. A newline or emoji sent to `keyboard` on an Android-TV target therefore surfaces as a 500 from `adbShell`, whereas the identical input on an Android phone now returns a clean 400 with an actionable message. The unified input-rejection taxonomy stops short of the one other backend that shares this exact sink.

packages/tool-server/src/utils/android-input.ts
```diff
@@ -0,0 +1,148 @@
...
+// `input text "100%safe"` silently types `100 afe`. Split the text so that every
+// `%` is the LAST character of its segment and issue one `input text` per segment:
+// within a segment a `%` is therefore never immediately followed by `s`, so
+// sendText can't fire that transform, and the segments concatenate on-device to
+// the exact input. A `%`-free string yields a single segment (one `input text`),
+// identical to before.
+//   "100%safe" → ["100%", "safe"] → "100%" + "safe" = "100%safe"
+//   "%s"       → ["%", "s"]        → "%" + "s"       = "%s"
+//   "%%"       → ["%", "%"]        → "%" + "%"       = "%%"
+function splitForVerbatimPercent(text: string): string[] {
```
hubgan, line 125: This handles the `input text` `%s`→space quirk. `blueprints/android-tv-control.ts` `type()` already handles the same quirk independently, with a different approach (`split(' ')` + `split(/(?<=%)/)` + `KEYCODE_SPACE` for spaces). Two separate implementations of the same on-device transport quirk in two files can drift — a future correction to one won't reach the other.

packages/tool-server/src/http.ts
```diff
@@ -717,6 +718,24 @@ export function createHttpApp(registry: Registry, options?: HttpAppOptions): Htt
...
+        // instead of a misleading 500.
+        //
+        // Ordering invariant: this check runs AFTER `findDependencyMissing`
+        // above, which walks the entire cause chain. That is unambiguous only
+        // because `InvalidToolInputError` is always thrown as a leaf (no
+        // `.cause`) — if a throw site ever nested a `DependencyMissingError` as
+        // the cause of an `InvalidToolInputError`, the earlier whole-chain scan
+        // would surface the dependency error and map the response to 424 first.
+        // Keep `InvalidToolInputError` causeless, or reorder these two checks.
+        const invalidInputErr = findErrorInCauseChain(err, InvalidToolInputError);
```
hubgan, line 736: The ordering invariant documented just above — that this 400 check is safe only because it runs after the whole-chain `findDependencyMissing` scan and because `InvalidToolInputError` is always a causeless leaf — isn't exercised by a test. No case constructs an error whose cause chain carries both a `DependencyMissingError` and an `InvalidToolInputError`, so reordering these two blocks (or a throw site that nests a dependency error under an `InvalidToolInputError`) would flip a 424 to a 400 while the suite stays green.

hubgan (commented):
Exercised this diff end-to-end against real devices (iOS iPhone 17 Pro, an Android emulator with `hw.keyboard=yes`, and a Chromium app) plus the full tool-server unit suite. The Android adb-input path and the 500→400 input-rejection taxonomy behave as intended, including the `%s`-verbatim handling and the per-backend error mapping. A few non-blocking consistency and coverage notes inline.

hubgan (approved):

packages/tool-server/src/utils/android-input.ts
latekvo: Fixed in 72f2c9c3. Both `assertTypeableAndroidText` rejections — the non-ASCII character and the newline (also a character this backend can't type) — now carry `KEYBOARD_CHARACTER_UNSUPPORTED`, so an un-typeable char on Android buckets with the same failure on iOS/chromium instead of falling into the generic `TOOL_INPUT_INVALID`.

Proved the divergence first (a focused repro asserted the pre-fix signal was `TOOL_INPUT_INVALID`), then re-ran after the change to confirm it flipped. `keyboard-error-taxonomy.test.ts` now pins both android text rejections to 400 + `KEYBOARD_CHARACTER_UNSUPPORTED` alongside the other backends' pins.

packages/tool-server/src/utils/android-input.ts
latekvo: Fixed in 72f2c9c3. `android-tv-control.ts` `type()` now runs the same `assertTypeableAndroidText` guard as the phone path before touching the sink, so a newline / emoji / non-ASCII char on an Android-TV target rejects as the same clean 400 (`InvalidToolInputError` + `KEYBOARD_CHARACTER_UNSUPPORTED`) instead of surfacing as a raw adb 500.

Proved the gap first: a unit repro showed `type("😀")` resolving normally and passing `input text '😀'` straight to the adb sink (where `sendText` NPEs on-device), and a newline reaching the sink too; after the fix the same inputs reject before any adb call. Two new tests in `android-tv-control.test.ts` pin the rejection type, telemetry code, and that no adb call is issued.

packages/tool-server/src/utils/android-input.ts
latekvo: Fixed in 72f2c9c3. `splitForVerbatimPercent` is now exported from `utils/android-input.ts` and the Android-TV `type()` routes its `%`-adjacency handling through it, so the transport quirk has a single implementation a future correction lands in once.

I deliberately kept the TV path's KEYCODE_SPACE space handling rather than unifying the whole body onto `injectAndroidText`: the space transport difference is a documented, test-pinned design choice on the TV side, and changing it would be an on-device behavioral change this PR has no TV-hardware verification for — whereas the `%s` quirk logic (the drift risk here) is pure string manipulation with identical semantics in both files. The existing TV tests (`hello world` → text/keyevent-62/text, `50%save` → `'50%'`+`'save'`, empty no-op) pass unchanged through the shared helper, proving the adb call sequences are identical.

packages/tool-server/src/http.ts
latekvo: Fixed in 72f2c9c3. Added a dual-class-chain case to `http-dep-gate.test.ts`: a tool throws an `InvalidToolInputError` whose `.cause` is a `DependencyMissingError`, and the test asserts the response is 424 (the whole-chain dependency scan wins). The invariant comment in `http.ts` now points at the pin.

Proved the test has teeth by mutation: temporarily reordering the two dispatcher checks (input-error check before `findDependencyMissing`) turns exactly this test red (400 instead of 424) while the other eight dep-gate cases stay green; restoring the order goes green again. So both failure modes you named — reordering the blocks, or a throw site nesting a dependency error under an `InvalidToolInputError` — now flip a test instead of silently flipping the status code.

latekvo:
Testing whether a video can be attached to a comment purely over the API 🎥

https://github.com/user-attachments/assets/53f6b583-6e1f-4f71-9316-30485af19ee5

senicko:
Testing programmatic (non-interactive) video attachment upload: a tiny synthetic clip was turned into a `File` in-page from base64 and dispatched as a `drop` `DragEvent` (with a real `DataTransfer`) straight at the `<file-attachment>` element - no manual drag-and-drop, no CDP `DOM.setFileInputFiles`, no direct call to `/upload/policies/assets`. GitHub's own JS on the page performed the real upload and returned this asset URL:

<video src="https://github.com/user-attachments/assets/728094ff-4ef6-40df-91bb-fde7f6c6d108"></video>

(Leaving this as a comment on a closed PR purely as a scratchpad for the experiment - feel free to delete.)

PR #457, opened 2026-07-02
test(screenshot-diff): pin that the OCR/font pass gets the normalized images

latekvo:
Follow-up test hardening for the merged #442 (`fix(screenshot-diff): rescale OCR text bounds into the normalized diff space`). Rides on a fresh branch off main.

## Gap
`diffPngFiles` (screenshot-diff.ts) hands the OCR/font pass the *normalized* (rescaled) images and each image's decoded->normalized region scale, so the rescaled text bounds and the pixels they crop share one coordinate space. #442 pinned the per-image region scales for a single baseline-larger fixture — but that fixture left the *current*-side assertions trivial: `normalizeToCommonSize` only ever downscales the *larger* image and returns the smaller one untouched, so on the untouched side decoded == normalized and the assertion holds whether production passes the raw or the normalized value.

## Tests
- Replaces the #442 region-scale test and the initial normalized-images test with one parametrized `it.each` over **both** scale directions (baseline-larger and current-larger). Each direction makes the downscaled side's `baselineImage`/`currentImage` **and** `baselineRegionScale`/`currentRegionScale` load-bearing, so all four wiring fields are pinned across the two cases. Verified by mutation: reverting any one of them to the raw decoded value / identity scale fails the matching direction (and only that direction), whereas before, reverting `currentImage` to the decoded value left the suite green.
- Adds a both-directions `hasPixelDiff` test (no diff -> `false`, a pixel change -> `true`), covering the one remaining load-bearing OCR-call input that had no wiring coverage: hardcoding it (e.g. to `false`, which silently disables text analysis for every real diff) previously left the whole suite green.

No production code changed.

### DIFF
--- a/packages/tool-server/test/screenshot-diff.test.ts
+++ b/packages/tool-server/test/screenshot-diff.test.ts
@@ -285,9 +285,8 @@ describe("diffPngFiles", () => {
   // baselineImage/currentImage (or the matching region scale) to the decoded value
   // would surface the raw size and fail here. The untouched side's assertions hold
   // for both the decoded and normalized value, so we exercise BOTH directions to pin
-  // both sides. Without this wiring (the pre-fix screenshot-diff.ts) OCR is handed
-  // the raw images and no region scales, and the half-image spurious "moved" change
-  // returns.
+  // both sides. Without this wiring the OCR pass would be handed the raw images and
+  // no region scales -- the mismatch #442 fixed.
   it.each([
     {
       larger: "baseline",
@@ -323,6 +322,38 @@ describe("diffPngFiles", () => {
           baselineRegionScale,
           currentRegionScale,
         })
+      );
+    }
+  );
+
+  // diffPngFiles always calls the OCR/font pass, but tells it whether the pixel diff
+  // found any changes via hasPixelDiff; the pass uses that to skip text analysis when
+  // nothing moved. Pin the wiring in both directions -- identical images -> false, a
+  // real pixel change -> true -- so hardcoding the source (e.g. to false, which would
+  // silently disable text analysis for every real diff) fails here instead of staying
+  // green.
+  it.each([
+    { name: "no pixel diff", changePixel: false, expected: false },
+    { name: "a pixel diff", changePixel: true, expected: true },
+  ])(
+    "tells OCR whether the pixel diff found changes ($name -> hasPixelDiff=$expected)",
+    async ({ changePixel, expected }) => {
+      const dir = await makeTempDir();
+      const baselinePath = path.join(dir, "baseline.png");
+      const currentPath = path.join(dir, "current.png");
+      await writePng(baselinePath, 2, 20, { r: 0, g: 0, b: 0 });
+      await writePng(
+        currentPath,
+        2,
+        20,
+        { r: 0, g: 0, b: 0 },
+        changePixel ? [{ x: 1, y: 10, rgb: { r: 255, g: 0, b: 0 } }] : []
+      );
+
+      await diffPngFiles({ baselinePath, currentPath, outputDir: dir });
+
+      expect(analyzeScreenshotTextChangesMock).toHaveBeenCalledWith(
+        expect.objectContaining({ hasPixelDiff: expected })
       );
     }
   );
--- a/packages/tool-server/test/screenshot-diff.test.ts
+++ b/packages/tool-server/test/screenshot-diff.test.ts
@@ -275,52 +275,57 @@ describe("diffPngFiles", () => {
     );
   });
 
-  it("hands OCR the decoded->normalized region scale for each image (diffPngFiles wiring)", async () => {
-    // Same 2:1 aspect, different resolution: normalizeToCommonSize downscales the
-    // larger baseline (480x240) to the current's size (240x120). diffPngFiles must
-    // compute each image's decoded->normalized scale and pass it to the OCR pass,
-    // so text bounds land in the shared pixel-diff coordinate space. Without this
-    // wiring (the pre-fix screenshot-diff.ts) OCR is called with no region scales
-    // and the half-image spurious "moved" change returns.
-    const dir = await makeTempDir();
-    const baselinePath = path.join(dir, "baseline.png");
-    const currentPath = path.join(dir, "current.png");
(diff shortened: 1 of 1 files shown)


hubgan (commented):
Reviewed the added test and verified it end to end. It is meaningful: reverting the production `baselineImage: baseline` wiring to `decodedBaseline` makes this test (and only this test) fail with `received width 480 / height 240`, so it correctly locks in the normalized-image wiring the fix introduced. I also exercised the real diff engine on live iPhone and Android captures at mismatched resolutions and confirmed the OCR/font pass receives the normalized images with no spurious text changes. A few non-blocking notes inline on coverage and duplication.

packages/tool-server/test/screenshot-diff.test.ts
```diff
@@ -297,6 +297,30 @@ describe("diffPngFiles", () => {
...
+    const currentPath = path.join(dir, "current.png");
+    await writePng(baselinePath, 480, 240, { r: 0, g: 0, b: 0 });
+    await writePng(currentPath, 240, 120, { r: 0, g: 0, b: 0 });
+
+    await diffPngFiles({ baselinePath, currentPath, outputDir: dir });
+
+    expect(analyzeScreenshotTextChangesMock).toHaveBeenCalledWith(
+      expect.objectContaining({
+        baselineImage: expect.objectContaining({ width: 240, height: 120 }),
+        currentImage: expect.objectContaining({ width: 240, height: 120 }),
```
hubgan: This `currentImage` assertion is effectively a tautology for this fixture, so it does not actually guard the current-image wiring. `normalizeToCommonSize` only downscales the larger image; with baseline 480x240 and current 240x120, the current image is already at the common size and passes through unchanged, so the raw decoded current and the normalized current are the same 240x120 object. The assertion therefore holds whether production passes `current` or `decodedCurrent`. I confirmed this by mutation: reverting `currentImage: current` to `currentImage: decodedCurrent` in the source leaves this test green, whereas the equivalent `baselineImage` mutation does fail it. As written, only the `baselineImage` line above is load-bearing, and a current-side regression would not be caught here.

packages/tool-server/test/screenshot-diff.test.ts
```diff
@@ -297,6 +297,30 @@ describe("diffPngFiles", () => {
       })
     );
   });
+
+  it("hands OCR the normalized images, not the raw decoded ones (diffPngFiles wiring)", async () => {
```
hubgan: This case shares its entire setup with the sibling test immediately above (`hands OCR the decoded->normalized region scale for each image`): the same 480x240 / 240x120 fixture, the same single `diffPngFiles` call, and the same `toHaveBeenCalledWith(expect.objectContaining(...))` against the same mock invocation. The two differ only in which fields of that one call they assert (`baselineImage` / `currentImage` here vs `baselineRegionScale` / `currentRegionScale` there). Flagging the duplicated fixture and call across the two cases.

packages/tool-server/test/screenshot-diff.test.ts
```diff
@@ -297,6 +297,30 @@ describe("diffPngFiles", () => {
...
+    // Same 2:1 aspect, different resolution: normalizeToCommonSize downscales the
+    // larger baseline (480x240) to the current's size (240x120). The OCR/font pass
+    // rescales its text bounds into that shared 240x120 space, so it must also be
+    // handed the NORMALIZED images to crop from. Feeding it the raw decoded
+    // baseline (480x240) instead would crop the wrong region while the bounds are
+    // already normalized. Pin that baselineImage/currentImage carry the normalized
```
hubgan: This comment states the test pins that `baselineImage` and `currentImage` both carry the normalized dimensions, but for this fixture only the `baselineImage` half can actually fail. Because the current image (240x120) is already the common size, the `currentImage` assertion holds for both the raw and the normalized current, so the plural claim here overstates what the test verifies.

hubgan (approved):

packages/tool-server/test/screenshot-diff.test.ts
latekvo: Fixed in 0205c070.

You're exactly right, and I reproduced your mutation: for the baseline-larger fixture `normalizeToCommonSize` returns `current` untouched (the smaller image is never resized), so `currentImage: { 240, 120 }` held whether prod passed `current` or `decodedCurrent` — reverting `currentImage: current` -> `decodedCurrent` left the old test green, while the `baselineImage` mutation failed it.

Replaced the single-direction test with a parametrized `it.each` over both scale directions. The new 'current is downscaled' case (baseline 240x120 vs current 480x240) makes `normalizeToCommonSize` downscale `current` instead, so `currentImage` (and `currentRegionScale`) become load-bearing there. Re-ran the mutation on the fix: `currentImage: current` -> `decodedCurrent` now fails the 'current is downscaled' case with received width 480 / height 240, and `baselineImage: baseline` -> `decodedBaseline` fails the 'baseline is downscaled' case — so both image-wiring lines are now guarded.

packages/tool-server/test/screenshot-diff.test.ts
latekvo: Fixed in 0205c070.

Folded the region-scale test and this image test into a single parametrized `it.each` over the two scale directions. Each case runs one fixture + one `diffPngFiles` call and asserts all four wiring fields (`baselineImage`/`currentImage`/`baselineRegionScale`/`currentRegionScale`) on that one mock invocation, so the duplicated fixture-and-call across the two tests is gone. As a bonus of merging them, each direction now makes the downscaled image's assertions load-bearing, which is what closed the current-side gap from the sibling thread.

packages/tool-server/test/screenshot-diff.test.ts
latekvo: Fixed in 0205c070.

The old comment's plural claim did overstate it — for a single baseline-larger fixture only the baseline half could fail. Rewrote the comment to say exactly that: `normalizeToCommonSize` only ever downscales the larger image and returns the smaller one untouched, so for any one fixture only the downscaled side's image and region scale differ from the raw decoded ones (i.e. are load-bearing); the untouched side's assertions hold for both values. The test now exercises both directions precisely so both sides get pinned.

PR #458, opened 2026-07-02
fix(cli): let a tool's own --args field win over the whole-payload flag

### LINKED ISSUE #452: flow-add-step: global `--args` shadows the per-field `--args`, so the documented per-flag form drops the step's arguments
## Summary

For `argent run flow-add-step`, the tool's own `--args` flag (the JSON string of arguments for the step's tool call) is **shadowed by the global `--args <json>` flag** ("Pass the entire payload as JSON (overrides flags)"). As a result, the **documented per-flag form — the exact example printed in `--help` — silently drops the step's arguments**, and the recorded/executed tool call runs with empty args.

## Environment
- @swmansion/argent 0.13.0
- Node v24.16.0, macOS (darwin)

## The collision
`argent run flow-add-step --help` documents:
```
--command <value>  string (required)  MCP tool name (e.g. "tap", "screenshot", "launch-app")
--args <value>     string  Tool arguments as a JSON string, e.g. '{"udid": "ABC", "x": 0.5, "y": 0.3}'.
...
Global flags:
--args <json>      Pass the entire payload as JSON (overrides flags)
```
Both the tool field and the global option are named `--args`, and the global one wins.

## Reproduction

```console
$ argent run flow-start-recording --name colltest --project_root /tmp/ws --executionPrerequisite "Settings open"
{ "message": "Started recording \"colltest\" flow", ... }

# A) Documented per-flag form (matches the --help example) — FAILS:
$ argent run flow-add-step --command gesture-tap --args '{"udid":"<UDID>","x":0.5,"y":0.35}'
[Tool:flow-add-step] [Tool:gesture-tap] Invalid params for tool "gesture-tap": [
  { "expected": "string", "code": "invalid_type", "path": ["udid"],
    "message": "Invalid input: expected string, received undefined" },
  { "expected": "number", "code": "invalid_type", "path": ["x"], ... } ]
# --command was honored, but the --args JSON was swallowed by the global payload parser,
# so gesture-tap ran with NO arguments.

# B) Only the full-payload form works — and note `args` must be a JSON *string*, not an object:
$ argent run flow-add-step --args '{"command":"gesture-tap","args":"{\"udid\":\"<UDID>\",\"x\":0.5,\"y\":0.35}"}'
{ "message": "Step added to \"colltest\" flow", "toolResult": { "tapped": true }, ... }
```

## Expected
The documented `flow-add-step --command <tool> --args '<json>'` form works and records/executes the step with the given arguments.

## Actual
The per-flag `--args` value is consumed by the global `--args` (whole-payload) option; `--command` is set but the step's arguments are lost, producing a misleading downstream validation error on the inner tool. Only the nested full-payload form (with `args` as a JSON-encoded string) succeeds.

## Suggested fix
Either rename the tool field (e.g. `--tool-args`) so it no longer collides with the global `--args`, or make the global whole-payload option non-shadowing for tools that declare their own `args` field. At minimum, fix the `--help` example, since it currently demonstrates the form that fails.

latekvo:
## Bug

`argent run flow-add-step` declares its own `args` field — a JSON string holding the recorded step's tool arguments. But the CLI flag parser hardcoded `--args` as a whole-payload escape hatch, so `--args` shadowed the tool's own `args` field. The documented per-flag form

```
argent run flow-add-step --command gesture-tap --args '{"udid":"...","x":0.5,"y":0.5}'
```

silently dropped the step arguments: the `--args` JSON was consumed as the whole payload and the tool's `args` field stayed `undefined`, producing a misleading downstream validation error. Only the nested full-payload form worked.

## Fix

Make `--args` schema-aware in `parseFlags()`:

- When the tool's schema declares a property named `args` (`schema.properties.args`), `--args <value>` is routed through the normal per-field handling and coerced by its declared type (a `string` for flow-add-step, so the raw JSON string passes straight into the `args` field). Both `--args value` and `--args=value` forms are supported.
- When the schema does NOT declare `args`, the existing whole-payload behavior is preserved exactly, including the `--args -` stdin sentinel.

`printToolHelp()` no longer prints the global whole-payload `--args <json>` / `--args -` lines for tools that declare their own `args` field, so the help stops contradicting the per-field `--args` shown in the schema flags.

The change is general: any tool with an `args` field benefits, with no special-casing by tool name.

## Verification

- `npm run build` (root tsc) — clean.
- `npm test -w @argent/cli` — 172 passed, including 6 new `parseFlags` cases covering schema-with-`args` (space-separated + inline forms, plus sibling fields) and schema-without-`args` (whole-payload preserved, `--args -` sentinel, no-schema fallback).
- Prettier + ESLint (`--max-warnings 0`) clean on the touched files.

Full CLI end-to-end (running `flow-add-step` against a live tool-server with an active flow recording) was not exercised here; the added unit tests prove the parser behavior.

Fixes #452
---

### CLI end-to-end verification

Ran the documented per-flag form against a live tool-server with an active flow recording:

- **Before (global argent 0.13.0):** `argent run flow-add-step --command gesture-tap --args '{"udid":"...","x":0.5,"y":0.35}'` → `Invalid params for tool "gesture-tap": udid expected string, received undefined ...` (args dropped).
- **After (this branch):** same command → `{ "message": "Step added to \"colltest\" flow", "toolResult": { "tapped": true } }`, and the recorded YAML contains the step **with** its args (`udid`/`x`/`y`). The per-flag form now works as documented.

### DIFF
--- a/packages/argent-cli/src/flag-parser.ts
+++ b/packages/argent-cli/src/flag-parser.ts
@@ -11,6 +11,12 @@
 //   --args '<json>'       (whole-payload escape hatch; merges with parsed flags)
 //   --args -              (read whole-payload JSON from stdin)
 //
+// Exception: when the tool's own schema declares a property named `args` (e.g.
+// flow-add-step, whose `args` is a JSON string of the step's tool arguments),
+// `--args <value>` is treated as that per-field value — coerced by its declared
+// type, exactly like any other field — NOT the whole-payload escape hatch. Such
+// a tool has no whole-payload shortcut; use individual flags / --<field>-json.
+//
 // Scalar field types come from JSON Schema: string, number, integer, boolean, enum.
 // Array fields: items.type must be a scalar to get a repeatable flag.
 // Object fields and arrays of objects fall through to --field-json.
@@ -127,7 +133,11 @@ export function parseFlags(argv: string[], schema: JsonSchema | undefined): Flag
     }
 
     // ── Whole-payload escape hatch ──
-    if (flag === "args") {
+    //
+    // Skipped when the tool declares its own `args` field: then `--args` is a
+    // normal per-field flag (handled below) and wins over the escape hatch, so
+    // the field's value isn't silently swallowed as the whole payload.
+    if (flag === "args" && properties.args === undefined) {
       const { value, nextIndex } =
         inlineValue !== undefined ? { value: inlineValue, nextIndex: i } : takeNext(i, "args");
       rawArgs = value;
--- a/packages/argent-cli/src/flag-parser.ts
+++ b/packages/argent-cli/src/flag-parser.ts
@@ -91,6 +91,13 @@ function parseJsonOrThrow(raw: string, label: string): unknown {
  */
 export function parseFlags(argv: string[], schema: JsonSchema | undefined): FlagParseResult {
   const properties = schema?.properties ?? {};
+  // The whole-payload `--args '<json>'` escape hatch only exists when the tool
+  // does NOT declare its own `args` field. When it does (e.g. flow-add-step),
+  // `--args` is that per-field flag, so the "or --args '<json>'" fallback the
+  // error messages suggest would be a dead end — that form re-enters per-field
+  // handling instead of the hatch, and only `--<field>-json` works. Emit the
+  // suggestion only when the hatch is actually available.
+  const argsHatchHint = properties.args === undefined ? " or --args '<json>'" : "";
   const args: Record<string, unknown> = {};
   const positional: string[] = [];
   let helpRequested = false;
@@ -165,7 +172,7 @@ export function parseFlags(argv: string[], schema: JsonSchema | undefined): Flag
         // A prior --${fieldName} (scalar-array form) already populated this
         // field; overwriting it here would silently discard those values.
         throw new FlagParseException(
-          `--${fieldName} and --${flag} cannot be mixed for the same field; pass it entirely as --${flag} '<json>' or --args '<json>'`
+          `--${fieldName} and --${flag} cannot be mixed for the same field; pass it entirely as --${flag} '<json>'${argsHatchHint}`
         );
       }
       args[fieldName] = parseJsonOrThrow(value, `--${flag}`);
@@ -209,7 +216,7 @@ export function parseFlags(argv: string[], schema: JsonSchema | undefined): Flag
       const itemType = propSchema.items?.type;
       if (!isScalarType(itemType)) {
         throw new FlagParseException(
-          `--${flag} is an array of objects; pass it as --${flag}-json '<json>' or --args '<json>'`
+          `--${flag} is an array of objects; pass it as --${flag}-json '<json>'${argsHatchHint}`
         );
       }
       const { value, nextIndex } =
@@ -223,7 +230,7 @@ export function parseFlags(argv: string[], schema: JsonSchema | undefined): Flag
         // a scalar-coercion error — matching the --field-json branch above,
         // which checks mixing before parsing the JSON.
         throw new FlagParseException(
-          `--${flag} and --${flag}-json cannot be mixed for the same field; pass it entirely as --${flag}-json '<json>' or --args '<json>'`
+          `--${flag} and --${flag}-json cannot be mixed for the same field; pass it entirely as --${flag}-json '<json>'${argsHatchHint}`
         );
       }
       const coerced = coerceScalar(value, itemType, flag);
@@ -240,7 +247,7 @@ export function parseFlags(argv: string[], schema: JsonSchema | undefined): Flag
     // ── Object field: must use --field-json ──
     if (propSchema?.type === "object") {
       throw new FlagParseException(
-        `--${flag} is an object; pass it as --${flag}-json '<json>' or --args '<json>'`
+        `--${flag} is an object; pass it as --${flag}-json '<json>'${argsHatchHint}`
       );
     }
 
--- a/packages/argent-cli/src/run.ts
+++ b/packages/argent-cli/src/run.ts
@@ -72,13 +72,21 @@ async function readStdin(): Promise<string> {
 
 function printToolHelp(meta: ToolMeta): void {
   const description = meta.description?.trim() ?? "";
+  const schema = meta.inputSchema as JsonSchema | undefined;
+  // A tool may declare its own `args` field (e.g. flow-add-step). In that case
+  // `--args` is a per-field flag shown in the schema block above, so don't also
+  // advertise the whole-payload `--args` escape hatch — it no longer applies and
+  // would contradict the per-field flag.
+  const hasOwnArgsField = schema?.properties?.args !== undefined;
   console.log(`argent run ${meta.name} [flags]`);
   if (description) console.log(`\n${description}\n`);
   console.log("Flags:");
-  console.log(formatSchemaUsage(meta.inputSchema as JsonSchema));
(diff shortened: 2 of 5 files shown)


packages/argent-cli/test/flag-parser.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+describe("parseFlags — whole-payload --args (no own `args` field)", () => {
+  it("keeps --args as the whole-payload escape hatch", () => {
+    const result = parseFlags(["--args", '{"udid":"X","x":0.5,"y":0.5}'], gestureTapSchema);
+    expect(result.rawArgs).toBe('{"udid":"X","x":0.5,"y":0.5}');
+    // `args` must not appear as a parsed field.
+    expect("args" in result.args).toBe(false);
+  });
+
+  it("keeps the --args - stdin sentinel as whole-payload", () => {
```
hubgan, line 261: `--args -` is covered here only for a tool with no `args` field. The case this change actually introduces is the opposite one: a tool that declares its own `args` (flow-add-step), where `--args -` now resolves to the literal field value `"-"` and leaves `rawArgs` null instead of reading stdin. That path isn't asserted anywhere, so a later refactor that moved the `-` sentinel handling ahead of the `properties.args === undefined` gate would silently start routing flow-add-step's `--args -` to stdin, and the suite would stay green.

packages/argent-cli/src/run.ts
```diff
@@ -72,13 +72,21 @@ async function readStdin(): Promise<string> {
...
+  const hasOwnArgsField = schema?.properties?.args !== undefined;
   console.log(`argent run ${meta.name} [flags]`);
   if (description) console.log(`\n${description}\n`);
   console.log("Flags:");
-  console.log(formatSchemaUsage(meta.inputSchema as JsonSchema));
+  console.log(formatSchemaUsage(schema));
   console.log("\nGlobal flags:");
-  console.log("  --args <json>          Pass the entire payload as JSON (overrides flags)");
-  console.log("  --args -               Read the entire payload as JSON from stdin");
+  if (!hasOwnArgsField) {
```
hubgan, line 86: This suppression has no test. Nothing asserts that a tool declaring its own `args` drops the whole-payload `--args` lines while a tool without one keeps them. It is half of the user-visible change in this PR, so inverting or removing the `hasOwnArgsField` guard later would re-advertise a `--args <json>` hatch that flow-add-step does not honor, with no failing test to catch it.

hubgan (commented):
A couple of test-coverage notes inline.

packages/argent-cli/test/flag-parser.test.ts
latekvo: Fixed in cc8f3a35.

Added two flag-parser cases asserting that for a tool that owns its `args` field, `--args -` (and the inline `--args=-`) resolve to the literal field value `"-"` with `rawArgs` staying null — i.e. NOT the stdin sentinel. Proven real by mutation: routing the `-` sentinel ahead of the `properties.args === undefined` gate left the existing six cases green, and made these two go red; reverting the mutation returns them to green.

packages/argent-cli/src/run.ts
latekvo: Fixed in cc8f3a35.

New `test/run-help.test.ts` drives `printToolHelp` through the real `run(..., "--help")` entry point (mocked tools-client, captured `console.log`) and asserts a tool declaring its own `args` field omits the whole-payload `--args <json>` / `--args -` lines, while a tool without one keeps them. Proven real by mutation: removing the `hasOwnArgsField` guard left the full existing suite green and made this new test go red; reverting returns it to green.

packages/argent-cli/test/run-help.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+// Drive `printToolHelp` through the real `run(..., "--help")` entry point with a
+// mocked tools-client, capturing console.log. This locks in the user-visible
+// half of the fix: a tool that declares its own `args` field must NOT advertise
+// the whole-payload `--args <json>` / `--args -` escape hatch (it no longer
+// applies), while a tool without one keeps it.
+
+const toolsClientMock = vi.hoisted(() => ({
+  fetchTool: vi.fn(),
+  callTool: vi.fn(),
```
hubgan, line 12: This `callTool` mock is only ever asserted as `not.toHaveBeenCalled()`, on the two `--help` paths. Nothing drives `run(["flow-add-step", "--command", "gesture-tap", "--args", "{...}"])` without `--help` to assert that `callTool` receives `{ command: "gesture-tap", args: "{...}" }`. The parser result and the help suppression are each locked in, but the end-to-end payload for the exact invocation from the bug report isn't guarded at the `run()` layer — so a future change that reconnected flag parsing to the payload builder incorrectly could reintroduce the dropped-`args` regression with this suite still green.

packages/argent-cli/src/flag-parser.ts
```diff
@@ -138,7 +144,11 @@ export function parseFlags(argv: string[], schema: JsonSchema | undefined): Flag
     }
 
     // ── Whole-payload escape hatch ──
-    if (flag === "args") {
+    //
+    // Skipped when the tool declares its own `args` field: then `--args` is a
+    // normal per-field flag (handled below) and wins over the escape hatch, so
+    // the field's value isn't silently swallowed as the whole payload.
+    if (flag === "args" && properties.args === undefined) {
```
hubgan, line 158: With this guard, any tool that declares a top-level `args` field no longer reaches the whole-payload escape hatch. Today that's only `flow-add-step`, whose `args` is a `string`, so it routes cleanly through the scalar branch. If a future tool ever declared its top-level `args` as an object or array, `--args '<json>'` would fall into the object/array branches, whose error text still suggests retrying with `--args '<json>'` — a dead end for such a tool, since that form no longer routes to the hatch and would just re-enter the same branch; only `--args-json` would work. Noting it as a latent edge, not a problem for the current tool set.

hubgan (commented):
Two non-blocking observations inline.

hubgan (approved):

packages/argent-cli/test/run-help.test.ts
latekvo: Fixed in dd994a83.

Added `test/run-flow-add-step-payload.test.ts`, which drives the whole `run()` path (not just the parser) against a real in-process tool-server and captures the exact POST body for `argent run flow-add-step --command gesture-tap --args '{...}'`, asserting the payload is exactly `{ command, args }` — both the space-separated and inline `--args=` forms. Proven real by mutation: reverting the parser's `properties.args === undefined` guard so `--args` is swallowed as the whole payload makes the captured body `{ command, udid, x, y }` (args dropped, its keys hoisted) and turns the test red — the precise #452 regression, now guarded at the `run()` layer.

packages/argent-cli/src/flag-parser.ts
latekvo: Fixed in 4b602b02.

Gated the trailing `or --args '<json>'` suggestion on the whole-payload hatch actually being available (`properties.args === undefined`), via a single `argsHatchHint`, so a tool that owns its `args` field no longer points at the dead-end form — the object / array-of-objects / `-json`-mixing messages now suggest only `--<field>-json`. Still latent for the current tool set (flow-add-step's `args` is a string, so it never reaches those branches), but the messages now honour the invariant this PR introduced.

New flag-parser cases pin that an object/array `args` owner suggests only `--args-json` (never `or --args '<json>'`), while a control tool without its own `args` keeps the suggestion. Proven by mutation: making `argsHatchHint` unconditional turns exactly the two owner cases red and leaves the control green.

PR #459, opened 2026-07-02
fix(native-devtools): terminal state for non-injectable Apple system apps

### LINKED ISSUE #453: native-devtools injection status never reaches a terminal state for non-injectable apps (restart→retry loops forever)
## Summary

For an app that can **never** be injected (e.g. an Apple built-in app, which can't load the Argent devtools dylib), `native-devtools-status` keeps reporting `requiresRestart: true` / `nextLaunchWillBeInjected: true`, and the native-* tools return "restart-app then retry". Restarting the app does **not** change the status — there is no terminal "this app cannot be injected" state, so an agent following the guidance loops `restart-app → retry → restart-app …` forever.

## Environment
- @swmansion/argent 0.13.0
- Node v24.16.0, macOS (darwin), iOS 26.5 simulator (iPhone 17 Pro)
- Target app: `com.apple.Preferences` (Settings) — a non-injectable built-in app

## Reproduction

```console
$ argent run native-devtools-status --args '{"udid":"<UDID>","bundleId":"com.apple.Preferences"}'
{ "envSetup": true, "appRunning": true, "connected": false,
  "requiresRestart": true, "nextLaunchWillBeInjected": true }

# Follow the guidance — restart the app:
$ argent run restart-app --args '{"udid":"<UDID>","bundleId":"com.apple.Preferences"}'
{ "restarted": true, "bundleId": "com.apple.Preferences" }

# Re-check: identical status — restart did nothing to advance state:
$ argent run native-devtools-status --args '{"udid":"<UDID>","bundleId":"com.apple.Preferences"}'
{ "envSetup": true, "appRunning": true, "connected": false,
  "requiresRestart": true, "nextLaunchWillBeInjected": true }
```

The native-* tools (e.g. `native-describe-screen`, `native-find-views`) correspondingly return `restart_required` / "Call restart-app then retry" for this app, which will never succeed.

## Expected
The status should distinguish **"injection pending — restart to apply"** from **"this app cannot be injected"** (unsigned-for-injection / Apple built-in / entitlement mismatch), and the native-* tools should return a terminal error for the latter instead of an unbounded restart→retry instruction.

## Impact
Minor for humans, but it's an infinite loop for autonomous agents driving these tools against a built-in app — they retry indefinitely with no signal to stop.

latekvo:
Fixes #453

## The bug

For an app that can never be injected (an Apple built-in app such as `com.apple.Preferences`), `native-devtools-status` kept reporting `requiresRestart: true` / `nextLaunchWillBeInjected: true`, and the native-* feature tools (native-describe-screen, native-find-views, etc.) kept returning "restart-app then retry". Restarting never changes the state, so an autonomous agent loops `restart-app → retry` forever. There was no terminal "this app cannot be injected" signal.

## Why these apps can't inject

Apple system/built-in apps (bundle ids under `com.apple.`) are platform binaries with library validation enabled; the simulator ignores `DYLD_INSERT_LIBRARIES` for them, so the Argent native devtools dylib can never load. This is a reliable, well-understood signal on the simulator. Third-party apps the user installs carry no such restriction and inject fine. So the fix treats bundle ids starting with `com.apple.` as non-injectable, via a general predicate rather than hardcoding the single bundle id from the issue.

## The fix

- New pure predicate `isInjectableBundleId(bundleId)` in `blueprints/native-devtools.ts` returning `false` for `com.apple.*` bundle ids (case-sensitive prefix, matching how bundle ids are compared), with a doc comment explaining the library-validation rationale.
- New failure code `NATIVE_DEVTOOLS_NOT_INJECTABLE` in `@argent/registry`.
- `precheckNativeDevtools(api, udid, bundleId)` (the 3-arg overload) now throws a terminal `FailureError` with that code, *before* the `requiresAppRestart → restart_required` branch, when the bundle id is non-injectable. Because every native-* feature tool awaits this precheck and propagates the throw, all 7 tools now surface a hard, terminal error instead of an unbounded restart→retry instruction — no change needed in the individual tool files. The 2-arg overload (no bundleId), used by `native-devtools-status`, is unchanged and never throws (guarded by `bundleId !== undefined`).
- `native-devtools-status` now *reports* the state instead of throwing: it adds an `injectable: boolean` field. For a non-injectable app it returns `injectable: false, requiresRestart: false, nextLaunchWillBeInjected: false` (with `envSetup`/`appRunning`/`connected` still measured), giving an agent that checks status first a terminal signal. The tool description documents the new field and that `injectable: false` is terminal (do not restart/retry; use the standard describe / screenshot / view-at-point tools).

## Verification

- `npm run build` (root tsc): clean.
- `vitest run` on `native-devtools-status.test.ts`: 13/13 passing. Added coverage for the predicate, the status terminal state for a `com.apple.*` app, and that the 3-arg precheck throws `NATIVE_DEVTOOLS_NOT_INJECTABLE` while the 2-arg overload does not throw for the same api.
- Telemetry `sanitize.test.ts` (iterates every registered failure code): 49/49 passing — the new code is accepted.
- Prettier + eslint (`--max-warnings 0`) clean on all touched files.

Device E2E on an iOS simulator with `com.apple.Preferences` (confirming `native-devtools-status` returns `injectable: false` and the native-* tools return the terminal error rather than looping restart-app) is pending and will be run separately.
---

### Device verification (iOS 18.5 simulator, com.apple.Preferences)

Reproduced before/after on a real simulator with a dev tool-server:

**Before (origin/main):** `native-devtools-status` for `com.apple.Preferences` → `requiresRestart:true, nextLaunchWillBeInjected:true`; `native-describe-screen` → `restart_required`; after `restart-app` the status is identical (the infinite loop).

**After (this branch):**
- `native-devtools-status` → `{envSetup:true, appRunning:true, connected:false, requiresRestart:false, nextLaunchWillBeInjected:false, injectable:false}` — terminal.
- `native-describe-screen` / `native-find-views` → terminal error: "com.apple.Preferences is an Apple system app ... can never be injected ... use the standard describe / screenshot / view-at-point tools instead." (no more restart→retry loop).

**Regression (injectable third-party app `xyz.blueskyweb.app`):** `native-devtools-status` → `injectable:true` with the normal `requiresRestart:true`; `native-describe-screen` → the normal `restart_required` (not the terminal error). The predicate does not false-positive on normal apps.

---

### Review follow-ups (@hubgan on-device review)

- **Recovery guidance no longer recommends `view-at-point`.** Both view-at-point tools run this same 3-arg precheck and re-throw `NATIVE_DEVTOOLS_NOT_INJECTABLE` for a `com.apple.*` bundle, so recommending them dead-ended the agent. All three surfaces (the precheck throw, the `describe` iOS fallback hint, and the `native-devtools-status` description) now share one verbatim "do not fall back to the native-* tools" warning (`NON_INJECTABLE_NATIVE_WARNING`) and recommend only `describe` and `screenshot`. (52b888ce)
- **Propagation tests for all six feature tools.** Added tests that drive each 3-arg feature tool's `execute()` with a `com.apple.*` bundle and assert the throw propagates out as `NATIVE_DEVTOOLS_NOT_INJECTABLE`, so a future try/catch inside a tool can't silently regress the terminal behavior (proven non-vacuous via a swallow-the-throw mutation). (52b888ce)
- **Degraded sim + system app keeps the re-boot hint.** When the ax-service is degraded, `describe` now returns the `boot-device force=true` guidance instead of the terminal screenshot hint - a proper re-boot may let the ax-service read the system app's full tree. `should_restart` is still never set, so the restart loop stays removed. (5ace8be3)

---

### Second review round (@hubgan) + follow-up hardening

- **Terminal check hoisted above env init in the precheck.** Injectability is a static property of the bundle id, so the `NATIVE_DEVTOOLS_NOT_INJECTABLE` throw now fires before the `givenUp`/`ensureEnvReady` plumbing - a broken env can no longer mask the terminal signal behind `init_failed`'s re-boot guidance, and no env work is spent on a system app. Regression tests cover both masked paths. (ca7093c6)
- **The same ordering applied to the other two surfaces.** `native-devtools-status` computes the injectable gate before the precheck (a given-up sim no longer hides `injectable: false` behind `init_failed`), and `describe`'s gate moved ahead of native-devtools service resolution (a downed ios-remote tunnel or dispose race no longer swallows the terminal hint). (bef6dfcb)
- **Unreachable-sim fallback.** If the status tool's `isAppRunning` probe fails on a non-injectable app (shut-down sim), it falls back to the precheck so a broken sim still surfaces structured `init_failed` - whose re-boot guidance IS corrective there - instead of a raw subprocess error; a healthy env rethrows the probe failure. (722b432a)
- **Warning text scoped to the tools it is true of.** `NON_INJECTABLE_NATIVE_WARNING` now names the six injection-based feature tools instead of sweeping in all `native-*` tools - the `native-profiler-*` tools never inject and keep working on system apps. (bef6dfcb)
- **launch-app / restart-app boundary pinned.** New tests drive both tools (local iOS and ios-remote dispatch, all four call sites) with `com.apple.Preferences` and the real precheck, asserting the terminal throw never reaches them - proven non-vacuous by a 3-arg mutation failing exactly the mutated sites. (bef6dfcb, 722b432a)


hubgan (commented):
A few inline notes from reviewing and testing this branch on-device (iPhone 16 Pro and Pixel 3a).

packages/tool-server/src/blueprints/native-devtools.ts
```diff
@@ -81,6 +98,24 @@ export async function precheckNativeDevtools(
...
 
+  // Terminal case: an app that can never be injected (Apple system app). Throw
+  // instead of returning a restart-required block so the native-* feature tools
+  // surface a hard error rather than instructing an unbounded restart→retry
+  // loop that can never succeed. The 2-arg overload (bundleId undefined), used
+  // by native-devtools-status, must NOT throw — it reports the state instead.
+  if (bundleId !== undefined && !isInjectableBundleId(bundleId)) {
+    throw new FailureError(
+      `${bundleId} is an Apple system app: it is a platform binary with library validation, so Argent native devtools can never be injected into it. ` +
+        "The native-* tools are unavailable for this app — use the standard describe / screenshot / view-at-point tools instead.",
```
hubgan: This recovery text recommends `view-at-point`, but the only view-at-point tools are `native-view-at-point` and `native-user-interactable-view-at-point`, and both call this same 3-arg `precheckNativeDevtools` and re-throw this identical `NATIVE_DEVTOOLS_NOT_INJECTABLE` for a `com.apple.*` bundle. I confirmed this on a real simulator: `native-describe-screen` and `native-find-views` (and, via the same code path, the two view-at-point tools) all surface this terminal error for `com.apple.Preferences`. So an agent that follows the guidance to `view-at-point` lands right back on the same dead-end. The other two options listed do work here: `describe` returned the full Settings tree via ax-service and `screenshot` is unaffected.

packages/tool-server/src/blueprints/native-devtools.ts
```diff
@@ -81,6 +98,24 @@ export async function precheckNativeDevtools(
     });
   }
 
+  // Terminal case: an app that can never be injected (Apple system app). Throw
+  // instead of returning a restart-required block so the native-* feature tools
+  // surface a hard error rather than instructing an unbounded restart→retry
+  // loop that can never succeed. The 2-arg overload (bundleId undefined), used
+  // by native-devtools-status, must NOT throw — it reports the state instead.
+  if (bundleId !== undefined && !isInjectableBundleId(bundleId)) {
```
hubgan: All six native-* feature tools depend on this throw propagating out of `precheckNativeDevtools` (the description notes no change was needed in the individual tool files). That propagation isn't exercised by any test: the suite asserts the 3-arg precheck throws and the 2-arg overload does not, but nothing drives a feature tool's `execute()` with a `com.apple.*` bundle. I verified on-device that it does propagate today for all six, but a later change that wrapped this precheck call in a try/catch inside a tool would regress it silently while the suite stays green.

packages/tool-server/src/tools/native-devtools/native-devtools-status.ts
```diff
@@ -31,14 +33,16 @@ export const nativeDevtoolsStatusTool: ToolDefinition<Params, Result> = {
...
+Returns { envSetup, appRunning, connected, requiresRestart, nextLaunchWillBeInjected, injectable }:
 - envSetup: DYLD_INSERT_LIBRARIES is configured in the simulator's launchd environment
 - appRunning: the target bundle currently has a running UIKit process on the simulator
 - connected: the dylib is active in the current running process for this bundleId
 - requiresRestart: the app is already running but its current process does not have native devtools injected
 - nextLaunchWillBeInjected: if you launch this bundle now, native devtools env setup is already in place
+- injectable: whether native devtools can ever be injected into this app. Apple system apps (bundle ids under com.apple.) are platform binaries with library validation, so the dylib can never load into them.
 
 Call this before using app-scoped native hierarchy tools or native-network-logs.
+If injectable is false: this is a TERMINAL state — the app can never be injected. Do NOT restart/retry; use the standard describe / screenshot / view-at-point tools instead.
```
hubgan: Same point as the precheck error text: this description tells the agent to `use the standard describe / screenshot / view-at-point tools instead`, but the view-at-point tools re-throw the non-injectable error through this same precheck, so that part of the advice dead-ends on the identical terminal error.

packages/tool-server/src/tools/describe/platforms/ios/index.ts
```diff
@@ -24,6 +28,17 @@ const TVOS_HINT =
...
+// Apple system apps (`com.apple.*`) can never load argent's injected dylib, so
+// the native-devtools fallback can't read their view hierarchy and restarting
+// them would never help — returning `should_restart` here puts the agent in an
+// unbounded restart-app → describe loop. Surface a terminal hint to use a
+// screenshot instead, mirroring the terminal state `native-devtools-status`
+// reports for these bundle ids.
+const NON_INJECTABLE_HINT =
+  "This is an Apple system app (com.apple.*), which cannot load argent's native-devtools " +
+  "instrumentation — the native view hierarchy is unavailable and restarting the app will NOT " +
+  "help. Use the `screenshot` tool to see the screen and interact by coordinate.";
```
hubgan: This hint recommends only `screenshot`, whereas the precheck throw in `native-devtools.ts` and the `native-devtools-status` description both recommend `describe / screenshot / view-at-point` for the same non-injectable condition. Three surfaces describe the same terminal state with a different set of recommended tools, so the agent's next step depends on which one it happened to call.

packages/tool-server/src/blueprints/native-devtools.ts
latekvo: Fixed in 52b888ce.

You're right - both view-at-point tools run this same 3-arg `precheckNativeDevtools` and re-throw `NATIVE_DEVTOOLS_NOT_INJECTABLE` for a `com.apple.*` bundle, so recommending them here just dead-ends the agent back on this identical error. The recovery text no longer recommends `view-at-point`. It now points only at `describe` (which reads these apps via the ax-service, no injection) and `screenshot`, and names the two view-at-point tools only inside an explicit "do not fall back to the native-* tools" warning.

Verified with a new test that drives each view-at-point tool's `execute()` on `com.apple.Preferences` (both surface the terminal error), plus a consistency test asserting the recommendation clause never names a native-* tool outside the warning.

packages/tool-server/src/blueprints/native-devtools.ts
latekvo: Fixed in 52b888ce.

Added tests that drive all six 3-arg feature tools' `execute()` with a `com.apple.*` bundle and assert the throw propagates out as `NATIVE_DEVTOOLS_NOT_INJECTABLE`: native-describe-screen, native-find-views, native-full-hierarchy, native-network-logs, native-view-at-point, and native-user-interactable-view-at-point.

To confirm the test actually guards the regression you described, I temporarily wrapped the precheck call in native-view-at-point in a try/catch that swallowed the throw - the new test failed with "expected undefined to be an instance of FailureError" - then reverted, and it passes again.

packages/tool-server/src/tools/native-devtools/native-devtools-status.ts
latekvo: Fixed in 52b888ce.

Same root cause as the precheck-text thread: the view-at-point tools re-throw this identical error through this precheck, so recommending them dead-ends. The status description now shares the exact recovery clause used by the precheck throw (via a single exported constant), so it recommends `describe`/`screenshot` and no longer lists `view-at-point`.

packages/tool-server/src/tools/describe/platforms/ios/index.ts
latekvo: Fixed in 52b888ce.

All three surfaces now share one verbatim warning - "Do not fall back to the native-* tools ..." (a single exported constant, `NON_INJECTABLE_NATIVE_WARNING`) - so none of them can drift into recommending a dead-end.

The positive recommendation is deliberately context-dependent rather than byte-identical. The precheck throw and the status description are reached before `describe` has been tried, so they recommend `describe` and `screenshot`. This describe fallback hint, though, is reached only after `describe`'s own ax-service path already returned empty, so it leads with `screenshot` - re-recommending `describe` from inside describe's own hint would be circular and risks reintroducing the very loop this PR removes. A consistency test asserts the shared warning appears verbatim across all three surfaces, and the describe-tool test asserts this hint contains it.

packages/tool-server/test/describe-tool.test.ts
```diff
@@ -255,6 +259,65 @@ describe("describe tool", () => {
     expect(elementLineCount(result.description)).toBe(0);
   });
 
+  it("does NOT return should_restart for a non-injectable Apple system app (no restart loop)", async () => {
```
hubgan, line 262: Both non-injectable `describe` tests use an empty AX tree (`elements: []`), so they only cover the fallback branch and never the more common case: a `com.apple.*` app whose accessibility tree is non-empty. That case returns early in `describeIos` at the `tree.children.length > 0` return, before the injectability gate is reached, and it's the intended behavior for Settings et al. — on an iOS 26 simulator `describe com.apple.Preferences` with Settings foreground returns ~30 elements with no hint and no `should_restart`. Because that path is untested, moving the gate above the early return would silently replace a real system-app tree with the screenshot hint and no test would catch it.

packages/tool-server/src/blueprints/native-devtools.ts
```diff
@@ -18,6 +18,57 @@ export type NativeDevtoolsTransport = "unix" | "tcp";
...
+export function isInjectableBundleId(bundleId: string): boolean {
+  return !bundleId.startsWith("com.apple.");
+}
+
+/**
+ * The invariant half of the non-injectable recovery guidance: which tools NOT
+ * to fall back to. Shared VERBATIM by every surface that reports this terminal
+ * state (this precheck's throw, the `describe` iOS fallback hint, and the
+ * `native-devtools-status` description) so none of them can drift into
+ * recommending a dead-end. Every native-* tool — notably the two view-at-point
```
hubgan: "Every native-* tool ... re-throws this identical error" isn't accurate: `native-devtools-status` is also a native-* tool, and it deliberately does not re-throw — it reports `injectable: false` instead (the precheck's own comment below documents that exception). A reader could take this to mean status throws for a system app too.

packages/tool-server/src/blueprints/native-devtools.ts
```diff
@@ -18,6 +18,57 @@ export type NativeDevtoolsTransport = "unix" | "tcp";
...
+ *
+ * Apple system / built-in apps (bundle ids under `com.apple.`) are platform
+ * binaries shipped with library validation enabled. The simulator refuses to
+ * honour `DYLD_INSERT_LIBRARIES` for them, so our dylib can never load — no
+ * amount of relaunching changes that. Third-party apps the user installs carry
+ * no such restriction and inject normally. Treating the `com.apple.` prefix as
+ * non-injectable gives the native-* tools a terminal signal instead of an
+ * unbounded restart-app → retry loop.
+ *
+ * Bundle ids are case-sensitive, so a case-sensitive prefix check is correct.
```
hubgan: The stated rationale is off. iOS treats bundle identifiers case-insensitively for launch/uniqueness, so "bundle ids are case-sensitive" isn't why a case-sensitive prefix check is safe. In practice it's fine because Apple's system-app ids are always lowercase and agents get real ids from list-devices/describe, but as written a value like `com.Apple.Preferences` would pass as injectable.

packages/tool-server/src/tools/native-devtools/native-devtools-status.ts
```diff
@@ -55,9 +60,25 @@ Fails if the simulator server is not running for the given UDID or the bundleId
...
     if (blocked) return blocked;
 
+    const injectable = isInjectableBundleId(params.bundleId);
     const appRunning = await api.isAppRunning(params.bundleId);
     const connected = api.isConnected(params.bundleId);
 
+    // Non-injectable apps (Apple system apps) can never load the dylib no matter
+    // how many times they relaunch. Report a terminal state so agents stop
+    // looping restart-app → retry: no restart is required and the next launch
+    // will not be injected either. envSetup/appRunning/connected stay measured.
```
hubgan: "envSetup/appRunning/connected stay measured" slightly overstates it for `envSetup`: on this branch it's read straight from the cached latch (`api.isEnvSetup()`), unlike the injectable path which reverifies when disconnected. No behavioral impact since nothing downstream uses it for a non-injectable app, but the comment reads as if `envSetup` were freshly measured here.

packages/tool-server/test/native-devtools-status.test.ts
```diff
@@ -145,8 +158,188 @@ describe("native-devtools-status tool", () => {
...
+    expect(NON_INJECTABLE_NATIVE_WARNING).toMatch(/Do not fall back to the native-\* tools/);
+    expect(NON_INJECTABLE_NATIVE_WARNING).toContain("native-view-at-point");
+    // The recommendation clause itself never names a native-* tool outside the
+    // warning, so nothing points the agent back at a dead-end.
+    const recommendationOnly = NON_INJECTABLE_RECOVERY.replace(NON_INJECTABLE_NATIVE_WARNING, "");
+    expect(recommendationOnly).not.toContain("native-");
+  });
+
+  it("all three surfaces carry the same dead-end warning verbatim", async () => {
+    // Finding #4: the precheck throw, the status description, and the describe
```
hubgan: "Finding #4" points at a review thread that won't mean anything to a future reader of the test. The invariant it describes (all three surfaces carry the same dead-end warning verbatim) stands on its own without the reference.

packages/tool-server/test/describe-tool.test.ts
```diff
@@ -255,6 +259,65 @@ describe("describe tool", () => {
...
+      { udid: "AAAAAAAA-AAAA-AAAA-AAAA-AAAAAAAAAAAA", bundleId: "com.apple.Preferences" }
+    );
+    expect(result.source).toBe("ax-service");
+    expect(result.should_restart).toBeUndefined();
+    expect(elementLineCount(result.description)).toBe(0);
+    expect(result.hint).toMatch(/system app/i);
+    // Reached only after describe's own AX path returned empty, so this hint
+    // leads with `screenshot` rather than re-recommending `describe`. It still
+    // carries the same native-* dead-end warning verbatim as the precheck throw
+    // and the native-devtools-status description (finding #4).
```
hubgan: Same here — the "(finding #4)" reference points at a review thread that won't be available to a later reader.

packages/tool-server/test/native-devtools-status.test.ts
```diff
@@ -145,8 +158,188 @@ describe("native-devtools-status tool", () => {
...
+  // The NATIVE_DEVTOOLS_NOT_INJECTABLE guard lives only in the shared precheck;
+  // every 3-arg feature tool relies on that throw propagating straight out of
+  // its execute() (none wraps the precheck in a try/catch). The precheck-level
+  // unit above proves the precheck throws, but not that each tool surfaces it —
+  // a later refactor that swallowed the throw inside a tool would leave that
+  // unit green while regressing the terminal behavior. Assert it at every tool
+  // boundary so that regression can't slip through.
+  const U = "44444444-4444-4444-4444-444444444444";
+  const SYSTEM_APP = "com.apple.Preferences";
+  const mkApi = () => makeNativeApi({ appRunning: true, connected: false }).api;
```
hubgan: In these propagation tests the `{ appRunning: true, connected: false }` config is inert: the non-injectable throw fires in the precheck before `appRunning`/`connected`/`requiresAppRestart` are consulted, so this mock state doesn't affect the assertion. Harmless, but it can read as if the restart logic were being exercised.

hubgan (commented):
Reviewed the diff and exercised it end-to-end against this branch's tool-server on iOS 26 simulators (iPad Pro 13" M5 and iPhone 17 Pro) plus an Android tablet. The behavior matches the intent: a `com.apple.*` app reports `injectable:false` with no `requiresRestart`, the native-* feature tools return the terminal error instead of "restart then retry", and `describe` never sets `should_restart` for a system app (confirmed both the empty-tree hint path and the real-tree path). Injectable apps are unaffected and Android is untouched. No correctness issues found.

A few smaller notes inline. The one worth acting on is the missing test coverage for a system app with a non-empty accessibility tree; the rest are comment/doc-accuracy nits.

packages/tool-server/test/describe-tool.test.ts
latekvo: Fixed in 756a1919.

Added a regression test for the populated-tree path: `describe com.apple.Preferences` with a NON-empty ax-service tree now asserts it returns the real tree (`source: ax-service`, the `General` element present, `should_restart` unset, no non-injectable hint) via the `tree.children.length > 0` early return, before the injectability gate. To confirm the guard is meaningful I temporarily hoisted the gate above the early return (returning an empty tree + the screenshot hint); the new test failed on the missing `General` element, then reverted and it passes.

packages/tool-server/src/blueprints/native-devtools.ts
latekvo: Fixed in 756a1919.

Replaced the incorrect "bundle ids are case-sensitive" rationale and made the check case-insensitive (`bundleId.toLowerCase().startsWith("com.apple.")`), so a stray `com.Apple.Preferences` is treated as the system app it is rather than slipping through as injectable and restart-looping. Added mixed-case assertions (`com.Apple.Preferences`, `COM.APPLE.PREFERENCES` → non-injectable); exact-prefix lookalikes like `com.appleseed.App` stay injectable under case-folding. Verified the new assertions fail against the old case-sensitive check.

packages/tool-server/src/blueprints/native-devtools.ts
latekvo: Fixed in 756a1919.

Right — `native-devtools-status` is a native-* tool that reports `injectable: false` rather than re-throwing. The jsdoc now scopes the claim to native-* *feature* tools and calls out `native-devtools-status` as the lone exception (2-arg precheck, reports instead of throws).

packages/tool-server/src/tools/native-devtools/native-devtools-status.ts
latekvo: Fixed in 756a1919.

Reworded so the comment no longer implies `envSetup` is freshly measured on the non-injectable path. It now notes `envSetup` is read from the cached latch (unlike the injectable path, which reverifies when disconnected), since there is no point reverifying the env for an app that can never inject.

packages/tool-server/test/native-devtools-status.test.ts
latekvo: Fixed in 756a1919.

Dropped the "Finding #4:" prefix — the invariant (all three surfaces carry the same dead-end warning verbatim) stands on its own.

packages/tool-server/test/describe-tool.test.ts
latekvo: Fixed in 756a1919.

Dropped the "(finding #4)" reference from the comment.

packages/tool-server/test/native-devtools-status.test.ts
latekvo: Fixed in 756a1919.

Switched the propagation tests to `makeNativeApi({})` (default state) with a note that the non-injectable throw fires before appRunning/connected/requiresAppRestart are consulted, so the mock device state is inert here. All six propagation tests still pass.

packages/tool-server/src/blueprints/native-devtools.ts
```diff
@@ -81,6 +138,24 @@ export async function precheckNativeDevtools(
     });
   }
 
+  // Terminal case: an app that can never be injected (Apple system app). Throw
+  // instead of returning a restart-required block so the native-* feature tools
+  // surface a hard error rather than instructing an unbounded restart→retry
+  // loop that can never succeed. The 2-arg overload (bundleId undefined), used
+  // by native-devtools-status, must NOT throw — it reports the state instead.
+  if (bundleId !== undefined && !isInjectableBundleId(bundleId)) {
```
hubgan: This terminal check runs after `ensureEnvReady()` (line 130). For a `com.apple.*` bundle, injectability is knowable without any env state, but if `ensureEnvReady()` throws transiently the precheck returns the `init_failed` block ("try shutting down and re-booting the simulator") before this branch is ever reached. For a system app that guidance is misleading - a reboot can never make it injectable - and the env-setup work is wasted, since these apps can never load the dylib regardless of env. The masking is bounded (the attempt cap stops it looping), so this is minor, but the terminal signal is the more accurate thing to surface for a system app.

packages/tool-server/src/blueprints/native-devtools.ts
```diff
@@ -81,6 +138,24 @@ export async function precheckNativeDevtools(
     });
   }
 
+  // Terminal case: an app that can never be injected (Apple system app). Throw
+  // instead of returning a restart-required block so the native-* feature tools
+  // surface a hard error rather than instructing an unbounded restart→retry
+  // loop that can never succeed. The 2-arg overload (bundleId undefined), used
+  // by native-devtools-status, must NOT throw — it reports the state instead.
```
hubgan: This comment names `native-devtools-status` as the 2-arg consumer, but `launch-app` (platforms/ios.ts:34, platforms/shared.ts:17) and `restart-app` (platforms/ios.ts:35, platforms/shared.ts:18) also call the 2-arg overload - and they do not "report the state", they go on to launch/restart the app. The load-bearing reason the 2-arg path must not throw is broader than status's reporting: launching or restarting an Apple system app is legitimate (env gets set up, the app just won't inject). As written, someone could read this as "status is the only 2-arg caller" and tighten the guard, which would break system-app launch/restart. (The sibling jsdoc at lines 50-51 is scoped correctly.)

packages/tool-server/test/native-devtools-status.test.ts
```diff
@@ -145,8 +158,196 @@ describe("native-devtools-status tool", () => {
...
+    expect(NON_INJECTABLE_RECOVERY).toContain(NON_INJECTABLE_NATIVE_WARNING);
+    expect(NON_INJECTABLE_NATIVE_WARNING).toMatch(/Do not fall back to the native-\* tools/);
+    expect(NON_INJECTABLE_NATIVE_WARNING).toContain("native-view-at-point");
+    // The recommendation clause itself never names a native-* tool outside the
+    // warning, so nothing points the agent back at a dead-end.
+    const recommendationOnly = NON_INJECTABLE_RECOVERY.replace(NON_INJECTABLE_NATIVE_WARNING, "");
+    expect(recommendationOnly).not.toContain("native-");
+  });
+
+  it("all three surfaces carry the same dead-end warning verbatim", async () => {
```
hubgan: The title says "all three surfaces carry the same dead-end warning verbatim", but this test only exercises two of them (the precheck throw and the status description); the describe fallback hint's copy is asserted separately in describe-tool.test.ts. All three surfaces are in fact covered across the two files, so there is no coverage gap - just flagging that the title claims a three-way equality that no single assertion in this file actually makes.

hubgan (commented):
A few minor, non-blocking notes inline.

hubgan (approved):

packages/tool-server/src/blueprints/native-devtools.ts
latekvo: Fixed in ca7093c6.

Hoisted the non-injectable terminal check to the top of the precheck, above both the `givenUp` early-return and `ensureEnvReady()` — injectability is a static property of the bundle id, so no env state is needed to decide it. A broken env can no longer mask the terminal signal behind `init_failed`'s re-boot guidance, and no env-setup work is spent on a system app.

Reproduced both masked paths first (given-up latch and a transiently-throwing `ensureEnvReady`, each with `com.apple.Preferences` — the precheck returned `init_failed` instead of throwing), then re-ran after the hoist: both now throw `NATIVE_DEVTOOLS_NOT_INJECTABLE`, and a new assertion pins that `ensureEnvReady` is never called for a non-injectable bundle. Both stay as regression tests.

packages/tool-server/src/blueprints/native-devtools.ts
latekvo: Fixed in ca7093c6.

The relocated comment now names all three 2-arg consumers and the broader reason the 2-arg path must not throw: `native-devtools-status` reports the state instead, and `launch-app` / `restart-app` run it too — launching or restarting a system app is legitimate, it just never injects. Also retitled the 2-arg precheck test from "(status path)" to "(status / launch-app / restart-app path)" so the tests don't repeat the same understatement.

packages/tool-server/test/native-devtools-status.test.ts
latekvo: Fixed in ca7093c6.

Retitled to "the precheck throw and the status description share the recovery guidance verbatim" — the two surfaces this test actually asserts — with the comment noting the third surface (the describe fallback hint) is asserted in describe-tool.test.ts. No assertion changes, since as you note all three surfaces are already covered across the two files.

PR #460, opened 2026-07-02
fix(cli): print help for init/update/uninstall instead of running them

### LINKED ISSUE #451: `<cmd> --help` runs the command instead of printing help for init/update/uninstall (uninstall --help opens the destructive prompt)
## Summary

`--help` / `-h` is **not** intercepted for the `init`, `update`, and `uninstall` subcommands. Instead of printing usage, the flag is passed straight through to the subcommand, which then **runs its real (side-effecting) flow**. Most seriously, `argent uninstall --help` opens the destructive *"Remove argent configuration from this workspace?"* prompt with **`Yes` pre-selected** — a user who types `--help` expecting docs and reflexively hits Enter can wipe their workspace config. In this session only a non-TTY stdin prevented the mutation from completing.

## Environment
- @swmansion/argent 0.13.0
- Node v24.16.0, macOS (darwin)

## Root cause
`dist/cli.js` dispatches on the command in a `switch`, and only the **top-level** `--help`/`-h` case calls `printHelp()`. For subcommands, the remaining args (including `--help`) are forwarded verbatim:

```js
case "update":            return (await loadInstaller()).update(rest);
case "uninstall":
case "remove":            return (await loadInstaller()).uninstall(rest);
case "init":
case "install":           return (await loadInstaller()).init(rest);
...
case "--help":
case "-h":
default:  printHelp(); ...   // only reached when --help is the top-level command
```

The installer subcommands don't short-circuit `--help` themselves, so it becomes a no-op flag and the command executes.

## Reproduction

```console
$ argent run --help          # CONTROL — correctly prints usage
Usage: argent run <tool> [flags]
Invoke a tool exposed by the argent tool-server. ...

$ argent update --help       # expected: help. actual: runs the update flow
┌   argent update
◇  Version check complete.
●  Installed: v0.13.0
●  Latest:    v0.13.0
◇  Configuration refreshed.       <-- real side effect
└  Update complete.

$ argent init --help         # expected: help. actual: launches the config wizard
┌   argent init
◇  Step 1: MCP Server Configuration
◆  Which editors should Argent be configured for?   <-- interactive multi-select

$ argent uninstall --help    # expected: help. actual: destructive prompt, Yes pre-selected
┌   argent uninstall
◆  Remove argent configuration from this workspace?
│  ● Yes / ○ No
```

(`init`/`uninstall` only aborted here because stdin was not a TTY; on a real terminal they proceed.)

## Expected
`--help`/`-h` is read-only for **every** subcommand and prints that subcommand's usage without any network call, config mutation, wizard, or destructive prompt.

## Suggested fix
Short-circuit `--help`/`-h` centrally in `cli.js` *before* dispatching to any subcommand (or have each installer subcommand check for it first). Given `uninstall` is destructive, this should be treated as a safety fix.

latekvo:
## The bug

`argent <cmd> --help` for the installer subcommands `init` / `install` / `update` / `uninstall` / `remove` did not print help — it ran the real, side-effecting command. `argent uninstall --help` even opened the destructive "Remove argent configuration from this workspace?" prompt (with Yes pre-selected).

Root cause: `main()` in `packages/argent/src/cli.ts` dispatched these subcommands and forwarded `rest` (including `--help`) to the installer functions, which have no help handling. Only the top-level `--help` case called `printHelp()`.

## The fix

Intercept `--help`/`-h` for exactly the installer command set before dispatching, and print a short per-subcommand usage block without loading or invoking any installer code (no network, no wizard, no prompt).

- New side-effect-free module `packages/argent/src/installer-help.ts` exports the pure decision function `installerHelpRequested(command, rest)` plus `printInstallerHelp(command)`.
- `cli.ts` calls the guard at the top of `main()` and returns early when help is requested.
- The other subcommands (`run`, `tools`, `server`, `link`, `flags`, `enable`, `disable`, `telemetry`) already handle `--help` themselves, so they are deliberately excluded — `argent run <tool> --help` still prints that tool's flags.

## Verification

- `npm run build` (root tsc) + `npm run build -w @swmansion/argent` — green.
- Ran the built CLI end-to-end (stdin from /dev/null): `init --help`, `update --help`, `uninstall --help`, `install --help`, `remove --help`, `uninstall -h` all print usage and exit 0 with no wizard/network/prompt.
- Regression: `argent --help` still prints top-level help; `argent run --help` still prints the `run` usage (not intercepted).
- Added unit tests in `packages/argent/test/installer-help.test.ts` for `installerHelpRequested` / `isInstallerCommand` (8 tests, all pass).
- `prettier` and `eslint` clean on the touched files; test typecheck passes.

Fixes #451


hubgan (commented):
Reviewed the installer-help short-circuit and exercised it end-to-end against the built CLI: the five installer subcommands now print usage and exit 0 without running, `argent uninstall --help` and `remove --help` leave an existing workspace's config byte-for-byte unchanged (no prompt, nothing deleted), and `run` / `tools` / `server` still handle their own `--help` unaffected. I also confirmed the pre-fix behaviour — `uninstall` derives non-interactivity only from `--yes`, so `--help` previously fell through to the destructive confirm. The fix is correct; the inline notes are mostly about test coverage of the dispatch path.

packages/argent/src/cli.ts
```diff
@@ -103,6 +104,17 @@ async function loadCli(): Promise<typeof Cli> {
 }
 
 async function main(): Promise<void> {
+  // The installer subcommands (init / install / update / uninstall / remove)
+  // forward their argv straight to the side-effecting installer functions,
+  // which do not short-circuit on `--help` — so `argent uninstall --help`
+  // would run the real (destructive) command. Intercept help for exactly that
+  // set before dispatching. All other subcommands handle `--help` themselves.
+  if (installerHelpRequested(command, rest)) {
```
hubgan, line 128: This guard is the actual protection against the destructive path, yet nothing tests it — the unit tests cover only the `installerHelpRequested` predicate, not this dispatch. Moving this block below the `switch`, dropping the `return`, or removing it would keep every existing test green while re-opening the exact bug the PR fixes (`argent uninstall --help` running the real uninstall). The behaviour with the highest regression cost is the one with no test around it.

packages/argent/src/installer-help.ts
```diff
@@ -0,0 +1,77 @@
...
+    usage: "argent remove [options]",
+    description: "Alias for `argent uninstall`.",
+  },
+};
+
+/**
+ * Print the short usage block for an installer subcommand. Read-only: no
+ * network, no wizard, no prompt.
+ */
+export function printInstallerHelp(command: InstallerCommand): void {
```
hubgan, line 213: `printInstallerHelp` itself has no test. Its contract is to print usage with no network, no wizard and no prompt, but neither the per-command output nor the absence of side effects is asserted, so a change that broke the wording or reintroduced a side effect here would pass unnoticed.

packages/argent/src/installer-help.ts
```diff
@@ -0,0 +1,77 @@
...
+  const { usage, description } = INSTALLER_USAGE[command];
+  console.log(`
+Usage: ${usage}
```
hubgan: The printed help only ever shows `argent <cmd> [options]` and never lists the options. `argent init --help` won't surface `--yes` / `--no-telemetry` / `--from`, and `argent uninstall --help` won't surface `--yes`, so the help doesn't actually tell the user which flags the command accepts.

packages/argent/src/installer-help.ts
```diff
@@ -0,0 +1,77 @@
...
+export const INSTALLER_COMMANDS = ["init", "install", "update", "uninstall", "remove"] as const;
+
+export type InstallerCommand = (typeof INSTALLER_COMMANDS)[number];
+
+export function isInstallerCommand(command: string | undefined): command is InstallerCommand {
+  return INSTALLER_COMMANDS.includes(command as InstallerCommand);
+}
+
+function isHelpFlag(arg: string): boolean {
+  return arg === "--help" || arg === "-h";
```
hubgan: Help is matched exactly on `--help` / `-h`, so `argent uninstall --HELP`, `argent uninstall --help=foo`, and the bareword `argent uninstall help` are not treated as help and still forward to the real, destructive uninstall (opening the removal prompt). `--help=foo` is a plausible fat-finger on a command that deletes workspace config. This mirrors the top-level `--help` handling, which is also exact-match.

packages/argent/src/installer-help.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+export function installerHelpRequested(command: string | undefined, rest: string[]): boolean {
+  return isInstallerCommand(command) && rest.some(isHelpFlag);
+}
+
+/** One-line usage + description for each installer subcommand. */
+const INSTALLER_USAGE: Record<InstallerCommand, { usage: string; description: string }> = {
+  init: {
+    usage: "argent init [options]",
+    description: "Initialize argent in the current workspace (MCP server + skills + rules).",
```
hubgan: These per-command descriptions duplicate the ones in the top-level `printHelp` in cli.ts — for example init reads "MCP server + skills + rules" here versus "MCP server + skills" there, and uninstall reads "from the current workspace" versus "from the workspace". Two independent copies of the same descriptions tend to drift apart over time.

packages/argent/src/cli.ts
latekvo: Fixed in 7daf6bd5. Added `packages/argent/test/cli-dispatch.test.ts`, a spawn-based E2E that transpiles the real dispatcher (its workspace imports are `import type`, so they erase — no workspace build needed) and stages fake lazily-loaded bundles which record if they run. It asserts `argent uninstall --help` prints usage and never loads the installer, while the control `argent uninstall` does — so dropping the `return`, moving the block below the `switch`, or removing it fails the suite. Verified by temporarily dropping the `return`: 4/5 cases go red, control stays green.

packages/argent/src/installer-help.ts
latekvo: Fixed in 7daf6bd5. Added a `printInstallerHelp` describe block that captures `console.log` and asserts the per-command output for every command — usage line, summary, the Options section (or the alias pointer for install/remove), and the footer — plus that its only observable effect is a single `console.log` (returns undefined, nothing else). The end-to-end side-effect guarantee (workspace config left byte-for-byte untouched) is covered by the new dispatch E2E in cli-dispatch.test.ts.

packages/argent/src/installer-help.ts
latekvo: Fixed in 7daf6bd5. `INSTALLER_COMMAND_META` now carries each command's real flags and `printInstallerHelp` renders an Options section: `init --help` lists `--yes, -y` / `--no-telemetry` / `--from <path>`, `update --help` lists `--yes, -y` / `--no-telemetry` / `--version <version>`, `uninstall --help` lists `--yes, -y`. The flags mirror exactly what init/update/uninstall parse. Aliases (install/remove) point at their target's `--help` rather than duplicating the list.

packages/argent/src/installer-help.ts
latekvo: Fixed in 7daf6bd5. `isHelpFlag` is now case-insensitive and also matches `--help=<value>`, and `installerHelpRequested` treats a bareword `help` in first position as help. So `argent uninstall --HELP`, `argent uninstall --help=foo`, and `argent uninstall help` all short-circuit to usage instead of the removal prompt (verified end-to-end). The bareword is honoured only in first position so a flag value like `--from help` isn't mistaken for a help request. I left the top-level `--help` as exact-match on purpose — that path is non-destructive (it just prints the top-level help), so the footgun is specific to the installer commands.

packages/argent/src/installer-help.ts
latekvo: Fixed in 7daf6bd5. The summaries now live once in `INSTALLER_COMMAND_META`; the top-level `printHelp` table in cli.ts interpolates `INSTALLER_COMMAND_META.<cmd>.summary`, so there is a single copy and the two can't drift. I reconciled the existing drift onto the accurate wording — init does install rules (Step 3), so `MCP server + skills + rules`, and `from the current workspace`.

packages/argent/src/installer-help.ts
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * Help is recognised from a help flag anywhere in `rest` (`--help`, `-h`,
+ * `--help=…`, case-insensitively) or the bareword `help` as the first argument
+ * (`argent uninstall help`). The bareword is only honoured in first position so
+ * a flag value such as `--from help` isn't mistaken for a help request.
+ */
+export function installerHelpRequested(command: string | undefined, rest: string[]): boolean {
+  if (!isInstallerCommand(command)) return false;
+  if (rest.some(isHelpFlag)) return true;
+  return rest[0]?.toLowerCase() === "help";
```
hubgan: A help flag is recognized anywhere in `rest`, but the bareword `help` only in first position. That asymmetry leaves one path that still reaches the real uninstall: `argent uninstall --yes help` (and `-y help`) evaluates to `false` here, so it falls through to the installer. Since `--yes`/`-y` sets `nonInteractive` in uninstall.ts and skips the `p.confirm` at "Remove argent configuration from this workspace?", the config is pruned with no prompt and no help. It's a narrow case: it needs `--yes` before a stray `help`, and `uninstall help --yes` is caught correctly. But it's the destructive command, so worth being aware of.

packages/argent/src/installer-help.ts
```diff
@@ -0,0 +1,160 @@
...
+export function isInstallerCommand(command: string | undefined): command is InstallerCommand {
+  return INSTALLER_COMMANDS.includes(command as InstallerCommand);
+}
+
+/**
+ * True when `arg` is an unambiguous help flag. Matched leniently — these guard
+ * destructive installer commands, so a fat-fingered `--help=foo` or an
+ * upper-case `--HELP` must still short-circuit to usage rather than fall
+ * through to the real (config-deleting) command.
+ */
```
hubgan, line 38: This comment says a fat-fingered help request "must still short-circuit to usage rather than fall through to the real (config-deleting) command," but a few plausible fat-fingers don't: `-help` (single dash), `—help` (em-dash, common from autocorrect or paste), and `/help` all return `false` and fall through. They stay interactive, so the confirmation prompt still guards the user, but the actual leniency is narrower than the comment describes.

packages/argent/src/installer-help.ts
```diff
@@ -0,0 +1,160 @@
...
+  description: "Opt out of anonymous telemetry for this run.",
+};
+
+/**
+ * How each installer subcommand is described in help — its summary (the sole
+ * copy, also rendered by the top-level table in cli.ts), usage, and options.
+ * The option lists are hand-maintained to mirror the flags each installer
+ * actually parses (see packages/argent-installer/src/{init,update,uninstall}.ts);
+ * keep them in sync when those flags change — nothing links them automatically.
+ */
```
hubgan, line 127: As this comment notes, nothing links these option lists to the real parsers. They match today (init: `--yes`/`-y`, `--no-telemetry`, `--from`; update: `--yes`/`-y`, `--no-telemetry`, `--version`; uninstall: `--yes`/`-y`), but no test asserts the table against `init`/`update`/`uninstall.ts`, so a later flag change there can silently make `--help` under- or over-report the real flags without any test going red.

packages/argent/test/cli-dispatch.test.ts
```diff
@@ -0,0 +1,178 @@
...
+function record(name) {
+  const marker = process.env.ARGENT_E2E_MARKER;
+  if (marker) writeFileSync(marker, name);
+}
+export function init() { record("init"); }
+export function update() { record("update"); }
+export function uninstall() { record("uninstall"); }
+`;
+
+const FAKE_NOOP_BUNDLE = [
```
hubgan: Because the fake `run`/`tools` exports write no marker, this harness can't assert the complementary guarantee, that a non-installer `--help` is forwarded to its own bundle rather than intercepted (e.g. `argent run <tool> --help` still reaching `run`). The predicate is unit-tested separately, but the dispatch wiring for the non-intercepted commands, and the unknown-command `exit(1)` / top-level `--help` arms right below the guard, aren't exercised here.

packages/argent/test/installer-help.test.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  it("only writes to stdout (no wizard / prompt / network)", () => {
+    // The whole contract of the help path is that it is side-effect-free. This
+    // asserts the function's sole observable effect is console.log; the
+    // end-to-end guarantee (config left untouched) is covered in
+    // cli-dispatch.test.ts.
+    const spy = vi.spyOn(console, "log").mockImplementation(() => {});
+    const result = printInstallerHelp("uninstall");
+    expect(result).toBeUndefined();
+    expect(spy).toHaveBeenCalledTimes(1);
```
hubgan: `toHaveBeenCalledTimes(1)` ties this test to the implementation detail that the output is emitted as a single `console.log` call. A refactor that produced identical output line-by-line would fail it despite no observable change. The `result` being `undefined` and the absence of a prompt are the assertions that capture the actual contract.

hubgan (commented):
Reviewed and verified this end-to-end: built the CLI and ran every installer `--help` form against the real dispatcher. Each prints usage and exits 0 without loading the installer bundle, and `uninstall --help` in a workspace populated with argent config leaves every file untouched. The option lists in `INSTALLER_COMMAND_META` match the real `init`/`update`/`uninstall` parsers, and non-installer commands (`run`/`tools`) still handle their own `--help`. A few low-severity notes inline.

hubgan (approved):

packages/argent/src/installer-help.ts
latekvo: Fixed in 59008a01. The bareword `help` now counts in any position unless it directly follows a value-taking flag (`--from` / `--version` / `--project-root`, kept per-command in `VALUE_TAKING_FLAGS`), so `uninstall --yes help` and `-y help` short-circuit to usage while `init --from help` still reaches the installer. Reproduced the fallthrough end-to-end before the fix (the staged dispatcher recorded `uninstall:--yes help` reaching the installer) and re-ran after — the installer is never loaded and usage prints. Also ran the real built CLI against a workspace populated with argent config: byte-for-byte untouched.

packages/argent/src/installer-help.ts
latekvo: Fixed in 59008a01, in both directions: `isHelpFlag` now also accepts single-dash `-help` and the smart-dash `—help` / `–help` spellings (a leading em/en dash normalizes to `--`), and the comment now enumerates exactly what is matched and states that anything else (`/help`, `--helpme`) falls through to the command — where the interactive confirmation still guards uninstall unless `--yes` was passed.

packages/argent/src/installer-help.ts
latekvo: Fixed in 59008a01. Added test/installer-flags-sync.test.ts: it reads the installer parser sources (init-args.ts, update.ts, uninstall.ts, install-targets.ts) and extracts every flag they consult through the parsing idioms in use (`args.includes`, `extractFlag`, and the `--flag` / `--flag=` lookahead pairs), then asserts set-equality with the META option lists — plus an explicit internal-flag allowlist for update's agent-internal `--project-root`. Verified it trips: reverting init's META to the pre-#428 list (no `--global`/`--local`) fails the test, which is exactly the drift the merge with main had just introduced; the meta now documents those flags.

packages/argent/test/cli-dispatch.test.ts
latekvo: Fixed in 59008a01. The fake bundles now all record which export ran and with which argv, so the harness asserts the complementary guarantees too: `run gesture-tap --help` reaches the run bundle with argv intact, `tools --help` reaches tools, top-level `--help` prints the table without loading any bundle, and an unknown command prints help and exits 1. Sensitivity mutation-checked: adding `run` to INSTALLER_COMMANDS fails the forwarding test.

packages/argent/test/installer-help.test.ts
latekvo: Fixed in 59008a01. The test no longer pins the console.log call count — it asserts the undefined return and that output was written, and the comment now says why the batching is out of scope.

PR #462, opened 2026-07-03
feat(tool-server): settings-permissions tool to grant/deny/reset app permissions

hubgan:
## What

Adds a `settings-permissions` MCP tool that edits an app's runtime permissions **out-of-band**, without navigating the system Settings UI. It's a test-setup / out-of-band tool — the default way to change a permission is still through the app (in-app toggle or the app-triggered dialog). Reach for this only when the app can't get there:

- **pre-authorize** (`grant`) or **deny** a permission before the app asks, so no dialog interrupts the flow
- **re-enable** one the user already denied (iOS never re-prompts once denied)
- **reset** it so the first-run dialog reappears on next launch

Supported permissions: camera, microphone, photos, contacts, notifications, calendar, location, location-always, media-library, motion, reminders.

## How

**iOS simulator** — `xcrun simctl privacy grant|revoke|reset <service> [bundleId]`.
- `notifications` has no simctl service → surfaced as an unsupported error.
- An "invalid service" failure (e.g. `camera` on an Xcode that doesn't model it) is wrapped as unsupported with a hint to list supported services.
- A shutdown simulator error carries a `boot-device` hint.

**Android** — `pm grant` / `pm revoke` (plus `clear-permission-flags` on reset) over adb.
- One abstract permission fans out to the concrete `android.permission.*` set (fine/coarse location, per-media reads); `location-always` grant also grants foreground location since background alone is unusable.
- A `pm path` preflight rejects an uninstalled package (`pm grant`/`revoke` exit 0 for a missing package, which would otherwise report a false success). A transport-level adb failure is rethrown rather than mislabeled as "not installed" (`isTerminalAdbError` is now exported for this).
- Partial results land in `skipped`; the action fails only when pm rejects **every** mapped permission.

## Supporting changes

- New failure codes: `ANDROID_SETTINGS_PERMISSION_FAILED`, `IOS_SETTINGS_PERMISSION_FAILED`, `SETTINGS_PERMISSION_UNSUPPORTED`.
- New `argent-settings-permissions` skill, plus cross-references from `argent-device-interact`, `argent-test-ui-flow`, and the argent rule (with a strong "prefer the app dialog over this tool" steer).
- `bundleId` is regex-validated (same shell-injection guard as launch-app/restart-app) and schema-required for grant/deny.

## Testing

`vitest` — **37/37 pass** (schema validation incl. shell-injection payloads, iOS branch, Android branch incl. partial/total pm failures and the preflight). `tsc --noEmit` clean for tool-server + registry + tests; prettier clean.


packages/tool-server/src/tools/settings-permissions/platforms/android.ts
```diff
@@ -0,0 +1,219 @@
...
+        result = await runPm(udid, `revoke ${pkg} ${perm}`);
+      } else {
+        // reset = revoke + drop the user-set/user-fixed flags so the app is
+        // back to "not asked yet" and the next request shows the dialog. Both
+        // steps must succeed for the perm to count as reset: a revoke whose
+        // flags survive leaves the app "user-denied" (no dialog on next ask),
+        // which is a deny, not a reset — report it in `skipped`, not `applied`.
+        result = await runPm(udid, `revoke ${pkg} ${perm}`);
+        if (result.ok) {
+          result = await runPm(udid, `clear-permission-flags ${pkg} ${perm} user-set user-fixed`);
```
latekvo: `reset` runs `pm revoke` and then, only if that succeeds, `pm clear-permission-flags` — two separate adb calls, with a permission counted as applied only when both succeed. A permission whose revoke succeeds but whose flag-clear fails has therefore already been revoked on the device, yet it is recorded as a failure. Two consequences:

- When the flag-clear fails for every mapped permission, `applied` is empty and the call throws the "pm rejected every mapped permission … the app must be installed and declare the permission in its manifest" error below. But pm did not reject anything — the revoke succeeded — and the manifest/install advice is wrong for a flag-clear failure. The app is left denied while the caller is told nothing changed.
- When it fails for only some permissions, the call returns success with those permissions in `skipped`, which the result type documents as permissions pm rejected / not declared in the manifest. Here they were accepted and revoked, so a caller reading `skipped` believes them untouched when they were actually left denied.

`clear-permission-flags` does not exist before Android 11 — it is absent from the API-29 AOSP `PackageManagerShellCommand`, and an unknown pm subcommand exits non-zero with "Unknown command". So on Android 10 and earlier every `reset` revokes and then reports failure, leaving the permission denied rather than reset. The tool's `capability` declares no minimum SDK, so those devices are in scope.

packages/tool-server/src/tools/settings-permissions/platforms/android.ts
```diff
@@ -0,0 +1,219 @@
...
+    // `package:...` for an installed app and exits non-zero otherwise. A
+    // transport-level failure (device offline / unauthorized / not found) is
+    // NOT a "not installed" verdict — rethrow adb's own error so the caller
+    // sees the real cause instead of a confidently wrong diagnosis.
+    let installed: string;
+    try {
+      installed = await adbShell(udid, `pm path ${pkg}`, { timeoutMs: 15_000 });
+    } catch (err) {
+      const message = err instanceof Error ? err.message : String(err);
+      if (isTerminalAdbError(message)) throw err;
```
latekvo: The `pm path` preflight rethrows only errors matching `isTerminalAdbError` (device unauthorized / not found / offline / no devices); every other thrown error falls through to `installed = ""` and is reported as "Package `<id>` is not installed." A 15s `pm path` timeout (SIGKILL, message ending `(killed=true signal=SIGKILL)`) and a package manager that is still initializing right after boot ("Could not access the Package Manager. Is the system running?") both fail to match those patterns, so a slow or transiently-unavailable device is told its app is uninstalled and the caller is pointed at the wrong remediation. This partly defeats the intent stated in the comment just above (avoid a "confidently wrong diagnosis").

packages/tool-server/src/tools/settings-permissions/platforms/ios.ts
```diff
@@ -0,0 +1,118 @@
...
+        },
+        { cause: err instanceof Error ? err : new Error(String(err)) }
+      );
+    }
+
+    return {
+      action,
+      permission,
+      ...(bundleId ? { bundleId } : {}),
+      applied: [service],
```
latekvo: On success `applied` is set to `[service]` unconditionally, but a device-wide `reset` (no `bundleId`) does not actually clear existing per-app TCC entries on recent iOS. Reproduced on iOS 18.5: after granting `camera` to two apps, `simctl privacy <udid> reset camera` with no bundle exits 0 but leaves every per-app row intact, while the tool returns `applied:["camera"]`; a per-app `reset camera <bundleId>` correctly removes the single row. So on the device-wide `reset` path the caller gets a truthful-looking `applied` for a change that did not happen. Unlike the Android branch, the iOS branch never inspects the resulting state — the prose warns about this runtime behavior, but the return value still reports success.

packages/tool-server/src/tools/settings-permissions/platforms/ios.ts
```diff
@@ -0,0 +1,118 @@
...
+    } catch (err) {
+      const detail = err instanceof Error ? err.message : String(err);
+      // Distinguish "this Xcode's simctl doesn't model that service" (a
+      // capability gap worth naming) from an ordinary simctl failure. The
+      // pattern is a best-effort heuristic: the rejection text comes from a
+      // CoreSimulator NSError whose exact wording we could not pin on an Xcode
+      // that rejects a service (this host's accepts everything we map, camera
+      // included). A miss just falls through to the generic branch with
+      // simctl's own text — nothing is lost, only the nicer hint.
+      const unsupportedService = /invalid.*service|unknown.*service/i.test(detail);
```
latekvo: This heuristic classifies a rejected simctl service with `/invalid.*service|unknown.*service/i`, but that wording does not appear in simctl's actual output. A service simctl will not accept produces `Simulator device failed to complete the requested operation. / Operation not permitted / Failed to set access` (NSPOSIXErrorDomain), which the regex does not match — so the friendly hint and the `SETTINGS_PERMISSION_UNSUPPORTED` code never fire and the error always buckets as a generic `IOS_SETTINGS_PERMISSION_FAILED`. The unit test covering this branch feeds a fabricated `"Invalid privacy service: camera"` string, so it asserts a classification that cannot occur against real simctl. The generic fallback still surfaces simctl's own text, so it is behaviorally benign, but the special-case branch and its test give false confidence.

packages/tool-server/test/settings-permissions.test.ts
```diff
@@ -0,0 +1,427 @@
...
+  it("denying location-always touches only the background permission", async () => {
+    const result = await androidImpl.handler(
+      {},
+      params({ action: "deny", permission: "location-always" }),
+      androidDevice
+    );
+    expect(result.applied).toEqual(["android.permission.ACCESS_BACKGROUND_LOCATION"]);
+  });
+
+  it("partial pm failures land in `skipped`, not an error", async () => {
```
latekvo: These pm-failure tests mock `adbShell` to resolve the `SecurityException` text as stdout with a zero exit, but on a real device (verified on API 34) `pm grant` for a non-declared permission exits 255 and writes the exception to stderr — so `adbShell` throws and the failure is handled in `runPm`'s catch branch, not the stdout-inspection branch these tests drive. The realistic (thrown) "partial failure lands in `skipped`" behavior is therefore not exercised; the one throw-path test asserts only the error code on a total failure. The stack-frame-stripping assertion here also runs against a resolved-stdout string rather than the thrown `err.message` where the stripping actually happens in production.

packages/tool-server/src/tools/settings-permissions/index.ts
```diff
@@ -0,0 +1,81 @@
...
+Permissions: camera, microphone, photos, contacts, notifications, calendar, location, location-always, media-library, motion, reminders.
+iOS simulator: runs \`xcrun simctl privacy\`. \`notifications\` is not supported (no such simctl service). \`reset\` with no bundleId asks simctl to reset the service for ALL apps — but on recent iOS runtimes that can leave existing per-app entries untouched, so prefer a per-app reset with bundleId.
+Android: runs \`pm grant\` / \`pm revoke\` (reset also clears the user-set permission flags) on the mapped \`android.permission.*\` runtime permissions. The app must be installed and declare them in its manifest; \`reminders\` has no Android equivalent; bundleId is always required.
+Some permission changes terminate the app if it is running (system behavior on both platforms) — set permissions before launching, or relaunch after.
+Returns { action, permission, bundleId, applied, skipped? }: \`applied\` lists the platform-level services/permissions actually changed; \`skipped\` (Android) lists mapped permissions pm rejected, e.g. ones the manifest doesn't declare. Fails if nothing could be applied.`,
```
latekvo: The description states the tool returns `{ action, permission, bundleId, applied, skipped? }`, but `bundleId` is omitted from the result on an iOS device-wide `reset` (no bundleId supplied), where the handler spreads it only when present. The SKILL doc writes it as `bundleId?`; this description implies it is always echoed back.

latekvo (commented):
Reviewed the new `settings-permissions` tool end-to-end — built the branch and ran the real iOS and Android handlers against an iOS 18.5 simulator and an API-34 Android emulator, plus the schema, docs, and test suite. It is well structured and the common grant/deny/per-app paths work as intended (verified observable state changes on both platforms). Inline notes below, mostly around reset semantics on each platform and the accuracy of a few error diagnoses.

packages/tool-server/src/utils/adb.ts
```diff
@@ -578,7 +578,14 @@ const TERMINAL_ADB_ERROR_PATTERNS: RegExp[] = [
...
 
-function isTerminalAdbError(message: string): boolean {
+/**
+ * True when an adb error message names a terminal device state (unauthorized /
+ * not found / offline / no devices) — a transport-level condition no retry or
+ * alternate interpretation fixes. Exported so callers that probe a device can
+ * distinguish "the command failed because the device is unreachable" from "the
+ * command ran and answered negatively".
+ */
+export function isTerminalAdbError(message: string): boolean {
```
latekvo, line 618: `isTerminalAdbError` is widened from module-private to `export`ed here, but no module imports it. The Android settings-permissions preflight uses only `adbShell` and `shellQuote` and lets adb errors propagate unchanged rather than classifying them, so it never calls this function. The commit message and PR description say the export exists so the preflight can tell a transport-level adb failure from a genuine "not installed" verdict — but the final `pm list packages` preflight doesn't do that classification, which leaves the widened visibility and the new JSDoc (addressed to "callers that probe a device") without a consumer.

packages/tool-server/src/tools/settings-permissions/platforms/ios.ts
```diff
@@ -0,0 +1,116 @@
...
+};
+
+// Tool permission → `simctl privacy` service. Identity for everything simctl
+// models, with two exceptions:
+// - `notifications` has NO simctl privacy service (notification authorization
+//   lives outside TCC), so it is null and surfaces a clear unsupported error.
+// - `camera` is not listed by every Xcode's simctl (the simulator has no
+//   camera hardware). It is passed through rather than pre-rejected so an
+//   Xcode that does support it works; on one that doesn't, simctl's own
+//   "invalid service" error is wrapped with a hint below.
```
latekvo: This comment says an unsupported `camera` surfaces as simctl's own "invalid service" error that is wrapped below. That conflicts with the rationale further down in this same file (lines 86-91) and with simctl's actual behavior: simctl rejects a service it doesn't model with a generic CoreSimulator NSError ("Failed to set access" / "Operation not permitted") that carries no reliable "invalid service" text — which is precisely why the hint keys off the service name instead of parsing the message. The two comments describe the same failure in contradictory ways.

packages/tool-server/src/tools/settings-permissions/platforms/android.ts
```diff
@@ -0,0 +1,210 @@
...
+    // API 36) and prints a `package:<name>` line only for installed packages.
+    // Because a *successful* run answers both "installed" and "not installed"
+    // (with vs without the line), any *thrown* error here is unambiguously a
+    // transport / timeout / package-manager-not-yet-up failure — never a "not
+    // installed" verdict. Let it propagate with adb's real cause rather than
+    // catching it and reporting a confidently wrong "not installed" (the failure
+    // mode of the old `pm path` probe, which exits non-zero for a missing
+    // package and so could not tell the two apart).
+    const listing = await adbShell(udid, `pm list packages ${pkg}`, { timeoutMs: 15_000 });
+    const installed = listing.split("\n").some((line) => line.trim() === `package:${bundleId}`);
```
latekvo, line 189: The installed-app check relies on an exact per-line match against `package:<bundleId>`. Because `pm list packages <pkg>` filters by substring, the normal output for an installed app is frequently multiple lines — the target plus any packages whose names contain it (e.g. `pm list packages com.google.android.gms` returns both `package:com.google.android.gms.supervision` and `package:com.google.android.gms`). The per-line `.some(...)` handles this correctly, but the tests only exercise the single-line, empty, and sibling-only-substring outputs — never the target-among-siblings shape — so a change to whole-output or first-line matching would keep every test green while making a real installed app read as "not installed".

packages/tool-server/src/utils/adb.ts
hubgan: Fixed in e00cd0a0 — reverted `isTerminalAdbError` back to module-private. The final `pm list packages` preflight deliberately lets adb errors propagate unchanged (any throw there is unambiguously transport/timeout/pm-not-up, never a "not installed" verdict), so it never needs the classifier. `adb.ts` is now identical to `main` again, and the JSDoc addressed to "callers that probe a device" is gone with it.

packages/tool-server/src/tools/settings-permissions/platforms/ios.ts
hubgan: Fixed in e00cd0a0 — reworded both camera comments so they agree with each other and with real simctl behavior. A rejected service fails with a generic `NSPOSIXErrorDomain` NSError ("Failed to set access" / "Operation not permitted"), not an "invalid service" string; the hint keys off the service name rather than parsing the message. Also clarified that camera support varies by simruntime independent of whether it appears in `simctl privacy`'s usage listing — verified on Xcode 26.5, where `camera` is absent from the usage text yet `grant camera` still succeeds.

packages/tool-server/src/tools/settings-permissions/platforms/android.ts
hubgan: Fixed in e00cd0a0 — added a `finds the target package among substring siblings` test: `pm list packages` returns three lines (`…app.helper`, `…app`, `…app.debug`) and the test asserts the exact-target line is matched and the grant proceeds. A regression to whole-output or first-line matching now fails, since the first line is a sibling. Also confirmed against a real device — `pm list packages com.google.android.gms` returns both `…gms.supervision` and `…gms`, and the preflight resolves the exact target and grants.

packages/tool-server/src/tools/settings-permissions/platforms/android.ts
```diff
@@ -0,0 +1,213 @@
...
+// exists depends on the app's manifest and the device's API level, and `pm`
+// itself is the authority on that.
+//
+// `reminders` is empty: iOS Reminders (EventKit) has no Android runtime
+// permission, so it surfaces an unsupported error rather than silently
+// no-opping.
+const ANDROID_PERMISSIONS: Record<PermissionName, string[]> = {
+  "camera": ["android.permission.CAMERA"],
+  "microphone": ["android.permission.RECORD_AUDIO"],
+  "photos": [
```
latekvo, line 35: On API 34+ the platform auto-adds `READ_MEDIA_VISUAL_USER_SELECTED` (a `dangerous` runtime permission) to any app that requests `READ_MEDIA_IMAGES`/`READ_MEDIA_VIDEO`, and the partial-access dialog grants it persistently while IMAGES/VIDEO get only a session-long grant. Because the `photos` mapping never includes it, `deny photos` revokes the three listed permissions and reports full success while the app still passes its partial-access check and keeps reading the user-selected media, and `reset photos` does not return the app to the not-yet-asked state (the next request shows the "keep/select more" flow, not the first-run dialog). Reproduced on an API 34 emulator: with USER_SELECTED granted, `deny photos` returned `applied: [READ_MEDIA_IMAGES, READ_MEDIA_VIDEO, READ_EXTERNAL_STORAGE]` and dumpsys still showed `READ_MEDIA_VISUAL_USER_SELECTED: granted=true` afterward. This only bites when partial access was previously established through the real dialog, but that dirty-state flow is one the skill doc explicitly targets ("re-enable one the user already denied", "reset dirty state").

packages/tool-server/src/tools/settings-permissions/platforms/android.ts
```diff
@@ -0,0 +1,213 @@
...
+
+// Resolve the concrete permission list for a (permission, action) pair.
+// `location-always` is the one action-dependent case: granting
+// ACCESS_BACKGROUND_LOCATION alone leaves the app unable to read location at
+// all (Android requires the foreground permissions too), while iOS's
+// `location-always` grants full always-access — so a grant fans out to
+// foreground + background to match the caller's intent. deny/reset stay
+// background-only: taking away "always" shouldn't also strip "while in use".
+function permissionsFor(permission: PermissionName, action: PermissionAction): string[] {
+  if (permission === "location-always" && action === "grant") {
```
latekvo, line 65: The reset half of the asymmetry documented above is unpinned: tests cover the grant fan-out and deny staying background-only, but no test resets `location-always`. Changing this condition to `action !== "deny"` — which makes reset revoke `ACCESS_FINE/COARSE_LOCATION` too, exactly the "taking away 'always' shouldn't also strip 'while in use'" behavior the comment forbids — leaves all 43 tests green (verified by running the suite against that mutation). The suite pins the other comment-contracts in this file (shutdown-beats-camera-hint, per-app reset, the deny side of this same rule), so the reset side is the one gap.

packages/tool-server/src/tools/settings-permissions/platforms/android.ts
```diff
@@ -0,0 +1,213 @@
...
+// non-zero exit throws from runAdb and is captured the same way.
+async function runPm(udid: string, pmArgs: string): Promise<PmResult> {
+  try {
+    const out = await adbShell(udid, `pm ${pmArgs}`, { timeoutMs: 15_000 });
+    const trimmed = out.trim();
+    if (trimmed && !/^Success/i.test(trimmed)) {
+      return { ok: false, detail: stripStackFrames(trimmed) };
+    }
+    return { ok: true, detail: trimmed };
+  } catch (err) {
```
latekvo, line 140: This catch folds every `adbShell` throw into the same per-permission failure shape as a pm rejection, which conflates two very different situations. If the device drops off mid-fan-out after one permission already succeeded (flaky USB / wireless adb), the call returns success with the dead-device permissions in `skipped` — which types.ts documents as entries "the package manager rejected (typically not declared in the app's manifest...)" — and the observed detail (e.g. `device 'emulator-5554' not found`) is dropped entirely on that path, so the caller is pointed at the manifest instead of the dead transport. Reproduced against the built module: `grant photos` with a disconnect after READ_MEDIA_IMAGES returns `{applied: [READ_MEDIA_IMAGES], skipped: [READ_MEDIA_VIDEO, READ_EXTERNAL_STORAGE]}` with no error. The catch also discards the FailureError classification `adbShell` already attached: on the all-rejected path the rethrow reports `error_kind: "subprocess"` with no `subprocessFailureMetadata`, so a wedged device timing out every pm call loses its `timeout` kind and exit/signal/command telemetry fields — the iOS branch of this same tool forwards exactly that metadata, and sibling Android branches (launch-app, open-url) let the classified adb error propagate.

packages/tool-server/src/tools/settings-permissions/platforms/android.ts
```diff
@@ -0,0 +1,213 @@
...
+      } else if (action === "deny") {
+        result = await runPm(udid, `revoke ${pkg} ${perm}`);
+      } else {
+        // reset = `pm revoke`, then a best-effort attempt to drop the
+        // user-set/user-fixed flags so the next request shows the dialog again.
+        // The revoke is the state-changing step and alone decides success: it is
+        // rejected only when pm genuinely refuses the permission (not declared /
+        // not a changeable type) — the manifest-style failure the aggregate
+        // error below describes. Clearing the flags is a refinement that must
+        // never demote a permission the revoke already changed:
```
latekvo, line 218: `clear-permission-flags` first appears in Android 13 — it is absent from the android11-release and android12-release `PackageManagerShellCommand` and present from android13-release — so this boundary (and the "API < 30" corollary below, the same claim in the skill doc, and the test comment) is off by two major versions. The ignore-the-outcome design holds either way, but the real consequence is wider than documented: on API 29-32, reset can never clear a user-fixed state (set automatically after repeated denies on 11/12), so a permission in "don't ask again" is revoked and reported reset while the dialog stays suppressed — and nothing in the result distinguishes that from a full reset. The skill doc's "does not exist before Android 11" actively misleads someone debugging exactly that on an Android 12 device.

packages/tool-server/src/tools/settings-permissions/platforms/ios.ts
```diff
@@ -0,0 +1,121 @@
...
+//   privacy`'s usage text lists it). It is passed through rather than
+//   pre-rejected so a runtime that accepts it works; a runtime that rejects it
+//   fails with a generic NSError (NSPOSIXErrorDomain, "Failed to set access" /
+//   "Operation not permitted") that carries no reliable "unsupported service"
+//   text, so the handler below keys a list-services hint off the service name
+//   rather than the message wording.
+const IOS_SERVICE: Record<PermissionName, string | null> = {
+  "camera": "camera",
+  "microphone": "microphone",
+  "photos": "photos",
```
latekvo: `photos` maps only to the `photos` TCC service, but add-only photo access lives in the separate `photos-add` service (`kTCCServicePhotosAdd`, listed separately by `simctl privacy`). After an app has been granted add-only access, `deny photos` and `reset photos` both return `applied: ["photos"]` while the `kTCCServicePhotosAdd` row survives untouched — reproduced on an iOS 18 simulator: granted `photos-add`, ran deny and then reset through the handler, and TCC.db still showed the add-only row at auth 2 after both reported success. For that state the documented deny/reset outcomes (the denied path, or the prompt reappearing) are reported as applied but don't happen.

packages/tool-server/src/tools/settings-permissions/platforms/ios.ts
```diff
@@ -0,0 +1,121 @@
...
+        },
+        { cause: err instanceof Error ? err : new Error(String(err)) }
+      );
+    }
+
+    return {
+      action,
+      permission,
+      bundleId,
+      applied: [service],
```
latekvo: For `location`, a grant to a bundle id that isn't installed exits 0 and returns `applied: ["location"]`, but nothing is recorded anywhere — no TCC row (location authorization doesn't live in TCC) and no locationd `clients.plist` entry, checked after settle time and across subsequent plist rewrites, with a whole-data-dir sweep finding no trace outside logs. Granting to an installed app writes `Authorization => 2` within seconds, and for TCC-backed services a pre-install grant persists in TCC.db and applies once the app is installed — `location` is the exception where grant-before-install silently applies nothing while the result claims it was applied.

packages/tool-server/src/tools/settings-permissions/index.ts
```diff
@@ -0,0 +1,70 @@
...
+      "App to change the permission for — required for every action. iOS: bundle id (e.g. com.example.app). Android: package name. `reset` is per-app too: simctl's device-wide reset (no bundleId) silently leaves existing per-app grants untouched on recent iOS, so the permission is always reset for this one app."
+    ),
+});
+
+type Params = z.infer<typeof zodSchema>;
+
+const capability: ToolCapability = {
+  // `simctl privacy` edits the simulator's TCC store — physical iPhones have no
+  // equivalent host-side switch, so no `device: true` on apple.
+  apple: { simulator: true },
```
latekvo, line 43: launch-app, restart-app, reinstall-app, and open-url all declare `appleRemote: { simulator: true }` with an `ios-remote` platform branch; this tool declares no remote capability. On a sim-remote setup an agent can install, launch, and open URLs, but the one tool in that family meant for permission test-setup returns a capability rejection, putting permission dialogs back into manual handling. This is the only tool in its sibling family without the remote branch.

packages/tool-server/src/tools/settings-permissions/index.ts
```diff
@@ -0,0 +1,70 @@
...
+  services: () => ({}),
+  execute: dispatchByPlatform<
+    SettingsPermissionsServices,
+    SettingsPermissionsServices,
+    Params,
+    SettingsPermissionsResult
+  >({
+    toolId: "settings-permissions",
+    capability,
+    ios: iosImpl,
```
latekvo, line 71: No test exercises this wiring: all 43 tests call `iosImpl.handler`/`androidImpl.handler` directly, so `execute`/`dispatchByPlatform`, the capability matrix, and each branch's `requires` are covered by nothing. Swapping these two lines to `ios: androidImpl, android: iosImpl` typechecks (both impls share the same `PlatformImpl` instantiation), passes lint, and leaves all 43 tests green — verified by running the suite against exactly that mutation — while the shipped tool would run `pm` against iOS UDIDs. Most cross-platform tools here test through `tool.execute` with shaped udids (launch-app, restart-app, describe, boot-device, tv-remote).

packages/tool-server/test/settings-permissions.test.ts
```diff
@@ -0,0 +1,591 @@
...
+const IOS_UDID = "AAAAAAAA-BBBB-CCCC-DDDD-EEEEEEEEEEEE";
+const ANDROID_SERIAL = "emulator-5554";
+
+const iosDevice: DeviceInfo = { id: IOS_UDID, platform: "ios", kind: "simulator" };
+const androidDevice: DeviceInfo = { id: ANDROID_SERIAL, platform: "android", kind: "emulator" };
+
+// FailureError attaches its FailureSignal under a non-enumerable symbol, so
+// toMatchObject can't see it — assert through the public accessor instead.
+function failsWith(code: string): (err: unknown) => boolean {
+  return (err) => getFailureSignal(err)?.error_code === code;
```
latekvo: If a `FAILURE_CODES` member ever resolves to `undefined` at test runtime — vitest loads `@argent/registry` from its built dist, so a stale local dist that predates the three new codes does exactly that — this comparison becomes `undefined === undefined`, and every `rejects.toSatisfy(failsWith(...))` in the file passes for any rejection, including TypeErrors from broken mocks. Verified end-to-end: with the code removed from the loaded dist, the full suite stayed green and a probe rejection carrying a plain Error satisfied the matcher. CI rebuilds before testing so it is shielded; a local `vitest run` without a rebuild is not, and nothing asserts the codes are strings.

packages/skills/skills/argent-settings-permissions/SKILL.md
```diff
@@ -0,0 +1,129 @@
...
+| The app has an **in-app control** for the permission (a toggle in its own settings screen)                       | **Tap it in the app** (`describe` → `gesture-tap`) — do NOT use this tool                     | It's real user behavior and exercises the flow you're testing. See `argent-device-interact`.                                                                         |
+| The app is **about to ask** (or just asked) and the system **permission dialog is on screen**                    | **Tap the dialog** (`Allow` / `Don't Allow` / `Allow While Using App`) — do NOT use this tool | The app-triggered prompt is the natural path; answering it is what a user does. `describe` exposes the dialog buttons; fall back to `screenshot` only if it doesn't. |
+| You need the permission **already granted/denied before the app runs**, so no dialog interrupts the flow         | **Use this tool** (`grant` / `deny`) before `launch-app`                                      | The app can't pre-set its own permission; a real user would do it in Settings. This is the core use case.                                                            |
```
latekvo: For `deny`, this row's promise — "already granted/denied before the app runs, so no dialog interrupts the flow" — doesn't hold on Android: `pm revoke` clears the grant but records no refusal (it sets no user-fixed flag, and dialog suppression keys off user-fixed/policy-fixed), so the app's next `requestPermissions` still shows the system dialog. That is faithful to Android's own single-deny semantics, but it silently diverges from the iOS half of this tool, where deny writes a TCC denial that answers the app without prompting. Nothing in the doc or tool description warns that a pre-launch deny keeps the dialog out of the flow on iOS only.

packages/skills/skills/argent-settings-permissions/SKILL.md
```diff
@@ -0,0 +1,129 @@
...
+
+**Rule of thumb:** if a human tester could flip it _inside the app_, do that. Reach for `settings-permissions` only for a change a human would otherwise make in the **system Settings app**.
+
+## Supported permissions & platform coverage
+
+| `permission`      | iOS simulator (`simctl privacy` service)                                                       | Android (`android.permission.*`)                                                                          |
+| ----------------- | ---------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------- |
+| `camera`          | `camera` — only if the installed Xcode's simctl models it (simulators have no camera hardware) | `CAMERA`                                                                                                  |
```
latekvo: This row — plus the gotcha at line 84, the closing note at line 93, and the `permission` describe-string in index.ts — attributes camera availability to "the installed Xcode's simctl", but the implementation comment (platforms/ios.ts lines 28-35) states support varies by simruntime, independent of whether the Xcode's `simctl privacy` usage text lists it. That is also the observed behavior: on Xcode 16.4 (and 26.5 per the earlier review round) the usage text omits `camera` while camera grants succeed. An agent following this wording will blame or switch the Xcode install, or treat the usage listing as authoritative — the error hint in ios.ts points at that same listing — when the actual variable is the target simulator's runtime.

packages/skills/skills/argent-settings-permissions/SKILL.md
```diff
@@ -0,0 +1,129 @@
...
+- The app must **declare** the permission in its manifest. `pm` rejects any mapped permission the manifest doesn't request; those come back in the result's `skipped` list. The action succeeds if **at least one** mapped permission sticks, and errors only if `pm` rejected **all** of them.
+
+## Gotchas
+
+- **Changing a permission can terminate a running app** (system behavior on both platforms). Prefer setting permissions **before** `launch-app`; if you change one while the app is running, `restart-app` afterward.
+- **Reset is per-app on both platforms** — pass `bundleId`; there is no reliable device-wide reset.
+- **A partial Android result is normal.** `applied` lists what actually changed; `skipped` lists mapped permissions `pm` rejected (usually not in the manifest, or gated by API level). Both together tell you what happened.
+- **`camera` on iOS** may be rejected by an Xcode whose simctl doesn't model it (simulators have no camera hardware). simctl reports this as a generic CoreSimulator error, so a `camera` failure always carries a hint to run `xcrun simctl privacy` and list the supported services.
```
latekvo: "Always" isn't true: the hint is deliberately suppressed when the camera failure is a shutdown-simulator error (platforms/ios.ts gates it on `!shutdownHint`, pinned by the "camera failure on a shutdown simulator" test). The closing note at line 93 has the same problem in a worse spot — it sits right after a failure list that includes "a shutdown simulator (iOS)" and still says a camera failure "additionally hints".

packages/skills/skills/argent-device-interact/SKILL.md
```diff
@@ -216,6 +216,8 @@ When using `screenshot` for permission or native modal navigation:
 - Tap one control at a time and inspect the returned auto-screenshot before doing anything else.
 - After the modal is dismissed, return to normal discovery with `describe`, `native-describe-screen`, or `debugger-component-tree`.
 
+> **Prefer the dialog over the Settings tool.** When the app triggers its own permission prompt, answering it here is the real user path — do that. Reach for the `argent-settings-permissions` tool only when you can't get to the change through the app: pre-authorize/deny a permission _before_ the app asks, re-enable one the user already denied (iOS won't re-prompt), or reset it so the prompt reappears. See the `argent-settings-permissions` skill.
```
latekvo: "Reach for the `argent-settings-permissions` tool" names the skill, not the tool — the tool id is `settings-permissions` (the sentence's own closing reference, "See the `argent-settings-permissions` skill", uses the name correctly). An agent searching its tool list for `argent-settings-permissions` finds nothing.

packages/tool-server/src/tools/settings-permissions/platforms/ios.ts
```diff
@@ -0,0 +1,244 @@
...
+      const detail = err instanceof Error ? err.message : String(err);
+      // `get_app_container` needs a booted device: on a shutdown/booting sim it
+      // fails with "Unable to lookup in current state: Shutdown" for installed
+      // AND missing apps, so that error is NOT a "not installed" verdict —
+      // return undefined so the guard is skipped and the privacy call surfaces
+      // the boot hint instead of misdirecting the agent to reinstall the app.
+      if (/current state:/i.test(detail)) return undefined;
+      // Any other non-zero exit on a usable device means the app is absent
+      // ("No such file or directory" / "... is not installed").
+      return false;
```
latekvo: The interface contract above (lines 80-88) reserves `false` for a definitive "not installed" and `undefined` for "the probe couldn't answer", but this catch routes only the shutdown shape to `undefined` — every other probe failure becomes a definitive verdict. Two shapes that aren't verdicts land here: a nonexistent/stale UDID (`get_app_container` fails with "Invalid device", and device resolution is shape-based, so a deleted sim's UDID reaches this path) and a probe timeout (killed, no "current state" text — a wedged CoreSimulator). Both produce "the app is not installed — install the app first, then grant" for a question the probe never answered. Verified end-to-end: on a random UUID, `grant location` reports the app isn't installed while `grant microphone` on the same UUID correctly surfaces simctl's "Invalid device". No false success — the call fails either way — but the steer is wrong in exactly the confidently-wrong-diagnosis way the guard exists to avoid.

packages/tool-server/src/tools/settings-permissions/platforms/android.ts
```diff
@@ -0,0 +1,251 @@
...
+//      manifest-style rejection that belongs in `skipped`/the aggregate error.
+//   2. the adb transport itself failed — the device dropped mid-fan-out
+//      (unauthorized / offline / not found) or the call was killed at its
+//      timeout. That is not "this permission was rejected"; it means every
+//      remaining call is unreliable and the observed cause (a dead device, a
+//      wedged pm) must reach the caller, not be relabelled as a manifest gap.
+// adbShell already classifies (2) into a FailureError with error_kind +
+// subprocess metadata, so we detect it and let that error propagate intact
+// rather than folding it into the per-permission failure shape.
+function isTransportFailure(err: unknown): boolean {
```
latekvo, line 106: This nets the four device-state patterns plus the timeout kind, but adb's client↔daemon leg fails with shapes that match neither: `adb: protocol fault (couldn't read status): Connection reset by peer` (an adb server restarting mid-command — a real event on hosts running mixed adb versions, where a version-mismatched client kills and restarts the shared server) and `adb: cannot connect to daemon at tcp:...: Connection refused`. Both classify as `error_kind: "subprocess"` and match no terminal pattern (verified against the built `adbShell` with a redirected `ADB_SERVER_SOCKET`), so they fold into the per-permission failure shape: a one-call blip mid-fan-out puts the lost permission into `skipped` under a success result — a deny whose revoke never executed reports success — and a daemon that stays down after the preflight produces the "every mapped runtime permission was rejected... manifest" aggregate, with the transport cause surviving only inside the parenthesized detail. The same failure at the preflight (line 166) propagates with adb's own classification, and the comment above (lines 80-84) defines "the adb transport itself failed" as exactly the category that must propagate — the daemon leg is in that category but escapes the net.

packages/tool-server/src/tools/settings-permissions/platforms/android.ts
```diff
@@ -0,0 +1,251 @@
...
+        // The revoke is the state-changing step and alone decides success: it is
+        // rejected only when pm genuinely refuses the permission (not declared /
+        // not a changeable type) — the manifest-style failure the aggregate
+        // error below describes. Clearing the flags is a refinement that must
+        // never demote a permission the revoke already changed:
+        //   - `clear-permission-flags` first appears in Android 13 (API 33) — it
+        //     is absent from android11/android12-release's PackageManagerShell-
+        //     Command and an unknown pm subcommand exits non-zero — so coupling
+        //     reset to it would make every reset on API < 33 falsely report
+        //     failure. (On API 29-32 this means reset can't clear a user-fixed
```
latekvo: This range is under-inclusive (it inherited the round-3 review comment's own phrasing): `FLAG_PERMISSION_USER_FIXED` exists since API 23 — it is what the checkbox-era "Don't ask again" sets on 23-29; API 30 only made it automatic after the second deny — and the flag-clear subcommand is absent everywhere below 33. So reset cannot clear a user-fixed state on API 23-32, not just 29-32; as written here and in the SKILL gotcha it implies sub-29 devices are unaffected. The capability matrix declares no Android version floor, so those levels are in scope.

packages/skills/skills/argent-settings-permissions/SKILL.md
```diff
@@ -0,0 +1,131 @@
...
+- **`camera` on iOS** may be rejected by a simulator **runtime** that doesn't model the service (it varies by simruntime, not by the installed Xcode - a runtime can accept `camera` even when the Xcode's `simctl privacy` usage text omits it). simctl reports a rejection as a generic CoreSimulator error, so a `camera` failure (unless it's the shutdown-simulator case, which gets the boot hint instead) carries a hint to run `xcrun simctl privacy` and list the supported services.
+- **`grant location` needs the app installed first (iOS).** Location authorization isn't stored in TCC and isn't applied to a bundle id until the app exists, so a pre-install `grant location` / `grant location-always` records nothing - the tool checks and errors clearly instead of reporting a false success. (TCC-backed services like `camera`/`photos` _can_ be granted before install; they persist and apply on install.)
```
latekvo: The remote branch this PR also adds can't keep this promise: `sim-remote` has no app-container verb, so the remote backend omits the install probe, and a pre-install `grant location` on a sim-remote target returns `applied: ["location"]` — the false success this sentence says the tool prevents. The exemption is stated only in code comments; neither this doc nor the tool description scopes the check to local simulators, and the only remote-path test grants `microphone`, so nothing exercises the remote guard-skip.

packages/skills/skills/argent-settings-permissions/SKILL.md
```diff
@@ -0,0 +1,131 @@
...
+| ----------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------ | --------------------------------------------------------------------------------------------------------- |
+| `camera`          | `camera` - only if the target simulator's **runtime** models it (varies by simruntime, not by the installed Xcode; simulators have no camera hardware) | `CAMERA`                                                                                                  |
+| `microphone`      | `microphone`                                                                                                                                           | `RECORD_AUDIO`                                                                                            |
+| `photos`          | `photos` + `photos-add` (add-only access is a separate TCC service; deny/reset clear both)                                                             | `READ_MEDIA_IMAGES` + `READ_MEDIA_VIDEO` + `READ_MEDIA_VISUAL_USER_SELECTED` + `READ_EXTERNAL_STORAGE`    |
```
latekvo: "deny/reset clear both" is stronger than the implementation guarantees: `photos-add` is deliberately best-effort — a secondary-service failure is skipped silently, it does not appear in the result's `skipped` (which types.ts pins as Android-only), and types.ts's `applied` doc still describes fan-out as an Android-only phenomenon. When the second call fails, the result is a success whose `applied` lists only `photos` — consistent with the code's stated best-effort intent, but not with this row's unconditional wording.

PR #471, opened 2026-07-06
fix(argent-lens): strip ELECTRON_RUN_AS_NODE from spawned Electron env

latekvo:
## Problem

QA report: **Argent Lens preview window never launches** — `propose_variant` / `await_user_selection` stage variants correctly, but the native window never appears and `await_user_selection` hangs. The tool-server log shows the child crashing at launch:

```
dist/preview-window/main.cjs:28
import_electron.app.setName("Argent Lens");
                    ^
TypeError: Cannot read properties of undefined (reading 'setName')
    ...
    at c._load (node:electron/js2c/node_init:2:18095)
```

## Root cause (reproduced & proven)

`import_electron.app` being `undefined` while the JS still executes is the exact signature of the Electron binary booting in **Node mode**. For a downloaded, non-fused Electron the only trigger is `ELECTRON_RUN_AS_NODE` in the environment. In Node mode `require("electron")` returns the executable *path string* (not the module), so `.app` is `undefined`.

The tool-server routinely runs under an Electron-based MCP host (VS Code, Cursor, the Codex desktop app), which sets `ELECTRON_RUN_AS_NODE=1` in the env of the Node children it spawns. `createPreviewWindowManager.ensureOpen()` spread `{ ...process.env }` straight into its Electron child, so the flag leaked in and the child booted headless as Node → crash before any window.

Verified against the real Electron binary + the real `main.cjs`:

| Env | Result |
|-----|--------|
| `ELECTRON_RUN_AS_NODE=1` | Crashes at `main.cjs:28` — **byte-identical to the report**, incl. the `node:electron/js2c/node_init` frame |
| unset | Boots as a real Electron GUI, gets past `setName`, only fails later at `loadURL` |

The wrapper-bundle / code-signing theory in the report was a **red herring** — the reporter never controlled for the spawn *environment*. The same node-mode surfaces differently depending on whether the `.app` wrapper built: plain path → `setName` crash (what they saw); wrapper path → Node rejects the Electron-only `--no-sandbox` flag (`bad option: --no-sandbox`). One fix cures both.

## Fix

Strip `ELECTRON_RUN_AS_NODE` before spawning, via a shared `electronGuiChildEnv()` helper, at **both** GUI-Electron spawn sites:
- `preview-window.ts` — the Lens window (the reported bug).
- `boot-electron.ts` — the same latent flaw would boot a headless CDP app with **no debug port**, timing out `boot-device --electronAppPath`.

## Verification

- **Root cause** reproduced byte-for-byte, both directions, against the real binary.
- **End-to-end**: ran the real fixed `ensureOpen()` path via ts-node with `ELECTRON_RUN_AS_NODE=1` leaked into the parent env against the real Electron binary — child boots as a real GUI app (`VERDICT: PASS`). Reverting the one-line fix flips it to `FAIL`, confirming the test discriminates.
- **Unit tests**: added `electron-env.test.ts` (helper contract) + a regression case in `preview-window-launch-failure.test.ts` asserting the spawn env strips the var and keeps the override. 21/21 pass.
- `tsc --noEmit` (build) and `tsc --noEmit -p tsconfig.test.json` (CI "Unit tests" gate) both clean; prettier + eslint clean on all changed files.

### DIFF
--- a/packages/tool-server/src/utils/electron-env.ts
+++ b/packages/tool-server/src/utils/electron-env.ts
@@ -1,0 +1,26 @@
+/**
+ * Build the environment for spawning a GUI Electron process as a child of the
+ * tool-server.
+ *
+ * The tool-server routinely runs under an Electron-based MCP host — VS Code,
+ * Cursor, the Codex desktop app, and friends all spawn their Node children
+ * (this tool-server, and the MCP stdio server that starts it) with
+ * `ELECTRON_RUN_AS_NODE=1` set in the environment. That flag tells the Electron
+ * binary to boot as a bare Node runtime instead of a GUI app.
+ *
+ * If it leaks into a GUI Electron child we launch, the child boots in Node
+ * mode: `require("electron")` returns the executable PATH string (not the
+ * Electron module), so `.app` is `undefined` and the process crashes at the
+ * first `app.*` call — e.g. `app.setName()` at the top of the Lens preview
+ * window's main entry. The failure is silent from the tool-server's side: the
+ * window never appears and a parked `await_user_selection` just hangs.
+ *
+ * Stripping the variable makes a GUI Electron child boot as a real Electron app
+ * regardless of what launched the tool-server. Any per-launch additions (the
+ * preview URL, logging flags) are layered on top via `overrides`.
+ */
+export function electronGuiChildEnv(overrides: Record<string, string> = {}): NodeJS.ProcessEnv {
+  const env: NodeJS.ProcessEnv = { ...process.env, ...overrides };
+  delete env.ELECTRON_RUN_AS_NODE;
+  return env;
+}
--- a/packages/tool-server/src/utils/electron-env.ts
+++ b/packages/tool-server/src/utils/electron-env.ts
@@ -18,9 +18,21 @@
  * Stripping the variable makes a GUI Electron child boot as a real Electron app
  * regardless of what launched the tool-server. Any per-launch additions (the
  * preview URL, logging flags) are layered on top via `overrides`.
+ *
+ * The strip is case-insensitive: Windows environment variables are
+ * case-insensitive, so a host may surface the flag under non-canonical casing
+ * (e.g. `Electron_Run_As_Node`). The plain object spread from `process.env`
+ * preserves that casing, so a case-sensitive `delete` of the canonical name
+ * would miss it and leak Node mode into the cross-platform `boot-electron`
+ * child. (On macOS/Linux the name is always the exact uppercase form, so the
+ * loop simply matches that one key.)
  */
 export function electronGuiChildEnv(overrides: Record<string, string> = {}): NodeJS.ProcessEnv {
   const env: NodeJS.ProcessEnv = { ...process.env, ...overrides };
-  delete env.ELECTRON_RUN_AS_NODE;
+  for (const key of Object.keys(env)) {
+    if (key.toLowerCase() === "electron_run_as_node") {
+      delete env[key];
+    }
+  }
   return env;
 }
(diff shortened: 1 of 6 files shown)


filip131311 (approved):
That's crazy tell me a human would find it anywhare as quicly

packages/tool-server/src/utils/electron-env.ts
```diff
@@ -0,0 +1,26 @@
...
+ * window's main entry. The failure is silent from the tool-server's side: the
+ * window never appears and a parked `await_user_selection` just hangs.
+ *
+ * Stripping the variable makes a GUI Electron child boot as a real Electron app
+ * regardless of what launched the tool-server. Any per-launch additions (the
+ * preview URL, logging flags) are layered on top via `overrides`.
+ */
+export function electronGuiChildEnv(overrides: Record<string, string> = {}): NodeJS.ProcessEnv {
+  const env: NodeJS.ProcessEnv = { ...process.env, ...overrides };
+  delete env.ELECTRON_RUN_AS_NODE;
```
hubgan: This `delete` operates on a plain object spread from `process.env`, which is case-sensitive, so it only removes an exactly-uppercase key. On Windows the OS environment is case-insensitive, so a host that put the flag into the env under a different casing would leave the copied key in place and the spawned GUI child would still boot in Node mode. In practice every host that sets this writes the canonical uppercase name and the Lens preview window is macOS-only, so this is a latent edge rather than a live bug — but `boot-electron` is cross-platform, so worth being aware of.

packages/tool-server/test/electron-env.test.ts
```diff
@@ -0,0 +1,42 @@
...
+describe("electronGuiChildEnv", () => {
+  const prev = process.env.ELECTRON_RUN_AS_NODE;
+  afterEach(() => {
+    if (prev === undefined) delete process.env.ELECTRON_RUN_AS_NODE;
+    else process.env.ELECTRON_RUN_AS_NODE = prev;
+  });
+
+  it("removes ELECTRON_RUN_AS_NODE inherited from the parent env", () => {
+    process.env.ELECTRON_RUN_AS_NODE = "1";
+    expect(electronGuiChildEnv().ELECTRON_RUN_AS_NODE).toBeUndefined();
```
hubgan, line 13: None of the cases here asserts that `electronGuiChildEnv` leaves the real `process.env` unmutated. The implementation is correct today (it deletes from a copy), but every case also restores the global in `afterEach`, so a future change that deleted the key from `process.env` directly — permanently stripping the flag from the tool-server process itself — would still pass this suite. The regression would be invisible at this layer.

hubgan (approved):

packages/tool-server/src/utils/electron-env.ts
latekvo: Fixed in 6aef3cb1 — the strip is now case-insensitive: it iterates the copied env and deletes any key that lowercases to `electron_run_as_node`, so a differently-cased flag from a Windows host no longer survives into the cross-platform `boot-electron` child. Added a regression case (`strips the flag regardless of key casing`) that seeds a mixed-case `Electron_Run_As_Node` and asserts no node-mode key remains; it fails against the old exact-case `delete`.

packages/tool-server/test/electron-env.test.ts
latekvo: Fixed in 6aef3cb1 — added a `does not mutate the caller's process.env` case that sets the flag, calls the helper, and asserts `process.env.ELECTRON_RUN_AS_NODE` is still `1` afterward. Confirmed it is a real guard, not decoration: temporarily swapping in a buggy impl that `delete`s from `process.env` directly makes exactly this case fail, while every other case (which restore the global in `afterEach`) stays green.

PR #473, opened 2026-07-06
fix(tools-client): unref spawned tool-server's stdout pipe so `argent run` exits

filip131311:
## Problem

If the tool-server is **not** already running the first time `argent run <tool>` is called, the CLI prints its result but **never exits** — it hangs indefinitely, even when the tool call succeeded.

## Root cause

`spawnToolsServer` (in `packages/argent-tools-client/src/launcher.ts`) launches the tool-server `detached` with `stdio: ["ignore", "pipe", logFd]` and calls `child.unref()`. But `child.unref()` only detaches the **process** handle — the piped `child.stdout` is a *separate* `net.Socket` handle. After reading the "listening" banner the code calls `child.stdout.resume()` (so the child doesn't hit EPIPE) but leaves that socket **ref'd**, which keeps the short-lived caller's event loop alive forever.

The server-reuse path never opens that pipe, so only the first (spawning) invocation hangs — exactly matching the report.

## Fix

Unref the stdout socket right after the ready banner is read:

```ts
child.stdout?.resume();
(child.stdout as { unref?: () => void } | null)?.unref?.();
```

Safe for both callers:
- **`argent run`** (short-lived): the loop drains the pipe while working, then exits. The tool-server already installs an EPIPE handler on stdout (`packages/tool-server/src/index.ts`), so it tolerates the parent closing the read end.
- **MCP launcher** (long-lived): its loop stays alive by other means, so the pipe still drains normally; the unref is a no-op for it.

## Testing

- New regression test `test/launcher-exit.test.ts` — **fails without the fix**, passes with it.
- End-to-end control: a real subprocess calling the launcher **hangs** with the unref removed and **exits cleanly** with it in place.
- Full local CI parity: `tsc --build`, `typecheck:tests` (all workspaces), `typecheck:scripts`, full test suite (12 packages, ~2000 tests) — all green. ESLint + Prettier clean.

### DIFF
--- a/packages/argent-tools-client/src/launcher.ts
+++ b/packages/argent-tools-client/src/launcher.ts
@@ -278,9 +278,19 @@ export function spawnToolsServer(
       if (match) {
         const actualPort = parseInt(match[1]!, 10);
         rl.close();
-        // Resume stdout so the pipe stays open and the child's console.log calls
-        // don't fail with EPIPE once the readline interface stops consuming it.
+        // Resume stdout so the pipe keeps draining and the child's console.log
+        // calls don't back up once the readline interface stops consuming it.
         child.stdout?.resume();
+        // ...but unref the pipe socket so it does NOT keep OUR event loop alive.
+        // `child.unref()` only detaches the process handle; the stdout pipe is a
+        // separate ref'd handle. Without this, a short-lived caller like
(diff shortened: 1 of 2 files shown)


jwajgelt (approved):
As long as tools server really handles closed `stdout` gracefully, this should be fine.

PR #477, opened 2026-07-07
feat(telemetry): distinguish tvOS and Android TV from iOS/Android

hubgan:
## Summary

Telemetry inferred the device platform purely from device-id **shape** on the per-tool-call hot path, so a tvOS simulator (same 8-4-4-4-12 UUID as an iPhone) and an Android TV emulator (same `emulator-NNNN` serial as a phone) were both reported as `ios` / `android` — TV usage was invisible in analytics.

This adds `tvos` and `android-tv` to the **telemetry** `PLATFORMS` enum and refines the inference so a TV target is attributed correctly.

## Approach

- The runtime kind (`mobile` / `tv`) is already resolved and memoized in two module-level caches (`runtimeKindCache` in `ios-devices.ts`, `androidRuntimeKindCache` in `adb.ts`), warmed by the usual TV-path tools (describe, screenshot, keyboard, launch/restart-app, the streaming blueprint, tv-control).
- New synchronous, **cache-only** readers (`getCachedSimulatorRuntimeKind`, `getCachedAndroidRuntimeKind`) let `refineTvPlatform` in `http.ts` map `ios → tvos` / `android → android-tv` with **no new `simctl`/`adb` round-trip** on the hot path.
- The registry **device** `Platform` (`@argent/registry`) is intentionally left unchanged — a TV stays a `runtimeKind`, not its own device platform, so capability gating and dispatch remain TV-agnostic. The two new values are confined to the telemetry boundary (`tool:invoke` / `tool:complete` / `tool:fail` and child sub-tool attribution).

## Behavior notes

- The **first** tool call on a device may land before the cache is warm and reports the coarse base platform (`ios`/`android`); subsequent calls, once describe/screenshot/streaming have run, report the TV variant. This is best-effort coarse telemetry by design.
- For an `emulator-NNNN` slot reclaimed by a different-form-factor AVD within one tool-server lifetime, the synchronous reader can momentarily report the prior occupant's kind until an async, avdName-checked path re-probes. Documented inline; no crash and no dispatch impact.

## Verification

- **Live E2E** against real booted devices (real `simctl`/`adb` → real cache → real http telemetry recording): Apple TV → `tvos` (cold-cache `ios`, warm `tvos`); iPhone → `ios`; Android phone → `android`. The `android-tv` path is symmetric and covered by unit/integration tests with real leanback-feature probing.
- Unit + integration tests added: sanitize allowlist, http platform inference (cold/warm/TV/child), and cache readers. Telemetry suite 192/192; tool-server affected suites green.
- `tsc` build, `typecheck:tests`, eslint (`--max-warnings 0`), and prettier all clean.

## Out of scope

- The `Telemetry.md` privacy-notice update (adding TV/Fire TV to the collected platform-family list) is intentionally **not** in this PR — it will come from a separate PR.

### DIFF
--- a/packages/tool-server/src/http.ts
+++ b/packages/tool-server/src/http.ts
@@ -128,7 +128,9 @@ type HttpFailureMeta = { platform?: TelemetryPlatform } & AiTelemetryProps;
  * `tvos` / `android-tv` here for reporting only when the cache already knows the
  * kind, and leave it coarse otherwise. The first tool call on a device may land
  * before the cache is warm and report the base platform; subsequent calls, once
- * describe/screenshot/streaming have run, report the TV variant.
+ * a describe/interaction path has warmed the runtime-kind cache (the per-platform
+ * warmers are listed on `getCachedSimulatorRuntimeKind` /
+ * `getCachedAndroidRuntimeKind`), report the TV variant.
  */
 function refineTvPlatform(basePlatform: DevicePlatform, deviceId: string): TelemetryPlatform {
   if (basePlatform === "ios" && getCachedSimulatorRuntimeKind(deviceId) === "tv") {
@@ -180,14 +182,21 @@ function extractInvocationMeta(
   return Object.keys(meta).length > 0 ? meta : null;
 }
 
-/** Coarse platform from a tool call's device arg (`udid` / `device_id` / `avdName`), or null. */
+/**
+ * Telemetry platform from a tool call's device arg, or null when it carries none.
+ * A `udid` / `device_id` resolves through the runtime-kind cache and refines to
+ * `tvos` / `android-tv` once that cache is warm (coarse `ios` / `android` until
+ * then); only the `avdName`-only fallback is unconditionally coarse.
+ */
 function platformFromArgs(data: unknown): TelemetryPlatform | null {
   if (!data || typeof data !== "object") return null;
   const deviceArg = extractDeviceArg(data);
   if (deviceArg) return inferPlatform(deviceArg) ?? null;
   // An `avdName`-only call (boot-device before the emulator exists) has no serial
-  // to resolve a runtime kind from, so it stays the coarse `android` platform —
-  // an Android TV AVD is attributed to `android-tv` on its first device-arg call.
+  // to resolve a runtime kind from, so it stays the coarse `android` platform.
+  // Once the emulator is up, later `udid` / `device_id` calls refine an Android TV
+  // AVD to `android-tv` — from the call after describe/launch has warmed the cache,
+  // not the first (see `refineTvPlatform`).
   if (typeof (data as Record<string, unknown>).avdName === "string") return "android";
   return null;
 }
--- a/packages/tool-server/src/http.ts
+++ b/packages/tool-server/src/http.ts
@@ -5,11 +5,17 @@ import {
   FAILURE_CODES,
   type FailureSignal,
   type FileInputSpec,
-  type Platform,
+  // The tool-server's coarse *device* platform (TV-agnostic). Telemetry's own
+  // Platform (imported below) is a superset that also carries `tvos`/`android-tv`.
+  type Platform as DevicePlatform,
   type Registry,
   type ResolvedFileInput,
 } from "@argent/registry";
-import { AI_CLIENTS, type AiTelemetryProps } from "@argent/telemetry";
+import {
+  AI_CLIENTS,
+  type AiTelemetryProps,
+  type Platform as TelemetryPlatform,
+} from "@argent/telemetry";
 import { ToolNotFoundError } from "@argent/registry";
 import { createIdleTimer, IDLE_CHECK_INTERVAL_MS } from "./utils/idle-timer";
 import { DependencyMissingError, ensureDeps } from "./utils/check-deps";
@@ -25,6 +31,8 @@ import {
   UnsupportedOperationError,
 } from "./utils/capability";
 import { resolveDevice } from "./utils/device-info";
+import { getCachedSimulatorRuntimeKind } from "./utils/ios-devices";
+import { getCachedAndroidRuntimeKind } from "./utils/adb";
 import type { Server as HttpServer } from "node:http";
 import {
   CHROMIUM_CDP_NAMESPACE,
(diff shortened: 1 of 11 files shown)


packages/tool-server/src/http.ts
```diff
@@ -150,10 +181,13 @@ function extractInvocationMeta(
 }
 
 /** Coarse platform from a tool call's device arg (`udid` / `device_id` / `avdName`), or null. */
-function platformFromArgs(data: unknown): Platform | null {
+function platformFromArgs(data: unknown): TelemetryPlatform | null {
```
latekvo, line 191: The JSDoc summary still calls the result the `Coarse platform`, but the `udid`/`device_id` branch just below now routes through `inferPlatform` -> `refineTvPlatform` and can return the refined `tvos`/`android-tv` values; only the `avdName`-only fallback stays coarse. For the function's primary path the description no longer matches what it returns.

packages/tool-server/src/http.ts
```diff
@@ -150,10 +181,13 @@ function extractInvocationMeta(
...
   if (!data || typeof data !== "object") return null;
   const deviceArg = extractDeviceArg(data);
   if (deviceArg) return inferPlatform(deviceArg) ?? null;
+  // An `avdName`-only call (boot-device before the emulator exists) has no serial
+  // to resolve a runtime kind from, so it stays the coarse `android` platform —
+  // an Android TV AVD is attributed to `android-tv` on its first device-arg call.
```
latekvo: This states an Android TV AVD `is attributed to android-tv on its first device-arg call`, but `refineTvPlatform` reads a cache that the first device-arg call has typically not warmed yet. Per that function's own docstring above, the first call on a device reports the coarse base platform, and only later calls (once describe/screenshot/streaming have warmed the cache) report the TV variant - so the two comments disagree about first-call behavior.

packages/tool-server/src/utils/ios-devices.ts
```diff
@@ -82,3 +82,22 @@ export async function getSimulatorRuntimeKind(udid: string): Promise<"mobile" |
...
 }
+
+/**
+ * Synchronous, cache-only view of a UDID's runtime kind: returns the memoized
+ * "mobile"/"tv" verdict if a prior `getSimulatorRuntimeKind` call resolved it,
+ * otherwise undefined. It NEVER runs `simctl` — callers on a latency-sensitive
+ * hot path (telemetry platform inference) use this to distinguish tvOS from iOS
+ * only when the kind is already known, and fall back to the coarse platform when
+ * it isn't, rather than paying a ~100ms probe per call. The cache is warmed as a
+ * side effect of the describe/screenshot/keyboard/streaming paths that any real
```
latekvo: This warmer set omits the Apple TV control path. The tvOS `tv-control` blueprint resolves the device via `listIosSimulators()` (blueprints/tv-control.ts:230), which does not populate `runtimeKindCache` - only `getSimulatorRuntimeKind`/`isTvOsSimulator` do. The Android side differs: `android-tv-control` (blueprints/android-tv-control.ts:183) warms its cache through `getAndroidRuntimeKind`. Effect: an Apple TV session driven only by `tv-remote` (with no describe/screenshot/keyboard/streaming) never warms this cache, so its events stay attributed to coarse `ios` and tvOS usage remains invisible, whereas the equivalent Android TV session is attributed `android-tv` from the second call. Coarser rather than wrong, but an asymmetry between the two TV platforms' coverage.

latekvo (approved):
Reviewed with real-hardware end-to-end verification. Approving - the telemetry-only refinement is sound, and the design (a 7-value telemetry `Platform` superset over the deliberately-coarse 5-value device `Platform`) is internally consistent.

What I verified:
- Real tvOS E2E on a booted Apple TV simulator: real `simctl` -> runtime-kind cache -> `createHttpApp` hot path records `platform: "tvos"`; iPhone -> `ios`; a cold cache-miss -> coarse `ios`; the failure path (404) also refines to `tvos`; and the real telemetry `sanitize` preserves `tvos` for both `tool:invoke` and `tool:fail`.
- Real Android E2E on a booted Pixel emulator: real `adb` detects the phone as `mobile`, and the hot path records coarse `android` (correctly not over-refined to `android-tv`).
- Full workspace build clean; telemetry 192/192; the tool-server suite is green apart from a pre-existing, load-sensitive `await-ui-element` 60ms real-timer flake in a file this PR does not touch (the change is purely additive to `ios-devices.ts`/`adb.ts`); eslint `--max-warnings 0` clean; both packages' `typecheck:tests` clean.
- Confirmed the synchronous cache-only readers are consumed only by `refineTvPlatform` (telemetry), so the documented Android slot-reuse mislabel is telemetry-bounded, self-healing, and has zero dispatch/capability impact.

A few non-blocking nits inline (coarser-not-wrong / doc accuracy).

packages/tool-server/src/http.ts
hubgan: Fixed in 1658b9cc — rewrote the JSDoc so the `udid`/`device_id` path is described as refining to `tvos`/`android-tv` once the runtime-kind cache is warm, with only the `avdName`-only fallback left as unconditionally coarse.

packages/tool-server/src/http.ts
hubgan: Fixed in 1658b9cc — corrected the comment: once the emulator is up, a later `udid`/`device_id` call refines an Android TV AVD to `android-tv` from the call *after* the runtime-kind cache is warmed, not the first one — now consistent with `refineTvPlatform`.

packages/tool-server/src/utils/ios-devices.ts
hubgan: Fixed in 1658b9cc — closed the asymmetry rather than just documenting it: the tvOS `tv-control` factory now warms `runtimeKindCache` via a new `cacheSimulatorRuntimeKind(udid, kind)` helper (using the `runtimeKind` it already resolved from `listIosSimulators()`), mirroring how the Android TV factory warms via `getAndroidRuntimeKind`. A tv-remote-only Apple TV session is now attributed `tvos` from the second call, symmetric with Android TV. The `getCachedSimulatorRuntimeKind` docstring now lists the tv-remote (tv-control) warming path too.

PR #478, opened 2026-07-07
feat(telemetry): track Argent Lens design-review funnel

hubgan:
## What

Adds three privacy-safe telemetry events for the Argent Lens design-review flow, capturing the human-decision funnel that the generic `tool:*` path can't see (submission is an HTTP POST, and the `await_user_selection` tool is hidden in `argent lens` CLI sessions):

- **`lens:preview_opened`** — a human loaded the preview UI (`GET /preview/`), deduped per round.
- **`lens:round_completed`** — the human submitted their picks (once per round, never on a resubmit).
- **`lens:round_abandoned`** — a staged round was discarded before submit (window closed, `argent lens` exited mid-review, or the round superseded).

## Privacy

Payloads carry only aggregate counts, booleans, durations, and the device `platform` enum — never element names, comment text, variant code, file paths, or raw device ids. The store emits privacy-safe aggregate structs that are decoupled from the telemetry package; structural assignability to the event props is enforced at the `track()` call site (drift is a compile error).

## Event fields

- **`preview_opened`**: `round, element_count, variant_count, is_cli_session, platform`
- **`round_completed`**: `round, element_count, variant_count, annotation_count, element_comment_count, skipped_comment_count, has_global_comment, is_cli_session, had_parked_await, round_duration_ms, platform`
- **`round_abandoned`**: `round, element_count, variant_count, had_parked_await, is_cli_session, platform`

## Testing

- Unit tests for all three events: per-round dedup, exactly-once completion (no resubmit re-emit), abandoned vs happy-path roll, CLI up-front open, and content-agnostic privacy assertions.
- Verified end-to-end against a live tool-server: all three events fire with correct aggregate values and zero content leakage through the real sanitize pipeline.


packages/tool-server/src/preview.ts
```diff
@@ -685,6 +693,24 @@ export function createPreviewRouter(registry: Registry): Router {
...
+    if ((snap.proposals.length > 0 || snap.cliSession) && snap.round !== lastOpenedRound) {
+      lastOpenedRound = snap.round;
+      track("lens:preview_opened", {
+        round: snap.round,
+        element_count: snap.proposals.length,
+        variant_count: snap.proposals.reduce((n, p) => n + p.variants.length, 0),
+        is_cli_session: snap.cliSession,
+        platform: snap.device ? classifyDevice(snap.device) : undefined,
+      });
+    }
```
latekvo: `lens:preview_opened` is emitted only here, on `GET /preview/` (a full page load), deduped per round via `lastOpenedRound`. In a multi-round `argent lens` CLI session the preview window is loaded exactly once and then reused for the whole session: the UI's round-change handler swaps rounds client-side off the `/preview/variants` poll without navigating, and because `await_user_selection` is hidden during a CLI session the window is never re-foregrounded after the initial open, so `GET /preview/` never fires again.

The result is one `lens:preview_opened` per CLI session versus one `lens:round_completed` (and its abandoned counterpart) per round, so any per-round open-to-complete funnel derived from these events shows more completions than opens for every session with two or more rounds. The comment above states the per-round dedup "still count[s] each new round's first view", which does not hold for the reused CLI window. The non-CLI (MCP) path is unaffected: there each round respawns the window and re-hits `GET /`.

latekvo (approved):
Reviewed end-to-end. Built the monorepo (`tsc --build` clean), ran the telemetry and tool-server suites plus both `typecheck:tests` gates (green - the one full-suite `await-ui-element` failure is a pre-existing timing flake that reproduces identically on unmodified `main` and is untouched here), and drove the real store -> real `sanitize()` pipeline with adversarial secret-laden inputs.

The privacy contract holds: every field is an aggregate count/boolean/duration or the coarse `platform` enum (`classifyDevice` never echoes a raw device id, even for a malformed one), with `sanitize()` as the enforced backstop. `roundCompleted`/`roundAbandoned` are exactly-once and mutually exclusive, the aggregate counts match their doc comments (including the asymmetric `element_comment_count` vs `skipped_comment_count` split), the compile-time drift guard between the store stats and the telemetry props genuinely fails on a mismatch, listener registration/removal is symmetric, and emission is throw-safe.

One item worth resolving is left inline: `lens:preview_opened` undercounts rounds in multi-round `argent lens` CLI sessions, which skews the open-to-complete funnel these events exist to measure. Approving, since the events are individually correct and privacy-safe and the rest of the change is thoroughly verified.

packages/tool-server/src/utils/variant-proposals.ts
```diff
@@ -296,6 +375,23 @@ export class VariantProposalStore {
 
   /** Begin a fresh round, discarding the previous one's proposals/selections. */
   reset(): void {
+    // Drop-off telemetry: a round with staged proposals that is being discarded
+    // WITHOUT a submit is an abandonment (window closed, `argent lens` exited
```
latekvo: In the MCP flow nothing ever routes a walk-away through `reset()`: its only production callers are `autoRollIfCompleted` (requires `completed`, so the `!completed` guard excludes it) and the `setCliSession` transition boundary. Closing the preview window touches neither the store nor this method (the window manager and the renderer's unload path never reach it), an await timeout leaves the round staged, and `shutdown()` in index.ts exits without a reset. So for a non-CLI round the "window closed" trigger listed here only materializes if a CLI session later begins in the same process — the abandonment is then timestamped at that sweep, potentially much later — and never materializes if the process exits first. Verified against a live tool-server on this branch: a staged, never-submitted MCP round produced no `lens:round_abandoned` through any number of polls, and only emitted (correctly labeled `is_cli_session:false`) when a CLI session begin happened to reset it. As shipped, the drop-off leg of the funnel effectively measures only CLI sessions; MCP abandonment lands in the unattributable residue of opens minus completions — the very thing this event exists to disambiguate.

packages/tool-server/src/utils/variant-proposals.ts
```diff
@@ -24,7 +24,8 @@
  * parked is preserved across that roll and delivered on the next await.
  */
 
-import { TypedEventEmitter } from "@argent/registry";
+import { TypedEventEmitter, type Platform } from "@argent/registry";
+import { classifyDevice } from "./device-info";
```
latekvo: This branch predates #477 (f38f0e38), which widened telemetry `PLATFORMS` with `tvos`/`android-tv` and routes tool:* events through `refineTvPlatform` in http.ts so TV sessions stop being counted as iOS/Android. `classifyDevice` is shape-only and can never return a TV value, and all four lens emit sites (here and in preview.ts) use it directly, so after merging onto current main a Lens round against an Apple TV simulator emits `platform:"ios"` on lens events while the same session's tool:* events say `tvos` — reintroducing for the lens funnel exactly the attribution blindness #477 removed. The merge is clean and both the build and the telemetry/tool-server suites stay green on the merged tree (I ran them), so nothing surfaces this at merge time.

packages/tool-server/src/utils/variant-proposals.ts
```diff
@@ -508,6 +604,16 @@ export class VariantProposalStore {
...
     if (transitioned) this.events.emit("cliSessionChanged", active);
+    // Count every `argent lens` invocation. Each invocation POSTs
+    // /preview/cli-session {active:true} exactly once, so one emit per begin ==
+    // one per invocation. Guarding on `transitioned` would MISS a begin that
+    // arrives while the store is still marked active because a prior session's
+    // clean end never landed (e.g. that `argent lens` was killed) — an
+    // under-count of the exact metric this event exists for. An end is
+    // active=false, so it never emits. Privacy-safe: just the picker size.
+    if (active) {
+      this.events.emit("cliSessionStarted", { agent_choice_count: this.lensAgents.length });
```
latekvo, line 660: The kill scenario this comment invokes to justify the unguarded emit is also the scenario in which everything around the emit degrades. After a SIGKILLed `argent lens` (no `active:false` POST — and the server has no dead-session detection; the lens-stream close handler only unhooks its own listeners), the store stays `cliSession=true`, so the next begin has `transitioned=false` and skips the boundary reset above: the dead session's staged, unsubmitted round is neither counted as abandoned nor cleared, and bleeds into the new invocation's first round, inflating its `round_completed` element/variant counts. Until that next transition, `is_cli_session` on `preview_opened`/`round_completed`/`round_abandoned` also reads true for rounds actually driven by an MCP flow, and the un-rolled round keeps `lastOpenedRound` in preview.ts, so the new session's first `preview_opened` is suppressed. Reproduced the no-sweep-on-re-begin against a live server on this branch (a staged round survived a second `{active:true}` POST with no `round_abandoned`). `lens:cli_session_started` itself stays correct — it is the surrounding events that mislabel.

packages/tool-server/src/utils/variant-proposals.ts
```diff
@@ -175,12 +230,36 @@ type StoreEvents = {
...
+  roundAbandoned: (stats: RoundAbandonedStats) => void;
   /**
    * Emitted when a CLI-driven Lens session is begun or ended (`argent lens`).
    * The tool-server's window manager listens: begin ⇒ open the window now, end
    * ⇒ close it. Carries the new active state so listeners need not re-snapshot.
    */
   cliSessionChanged: (active: boolean) => void;
+  /**
+   * Emitted once on a CLI session BEGIN transition (not on end, not on a
+   * re-begin while already active) — i.e. once per `argent lens` invocation.
```
latekvo: This docblock says the event fires "not on a re-begin while already active", but `setCliSession` deliberately emits on every `active=true` call — the comment at the emit site and the test ("emits once per begin (every invocation) ... never on session end") both pin the opposite behavior, and a no-transition re-begin does emit (verified live). A future reader taking this docblock at face value and adding a `transitioned` guard would reintroduce the exact under-count the emit-site comment warns about.

packages/tool-server/src/utils/variant-proposals.ts
```diff
@@ -163,6 +164,60 @@ export interface StoreSnapshot {
...
 }
 
+/**
+ * Privacy-safe aggregate shape of one finished round, carried by the
+ * `roundCompleted` store event so the tool-server's composition layer can emit
+ * it as `lens:round_completed` telemetry WITHOUT the store depending on the
+ * telemetry package. Only counts/booleans/durations — never element names,
+ * comment text, variant code, file paths, or device identifiers. Structurally
+ * assignable to `LensRoundCompletedProps` in @argent/telemetry; if the two
+ * drift, the `track()` call site in index.ts fails to compile.
```
latekvo: "Fails to compile" holds in only one direction. The relay call sites in index.ts pass these stats as variables, so excess-property checking doesn't apply: adding a field to this struct (and its emit) without extending `LensRoundCompletedProps` compiles cleanly, and `sanitize()` then silently drops the new field — that drift shows up as a metric that never arrives, not as a compile error. Only missing, renamed, or re-typed fields fail the build.

packages/tool-server/src/utils/variant-proposals.ts
```diff
@@ -691,6 +804,36 @@ export class VariantProposalStore {
...
     this.events.emit("changed");
     this.events.emit("selectionSubmitted");
+    // Count the human decision exactly once per round: a resubmit updates the
+    // same frozen outcome in place (the submit route is unauthenticated and the
+    // UI re-enables its button), so `!wasResubmit` guards against double-counting.
+    // Everything here is a privacy-safe aggregate derived from the just-frozen
+    // outcome and the still-intact proposals (reset happens later, on the next
+    // propose/roll), so no element names, comment text, or paths ever leave.
+    const outcome = this.lastOutcome;
+    if (!wasResubmit && outcome) {
```
latekvo, line 883: A stale duplicate submit mints a completion for a round the human never reviewed. The submit route is unauthenticated and carries no round token, and in the non-CLI path the UI neither clears the sticky usage flags after a successful submit (that block is `if (vp.cliSession)`-guarded) nor stays disabled (`completeBtn.disabled = false` runs in the `finally`). So: the human submits round N from a browser tab, the agent proposes again (auto-roll to N+1), and a second click landing before the tab's next poll passes the proposals-staged check with every selection filtered out. Reproduced against a live tool-server on this branch: the second submit returned `{ok:true, resolved:0}` and emitted `lens:round_completed` for round N+1 with `inspector_used:true`/`offscreen_revealed:true` carried over from round N — a phantom completion wearing the previous round's usage flags, for a round whose real review is still pending. The empty-completion itself is pre-existing store behavior; what's new is that it now manufactures a funnel event.

packages/tool-server/src/index.ts
```diff
@@ -223,6 +251,9 @@ export function start(): void {
     variantProposalStore.events.off("awaitParked", onAwaitParked);
     variantProposalStore.events.off("selectionSubmitted", onSelectionSubmitted);
     variantProposalStore.events.off("cliSessionChanged", onCliSessionChanged);
+    variantProposalStore.events.off("roundCompleted", onRoundCompleted);
+    variantProposalStore.events.off("roundAbandoned", onRoundAbandoned);
+    variantProposalStore.events.off("cliSessionStarted", onCliSessionStarted);
```
latekvo: These listeners come off at the top of `shutdown()`, but the HTTP server keeps accepting requests until `server.close()` at the bottom — after the owned-simulator drain (real `simctl shutdown` calls, potentially seconds) and the telemetry capture/drain. The submit and cli-session routes touch only the store, so during that window a human clicking Complete gets a 200 and the store emits `roundCompleted` to zero listeners: the event is silently lost while `toolserver:stop` from the very same shutdown is still delivered. The same applies to a `roundAbandoned` from an `argent lens` exiting concurrently. Separately, a round still staged at process exit produces no abandonment signal at all, since shutdown never resets the store — "server died mid-review" is invisible to the funnel.

packages/tool-server/src/index.ts
```diff
@@ -207,9 +212,32 @@ export function start(): void {
...
+  // The generic tool:* path counts per-tool-call and lens:preview_opened counts
+  // per-round, so neither can answer "how many `argent lens` runs / unique users"
+  // — this per-invocation marker does. Carries only an aggregate count.
+  const onCliSessionStarted = (stats: CliSessionStartedStats): void => {
+    telemetryTrack("lens:cli_session_started", stats);
+  };
   variantProposalStore.events.on("awaitParked", onAwaitParked);
   variantProposalStore.events.on("selectionSubmitted", onSelectionSubmitted);
   variantProposalStore.events.on("cliSessionChanged", onCliSessionChanged);
+  variantProposalStore.events.on("roundCompleted", onRoundCompleted);
```
latekvo, line 298: Nothing exercises this relay at runtime: startup-telemetry.test.ts boots `start()` with `track` mocked but never emits a store event, and no other test asserts that a given store event maps to a given telemetry event name. Typos are compile-caught, but two failure classes ship green: dropping one of these registrations, and cross-wiring — `RoundAbandonedStats` is structurally assignable to `LensPreviewOpenedProps` (every required prop present), so relaying `roundAbandoned` as `"lens:preview_opened"` type-checks and would silently mislabel the funnel.

packages/tool-server/src/preview.ts
```diff
@@ -255,7 +293,13 @@ export function createPreviewRouter(registry: Registry): Router {
   // Live proposal state for the UI to poll. Invisible to MCP (only /tools).
   router.get("/variants", (_req: Request, res: Response) => {
     res.set("Cache-Control", "no-store");
-    res.json(variantProposalStore.snapshot());
+    // This poll is only ever issued by an OPEN preview window (the agent mutates
```
latekvo: "Only ever issued by an OPEN preview window" doesn't hold for browsers, which are a documented consumer of this UI: `notifyWindowUnavailable` tells the agent to hand the human the `/preview/` URL, and `argent lens` prints it to the terminal. The page polls on an unconditional 1.2s `setInterval` with no `visibilitychange` gating and no teardown, so a forgotten background tab keeps claiming each later round's `preview_opened` (the dedup only collapses it to one phantom per round, and the server cannot tell a tab's poll from the Electron renderer's). Every round staged after the tab is abandoned counts as opened with nobody looking — verified that a windowless HTTP poller mints these events against a live server on this branch. This depresses open-to-completed conversion for as long as such a tab lives.

packages/telemetry/src/events.ts
```diff
@@ -142,6 +142,100 @@ export interface ToolserverStopProps extends FailureTelemetryProps {
...
+export interface LensPreviewOpenedProps {
+  /** Proposal round on screen when the preview UI was loaded. */
+  round: number;
+  /** Elements with staged proposals at open time (0 for a CLI up-front open). */
+  element_count: number;
+  /** Total variants staged across all elements at open time (0 for a CLI up-front open). */
+  variant_count: number;
+  /** Whether an `argent lens` CLI session owns the window (vs the MCP path). */
+  is_cli_session: boolean;
+  /** Device platform the variants target; omitted until a device is bound (e.g. a CLI up-front open). */
```
latekvo: Two caveats on these fields beyond the documented round-1 zero. For any poll-claimed round — every CLI round after the first, or an MCP window reused within the close delay — the counts freeze at the first ~1.2s poll tick, i.e. typically after the agent's first `propose_variant` of the round but before the rest, so they systematically under-report relative to fresh-window MCP opens (which fire on `GET /` after staging settles), and the two populations are not comparable. And `platform` is not reliably omitted for a CLI up-front open: the store's `device` deliberately survives `reset()`, so a CLI session begun after an earlier flow on another device reports that flow's platform — on a live run of this branch, the CLI up-front open emitted `platform:"chromium"` (stale from a prior round) alongside `element_count:0`.

packages/telemetry/src/events.ts
```diff
@@ -142,6 +142,100 @@ export interface ToolserverStopProps extends FailureTelemetryProps {
...
+// Fired ONCE per `argent lens` CLI invocation, at the session-begin transition.
+// The generic tool:* path counts the agent's propose_variant/await_user_selection
+// calls (which fire many times per session), and lens:preview_opened fires once
+// PER ROUND — so neither can count how many times a human ran `argent lens`.
+// This is the per-invocation marker: a plain count of these events is the
+// invocation total, and distinct telemetry ids over them are the unique-user
+// population for the tool. Privacy-safe: only an aggregate count, no PII.
+export interface LensCliSessionStartedProps {
+  /**
+   * Coding-agent choices offered in the window's picker (0/1 when no picker is
```
latekvo: The "0/1" here overstates what the CLI can send: `planAgent` in lens.ts resolves the agent CLI-side (and the POST sends an empty `agents` array, count 0) for `--agent`, for a remembered-and-still-installed choice, and for a single installed agent; the `choose` list it forwards is only ever length >= 2. So 1 is unreachable via `argent lens` — real values are 0 (no picker) or >= 2 (picker shown) — and a count of 1 in the data would indicate a hand-crafted POST rather than the documented single-installed-agent case.

packages/telemetry/src/events.ts
```diff
@@ -142,6 +142,100 @@ export interface ToolserverStopProps extends FailureTelemetryProps {
...
 }
 
+// Lens (variant-proposal) events
+//
+// The agent CALLS `propose_variant` / `await_user_selection`, which the generic
+// tool:invoke/complete/fail path already counts — but those calls only prove the
+// AGENT acted. They can't tell us whether a HUMAN opened the preview or what they
+// decided. These events capture the human side of the funnel: preview loaded →
+// round decided OR abandoned. All carry only privacy-safe aggregate counts /
+// booleans / durations plus the device `platform` enum — never element names,
```
latekvo, line 184: Telemetry.md (the versioned "Argent Privacy Notice", with its own changes section) enumerates collected-data categories under "What we collect", and none of its bullets covers what these events add: human design-review interaction — per-round comment/annotation counts, decision durations, inspector/off-screen usage, and the installed-coding-agent count. Previous event additions instrumented categories the notice already listed (tool invocations, process lifetime); this is the first that expands the kind of data collected without extending the notice (which is also already stale on platforms — it names "Android, iOS, or a Chromium-based target").

packages/tool-server/test/variant-proposals.test.ts
```diff
@@ -510,3 +530,306 @@ describe("VariantProposalStore — Lens-owned devices", () => {
...
+    // never landed (e.g. `argent lens` was killed) — is a real new invocation
+    // and MUST count. A transition-only guard would miss it and under-count the
+    // very metric this event exists for. Its payload must also reflect THIS
+    // begin's picker size, not a stale one.
+    s.setCliSession(true, [{ id: "cursor", name: "Cursor" }]); // invocation 3, no transition (1 agent)
+    expect(stats.map((x) => x.agent_choice_count)).toEqual([1, 0, 1]);
+  });
+});
+
+describe("VariantProposalStore — roundAbandoned telemetry event", () => {
```
latekvo, line 822: These cases all trigger the event via a direct `reset()`, so `is_cli_session` is only ever asserted false. The flagship path the event was added for — `argent lens` exited mid-review, i.e. `setCliSession(false)` with a staged round — is pinned nowhere, and its correct `is_cli_session:true` attribution depends on ordering inside `setCliSession` (the transition reset runs before `this.cliSession = active` is assigned). Hoisting that assignment above the transition block would flip the label on every CLI-exit abandonment with the whole suite staying green. The begin-direction sweep (a leftover non-CLI round swept at session start, labeled false) is equally unpinned.

packages/tool-server/src/preview.ts
latekvo: Fixed in 379f92ba. `lens:preview_opened` is now driven by an explicit, visibility-gated client trigger (`POST /preview/opened`) that the UI posts when it renders a round in a visible window, so a reused CLI window emits one open per round. `GET /` and the `/variants` poll no longer emit.

packages/tool-server/src/utils/variant-proposals.ts
latekvo: Fixed in 379f92ba. The `reset()` comment now lists the real triggers (CLI session boundary, superseded round, shutdown flush) and is explicit that a bare-MCP window-close walk-away is not routed through `reset()`. Server shutdown now flushes a still-staged round as an abandonment (`flushAbandonedRound`), so "server died mid-review" is captured instead of landing in the unattributable residue.

packages/tool-server/src/utils/variant-proposals.ts
latekvo: Fixed in 379f92ba. Merged `main` and factored `refineTvPlatform` into a shared `utils/telemetry-platform` (http.ts now imports it too). All four lens emit sites go through `classifyDeviceForTelemetry`, so a Lens session on a TV target reports `tvos`/`android-tv` consistently with its `tool:*` events once the runtime-kind cache is warm.

packages/tool-server/src/utils/variant-proposals.ts
latekvo: Fixed in 379f92ba. `setCliSession` now sweeps leftover round state on every begin (`active || transitioned`), not only a clean transition. A killed prior session's staged round is discarded as an abandonment (`is_cli_session:true`) and can no longer bleed into the new invocation's first round. Covered by a dead-session re-begin test.

packages/tool-server/src/utils/variant-proposals.ts
latekvo: Fixed in 379f92ba. The docblock now states it fires on every begin — including a re-begin while still marked active from a killed predecessor — matching the emit-site comment and the tests.

packages/tool-server/src/utils/variant-proposals.ts
latekvo: Fixed in 379f92ba. Added a bidirectional compile-time key-set guard in index.ts (`SameKeys` over the three relayed stat structs vs their `EventPropertyMap` entries), so an added/removed/renamed field on either side now fails the build; the struct comment is corrected to say so. Verified the guard bites by temporarily adding a probe field (tsc errored at the guard).

packages/tool-server/src/utils/variant-proposals.ts
latekvo: Fixed in 379f92ba. The submit now carries the round the UI built it against; the store rejects a submit whose round has since rolled (returns `stale`, no completion and no emit). The route forwards the token, the UI sends it, and the per-round usage flags reset on both paths. Verified end-to-end against a live router: the stale click returns `{stale:true, resolved:0}` and emits no phantom `lens:round_completed`.

packages/tool-server/src/index.ts
latekvo: Fixed in 379f92ba. The store-event relays now stay attached until after `server.close()`, so a submit that completes while the server is still accepting is relayed; and shutdown flushes a staged round as an abandonment before the telemetry drain. A submit that lands after the drain begins is inherently unobservable (the queue is closing) — noted honestly in the comment.

packages/tool-server/src/index.ts
latekvo: Fixed in 379f92ba. Added `lens-relay-telemetry.test.ts`: it boots `start()` and emits each store event, asserting it lands as the correct telemetry event name — catching both a dropped registration and cross-wiring (incl. the `RoundAbandonedStats`→`lens:preview_opened` mislabel).

packages/tool-server/src/preview.ts
latekvo: Fixed in 379f92ba. The `/variants` poll no longer emits. `lens:preview_opened` is posted by the client only when `document.visibilityState === 'visible'`, so a forgotten background tab no longer claims each later round's open.

packages/telemetry/src/events.ts
latekvo: Fixed in 379f92ba. All opens now go through the single visibility-gated client trigger, so MCP and CLI rounds are sampled at the same point (no fresh-load vs poll-tick split — the two populations are comparable). `platform` is omitted whenever `element_count` is 0, so a CLI up-front open can no longer inherit a prior flow's (surviving) device. Field docs updated; verified the up-front open emits `element_count:0` with no `platform`.

packages/telemetry/src/events.ts
latekvo: Fixed in 379f92ba. The doc now says the real values are 0 (no picker) or >= 2 (picker shown) — a lone installed agent is auto-resolved CLI-side and posts an empty list — and that a 1 in the data would indicate a hand-crafted POST rather than the single-installed-agent case.

packages/telemetry/src/events.ts
latekvo: Fixed in 379f92ba. Added a "Design review (Argent Lens)" category to Telemetry.md covering the per-round aggregate comment/annotation counts, decision durations, inspector/off-screen usage, and the coding-agent count, and bumped the version. (The platform bullet was already refreshed to tvOS/Android TV/VegaOS on `main`, now merged in.)

packages/tool-server/test/variant-proposals.test.ts
latekvo: Fixed in 379f92ba. Added tests pinning the CLI-exit abandonment (`is_cli_session:true`, which depends on `reset()` reading `cliSession` before `setCliSession` flips it), the begin-direction sweep (`is_cli_session:false`), and the dead-session re-begin sweep.

latekvo (approved):
Re-reviewed after the fixes in 379f92ba (＋ a `main` merge to resolve a shutdown conflict). All fourteen inline findings are addressed and every one is verified — not just believed:

- **preview_opened** is now a single, visibility-gated client signal (`POST /preview/opened`): one open per round in a reused CLI window, no background-tab overcount, consistent count sampling across MCP/CLI, and `platform` omitted for a zero-proposal open so a surviving device can't attribute a CLI up-front open. `GET /` and the `/variants` poll no longer emit.
- **TV attribution** goes through a shared `refineTvPlatform` (factored out of http.ts), so a Lens session on a TV target reports `tvos`/`android-tv` like its `tool:*` events.
- **Phantom completion** is closed by a round token the store rejects when the round has rolled (route forwards it, UI sends it, usage flags reset on both paths).
- **Dead CLI session**: leftover round state is swept on every begin, so a killed predecessor's staged round is counted abandoned and can't bleed forward.
- **Shutdown**: relays stay attached until after `server.close()`, and a still-staged round is flushed as an abandonment before the drain.
- **Drift guard** is now bidirectional (compile-time key-set check); the docblocks (cliSessionStarted, drift, agent_choice_count 0/1) and the reset() abandonment comment are corrected; the privacy notice gains a design-review category.

Verification: `tsc --build` clean; telemetry (259) and full tool-server (2201) suites green; both `typecheck:tests` gates, ESLint, and Prettier pass; the drift guard proven to bite via a probe field; and a live end-to-end run against the real preview router + store + relay + `sanitize()` pipeline confirmed each event fires with correct aggregate values, the stale submit is rejected with no phantom completion, the CLI-exit abandonment is labeled `is_cli_session:true`, the up-front open omits the stale platform, and no element name / comment text leaks. CI is green on the head commit.

Approving.

PR #480, opened 2026-07-08
fix(tool-server): cap concurrent simctl device listing

### LINKED ISSUE #327: tool-server.cjs is orphaned (reparented to launchd) when the argent mcp process that spawned it exits
**Environment:** `@swmansion/argent` 0.11.0, Node v24.12.0 (nvm), macOS 26.5, Apple Silicon

## Bug

When the `argent mcp` process that **first spawns** `tool-server.cjs` exits (e.g. its MCP client is a short-lived headless run that finishes), the tool-server does not exit. It re-parents to launchd (ppid 1) and runs indefinitely. There appears to be no parent-death detection (stdin EOF / ppid watchdog) and no idle timeout.

## Repro (verified)

1. Ensure no `tool-server.cjs start` process is running:
   `ps -Ao pid,ppid,command | grep '[t]ool-server.cjs'` → empty.
2. In a directory whose `.mcp.json` registers argent as a stdio server:
   ```json
   {"mcpServers":{"argent":{"type":"stdio","command":"argent","args":["mcp"]}}}
   ```
3. Run any short-lived MCP client from that directory, e.g.:
   `echo "say hi" | claude -p --model haiku --output-format text`
4. After the client exits: `ps -Ao pid,ppid,command | grep '[t]ool-server.cjs'` shows a tool-server with **ppid 1** that never exits.

## Why it's intermittent

If a tool-server already exists, a new `argent mcp` reuses it and exits cleanly — no leak. Only the client that first spawns the tool-server orphans it.

Additionally, spawn is **not mutually exclusive**: multiple `argent mcp` processes starting concurrently each spawn their own tool-server (we observed 4 simultaneous spawns within the same second after killing the existing one). Combined with the missing cleanup, automation that repeatedly launches short-lived MCP clients accumulates orphans over time.

## Real-world impact

An automation daemon on my machine invoked headless `claude -p` ~2×/minute in a project with argent configured. Within hours: **123 orphaned tool-servers** and, with simulators in use, **330 concurrent `simctl`/CoreSimulator processes**, load average **445**, memory exhausted — the machine was barely usable (~30% system CPU from simctl spawning alone).

With no simulator booted an orphan idles cheaply (it just never exits); the simctl storm occurs when simulators are in use.

## Suggested fixes

- tool-server exits when its last client disconnects (stdin/IPC EOF), and/or watches `process.ppid` and exits when re-parented to 1, and/or an idle TTL;
- `argent mcp` reaps the tool-server it spawned on shutdown;
- serialize tool-server spawn (lockfile / port probe) so concurrent clients can't each create one.

Any one of the first group would cap the damage; spawn serialization would make the leak self-limiting even if cleanup fails.

**Workaround** for anyone hitting this: kill orphans with
`for pid in $(ps -Ao pid,ppid,command | awk '/argent\/dist\/tool-server/ && $2 == 1 {print $1}'); do kill $pid; done`
and pass `--strict-mcp-config --mcp-config '{"mcpServers":{}}'` to headless `claude -p` runs that don't need tools.

iroiro147:
## Summary

Fixes #327.

Adds a host-wide filesystem lock around local `xcrun simctl list devices --json` so multiple Argent tool-server processes do not spawn concurrent CoreSimulator discovery calls. Both `listIosSimulators()` and the simulator watcher now share `readSimctlDevices()`.

The lock path also handles the two race cases found during review:

- fresh empty/partial lock files are treated as live using filesystem `mtimeMs` fallback
- stale lock cleanup is serialized through a sidecar `.cleanup` lock so competing stale cleaners cannot delete a fresh holder's lock

## Tests

- `npm test -w @argent/tool-server -- --run test/ios-devices-runtime-kind.test.ts test/list-devices-deadline.test.ts test/simulator-watcher-init-failure.test.ts test/list-devices.test.ts`
- `npm run build`
- `npm run build -w @argent/tool-server`
- `npm run lint -- --no-warn-ignored packages/tool-server/src/utils/ios-devices.ts packages/tool-server/src/utils/simctl-config.ts packages/tool-server/src/utils/simulator-watcher.ts packages/tool-server/src/tools/devices/list-devices.ts packages/tool-server/test/ios-devices-runtime-kind.test.ts packages/tool-server/test/list-devices-deadline.test.ts packages/tool-server/test/simulator-watcher-init-failure.test.ts`
- `npx prettier --check packages/tool-server/src/utils/ios-devices.ts packages/tool-server/src/utils/simctl-config.ts packages/tool-server/src/utils/simulator-watcher.ts packages/tool-server/src/tools/devices/list-devices.ts packages/tool-server/test/ios-devices-runtime-kind.test.ts packages/tool-server/test/list-devices-deadline.test.ts packages/tool-server/test/simulator-watcher-init-failure.test.ts`
- `git diff --check`

Note: I did not use the full `@argent/tool-server` Vitest suite as a local gate because this environment blocks several unrelated socket/process-list tests with `listen EPERM` / host-tool permission errors. The focused tool-server slice above is clean.

### DIFF
--- a/packages/tool-server/src/utils/ios-devices.ts
+++ b/packages/tool-server/src/utils/ios-devices.ts
@@ -1,4 +1,5 @@
 import { execFile } from "node:child_process";
+import { randomUUID } from "node:crypto";
 import { open, readFile, stat, unlink, type FileHandle } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
@@ -33,7 +34,24 @@ export interface SimctlOutput {
 }
 
 const DEFAULT_SIMCTL_LIST_DEVICES_LOCK_PATH = join(tmpdir(), "argent-simctl-list-devices.lock");
+const SIMCTL_LIST_DEVICES_LOCK_LIVE_OWNER_MAX_MS = SIMCTL_LIST_DEVICES_LOCK_STALE_MS * 4;
 let simctlListDevicesLockPath: string | null = DEFAULT_SIMCTL_LIST_DEVICES_LOCK_PATH;
+
+interface SimctlListDevicesLockMetadata {
+  createdAt: number;
+  ownerId: string;
+  pid: number;
+}
+
+class SimctlListDevicesLockError extends Error {
+  constructor(
+    message: string,
+    readonly cause?: unknown
+  ) {
+    super(message);
+    this.name = "SimctlListDevicesLockError";
+  }
+}
 
 function sleep(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms));
@@ -43,28 +61,153 @@ function isNodeError(err: unknown): err is NodeJS.ErrnoException {
   return err instanceof Error && "code" in err;
 }
 
+function isSimctlListDevicesLockError(err: unknown): err is SimctlListDevicesLockError {
+  return err instanceof SimctlListDevicesLockError;
+}
+
+function isProcessAlive(pid: number): boolean {
+  if (!Number.isInteger(pid) || pid <= 0) return false;
+  try {
+    process.kill(pid, 0);
+    return true;
+  } catch (err) {
+    return isNodeError(err) && err.code === "EPERM";
+  }
+}
+
+function isLockOwnedByLiveProcess(metadata: SimctlListDevicesLockMetadata, now: number): boolean {
+  if (now - metadata.createdAt >= SIMCTL_LIST_DEVICES_LOCK_LIVE_OWNER_MAX_MS) return false;
+  return isProcessAlive(metadata.pid);
+}
+
+function parseSimctlListDevicesLockMetadata(raw: string): SimctlListDevicesLockMetadata | null {
+  try {
+    const parsed = JSON.parse(raw) as {
+      createdAt?: unknown;
+      ownerId?: unknown;
+      pid?: unknown;
+    };
+    if (
+      typeof parsed.createdAt === "number" &&
+      typeof parsed.ownerId === "string" &&
+      typeof parsed.pid === "number"
+    ) {
+      return {
+        createdAt: parsed.createdAt,
+        ownerId: parsed.ownerId,
+        pid: parsed.pid,
+      };
+    }
+  } catch {
+    return null;
+  }
+  return null;
+}
+
+async function readSimctlListDevicesLockMetadata(
+  lockPath: string
+): Promise<SimctlListDevicesLockMetadata | null> {
+  try {
+    return parseSimctlListDevicesLockMetadata(await readFile(lockPath, "utf8"));
+  } catch (err) {
+    if (isNodeError(err) && err.code === "ENOENT") return null;
+    throw err;
+  }
+}
+
+async function getSimctlListDevicesLockCreatedAt(lockPath: string): Promise<number | undefined> {
+  const metadata = await readSimctlListDevicesLockMetadata(lockPath);
+  if (metadata) return metadata.createdAt;
+
+  try {
+    return (await stat(lockPath)).mtimeMs;
+  } catch (err) {
+    if (isNodeError(err) && err.code === "ENOENT") return undefined;
+    throw err;
+  }
+}
+
+async function acquireSimctlListDevicesLock(
+  lockPath: string,
+  now: number
+): Promise<SimctlListDevicesLockMetadata | null> {
+  const metadata: SimctlListDevicesLockMetadata = {
+    createdAt: now,
+    ownerId: randomUUID(),
+    pid: process.pid,
+  };
+  let handle: FileHandle | undefined;
+
+  try {
+    handle = await open(lockPath, "wx");
+    await handle.writeFile(JSON.stringify(metadata));
+    return metadata;
+  } catch (err) {
+    if (handle) {
+      await handle.close().catch(() => {});
+      await unlink(lockPath).catch(() => {});
+    }
+    if (isNodeError(err) && err.code === "EEXIST") return null;
+    throw new SimctlListDevicesLockError(
+      `failed to acquire simctl list devices lock at ${lockPath}`,
+      err
+    );
+  } finally {
+    await handle?.close().catch(() => {});
+  }
+}
+
+async function releaseSimctlListDevicesLock(
+  lockPath: string,
+  metadata: SimctlListDevicesLockMetadata
+): Promise<void> {
+  const currentMetadata = await readSimctlListDevicesLockMetadata(lockPath);
+  if (currentMetadata?.ownerId !== metadata.ownerId) return;
+
+  try {
+    await unlink(lockPath);
+  } catch (err) {
+    if (!isNodeError(err) || err.code !== "ENOENT") {
+      throw new SimctlListDevicesLockError(
+        `failed to release simctl list devices lock at ${lockPath}`,
+        err
+      );
+    }
+  }
+}
+
+async function removeStaleCleanupLock(cleanupLockPath: string, now: number): Promise<void> {
+  const createdAt = await getSimctlListDevicesLockCreatedAt(cleanupLockPath);
+  if (createdAt === undefined || now - createdAt < SIMCTL_LIST_DEVICES_LOCK_STALE_MS) return;
+
+  const metadata = await readSimctlListDevicesLockMetadata(cleanupLockPath);
+  if (metadata && isLockOwnedByLiveProcess(metadata, now)) return;
+
+  try {
+    await unlink(cleanupLockPath);
+  } catch (err) {
+    if (!isNodeError(err) || err.code !== "ENOENT") {
+      throw new SimctlListDevicesLockError(
+        `failed to remove stale simctl list devices cleanup lock at ${cleanupLockPath}`,
+        err
+      );
+    }
+  }
+}
+
 async function removeStaleSimctlListDevicesLock(lockPath: string, now: number): Promise<void> {
   const cleanupLockPath = `${lockPath}.cleanup`;
-  let cleanupHandle: FileHandle | undefined;
-  try {
-    cleanupHandle = await open(cleanupLockPath, "wx");
-    await cleanupHandle.writeFile(JSON.stringify({ pid: process.pid, createdAt: now }));
-  } catch (err) {
-    if (cleanupHandle) {
-      await cleanupHandle.close().catch(() => {});
-      await unlink(cleanupLockPath).catch(() => {});
-    }
-    if (isNodeError(err) && err.code === "EEXIST") return;
-    throw err;
+  let cleanupMetadata = await acquireSimctlListDevicesLock(cleanupLockPath, now);
+  if (!cleanupMetadata) {
+    await removeStaleCleanupLock(cleanupLockPath, now);
+    cleanupMetadata = await acquireSimctlListDevicesLock(cleanupLockPath, Date.now());
+    if (!cleanupMetadata) return;
   }
 
   try {
     await removeStaleSimctlListDevicesLockWithCleanupLock(lockPath, now);
   } finally {
-    await cleanupHandle.close().catch(() => {});
-    await unlink(cleanupLockPath).catch((err: unknown) => {
-      if (!isNodeError(err) || err.code !== "ENOENT") throw err;
-    });
+    await releaseSimctlListDevicesLock(cleanupLockPath, cleanupMetadata);
   }
 }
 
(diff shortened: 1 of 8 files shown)


packages/tool-server/src/utils/ios-devices.ts
```diff
@@ -19,21 +28,136 @@ interface SimctlDevice {
...
+  try {
+    await removeStaleSimctlListDevicesLockWithCleanupLock(lockPath, now);
+  } finally {
+    await cleanupHandle.close().catch(() => {});
+    await unlink(cleanupLockPath).catch((err: unknown) => {
+      if (!isNodeError(err) || err.code !== "ENOENT") throw err;
+    });
+  }
+}
+
```
hubgan: The main lock self-heals via the createdAt/mtime + 15s staleness check,
but the `.cleanup` sidecar does not. Line 57 is a blind `if (EEXIST) return` with no
age check, and line 65 is the only code that ever removes the file. If a process is
SIGKILL'd / crashes in the window between `open(cleanupLockPath, "wx")` and the `finally` unlink, `<lock>.cleanup` is orphaned permanently.

packages/tool-server/src/utils/ios-devices.ts
hubgan, line 313: Before this PR, listing iOS sims only needed to run simctl. Now it also needs to create a file in the temp folder first. If that folder is read-only or full (rare, but real on locked-down/CI machines), creating the file fails, and the code treats that identically to "no simulators found" and returns an empty list with no error message. So your booted device just disappears from the list with zero explanation. The same silent-empty happens if 3+ processes fight over the lock file and a loser times out after 12s.

packages/tool-server/src/utils/ios-devices.ts
```diff
@@ -19,21 +28,136 @@ interface SimctlDevice {
...
+  }
+
+  if (now - createdAt < SIMCTL_LIST_DEVICES_LOCK_STALE_MS) return;
+
+  try {
+    await unlink(lockPath);
+  } catch (err) {
+    if (!isNodeError(err) || err.code !== "ENOENT") throw err;
+  }
+}
```
hubgan, line 237: Both delete sites `unlink(lockPath)` by path, ignoring only ENOENT, without
confirming the file is still the same lock they created/inspected. The 15s-stale vs
10s-timeout margin assumes a live holder always releases by ~10s. But the exact scenario
this lock defends against - a wedged CoreSimulatorService - can leave a `simctl` stuck
past 15s while its owner is still alive. Interleaving:

1. P1 holds the lock (createdAt=T0), its `simctl` wedges past its 10s timeout, still awaited.
2. At T0+15s a waiter P2 sees createdAt aged >=15s, unlinks P1's LIVE lock, acquires, and runs a SECOND concurrent `simctl` - the single-simctl guarantee is violated.
3. When P1's `simctl` finally dies, P1's `finally` unlink deletes whatever is at the path, which is now P2's (or a successor's) fresh lock - an ABA delete.

packages/tool-server/test/ios-devices-runtime-kind.test.ts
hubgan, line 1: The tests only cover the case where simctl succeeds. So if someone later accidentally broke the error path, every test would still pass and iOS listing would quietly freeze for 15s. It's not a bug, but one of the most important behavior has no test.

hubgan (commented):
Thanks for taking the time to work on this. I appreciate the contribution. There are a few areas that need some additional refinement before this is ready to merge.

iroiro147:
Thanks for the detailed review, Hubert. I pushed a follow-up commit that reworks the simctl-list lock handling around explicit ownership metadata.

What changed:

- the main lock and cleanup sidecar now carry an owner token, pid, and createdAt
- stale cleanup no longer removes a lock while the recorded owner process is still within the live-owner grace window
- very old locks/cleanup sidecars can still be recovered even if the recorded pid is alive again, which covers pid reuse after a crash
- release only unlinks when the current owner token still matches, so an old holder cannot delete a successor lock
- `listIosSimulators()` no longer hides lock acquisition/wait failures as an empty simulator list, while simctl/xcrun failures still preserve the existing empty-list fallback
- the aggregate `list-devices` tool catches that iOS branch failure so Android/Chromium/Vega results are still returned
- added coverage for live-owner stale locks, pid-reuse recovery, dead-owner stale locks, orphaned cleanup sidecars, lock acquisition failures, simctl failure fallback, and aggregate list-devices degradation

Re-ran:

- `npm test -w @argent/tool-server -- --run test/ios-devices-runtime-kind.test.ts test/list-devices.test.ts test/list-devices-deadline.test.ts test/simulator-watcher-init-failure.test.ts`
- `npm run build -w @argent/tool-server`
- `npm run lint`
- `npx prettier --check packages/tool-server/src/utils/ios-devices.ts packages/tool-server/src/tools/devices/list-devices.ts packages/tool-server/test/ios-devices-runtime-kind.test.ts packages/tool-server/test/list-devices.test.ts`
- `git diff --check`

packages/tool-server/src/utils/ios-devices.ts
iroiro147: Addressed in 8e53dfa. The cleanup sidecar now carries the same ownership metadata as the main lock, checks staleness plus live-owner state, and can recover an orphaned sidecar before retrying cleanup acquisition. Added coverage for orphaned and very-old/PID-reused cleanup sidecars.

packages/tool-server/src/utils/ios-devices.ts
iroiro147: Addressed in 8e53dfa. Lock acquisition and wait failures now propagate from listIosSimulators() instead of becoming an empty simulator list. The aggregate list-devices tool catches only that iOS branch failure so Android, Chromium, and Vega results remain available; simctl/xcrun failures retain the existing empty-list fallback. Both paths have regression coverage.

packages/tool-server/src/utils/ios-devices.ts
iroiro147: Addressed in 8e53dfa. Locks now include an owner token, PID, and creation time. Stale cleanup preserves a lock while its recorded owner is alive within the grace window, very old locks remain recoverable for PID reuse, and release unlinks only when the on-disk owner token still matches, preventing an old holder from deleting a successor lock. Added live-owner, dead-owner, PID-reuse, and release-safety coverage.

packages/tool-server/test/ios-devices-runtime-kind.test.ts
iroiro147: Addressed in 8e53dfa. Added explicit tests showing lock acquisition failures are surfaced, simctl/xcrun failures still produce the established empty-list fallback, and aggregate list-devices degrades only the iOS branch while returning other platforms.

packages/tool-server/src/utils/ios-devices.ts
```diff
@@ -51,7 +305,8 @@ export async function listIosSimulators(): Promise<IosSimulator[]> {
       }
     }
     return out;
-  } catch {
+  } catch (err) {
+    if (isSimctlListDevicesLockError(err)) throw err;
     return [];
```
hubgan, line 310: Before this PR, `listIosSimulators()` could never throw — any failure ended up as an empty list, and every caller was written to treat "no simulators found" as a safe outcome. This change makes it re-throw `SimctlListDevicesLockError` when the 12-second lock wait runs out. Only the `list-devices` caller was updated to catch that.

The other callers reach this function indirectly through `isTvOsSimulator()` → `getSimulatorRuntimeKind()`, and none of them expect it to throw:

- the simulator-server factory that every tap / swipe / keyboard / rotate goes through (`blueprints/simulator-server.ts:343`)
- `describe` (`tools/describe/index.ts:106`)
- `screenshot` (`tools/screenshot/index.ts:155`)
- `await-ui-element` (`tools/await-ui-element/index.ts:386`)
- `await-screen-idle` (`tools/await-screen-idle/index.ts:139`)
- `keyboard` (`tools/keyboard/platforms/ios.ts:16`)
- `tv-control` (`blueprints/tv-control.ts:230`)

So when the lock is busy — which is exactly the "many tool-servers at once" situation this PR is built to handle — the first of these tools to run on a fresh process fails with an internal error instead of quietly treating the device as a normal phone the way it used to. It only happens on the first, uncached lookup for a given device, but that's enough to break a tap or a screenshot.

I reproduced this end to end: with the lock held by another process, `isTvOsSimulator()` throws after ~12 seconds rather than returning false, and that error travels all the way up to the tool call.

packages/tool-server/src/utils/simctl-config.ts
```diff
@@ -5,3 +5,19 @@
...
+
+/** Wait briefly for another argent process' simctl-list probe to finish instead
+ * of spawning another one. This stays below list-devices' branch deadline when
+ * combined with SIMCTL_LIST_DEVICES_TIMEOUT_MS. */
+export const SIMCTL_LIST_DEVICES_LOCK_WAIT_MS = 12_000;
+
+/** Recover from a crashed process that acquired the discovery lock but never
+ * removed it. This is above the subprocess timeout so a healthy holder is not
+ * treated as stale under normal load. */
+export const SIMCTL_LIST_DEVICES_LOCK_STALE_MS = 15_000;
```
hubgan, line 21: These two timeouts work against each other. A process waiting for the lock gives up after `SIMCTL_LIST_DEVICES_LOCK_WAIT_MS` (12s), but a lock is only treated as abandoned once it's older than `SIMCTL_LIST_DEVICES_LOCK_STALE_MS` (15s).

That ordering means a crashed lock can't always be recovered. If the process holding the lock dies without releasing it, a waiter that started within the first ~3 seconds of that lock's life reaches its own 12-second deadline and fails *before* the 15-second staleness point is ever reached — so it gives up instead of cleaning up the dead lock and continuing.

I confirmed this: I planted a lock file whose recorded owner process no longer exists, and a fresh caller still waited the full 12 seconds and failed, even though the "owner" was already dead. The reason is that the code checks the 15-second age gate first and only checks whether the owner is still alive after that gate passes — so a provably-dead holder isn't reclaimed until 15 seconds have elapsed. For a single waiter to reliably clean up after a crashed holder, its wait window needs to be able to outlast the staleness window.

This matters because it undoes part of the crash-recovery that the lock is supposed to provide, and it combines with the previous comment: a crashed or slow holder turns into the same `SimctlListDevicesLockError` that the interaction tools don't handle.

iroiro147:
@hubgan I owe you an apology first: you left two careful reviews and the second
one has sat unanswered since July. That is on me. I have now gone through both
findings properly, and I am closing this PR — with the reasoning, because you
should get something back for the time you put in.

**Both of your findings are correct.**

*On the throw breaking other callers.* You are right, and the list is slightly
longer than the seven you named. The two extra probe sites are
`tools/describe/platforms/ios/index.ts:82` — reached only from `preview.ts:651`,
the one `describeIos` caller that does not pre-resolve `isTvOs` — and
`utils/ios-host.ts:138` in `setupNativeDevtoolsEnvLocal`. Those two degrade
rather than crash: the preview route has its own `try/catch` and returns a 500,
and every path into the native-devtools env setup already wraps it. So it is your
seven that break hard, plus two more that fail less loudly.

The root cause is worth naming, because it is the interesting part: the throw was
added to fix *your earlier comment* about `listIosSimulators()` silently
returning an empty list on a read-only tmpdir. Fixing the visibility problem at
the primitive is what broke nine callers, and patching the nine call sites one by
one is how you end up missing two of them. The correct answer to both of your
comments at once is that `listIosSimulators()` should never throw — log the lock
failure and return `[]` — which was its contract before this PR and, as it turns
out, is still its contract on `main` today.

*On the two timeouts.* Also correct, and the window is exactly three seconds. The
reclaim path checks age before liveness:

```ts
const createdAt = await getSimctlListDevicesLockCreatedAt(lockPath);
if (createdAt === undefined) return;
if (now - createdAt < SIMCTL_LIST_DEVICES_LOCK_STALE_MS) return;   // 15s gate
const metadata = await readSimctlListDevicesLockMetadata(lockPath);
if (metadata && isLockOwnedByLiveProcess(metadata, now)) return;   // liveness, too late
```

A waiter's last reclaim attempt is at `start + 12_000`, and it only gets past the
gate if `start + 12_000 - createdAt >= 15_000`. So any waiter that begins less
than `STALE_MS - WAIT_MS` = 3s after the holder dies burns its full twelve
seconds and throws, while one cheap `isProcessAlive` call would have told it the
owner was already gone. Checking metadata liveness first and falling back to the
age gate only when there is no metadata fixes it without touching either
constant. `removeStaleCleanupLock` has the same ordering, so it is one root cause
in two places.

You should also know neither is covered by the tests I added — every stale-lock
test calls the internal helper directly with a precomputed `now`, which
structurally cannot exercise the wait loop's real timing. That is why both of
these shipped past me.

**Why I am closing rather than fixing.** `main` has moved underneath this PR.
`readSimctlDevices()` no longer exists; `ios-devices.ts` is now 155 lines built
around a per-device-set fan-out — `Promise.all(sets.map(listDeviceSetSimulators))`
— with physical-device and external-provider support, and each set's call already
carries `killSignal: SIMCTL_KILL_SIGNAL` for the wedged-`CoreSimulatorService`
case my version relied on `SIGTERM` for. There is no longer a single call for one
global lock to wrap, so this is not a rebase, it is a rewrite against a different
shape. And `listDeviceSetSimulators` returns `[]` on any failure, which is the
fail-open contract your review was defending.

If capping concurrent `simctl list` is still worth having — the fan-out arguably
made it more pressing, not less, since one call became N — I would be glad to
propose it fresh against the current shape: per-device-set, liveness-first
reclaim with no separate wait/stale duality, and on give-up run `simctl` anyway
rather than fail the caller. Concurrent `simctl list` calls are expensive, not
unsafe, so failing open is the right trade. Happy to open an issue for that first
rather than arrive with another large PR. Thanks again for the reviews — both
findings were right and the second one was the more important of the two.

PR #486, opened 2026-07-09
fix(tool-server): press named key after text in keyboard tool

j-piasecki:
## Problem

A `keyboard` call combining `text` and `key` pressed the named key **before** typing the text — the reverse of the intended "type, then submit". Every platform branch had the same ordering (copied verbatim from the original iOS implementation into each port), and nothing in the schema or description documented it. Ironically, run-sequence's own description lists *"typing then pressing enter"* as a canonical use case.

The observed failure: an agent ran `{ text: "jane@test.dev", key: "enter" }` against a freshly-focused email field in a React Native dev build. Enter fired first into the still-empty field, submitting and blurring it; the text was then typed with nothing focused. RN's `RCTKeyCommands` matches Cmd-registered dev commands even with no modifier held when there is no first responder (`_flags == flags || flags == 0` in `matchesInput:flags:`), so the bare `d` in ".dev" toggled the dev menu mid-sequence, and every subsequent tap landed on the dev-menu overlay.

## Fix

- All executing branches now type `text` first and press `key` after it:
  - `simulator-server-keys.ts` (iOS simulator / Android emulator)
  - `keyboard/platforms/chromium.ts` (Electron / CDP)
  - `keyboard/platforms/vega.ts` (Fire TV)
  - (the TV branch already rejects `key`; unchanged)
- Unknown key names are still rejected **before** any text is typed — each branch resolves the named key up front instead of relying on the old key-first order for fail-fast. The Vega branch gained `resolveVegaNamedKeycode` in `vega-input.ts` so validation is separable from injection.
- The ordering is now documented in the keyboard tool's `key` param description and tool description, and in run-sequence's args table. The run-sequence "type text and submit" example now shows the combined single-step form, since that is the pattern agents naturally reach for.

## Verification

- New regression suite `test/keyboard-key-order.test.ts`: 6 tests asserting key-after-text ordering and fail-fast on unknown keys across all three branches.
- Full tool-server suite: 1995/1996 pass — the single failure (`await-ui-element` Android `hidden` edge case) fails identically on clean `main`, as do the two `tsc` errors from a stale `@argent/telemetry` build.


filip131311 (approved):
this is clearly a good change thank you so much

PR #487, opened 2026-07-10
ci: ship TCP-transport iOS binaries in the published package

jwajgelt:
## Why

argent's `ios-remote` (sim-remote) support drives **ax-service** and **native-devtools** over a TCP socket, resolving them from `bin/darwin/tcp/ax-service` and `dylibs/tcp/*.dylib`. `bundle-tools.cjs` already stages those into the published package **when present**, but `download-native-binaries.sh` never fetched them — so `@swmansion/argent` 0.15.0 shipped with no TCP variant at all, and every `ios-remote` `describe` / `native-*` call failed with "not found".

## What

- **`download-native-binaries.sh`** — fetch the TCP dylibs tarball into `dylibs/tcp/` and `tcp-ax-service` into `bin/darwin/tcp/ax-service` (best-effort: skipped with a warning on pre-sim-remote release tags), with the same Mach-O platform + codesign verification applied to the iOS/tvOS assets.
- **`publish.yml` / `publish-next.yml`** — hard `test -x`/`test -f` gates for the four TCP files so a missing build fails the publish loudly instead of silently regressing (mirrors the existing tvOS gates).
- **`package.json`** — add `build:ios-binaries:tcp` and wire it into `pack:mcp:local` so local packs match CI.

`bundle-tools.cjs` needs no change — its `ax-service (tcp)` asset entry and recursive `dylibs/` copy already handle the `tcp/` slot.

## Verified locally

- `npm pack --dry-run -w @swmansion/argent` now lists `bin/darwin/tcp/ax-service` + `dylibs/tcp/{libArgentInjectionBootstrap,libKeyboardPatch,libNativeDevtoolsIos}.dylib`.
- `package-manifest.test.ts` (download↔bundle-tools path agreement) and `native-devtools-ios/index.test.ts` (TCP path resolution) pass.
- End-to-end against a real sim-remote device: with the TCP binaries in place, `describe` returns the full ax tree and `native-describe-screen` / `native-find-views` / `native-full-hierarchy` all return `status: ok`.

## Paired PR

The producer side — the workflow that builds, signs, and uploads these TCP release assets — is **software-mansion/argent-private#26**. Both must land for a release to ship the TCP binaries.

## Known follow-up (out of scope here)

`axServiceBinaryPathTcp()` resolves `bin/<hostPlatformKey()>/tcp/ax-service`, and `bundle-tools.cjs` stages the TCP ax-service **darwin-only**. So `describe` via ios-remote works from a macOS argent host (verified), but a **Linux** host would hit "not found" at `bin/linux/tcp/`. The TCP dylibs are fine cross-host (flat `dylibs/tcp/`, un-gated). Full Linux-host support means staging the darwin TCP ax-service into each host's `tcp/` dir, modeled on the per-platform `simulator-server` loop — left as a follow-up.


filip131311 (approved):
👍

PR #497, opened 2026-07-14
feat(flow): multi-tap via tap times and gesture-tap clickCount

j-piasecki:
Flows can now express double-taps declaratively:

```yaml
- tap: { on: Login, times: 2 }          # selector double-tap
- tap: { x: 0.5, y: 0.57, times: 2 }    # coordinate double-tap
```

The selector nests under `on` so option keys never share a map with selector fields — zod would silently strip them; mixing errors with a hint instead. `times: 1` normalizes to absent and the serializer emits the minimal spelling, keeping parse/serialize exact inverses.

The taps dispatch as **one gesture** via `gesture-tap`'s new `clickCount` (1–10): N Down/Up pairs ~100 ms apart on touch, escalating CDP `clickCount` per click on Chromium (what makes `dblclick` actually fire) — never N separate calls whose RPC gaps could miss the OS double-tap window.

The recorder carries a recorded `clickCount` into `times`, so a recorded double-tap can't silently replay as a single tap.

### DIFF
--- a/packages/skills/skills/argent-create-flow/SKILL.md
+++ b/packages/skills/skills/argent-create-flow/SKILL.md
@@ -23,7 +23,7 @@ Beyond raw `tool:` steps and `echo:`, flows support declarative directives inter
 | Directive    | YAML                                                                                                     | Meaning                                                                 |
 | ------------ | -------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------- |
 | `launch`     | `- launch: com.acme.app` or `- launch: { ios: …, android: … }`                                           | start the app from scratch (terminate + relaunch) and wait until ready  |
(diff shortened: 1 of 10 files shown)


latekvo (approved):
Clean. A little personal preference that's debatable: I'd prefer a regular agent to always default to coordinate based navigation:

- Agent already knows exact coord thanks to `describe`
- When our `describe` coord readout is bugged and the agent is using the coord input method, the agent can reliably recover
- When the readout is bugged and the agent uses label instead, there's predictable way for the agent to recover from such scenario
- This approach also fixes a possibility of agent trying to use a selector that would have ambiguous target, like multiple buttons with the same label or something.

But this isn't a requirement and it's up for discussion.

j-piasecki:
Raw `gesture-tap` tool always accepts the coordinates and not identifiers/text. That resolution is a part of the flow runner. Or did you mean something else?

PR #498, opened 2026-07-14
feat(flow): conditional when blocks

j-piasecki:
## What

A `when:` block runs a step list only when one condition holds — a UI
condition (`exists`/`visible`/`hidden`/`text`, the await/assert shapes)
or a static `platform` test. **No else**, by design: a block exists to
restore determinism (dismiss an interstitial, reconverge), never to
test two paths. Tap-if-present is a one-step block:
`when: { visible: "Got it" }` + `steps: [tap: "Got it"]`.

No per-step `optional:` — `when:` covers every directive once. The key
is rejected at parse with a pointer to `when:`, not silently dropped (a
Maestro habit that would leave a "can't fail" step hard-stopping the
flow).

## How

- The guard reuses the assert engine at the 1s grace (`timeout` is a
parse error), so a skipped block barely costs a clean run.
- Verdicts are evidence-based: no trustworthy read in the window →
indeterminate → the step **errors** ("could not evaluate" is not
"condition false"); a transient read failure after trusted reads showed
the condition false stays a clean skip.
- `platform` folds `ios-remote` into `ios` — the parser rejects that
spelling, so the fold is the only way a guard matches a remote sim.
- A skipped block reports one line per authored step, same shape for
unmet guard, hard stop, or cancellation; everything goes through
`pushReport`, so when lines stream live. Failures inside an entered
block hard-stop as usual.
- Results carry an explicit `aborted` flag; `ok` keys off
fail/error/cancel, not skip lines.
- Nesting is capped at 20, so a cyclic YAML alias is a structured parse
error, not a stack overflow.

## Testing

`vitest` 2546 pass (17 when-specific); `tsc --noEmit` (tool-server,
argent-mcp, argent-cli), eslint, prettier clean.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -45,6 +45,15 @@ export interface DirectiveOutcome {
   reason?: string;
   /** The run was cancelled mid-step — reported as a skip, not a step failure. */
   aborted?: boolean;
+  /**
+   * The condition could not be evaluated — unknown, not false: the window
+   * never produced a trustworthy read (every fetch threw or returned a
+   * blind/degraded tree), or a `hidden` check ended on blind reads after the
+   * element had matched. Read by the `when:` guard probe, which must error
+   * rather than silently skip a block a broken tree source can't vouch for;
+   * a plain `assert` reports it as an ordinary failure.
+   */
+  indeterminate?: boolean;
 }
 
 /**
@@ -177,6 +186,25 @@ function axisFullyInside(
 // Playwright's web-first assertions, assert retries for a short grace window so
 // it absorbs that latency; a genuinely-false assertion still fails quickly.
 const DEFAULT_ASSERT_TIMEOUT_MS = 1000;
+
+/**
+ * Evaluate a `when:` block's UI guard — the same engine as `assert`, on the
+ * same assert grace window: a skipped block must not add an await-sized dead
+ * wait to every clean run. `ok` is "condition met"; `indeterminate`
+ * distinguishes an unreadable tree (the caller errors — unknown is not false)
+ * from a plainly unmet condition (the caller skips).
+ */
+export function probeWhenCondition(
+  env: ActionEnv,
+  cond: {
+    condition: WaitCondition;
+    selector: FlowSelector;
+    expectedText?: string;
+    textMatch?: TextMatchMode;
+  }
+): Promise<DirectiveOutcome> {
+  return waitForCondition(env, cond, DEFAULT_ASSERT_TIMEOUT_MS);
+}
 
 /**
  * The strict selectors a flow selector resolves through, in order. A loose
@@ -769,11 +797,35 @@ async function waitForCondition(
     }
   }
 
-  if (fetchError) return { ok: false, reason: `could not read the UI tree: ${fetchError}` };
-  if (step.condition === "hidden" && !everTrustedRead) {
+  // Unknown, not false: the window never produced a read we can trust —
+  // every fetch either threw or returned a blind tree (empty + degraded
+  // hint, or empty after the selector had matched). This is not specific to
+  // `hidden`: a probe that never got a trustworthy look at the screen cannot
+  // vouch for "condition false" for any condition. Conversely, once a trusted
+  // read exists, a transient fetch error on the trailing polls does NOT make
+  // the outcome unknown — the reads that succeeded already showed the
+  // condition false, and a last-poll blip must not flip a clean skip into a
+  // hard error.
+  if (!everTrustedRead) {
     return {
       ok: false,
-      reason: "could not confirm the element is hidden — the UI tree was empty or unreadable",
+      indeterminate: true,
+      reason: fetchError
+        ? `could not read the UI tree: ${fetchError}`
+        : "could not evaluate the condition — every read of the UI tree was empty or degraded",
+    };
+  }
+  // `hidden` with an evidence gap: the element matched on an earlier trusted
+  // read and every read since was blind or failed, so gone-ness can't be
+  // confirmed — also unknown, not false. (A trusted read WITHOUT a visible
+  // match would have satisfied `hidden` inside the loop; a final trusted
+  // read WITH one falls through to the determinate "still visible" below.)
+  if (step.condition === "hidden" && !lastMatches.some(isVisible)) {
+    return {
+      ok: false,
+      indeterminate: true,
+      reason:
+        "could not confirm the element is hidden — it was visible earlier, but the last UI reads were empty or unreadable",
     };
   }
   return {
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -48,10 +48,10 @@ export interface DirectiveOutcome {
   /**
    * The condition could not be evaluated — unknown, not false: the window
    * never produced a trustworthy read (every fetch threw or returned a
-   * blind/degraded tree), or a `hidden` check ended on blind reads after the
-   * element had matched. Read by the `when:` guard probe, which must error
-   * rather than silently skip a block a broken tree source can't vouch for;
-   * a plain `assert` reports it as an ordinary failure.
+   * blind/degraded tree), or a `hidden` check ended on a blind or failed
+   * read after the element had matched. Read by the `when:` guard probe,
+   * which must error rather than silently skip a block a broken tree source
+   * can't vouch for; a plain `assert` reports it as an ordinary failure.
    */
   indeterminate?: boolean;
 }
@@ -765,6 +765,10 @@ async function waitForCondition(
   let fetchError: string | undefined;
   let everMatched = false;
   let everTrustedRead = false;
+  // Whether the LAST completed read attempt was trusted — assigned on every
+  // pass through the loop (true on a trusted fetch, false on a blind one or a
+  // throw), so post-loop it describes the final poll.
+  let lastReadTrusted: boolean;
   let finalPoll = false;
 
   for (;;) {
@@ -777,6 +781,7 @@ async function waitForCondition(
       const blind =
         data.tree.children.length === 0 && Boolean(data.hint || data.should_restart || everMatched);
       everTrustedRead ||= !blind;
+      lastReadTrusted = !blind;
       if (
         !blind &&
         evaluateCondition(step.condition, step.expectedText, lastMatches, step.textMatch)
@@ -785,6 +790,9 @@ async function waitForCondition(
       }
     } catch (err) {
       fetchError = err instanceof Error ? err.message : String(err);
+      // A throw is as blind as an empty tree — `lastMatches` still holds the
+      // previous successful read, which must not pass for current evidence.
+      lastReadTrusted = false;
     }
     if (Date.now() >= deadline) {
       if (finalPoll) break;
@@ -805,7 +813,10 @@ async function waitForCondition(
   // read exists, a transient fetch error on the trailing polls does NOT make
   // the outcome unknown — the reads that succeeded already showed the
   // condition false, and a last-poll blip must not flip a clean skip into a
-  // hard error.
+  // hard error. (`hidden` is the one exception, handled below: there
+  // "condition false" means the element was VISIBLE, and the element leaving
+  // is exactly the transition `hidden` waits for — so visibility seen before
+  // the reads went dark cannot vouch for it still being on screen.)
   if (!everTrustedRead) {
     return {
       ok: false,
@@ -816,16 +827,18 @@ async function waitForCondition(
     };
   }
   // `hidden` with an evidence gap: the element matched on an earlier trusted
-  // read and every read since was blind or failed, so gone-ness can't be
+  // read and the FINAL read attempt was blind or threw, so gone-ness can't be
   // confirmed — also unknown, not false. (A trusted read WITHOUT a visible
-  // match would have satisfied `hidden` inside the loop; a final trusted
-  // read WITH one falls through to the determinate "still visible" below.)
-  if (step.condition === "hidden" && !lastMatches.some(isVisible)) {
+  // match would have satisfied `hidden` inside the loop, so a trusted final
+  // read implies it saw the element — that falls through to the determinate
+  // "still visible" below with `lastMatches` fresh from that read.)
+  if (step.condition === "hidden" && !lastReadTrusted) {
     return {
       ok: false,
       indeterminate: true,
-      reason:
-        "could not confirm the element is hidden — it was visible earlier, but the last UI reads were empty or unreadable",
+      reason: fetchError
+        ? `could not confirm the element is hidden — it was visible earlier, but the last UI read failed: ${fetchError}`
+        : "could not confirm the element is hidden — it was visible earlier, but the last UI reads were empty",
     };
   }
   return {
@@ -856,13 +869,11 @@ function assertReason(
         ? `element(s) matched ${sel} but none was visible (zero-area frame)`
         : `no element matched selector ${sel}`;
     case "hidden":
-      // No visible match on the final read means the deadline was spent on
-      // blind reads (blank trees the poll refuses to trust once the element
-      // has matched — see waitForCondition): the element may well be gone,
-      // but that can't be confirmed, so don't claim it was still on screen.
-      return matches.some(isVisible)
-        ? `an element matching ${sel} was still visible`
-        : `could not confirm the element is hidden — it was visible earlier, but the last UI reads returned an empty tree`;
+      // Reached only when the final read was trusted (waitForCondition
+      // returns indeterminate when it was blind or threw), and a trusted read
+      // without a visible match satisfies `hidden` inside the poll loop — so
+      // `matches` holds what that read saw: the element, still on screen.
+      return `an element matching ${sel} was still visible`;
     case "text": {
       const first = firstInReadingOrder(matches.filter(isVisible)) ?? firstInReadingOrder(matches);
       if (!first) return `no element matched selector ${sel}`;
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -187,6 +187,17 @@ function axisFullyInside(
 // it absorbs that latency; a genuinely-false assertion still fails quickly.
 const DEFAULT_ASSERT_TIMEOUT_MS = 1000;
 
+// Evidence-gap bound for `waitForCondition`'s post-timeout verdict: how far
+// behind the loop's exit the last TRUSTED read may lie before a determinate
+// "condition false" verdict stops being honest. Two poll intervals budgets
+// the worst genuine last-poll blip — up to one interval of sleep since the
+// last clean read, plus an interval's worth of latency for the deadline poll
+// and its back-to-back final retry both failing.
+// Anything longer means consecutive polls went dark, and a verdict narrated
+// from the reads before the darkness would describe a screen nobody saw at
+// the deadline.
+const CONDITION_DARK_TAIL_TOLERANCE_MS = POLL_INTERVAL_MS * 2;
+
 /**
  * Evaluate a `when:` block's UI guard — the same engine as `assert`, on the
  * same assert grace window: a skipped block must not add an await-sized dead
@@ -764,7 +775,10 @@ async function waitForCondition(
   let lastMatches: ReturnType<typeof findAll> = [];
   let fetchError: string | undefined;
   let everMatched = false;
-  let everTrustedRead = false;
+  // Date.now() of the most recent TRUSTED read — undefined until one lands.
+  // Post-loop it anchors the dark-tail measurement: how long the window's
+  // final stretch went without a trustworthy look at the screen.
+  let lastTrustedReadAt: number | undefined;
   // Whether the LAST completed read attempt was trusted — assigned on every
   // pass through the loop (true on a trusted fetch, false on a blind one or a
   // throw), so post-loop it describes the final poll.
@@ -780,7 +794,7 @@ async function waitForCondition(
       everMatched ||= lastMatches.length > 0;
       const blind =
         data.tree.children.length === 0 && Boolean(data.hint || data.should_restart || everMatched);
-      everTrustedRead ||= !blind;
+      if (!blind) lastTrustedReadAt = Date.now();
       lastReadTrusted = !blind;
       if (
         !blind &&
@@ -805,19 +819,32 @@ async function waitForCondition(
     }
   }
 
-  // Unknown, not false: the window never produced a read we can trust —
-  // every fetch either threw or returned a blind tree (empty + degraded
-  // hint, or empty after the selector had matched). This is not specific to
-  // `hidden`: a probe that never got a trustworthy look at the screen cannot
-  // vouch for "condition false" for any condition. Conversely, once a trusted
-  // read exists, a transient fetch error on the trailing polls does NOT make
-  // the outcome unknown — the reads that succeeded already showed the
-  // condition false, and a last-poll blip must not flip a clean skip into a
-  // hard error. (`hidden` is the one exception, handled below: there
-  // "condition false" means the element was VISIBLE, and the element leaving
-  // is exactly the transition `hidden` waits for — so visibility seen before
-  // the reads went dark cannot vouch for it still being on screen.)
-  if (!everTrustedRead) {
+  // Post-timeout verdict — unknown must not masquerade as false. Three tiers
+  // of evidence quality:
+  //
+  // 1. No trusted read in the whole window: every fetch either threw or
+  //    returned a blind tree (empty + degraded hint, or empty after the
+  //    selector had matched). A probe that never got a trustworthy look at
+  //    the screen cannot vouch for "condition false" for ANY condition.
+  // 2. Trusted reads existed but the window went dark at the end: the FINAL
+  //    read attempt was blind or threw AND the last trusted read lies more
+  //    than {@link CONDITION_DARK_TAIL_TOLERANCE_MS} behind the loop's exit.
+  //    The condition becoming true is exactly the transition being waited on,
+  //    so a "condition false" observation from before the reads went dark
+  //    says nothing about the deadline — a determinate verdict built from it
+  //    would let a dying tree source fake a clean report (and green-skip a
+  //    `when:` guard whose dismissal target may well be on screen). `hidden`
+  //    is held to a stricter bar: there "condition false" means the element
+  //    was VISIBLE, and the element leaving is the transition itself — so ANY
+  //    untrusted final read, however short the tail, leaves gone-ness
+  //    unconfirmable.
+  // 3. Dark tail within the tolerance — a genuine last-poll blip: trusted
+  //    reads showed the condition false until at most ~one poll interval
+  //    before the deadline, so they still describe the window and a transient
+  //    fetch error on the trailing poll must not flip a clean skip into a
+  //    hard error. The determinate verdict stands, with the failed final read
+  //    appended so the error is not silently dropped from the report.
+  if (lastTrustedReadAt === undefined) {
     return {
       ok: false,
       indeterminate: true,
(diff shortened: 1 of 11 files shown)


packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -696,20 +739,146 @@ async function execSteps(
...
+      pushReport(state, {
+        index,
+        kind: "when",
+        status: "error",
+        reason: `could not evaluate when guard (${label}): ${probe.reason}`,
+        flow: sourceFlow,
+        target,
+      });
+      state.stopped = true;
+      return;
```
latekvo, line 891: When the guard probe comes back indeterminate, this branch emits the `when:error` marker and hard-stops, but returns without expanding the block's own guarded steps. Every other path where the block does not run expands them one line per authored step: the unmet-guard branch just below and the aborted branch above both call `reportBlockSkipped`, and so does the outer-loop hard-stop in `execSteps`. Because of that, the same authored block produces a different number of report lines depending on why it didn't run: an unmet two-step block reports `when:skip, <step>:skip, <step>:skip`, whereas an errored guard reports only `when:error` with the two guarded steps missing entirely. That contradicts the invariant `reportBlockSkipped` documents ('a run where the block was skipped (unmet guard, hard stop, or cancellation) produces the same report shape (one line per authored step) ... reports stay comparable run-to-run'), and the same omission recurs when the errored guard is nested, dropping its inner subtree. The run verdict is unaffected (still ok:false); the divergence is in the report shape and the skipped/errored tallies.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -756,11 +784,35 @@ async function waitForCondition(
...
+        ? `could not read the UI tree: ${fetchError}`
+        : "could not evaluate the condition — every read of the UI tree was empty or degraded",
+    };
+  }
+  // `hidden` with an evidence gap: the element matched on an earlier trusted
+  // read and every read since was blind or failed, so gone-ness can't be
+  // confirmed — also unknown, not false. (A trusted read WITHOUT a visible
+  // match would have satisfied `hidden` inside the loop; a final trusted
+  // read WITH one falls through to the determinate "still visible" below.)
+  if (step.condition === "hidden" && !lastMatches.some(isVisible)) {
```
latekvo: This `hidden` indeterminate branch is gated on `!lastMatches.some(isVisible)`, but the fetch `catch` earlier in the loop only records `fetchError` and leaves `lastMatches` holding the matches from the last successful read. So when the element matched on an early trusted read and then every remaining read in the grace window throws (for example a native-devtools disconnect, which `fetchFlowTree` surfaces as a throw), `lastMatches` still contains the visible match, this branch never fires, and evaluation falls through to the determinate-false return: the guard reports `when:skip` with reason 'condition not met' and the run stays green. The morally identical case where the post-match reads instead return an empty tree does reach this branch and errors. Concretely, a `when: { hidden: Spinner }` block with Spinner visible on read 1 and every later read throwing yields `when:skip` / ok:true, while the same block whose later reads return an empty degraded tree yields `when:error` / ok:false. That is the reverse of the stated intent that an unreadable tree must error rather than silently turn a guarded dismissal into a green no-op, and the 'condition not met' reason reports an unevaluable read as a confirmed-false condition.

latekvo (commented):
A couple of observations on the reporting/evaluation paths, left inline.

packages/tool-server/src/tools/flows/flow-run.ts
j-piasecki: Confirmed — the indeterminate-guard branch returned before expanding the block, so an errored guard reported a different shape (and skipped tally) than the unmet/aborted/hard-stop paths, and dropped nested subtrees. Fixed in `execWhenStep` by calling `reportBlockSkipped` on the guarded steps with reason "when guard errored" before returning, matching the sibling paths; the `reportBlockSkipped` doc now lists the errored guard among the covered cases. Updated the three existing errored-guard tests to assert the expanded shape and added a nested-when test covering subtree expansion and the skipped/errored tallies.

Fixed in 7e48565b6d4e6b8a4a6340645ee3c7b728f7bccd.

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: Confirmed — the hidden indeterminate branch inspected `lastMatches`, which a throwing fetch never updates, so a visible match from an early trusted read survived a throw-tail and pushed evaluation to determinate false (`when:skip`, run green), while the equivalent blind-empty tail errored. Fixed by tracking whether the last completed read attempt was trusted (false on a blind fetch or a throw) and gating the branch on that, so a hidden check whose final read attempt was blind or threw is indeterminate either way: the when guard errors, assert/await fail with "could not confirm", and the reason now carries the fetch error for the throw case. The trailing-blip tolerance for non-hidden conditions is deliberately unchanged — there a trusted read showing the condition false stays valid evidence through a last-poll fetch error; for hidden it cannot, since the element leaving is exactly the transition being waited for. With the new gate, the determinate "still visible" path is only reachable off a trusted final read (a trusted read with no visible match satisfies hidden inside the loop), which also made `assertReason`'s "could not confirm" arm dead code — removed. Tests cover the throw-tail guard error, the assert-level reason, the non-hidden trailing-throw clean skip, and a mid-window throw recovering to a trusted "still visible" read.

Fixed in 8fa6ba3d.

filip131311 (commented):
Reviewed this by packing the branch (`npm run pack`) and driving the resulting tarball against a live iOS simulator (iPhone 17, iOS 26.4) and Android emulator (Pixel 9) with a real RN app. The feature itself holds up well — guard met/unmet, `platform` guards flipping correctly across ios/android, nesting, `hidden`/`text` guards, failures inside an entered block hard-stopping, exit codes, and all the parse rejections (`else`, `optional`, `timeout`, empty `steps`, two condition keys, `ios-remote`, and the cyclic alias hitting the depth cap and rendering `[cyclic entry]`). The no-else / no-`optional` design reads well and the rejection messages point where they should.

One real bug, in the reporting rather than the execution.

### A skipped `echo` inside a `when:` block prints as if it ran

Execution is correct — the block is genuinely skipped. Only the report lies. From a live run of a flow whose block is guarded on an absent element:

```
·  3 when visible "Nonexistent Interstitial XYZ" — condition not met (visible text="Nonexistent Interstitial XYZ") — block skipped (2 steps)
·  4 tap "Nonexistent Interstitial XYZ" — when block skipped
›  THIS MUST NOT RUN
✓  5 assert visible "Explore"
```

Both steps in the block were skipped. The `tap` says so; the `echo` prints its message exactly as if it had executed.

**Why:** `reportBlockSkipped` (`flow-run.ts:795`) deliberately attaches `message` for echo steps, but echo is rendered as narration — no index, no status glyph, just `› message` — so it is the one report line carrying no status at all. The renderers print the message whenever one is present and never check `status`: `argent-cli/src/flow.ts:329` (batch) and `:420` (live), plus `argent-mcp/src/content.ts:257`. A skipped echo and an executed echo are therefore byte-identical. Those renderers are untouched by this PR; it is the new `reportBlockSkipped` that starts feeding skipped messages into them.

**Why it's more than cosmetic:** the report is the artifact a human or an agent reads to decide what happened. An author who writes the natural thing gets two adjacent lines that contradict each other, with the false one in their own voice:

```
·  2 when visible "General" — condition not met (visible text="General") — block skipped (2 steps)
›  guard met - entering block
```

**The inconsistency:** the two skip paths disagree, so an echo is either over- or under-reported depending on which one skipped it:

| echo state | what prints |
| --- | --- |
| actually ran | `› msg` |
| skipped inside an unmet `when:` block | `› msg` — indistinguishable |
| skipped after a hard stop | nothing at all |

The third case differs because the top-level hard-stop path builds its skip report without `message`, so the echo vanishes. That sits awkwardly against the PR description's "one line per authored step, same shape for unmet guard, hard stop, or cancellation".

**Suggested fix:** have the renderers emit a skip line instead of the narration when `status === "skip"`, rather than dropping the message in `reportBlockSkipped`. That keeps one line per authored step (what the PR is going for) and also fixes the hard-stop case where the echo currently disappears entirely.

### Not blocking, but worth knowing: the `indeterminate` net has a hole

`blind = children.length === 0 && Boolean(hint || should_restart || everMatched)` means **an empty tree with no `hint` counts as a trusted read**. This is identical on `main`, so it is not a regression from this PR — but it matters more here, because an unmet guard is now a silent, green skip, which is precisely the "broken tree source turns every guarded dismissal into a green no-op" outcome the indeterminate design sets out to prevent.

Live, on a clean simulator, a flow that launches an Apple system app (never injectable — library validation) passes end to end with exit 0 while the element is plainly on screen:

```
✓  1 tool launch-app                      (com.apple.Preferences)
·  2 when visible "General" — condition not met — block skipped (2 steps)
✓  3 assert hidden "General"              ← false pass; `describe` shows General on screen
```

`resolveNativeTargetApp` resolves the system app (it reports `connected: true`), `queryViewHierarchy` returns no windows, and the resulting empty-and-unhinted tree is treated as trustworthy evidence. Worth noting `flow-ios-tree.ts`'s docstring claims the source "throws rather than letting the caller degrade" — it doesn't in this case. The new tests cover the empty-tree-**plus-hint** case; the no-hint case is the untested one. Candidate fix for a follow-up: treat an empty full-hierarchy result as blind, or have the iOS source throw when the resolved target isn't injectable / returns no windows.

latekvo:
This will run Doom.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
```diff
@@ -92,6 +92,17 @@ You can still edit the .yaml file directly afterwards to remove or reorder steps
...
         case "wait":
           return `${n}. wait: ${step.ms}ms`;
+        case "when": {
+          // Mirror the await/assert rendering above — selectorLabel spelling,
+          // expected text included for text guards.
+          const cond =
+            step.condition.kind === "platform"
+              ? `platform ${step.condition.platform}`
+              : step.condition.condition === "text"
+                ? `text ${selectorLabel(step.condition.selector)} == "${step.condition.expectedText ?? ""}"`
```
latekvo: This `when` branch renders every text guard as `text <selector> == "<expectedText>"`, ignoring `step.condition.textMatch`. The `await`/`assert` branch just above distinguishes the three modes — `matches /…/`, `== "…"` for equals, and `contains "…"`. As a result a `when` guard authored with `contains` is shown in the recording summary as an exact-equality check, and a `matches` guard has its regex source printed as a literal `== "…"` string, misrepresenting a substring or regex check as exact-string equality. The comment's claim that this mirrors the await/assert rendering does not hold for the text case.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -370,10 +394,23 @@ type YamlScrollBody =
...
+/**
+ * A `when:` guard body: exactly one UI condition key (the await/assert shapes,
+ * no `timeout` — evaluation always uses the assert grace) or `{ platform }`.
+ */
+type YamlWhenBody =
+  | { exists: YamlSelector }
+  | { visible: YamlSelector }
+  | { hidden: YamlSelector }
+  | { text: { in: YamlSelector; contains: string } }
+  | { text: { in: YamlSelector; equals: string } }
```
latekvo: `YamlWhenBody`'s `text` variant lists only `contains` and `equals`, but a `when` guard's condition is parsed by `parseWaitFields(…, "when")`, which also accepts `matches` (and `YamlWaitCondition` above includes the `matches` variant). A `when: { text: { in: …, matches: … } }` guard parses and round-trips at runtime, so this type is narrower than the values the parser actually produces; the gap is currently masked only by the `as YamlWhenBody` cast in `toYamlStep` and would surface as a compile error if that cast were removed.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
latekvo: This text-guard branch always renders `== "<expectedText>"`, ignoring `step.condition.textMatch`. The await/assert branch just above splits on `matches` / `equals` / `contains`, but here a `contains` guard is summarized as an exact-equality check and a `matches` guard as a literal string rather than a regex. Since this summary is the text an author reads before hand-editing the recorded YAML, it misstates what the guard actually checks. The interpolated value is also not passed through `JSON.stringify` the way the await/assert branch does, so an `expectedText` containing a `"` produces malformed output (e.g. `== "He said "hi""`).

packages/tool-server/src/tools/flows/flow-finish-recording.ts
```diff
@@ -92,6 +92,17 @@ You can still edit the .yaml file directly afterwards to remove or reorder steps
...
+                : `${step.condition.condition} ${selectorLabel(step.condition.selector)}`;
+          return `${n}. when: ${cond} (${step.steps.length} steps)`;
```
latekvo: `(${step.steps.length} steps)` is not pluralized, so a single-step block renders `(1 steps)`. The runtime skip reason in flow-run.ts pluralizes the same count (`n === 1 ? "" : "s"`), so the two surfaces disagree for a one-step block.

packages/tool-server/src/tools/flows/flow-utils.ts
latekvo: `YamlWhenBody` lists only the `contains` and `equals` text variants, omitting the `matches` shape that `YamlWaitBody` above carries. A `when` text guard does accept `matches` at runtime (`parseWaitFields` handles it) and `toYamlStep` emits `{ text: { in, matches } }` for such a guard — this only type-checks because it is forced through the `as YamlWhenBody` cast in `toYamlStep`, which hides that the type does not cover a value the code actually produces.

latekvo (commented):
A few notes on the recording-summary and type surfaces for `when:` guards. The runtime path (parse, guard evaluation, reporting, verdict, CLI rendering) all looks correct and is well covered by the new tests.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
j-piasecki: Confirmed — the `when` branch hard-coded `== "…"`, so a `contains` guard displayed as an exact-equality check and a `matches` guard's regex source printed as a literal string; the mirror-the-await/assert-rendering comment did not hold for the text case. Fixed by extracting the await/assert text tail into a shared `textConditionLabel` used by both branches, so the comparator spelling (`contains "…"`, `== "…"`, `matches /…/`) and the JSON quoting are one definition and cannot drift again. The summary test gains a when-guard case asserting all three comparators.

Fixed in c8d28d1b.

j-piasecki: Same finding as the earlier thread on this line — fixed in c8d28d1b by sharing the await/assert text tail (`textConditionLabel`) between both branches. The quoting half is covered by the same change: the shared helper JSON-quotes literal expectations, so an embedded `"` renders escaped instead of malformed; the new test's expected text carries an embedded quote and a newline.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Confirmed — the hand-listed text variants omitted `matches`, which `parseWaitFields` accepts for when guards and `toYamlStep` emits, masked only by the `as YamlWhenBody` cast. Fixed by deriving the UI arm from `YamlWaitCondition` (`type YamlWhenBody = YamlWaitCondition | { platform: WhenPlatform }`) and dropping the cast — a condition shape added to await/assert is a when-guard shape by construction, and any future divergence is a compile error at the assignment. `timeout` stays excluded by construction: `YamlWaitCondition` never carried it (the await step type adds it as a sibling key). The when round-trip test gains a `matches` guard.

Fixed in 69d6a2e7.

j-piasecki: Same finding as the earlier thread on this line — fixed in 69d6a2e7: `YamlWhenBody`'s UI arm is now derived from `YamlWaitCondition` and the `as YamlWhenBody` cast in `toYamlStep` is dropped.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
j-piasecki: Confirmed — the summary always spelled `steps`, while flow-run's skip reason pluralizes the same count, so the two surfaces disagreed for a one-step block. Fixed by matching the runtime spelling (`1 step` / `2 steps`); the summary test now asserts both forms.

Fixed in 1ad115de.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -769,11 +805,40 @@ async function waitForCondition(
...
+        : "could not evaluate the condition — every read of the UI tree was empty or degraded",
+    };
+  }
+  // `hidden` with an evidence gap: the element matched on an earlier trusted
+  // read and the FINAL read attempt was blind or threw, so gone-ness can't be
+  // confirmed — also unknown, not false. (A trusted read WITHOUT a visible
+  // match would have satisfied `hidden` inside the loop, so a trusted final
+  // read implies it saw the element — that falls through to the determinate
+  // "still visible" below with `lastMatches` fresh from that read.)
+  if (step.condition === "hidden" && !lastReadTrusted) {
```
latekvo: This evidence-gap treatment is limited to `hidden`, so for `exists`/`visible`/`text` a window whose reads die after one early trusted read falls through to the determinate `assertReason` verdict, built from `lastMatches` captured on that early read. On main, the post-loop `if (fetchError) ...` surfaced `could not read the UI tree: <err>` for every condition whenever the final attempt threw; at this head, `await: { exists: Done, timeout: 120000 }` whose tree source dies after the first poll reports `no element matched selector text="Done"` - a determinate-sounding verdict about a screen that was unreadable for essentially the whole window, with the fetch error absent from the report. For a `text` condition the reason even quotes the element's text from that first read as if it were the state at the deadline (`element matched ... but its text was "Loading"`). The comment above justifies the `hidden` carve-out with the element leaving being exactly the transition `hidden` waits for - but for an `await`, the condition becoming true is equally the transition being awaited, and the dark tail hides it the same way; a last-poll blip does not describe a tail where 11 of 12 polls threw. The same read shape inside a `when:` guard probe reports `condition not met` and skips the block green, so a tree source that dies mid-probe turns a guarded dismissal into exactly the silent green no-op the indeterminate design exists to prevent. Reproduced with a mocked `fetchFlowTree` (read 1 trusted without the element, reads 2+ throwing): `await exists` fails with `no element matched selector text="Done"` (12 reads, 11 thrown, error dropped); the same sequence as a `when` guard yields `when:skip` with `ok: true`.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -726,21 +772,149 @@ async function execSteps(
...
+  let met: boolean;
+  if (step.condition.kind === "platform") {
+    // "ios-remote" is an iOS simulator driven through sim-remote — for a
+    // platform guard it IS ios. The parser deliberately rejects "ios-remote"
+    // as a guard spelling, so without this fold no guard could ever match on
+    // a remote sim and iOS-only blocks would silently skip there.
+    const platform = state.device.platform === "ios-remote" ? "ios" : state.device.platform;
+    met = platform === step.condition.platform;
+  } else {
+    const probe = await probeWhenCondition(state, step.condition);
```
latekvo, line 867: A `text` guard whose expected string uses a `{{secret:NAME}}` placeholder is compared literally by this probe - placeholders resolve only inside the text-entry tools (a flow `type:` step routes through `keyboard`, which resolves them), never in condition evaluation. Such a guard can never match what's on screen, so the block silently skips and the run stays green: with `ARGENT_SECRET_USERNAME` exported and the screen showing exactly that value, `when: { text: { in: account, equals: "{{secret:USERNAME}}" } }` reports `condition not met (...) — block skipped` and the guarded steps never run (reproduced at this head). The same mistake in an `assert` fails loudly and gets discovered on the first run; inside a `when:` guard it becomes a permanent green no-op - the step the author believes is live can never fire, which is the exact failure mode the parse-time `optional:` rejection in this PR argues against.

packages/tool-server/test/flows/flow-when.test.ts
```diff
@@ -0,0 +1,654 @@
...
+      "echo:skip",
+    ]);
+    // The trailing top-level echo is skipped by the hard stop but keeps its
+    // message, so the renderer can mark it skipped instead of dropping the
+    // line (or, worse, printing it as if it had run).
+    expect(result.steps[2].message).toBe("never reached");
+    expect(result.ok).toBe(false);
+  });
+
+  it("evaluates a platform guard statically against the resolved device", async () => {
```
latekvo, line 425: Every execution test in this file runs against an iOS-UDID-shaped device, so platform-guard coverage is ios-matches / android-skips on an iOS device only. The `ios-remote` to `ios` fold in `execWhenStep` - whose own comment notes that without it iOS-only blocks would silently skip on remote simulators - has no test, and neither does a positive platform match on an android/chromium/vega device. The fold is reachable in this same harness (passing `device: "remote:<udid>"` resolves the device as `ios-remote`; verified it enters the block today). If the fold regresses, every `when: { platform: ios }` block silently skips on remote simulators while the whole suite stays green - a silent green no-op of the class this PR's indeterminate-guard handling is specifically designed to prevent.

latekvo (commented):
Ran the branch end-to-end against a live iOS simulator (flow-execute over HTTP plus the CLI renderer) - the guard/skip/error paths all behave as described. A few observations left inline. Written against c2111fe0 just as the three summary/type fixes were landing; re-checked against 1ad115de - those commits touch the recording summary and YamlWhenBody only, so all five comments still apply.

packages/tool-server/test/flows/flow-when.test.ts
```diff
@@ -0,0 +1,654 @@
...
+    expect(() => parseFlow("steps:\n  - tap: A\n    optional: true\n")).toThrow(
+      /optional is not supported — guard the step with a when: block/i
+    );
+    expect(() => parseFlow("steps:\n  - await: { visible: X }\n    optional: false\n")).toThrow(
+      /optional is not supported/i
+    );
+  });
+});
+
+describe("when: execution", () => {
```
latekvo, line 256: None of these execution tests exercises the abort paths added for `when` blocks - the guard-probe abort branch in `execWhenStep`, the pre-step-abort expansion, and the post-hard-stop expansion (the hard-stop test that exists places an echo, not a `when`, after the failure) - and no test anywhere asserts the new `FlowRunResult.aborted` field. The probe-abort branch is silently deletable today: `ABORTED_OUTCOME` carries no `indeterminate`, so without that branch a cancelled probe falls through to `met = false` and the report claims the determinate `condition not met (...) — block skipped` for a screen the probe never read, with the entire suite staying green.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -726,21 +772,149 @@ async function execSteps(
...
+    return;
+  }
+
+  // Marker for the block, then the guarded steps inline — same scope, same
+  // fragment attribution, failures hard-stop as anywhere else.
+  pushReport(state, {
+    index,
+    kind: "when",
+    status: "pass",
+    reason: `condition met (${label})`,
```
latekvo, line 916: This is the only `status: "pass"` report in the runner that carries a `reason` - `StepReport.reason` is documented above as machine-readable explanation for when the step did not pass, and that contract held on main. A consumer of `FlowRunResult.steps` that follows the documented rule (reason present means the step did not pass cleanly) now misclassifies the entered-block marker of every `when:` block in an otherwise green run.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: To be clear about how this relates to the resolved thread from the 14th: that thread was about `hidden` trusting stale `lastMatches` through a throw-tail, and the fix reply there called the trailing tolerance for non-hidden conditions deliberate - a trusted read showing the condition false staying valid evidence through a last-poll fetch error. This comment is about the case that rationale doesn't reach: under `await`, a trusted read showing the condition false is the expected starting state, not evidence about the deadline, and the tail here isn't a last-poll blip but the whole remaining window (11 of 12 polls thrown in the repro). Separate from the verdict itself, the reason string is a regression on its own: main surfaced `could not read the UI tree: <err>` in this situation, and this head drops the fetch error from the report entirely.

packages/tool-server/test/flows/flow-when.test.ts
latekvo: For the record, this doesn't contradict the live verification in the earlier review - platform guards were exercised manually across ios/android there, and the parser's rejection of the `ios-remote` spelling was covered as well. The gap is that nothing in the suite pins the runtime fold itself: it was verified by hand once, and after this merges it can regress without any test noticing.

latekvo (approved):
Neat, thanks ❤️ 

<sub>just resolve the inlines first please. Apologies for the two duplicated comments, this won't happen again</sub>

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: Fixed in 06c55072. Extended the indeterminate treatment to every condition, not just `hidden`.

`waitForCondition` now tracks the timestamp of the last *trusted* read; the post-timeout verdict is three-tiered:

- no trusted read ever → indeterminate (unchanged);
- final read untrusted **and** the last trusted read lies more than two poll intervals behind the loop's exit (a dark tail, not a last-poll blip) → indeterminate for *any* condition, with the fetch error back in the reason (`could not evaluate the condition — the UI tree was unreadable for the final <N>ms of the window: <err>`);
- within that tolerance → the determinate verdict stands (preserving the deliberate blip tolerance from the earlier thread), but with the failed final read appended to the reason so the error is no longer silently dropped.

`hidden` keeps its stricter any-untrusted-final-read carve-out. Your repro (1 trusted read without the element, reads 2+ throwing) now errors the `when:` guard instead of green-skipping, and reports the screen was unreadable for an `await` instead of the stale `no element matched` verdict. Covered by new tests in `flow-hidden-blank-reads.test.ts` (dark-tail diagnostics for non-hidden conditions + the blip-tolerance case) and `flow-when.test.ts` (the same shape erroring a guard).

packages/tool-server/src/tools/flows/flow-run.ts
j-piasecki: Fixed in 06c55072 — rejected at parse, matching this PR's `optional:` / `when.timeout` rejections rather than becoming a permanent green no-op.

`parseWhenCondition` now fails a guard whose expected text *or* any selector string contains `{{secret:`. Secrets resolve only in the text-entry tools, so such a guard tests literal placeholder text that is never on screen — permanently false for `exists`/`visible`/`text` (the silent skip you described), and vacuously true for `hidden` (the block would always run). `assert` / `await` / `type:` are untouched and still carry placeholders verbatim (the keyboard tool resolves them). Tests in `flow-when.test.ts` cover the `equals`/`contains` expected-text shapes, every selector channel, the `hidden` inversion, and a sanity check that a `type:` step still parses its placeholder.

packages/tool-server/test/flows/flow-when.test.ts
j-piasecki: Added in 06c55072. `flow-when.test.ts` now runs a `remote:<udid>` device (resolves as `ios-remote`) and pins the fold directly: `when: { platform: ios }` enters and fires its step while `when: { platform: android }` skips in the same run. Also added positive platform matches on an android serial (`emulator-5554`) and a vega device (`amazon-…`) — the vega block uses echo steps, since touch directives are rejected there before execution. Verified each fails if the `ios-remote → ios` fold line is removed (the guard then skips and the tap/echo never fires).

packages/tool-server/test/flows/flow-when.test.ts
j-piasecki: Added in 06c55072. `flow-when.test.ts` now covers all three abort paths plus the `aborted` field:

- **guard-probe cancellation** — abort fired from inside the mocked tree fetch mid-probe: the step reports `run aborted` (never the determinate `condition not met`), the authored steps expand one skip line each, the guarded steps never run, and `result.aborted === true`. Confirmed this test fails if the `probe.aborted` branch is deleted — the cancelled probe then falls through to `met = false` and misreports the determinate skip, exactly as you called out (`ABORTED_OUTCOME` carries no `indeterminate`).
- **pre-step abort** — run already cancelled before a `when` block: marker + authored steps (nested block included) expand as `run aborted`, `aborted === true`.
- **post-hard-stop** — a failing step before a `when` block: the block expands one line per authored step with `aborted` absent, pinning hard-stop and cancellation as distinct.
- **baseline** — a clean green run asserts `"aborted" in result === false`.

packages/tool-server/src/tools/flows/flow-run.ts
j-piasecki: I looked into this and the premise doesn't hold, so I corrected the doc rather than the behavior.

Passing **snapshot** steps already carry a `reason` — `flow-visual.ts` returns `{ status: "pass", reason: "diff X% ≤ Y% (…)" }` (and `baseline written/updated`), surfaced into `StepReport` via the `snapshot` case in `execLeafStep`. That's all on `main`, so the `when:` marker isn't the first `status: "pass"` report with a `reason`, and the "reason ⇒ didn't pass cleanly" contract didn't actually hold before this PR either.

The two consumers of `FlowRunResult.steps` — `argent-cli/src/flow.ts` and `argent-mcp/src/content.ts` — both render `reason` as an unconditional ` — <reason>` suffix regardless of `status`, so neither implements the documented rule and nothing is misclassified in practice; on a green snapshot run today the reason is already shown, informatively.

So the only real (and pre-existing) issue was the stale field doc. Reworded in 06c55072 to describe the field as used: always set when the step did not pass, and also set on some passing reports whose result is self-narrating — the `when:` guard marker and snapshot passes. Happy to drop the marker's `reason` instead if you'd prefer pass reports never carry one, but that would also mean stripping it from snapshots to be consistent.

PR #500, opened 2026-07-14
feat(flow): long-press directive

j-piasecki:
Press-and-hold on an element — context menus, reorder activation — with
the same resolution behavior as `tap` (settled tree, auto-wait, no
auto-scroll) and the same options shape: a bare/strict selector, or
`long-press: { on: <selector>, duration: <ms> }` with the selector
nested under `on` so option keys never mix with selector fields.

The default 800ms clears both platforms' recognizers (iOS 500ms,
Android ~400ms). Touch platforms dispatch one `gesture-custom` train
(`Down`, then `Up` delayed by the duration) so the hold length is
exact; Chromium maps to a mouse press-hold-release (`gesture-drag`,
from == to) — a desktop context menu is a right-click and is
deliberately not aliased. Vega rejects upfront with the other touch
directives. No coordinate form: a point long-press stays the
`gesture-custom` escape hatch.

### DIFF
--- a/packages/skills/skills/argent-create-flow/SKILL.md
+++ b/packages/skills/skills/argent-create-flow/SKILL.md
@@ -20,17 +20,18 @@ Both run via `argent flow run <name>` — a fragment simply runs against whateve
 
 Beyond raw `tool:` steps and `echo:`, flows support declarative directives interpreted by the runner (they are **not** agent-callable tools). **Every directive hard-stops the flow on failure**; later steps are reported `skip`.
 
-| Directive   | YAML                                                                                                     | Meaning                                                                 |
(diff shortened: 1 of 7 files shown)


filip131311 (approved):
you will know better if this is a problem but: 



Recording never produces long-press:. flow-add-step only rewrites gesture-tap/restart-app/flow-execute into directives, so an agent performing a long-press mid-recording records it as a raw coordinate gesture-custom tool step — non-portable, no selector capture. Consistent with the existing rewrite whitelist, but worth a follow-up issue since long-press is precisely the gesture agents will now perform via gesture-custom.

j-piasecki:
I don't think it is - there are more things that cannot be recorded directly (how do you distinguish a swipe from a scroll?). The agent will update the recorded flow after it's done, so if it knows it wanted to perform a long press, it should update it.

PR #503, opened 2026-07-14
feat(flow): regex matching via matches in assertions and selectors

j-piasecki:
## What

- Add `matches` as a case-sensitive JavaScript regex comparator for `await` and `assert` text conditions.
- Support `{ text: { matches: ... } }` in every flow selector slot, combined with `id` and `role`.
- Preserve regex matchers through parsing, YAML serialization, recording summaries, and run reports.

## Correctness

- Prevent empty-matchable patterns from matching unlabeled nodes or empty assertion text.
- Reject unknown raw `await-ui-element` selector keys instead of silently dropping constraints.
- Reject selector combinations YAML cannot round-trip without changing their semantics.
- Document native label aggregation accurately and keep action ranking consistent.

## Maintenance

- Share selector-field and regex-syntax validation across both matcher parse paths.
- Compile selector regexes once per tree walk and full-match regexes once per ranking pass.
- Make comparator serialization exhaustive, extract the duplicated flow test harness, and scope slow-test timeouts narrowly.

## Verification

- Full tool-server suite: 2,588 tests across 263 files.
- Source and test typechecks, ESLint, Prettier, and skill-copy validation pass.


packages/skills/skills/argent-create-flow/SKILL.md
```diff
@@ -36,7 +36,9 @@ Beyond raw `tool:` steps and `echo:`, flows support declarative directives inter
...
+
+Selectors resolve against the **full native hierarchy** (iOS: the UIView tree; Android: the complete accessibility hierarchy including not-important views) — strictly more than `describe` or the raw `await-ui-element` tool see (both use the trimmed tree), with complete `testID`/`resource-id` coverage. So an `id` selector works even when `describe` collapses or omits the element — don't fall back to coordinate taps just because a testID isn't visible in `describe` output. And when several elements match — including wrappers whose native text aggregates descendant content — the action directives (`tap`, `type`, `scroll-to`) pick the **most specific** match: an exact text/identifier match beats a substring hit (for a regex matcher, a pattern consuming the element's whole text counts as exact), then the smallest frame wins.
```
latekvo: Just to be clear, my comment from #497 DOES NOT apply to this text block, as `flow` files should always prefer by-label taps over coord taps. Just wanted to clear this out :)

latekvo: This text looks bloated to me but this isn't a core skill so it's not a huge deal.

latekvo (approved):
great idea, gg :shipit:

packages/skills/skills/argent-create-flow/SKILL.md
j-piasecki: Reduced it slightly

PR #504, opened 2026-07-15
feat(flow): coordinate targets for tap and long-press

j-piasecki:
## Summary

This adds raw normalized screen coordinates as first-class targets for both `tap` and `long-press`. Gestures can now target either an element selector or a point, while keeping the existing selector auto-wait and off-screen diagnostics.

Supported forms include:

```yaml
- tap: { x: 0.5, y: 0.6 }
- tap: { on: { x: 0.5, y: 0.6 }, times: 2 }
- long-press: { x: 0.5, y: 0.6 }
- long-press: { on: { x: 0.5, y: 0.6 }, duration: 1200 }
```

## What changed

- Introduced a shared gesture-target parser for selector and point targets.
- Added raw-point execution for `long-press` and shared point resolution between tap and long-press.
- Validate coordinates as finite normalized `0–1` screen fractions, rejecting pixels, malformed points, mixed selector/coordinate maps, and unknown point keys.
- Canonicalize option-bearing coordinate gestures under `on`, keeping option keys separate from target fields.
- Preserve coordinate targets in recording summaries and flow progress reports instead of displaying an undefined selector.
- Added serialization validation so recording cannot persist a flow that would fail when parsed for replay.
- Updated the flow authoring skill and tool description with the new coordinate forms.

## Behavior notes

- Raw points act immediately and do not resolve or observe the UI tree.
- Selector targets retain settling, auto-wait, and the existing `scroll-to` hint when an element cannot be found.
- Observation directives such as `type`, `await`, `assert`, and `scroll-to` remain selector-only.
- The mixed spelling `tap: { x, y, times }` is now intentionally rejected; use `tap: { on: { x, y }, times }` instead.

## Validation

- Full tool-server suite: 2,618 tests passed across 266 files.
- Added parser, serialization, recording, touch execution, Chromium execution, reporting, and malformed-input coverage.
- Tool-server build, test typecheck, ESLint, Prettier, and skill validation passed.


filip131311 (approved):
Approving — the feature is well-built and I verified it end-to-end on real devices from a packaged `.tgz`. The parser, the raw-point execution path, and the reporting fix all do exactly what the description says.

One thing must be resolved before this ships, and it is not a code defect — it is a claim in the description that expired while the PR sat.

## ⚠️ WARNING — `tap: { x, y, times }` is NOT unreleased. It shipped in 0.16.0.

The description says:

> The unreleased mixed spelling `tap: { x, y, times }` is intentionally rejected; use `tap: { on: { x, y }, times }` instead.

That was true when this PR was opened, and is not true now:

| | |
|---|---|
| PR opened | `2026-07-15T07:25:01Z` |
| **`0.16.0` published to npm** | **`2026-07-15T19:04:34Z`** (~12h later) |

The spelling was introduced by #497 (`5082b630`), `git tag --contains` → `v0.16.0`, and the **published** tarball carries the old parser (`npm pack @swmansion/argent@0.16.0` → `dist/tool-server.cjs` contains the string `a coordinate tap takes only { x, y, times }`). So this is a breaking change against a released version, not a cleanup of an unreleased spelling.

### It is machine-reachable, not just hand-authored

This is the part that raises the stakes. `flow-add-step.ts` emits the coordinate step for any tap with no stable selector:

```ts
step = { kind: "tap", x: args.x as number, y: args.y as number, ...tapTimes };
```

So **0.16.0's own recorder writes `tap: { x, y, times: 2 }`** whenever a user records a double-tap on an element with no stable selector — precisely the unlabeled-element case coordinate taps exist for, and the case #511 made *more* common by routing invisible-only text selectors to coordinates. Those are committed `.yaml` files in users' repos.

### Verified two-way incompatible (live, both directions)

- **0.16.0-recorded flow → this PR's parser:**
  `Unrecognized flow entry (the tap options form takes a nested point — e.g. tap: { on: { x: 0.5, y: 0.5 }, times: 2 }): {"tap":{"x":0.5,"y":0.6,"times":2}}`
- **This PR's recorded flow → 0.16.0's parser** (ran a 0.16.0 tool-server against a flow this branch recorded):
  `Unrecognized flow entry (tap.on: selector has unknown keys 'x', 'y' — allowed keys: text, id, identifier, role): {"x":0.42,"y":0.62}`

### Why this is a decision and not a blocker

Two things genuinely cap the severity, and they are why I'm approving rather than requesting changes:

1. **It fails loudly, at parse time, with the exact fix in the message.** No silent misfire, no wrong-element tap — the failure mode is a stopped flow with `tap: { on: { x, y }, times: 2 }` printed in the error.
2. **It is deliberate and tested** — `expect(() => parseFlow('… tap: { x: 0.5, y: 0.5, times: 2 }')).toThrow(…)` is in the diff. This was a choice, not an oversight.

**What I'd ask for (your call which):**
- **(a)** Keep the rejection and land it as a **documented breaking change** in the 0.17.0 release notes — cheap, honest, and the error message already does the migration teaching; or
- **(b)** Accept `{ x, y, times }` as a deprecated alias that parses to the canonical form and re-serializes under `on:`, so 0.16.0-recorded flows keep replaying.

Either way, please **correct the "unreleased" line in the description** so the rationale in the merge commit isn't wrong for whoever reads it next.

## Minor — the skill never documents the coordinate options form

`packages/skills/skills/argent-create-flow/SKILL.md` documents these three:

```yaml
- tap: { x: 0.5, y: 0.57 }
- tap: { on: Login, times: 2 }
- long-press: { on: <sel>, duration: 1200 }
```

…but never `tap: { on: { x, y }, times }` or `long-press: { on: { x, y }, duration }` — which after this PR are the **only** spellings for a coordinate double-tap and a timed coordinate long-press. Since the intuitive spelling is now a hard parse error, and this skill is exactly what an agent reads while authoring a flow, an agent following the docs will write the rejected form and hit the error. `flow-run.ts`'s tool description does cover it (`bare { x, y } or on: { x, y }`) — only the skill lags. One row each fixes it.

## Nit (non-blocking)

`FlowStep` lets `selector` and `x`/`y` coexist at the type level, so invalid states are representable and guarded at runtime in three places (`parseTarget`, `targetToYaml`, `resolveTargetPoint`). A discriminated union on the target would delete those guards. Not worth holding the PR for.

---

## Testing

Built the real package from this branch **merged onto current main** (merges clean; no conflict with #511) and installed the `.tgz` into an isolated prefix with its own tool-server.

- **Suite:** 271 files / **2678 tests pass**. ESLint, `tsc --noEmit`, Prettier all clean.
- **iOS** — iPhone 17e, iOS 26.4, driven from the built tarball:
  - At the **identical point** `(0.65, 0.22)` in Maps: plain coordinate `tap` → pin unchanged (`88-231`); coordinate `long-press` → **new pin `Kątno 66A`**. Only the directive differed.
  - `long-press: { on: { x: 0.72, y: 0.22 }, duration: 1500 }` → new pin `87-620` from a no-pin state.
  - 3× `tap: { on: { x: 0.5, y: 0.3 }, times: 2 }` → map zoomed ~3 levels (village-level detail); 3× single coordinate tap at the same point → no zoom. Confirms `times` reaches `clickCount` as a real OS-recognized double-tap at a raw point.
- **Android** — Pixel_9_2:
  - `tap: { x: 0.5, y: 0.307 }` navigates Settings → Internet.
  - `long-press: { x: 0.5, y: 0.35 }` on empty launcher space opens the Wallpaper/Widgets menu; a plain tap at the same point does not. `{ on, duration: 1500 }` likewise.
- **Validation matrix (on-device):** all rejected with specific messages — pixels `{x:250,y:300}`, `{x,y,times}`, mixed selector+coords, `x` without `y`, `.nan`, `.inf`, unknown point key, negative, string coords, and both long-press equivalents. Boundaries `{x:0,y:0}` and `{x:1,y:1}` accepted.
- **Recording round-trip:** recorder writes `tap: { x: 0.42, y: 0.62 }` and `tap: { on: { x: 0.42, y: 0.62 }, times: 2 }`, and replay is green — the serializer and parser agree.

Two environment notes for anyone reproducing (neither is a fault of this PR): iOS system apps are `injectable: false`, so the flow `launch`/`await` gate can't be used with Settings/Maps — coordinate-only fragment flows work there, and Android is unaffected since uiautomator needs no injection. And a second tool-server contends with a running one on the fixed per-device injection socket `/tmp/argent-nd-<UDID>.sock`.

j-piasecki:
Updated the skill to mention `tap: { on: { x, y }, times }` syntax and the PR description to say that this is an intentional breaking change.

Unfortunately, I didn't make it in time before the 0.16 release 😞, but given that the new flow is experimental, it shouldn't be that much of a problem.

PR #429, opened 2026-06-30
feat(flow): support running end-to-end flows without agent

j-piasecki:
# Flow E2E

This PR turns flows from replayable tool-call recordings into a declarative, portable e2e test system that runs headlessly via `argent flow run <name>` — no LLM in the loop, non-zero exit on failure, suitable for CI.

Agents can still replay flows via `flow-execute` and get the full output from each step after it finishes.

## Flow format

Flows are YAML files in `.argent/flows/`, now built from **directives** interpreted by the runner rather than raw tool calls:

- `launch` — terminate + relaunch the app under test (per-platform app id map supported); its presence marks a flow as **e2e** (controls its own start state) vs. a **fragment** (runs against current state, declares an `executionPrerequisite`, composable via `run:` with cycle/depth guards).
- `tap`, `type` (focus handshake + optional submit), `scroll-to` (momentum-free, increments by half the container, `within:` for nested scrollers), `await`/`assert` (`visible`/`exists`/`hidden`/`text` conditions), `wait`, `snapshot`, `echo`, and raw `tool:` as escape hatch — still fully supported, so previously recorded flows keep working unchanged.
- **Selectors** are `{ text, identifier, role }`; a bare string is a loose selector resolving identifier-first, then visible text. Any directive failure hard-stops the flow; later steps report `skip`.

## Execution engine

- Flows store no device id — the runner binds a device at replay (explicit `--device`/`--platform` or the single booted one).
- Selector resolution runs against **platform-unified UI trees**: iOS native view hierarchy (full testID coverage, gated on native-devtools connecting), Android full accessibility hierarchy (full resource-id coverage), Chromium CDP DOM, and the Vega automation toolkit. A shared flatten/text-hoisting pass lets `text` assertions read what an identified container visibly shows. Vega is remote-driven, so touch directives fail with guidance to use `tv-remote`/`keyboard`.
- Steps wait for the tree to **settle** (two consecutive identical reads) before acting; `launch` additionally gates on platform tree-source readiness instead of silently degrading.

## Visual snapshots

`snapshot` diffs a full-resolution screenshot against baselines in `.argent/flows/__baselines__/<flow>/`, keyed by platform + resolution, with configurable `maxMismatch` and `--update-baselines`. The status bar is pinned for the whole run (iOS `simctl status_bar`, Android demo mode) so clocks/battery don't drive diffs. Baseline/current/diff images are returned as artifacts on failure (and the written baseline on first capture/update); a clean pass carries none, keeping reports and agent context lean.

[...]

PR #505, opened 2026-07-15
feat(tool-server): type secrets via {{secret:NAME}} placeholders — plaintext never enters agent context

filip131311:
## Why

Feedback: unlike Stagehand's `act("type %pw% into password", { variables: { pw } })`, argent had no way to type a credential without the LLM seeing the plaintext — which disqualifies it as a CI agent. In the MCP model the agent composes the `keyboard` call itself, so a plaintext password lands in **six** places: the agent transcript, `~/.argent/mcp-calls.log` (spyLog records full args + result), the flag-gated tool-server event log, the `typed` result echo, recorded flow YAMLs (committed to the repo), and the post-type auto-screenshot (plaintext back into context as pixels for non-secure fields).

## What

`keyboard` and `paste` now accept **`{{secret:NAME}}`** placeholders in `text`, resolved server-side from **`ARGENT_SECRET_NAME`** environment variables:

```yaml
# GitHub Actions
env:
  ARGENT_SECRET_APP_PASSWORD: ${{ secrets.APP_PASSWORD }}
```

```json
{ "udid": "<UDID>", "text": "{{secret:APP_PASSWORD}}", "key": "enter" }
```

**Resolution happens inside the tool's `execute`** — after every existing logging boundary — so spyLog, the event log, the transcript, and recorded flows all see only the placeholder, with near-zero extra redaction work:

- `run-sequence` keyboard steps and flow `type` steps go through `invokeSubTool` into the same `execute` → covered for free. A recorded login flow stores the placeholder verbatim, so the committed YAML is secret-free — and agent-less `argent flow run` (#429) resolves it in CI from the job's env, no LLM in the loop at all.
- The mandatory `ARGENT_SECRET_` prefix is an allowlist by construction: a prompt-injected agent can never exfiltrate `AWS_SECRET_ACCESS_KEY` or `GITHUB_TOKEN` through this mechanism — only vars the user deliberately exposed under the prefix resolve.
- `typed` echoes the placeholder, never the value.
- Backend errors are scrubbed at the resolution site (e.g. a failed Android `input text` echoes the device-side command line containing the secret) — message and stack, in place, preserving the error class and with it the HTTP-status/telemetry mapping.
- Unknown placeholder → 400 listing the **names** of available secrets, never values (`SECRET_PLACEHOLDER_UNKNOWN`).
- **Auto-screenshot skip:** the MCP layer deep-scans args for `{{secret:` and replaces the after-action screenshot with a text note — otherwise a non-secure-entry field hands the plaintext straight back to the model as an image. Covers nested `run-sequence` steps too.
- Remote-routed setups resolve on the machine that types (the tool-server's env) — documented in the skill.
- Inert unless a placeholder is used; no feature flag.

## Docs

- `keyboard` / `paste` tool descriptions
- `argent-device-interact` skill: "Typing secrets" section (incl. *don't* `describe`/`screenshot` the field you just filled — submit first)
- `argent-create-flow` skill: never record a real credential; placeholder flows stay secret-free in the repo

## Known residual risks (shared with Stagehand-style designs)

- A prompt-injected agent could direct `{{secret:X}}` into an attacker-visible field and screenshot it — inherent to any "agent types secrets" design; the prefix allowlist and no-read-back keep the blast radius to secrets deliberately exposed.
- `describe` can read back a non-secure field's value after typing; iOS secure fields briefly show the last typed character; Android typing puts the resolved string on the device-side `input` command line. Acceptable for emulator CI; noted in the skill.
- `keys` in the result still reflects the real keystroke count (secret length) — kept for debuggability.

## Tests

- `keyboard-secrets.test.ts`: resolution (embedded/repeated placeholders, malformed shapes ignored), unknown-name error lists names but never values, redaction preserves error class, full `execute` round-trips on the chromium branch asserting the plaintext appears nowhere in results or thrown errors, paste via mocked `sendCommand`.
- `auto-screenshot.test.ts`: `containsSecretPlaceholder` on flat/nested/unserializable args (fails safe).
- Full suites green: tool-server 2550, argent-mcp 76, registry 87; builds + `typecheck:tests` + eslint + prettier clean.


filip131311:
Superseded by #506 — this branch was accidentally created on top of `filip/fix-pnpm-local-init` (concurrent session in the same checkout), so the diff here included unrelated installer commits from #501. #506 is the same single commit rebased onto main, containing only the secret-placeholder work.

PR #507, opened 2026-07-15
feat(flow): allow to compose all flows via `run:`

j-piasecki:
## Summary

`run:` accepted **fragments only** — an e2e flow (one starting with `launch:`)
was rejected as a `run:` target. This lifts that: an e2e flow can now be nested
via `run:`, its `launch` running inline (a real terminate + relaunch on native).
Composition becomes uniform — any flow can be a reusable building block.

## What changed

- **`flow-run.ts`**: dropped the `isE2eFlow` rejection in `execRunStep`; cycle /
  `MAX_RUN_DEPTH` guards still bound recursion.
- **`flow-add-step.ts`**: the recorder now captures a recorded `flow-execute` of
  an e2e sibling as a portable `run: <name>` directive (was: raw `tool:` step).
- **`flow-utils.ts`, `argent-create-flow/SKILL.md`**: updated docs/skill that
  said e2e flows can't be `run:` targets.

## Known limitation — Chromium

Chromium `launch` boots an Electron app from a *path*, wired to the top-level
flow only — so a nested **chromium** e2e flow doesn't boot its own app. Native
nesting is correct; chromium nested e2e is unsupported.

## Testing

Rewrote the two affected tests (composition + recorder). Full flows suite green
(349 tests); `tsc --noEmit` clean.

### DIFF
--- a/packages/skills/skills/argent-create-flow/SKILL.md
+++ b/packages/skills/skills/argent-create-flow/SKILL.md
@@ -11,7 +11,7 @@ Flows store **no device id**: the runner binds a device (the single booted one, 
 
 **Two flow types**
 
-- **e2e** — begins with a `launch:` step, which starts that app from scratch (terminate + relaunch), so the flow controls its own start state. No `executionPrerequisite`. May `run:` fragments; cannot itself be a `run:` target. Record one by adding a `restart-app` of the app under test as the **first** step — it is captured as the `launch` step.
+- **e2e** — begins with a `launch:` step, which starts that app from scratch (terminate + relaunch), so the flow controls its own start state. No `executionPrerequisite`. May `run:` other flows, and (on iOS/Android) may itself be a `run:` target — when nested, its `launch` runs inline, restarting the app for that sub-scenario. **Chromium is the exception:** the runner boots one Electron app per run (the top-level flow's), so a nested chromium e2e flow's `launch` can't boot its own instance and fails the run — keep chromium e2e flows top-level. Record one by adding a `restart-app` of the app under test as the **first** step — it is captured as the `launch` step.
 - **fragment** — doesn't begin with a launch; runs against the device's current state. May declare an `executionPrerequisite` (a documented entry-state contract). Invoked from other flows via a `run:` step, or directly by you at any time.
 
 Both run via `argent flow run <name>` — a fragment simply runs against whatever is on screen (its prerequisite is printed as a reminder). Only e2e flows are meaningful CI/suite entries, since only they give a deterministic verdict from a clean start.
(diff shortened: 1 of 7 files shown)


filip131311 (approved):
Approving — the native half does exactly what it claims, verified live on a simulator and an emulator from a real `npm run pack:mcp` tarball (not just unit tests).

### What I verified

The decisive test: a parent e2e navigates deep, then `run:`s a nested **e2e** whose `launch` must genuinely restart the app.

- **Android emulator** — 10/10. Parent went deep into Settings (Brightness visible), the nested `launch` really restarted the app, Brightness gone and the root list back.
- **iOS simulator** — 10/10 against a real RN app (Expo `my-app`): Explore tab → nested `launch` → back on Home.
- **A/B vs pre-PR** (released 0.16.0): the identical flow errors with `"nested-e2e" is an e2e flow ... only fragments can be run from another flow`; this build passes. Clean proof the change is what enables it.
- **Recorder** — pre-PR emitted `tool: flow-execute` with a baked-in `device:` and an absolute `project_root`; this PR emits a portable `run: nested-e2e` that replays green. Real portability win.
- **Guards intact** with e2e on both sides: cycle (`a→b→a`), self-reference, and `MAX_RUN_DEPTH` firing at exactly 20.
- Fragment composition unregressed; 349 flows / 4074 total tests pass; lint + `tsc` clean. (Three suite failures are unrelated: two `scripts/*.test.mjs` are node:test files that only run under `npm run test:scripts`, and `lens-tools-platform-gate` passes in isolation — load-order flake.)

Backward compatible: this only widens what's accepted, and existing raw `tool: flow-execute` steps keep working.

---

### ⚠️ Warnings before merge

**1. The chromium recorder rewrite is a regression, not just an unsupported gap.**

`captureRunTarget` has no platform awareness, so it silently rewrites a chromium e2e `flow-execute` into `run:` **with no warning** — and `run:` never boots the nested app, because `chromiumBootSpec` only consults the *top-level* flow.

I isolated this with two distinguishable Electron apps against a booted chromium device, the nested flow declaring a different app (B) than the parent (A):

- **pre-PR** (`tool: flow-execute`) → **PASS**. The inner flow booted its own instance (`chromium-cdp-64675` — a distinct device id from the bound one) and asserted app B's marker, `ok: true`.
- **post-PR** (`run:`) → **FAIL**. The nested `launch` reported ✓ while booting nothing; the steps ran against app A.

So a recording that passes live now produces a flow that behaves differently on replay. The PR body calls chromium nested e2e "unsupported", but the recorder happily *emits* it.

The fix is small: `captureRunTarget` already parses the target and throws the result away. Keep it, and when the target `isE2eFlow` and its leading launch declares a chromium entry (`chromiumLaunchSpec(first.app)` non-null), keep the raw step plus the existing warning.

**2. The nested chromium `launch` passes silently.**

`runLaunch` returns `{ok: true}` when `state.chromiumBooted`, so the wrong-app case only surfaces if the nested flow happens to assert something. Without an assert it's a green run against the wrong app. It should fail loudly instead.

**3. The shipped skill doc overstates.**

`argent-create-flow/SKILL.md` now says an e2e flow "may itself be a `run:` target — when nested, its `launch` runs inline, restarting the app for that sub-scenario", with no chromium caveat. That's true on native and false on chromium. The limitation currently lives only in the PR description, not in the agent-facing doc that agents actually read.

---

Also: the branch is based on 0.15.0 while main is now 0.16.0 — worth a rebase.

Non-blocking, found while testing and not caused by this PR: iOS flows resolve selectors from the injected native-devtools hierarchy rather than the ax-service `describe` tree, so `com.apple.*` apps can never be flow targets — which means this skill's own worked example (`launch: com.apple.Preferences` + `tap: { text: General }`) can't replay on iOS. Probably deserves its own issue.

j-piasecki:
I kept the recorder rewriting the `flow-execute` to `run`, but added an explicit error when it's used with Chromium, so it fails loudly. The inner flow running on its own instance, which dies when it ends, is not the correct result.

The proper chromium lifecycle handling is something I'd like to do soon-ish.

PR #508, opened 2026-07-15
fix(tool-server): resolve `ps` by absolute path so VVD dedup survives a sanitized PATH

finloop:
## What

A running Vega VVD registers a second adb transport (`emulator-<port>`), so it can surface in `list-devices` **twice** — once as `platform:"vega"` and once as `platform:"android"`. There's dedup logic (`resolveVvdShadowAdbSerials` + `filterVvdShadowsFromAndroid`) that drops the adb "shadow" row by matching the running VVD's console port — read from the process table via `ps` — against the adb serial's port.

That dedup was spawning a **bare `"ps"`**, which relies on `PATH`.

## Why it broke

An MCP server is frequently launched from a GUI / launchd context that inherits a **sanitized `PATH`** omitting `/bin`. There, `execFile("ps", …)` fails with `ENOENT`; the surrounding `catch` returns an empty port set, and the dedup silently no-ops — so the VVD shows up as two devices. It works fine from an interactive shell (where `/bin` is on `PATH`), which is why it doesn't reproduce in a terminal or in the existing tests.

Observed: `list-devices` returned both a phantom `platform:"android"` `emulator-5554` and the real `platform:"vega"` device for a single running VVD.

## Fix

Pin `ps` to its canonical absolute location, matching how the codebase already hardcodes `/bin/sh` (`android-binary.ts`, `vega-cli.ts`, `check-deps.ts`):

```ts
export const PS_BIN = ["/bin/ps", "/usr/bin/ps"].find((p) => existsSync(p)) ?? "ps";
```

Applied to all four `ps` spawn sites:
- `vega-process.ts` — `listRunningVvdConsolePorts`, `listRunningVvdPids`
- `vega-cli.ts` — `collectDescendantPids`, `pgidMembers` (same latent PATH fragility in the process-reaping paths)

The bare-`"ps"` fallback is retained only if neither canonical path exists, preserving `PATH` resolution for an atypical layout.

## Tests

- New regression test in `vega-process.test.ts` that strips `/bin` and `/usr/bin` from `PATH` and asserts the probe still spawns `ps` (a bare `"ps"` ENOENTs there — verified the test fails against the pre-fix code).
- Updated the `list-devices.test.ts` mock to match `ps` by basename, since production now passes an absolute path.
- `npm run build`, typecheck, lint (`--max-warnings 0`), and the full vega/list-devices suite (47 tests) all pass.

## Verification

End-to-end `list-devices.execute()` under a launchd-style minimal `PATH` now returns exactly **one** device (the Vega), with no android shadow. `PS_BIN` resolves to `/bin/ps` and reads the VVD's console port even with `/bin` removed from `PATH`.

### DIFF
--- a/packages/tool-server/src/utils/vega-cli.ts
+++ b/packages/tool-server/src/utils/vega-cli.ts
@@ -184,10 +184,17 @@ async function reapVegaGroup(child: ChildProcess): Promise<void> {
  * group kill is the primary mechanism; this is insurance).
  */
 async function collectDescendantPids(rootPid: number): Promise<number[]> {
-  const { stdout } = await execFileAsync(PS_BIN, ["-A", "-o", "pid=,ppid="], {
-    timeout: 1_500,
-    maxBuffer: 16 * 1024 * 1024,
-  });
+  let stdout: string;
+  try {
+    ({ stdout } = await execFileAsync(PS_BIN, ["-A", "-o", "pid=,ppid="], {
+      timeout: 1_500,
+      maxBuffer: 16 * 1024 * 1024,
+    }));
+  } catch {
+    // `ps` unavailable / timed out — honor the documented [] contract (the group
+    // kill is the primary mechanism; this walk is only insurance).
+    return [];
+  }
   const childrenByParent = new Map<number, number[]>();
   for (const line of stdout.split("\n")) {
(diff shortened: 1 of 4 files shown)


latekvo (approved):
GG :shipit:

PR #515, opened 2026-07-16
feat(screenshot-diff): support full-image comparisons

j-piasecki:
## Summary

- Add an explicit top-mask policy to the shared PNG differ.
- Preserve status-bar masking by default while allowing isolated image regions to compare every row.
- Add real-PNG regressions for top-band changes and one-row images.

This prepares the differ for cropped comparisons without changing existing callers.

## Testing

- Screenshot-diff tests
- Tool-server build

### DIFF
--- a/packages/tool-server/test/screenshot-diff.test.ts
+++ b/packages/tool-server/test/screenshot-diff.test.ts
@@ -204,7 +204,12 @@ describe("diffPngFiles", () => {
       ...rectPixels(4, 20, 4, 4, { r: 255, g: 0, b: 0 }),
     ]);
 
-    const result = await diffPngFiles({ baselinePath, currentPath, outputDir: dir });
+    const result = await diffPngFiles({
+      baselinePath,
+      currentPath,
+      outputDir: dir,
+      topMask: "status-bar",
+    });
 
     expect(result.regions).toEqual([
       expect.objectContaining({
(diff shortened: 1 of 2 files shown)


packages/tool-server/test/screenshot-diff.test.ts
```diff
@@ -218,6 +218,60 @@ describe("diffPngFiles", () => {
     expect(readRgb(diff, 5, 21)).toEqual({ r: 0, g: 200, b: 0 });
   });
 
+  it("compares the full image when the top mask is disabled", async () => {
```
latekvo, line 226: The only `topMask` value the suite exercises is `"none"` - explicit `"status-bar"` is never passed anywhere, and the masked assertions in this file all rely on the option being omitted. Since the omitted case is normalized to `"status-bar"` at the call site before the policy check, the tests can't distinguish an explicit `"status-bar"` from a policy value that disables masking. Concretely: a refactor that moves the defaulting into `resolveDiffSettings` and checks the parameter's truthiness instead of comparing against `"none"` makes an explicit `"status-bar"` stop masking entirely while all 16 tests in this file stay green (ran that mutation locally to confirm).

j-piasecki: Updated the test to pass "status-bar" explicitly in [16fe355](https://github.com/software-mansion/argent/commit/16fe3555ec2cdbf5f28e7851c2346253c2ee4c93)

latekvo:
In case this change affects screenshot diff output, could we please get a before and after image?

j-piasecki:
|Baseline|Changed|No mask|Default mask|
|-|-|-|-|
|Simulator Screenshot - iPhone 17 Pro - 2026-07-16 at 15 51 37|Simulator Screenshot - iPhone 17 Pro - 2026-07-16 at 15 51 32|after-no-top-mask-full|before-status-bar-mask-full|

Is this what you meant?

latekvo (approved):
Reviewed the top-mask policy change. Verified against a fresh tool-server build (declaration emit clean), the full test suite (2667/2667 passing, including the 16 screenshot-diff tests), and traced the `topMask` behavior end to end.

- Existing callers omit `topMask` and resolve to `status-bar`, yielding a settings object identical to the previous code, so current behavior is unchanged.
- The `none` path was checked across every downstream consumer of the top-band cutoff: the pixel-scan loop bounds, the region tracer reaching y=0, rectangle stroke clamping, the OCR top-cutoff gate, size normalization, and 0/1-row images. All correct on both paths.
- Both new tests are numerically exact and act as real regression guards for the plumbing.

Clean and well-scoped. Approving.

PR #516, opened 2026-07-17
feat(flow): snapshot cropOn for single-element visual diffs

j-piasecki:
`snapshot` gains an optional `cropOn: <selector>` — diff one element's region instead of the whole screen:

`- snapshot: { name: cart-total, cropOn: "Order summary" }`

- The selector resolves like a directive target: settled tree, auto-wait, the standard
  not-found failure. Selector-only — no point form (a point has no extent to crop to).
- The frame resolves **before** the capture, and the **cropped** image is what gets
  compared, stored as the baseline, and reported as the `current` artifact. The pixel
  rect is the normalized frame × capture size; the crop is a pngjs `bitblt` written to
  a swept scratch dir and registered only when a failure returns it.
- The baseline key stays on the **full** capture's WxH (device-class identity), so the
  previously unreachable `dimensionMismatch` guard becomes load-bearing: an element that
  grew or shrank fails with "nothing was compared" plus a crop-fixed-size-containers
  hint. Deliberately strict — a 1px layout drift fails on dimensions, not content (the
  same tradeoff as Playwright element screenshots).
- A tree-source outage while resolving `cropOn` errors the step instead of degrading to
  a full-screen compare against a cropped baseline.
- No tool-schema changes; `snapshotKey` and `--output` behavior ride along unchanged.

Tests: parse/serialize round-trip + rejection cases, and runSnapshot coverage with real
PNGs (crop dims, key vs. content, artifact retention/sweep, unresolved selector, abort,
size-drift message, sub-pixel region).


filip131311 (changes requested):
Reviewed the code and tested the **packaged tgz** (built from this branch via `npm run pack`, installed into an isolated prefix) with `argent flow run` against a real **iOS simulator** (iPhone 17 Pro, an Expo app) and a real **Android emulator** (Pixel_9, Settings). Build, lint and the suite are green (2681 tests / 271 files).

The implementation is careful and most of it holds up under live testing — but the safety property the PR is built around does not hold, and I could demonstrate that on device. Details below.

## What I verified working (live, from the tgz)

- **The crop is exact and the artifact is what was compared.** `cropOn: Network` → a 436x71 baseline out of a 1080x2424 capture, stored as `network-row__android-1080x2424.png`: content is the crop, filename is the device class, exactly as designed. `--output` exported cropped baseline/current/diff.
- **DPR genuinely doesn't enter.** On iOS the same element math produced 1014x72 for a 338x24pt element at DPR 3.
- **The crop follows the element.** After scrolling the row up, the cropped snapshot still matched its baseline at `diff 0.00%`, where a full-screen snapshot would have failed outright. This is the feature earning its keep.
- **No regression to full-screen snapshots** (`topMask: "status-bar"` preserved, 0.00% on repeat runs); missing-baseline still fails without seeding; an unresolvable `cropOn` fails with the standard `scroll-to` hint and without capturing; scratch dirs are swept on pass and retain exactly the registered file on failure.
- **`waitForFrame`/`cropPngFile` rect math, bitblt bounds-safety, and the YAML round-trip** across every selector form all checked out.

## Finding 1 (blocking): the `dimensionMismatch` guard never fires for the case it's advertised for

The PR description says the guard "becomes load-bearing: an element that grew or shrank fails with `nothing was compared`", and `SKILL.md:69` promises "an element that grew or shrank **by a pixel** fails on dimensions ... rather than on content". Neither holds.

`diffPngFiles` runs `normalizeToCommonSize` (`screenshot-diff.ts:135`) **before** it ever builds a `dimensionMismatch`, and that helper only bails when the **aspect ratios** diverge by more than 1% (`resize.ts:24`, `ASPECT_RATIO_TOLERANCE = 0.01`). Any crop that preserves its aspect ratio is Lanczos-resampled onto the baseline's size and compared as if nothing changed — so `flow-visual.ts:277` is unreachable for proportional size drift.

Measured live on the emulator, live element 436x71 vs doctored baselines:

| baseline | reality | result |
| --- | --- | --- |
| 437x71 | 1px drift — the doc's literal claim | **PASS** `diff 0.14%` |
| 480x78 | ~10% shrink | **PASS** `diff 0.33%` |
| 872x142 | element **halved** | **PASS** `diff 0.01%` |

So `- snapshot: { name: cart-badge, cropOn: "Total" }` over a badge that doubles in size reports `pass` / `diff 0.00% <= 0.5%`. The regression the step exists to catch is the one it cannot see.

This is inert on `main` (the full-screen key embeds the capture dims, so drift changes the filename first) — it only bites under `cropOn`. The guard *is* reachable, just not where advertised: a scroll that clipped the element's visible extent (436x71 -> 436x60, aspect change) failed correctly with the new size-drift message.

**Fix direction:** under `cropOn`, compare the crop's dims against the baseline PNG's dims directly before calling the differ, or thread a `normalize: false` option through `diffPngFiles`. Size-normalization is right for its original purpose (0.3x baseline vs 1.0x capture) and shouldn't apply to crops at all. The SKILL.md sentence needs to match whatever lands.

## Finding 2: the baseline key carries no crop identity, so same-name crops collide silently

`flow-visual.ts:189` keys on `<name>__<platform>-<fullW>x<fullH>`. Nothing distinguishes one `cropOn` selector from another, and there's no duplicate-snapshot-name validation. Confirmed live:

```yaml
- snapshot: { name: row, cropOn: "Network & internet" }   # ✓ baseline written
- snapshot: { name: row, cropOn: "Connected devices" }    # ✓ baseline updated  <- same file
```

Both steps reported success under `--update-baselines` and left **one file on disk** — the first element's baseline silently discarded. On a normal run the two steps then compare against each other's images, and `--update-baselines` never converges (each pass fixes one step and breaks the other).

`main` made this unlikely (one snapshot per screen); `cropOn`'s natural pattern is several crops of the *same* screen, which actively invites reusing that screen's name. Finding 1 amplifies it: any two crops of similar aspect now normalize into a comparison rather than failing. Folding a short hash of `describeSelector(cropOn)` into `snapshotKey`, or rejecting duplicate names at parse time, closes it.

Credit where due: the loud case works — mixing a full and a cropped snapshot under one name failed with `nothing was compared` rather than going green.

## Smaller points

- **`flow-execute`'s tool description (`flow-run.ts:392`) needs a `cropOn` mention.** It enumerates every other directive with its options inline (`long-press: { on, duration }`, `scroll-to`'s momentum-free behavior), but still describes `snapshot` as a plain screenshot diff. It's the only other place documenting the directive; SKILL.md is updated.
- **The new section's own example works against its advice.** `cropOn: "Order summary"` is a text selector, and selector resolution picks the *smallest* matching node — the label, whose frame is text-metrics-sized — not the fixed-size container the same sentence tells you to crop. An `id` selector would match the prose.
- **`topMask: "none"` (`flow-visual.ts:268`)** is the right call for an arbitrary crop, but a crop overlapping the screen's status-bar band (a top nav/header — a natural target) now leans entirely on `pinStatusBar`, which is documented as best-effort with failures swallowed. Masking only crop rows that overlap the screen's top band would be strictly better; a doc note is the minimum.
- **The sub-pixel-crop failure returns neither `snapshotKey` nor artifacts**, so `--output` exports nothing and there's no image to look at.

## Testing notes for anyone reproducing this

- `com.apple.Preferences` (and any Apple system app) can never be injected, so flow selectors can't resolve against it — system apps are unusable as a `cropOn` testbed on iOS.
- With several argent installs on one machine, tool-servers contend for the same per-UDID `/tmp/argent-nd-<UDID>.sock`, which surfaces as the "stale or duplicate argent server may be holding the devtools connection" error. Restarting the server you want to win is the workaround.

packages/tool-server/src/tools/flows/flow-visual.ts
```diff
@@ -116,97 +198,171 @@ export async function runSnapshot(
...
+  const currentArtifact = async (): Promise<ArtifactHandle> => {
+    if (cropDir === undefined) return shot.image;
+    keepCropped = true;
+    // Explicit filename: the crop file's basename IS the baseline's `key`, and
+    // a remote client materializes downloads by filename — defaulting to the
+    // basename would land `current` on the same cache path as `baseline` and
+    // clobber it (the diff artifact below disambiguates the same way).
+    return store.register(currentPath, {
+      mimeType: "image/png",
+      filename: `${snapshotKey}-current.png`,
```
latekvo, line 234: The cropped `current` artifact is registered here with an explicit filename that sets it apart from the baseline. This override is load-bearing: the cropped file on disk shares the baseline's basename (`<snapshotKey>.png`, since `croppedPath = path.join(cropDir, key)`), and `register` defaults an artifact's download filename to that basename (`registry/src/artifacts.ts` — `filename = opts?.filename ?? basename(hostPath)`). On the over-threshold and dimension-mismatch failure paths both `baseline` and `current` are registered, so without this line both handles would carry the identical filename and a remote client materializing downloads into a filename-keyed cache would land `current` on the same path as `baseline` and clobber it.

Nothing in the test suite asserts this. The cropOn tests check the `current` handle's `hostPath`, its pixel size, and scratch-dir cleanup, none of which change if this override is removed — the on-disk file and its basename are unaffected, only the handle's `filename` field would silently revert to the colliding default. The one `filename` assertion in the file is on the `diff` artifact of the full-screen path. As a result, a change dropping this override would reintroduce the exact baseline/current collision on remote materialization while the entire suite stays green.

j-piasecki: Pinned in c28a0b90: the over-threshold and dimension-mismatch tests now assert the `current` handle's `filename` is `${cropKey}-current.png` and the baseline's is `${cropKey}.png`. Verified sensitivity by temporarily dropping the override — exactly those assertions fail, with `current` reverting to the colliding baseline filename.

latekvo:
[[Diplomat](https://github.com/latekvo/Diplomat)]: Ran a full end-to-end review sweep on this branch at `64d41d97` and it came back clean.

What I exercised:
- **The real crop → differ pipeline end to end** (device I/O stubbed, but real `diffPngFiles` / `cropPngFile` / PNG encode-decode-bitblt): adopting a cropped baseline, a matching crop passing at `0.00%`, a content change failing with a real annotated context diff, and the size-drift hard-fail landing correctly — including the proportional (same-aspect) resize case, which `normalizeSizes: false` now catches instead of resampling away.
- **Full `tool-server` suite green.** The one red is a pre-existing ~1ms real-timer flake in `flow-type-focus.test.ts` (untouched by this PR, and passing on isolated re-runs), not a regression here.
- Lenses swept without a finding: crop-rect / differ correctness, temp-dir lifecycle / cleanup / abort / error paths, `cropOn` parse-serialize round-trip and validation, cross-feature behavior (`when` / `run` fragments, `updateBaselines`, same-name key disambiguation), platform variance (iOS / Android / Chromium / Vega — the crop reuses the same normalized frame↔capture space tap resolution already relies on), and path-safety of the new key and download filenames.
- The earlier review points all check out at this head: the size-drift guard, crop identity folded into the baseline key, sub-pixel region handling, crop masking policy, the distinct `current` download filename, and the docs.

Thanks for the careful, well-tested change!

latekvo (approved):

filip131311 (approved):
sorry for blocking

PR #517, opened 2026-07-17
feat(screen-recording): screen-recording-start/stop tools with per-call stop reminders

latekvo:
## What

Adds a `screen-recording-start` / `screen-recording-stop` tool pair (naming follows `native-profiler-start/stop`) that records the device screen to an mp4 and returns it as a downloadable artifact.

Capture runs through **simulator-server** — the same backend `screenshot` and every interaction tool already drive. simulator-server publishes the device screen as an MJPEG stream (`/stream.mjpeg`), so recording needs no platform-specific recorder: the tool-server subscribes to that stream, paces the frames onto a fixed 30 fps timeline, and pipes them into a single ffmpeg process that encodes — and watermarks — straight to the final mp4. The stream is multi-subscriber, so a preview window can stay open on the same device while recording.

- **One path for every device simulator-server drives** — iOS simulators, Android emulators, and physical Android devices.
- **Native resolution, constant 30 fps** — h264, `+faststart`, no post-processing pass.
- Session state lives in a per-device `ScreenRecordingSession` blueprint (mirrors the native-profiler session): time-limit cap with auto-finalize, unexpected-exit recovery (a capture that ended on its own is still retrievable via stop), always-startable state reset on stop, and dispose-on-shutdown cleanup through the registry.

## Why the frames are paced host-side

A device only emits a frame when the screen **changes**. Pointing ffmpeg straight at the stream therefore produces a video whose length is the sum of the moments something moved — a still screen collapses to a fraction of a second — and an ffmpeg blocked on a silent socket does not answer a stop signal promptly either (measured: 12 s late). Re-emitting the last frame on a wall-clock schedule fixes both: **video duration always equals real elapsed time**, and stop is a stdin EOF that finalizes in ~60 ms. Duplicate frames cost almost nothing once encoded (a 10 s recording of a static screen is ~1 MB).

## Finishing pass folded into capture

The Argent watermark (logo + `By @swmansion`, bottom-left, sized relative to the frame) is stamped by the same ffmpeg invocation that encodes the capture, so there is no second encode and no temp intermediate. Its colour is chosen **per pixel** from the background luma underneath it — the white logo over dark areas, a runtime-derived near-black copy over light areas — so it keeps maximum contrast against whatever is behind it. It sits at 20 % opacity and is opt-out via `argent disable video-watermark` (read when the recording starts). Frame geometry comes from the first JPEG's SOF header, so no `ffprobe` is needed.

ffmpeg is now the recorder rather than an optional finisher, so a missing ffmpeg fails `screen-recording-start` up front with an install hint instead of silently degrading the output at stop time.

## Reminder plumbing

While a recording is running — or ended on its own but was never retrieved — **every** tool result carries a `note` telling the agent the recording is still going and how to stop it. It rides the same `note` field as the update reminder but is per-call and unsuppressed; both notes compose when they coincide. The note text switches to "already ended, retrieve the file" once the cap fires or the capture dies.

## Skill

New `argent-screen-recording` skill (+ routing stanza in `rules/argent.md`): documents the start → interact → stop workflow and instructs the agent to schedule its own reminder/wakeup for the expected end of the recording immediately after starting one (harness reminder tool if available, otherwise a background `sleep` whose completion notification pulls it back). Scores 10/10 on `scripts/grade-skills.mjs`.

## Verification

- Unit: 45 tests across `screen-recording.test.ts`, `mjpeg-stream.test.ts`, `watermark.test.ts`, `http-recording-note.test.ts`; full tool-server suite **2808 tests / 280 files green**; `tsc --noEmit -p tsconfig.test.json`, root `tsc --build`, eslint, prettier all clean.
- E2E on a live tool-server (`POST /tools/...`), driving real interactions between start and stop:
  - **iOS sim** (iPhone 16 Pro Max, iOS 18.5): 1320×2868 h264, `r_frame_rate=30/1`, 338 frames, container duration 11.27 s vs reported `durationMs` 11314 — **within 50 ms of wall clock**. Extracted frames show real screen content (mid app-switcher animation) with the watermark rendered light-over-dark.
  - **Android emulator** (Pixel 3a, API 34): 1080×2220 h264, 30 fps, 353 frames, duration 11.77 s vs `durationMs` 11793 — within 30 ms. 512 KB for a mostly-static home screen.
  - **Static screen** (no interaction at all, 10 s): 300 frames, duration exactly 10.000 s — the case the previous `recordVideo` implementation reduced to ~0.03 s.
  - **Long silence then activity** (15 s untouched, then an app launch): stream survived, no warning, frames after the silence show the new screen. This caught a real bug during review — the connect timeout was also acting as a socket *inactivity* timeout and tore the stream down at 10 s of stillness; fixed and covered by a regression test that fails without the fix.
  - **Time-limit cap**: `timeLimitSeconds: 5` self-finalized at 4.97 s / 149 frames, the per-call note flipped to "already ended (it hit its 5s time limit)", and a stop issued 11 s later still returned the video with the matching warning.

## Not supported (capability-gated with clear errors)

tvOS simulators, physical iPhones, Chromium apps, Vega, and remote (`remote:`-prefixed) simulators — none of them expose a readable frame stream.


packages/tool-server/src/tools/screen-recording/session-guards.ts
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * Reject a start while a capture is live OR while another start/stop is
+ * mid-flight on the same session. The pending flags are set synchronously
+ * before the first await of start/stop, closing the check-then-stamp gap that
+ * would otherwise let two overlapping calls both pass this guard and
+ * cross-corrupt the shared session state.
+ */
+export function assertNoActiveRecording(api: ScreenRecordingSessionApi, stage: string): void {
+  if (api.recordingActive || api.startPending || api.stopPending) {
```
hubgan: **Blocking.** This admission check gates on `recordingActive || startPending || stopPending` but not on `pendingRetrieval`. After the time-limit cap fires (or the encoder dies) the session sits in `pendingRetrieval` with a finalized video still on disk — the state that `assertStoppableSession` and the reminder note both treat as recoverable via `screen-recording-stop`. A new start in that state is accepted, and `startCaptureLocked` then overwrites `outputFile`/`logoFile` and clears `pendingRetrieval`, so the earlier recording's video is orphaned in the temp dir (no longer retrievable via stop), its watermark logo temp leaks, and its "retrieve me" reminder is replaced.

Reproduced live: start with a 5s cap, let the cap fire, then start again without stopping — the second start is admitted, stop returns the second video, and the first mp4 is left orphaned on disk. Silent data loss on a natural start-after-cap flow, so flagging as blocking.

packages/tool-server/src/tools/screen-recording/mjpeg-stream.ts
```diff
@@ -0,0 +1,230 @@
...
+            }
+          }
+          buffer = buffer.subarray(frameEnd);
+        }
+        // A stream that never yields a delimiter (wrong endpoint, proxy that
+        // rewrites the body) must not grow the heap without bound.
+        if (buffer.length > MAX_BUFFER_BYTES) buffer = Buffer.alloc(0);
+      });
+
+      const fail = (err: Error) => {
```
hubgan, line 130: **Non-blocking.** `fail()` records the stream error on `state.error` but never settles a pending `waitForFirstFrame`. If the connection errors or aborts after the stream object has resolved but before the first frame arrives (simulator-server restart, or the device screen going off mid-handshake), `waitForFirstFrame` does not reject on that error — it blocks the full `FIRST_FRAME_TIMEOUT_MS` (10s) and then throws the timeout message ("No frame arrived... Is the device booted and its screen on?"), which masks the real cause. The recording still aborts cleanly with no leak, so the impact is a 10s stall plus a misleading diagnostic rather than a correctness problem. Non-blocking.

packages/tool-server/src/tools/screen-recording/watermark.ts
```diff
@@ -0,0 +1,137 @@
...
+const even = (n: number) => 2 * Math.round(n / 2);
+
+/** The bottom-left watermark rectangle in pixels (all even) for a frame size. */
+export function computeWatermarkBox({ width, height }: Dimensions): WatermarkBox {
+  const w = Math.max(2, even(width * WATERMARK_WIDTH_FRACTION));
+  const h = Math.max(2, even(w / LOGO_ASPECT));
+  const margin = even(width * WATERMARK_MARGIN_FRACTION);
+  const bottomMargin = even(width * WATERMARK_BOTTOM_MARGIN_FRACTION);
+  const x = Math.max(0, Math.min(even(margin), even(width - w)));
+  const y = Math.max(0, Math.min(even(height - h - bottomMargin), even(height - h)));
```
hubgan: **Non-blocking (low).** `computeWatermarkBox` clamps `x`/`y` into the frame but never bounds `w`/`h`, so for pathological inputs the crop rectangle can exceed the frame: an aspect ratio beyond ~11:1 makes `h` taller than the frame, and a sub-~55px width with an odd height pushes `y + h` one past the bottom. ffmpeg's `crop` then fails (code -22) and aborts the whole recording instead of degrading to no-watermark. Not reachable at any real device resolution (every real frame leaves a comfortable margin), so this is a hardening note on the box geometry rather than a live defect. Non-blocking.

packages/tool-server/src/tools/screen-recording/session-guards.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat)]: Confirmed and fixed in 08e7e71908eb0725c7499cc151ca322b706afe6c.

Reproduced the orphaning at the unit level: start with a 5s cap, let it fire so the session lands in `pendingRetrieval` with the finalized mp4 still at `outputFile`, then start again — the second start was admitted, `startCaptureLocked` overwrote `outputFile`/`logoFile` and cleared `pendingRetrieval`, and the first video was left unreachable via stop.

The admission guard now also rejects when `pendingRetrieval && outputFile !== null`, with a message pointing the caller at `screen-recording-stop` to hand the finalized video over before starting a new capture. Regression test (`rejects a start while a capped recording still awaits retrieval, keeping the first video reachable`) fails without the guard change (the second start is admitted and `outputFile` is clobbered) and passes with it.

packages/tool-server/src/tools/screen-recording/mjpeg-stream.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat)]: Confirmed and fixed in 2c1cec38ead3cc9168148a1cf6cc089dda85b0e4.

Reproduced against a real local server that sends the response headers (so the stream object resolves) then destroys the socket before any complete frame arrives: `waitForFirstFrame(5000)` blocked the full 5s and then threw the misleading "No frame arrived… is the screen on?" timeout — measured 5015ms in the repro.

`fail()` now rejects a pending first-frame waiter the moment the drop is observed (and `waitForFirstFrame` short-circuits if the error already landed), both with a network-kind `SCREEN_RECORDING_STREAM_UNAVAILABLE` failure that names the drop. `close()` drops the waiter hooks first so our own teardown can't spuriously reject. Regression test asserts the reject now lands in <2s instead of after the timeout.

packages/tool-server/src/tools/screen-recording/watermark.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat)]: Confirmed (as a hardening note — not reachable at real device resolutions) and fixed in fd037123f43668851b56cf879c2808244d6e0e63.

`computeWatermarkBox({ width: 1000, height: 50 })` returns `h = 88 > 50`, so `y + h` exceeds the frame and ffmpeg's `crop` aborts (-22), taking the whole recording down rather than degrading to no-watermark.

`w`/`h` are now capped to the frame, and the fitting clamps round DOWN to even so the box can never spill past any edge (an even value rounded up was the "one past the bottom" edge you noted). Real resolutions are unaffected — the existing exact-geometry test (`886x1920 -> {254,78,26,1826}`) still holds — and a new regression test covers the wide-short and sub-55px odd-height cases.

packages/tool-server/src/tools/screen-recording/capture.ts
```diff
@@ -0,0 +1,498 @@
...
+  const stream = await openMjpegStream(params.streamUrl, STREAM_CONNECT_TIMEOUT_MS);
+  let logoFile: string | null = null;
+  let watermarkSkipped: string | null = null;
+  let child: ReturnType<typeof spawn>;
+  try {
+    // The first frame proves the device is actually drawing, and its JPEG
+    // header carries the frame size the watermark geometry needs — no ffprobe
+    // pass over a file that does not exist yet.
+    const firstFrame = await stream.waitForFirstFrame(FIRST_FRAME_TIMEOUT_MS);
+    const dims = params.watermark ? readJpegDimensions(firstFrame) : null;
```
hubgan, line 230: Frame geometry is read once here from the first frame and `buildWatermarkGraph(dims)` bakes absolute crops (`crop=w:h:x:y`, watermark.ts) from it. If the device rotates mid-recording and simulator-server begins emitting landscape-dimensioned frames, that absolute crop can no longer be satisfied and ffmpeg aborts the filtergraph. On the default (watermark-on) path this ends the encoder early and the capture returns a truncated mp4. I reproduced it by feeding portrait frames then landscape frames of the same size through the exact graph (`Invalid too big or non positive size ... Error reinitializing filters!`, output truncated). The no-watermark branch survives because its `crop=trunc(iw/2)*2` re-evaluates per frame, so the two branches behave differently on rotation.

packages/tool-server/src/tools/screen-recording/session-guards.ts
```diff
@@ -0,0 +1,160 @@
...
+ * waiting to be handed over (the `pendingRetrieval` state a cap or crash leaves
+ * behind). The pending flags are set synchronously before the first await of
+ * start/stop, closing the check-then-stamp gap that would otherwise let two
+ * overlapping calls both pass this guard and cross-corrupt the shared session
+ * state. Guarding `pendingRetrieval` too stops a start-after-cap from
+ * overwriting the earlier recording's `outputFile`/`logoFile` and orphaning its
+ * video on disk (stop still recovers it — the guard points the caller there).
+ */
+export function assertNoActiveRecording(api: ScreenRecordingSessionApi, stage: string): void {
+  const awaitingRetrieval = api.pendingRetrieval && api.outputFile !== null;
```
hubgan, line 41: This new admission gate rejects a start while `pendingRetrieval` holds (a capped or crashed recording awaiting retrieval). The guard itself is sound, but it collides with the touch-visualizer test added in the stacked pointer PR (#549), which starts a second recording after the cap fires without an intervening stop. That test passes today, but once this guard is in its merge base it throws `SCREEN_RECORDING_ALREADY_ACTIVE`, so the correctly-integrated tip of the stack has a red suite. Worth coordinating with #549 before the stack merges.

packages/tool-server/src/tools/screen-recording/capture.ts
```diff
@@ -0,0 +1,498 @@
...
+    clearTimeout(api.recordingTimeout);
+    api.recordingTimeout = null;
+  }
+
+  const outputFile = api.outputFile!;
+  const logoFile = api.logoFile;
+  const startedAtMs = api.wallClockStartMs;
+  const endedEarly = api.recordingTimedOut || api.recordingExitedUnexpectedly;
+  // Read before finalizing: closing the stream ourselves would look like a drop.
+  const streamError = api.frameStream?.error ?? null;
```
hubgan: On the recovery stop path (after the cap fired or the encoder exited), `api.frameStream` has already been nulled by the earlier `stopPump`/finalize, so `streamError` here is always null and the 'frame stream dropped, video may freeze' warning further down can never surface on those paths. Diagnostic-only, but a stream drop that coincided with a cap/crash is reported without the freeze hint.

hubgan (commented):
Notes inline.

packages/tool-server/src/tools/screen-recording/capture.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat)]: I could not reproduce this against the shipping pipeline — it looks like a non-issue as coded. The capture feeds whole JPEGs through `-f image2pipe`, and ffmpeg's `crop` filter *reinitializes and clamps* its offsets when the input frame size changes rather than aborting. I ran the exact production path — `buildWatermarkGraph({1170×2532})` + the real `ffmpegArgs` (image2pipe → looped logo → maskedmerge → overlay → libx264/yuv420p/+faststart) — and fed 10 portrait (1170×2532) frames then 10 landscape (2532×1170) frames mid-stream:

- exit code **0**, no `Error reinitializing filters` / `Invalid too big or non positive size`;
- valid, non-truncated mp4 out (`ffprobe`: 1170×2532, 19 packets).

Verbose logs show the graph receiving the landscape frame (`w:2532 h:1170`) and the mask crop reinitializing to `w:334 h:102` with no error (`[Parsed_crop_0] w:2532 h:1170 -> w:334 h:102`), i.e. the absolute `crop=w:h:x:y` offsets are clamped, not rejected.

The reproduction you saw (`Error reinitializing filters!`) doesn't arise through `image2pipe` + this graph — it may have used a different feeding/filter config. If you have the exact command that produced it I'm happy to reconcile, but against the code on this branch (`4bc271cb`) a mid-recording rotation yields a valid recording, so I'm not applying a change here. (Separately unverified: whether the iOS-sim stream even changes frame dims on rotation.)

packages/tool-server/src/tools/screen-recording/session-guards.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat)]: Agreed the guard is correct, and I checked the collision empirically — it doesn't hold against #549 as it currently stands. I scanned all 55 `it()` blocks in #549's `screen-recording.test.ts` (branch `feat/screen-recording-pointer`, tip `5fbc5295`) for any test that starts a second recording after a `pendingRetrieval`-inducing event (cap fire via `advanceTimersByTimeAsync(5_000)` or `child.exit(1)`): there are none. The touch-visualizer test you're thinking of — `"restores the overlay when the time-limit cap fires"` — starts exactly one recording, advances the cap, and asserts `pointer.disable` + `recordingTimedOut`; it never calls start again, so the `pendingRetrieval` guard doesn't fire for it. #549's base also doesn't yet contain this guard commit (`08e7e719`).

So there's nothing to change in #517 (the guard is sound) and nothing red in #549's current suite. If a start-after-cap test is added to the stack later it should call `screen-recording-stop` first (now the correct contract). Leaving this open for you to confirm rather than resolving a cross-PR coordination note myself.

packages/tool-server/src/tools/screen-recording/capture.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat)]: Confirmed and fixed in 220cafb2b76bc141a38ec114cd9b372c404f8855.

You're right on both counts: on the cap/crash recovery paths `api.frameStream` is already nulled by the earlier `stopPump`, so `streamError` here is always null — and even when it isn't, the `if (streamError && !warning)` gate lets the more specific cap/exit notice mask it. Reproduced with a unit test: a drop set before a 5s cap fires produced a warning of only `"Recording already ended at its 5s time limit…"`, with no freeze hint.

Fix: `stopPump` now stashes `frameStream.error` into a new `lastFrameStreamError` before nulling the reference (a clean close reports none, so no phantom drop), `stopCapture` reads it as a fallback, and the freeze hint now *appends* alongside the cap/exit notice instead of gating on `!warning`. The regression test asserts a drop-then-cap warning carries both the time-limit notice and the "may freeze" hint. Full suite green (screen-recording 72/72).

PR #519, opened 2026-07-17
fix(flow): error a no-windows iOS read instead of a false hidden pass

j-piasecki:
## Problem
On iOS, a flow `assert hidden` (and an unmet `when` guard) could pass /
green-skip against a screen where the element was plainly visible — whenever the
target's view hierarchy can't be read. A flow that launches a non-injectable app
(e.g. an Apple system app, where library validation blocks injection) runs to
exit 0 with a false pass while `describe` still shows the element on screen.

## Root cause
`queryFullHierarchyTree` returned an empty tree as a *successful* read when
`getFullHierarchy` yields no windows, and the iOS source never sets a
degraded-read hint. The runner's blind-read guard then falls back to
`everMatched` alone, so an element that was *never seen* leaves the empty tree
"trusted" — and an empty tree is exactly what a `hidden`/absent check accepts as
satisfied.

## Fix
Throw on a no-windows read instead of adapting it to an empty tree — consistent
with the existing unavailable / requiresAppRestart guards. The caller's retry
loop rides out a transient blank; a persistent one surfaces as the step's
failure (`assert` fails, `when` guard errors) with an actionable reason, never a
false pass. Keys on raw `windows`, not the flattened child count, so a genuinely
sparse screen (windows present, no matching leaves) still resolves to "not
found".

## Tests
- New `flow-ios-tree-no-windows.test.ts`: throws on empty / missing windows,
  returns a tree when windows are present, still throws on an explicit RPC error.
- `flow-hidden-blank-reads.test.ts`: adds the never-seen case (element never
  matched + dark source) → `hidden` fails with the outage reason, not a pass.

### DIFF
--- a/packages/tool-server/test/flows/flow-hidden-blank-reads.test.ts
+++ b/packages/tool-server/test/flows/flow-hidden-blank-reads.test.ts
@@ -133,6 +133,35 @@ describe("hidden timeout diagnostics", () => {
     expect(result.steps[0].reason).not.toMatch(/still visible/);
   });
 
+  it("does not pass a hidden assert when the element was NEVER seen and the tree source is dark", async () => {
+    // The repro that motivated the flow-ios-tree no-windows guard: a
+    // non-injectable target (an Apple system app — library validation) yields a
+    // tree source that cannot read the hierarchy. The element is never seen at
+    // all, so `everMatched` never flips and the blind-read guard's
+    // everMatched-only backstop can't catch it. The source therefore throws on a
+    // no-windows read: every fetch rejects, so the assert reports the outage
+    // instead of taking an empty tree as proof the element is gone.
+    currentFetch = () => {
+      throw new Error(
+        "getFullHierarchy returned no windows for com.apple.Preferences — the app is not injectable"
+      );
+    };
+
+    await writeFlow("never-seen-hidden", {
+      executionPrerequisite: "",
+      steps: [{ kind: "assert", condition: "hidden", selector: { identifier: "General" } }],
+    });
+
+    const result = await run("never-seen-hidden");
+
+    expect(result.ok).toBe(false);
+    expect(result.steps[0].status).toBe("fail");
+    expect(result.steps[0].reason).toMatch(/could not read the UI tree/);
+    expect(result.steps[0].reason).toMatch(/not injectable/);
+    // Must NOT read as a confirmed-hidden pass.
+    expect(result.steps[0].reason).not.toMatch(/still visible/);
+  });
+
   it("still reports a genuinely visible element as still visible", async () => {
     currentFetch = () => ({
       tree: screen([
--- a/packages/tool-server/test/flows/flow-hidden-blank-reads.test.ts
+++ b/packages/tool-server/test/flows/flow-hidden-blank-reads.test.ts
@@ -134,13 +134,18 @@ describe("hidden timeout diagnostics", () => {
   });
 
   it("does not pass a hidden assert when the element was NEVER seen and the tree source is dark", async () => {
-    // The repro that motivated the flow-ios-tree no-windows guard: a
-    // non-injectable target (an Apple system app — library validation) yields a
-    // tree source that cannot read the hierarchy. The element is never seen at
-    // all, so `everMatched` never flips and the blind-read guard's
-    // everMatched-only backstop can't catch it. The source therefore throws on a
-    // no-windows read: every fetch rejects, so the assert reports the outage
-    // instead of taking an empty tree as proof the element is gone.
+    // The runner's half of the no-windows fix: a `hidden` assert whose element
+    // is NEVER seen, so `everMatched` never flips and the blind-read guard's
+    // everMatched-only backstop can't catch it — the only defense is the tree
+    // source refusing the read. The rejecting fetch is SCRIPTED here (this
+    // file mocks fetchFlowTree wholesale; the message just mirrors
+    // flow-ios-tree's no-windows guard, which flow-ios-tree-no-windows.test.ts
+    // exercises at the unit level and flow-hidden-no-windows-e2e.test.ts
+    // end-to-end). What this case locks in is the caller's contract with that
+    // upstream throw: when every fetch rejects, the assert fails with the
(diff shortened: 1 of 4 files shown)


packages/tool-server/test/flows/flow-ios-tree-no-windows.test.ts
```diff
@@ -0,0 +1,93 @@
...
+  it("throws when the payload carries no windows field at all", async () => {
+    const registry = registryFor(nativeApi(() => ({})));
+    await expect(queryFullHierarchyTree(registry, IOS_DEVICE)).rejects.toThrow(/no windows/i);
+  });
+
+  it("returns a tree when windows are present", async () => {
+    const registry = registryFor(nativeApi(() => ({ windows: [windowSpanning()] })));
+    const { tree, source } = await queryFullHierarchyTree(registry, IOS_DEVICE);
+    expect(source).toBe("native-devtools");
+    expect(tree.children.length).toBeGreaterThan(0);
```
latekvo, line 101: This is the only positive case, and its `windowSpanning()` fixture always flattens to a non-empty tree (a leaf child). The guard's central invariant — keying on the raw `windows` array rather than the flattened child count, so that a windows-present-but-zero-leaf screen resolves to a trusted "not found" tree instead of throwing — is never exercised. A guard rewritten to throw when the *flattened* tree has zero children (the exact mistake the source comment warns against) still satisfies every assertion in this file, so a regression against "key on raw windows, not flattened children" would land invisibly and turn every genuinely sparse screen into an unreadable-tree error.

j-piasecki: Good catch — a guard rewritten to key on the flattened child count would indeed have passed every assertion here. Addressed in 8d4533c3: added a `windowBare()` fixture (a `UIWindow` wrapping one anonymous `UIView`, so nothing in the subtree carries an identifier, label, semantic role, or first-responder flag and the flatten emits zero leaves) and a test asserting the read *resolves* to a trusted empty tree (`source: "native-devtools"`, `children.length === 0`) instead of throwing. Sensitivity-checked it by temporarily rewriting the guard to throw on zero flattened children — exactly and only the new test fails.

packages/tool-server/test/flows/flow-hidden-blank-reads.test.ts
```diff
@@ -133,6 +133,35 @@ describe("hidden timeout diagnostics", () => {
...
+    // tree source that cannot read the hierarchy. The element is never seen at
+    // all, so `everMatched` never flips and the blind-read guard's
+    // everMatched-only backstop can't catch it. The source therefore throws on a
+    // no-windows read: every fetch rejects, so the assert reports the outage
+    // instead of taking an empty tree as proof the element is gone.
+    currentFetch = () => {
+      throw new Error(
+        "getFullHierarchy returned no windows for com.apple.Preferences — the app is not injectable"
+      );
+    };
```
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat)]: This test mocks `fetchFlowTree` wholesale (the `vi.mock("../../src/tools/flows/flow-tree", …)` at the top of the file) and its `currentFetch` unconditionally throws a hand-written "no windows … not injectable" string, so it never reaches the real `queryFullHierarchyTree` — the function this PR adds the no-windows guard to. Reverting the new guard block in `flow-ios-tree.ts` leaves this entire file green; only `flow-ios-tree-no-windows.test.ts` starts failing. So despite the docstring ("the repro that motivated the flow-ios-tree no-windows guard", "the source therefore throws on a no-windows read"), the test neither constructs a no-windows read nor runs the guard, and the `/not injectable/` assertion matches only the literal string the mock itself throws rather than any guard-produced behavior. The regression this PR fixes is gated solely by `flow-ios-tree-no-windows.test.ts`; what this case actually locks in is the caller's handling of an already-throwing fetch for a never-seen `hidden` assert — behavior that predates the fix — which the docstring's framing does not convey.

j-piasecki: Right on all counts — reverting the guard did leave this whole file green, and the docstring oversold what this case exercises. Addressed in e4219153, two parts:

- Rewrote this docstring to say what the case actually locks in: the runner's contract with an upstream throw for a never-seen `hidden` assert — with the throw explicitly called out as scripted (the mock's message just mirrors the guard's), and the `/not injectable/` check framed as pinning the fetch error's text landing in the step reason, not guard output.
- Added `flow-hidden-no-windows-e2e.test.ts` so the regression is no longer gated solely by the unit file: it drives the runner through the **real** `fetchFlowTree` → `queryFullHierarchyTree` — only the native-devtools service seam is faked, with `queryViewHierarchy` returning `{ windows: [] }` — and asserts the `hidden` step fails quoting the guard's reason (`/returned no windows for com\.example\.app/`, a string only the real guard produces here).

Sensitivity-checked by temporarily reverting the guard to a plain empty-tree return: the new e2e test fails on `result.ok` flipping `true` (the exact false pass the PR fixes), `flow-ios-tree-no-windows.test.ts` fails its throw cases, and this file stays green — matching your observation. Guard restored byte-for-byte; full flows suite 448/448.

latekvo:
[[Diplomat](https://github.com/latekvo/Diplomat)]: Ran a full end-to-end review sweep across the lenses — correctness, edge cases, ripple effects and cross-platform parity, and test quality — over two independent verification passes. That included a sensitivity check that temporarily reverts the guard to confirm the new tests genuinely gate the fix (they do: the empty-array and missing-`windows` unit cases and the `flow-hidden-no-windows-e2e` false-pass case all fail without it, while `flow-hidden-blank-reads` stays green — exactly as the docstrings describe), tracing every `fetchFlowTree` caller to confirm they each handle the new throw, and the full `test/flows` suite (502/502).

Everything came back completely clean — the no-windows guard turns an unreadable iOS read into an actionable step failure (`assert` fails, `when` guard errors) instead of a false `hidden` pass, and it correctly keys on the raw `windows` array so a genuinely sparse screen still resolves. Thank you for contributing!

latekvo (approved):
Thanks 🐈

PR #524, opened 2026-07-20
feat(flow): indent nested steps in reports by block depth

j-piasecki:
Flow reports rendered every step flat, hiding which steps ran inside a `when:` block or a `run:` fragment. The tool-server now stamps `StepReport.depth` (+1 per block directive, omitted at top level) and the CLI/MCP renderers indent step labels and under-step lines by it:

```
  ✓  1 launch
  ✓  2 when visible "What's new" — condition met
  ✓  3   tap "Skip"
  ✓  4   run dismiss [dismiss]
  ✓  5     tap "OK" [dismiss]
```

Invariants:

- **Wire-compatible both ways** — a flow with no block directives reports byte-identical to the pre-depth shape; a pre-depth server renders flat as before.
- **Skipped ≡ entered** — skipped/errored/aborted blocks report the same depths as entered ones, keeping reports comparable run-to-run.
- **Untrusted wire data** — renderers clamp depth (display cap, not a producer bound): hostile values render flat or at the cap, never throw or allocate huge strings.

Also fixes a pre-existing bug: under-step warning/artifact lines now stay aligned past 99 steps.

Tests pin depth on every producer path (entered/skipped/errored/aborted blocks, `run:` success/failure markers, `run:`-in-`when:`) and renderer alignment incl. the clamp boundary.

### DIFF
--- a/packages/argent-cli/src/flow.ts
+++ b/packages/argent-cli/src/flow.ts
@@ -29,6 +29,12 @@ export interface StepReport {
   message?: string;
   /** Human-readable step target (selector / snapshot name), set by the runner. */
   target?: string;
+  /**
+   * Nesting depth: absent/0 at top level, +1 inside each block directive
+   * (`when:` guarded steps, `run:` fragment steps). Renderers indent by it; a
+   * pre-depth tool-server sends none and the report renders flat, as before.
+   */
+  depth?: number;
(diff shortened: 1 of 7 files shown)


latekvo (approved):
You're amazing ❤️

PR #525, opened 2026-07-20
fix(tool-server): hard-kill (SIGKILL) timed-out iOS simctl calls so describe can't hang

latekvo:
## What happened

An iOS `describe` call hung for **~24 minutes** on the benchmark host, logging the incoming request and then going completely silent until the tool-server was killed.

The iOS `describe` path shells out to `xcrun simctl` for nearly every step — `isTvOsSimulator` (`simctl list`), the ax-service bootstrap (`simctl spawn defaults …`), and the native-devtools env setup (`simctl spawn launchctl …`). Every one is guarded by `execFileAsync(…, { timeout: 10_000 })`, so on paper each is bounded to 10s.

The problem is in Node itself: **`execFile`'s `timeout` sends its `killSignal` (default `SIGTERM`) exactly once and never escalates to `SIGKILL`.** When CoreSimulatorService wedges (which rapid clone/boot/delete churn induces), the `simctl` child ignores SIGTERM, its stdio never closes, and the promise never settles — so the "10-second timeout" never actually fires. There is no outer deadline on the `describe` tool to catch it, so the call hangs until the process is killed.

## Why this is the root cause (not a guess)

- **Static elimination:** every JS-level guard on the describe path (ax-service's 10s connect/query timers, native-devtools' 5s RPC timers) *would* have fired and let `describe` return — and it never returned. That leaves only an `execFileAsync` whose OS-level kill failed to settle the promise.
- **Server logs:** the hung process logged the POST, then nothing for 24 min until the kill — the exact signature of a stuck subprocess await with no error thrown.
- **Existing precedent:** `adb.ts` already fixes this identical class of bug for Android via `ADB_KILL_SIGNAL = "SIGKILL"` (see its comment: "Node's execFile default kill signal is SIGTERM, which an `adb` process blocked on a hung daemon can ignore — leaving the parent waiting past the deadline"). The same fix was simply never applied to the `simctl` path.
- **Deterministic repro:** `execFileAsync('sh', ['-c', 'trap "" TERM; …'], { timeout: 2000 })` never settles; adding `killSignal: "SIGKILL"` makes it reject at ~2006ms. Reproduces both the hang and the fix.

## The fix

- Add `SIMCTL_KILL_SIGNAL = "SIGKILL"` to `simctl-config.ts` and pass it alongside the timeout on every describe-path `simctl` invocation (`ax-prefs.ts`, `ios-devices.ts`, `ios-host.ts`), mirroring `ADB_KILL_SIGNAL`.
- Add the missing `timeout` + kill signal to `listRunningUIKitApplicationBundleIds`, which had **no timeout at all** (a latent forever-hang on the `native-devtools-status` path).
- Add `test/ios-simctl-hardening.test.ts` pinning every describe-path `xcrun simctl` call to `killSignal: SIGKILL` + a finite timeout.

## Verification

- `tsc --noEmit` clean (src + tests).
- Full tool-server suite: **2670/2670 pass**.
- Confirmed the new test is a real guard: stripping the flag makes it fail, restoring makes it pass.

## Scope note

This fixes the **proven** describe / native-devtools path. The same latent hang exists in other `simctl` call sites (`boot-device`, `launch-app`, `screenshot`, etc.) that use `execFileAsync` with the default signal — extending SIGKILL to the whole `simctl` surface is a reasonable follow-up but was kept out of this PR to stay scoped to the reported hang.

One caveat: if `simctl` is ever stuck in true uninterruptible kernel state, even SIGKILL can't reap it until the syscall returns; the fully robust version would detach the await from the child via an independent deadline. SIGKILL matches the `adb` precedent and handles the realistic CoreSimulator-wedge case.

### DIFF
--- a/packages/tool-server/src/utils/ax-prefs.ts
+++ b/packages/tool-server/src/utils/ax-prefs.ts
@@ -3,7 +3,7 @@ import * as os from "node:os";
 import * as path from "node:path";
 import { execFile } from "node:child_process";
 import { promisify } from "node:util";
-import { SIMCTL_SPAWN_TIMEOUT_MS } from "./simctl-config";
+import { SIMCTL_KILL_SIGNAL, SIMCTL_SPAWN_TIMEOUT_MS } from "./simctl-config";
 
 const execFileAsync = promisify(execFile);
 
@@ -21,7 +21,7 @@ export async function ensureAutomationEnabled(udid: string): Promise<void> {
       "-bool",
       "true",
     ],
-    { timeout: SIMCTL_SPAWN_TIMEOUT_MS }
+    { timeout: SIMCTL_SPAWN_TIMEOUT_MS, killSignal: SIMCTL_KILL_SIGNAL }
   );
 }
 
@@ -48,7 +48,7 @@ export async function isEntitlementBypassActive(udid: string): Promise<boolean> 
       "com.apple.Accessibility",
       "IgnoreAXServerEntitlements",
     ],
-    { timeout: SIMCTL_SPAWN_TIMEOUT_MS }
+    { timeout: SIMCTL_SPAWN_TIMEOUT_MS, killSignal: SIMCTL_KILL_SIGNAL }
   )
     .then(({ stdout }) => stdout.trim() === "1")
     .catch(() => false);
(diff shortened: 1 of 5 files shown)


filip131311 (approved):

filip131311 (changes requested):
comment

latekvo:
In-person concern about appropriateness of these timeouts researched and resolved. None of these calls are supposed to last more than a second, a couple at most, 10s is very generous. 
There is zero need for these terminations to be SIGTERM - if any of them take more than 10s then its already too late for SIGTERM and SIGKILL is required usually.
<sub>a human wrote this message, i just like to be formal in github or something idk</sub>

Issue #330, opened 2026-06-12
Argent iOS native profiler export failing


miguelartazos:
Hi Software Mansion team,

I’m using Argent 0.11.0 with Xcode 26.5 / iOS 26.5 Simulator.

native-profiler-start works only if I pass:
{ "app_process": "Myappname+" }

Auto-detection seems to pick CFBundleExecutable = Guiote, but xctrace only attaches by display name.

After start, native-profiler-stop says:
No active native profiling session found

The trace contains:
event_data_<pid>.oa

but export fails:
Export failed: Document Missing Template Error

Any known workaround? Should Argent retry CFBundleDisplayName, or is this likely an Xcode 26.5 xctrace issue?
Thanks.

benjaminharringtonrose:
Same issue for me ^

filip131311:
Hello, @miguelartazos and @benjaminharringtonrose i have merged #333 that will help on this specific issue, but I am afraid it is not the end of the problems with profiling on `Xcode 26.5`. Would you mind running a little test for me? 

could you run: 
```
xctrace record \
  --template "your_template" \
  --device <UDID> --attach <PID> \
  --output /tmp/test.trace --no-prompt
```

and then try to stop it (by `CRL+C`)?

miguelartazos:
Hi @filip131311, I ran the manual test on Xcode 26.5 build 17F42 / xctrace 16.0 (17F42), iOS 26.5 Simulator.

Command:
xcrun xctrace record \
  --template "Time Profiler" \
  --device 9D3A2BB6-8FE6-438E-81B9-71C35F13B3E6 \
  --attach 15550 \
  --output /tmp/argent-xctrace-manual-test.trace \
  --no-prompt

It started and attached successfully:
Starting recording with the Time Profiler template. Attaching to: Guiñote+ (15550).

After Ctrl+C, xctrace hung for >90s during stop/packaging. I terminated it. The trace only contained:
Trace1.run/RunIssues.storedata

Then:
xcrun xctrace export --input /tmp/argent-xctrace-manual-test.trace --toc

returned:
Export failed: Document Missing Template Error

So this part appears reproducible outside Argent on Xcode 26.5.

filip131311:
Hey @benjaminharringtonrose @miguelartazos 

I believe i found a workaround for `xctrace` not working, the experience might be minimally degraded because I need to manually filter additional data but the newest version (0.12.1) should work for you. 

thank you so much for supporting us <3

Issue #250, opened 2026-05-22
native-profiler-stop reports 'No active native profiling session found' immediately after native-profiler-start (v0.8.0)


royatluscii:
## Summary

`native-profiler-stop` consistently fails with `[Tool:native-profiler-stop] No active native profiling session found. Call native-profiler-start first.` even when called shortly after a successful `native-profiler-start` against the same `device_id`. The recording session is not being tracked between the two MCP tool calls.

A secondary issue: the bundled default `Argent.tracetemplate` uses `rsrc://templateProcessorTrace` and produces trace schemas (kdebug / processor-trace) that don't match the schemas `native-profiler-analyze` parses for CPU and hangs — so even when a session does survive, the default template returns an empty report.

## Environment

- Argent: **0.8.0**
- Host: macOS (Apple Silicon)
- Xcode: **26.4.1** (build 17E202), `xctrace version 16.0`
- Target: iOS 17.5 Simulator, iPhone 15 Pro (UDID `A41F04F1-4BC3-4903-BFF4-C7746FA13535`)
- Target app: `com.omronhealthcare.omronconnect`

## Steps to reproduce

1. Boot an iOS simulator and launch any app on it via `launch-app`.
2. Call `native-profiler-start` with the device's UDID. Tool returns:
   ```json
   { "status": "recording", "pid": 51559, "traceFile": "/var/folders/.../native-profiler-20260522-134212.trace" }
   ```
3. Drive ~30s of interaction (gesture-tap, gesture-swipe, etc.).
4. Call `native-profiler-stop` with the same `device_id`.

### Expected

`native-profiler-stop` exports CPU / hangs / leaks XML and `native-profiler-analyze` produces a populated report.

### Actual

`native-profiler-stop` returns:

```
[Tool:native-profiler-stop] No active native profiling session found. Call native-profiler-start first.
```

The trace bundle exists at the path returned by `native-profiler-start`, but is incomplete — some runs contain an `event_data_*.oa`, others contain only `Trace1.run/RunIssues.storedata` (no event data captured at all).

## Reproduction matrix

Reproduced with:

- The bundled default template (`dist/Argent.tracetemplate` → `rsrc://templateProcessorTrace`).
- A custom template passed via `template_path`, based on Apple's stock `Time Profiler.tracetemplate` (which includes `coresampler` + `com.apple.dt-perfteam.hangs`).
[...]

Issue #248, opened 2026-05-22
native-profiler-stop can't find the session created by native-profiler-start (iOS sim)


csori99:
Hey, ran into this while profiling a React Native app on the iOS simulator and wanted to flag it.

`native-profiler-start` works fine (xctrace spawns, starts recording, returns a pid and a trace path). But `native-profiler-stop` always fails with `No active native profiling session found. Call native-profiler-start first.`, so I can never export a trace through the tool.

Dug into it a bit:

- After calling start, the xctrace process stays alive. I watched it from a plain shell (no argent tool calls in between) for 6+ seconds and it kept running and recording.
- Calling stop after that still says "no active session", and the xctrace process gets killed right around then.
- So start and stop don't seem to share the same session state. stop looks like it resolves a fresh/empty session (profilingActive is false) instead of the one start created, and the real session gets disposed (dispose() SIGKILLs the live xctrace).

Only one `tool-server.cjs` and one `argent mcp` process running, so it's not a duplicate-instance thing. Happens whether or not there are other tool calls (gesture-tap, screenshot) in between.

Repro:
1. Boot an iOS simulator with an app running.
2. `native-profiler-start` (succeeds, recording).
3. interact or just wait a few seconds.
4. `native-profiler-stop` -> "No active native profiling session found".

Expected: stop finds the active session and exports the trace.
Actual: stop never finds it, and the recording gets killed.

Environment:
- argent 0.8.0 (also reproduced on 0.7.1)
- macOS, Xcode 26.4, xctrace 16.0
- iOS 26.4 simulator, React Native app (bridgeless, Hermes)

One side note in case it is related: running xctrace directly with the bundled Argent.tracetemplate records ok but won't finalize/package the trace on stop (it hangs, and the bundle is incomplete so export fails with "Document Missing Template Error"). Not sure if that is the same underlying issue or a separate xctrace/simulator thing on this machine, but the start/stop session mismatch above reproduces on its own.

The react profiler tools work great by the way, this is just the native side.

csori99:
Did some more digging in the 0.8.0 source in case it helps narrow it down.

The native-profiler session blueprint never emits a `terminated` event, while react-profiler-session emits one on CDP disconnect. The native session also doesn't declare dependencies, so it shouldn't be torn down by a cascade either. And the registry caches RUNNING instances by URN, so start's session should still be cached and RUNNING when stop resolves the same urn (`NativeProfilerSession:<deviceId>`).

[...]

PR #529, opened 2026-07-20
fix(ios-profiler): stop native-profiler-stop being aborted mid-packaging

L65FREAD:
## Problem

`native-profiler-stop` fails with `No active native profiling session found. Call native-profiler-start first.` while xctrace is still recording — #248, #250, #330. Those were closed on the 0.12.1 `--all-processes` workaround (#380), which fixed trace finalization but not this.

The tool isn't declared `longRunning`, so the MCP adapter applies its per-request timeout:

```js
fetchTimeoutMs: meta?.longRunning ? null : FETCH_TIMEOUT_MS,   // mcp-server.ts:222
```

A stop that runs past 30s is aborted and retried (`MAX_RETRIES = 4`). The retry runs against the session state the first call has already cleared, hits the `!api.profilingActive` branch and throws. The message points at session tracking; the actual failure is the timeout.

Every failed stop I have logs for lands at 30–34s: `durationMs: 33639` in [#330](https://github.com/software-mansion/argent/issues/330), and 30276 / 30605 / 31019 / 31914 locally.

Stop can't fit in 30s on Xcode 26.4+. Since #380 the capture is host-wide rather than app-scoped, so the bundle is much larger — ~55MB per 10s of recording here, 385MB for 90s — and xctrace needs ~22s to package even the 10s case before the four export passes start.

`STOP_GRACE_MS` has the same problem from the other side: at 30s xctrace is SIGTERM'd mid-package, and the bundle it leaves behind fails to export.

## Fix

Declare the tool `longRunning`, and raise `STOP_GRACE_MS` to 300s. The grace is a ceiling rather than a fixed wait — a fast package still returns as soon as xctrace exits — so this only changes captures that were previously killed.

With both applied a 90s capture stops and exports cleanly in ~2–4 min, and a 40s capture returns inline.

## Note

The abort-and-retry sequence is inferred rather than captured directly. What's observed: `MAX_RETRIES = 4`, the uniform ~30s durations, and `tool-server.log` showing the `did not respond to SIGINT` warning — written after the throw site — logged 184ms *before* the failure was returned. One execution can't produce that ordering.

## Testing

- `npm run build` and eslint clean on changed files
- `npm test -w @argent/tool-server` — 2727 tests / 273 files pass
- Verified on Xcode 26.5 / iOS 26.5 simulator against a bridgeless RN app: start → interact → stop → analyze now returns a populated report with `errors: {}`

### DIFF
--- a/packages/tool-server/src/tools/profiler/native-profiler/native-profiler-stop.ts
+++ b/packages/tool-server/src/tools/profiler/native-profiler/native-profiler-stop.ts
@@ -79,6 +79,8 @@ function registerTrace(store: ArtifactStore, traceFile: string): Promise<Artifac
 export const nativeProfilerStopTool: ToolDefinition<z.infer<typeof zodSchema>, StopResult> = {
   id: "native-profiler-stop",
   capability,
+  // Packaging plus the export passes routinely exceed the 30s fetch timeout.
+  longRunning: true,
   description: `Stop native profiling and export trace data.
 iOS: sends SIGINT to xctrace, waits for packaging, then exports CPU, hangs, and leaks XML.
 Android: sends SIGTERM to the perfetto daemon, polls /proc/<pid>, then \`adb pull\`s the .pftrace.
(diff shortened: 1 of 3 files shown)


filip131311 (approved):
thank you

PR #530, opened 2026-07-20
feat(map): `argent map` — automated screen-graph crawler + two-tab preview window

latekvo:
## What

`argent map <bundleId>` walks a running app on a simulator or emulator, taps its way through every reachable screen, and builds a directed graph of the app's navigation. The graph renders live in the Argent preview (Electron) window, which now has two tabs: **Lens** (the existing UI, unchanged) and **Map**.

It's behind a new `argent-map` feature flag (`argent enable argent-map`), off by default while it matures - the same pattern as `argent-lens`.

```
argent map xyz.blueskyweb.app --udid <id> --max-screens 20 --budget 300 \
  --deep-link bluesky://notifications --deep-link bluesky://search --json=graph.json
```

- **Crawls** the app: describe the screen, enumerate tappable elements, tap, detect the new screen, record a node and an edge, recurse. It backtracks with hardware/nav back where it can and relaunch-and-replay where it can't.
- **Fingerprints screens structurally** (roles, identifiers, coarse frames; dynamic text excluded) so revisits collapse into one node and the graph closes.
- **Streams progress** to the CLI and renders the graph live in the Map tab: screenshot-thumbnail nodes, bezier edges with action labels, pan/zoom, a node-detail overlay and a status strip. `--json[=path]` exports the full graph.

## An app is a graph, not a tree

The obvious layout is a tree, with nodes in columns by tap-distance from the launch screen. That's wrong, because depth is a traversal artifact rather than a property of a screen: a screen is reachable by many paths of different lengths, and apps have direct jumps and deep links between arbitrary parts. So this lays it out as a real graph.

- **`MapScreenNode` carries no depth.** Layout is a deterministic, stable, incremental force-directed placement - a string-hash PRNG seeded per node id, never `Math.random`, with nodes freezing after their structural revision so the skeleton is pinned and only fresh nodes move. There is no live per-frame physics or jitter. Depth survives only as a live "how many taps deep the crawl is right now" figure in the CLI progress line, and as the `--max-depth` crawl bound.
- **Multiple entry points.** `MapCrawlState` replaces the single `rootId` with `entryPoints[]`; the launch screen and every deep-link-reached screen are marked `entry: true` and ringed in the UI.
- **Deep-link seeding** via repeatable `--deep-link <url>`. After the launch crawl, each link is opened and fingerprinted: a new screen becomes a fresh entry the crawl explores, a known screen is just flagged as an entry (a deep link is a jump, not a tap, so there's no fake edge from the root), and a link that fails or bounces to Safari is skipped.

## Structure

- **Crawler** - `packages/tool-server/src/tools/map/` (fingerprint, actions, crawler, driver, stable-tree), exposed as the long-running `map-app` tool, plus a `mapSessionStore` singleton and `/preview/map` + screenshot routes.
- **CLI** - `packages/argent-cli/src/map.ts`, wired into the dispatcher with the `argent-map` flag in `FLAG_REGISTRY`.
- **UI** - additive Lens/Map tab strip and force-directed graph view in `packages/ui/index.html` and `theme.css`. Lens behavior is untouched.

## Things that only showed up under real crawling

- **iOS accessibility trees oscillate on a settled screen** - content nodes drop out and come back seconds apart - so a single describe would flip a screen's identity. The driver samples several times and keeps the fullest snapshot. Two details keep "fullest" honest: the early exit only fires once it has watched the tree climb to a peak and hold there (so a leading sparse phase can't lock the capture in before the fuller phase is sampled), and fullness is measured over the same nodes the fingerprint keys on (so a fading scroll indicator, counted by a raw walk but excluded from the key, can't bias the pick and split one screen across visits).
- **Leaving the app is detected whenever the device can answer "is my app frontmost".** iOS `describe` reads whichever app is frontmost, so a tap that opened Safari, Settings or the share sheet returned that app's tree looking perfectly in-app, and the crawler would map and descend into it. `driver.isTargetForeground()` (native-devtools app state on iOS, resumed-activity on Android) is the authoritative signal: a confident "not foreground" is recorded as leaving the app and triggers a relaunch, before the Android resource-id heuristic can seed a foreign package; a confident "foreground" is trusted even when the screen's only qualified ids belong to a third-party SDK (an embedded player, a maps view) that the resource-id fallback would misread. The heuristic only runs when foreground state is unknown. Known gap: on iOS that state is unanswerable for an app that never connected to native-devtools (a non-RN app, a simulator not booted through argent, or before the first relaunch under the injected env), and the resource-id fallback is Android-only — so a non-RN iOS app can still have a tap into Safari/Settings mapped as one of its own screens. What is now closed is the worst case of that: a mistyped or uninstalled bundle id, where the launch fails and describe returns whatever app was already frontmost. The crawl refuses to start unless the launch succeeded or the device positively confirms the target is on screen, so it can no longer map a stranger's app and report `completed`.
- The force layout can place nodes at **negative world coordinates**, so the edge SVG viewBox now spans the full `[minX..maxX]×[minY..maxY]` extent and arcs into those nodes aren't clipped at the world origin. Cards keep their exact positions.
- **Disconnected entry points spread out.** Deep-link entries arrive one per structural revision, so the neighbourless-seed counter now persists across revisions and each takes the next golden-angle slot instead of piling into one wedge and overlapping permanently.
- **Screen cards never occlude each other.** The cards are opaque, but soft range-capped repulsion couldn't guarantee clearance on dense graphs, and streaming the graph one node per revision let two edge-connected cards freeze while still stacked. The layout now enforces a tokenized minimum separation: an in-loop collision force plus a fixed-order Gauss-Seidel de-overlap pass that resolves every penetrating pair, frozen ones included, along the axis of least penetration. The budget scales with node count, so a cold single-shot render of a ~100-node completed graph fully un-stacks, while the loop early-breaks on convergence so the common already-separated case costs one sweep. Convergence is detected against a sub-pixel tolerance rather than exact zero: resolving a pair drives it to *exact* contact, where float residue leaves the penetration marginally positive, so an exact test re-flags every already-resolved pair and the break can never fire. Measured re-solving a settled layout at 200 nodes: 1200 of 1200 passes without the tolerance, 1 with it, and 0 card-footprint intersections either way. It stays a deterministic per-revision solve, so an unchanged graph is not re-solved at all and never jitters. The determinism is per-*sequence*, not per-graph: freeze grouping and the seed anchor centroid depend on the order revisions arrived, so a cold render of a finished graph lands on a different (equally valid, equally stable) arrangement than the one watched growing.
- The app is **restarted at crawl start**, so a leftover screen can't poison the root, and **back-navigation adopts** whatever known screen it lands on instead of paying a restart per sheet dismissal.
- **Reachability doesn't depend on discovery order.** Since depth is only a crawl budget, a screen reachable within it has to be mapped however it's found. A hub discovered the long way first and then reached by a shorter tap path now re-propagates that shorter depth down the already-explored graph and revives any descendant that was capped at the budget but now fits.
- Transient device errors **degrade the affected step** rather than killing the run: the crawl always returns a partial map, and a crawl that fails after mapping screens still writes the requested `--json` partial graph, matching the cancel and success paths.
- The Map tab **disables Lens's key forwarding** so keystrokes can't leak into the crawled app. Device-sourced titles and labels are control-byte-sanitized before reaching the terminal, both on stdout and in the `--json` artifact (`JSON.stringify` escapes only C0 and leaves the DEL/C1 escape-sequence introducers literal). `--deep-link` and the numeric flags are capped client-side, so an over-cap value fails with a one-line message instead of a ZodError blob.

## Verification

- Root `tsc --build` and `tsc --noEmit -p tool-server/tsconfig.test.json`: clean.
- Tests: tool-server and CLI suites plus configuration-core and tools-client, all passing. New tests cover the crawler, store, preview routes, CLI parser, stable-tree sampler, and the crawl-side fixes above.
- **Not covered by tests: `packages/ui/index.html`.** The force layout, the de-overlap pass, the viewBox extent and the key-forwarding gate have no automated coverage — there is no headless-browser harness in this repo. Those were verified by extracting the layout functions and re-deriving them in Node (numbers quoted below); adding a real harness is a follow-up.
- prettier and eslint: clean.
- **E2E on Bluesky** (`~/dev/social-app`, iOS 18.5 simulator), re-run against the final commit: `map-app xyz.blueskyweb.app` with two deep-link seeds produces a real graph of 17 screens and 3 entry points (`s0`, the launch feed, plus `bluesky://notifications` and `bluesky://search`), with no `depth` on any node. The pure deep-link entry has zero inbound edges, and a deep link landing on an already-tapped screen just flags it as an entry and keeps its real edge. Restart-replay backtracking ran 13 times and the crawl finished `completed` with no spurious "outside" nodes. The force-directed render is covered by headless-Chromium layout harnesses across streaming, completed, disconnected-entry and 100-node cold-load graphs.

Draft, since `argent map` is flag-gated and this is the feature's first cut.

---

## Review pass

Fixed in this branch after a full review of the diff. Each behavioural fix has a test that fails on the pre-fix code (verified by mutation).

**Action enumeration**
- `MIN_TAP_AREA` was an *area* floor (`w*h < 0.005`), which is the wrong shape for the thing it documented. A 44x44pt button — Apple's HIG minimum — is `0.1 x 0.046 = 0.0046` on an iPhone 16 Pro Max and was **dropped**, while a full-width 6pt divider (`1.0 x 0.006 = 0.006`) was **kept**. Now a per-axis minimum, so the two cases land the right way round.
- The destructive-action guard (`log out`/`sign out`/`delete`) only looked at `label` and `value`, so an icon-only control whose sole tell is `com.app:id/logout_button` was tapped. It now includes `identifier` — and splits identity separators first, because `_` is a word character so `\blogout\b` does not match `logout_button` on its own. `undelete_item` / `deleted_items` still stay tappable.
- `deriveMapSelector` skipped the `value` fallback that its twin `deriveSelector` documents, so an element whose only stable text is its value replayed by frame. Now mirrors the twin.
- `isTappableIosRole` matched `cell`, `menuitem`, `tab` and guarded against `table` — none of which the iOS describe adapters can emit (both derive role from `mapNativeTraitsToDescribeRole`, whose whole output set is `AXHeading`/`AXButton`/`AXTextField`/`AXLink`/`AXImage`/`AXStaticText`/`AXTabBar`/`AXAdjustable`/`AXGroup`). Four dead branches removed and the doc corrected; the real consequence — a trait-less list row arrives as `AXGroup` and is not enumerated — is now written down and pinned as a deliberate v1 trade-off. Tests that used `AXCell`/`AXTable` fixtures were pinning unreachable behaviour and now use emittable roles.

**Crawler**
- The deep-link re-root was strictly weaker than its own tap-path twin: gated on the landed screen's owed actions, and never calling `repropagateShorterDepth`. A deep link onto a fully-explored hub therefore left depth-capped descendants dead — exactly the deep screens deep links exist to reach. It now re-roots and re-propagates regardless of the hub's own state, matching the twin. (The "Known limitation" comment justified the omission by saying recovery "would need re-propagating depth through already-traversed edges" — that mechanism was already in the file.)
- Forward taps used the frame recorded when a screen was *first* enumerated, ignoring the selector stored alongside it. Actions are enumerated once but consumed across many later visits, and `screenKey` ignores labels while bucketing frames to 5%, so a reordered list re-keys identically — the tap then hit whatever now occupied that rectangle, recording an edge with one row's label and another row's destination. Forward taps now resolve the selector against the screen actually being stood on, falling back to the recorded frame when there is no current capture.
- `isTargetForeground` and `screenshot` were awaited unguarded, though the `CrawlDriver` contract says every method may reject — a rejection would have ended the whole crawl, which the crawl contract forbids. Both now degrade.

**Tool / contract**
- tvOS and Android TV targets passed the capability gate (a tvOS sim is platform `ios`, an Android TV emulator is platform `android`), then failed deep inside with a misleading "check that the bundle id is correct". Both are now rejected up front via `UnsupportedOperationError`, matching how every sibling screen-driving tool probes.
- `MapCrawlState`'s doc claimed the tool result was "this shape minus `screenshotPath`". It is a much smaller summary — nodes trimmed to 4 of 10 fields, `edges` collapsed to a count, seven top-level fields absent. Corrected.
- `MapSelector.label` was documented "(exact)" but resolves through the shared case-insensitive *substring* matcher against label or value. Corrected.
- `stats.restarts` counts attempted recovery navigations including deep-link re-opens, and one recovery can add more than one — documented.
- `maxActionsPerScreen` and `maxDepth` now state their caps like `maxScreens`/`timeBudgetS` already did; `maxDepth`'s description no longer says "from the start screen" (depth is measured from each screen's own entry); `deepLinks` rejects empty strings, which the CLI already did.

**UI**
- De-overlap convergence tolerance (measured above), so a streaming crawl stops spending the full O(N²)×6N budget on every poll.
- Switching tabs with a key held stranded it down on the device: the key gate drops the *keyup* too, so `releaseAllInput()` now runs on tab switch, as it already did on window blur.
- Escape on the Map tab also reached Lens's own Escape handler (whose guard only checks for a focused text field, which does not apply on Map), tearing down an in-progress comment. Now handled in the capture phase and confined to Map.
- `mapEdgePathD` produced `d="M NaN NaN ..."` for two nodes at identical centres: `|| 1` guards the division but leaves a zero unit vector, and `exit()` then returns `Infinity` on both axes, so the browser discarded the whole path and the edge silently vanished. Degenerate pairs now use a fixed direction.

**Docs**
- `argent map` added to the CLI Reference in both READMEs; the two `MAP_*` failure codes got the group comment every neighbouring group has.

**Known gaps left open** (deliberately, not oversights): no `argent-map` skill or `rules/argent.md` routing entry, so agents discover the tool only through the tool list; the Map tab renders for users with the flag off; budget/cap truncation is still indistinguishable from a clean finish in the durable outputs (`MapCrawlStatus` has no `truncated` member); `map/index.ts` and `map/driver.ts` still have no unit tests; and the `mapUrl` in the tool result is a hardcoded `127.0.0.1`, which is wrong for an MCP client talking to a remote tool-server over `argent link`.


latekvo:
Idea ditched for now. Might come back later.

PR #532, opened 2026-07-21
fix(tool-server): stop unix native-devtools listen from crashing the process

filip131311:
## Summary

Fixes the **real multi-user crash source** surfaced by the new 0.16.0 tool-server crash telemetry (PR #494).

Analyzing `toolserver:stop reason=crash` in PostHog (project Argent, 30d): 936 crashes / 48 users, 0 unclassified. Ranked by **users** (not events), the widespread bug is fingerprint **`dc063fe2`** — `EADDRINUSE` + `EEXIST` at `crash_phase=startup`, across ~14 users. (The 774-crash `a0517e9a` loop is a single machine tight-looping on an occupied HTTP port — high volume, one user — tracked separately.)

## Root cause

The native-devtools socket server's **unix branch** did a bare `server.listen(socketPath)` with **no `'error'` handler** — unlike the TCP branch immediately above it and unlike `ax-service.ts`'s `startListener`, both of which wrap listen in a promise with `server.once("error", reject)`.

So a bind failure — `EEXIST` from a stale `/tmp/argent-nd-<UDID>.sock` a crashed run left behind, or `EADDRINUSE` from a concurrent instance racing on the same per-UDID path (see the multi-agent socket-contention case) — fired an unhandled `'error'` EventEmitter event. Node rethrows that as an uncaught exception, taking down the **entire tool-server** (every device's service with it). That is exactly the `EADDRINUSE`/`EEXIST`-at-startup signature of `dc063fe2`.

## Fix

Wrap the unix listen in the same promise + error-handler pattern the TCP branch already uses, so a bind failure **rejects out of the factory**. The registry catches that and turns it into a retryable `ServiceInitializationError` (`packages/registry/src/registry.ts`) instead of a process crash. The existing pre-unlink (a few lines above) still clears the common stale-socket case, and the watcher retries on its next poll.

```diff
-      server.listen(socketPath);
+      await new Promise<void>((resolve, reject) => {
+        server.once("error", reject);
+        server.listen(socketPath, () => {
+          server.off("error", reject);
+          resolve();
+        });
+      });
```

## Testing

- **New regression test** (`native-devtools-factory-cleanup.test.ts`): occupies the per-UDID socket path with a directory so the bind deterministically fails, and asserts the factory **rejects** rather than emitting an unhandled `'error'`.
- Full tool-server suite: **2727 tests green**; typecheck + eslint clean.
- **Live A/B against the compiled `dist`** (the shipped surface; this crash path is pure JS, before any native binary is touched):
  - Buggy build → `uncaughtException EADDRINUSE` → **process exits 42** (crash reproduced).
  - Fixed build → **handled rejection, process survives, exit 0**.

## Scope note

This PR targets `dc063fe2` (the genuine multi-user defect). The single-machine `a0517e9a` restart loop on the tool-server's own HTTP port is environmental (a contested fixed port on one host) and would need port-selection/back-off changes to the main `app.listen` path; it is intentionally out of scope here.

### DIFF
--- a/packages/tool-server/src/blueprints/native-devtools.ts
+++ b/packages/tool-server/src/blueprints/native-devtools.ts
@@ -567,7 +567,19 @@ export const nativeDevtoolsBlueprint: ServiceBlueprint<NativeDevtoolsApi, Device
       // written — lands on our listener.
       await host.startProxy(udid, endpoint.port!);
     } else {
-      server.listen(socketPath);
+      // Mirror the TCP branch above (and ax-service's startListener): attach an
+      // error handler so a bind failure — EEXIST from a stale socket a crashed
+      // run left behind (the pre-unlink above covers the common case) or
+      // EADDRINUSE from a concurrent instance racing on the same per-UDID path —
+      // rejects into the watcher-retry path instead of surfacing as an unhandled
+      // 'error' event that crashes the entire tool-server.
+      await new Promise<void>((resolve, reject) => {
+        server.once("error", reject);
+        server.listen(socketPath, () => {
+          server.off("error", reject);
+          resolve();
+        });
+      });
     }
 
     // Tolerate ensureEnv failure: throwing here would leak `server` — the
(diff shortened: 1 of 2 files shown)


filip131311:
Closing as superseded by #531, which landed the same fix (native-devtools unix socket bind had no `error` handler → unhandled 'error' event → tool-server crash). #531 is a strict superset of this branch: it additionally self-heals with an unlink+retry-once on EADDRINUSE/EEXIST and rejects with a coded `NATIVE_DEVTOOLS_SOCKET_BIND_FAILED` FailureError for clean telemetry classification. Nothing here adds coverage #531 doesn't already have. Verified against 0.16.0 crash telemetry: #531 covers the entire 51-user startup-socket-bind crash class.

PR #533, opened 2026-07-21
fix(tool-server): defer to an existing server on EADDRINUSE instead of crash-looping

filip131311:
## Summary

Fixes **bucket B** of the 0.16.0 crash telemetry — the `toolserver_bind` / `EADDRINUSE` restart loop. It's the largest crash **volume** (~817 events, one machine alone = 785) even though it's a narrow user set (10). Companion to #531 (which fixed the widespread multi-user bucket A).

## Root cause (reproduced live)

The main HTTP listener's `server.on("error")` handler routed **every** bind error through `crashShutdown` (exit 1). When a second tool-server starts on an already-owned port, that exit just feeds a supervisor respawn → re-bind → crash, forever.

The telemetry signature was puzzling: all these crashes are `crash_phase=serving` with ~1s uptime and **0 tool calls** — i.e. the listener *bound successfully* and then died. I reproduced it: on macOS, a **second tool-server process's** `listen()` momentarily **succeeds** (the `listening` callback runs → `listening=true` → phase `serving`), and **then** the server emits an asynchronous `EADDRINUSE`. So the redundant instance binds, immediately errors, crashes, and the supervisor loops it (~1s each, 785× on one machine).

Live A/B on the compiled `dist` (isolated port, telemetry disabled), server A up then a second server B on the same port:

```
buggy (current 0.16.0):  B: "Tools server listening ..."  → "Failed to bind ... EADDRINUSE"  → exit 1  (→ respawn loop)
fixed:                   B: "Tools server listening ..."  → "deferring to it (redundant instance)" → exit 0
```

A survives in both.

## Fix

Handle `EADDRINUSE` specifically instead of crash-looping:

- Probe the port for a healthy argent tool-server (`GET /tools` → 200, or 401 when a token is configured).
- **If a healthy peer answers**, this instance is redundant → shut down **cleanly** (`shutdown(0)`) and defer to it. This removes it from the `reason:"crash"` restart-loop population entirely.
- **If nothing healthy owns the port** (a foreign holder, or a wedged peer that doesn't answer within the probe timeout) → still crash, so a genuinely stuck port stays visible.
- `EACCES` and all other server errors crash exactly as before.

It also corrects the now-false comment claiming these crashes are phased `startup` ("the listen callback never ran"). Telemetry and the live repro prove the bind succeeds first, so `listening=true` and the phase is genuinely `serving`.

## Testing

- **New unit tests** for `probeArgentToolServer`: peer answering 200 or 401 → `true`; foreign 500, connection-refused, and wedged/timeout → `false`.
- Full tool-server suite **2731 green**; `tsc` + `eslint` clean.
- **Live A/B** above, on the built dist.

## Note on scope

This is a robustness fix that breaks the loop for the redundant-instance case (the dominant one). The precise reason macOS lets a second process's bind succeed-then-error asynchronously (vs. failing immediately, as same-process contention does) is a kernel/loopback timing detail; the fix does not depend on it — any `EADDRINUSE`, immediate or deferred, now defers-or-crashes deterministically rather than hot-looping.

### DIFF
--- a/packages/tool-server/src/index.ts
+++ b/packages/tool-server/src/index.ts
@@ -1,4 +1,5 @@
 import * as path from "node:path";
+import * as http from "node:http";
 import { homedir } from "node:os";
 import { isFlagEnabled } from "@argent/configuration-core";
 import { FAILURE_CODES, attachRegistryLogger, type FailureSignal } from "@argent/registry";
@@ -85,6 +86,34 @@ function initializeStdioTimestampWrapper(): void {
     stream.write = ((chunk: string | Uint8Array, ...rest: unknown[]) =>
       orig(`[${new Date().toISOString()}] ${chunk}`, ...(rest as []))) as typeof stream.write;
   }
+}
+
+/**
+ * Probe whether a healthy argent tool-server is already listening on
+ * `host:port`. Used to tell a redundant/overlapping instance (which lost the
+ * bind with EADDRINUSE) that the port is owned by a live peer, so it can defer
+ * cleanly instead of crash-looping.
+ *
+ * An argent tool-server answers `GET /tools` with 200 (the tools JSON) or, when
+ * an auth token is configured, 401 — either proves a live argent peer. No token
+ * is sent: reachability + argent-shaped response is all we need. A wedged peer
+ * that never answers within the timeout resolves `false`, so we still surface a
+ * genuinely stuck port as a crash rather than deferring to a dead server.
+ */
+export function probeArgentToolServer(host: string, port: number, timeoutMs = 500): Promise<boolean> {
+  return new Promise((resolve) => {
+    const req = http.request({ host, port, path: "/tools", method: "GET", timeout: timeoutMs }, (res) => {
+      const isArgentPeer = res.statusCode === 200 || res.statusCode === 401;
+      res.resume(); // drain the response so the socket can close
+      resolve(isArgentPeer);
+    });
+    req.on("error", () => resolve(false)); // connection refused / reset / non-HTTP
+    req.on("timeout", () => {
+      req.destroy();
+      resolve(false);
+    });
+    req.end();
+  });
 }
 
 export function start(): void {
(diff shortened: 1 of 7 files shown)


packages/tool-server/src/index.ts
```diff
@@ -480,39 +516,65 @@ export function start(): void {
...
-            failure_area: "tool_server",
-            error_kind: "crash",
-          },
-          err
-        );
+        const crashBind = () =>
+          crashShutdown(`Failed to bind ${HOST}:${PORT}`, `${code}${err.message}`, bindSignal, err);
+
+        if (err.code === "EADDRINUSE") {
+          void probeArgentToolServer(HOST, PORT).then((isArgentPeer) => {
```
latekvo, line 526: This issues a live network request to `HOST:PORT` from inside the bind-error handler. `test/bind-failure-telemetry.test.ts` (unchanged by this PR) emits an EADDRINUSE for the default port 3001 and does not mock `node:http`, so the probe now fires a real `GET http://127.0.0.1:3001/tools` during that test. When anything argent-shaped answers on 3001 — which is exactly the port a developer's own tool-server binds by default — the probe returns `true`, the handler defers instead of crashing, and that test's `reason:"crash"` / `exit(1)` assertions fail. Reproduced locally: with 3001 free the test passes; with a 200-returning stub on 3001 it fails, the assertion flipping to `reason:"signal"`. The test outcome (and the ~500ms real probe it now performs) depends on whatever happens to be listening on 3001 when the suite runs.

packages/tool-server/src/index.ts
```diff
@@ -480,39 +516,65 @@ export function start(): void {
...
+            );
+            eventLog?.info({
+              type: "tool_server.deferred_to_existing",
+              msg: `Deferred to an existing tool-server on ${HOST}:${PORT}.`,
+              host: HOST,
+              port: PORT,
+            });
+            // Clean stop (not a crash): removes this redundant instance from the
+            // reason:"crash" restart-loop population.
+            void shutdown?.(0);
```
latekvo, line 546: On this defer path `shutdownReason` is never assigned, so it keeps its default `"signal"`, and the resulting `toolserver:stop` is recorded with `reason:"signal"` — the same value used for an ordinary SIGINT/SIGTERM shutdown. A redundant instance that defers is therefore indistinguishable, in the stop-reason telemetry, from a normal clean shutdown. If an external supervisor keeps relaunching on any exit, the resulting ~1s-uptime, 0-tool-call deferral loop surfaces as benign `reason:"signal"` churn rather than being identifiable as the redundant-instance population this change is meant to isolate out of the crash bucket.

packages/tool-server/src/index.ts
```diff
@@ -480,39 +516,65 @@ export function start(): void {
...
+      // owns the port (a foreign holder, or a wedged peer) do we still crash, so
+      // a genuinely stuck port stays visible.
+      const bindSignal: FailureSignal = {
+        error_code: FAILURE_CODES.ARGENT_UNCLASSIFIED_FAILURE,
+        failure_stage: "toolserver_bind",
+        failure_area: "tool_server",
+        error_kind: "crash",
+      };
       server.on("error", (err: NodeJS.ErrnoException) => {
         eventLog?.error({
```
latekvo: This `tool_server.bind_failed` record (tagged `error_kind:"crash"`) is emitted unconditionally at the top of the handler, before the EADDRINUSE branch decides whether to defer. On the clean-defer path the event log therefore contains a crash-tagged `bind_failed` immediately followed by `deferred_to_existing` and a clean exit 0. Any consumer counting `bind_failed` / `error_kind:"crash"` event-log entries still counts every clean deferral as a bind crash, which contradicts the actual outcome for that instance.

packages/tool-server/src/index.ts
```diff
@@ -480,39 +516,65 @@ export function start(): void {
...
-            failure_stage: "toolserver_bind",
-            failure_area: "tool_server",
-            error_kind: "crash",
-          },
-          err
-        );
+        const crashBind = () =>
+          crashShutdown(`Failed to bind ${HOST}:${PORT}`, `${code}${err.message}`, bindSignal, err);
+
+        if (err.code === "EADDRINUSE") {
```
latekvo, line 525: The EADDRINUSE → probe → `shutdown(0)` defer branch introduced here — the behavior this PR exists to add — has no test coverage. The new `probe-argent-tool-server.test.ts` exercises only the pure probe function; nothing asserts the handler's success path (probe returns true → `deferred_to_existing` logged → `shutdown(0)` → exit 0 → no `reason:"crash"` stop). The existing `bind-failure-telemetry.test.ts` reaches only the crash leg, and only incidentally (its un-mocked probe to a dead 3001 is refused). A regression on the defer path would pass with the whole suite green.

PR #536, opened 2026-07-21
fix: get ax-service working on ios-remote (sim-remote --json spawn + TCP artifact diagnostics)

jwajgelt:
## Summary

Three fixes that together get `describe` (and the other ax-service / native-devtools tools) working against **ios-remote** (`sim-remote`) devices — including from a **Linux** host. Reproduced end-to-end against a live remote iPhone 17 simulator over the QUIC relay.

Before: `argent run describe --udid remote:<UDID>` returned an empty `ROOT AXGroup` with a misleading "reboot the simulator" hint.

### 1. `sim-remote spawn` output was never valid JSON (`fix(tool-server)`)

`simctlSpawn` parses the CLI's stdout as one `{exit_code,stdout,stderr}` (or `{pid}`) JSON object, but `sim-remote spawn` only emits that when passed `--json`. Without it a **non-detached** spawn streams the child's raw output live, so `JSON.parse` throws `Unexpected end of JSON input`.

This broke ios-remote describe at the first step: `remoteIosHost.bootstrapAx` runs `defaults write com.apple.Accessibility …` through a non-detached `simctlSpawn`, which failed before the ax-service daemon ever started — surfacing only as an empty AX tree. Fix: pass `--json` unconditionally (no-op for the already-JSON `--detach` path).

### 2. The tcp ax-service was resolved/packaged under a host-platform dir (`fix(native-devtools-ios)`)

The tcp `ax-service` is a **darwin/iOS-simulator** binary that is uploaded to and `simctl spawn`d on the **remote** macOS orchestrator, so it's needed on any *host* platform. But it was resolved (and packaged) under `bin/<hostPlatform>/tcp` via `platformBinDir()`, so a **Linux host looked at `bin/linux/tcp` and never found the darwin-built binary** — the ios-remote path was effectively macOS-host-only. (The tcp dylibs were already neutral at `dylibs/tcp`.)

Moved it to a platform-neutral `bin/tcp`, mirroring `dylibs/tcp`, and rewired every producer/consumer:
- resolver: `platformTcpBinDir()` → `tcpBinDir()` = `BIN_DIR/tcp`
- `native-devtools-ios/scripts/build.sh`: `bin/darwin/tcp` → `bin/tcp`
- `scripts/download-native-binaries.sh`: `TCP_BIN_DIR` → `bin/tcp`
- `packages/argent/scripts/bundle-tools.cjs`: copy from/to `bin/tcp`
- tests assert the neutral location on both linux and darwin

### 3. Missing tcp artifacts were undiagnosable (`fix(native-devtools-ios)`)

When a tcp artifact is absent (local/dev build, older package), the bare `... not found` error was thrown deep in the ax-service factory, where `describeIos` swallowed it and returned an empty tree plus `DEGRADED_HINT` — telling the agent to `boot-device --force`, a dead end. The only escape was discovering `ARGENT_SIMULATOR_SERVER_TCP_DIR` by reading source. Fix: `requireTcpArtifact` throws an actionable error naming the exact path, the override env var, and how to build it; `describeIos` surfaces that instead of the reboot hint.

## Testing

- New `sim-remote-spawn-json.test.ts` — asserts `--json` on both plain and `--bin/--detach` argv, before the `--` separator.
- `native-devtools-ios/test/index.test.ts` — the tcp binary resolves from the neutral `bin/tcp` on **both** linux and darwin; the not-found errors name their override env var.
- `native-devtools-ios` (16) + ios/describe tool-server suites pass; typecheck, eslint, prettier clean.
- Manual end-to-end against `remote:<UDID>` (darwin host): binary at `bin/tcp` → full tree from `source: "ax-service"`, no hint; binary removed → actionable `ARGENT_SIMULATOR_SERVER_TCP_DIR` hint.

### DIFF
--- a/packages/argent/scripts/bundle-tools.cjs
+++ b/packages/argent/scripts/bundle-tools.cjs
@@ -100,10 +100,13 @@ const ESM_REQUIRE_BANNER = {
 // only lives under darwin/.
 const BIN_SRC_ROOT = path.resolve(WORKSPACE_ROOT, "packages/native-devtools-ios/bin");
 const AX_BIN_SRC = path.resolve(BIN_SRC_ROOT, "darwin/ax-service");
-const AX_TCP_BIN_SRC = path.resolve(BIN_SRC_ROOT, "darwin/tcp/ax-service");
+// The tcp ax-service lives in a platform-NEUTRAL bin/tcp/ (not bin/darwin/tcp/):
+// it is uploaded to the remote macOS orchestrator, so it must resolve from any
+// host platform. Matches tcpBinDir() in native-devtools-ios/src/index.ts.
+const AX_TCP_BIN_SRC = path.resolve(BIN_SRC_ROOT, "tcp/ax-service");
 const BIN_DIR = path.resolve(__dirname, "../bin");
 const AX_BIN_DEST = path.resolve(BIN_DIR, "darwin/ax-service");
-const AX_TCP_BIN_DEST = path.resolve(BIN_DIR, "darwin/tcp/ax-service");
+const AX_TCP_BIN_DEST = path.resolve(BIN_DIR, "tcp/ax-service");
 // tvOS control binaries. Both are macOS-only: tvos-ax-service runs inside an
 // appletvsimulator, tvos-hid-daemon runs on the host. Unix-socket only.
 const TVOS_AX_BIN_SRC = path.resolve(BIN_SRC_ROOT, "darwin/tvos-ax-service");
--- a/packages/native-devtools-ios/scripts/build.sh
+++ b/packages/native-devtools-ios/scripts/build.sh
@@ -54,7 +54,10 @@ SRC_DIR="${SUBMODULE_DIR}/Sources/NativeDevtoolsIos"
 
 if [[ "$TRANSPORT" == "tcp" ]]; then
   DEST_DIR="${ROOT_DIR}/dylibs/tcp"
-  BIN_DIR="${ROOT_DIR}/bin/darwin/tcp"
+  # Platform-neutral: the tcp ax-service is an iOS-sim binary uploaded to the
+  # remote macOS orchestrator, so it must be resolvable from any host platform
+  # (not nested under bin/darwin/). Matches tcpBinDir() in src/index.ts.
+  BIN_DIR="${ROOT_DIR}/bin/tcp"
 else
   DEST_DIR="${ROOT_DIR}/dylibs"
   BIN_DIR="${ROOT_DIR}/bin/darwin"
--- a/packages/native-devtools-ios/src/index.ts
+++ b/packages/native-devtools-ios/src/index.ts
@@ -36,6 +36,29 @@ function requireDylibIn(dir: string, name: string): string {
   return p;
 }
 
+// The TCP-transport artifacts (`*Tcp` accessors) are only needed for ios-remote
+// (sim-remote) devices, and are built/downloaded separately from the default
+// unix-socket set — so they can be legitimately absent in a local/dev build or
+// an older package. When one is missing the caller (spawnAxDaemonRemote /
+// native-devtools setup) throws deep inside the ax-service factory, where
+// `describe` swallows it into a "reboot the simulator" hint that has nothing to
+// do with the real cause. Give an actionable, self-explaining error instead:
+// name the exact path, the env override that relocates the lookup, and how to
+// produce the binary.
+function requireTcpArtifact(dir: string, name: string, kind: "binary" | "dylib", envVar: string): string {
+  const p = path.join(dir, name);
+  if (!fs.existsSync(p)) {
+    throw new Error(
+      `TCP-transport ${kind} not found: ${p}. This ${kind} is required to drive an ` +
+        `ios-remote (sim-remote) device over the QUIC tunnel. It ships in the argent ` +
+        `package under bin/<platform>/tcp/ and dylibs/tcp/; if you are running a local ` +
+        `build, produce it with \`npm run build:ios-binaries:tcp\`. To point the lookup ` +
+        `at a directory that already contains it, set ${envVar}=<dir>.`
+    );
+  }
+  return p;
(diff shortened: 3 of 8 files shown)


filip131311 (approved):
I can not see anything wrong with this expect verbosity of 
```js 
return requireTcpArtifact(
    DYLIB_TCP_DIR,
    "libArgentInjectionBootstrap.dylib",
    "dylib",
    "ARGENT_NATIVE_DEVTOOLS_TCP_DIR"
  );
```

My agent was able to provide a correct describe

filip131311 (changes requested):
jk I have some problem give me a sec

filip131311 (approved):
does not seem to be a problem with the PR

filip131311:
Re: [my review above](https://github.com/software-mansion/argent/pull/536#pullrequestreview-4746215255) — context on the problem I hit while live-testing this PR, and why it is **not** a defect in this PR:

## TL;DR

`describe --udid remote:<UDID>` currently returns an empty `ROOT AXGroup` + the reboot hint on the Staging fleet, but the failure is in the **sim-remote tunnel layer**, one step *below* everything this PR touches. The fleet is mid-rollout and no published client currently matches the Staging orchestrator's tunnel-mux protocol.

## What this PR fixes — verified live against a remote iPhone 17 (Staging, router `:3020`)

- ✅ `sim-remote spawn … --json`: the exact calls that used to break (`bootstrapAx`'s `defaults write`) now round-trip clean `{exit_code,stdout,stderr}` JSON; all three `com.apple.Accessibility` flags get applied on the remote sim.
- ✅ TCP ax-service resolves from the platform-neutral `bin/tcp/` (real signed release artifact via the patched `download-native-binaries.sh`), bundles correctly via `bundle-tools.cjs`, uploads and `simctl spawn`s on the orchestrator, and dials back out of the simulator (confirmed with a manual spawn: `ax-service: failed to connect to 127.0.0.1:6100: Connection refused` — i.e. the binary runs in-sim and reaches the dial step).
- ✅ `describe` reports `source: "ax-service"` instead of dying before the daemon ever started.

## The actual current blocker (infra, not this PR)

The reverse tunnel that the ax-service needs to dial back to the host fails to establish:

```
$ sim-remote proxy start <UDID> 6100
error: tunnel-mux connection closed

# ~/Library/Logs/radon/sim-remote-daemon.log
tunnel-mux: inbound accept ended: closed by peer: protocol-version-mismatch (code 16)
```

Root cause is version skew from today's staged rollout in radon-cloud:

- The Staging fleet's **orchestrator** was redeployed this morning (run 29820121911, 09:54Z) with a newer tunnel-mux protocol.
- Every client binary that can authenticate against the Staging router (`:3020`) predates that protocol. The newer client builds (incl. the `sim-remote-releases` `daily`) are built with the **Filament** environment's PKI, so they fail TLS against `:3020` with a misleading `could not reach sim-router` (client binaries pin their environment's CA at build time).
- The first client build that both auths to Staging *and* should speak the new protocol is run 29839588340 (main, Staging env) — it logs in and attaches fine, but the matching fleet deploy (run 29839607457) is still in progress, so the orchestrator side hasn't caught up yet.

So with the ax-service unable to dial back, `describeIos` lands in its catch path → empty tree + `DEGRADED_HINT`. Once the fleet deploy lands I'll rerun the happy path (full tree from `source: "ax-service"`) and the negative path (missing TCP artifact → the new actionable `ARGENT_SIMULATOR_SERVER_TCP_DIR` error) and follow up here.

*(Posted by Claude on Filip's behalf while validating this PR end-to-end.)*

filip131311:
Follow-up to my [earlier comment](https://github.com/software-mansion/argent/pull/536#issuecomment-5035899861): the radon-cloud Staging fleet deploy landed and the new `daily` release binaries (radon-cloud, default `--server` = Staging `:3020`) resolve the tunnel-mux mismatch. Re-ran the full validation of this PR against a live remote iPhone 17:

**Happy path** — cold tool-server, `describe --udid remote:<UDID>` on the Messages conversation list:

```
Source: ax-service
  AXGroup "Conversations" id="ConversationList"  (0.000, 0.000, 1.000, 1.000)
  AXGroup "Messages"  (0.040, 0.137, 0.401, 0.047)
  AXGroup "+1 (888) 555-1212, 12/31/00 "  (0.000, 0.192, 1.000, 0.099)
  ...
  AXTextField "Search" id="magnifyingglass"  (0.114, 0.929, 0.051, 0.022)
```

Full tree from `source: "ax-service"`, no degraded hint — matches the on-screen content exactly.

**Negative path** — fresh tool-server with `ARGENT_SIMULATOR_SERVER_TCP_DIR` pointed at an empty dir:

```
hint: [AXService:remote:<UDID>] TCP-transport binary not found: <dir>/ax-service. This binary is
required to drive an ios-remote (sim-remote) device over the QUIC tunnel. It ships in the argent
package under bin/tcp/ and dylibs/tcp/ (platform-neutral); if you are running a local build,
produce it with `npm run build:ios-binaries:tcp`. To point the lookup at a directory that already
contains it, set ARGENT_SIMULATOR_SERVER_TCP_DIR=<dir>.
```

The actionable message surfaces verbatim instead of the misleading force-reboot hint. Both behaviors are exactly as described in the PR. ✅

Two review notes:

1. **Publish workflows still gate on the old path**: `.github/workflows/publish.yml:96` and `publish-next.yml:132` run `test -x packages/argent/bin/darwin/tcp/ax-service`, but after this PR the bundled binary lands at `bin/tcp/ax-service` (verified in a local bundle — `bin/darwin/tcp/` doesn't exist anymore). Both publish workflows will fail; they need the neutral path.
2. *(minor, pre-existing)* In the ax-service blueprint, `startProxy`/`spawnAxDaemon` failures after `startListener` leak the host listener (and for spawn, the active proxy) — teardown only wraps `waitForDaemonConnection`. The new missing-artifact throw makes that path an expected diagnosable state, so it may be worth a follow-up try/catch.

*(Posted by Claude on Filip's behalf.)*

PR #538, opened 2026-07-21
feat(ios): physical iOS 27 support

latekvo:
## What

Adds physical-iPhone control to argent by routing it through radon's native `ios_device` simulator-server controller (**software-mansion/radon#146**). A tethered iPhone becomes just another sim-server subcommand, exactly like an iOS simulator or a physical Android phone.

This supersedes #369 (now closed), which built a CoreDevice engine inside argent itself: a base64-embedded pymobiledevice3 Python sidecar plus a privileged root tunnel, and one-shot screenshots with no live streaming or recording. Putting the engine in the component that already speaks device wire protocols lets physical iOS reuse radon's media pipeline and keeps argent to routing and discovery. None of #369's Python or root-tunnel apparatus exists on `main`, so nothing of it is carried here - this PR is the complete feature against `main` on its own.

## What argent does

- **Discovery.** `list-devices` gains connected physical iPhones, found via `xcrun devicectl list devices` (`listIosDevices` / `parsePhysicalIosDevices`). Only currently-reachable iOS devices are kept - devicectl also enumerates the host's *simulators* as `platform: "iOS"`, so entries are filtered on the physical ECID UDID shape and on carrying a transport. iOS entries now carry `kind` (`"simulator"` / `"device"`) and physical ones a `productType`. Discovery is skipped entirely unless the flag is on.
- **Classification.** `classifyDevice` recognises the physical UDID shape (`8hex-16hex`) as iOS, and `resolveDevice` resolves it to `kind: "device"` - the same shape-based split Android already uses for emulator vs phone. `isPhysicalIosUdid` / `isPhysicalIos` are the shared predicates.
- **`subcommandForDevice`** returns `ios_device` for a physical iPhone, mirroring `android_device`. The normal sim-server spawn / `api_ready` / stdin-command path drives it from there.
- **Input and capture over the sim-server.** `gesture-tap`, `gesture-swipe`, `keyboard`, `button`, `rotate`, `screenshot` and `screen-recording-start` / `-stop` resolve a physical iPhone to `simulatorServerRef` and go through the shared `sendCommand` / `httpScreenshot` transport, with no special-cased branch. `gesture-swipe` settles on device, since the controller replays the eased Move samples. `button` rejects `appSwitch` and `actionButton`, which have no hardware HID equivalent, before spawning anything. `screenshot` skips its tvOS probe for a hardware UDID, which can never be a sim runtime.
- **App lifecycle over devicectl.** `launch-app`, `restart-app`, `reinstall-app` and `open-url` shell `xcrun devicectl` for `kind: "device"` instead of `simctl`. These don't go through the sim-server factory, so each calls the flag assertion directly.
- **`describe`** reads the on-screen accessibility tree from the sim-server's `/api/ax-tree` endpoint - radon's native axAudit over CoreDevice DTX - through `ios-coredevice-ax-adapter.ts`. One USB session drives both input and describe, so there's no second tunnel and no pmd3 involved. The fetch is bounded by an `AbortSignal.timeout`, like the screenshot path.
  - **Frames are approximate, and results are not comparable between calls.** radon#146 sends captions + reading order only; Apple exposes no per-element geometry app-free on hardware. The adapter therefore interpolates *every* frame (full-width rows, top to bottom). Separately, the device's own VoiceOver cursor rotates the element order by one on each read, so a given element's synthesised frame changes between calls and two `describe` results cannot be diffed. The describe hint, README and tool docs all say so. The adapter still honours real rects if a future sim-server sends any, so landing geometry is a producer-side change.
- **Opt-in** stays behind the `physical-ios-devices` flag. The sim-server factory is the chokepoint every CoreDevice-backed tool resolves through, raising `CORE_DEVICE_FLAG_DISABLED` when it's off; `list-devices` checks the flag before discovering at all, and the four devicectl tools call the same assertion directly.
- **`boot-device`** on a physical UDID has nothing to boot, so it resolves (spawns) the sim-server as an explicit "prepare" step - the flag gate and an unauthorized device surface there rather than on the first tap.
- **What stays off.** `gesture-pinch`, `gesture-rotate`, `tv-remote`, the seven `native-*` devtools and the `native-profiler-*` / profiler query tools set `apple.device: false`, so the capability gate rejects a physical UDID with a 400 rather than failing deep. Each declaration records why: the native-devtools family injects its dylib with DYLD via `simctl spawn`, which cannot reach a signed app on hardware; the two multi-touch gestures need two simultaneous contacts and the CoreDevice digitizer is single-contact; `native-profiler-*` resolves its target through `simctl spawn` / `simctl listapps`. (`paste` is not in this list: it is absent from `setup-registry.ts` on every platform, so its gate never runs at all.) `gesture-custom` stays *enabled* and rejects only its own two-contact half per request. `ax-service` and `native-devtools` also refuse `kind: "device"` at the blueprint, as a backstop for direct resolution.
- **README** gains a "Physical iOS devices (experimental)" section covering requirements, setup and the current limitations.

## Deliberately left out

- **Element geometry for `describe`.** Tracked on radon#146; until then every frame is interpolated.
- **Multi-touch.** Pinch and rotate need a second contact the CoreDevice digitizer does not expose.
- **Distribution.** Nothing - this needs no argent-side step. radon#146 merged on 2026-07-28 and ships `ios_device` in the macOS argent sim-server binary; `scripts/download-simulator-server.sh` pulls the rolling `radon-main` tag rather than a pinned version, so there is no binary version to bump. The published `simulator-server-argent-macos` asset (rebuilt 2026-07-29 15:21Z) carries `ios_device`, `ios_axtree`, `/api/ax-tree` and `device_controller/ios_device/axaudit.rs`, so a release cut from here bundles the controller. An older bundled binary still exits 1 with "Unrecognized argument: ios_device"; that is classified and told to upgrade rather than reading as a crash.

## Verification

Hardware (physical iPhone, iOS 27, USB), end to end through argent's real tool-server (`POST /tools/*`), pointed at a local radon build via `ARGENT_SIMULATOR_SERVER_DIR`:

- `screenshot` on home and Settings, pixel-verified as live frames (a Settings frame differs from home).
- `gesture-tap` and `gesture-swipe` change the on-screen tree; `button home` returns to home; `button appSwitch|actionButton|back` reject cleanly with a 400 and the driver survives and still screenshots.
- `describe` over `/api/ax-tree` reflects the frontmost app (home vs Settings), with `source: coredevice-ax`. Repeated calls return the same elements rotated by one - the VoiceOver-cursor behaviour documented above, not a stability guarantee.
- Adversarial inputs (out-of-range normalized taps, a gated tool) are handled without killing the driver, and no pmd3 processes are spawned at any point.
- An adversarial E2E sweep converged clean over consecutive runs.

Local checks on the current head: `tsc --build`, `tsc -p packages/tool-server/tsconfig.test.json`, `eslint . --max-warnings 0`, `prettier --check`, `npm run typecheck:scripts`, `npm run test:scripts` (59 pass) and the full vitest suite (4800 pass, 1 skipped).

## Review pass

A review of this branch found three defects, each reproduced before being fixed and pinned by a test that fails on the un-changed code:

- **A malformed `/api/ax-tree` payload crashed the describe adapter with a bare `TypeError`** - a 500 carrying no `error_code`, which reads as an argent bug rather than a device answering out of contract. `{"elements":{}}` and `{"elements":"str"}` gave "els.map is not a function", `[null]` gave "Cannot read properties of null", a numeric caption gave "caption.split is not a function", and a non-numeric screen width escaped the `sw <= 0` guard as `NaN` and surfaced a raw `ZodError`. The `?? []` only ever guarded null/undefined. The response is now validated where it enters, the way describe's other adapter validates its own raw payload.
- **`open-url` reported success for a URL it never opened.** The url was passed straight through as the last argv element of `xcrun devicectl device process openURL`; that command with `--help` in the url position exits 0 and prints help, so the tool returned `opened: true`. The schema now requires an RFC 3986 scheme - the same guard `launch-app` already applies to `bundleId`, and one that also rejects the schemeless input no backend could have opened.
- **A bundled simulator-server predating the controller looked like a crash.** The released binary exits 1 with "Unrecognized argument: ios_device"; that surfaced as a generic "exited before becoming ready", bucketed with genuine crashes, and never said the binary ships inside argent. It now names the cause, says to upgrade, and has its own failure code.

The same pass compared each new physical-iOS path against its already-pinned physical-Android twin. The capability gate was the largest gap: the simulator-only list was hand-written under a comment claiming it was "enumerated rather than sampled", but `settings-permissions` was missing, and giving it `apple.device: true` left the whole suite green while a physical iPhone walked into `simctl privacy` with a hardware UDID. That set is now derived from the registry, so the claim holds by construction and future tools are covered without being listed. Also newly pinned, each against the mutation it catches: the `shutdownDevice` / `shutdownOwnedDevice` physical-iPhone refusal, `getSimulatorRuntimeKind`'s hardware short-circuit, the `ax-service` blueprint backstop, and `httpAxTree`'s error branch (deleting it turned an HTTP 500 into an empty element list, so `describe` reported "the screen may be off or locked" for a read that never happened).

## Second review pass

A second pass reviewed the branch again, including the first pass's own fixes. Same bar: every defect reproduced before being touched, the same reproduction re-run after, and a test that fails on the un-fixed code.

**Reported success for something that did not happen, or could not:**

- **Flows accepted a physical iPhone and then failed every step.** Treating `state: "connected"` as ready made an attached iPhone a run target - explicitly, and as auto-detection's single match - but `fetchFlowTree` sends every iOS device to the native view hierarchy, which `nativeDevtoolsBlueprint` refuses for `kind: "device"`. Every selector step failed with a devtools error, and a `launch` step answered "restart the argent server and try again", advice that can never work. The runner now refuses a hardware udid at device resolution with the permanent reason, and auto-detection ignores it rather than turning a lone simulator into "2 booted devices matched".
- **`await-screen-idle` could never settle on a screen of 120+ elements.** Sorting the signature cancels the read's rotation only while the whole screen fits in one read; `describe` asks for at most 120 and the walk advances the VoiceOver cursor per read, so past that each poll covers a *different window*. A still screen of 121 elements took 76 polls, spent the full 15s, and returned `settled: false` saying nothing. It now recognises a truncated read and answers immediately with the cause.
- **`httpAxTree` and `httpScreenshot` died with a bare `TypeError` on a JSON `null` body** - including on the 5xx responses a server that cannot answer is likeliest to send. That is the exact class the axAudit schema was added to prevent; the schema ran *after* the in-band `{ error }` read.
- **Every `devicectl` failure was reported as a signing problem** and the child's own output was discarded, so a locked phone was sent to re-sign a bundle that was fine. `subprocessFailureMetadata` records only exit code and signal, so nothing else carried devicectl's explanation.
- **`shutdownOwnedDevice` reached for a physical Android phone** while the comment beside it and the test covering it both described a physical-Android exclusion that did not exist.

**Reachable but wrong:**

- `describe` discarded the VoiceOver state tokens it strips from the caption, so an enabled and a disabled control came back byte-identical - while the hint promises the traits are exact. They now land on `disabled` / `selected`.
- `gesture-custom` resolved (spawned) the CoreDevice session before rejecting a two-contact request, so a device whose session cannot start answered with a transport error instead of "this gesture needs two contacts" - and falsified `run-sequence`'s own new comment.
- `await-ui-element` was reachable on hardware unadapted: the 5s simulator default against ~2s reads, and a description promising frame and reading-order semantics that do not hold there.
- `profiler-load`'s `load_native` mode built a `NativeProfilerSession` for a hardware udid that every reader of it rejects.
- The flow refusal answered HTTP 500 where every sibling refusal answers 400.
- The caption parser promoted a trait to an element's `value` on an all-trait caption (`"Dimmed, Button"` → `{ label: "Dimmed", value: "Button" }`).

**Docs that contradicted the code**, each now derived or pinned: the `argent-screen-recording` skill listing physical iPhones as unsupported (the skill is what an agent reads before recording, so the capability underneath was never consulted); three separate stale tool enumerations; a claimed gating rule the gate does not implement; four mutually inconsistent accounts of which tools shell `devicectl`; and a "saves a `simctl` probe" rationale for a guard that saves nothing.

**Coverage.** A mutation sweep found the physical-iOS opt-in gate itself unpinned - deleting `assertPhysicalIosEnabled` from `simulatorServerRef` left all 3176 tests green while taking the flag off screenshot, gesture-*, keyboard, button, rotate, describe, screen-recording and boot-device at once. Also newly pinned, each against the mutation that exposed it: `list-devices`' flag gate, `kind` tag and readiness rank; `listIosDevices`' darwin guard, `--quiet` and `killSignal` (its simctl twin pins all three, but that file's filter keys on `simctl` and could never see a devicectl spawn); `boot-device`'s `force` recycling and its await; `describe`'s empty and truncated hints, element ceiling and read bound; the placeholder-frames warning; the second-touch guard's `y2` half; and the adapter's clamp, arity guard and value-split restriction.

Local: `tsc --build`, the test tsconfig, `eslint . --max-warnings 0`, `prettier --check`, and the full vitest suite (3210 pass, 1 skipped). All CI checks green, including macOS + iOS E2E.

radon#146 has landed (merged 2026-07-28), so this no longer waits on it. Its `idevice` dependency is pinned pre-1.0 (`=0.1.65`) - flagged there, and now a shipped dependency rather than an open question.


packages/tool-server/src/tools/gesture-rotate/index.ts
```diff
@@ -84,7 +84,7 @@ interface Result {
 }
 
 const capability: ToolCapability = {
-  apple: { simulator: true, device: true },
+  apple: { simulator: true, device: false },
```
filip131311, line 73: why?

packages/tool-server/src/tools/keyboard/index.ts
```diff
@@ -42,7 +42,7 @@ const zodSchema = z.object({
 type Params = z.infer<typeof zodSchema>;
 
 const capability: ToolCapability = {
-  apple: { simulator: true, device: true },
+  apple: { simulator: true, device: false },
```
filip131311: ?? wait what?

filip131311 (changes requested):
This is not ready keyboard, pinch custom gestures is not working

 argent has issues with just installinga and testing an app and this is a problem as well 

we will need to push this to 0.19.0

packages/tool-server/src/tools/gesture-rotate/index.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Not a choice - it's a wire-protocol limit, so this one stays `false`.

Physical-iPhone input goes through radon's `ios_device` controller (software-mansion/radon#146), which drives the device's `mainTouchscreen` over `UniversalHidService`. That report carries **one contact**; the device's only other digitizer-sounding surface, `touchscreenGesture`, enumerates as HID usage page `0x01` "Generic Desktop" / usage `0x02` "Mouse" - the mirroring-window pointer, not a second finger. `gesture-rotate` is defined as two contacts orbiting a center, and there is no single-touch approximation of it worth shipping, so the tool is gated rather than half-implemented.

Verified against this branch's own tool-server (`POST /tools/*`, physical UDID `00008120-...`):

```
gesture-rotate -> 400  Tool 'gesture-rotate' is not supported on ios device (kind 'device' not supported).
gesture-pinch  -> 400  same
gesture-custom -> passes the gate (single-touch Down/Up reaches the sim-server resolution)
keyboard       -> passes the gate
```

So the neighbouring single-touch tools are not caught by this: `gesture-custom` stays **enabled** on hardware and rejects only its own `x2`/`y2` half per request (dfbef071), pinned by `physical-ios-followups.test.ts` - reverting that file to its pre-fix state fails 2 cases.

The reason was implicit when you asked; 56dbf9cd writes it into the capability declaration itself so the next reader doesn't have to.

packages/tool-server/src/tools/keyboard/index.ts
latekvo: Found a way around this in be0177cf

---

\[[Diplomat](https://github.com/latekvo/Diplomat)\]: **Fixed in be0177cf.**

radon#146 implements keyboard over `IndigoHidService` keyboard usages and hardware-verified it ("hello" typed into Spotlight); argent was declaring `apple.device: false` anyway, so the controller's keyboard surface had no consumer. `keyboard` now declares `apple.device: true` and a physical iPhone routes to `typeSimulatorServer` - `getSimulatorRuntimeKind` short-circuits a hardware UDID, so the tvOS probe in `keyboard/platforms/ios.ts` never shells `simctl` for it.

Reproduced both directions against this branch's tool-server, physical UDID `00008120-...`:

```
HEAD          POST /tools/keyboard {"text":"hi"} -> past the capability gate, into simulator-server resolution
keyboard cap  reverted to device:false          -> 400 Tool 'keyboard' is not supported on ios device (kind 'device' not supported)
```

That revert is also the mutation check on the test: with `apple.device` flipped back, `physical-ios-followups.test.ts:513` fails with exactly the 400 above, so the fix is pinned rather than just applied.

The same commit turned on `rotate`, `open-url`, `restart-app` and `screenshot-diff`; 248c186e added screen recording, `reinstall-app` and `await-screen-idle`. `gesture-pinch`/`gesture-rotate` are the only input tools still off, for the single-contact reason on the other thread.

latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat)\]: Re: the `CHANGES_REQUESTED` review - going through it point by point. Everything below was reproduced against this branch's own tool-server before and after the fix, and each fix has a mutation check (revert it, the suite goes red).

**`keyboard`** - real, fixed in `be0177cf`. radon#146 implements and hardware-verifies keyboard over `IndigoHidService`; argent was still declaring `apple.device: false`, so nothing drove it. Reverting the capability reproduces exactly what you saw:
```
POST /tools/keyboard {"udid":"00008120-…","text":"hi"}  ->  400 Tool 'keyboard' is not supported on ios device
```
and on HEAD the same call goes past the gate into simulator-server resolution. `physical-ios-followups.test.ts:513` fails on the reverted capability.

**custom gestures** - real, fixed in `dfbef071`. `gesture-custom` is now enabled on hardware and rejects only the `x2`/`y2` half, per request, before the first event goes out - so long press, drag-and-drop and custom scroll work and no contact is left down on a rejection. Reverting that file fails 2 cases in the same test file.

**`pinch`** - not fixable here, and this is the one I'm dismissing rather than fixing. The `ios_device` controller drives `mainTouchscreen` over `UniversalHidService`, whose report carries **one contact**; `touchscreenGesture` (the other candidate surface) enumerates as Generic Desktop / Mouse - the mirroring pointer, not a second finger. `gesture-pinch` and `gesture-rotate` therefore reject at the capability gate with a clean 400 instead of failing deep, and `56dbf9cd` writes that reason into the two capability declarations.

**installing and testing an app** - real, fixed in `248c186e`. At the commit you reviewed, `reinstall-app` threw `UnsupportedOperationError` for `kind: "device"` outright ("out of scope for the CoreDevice path"), so there was no way to get a build onto the phone through argent. It now uninstalls + installs through `xcrun devicectl device install app`, with the signing requirement named up front in the failure message rather than buried in devicectl's output. Restoring the old file fails 5 cases in `physical-ios-devicectl.test.ts`.

**`argent flags` blurb** - found while checking the above and fixed in `903c6163`: it still enumerated "screenshot, tap, swipe, hardware buttons, describe, and launch-app", which stopped being the supported set once the tools above landed, with nothing failing. It no longer lists tools, and a registry-derived assertion now fails if a tool id reappears in it.

**0.19.0** - your call, and I'm not touching the milestone. One input for it though: the distribution caveat in the PR description is out of date. radon#146 merged on 2026-07-28, and `scripts/download-simulator-server.sh` pulls from the rolling `radon-main` tag rather than a pinned version - the published `simulator-server-argent-macos` asset was rebuilt 2026-07-29 15:21Z and `strings` on it shows `ios_device`, `ios_axtree`, `/api/ax-tree` and `device_controller/ios_device/axaudit.rs`. So there is no binary version bump left to do; a release cut from here bundles the controller. I've corrected the PR description accordingly.

Still standing on hardware: multi-touch, and `describe` frames being interpolated (radon#146 follow-up) - both documented in the README's limitations section.

packages/tool-server/src/tools/gesture-rotate/index.ts
latekvo: Discussed in person. Currently looking into ways around this using xctest.

latekvo:
Found a way around most issues i previously believed to be dead ends. Looking into ways around the still not functional multi-touch gestures.

All installation issues should be fixed now.

Marked as ready too quickly, converting to draft for now.

ETA tuesday if i find a way for multitouch through CoreDevice, longer if it'll require xctest, long if we'll have to switch to pure xctest, but that'd be worst case.

started researching pre 27 support  just now, no ETA yet. Of course we know it's trivially possible, but i'm still looking into how cleanly we can integrate it as a fallback into current architecture

packages/tool-server/src/tools/devices/list-devices.ts
```diff
@@ -250,7 +279,29 @@ Booted/ready devices are listed first. Platforms whose CLI is unavailable are si
...
+    const iosTagged: IosDevice[] = [
+      ...ios.map(
+        (s): IosDevice => ({
+          platform: "ios",
+          kind: "simulator",
+          udid: s.udid,
+          name: s.name,
+          state: s.state,
+          runtime: s.runtime,
+          runtimeKind: s.runtimeKind,
```
latekvo, line 262: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `list-devices` stopped emitting `deviceSet` on iOS entries while its description still promises the tag.

Pre-PR the iOS mapping was `const iosTagged: IosDevice[] = ios.map((s) => ({ platform: "ios", ...s }))` with `type IosDevice = IosSimulator & { platform: "ios" }`, so every field of `IosSimulator` — including `deviceSet?: string` (ios-devices.ts:22, populated at ios-devices.ts:82 for simulators in an additional CoreSimulator set) — reached the tool's output. The PR replaces the spread with a hand-enumerated object literal and a hand-written `IosDevice` type that has no `deviceSet` member, so the field is now dropped for every simulator. The tool's own description, untouched at list-devices.ts:232, still says: "iOS simulators from an additional CoreSimulator device set (the 'ios.additionalDeviceSets' configuration — e.g. devices created by Radon IDE) are listed alongside default-set ones, tagged with their owning 'deviceSet' path". Nothing catches it: ios-device-sets.test.ts:166-167 asserts `deviceSet` on `listIosSimulators()` output, not on the tool result, and the new physical-ios tests only inspect `udid`/`kind`/`state`/`productType`/`runtime`.

**Trigger.** Configure `ios.additionalDeviceSets` (e.g. a Radon IDE simulator) and call `list-devices`. Before this PR the Radon-set entry came back as `{ platform: "ios", udid, name, state, runtime, deviceSet: "/Users/…/Radon/Devices" }`; after it the same entry comes back without `deviceSet`, so an agent following the description has no way to tell a headless additional-set simulator from a default-set one.

**Checked against.** NEW list-devices.ts:282-293: `const iosTagged: IosDevice[] = [ ...ios.map((s): IosDevice => ({ platform: "ios", kind: "simulator", udid: s.udid, name: s.name, state: s.state, runtime: s.runtime, runtimeKind: s.runtimeKind, })), …` — no `deviceSet`. OLD (git show 23b9ae5c:…/list-devices.ts:19,253): `type IosDevice = IosSimulator & { platform: "ios" };` / `const iosTagged: IosDevice[] = ios.map((s) => ({ platform: "ios", ...s }));`. Source field, unchanged, ios-devices.ts:21-22: `/** Owning CoreSimulator device-set directory; absent for the default set. */ deviceSet?: string;`.

packages/tool-server/src/blueprints/simulator-server.ts
```diff
@@ -216,29 +263,60 @@ async function spawnSimulatorServerProcess(
...
+    // a device that isn't paired / has Developer Mode off, and a bundled
+    // simulator-server too old to know this subcommand. Keep a bounded tail to
+    // put in the error. Bounded because a crash loop can emit unbounded output.
+    let stderrTail = "";
     proc.stderr?.on("data", (data: Buffer) => {
       process.stderr.write(`[sim ${udidTag}] ${data}`);
+      stderrTail = (stderrTail + data.toString()).slice(-2000);
     });
 
     proc.on("exit", (code, signal) => {
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The new stderr-based failure classification reads `stderrTail` from `proc.on("exit")`, which can fire before the stderr data is delivered.

`stderrTail` is accumulated in a `proc.stderr.on("data")` handler and consumed inside `proc.on("exit")`. Node emits `'exit'` from the process handle's `onexit` callback and only then schedules `flushStdio` on `process.nextTick`, so data still sitting unread in the stderr pipe is not guaranteed to have reached the `data` handler when `'exit'` runs (the documented reason `'close'` exists — "when the 'exit' event is triggered, child process stdio streams might still be open"). When that race loses, `detail` is the empty string, `unsupportedSubcommand` is false, and the error falls back to the generic `SIMULATOR_SERVER_READY_EXITED` with no output and no upgrade advice — i.e. exactly the two cases the comment above it names as the ones a user is most likely to hit (an old bundled binary, an unpaired / Developer-Mode-off device) are the ones that can silently lose their diagnosis. The test pins the happy ordering only, by pushing to `fakeProc.stderr` in one `setImmediate` and emitting `exit` in the next, so it cannot observe the race.

**Trigger.** A bundled simulator-server that predates the controller writes "Unrecognized argument: ios_device" to stderr and exits 1 immediately. If libuv has not yet read that pipe when the process-exit callback runs, `stderrTail` is "", so the caller gets "simulator-server (ios_device) exited before becoming ready with code 1." bucketed as SIMULATOR_SERVER_READY_EXITED instead of SIMULATOR_SERVER_SUBCOMMAND_UNSUPPORTED plus "Upgrade argent to a version that ships it."

**Checked against.** packages/tool-server/src/blueprints/simulator-server.ts:272-276 `let stderrTail = ""; proc.stderr?.on("data", (data: Buffer) => { …; stderrTail = (stderrTail + data.toString()).slice(-2000); });` then :278-286 `proc.on("exit", (code, signal) => { const detail = stderrTail.trim().split("\n")… ; const unsupportedSubcommand = /unrecognized (?:argument|subcommand|command)/i.test(detail);`. Compare the sibling exit listener at :463 which reads nothing from stdio. The test forces the ordering: test/simulator-server-blueprint.test.ts `setImmediate(() => { fakeProc.stderr.push(stderr); setImmediate(() => fakeProc.emit("exit", code, null)); });`

packages/skills/rules/argent.md
```diff
@@ -60,7 +60,7 @@ Before booting, running, or interacting with any app, call `list-devices` first
 Decision order:
```
latekvo, line 62: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The always-on tapping rule still bans the only tap workflow that works on a physical iPhone.

This PR's physical-iOS design makes `screenshot` the sole authority for positions on hardware and `describe` explicitly non-positional: the describe hint says "Use this to learn WHAT is on screen, never WHERE. Locate anything you intend to tap with screenshot first — screenshot is authoritative for positions" (describe/platforms/ios/index.ts, PHYSICAL_IOS_AX_HINT), the formatter header says "Use them only to read order — never as tap coordinates. Locate anything you intend to tap with `screenshot` first, then pass those coordinates to gesture-tap / gesture-swipe" (format-tree.ts:203-208), and the README repeats it. The PR edited `packages/skills/rules/argent.md` (the `alwaysApply: true` rule) for device selection at line 63, and edited `argent-device-interact/SKILL.md` and `argent-ios-simulator-setup/SKILL.md` for the new hardware shape — but left the `<tapping_rule>` in the same rule file, and the same skill's navigation rules, stating the exact opposite.

**Trigger.** An agent drives a connected iPhone. Per rules/argent.md:39 and :48 it must not take coordinates from a screenshot and must tap from a discovery tool; `describe` returns full-width placeholder rows (`{x:0.04, width:0.92, height:0.05}` synthesised from list position by `approxFrame`) whose y also rotates between calls. Following the rule, the agent taps the centre of a synthesised row and hits whatever is actually at that y.

**Checked against.** packages/skills/rules/argent.md:39 `<important>**Never** derive tap coordinates from a screenshot</important>` and :48 "…Do not tap if you have not called a discovery tool in the current step. Screenshots alone are never sufficient for coordinates." vs. this PR's format-tree.ts:203-208 "FRAMES ARE PLACEHOLDERS, not measurements … never as tap coordinates. Locate anything you intend to tap with `screenshot` first". The same file's line 63 WAS updated by this PR ("a connected physical iPhone is a `platform: "ios"` entry with `kind: "device"`…"), so the file was in scope.

packages/tool-server/test/await-screen-idle-physical-ios.test.ts
```diff
@@ -0,0 +1,168 @@
...
+    }));
+    const p = tool.execute({}, { udid: "R5CT30ABCDE" } as never) as Promise<{
+      settled: boolean;
+      note?: string;
+    }>;
+    let done = false;
+    void p.then(() => (done = true));
+    for (let i = 0; i < 400 && !done; i++) await vi.advanceTimersByTimeAsync(250);
+    const r = await p;
+    expect(r.note, "the CoreDevice element ceiling is not Android's").toBeUndefined();
```
latekvo, line 143: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The Android control case cannot fail on the mutation its own comment names.

The test's comment says it exists to pin the platform half of `rotatingRead`: "Dropping the platform half would give Android the order- and frame-free signature … and the CoreDevice truncation refusal, on a tree that has neither problem." But its two assertions survive that mutation. With `rotatingRead = device.kind === "device"` (platform half deleted from await-screen-idle/index.ts:201), the serial `R5CT30ABCDE` resolves to `{platform:"android", kind:"device"}` so `rotatingRead` becomes true: `sawFullRead` is set (125 >= 120) and the signature becomes sorted/frame-free — but `describeAndroid` returns a byte-identical tree on every poll, so the sorted signature is also identical and the wait still settles. `settled` is then `true`, and the note is attached only when `!settled` (index.ts:242), so `note` stays `undefined`. Both `expect(r.note).toBeUndefined()` and `expect(r.settled).toBe(true)` pass.

**Trigger.** Delete `device.platform === "ios" && ` from `const rotatingRead = device.platform === "ios" && device.kind === "device";` (await-screen-idle/index.ts:201) and run this file: it stays green, even though every physical Android phone now gets the frame-free signature (which stops catching a pure-motion animation) and the 15s CoreDevice timeout default instead of 3s.

**Checked against.** Test, await-screen-idle-physical-ios.test.ts:109-139: the Android stub is `describeAndroid.mockImplementation(async () => ({ source: "uiautomator", tree: { … children: Array.from({ length: rows }, …) } }))` — the same tree every call — and the only assertions are `expect(r.note, …).toBeUndefined(); expect(r.settled).toBe(true);`. Production, await-screen-idle/index.ts:237-243: `const settled = poll.result === true; return { settled, waitedMs, polls, ...(!settled && sawFullRead ? { note: TRUNCATED_READ_NOTE } : {}) };` — the note is unreachable whenever the screen settles.

packages/tool-server/test/physical-ios-followups.test.ts
```diff
@@ -0,0 +1,873 @@
...
+    expect(androidPhone.kind).toBe("device");
+    expect(() =>
+      assertSupported("native-profiler-start", nativeProfilerStartTool.capability, androidEmu)
+    ).not.toThrow();
+    expect(() =>
+      assertSupported("native-profiler-start", nativeProfilerStartTool.capability, androidPhone)
+    ).not.toThrow();
+  });
+
+  // Every tool whose backend is simulator-only, enumerated rather than sampled:
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The "enumerated rather than sampled" simulator-only list is still incomplete.

The comment claims the list below it is exhaustive — "Every tool whose backend is simulator-only, enumerated rather than sampled: an unlisted one is how a physical iPhone reaches a `simctl spawn` / xctrace path" — and the test is named "every simulator-only tool rejects a physical iPhone". Deriving the set from source gives 17 tools that declare `apple.simulator: true` without `apple.device: true`; the list names 16. `settings-permissions` (tools/settings-permissions/index.ts:49, `apple: { simulator: true }`) is missing — the very tool the PR body calls out as the gap that made this list's claim false. The PR body says the problem was fixed ("That set is now derived from the registry, so the claim holds by construction"), but that only happened in the new physical-ios-capability-sweep.test.ts; the hand-written list and its exhaustiveness claim were left in place here, so the repo now carries a derived set and a stale hand-written one making the same promise.

**Trigger.** Add a new simulator-only tool (or read the file today): the comment tells the next contributor this list is the exhaustive record of simulator-only tools, while `settings-permissions` — an existing, registered, simulator-only tool — is absent from it.

**Checked against.** physical-ios-followups.test.ts:787-806: `// Every tool whose backend is simulator-only, enumerated rather than sampled:` followed by a 16-entry `simulatorOnly` array (paste, gesture-pinch, gesture-rotate, tv-remote, 7 native-*, 3 native-profiler-*, profiler-stack-query, profiler-combined-report). Missing member: tools/settings-permissions/index.ts:49 `apple: { simulator: true },`. The new sweep, physical-ios-capability-sweep.test.ts, asserts `expect(ids, …).toContain("settings-permissions")` and `expect(simulatorOnly.length).toBeGreaterThanOrEqual(16)` — i.e. it knows about the 17th.

packages/tool-server/src/tools/paste/index.ts
```diff
@@ -27,8 +27,15 @@ interface Result {
 // Capability gate (HTTP layer + dispatchByPlatform-style consumers) rejects
 // Android serials with "Tool 'paste' is not supported on android". The handler
 // itself is iOS-only — no platforms/ split needed.
+//
+// `apple.device: false`: this sends the sim-server's `paste` command, and the
+// `ios_device` controller's clipboard-write arm sits behind a cargo feature the
+// shipped simulator-server binary is not built with. `sendCommand` is
+// fire-and-forget — the same property the tvOS guard in the simulator-server
+// blueprint exists for — so without this gate the tool would report
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `paste`'s new `apple.device: false` comment credits a gate that never runs.

The added rationale says "without this gate the tool would report `pasted: true` for text that never reached the phone". `pasteTool` is not registered anywhere in production: grepping the whole repo for `pasteTool` outside its own file yields only three test files, `setup-registry.ts` contains no `paste`, and `run-sequence`'s ALLOWED_TOOLS (run-sequence/index.ts:9-23) does not include it — so `assertSupported` is never invoked for `paste` and the capability edit cannot prevent anything. The PR body states this correctly ("`paste` is not in this list: it is absent from `setup-registry.ts` on every platform, so its gate never runs at all"), which is the opposite of what the comment beside the code says. Separately, physical-ios-capability-sweep.test.ts:29 explains paste's manual addition as "(macOS-only registration)" — also wrong, and not what the file it cites says: interaction-messages.test.ts keeps the macOS-only note for the Lens tools and a distinct "These definitions intentionally exist outside createRegistry" note for `paste` and `simulator-server`.

**Trigger.** Read paste/index.ts:31-36 and conclude the gate protects hardware from a silent false success; there is no path that reaches it, because the tool is never in the registry the HTTP capability check reads from.

**Checked against.** paste/index.ts:31-38: "…`sendCommand` is fire-and-forget … so without this gate the tool would report `pasted: true` for text that never reached the phone." + `apple: { simulator: true, device: false },`. `grep -rn "pasteTool" packages` → only src/tools/paste/index.ts:42 (the definition) and test/{physical-ios-followups,physical-ios-capability-sweep,keyboard-secrets,interaction-messages}.test.ts. physical-ios-capability-sweep.test.ts:27-30: "`paste` is defined outside `createRegistry` (macOS-only registration)" vs interaction-messages.test.ts:27-33 which separates the Lens macOS note from "These definitions intentionally exist outside createRegistry."

latekvo (commented):

packages/tool-server/src/tools/devices/list-devices.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: **Fixed in e6e8a8cf.**

`deviceSet` is back on iOS simulator entries: added to the hand-enumerated `IosDevice` type and spread from `listIosSimulators()` output (`...(s.deviceSet ? { deviceSet: s.deviceSet } : {})`), so an additional-set sim now comes back from the tool tagged with its owning set path and the description's promise holds again.

Pinned by a new tool-result-level test, `test/list-devices-deviceset.test.ts`, which mocks `listIosSimulators` with one default-set and one Radon-set entry and asserts the tag on the tool output (the existing ios-device-sets assertions only covered the util). Mutation check: deleting the spread line fails that test with exactly the missing-tag assertion.

packages/tool-server/src/blueprints/simulator-server.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: **Fixed in e6e8a8cf.**

The failure-classification listener moved from `proc.on("exit")` to `proc.on("close")`, which fires only after the stdio streams have ended — so the stderr tail it reads can no longer race empty against libuv's pending pipe reads. The sibling `exit` listener at :46x reads no stdio and stays put.

New repro in `simulator-server-blueprint.test.ts`: `failWithStderrAfterExit` emits `exit` *before* pushing the "Unrecognized argument" stderr and only then `close`. On the old code the rejection settles on `exit` with an empty detail and lands in SIMULATOR_SERVER_READY_EXITED — the test fails. On the fix it classifies SIMULATOR_SERVER_SUBCOMMAND_UNSUPPORTED with the upgrade advice. The pre-existing `failWith` helper was also updated to emit the realistic exit-then-close pair.

packages/skills/rules/argent.md
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: **Fixed in e6e8a8cf.**

The `<tapping_rule>` in `rules/argent.md` now carries an explicit physical-iPhone exception right after the retry rule: on a connected iPhone (`platform: "ios"`, `kind: "device"`) the rule is inverted — `describe` is for WHAT is on screen only, and anything you intend to tap is located with `screenshot` first. That matches PHYSICAL_IOS_AX_HINT and the format-tree header instead of contradicting them, and line 63's device-selection text already described the same shape.

packages/tool-server/test/await-screen-idle-physical-ios.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: **Fixed in e6e8a8cf.**

The Android control was rebuilt so the named mutation actually fails it. It now feeds a frame-only animation: two alternating trees whose labels stand still but whose element y positions shift every poll, with rows above the LIMIT so the truncation precondition also holds. Assertions: `settled === false` (a moving screen is not idle) and `note === undefined` (the CoreDevice ceiling is not Android's).

Mutation check: deleting `device.platform === "ios" &&` from `rotatingRead` makes the order- and frame-free sorted signature equate both trees, the wait reports settled mid-animation, and the test fails on `expected false to be false`'s negation — confirmed by running the suite with the narrowing dropped (1 failed). With the narrowing intact all 4 cases pass.

packages/tool-server/test/physical-ios-followups.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: **Fixed in e6e8a8cf.**

Two changes. `settings-permissions` is now in the pinned list, so the file no longer disagrees with the registry about its own membership. And the comment no longer claims exhaustiveness at all: it now says the list does per-tool pinning of each capability object, while the "every simulator-only tool rejects a physical iPhone" claim is owned — by construction, over the whole registry — by `physical-ios-capability-sweep.test.ts`. The test name was softened accordingly ("each known simulator-only tool...") so neither the comment nor the name promises more than the mechanism delivers.

packages/tool-server/src/tools/paste/index.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: **Fixed in e6e8a8cf.**

The `apple.device: false` rationale in `paste/index.ts` no longer credits a gate that never runs: it states plainly that the declaration is inert today because the tool is defined outside `createRegistry` and no dispatch path consults it, and that the declaration exists so a future registration cannot silently enable hardware without revisiting the fire-and-forget gap. The wrong "(macOS-only registration)" parenthetical in `physical-ios-capability-sweep.test.ts` was replaced with what interaction-messages.test.ts actually says — paste belongs to the definitions that intentionally exist outside the registry.

latekvo:
Closing for now, will be relevant after the xctest based approach is merged.

PR #541, opened 2026-07-22
fix(tool-server): tag gesture-pinch/gesture-rotate as ios/android in run-sequence

j-piasecki:
Both gestures appear to be supported on Android:

https://github.com/software-mansion/argent/blob/5d3f7bc62b278c03f3e63b45387cb41b937185c9/packages/tool-server/src/tools/gesture-pinch/index.ts#L54

https://github.com/software-mansion/argent/blob/5d3f7bc62b278c03f3e63b45387cb41b937185c9/packages/tool-server/src/tools/gesture-rotate/index.ts#L44


latekvo:
Ran the full E2E sweep across the review lenses (correctness, scope, edge cases, ripple effects). Confirmed both tools declare Android capability, dispatch through the same two-finger `sendTouchEvent` channel as `gesture-custom` (already tagged ios/android), these were the only two stale `[ios only]` tags in the repo, and no test asserts on the description text. Returned perfectly clean. Thanks for the fix!

latekvo (approved):

PR #545, opened 2026-07-22
feat(flow): pinch directive

j-piasecki:
Adds a selector-targeted pinch-zoom directive to flows: `pinch: { on?, scale }`. `on` resolves like `tap` (settled tree + auto-wait); omitted, the pinch centers on the screen. `scale > 1` zooms in, `< 1` zooms out.

## Design

**Options-map only, no bare form.** `tap: "Map"` is unambiguous; `pinch: "Map"` isn't (in or out?), so the parser requires `{ on?, scale }` and rejects a bare selector with authoring guidance.

**Validity checks, not policy caps.** `scale` must be finite, > 0, and ≠ 1 (a no-op scale is almost certainly a mistake) — but there is deliberately no magnitude cap. An extreme scale decomposes logarithmically into equal-ratio sub-gestures of ≤4× each, chained with a 250 ms recognizer-reset delay (`scale: 1e6` is just 10 gestures).

**Pure geometry module.** All placement math lives in `flow-pinch-geometry.ts` as deterministic data-in/data-out functions (no device access), so tests drive it with synthetic centers, spans, and edge guards. Per sub-gesture, a horizontal and a vertical candidate are built from the axis-matching target dimension (normalized X and Y are different units, so never `min(width, height)`), and the better one is dispatched.

**Edge safety scores, it never rejects.** Per-platform system-edge guards (wider on Android's back-gesture zones) demote candidates in ranking but never fail a step: viable travel beats edge safety (a risky-but-perceptible gesture beats a safe no-op the recognizer ignores), then fully-safe with the most travel, then motion parallel to the violated edge (edge swipes need motion *away* from the edge). Guards are resolved once per directive and passed as data — the seam for a future per-device query. The only geometry failure is literally zero on-screen finger travel; a tiny target is still attempted.

**Centroid drift near edges.** When expanding fingers would leave the screen, the centroid drifts inward — only on the moving axis, and only as far as needed. `gesture-pinch` grows optional `endCenterX`/`endCenterY` to interpolate the centroid alongside separation; omitted, behavior is unchanged.

**Open-loop by design.** There is no "current zoom" to read back, so the directive doesn't try to verify the multiplier — flows assert on the resulting UI instead.

**Platform rejections.** Vega is remote-driven (no touch input) and Chromium has no uniform pinch-zoom mapping for a desktop app (ctrl+wheel or app-specific controls) — both fail upfront with guidance to drive the app's zoom UI directly.

## Tests

Unit tests cover the geometry module (decomposition, candidate construction, edge-guard ranking), the directive end-to-end against a fake device, and the `gesture-pinch` centroid-drift interpolation.


packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -700,6 +720,69 @@ async function runLongPress(
...
+    centerY: center.y,
+    startDistance: selected.start,
+    endDistance: selected.end,
+    angle: selected.angle,
+  };
+  // Centroid drift rides the gesture only on the moving axis, and only when
+  // the clamp actually moved it.
+  const startCenter = selected.angle === 0 ? center.x : center.y;
+  if (selected.endCenter !== startCenter) {
+    args[selected.angle === 0 ? "endCenterX" : "endCenterY"] = selected.endCenter;
```
latekvo, line 775: The vertical-axis branch of this centroid-drift emit (angle 90 -> `endCenterY`) has no flow-level test exercising it. In flow-pinch.test.ts the only assertion that checks an emitted drift value uses a horizontal (angle-0) target, and both `endCenterY` assertions are `not.toHaveProperty` absence checks. Driving the real flow runner with a tall, narrow target near the bottom edge (frame x 0.475 / w 0.05, y 0.55 / h 0.45) selects angle 90 and emits `endCenterY: 0.5` drifted from `centerY: 0.775`, so the branch is reachable in practice. Because nothing pins it, a change that emitted `endCenterX` here regardless of the selected axis, or that compared `endCenter` against `center.x` unconditionally and so suppressed the drift, keeps all 38 tests green while pushing the end pointers off-screen for such targets.

packages/tool-server/src/tools/flows/flow-pinch-geometry.ts
```diff
@@ -0,0 +1,188 @@
...
+ * 4. then greater starting clearance, then greater travel.
+ * Travel compares raw normalized units across axes (DeviceInfo carries no
+ * viewport to equalize them) — an accepted marginal bias. When every candidate
+ * is sub-floor the safety ranking still decides — the least-bad candidate is
+ * dispatched. Undefined only when no candidate has positive motion — a
+ * violated guard is accepted (attempted), never rejected.
+ */
+export function selectPinchCandidate(candidates: PinchCandidate[]): PinchCandidate | undefined {
+  if (candidates.length === 0) return undefined;
+  const rank = (c: PinchCandidate): number => (c.fullyEdgeSafe ? 0 : c.axisSafe ? 1 : 2);
```
latekvo, line 179: The three-tier safety precedence encoded here (fullyEdgeSafe over axisSafe over neither) is not pinned by a test. Collapsing the middle tier in either direction — ranking axisSafe the same as fullyEdgeSafe, or the same as fully-unsafe — leaves all 38 tests green, because the existing selection tests are decided by the viability / travel / clearance tie-breakers rather than by the tier boundary itself. As a result, a regression that stopped preferring a fully-edge-safe candidate over a higher-travel axis-unsafe one (i.e. a gesture that starts its motion into a system-edge zone) would go undetected.

latekvo (commented):
A couple of test-coverage observations on the pinch geometry — the runtime behavior looks correct; these are about regressions the suite would not catch.

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: Good catch — covered in c3d0ff6a. I added a flow-level test using the suggested tall, narrow bottom-edge frame. It selects angle 90 and asserts centerY 0.775 → endCenterY 0.5, along with the absence of endCenterX.

packages/tool-server/src/tools/flows/flow-pinch-geometry.ts
j-piasecki: Covered in c3d0ff6a. I added direct tests for both safety-tier boundaries: fullyEdgeSafe > axisSafe and axisSafe > unsafe. In each case the losing candidate has better clearance and travel, so the safety rank itself is what decides the result.

packages/skills/skills/argent-create-flow/SKILL.md
latekvo, line 1: This entire SKILL is getting rather large, has loads of bloat, but that's probably a matter for a separate PR.

latekvo (approved):
Nice. Merge when ready ❤️

PR #546, opened 2026-07-22
feat(flow): rotate directive

j-piasecki:
# feat(flow): rotate directive

Adds a selector-targeted two-finger rotation directive to flows: `rotate: { on?, by }`. `on` resolves like `tap` (settled tree + auto-wait); omitted, the rotation centers on the screen. `by` is degrees, positive = clockwise. Distinct from `tool: rotate`, which changes device orientation — the docs and step labels call this out.

## Design

**One continuous gesture, fixed pace, no duration option.** Fingers orbit at ~90°/300 ms; duration derives from the angle. Unlike pinch there is no sub-gesture chaining — any angle dispatches as one sweep, so the angular delta is exact with no settle delays. `by` is parse-bounded to ±3000°: the largest sweep the 10 s single-gesture envelope delivers at that pace, derived from the constants rather than picked as policy — rejected at parse, never silently clamped. A consequence embraced here: frame count always scales with the angle, so a speed-sensitive recording stays a raw `tool:` step (the create-flow skill's polish list covers the conversion).

**Fixed centroid, constant physical radius.** Drifting the centroid would couple pan into the turn; growing the radius would couple pinch. Both stay fixed — the trade is a smaller orbit near edges and small targets, in exchange for a pure rotation.

**Physical circle, not a normalized-space ellipse.** A single normalized radius on a non-square screen traces a physical ellipse, coupling an unintended pinch into every turn. `gesture-rotate` grows optional per-axis `radiusX`/`radiusY` (fractions of width/height with `radiusX·W = radiusY·H`), and the flow tree adapters surface screen dimensions via an optional `screen` field on `DescribeTreeData` so the directive can compute the aspect. Missing dimensions fail soft to the legacy single-radius call rather than erroring.

**Placement reuses pinch's edge model.** The initial finger axis (horizontal or vertical) is scored against the same per-platform system-edge guards, preferring a start parallel to the nearest violated edge and staying inside the target's axis-matching span. Guards only score placement — they never shrink the radius, and a violated guard is attempted, never rejected. The only geometry failure is no positive on-screen orbit radius.

**Abort correctness.** `gesture-rotate` now honors `ctx.signal`: each frame checks it, lifts the synthetic fingers with a terminal Up (so a cancelled run doesn't leave them held on the glass), then throws with name `AbortError`; the directive maps a mid-gesture abort to the run's aborted skip rather than a step failure.

**Explicit `inputSchema`.** The radius requirement is a zod `.refine()` cross-field rule, which `z.toJSONSchema` drops — so the tool ships a derived schema plus an explicit `anyOf` (`radius` or `radiusX`+`radiusY`) to keep invalid calls rejected at the schema layer.

**Platform rejections.** Vega is remote-driven (no touch) and Chromium has no rotate-gesture idiom at all — both fail upfront with guidance to drive the app's rotate controls directly.

## Tests

Unit tests cover the geometry module (duration derivation, candidate construction, aspect handling, edge-guard ranking), the directive end-to-end against a fake device, and `gesture-rotate`'s per-axis radii and abort path.


packages/tool-server/src/tools/gesture-rotate/index.ts
```diff
@@ -1,35 +1,73 @@
...
+  })
+  .refine((p) => p.radius !== undefined || p.radiusX !== undefined, {
+    message: "Pass radius, or both radiusX and radiusY.",
+  });
+
+// Explicit because the auto-derived JSON Schema loses the .refine() cross-field
+// rule — without the anyOf, a call omitting radius/radiusX/radiusY looks valid.
+const inputSchema = {
+  ...zodObjectToJsonSchema(zodSchema),
+  anyOf: [{ required: ["radius"] }, { required: ["radiusX", "radiusY"] }],
```
latekvo: The advertised `anyOf` is looser than the two `.refine()` rules just above it. Its `{ required: ["radius"] }` branch is satisfied by the mere presence of `radius`, so `{ radius, radiusX }` and `{ radius, radiusY }` (a `radius` plus only one of the per-axis values) validate against this JSON Schema — yet the first refine rejects them at runtime, since `radiusX` and `radiusY` have to appear together. `registry.invokeTool` runs `zodSchema.safeParse` before `execute` on every dispatch path, so a client or model that trusts the advertised `inputSchema` will treat those two shapes as valid and then hit an `Invalid params` rejection. The comment above says the `anyOf` restores the cross-field rule that `z.toJSONSchema` drops, but it only encodes "at least one radius spelling is present", not the "radiusX and radiusY together" half of the refine.

packages/tool-server/test/flows/flow-rotate.test.ts
```diff
@@ -0,0 +1,426 @@
...
+
+    const result = await run("rotate-map");
+
+    expect(result.ok).toBe(true);
+    expect(result.calls).toHaveLength(1);
+    const { tool, args } = result.calls[0]!;
+    expect(tool).toBe("gesture-rotate");
+    expect(args.udid).toBe(DEVICE);
+    expect(args.centerX).toBeCloseTo(0.5, 9);
+    expect(args.centerY).toBeCloseTo(0.5, 9);
```
latekvo: Both execution tests assert `centerX` and `centerY` against the same number (0.5): the `Map` frame `{ x: 0.1, y: 0.3, width: 0.8, height: 0.4 }` centres at (0.5, 0.5) and the no-selector case also defaults to (0.5, 0.5). Because the two expected coordinates are identical, these assertions can't tell `centerX` and `centerY` apart if they were transposed in the `runRotate` dispatch — the exact class of mistake they sit here to catch would still pass green. No execution test resolves a selector to a point whose x and y centres differ.

latekvo (commented):
A couple of observations from reviewing the rotate directive, left inline.

packages/skills/skills/argent-create-flow/SKILL.md
```diff
@@ -28,6 +28,7 @@ Beyond raw `tool:` steps and `echo:`, flows support declarative directives inter
...
 | `scroll-to`  | `- scroll-to: "Order #1234"` (scrolls down) or `- scroll-to: { target: …, direction: right, within: … }`                                                             | momentum-free scroll until the target is visible                                                                                                                |
 | `pinch`      | `- pinch: { on: "Map", scale: 3 }` or `- pinch: { scale: 0.5 }`                                                                                                      | two-finger zoom in (`scale` > 1) or out (`< 1`); big scales chain gestures; `on` optional — defaults to screen center; open-loop — assert the visible result    |
+| `rotate`     | `- rotate: { on: "Map", by: 90 }` or `- rotate: { by: -45 }`                                                                                                         | two-finger rotation by degrees (+ CW, − CCW, within ±3000°; options map only); `on` optional — screen center default; not `tool: rotate` (orientation)          |
```
latekvo, line 31: >within ±3000°

~~Genuinely curious, why is it capped at such a weird angle?~~ Nevermind found it. Fair.

packages/tool-server/src/tools/gesture-rotate/index.ts
j-piasecki: Fixed in 5e20f9db — the anyOf now encodes both halves of the refines: `[{ required: ["radiusX", "radiusY"] }, { required: ["radius"], not: { anyOf: [{ required: ["radiusX"] }, { required: ["radiusY"] }] } }]`, so `{radius, radiusX}` / `{radius, radiusY}` are now rejected at the schema layer too, matching what `safeParse` does at dispatch. The schema test was also reworked to validate all eight radius presence combinations against both the advertised anyOf and the zod schema, so the two can't silently drift apart again.

packages/tool-server/test/flows/flow-rotate.test.ts
j-piasecki: Fixed in 862e4361 — the element-targeted test's frame now centres at (0.4, 0.6), so `centerX`/`centerY` differ both from each other and from the no-selector 0.5 default (a single swapped coordinate fails too, not just a full transpose). Sensitivity checked by temporarily transposing the dispatch in `runRotate`: only this test goes red.

latekvo (approved):

latekvo:
Thanks for the PR. I have a bad gut feeling about the `packages/skills/skills/argent-create-flow/SKILL.md` growing in size soo much with these massive oneliner examples, but this should probably be addressed in a separate PR.

j-piasecki:
Yes, I'll open one addressing the skill once all the other directives are ready. (Unless you prefer to do it now).

latekvo:
No worries, feel free to do it whenever suits you best :)

PR #548, opened 2026-07-22
feat(screen-recording): trim static stretches out of the recording

latekvo:
> Stacked on #517 (`feat/screen-recording`). Review/merge that first; this PR's diff is against that branch.

## What

Trims stretches where the screen doesn't change out of the recorded video. A recording left running across waits, slow navigation, or agent thinking time used to come back padded with dead air; now a long session with only brief activity comes back as a short clip.

## How

The frame pump already re-emits the last frame on a wall-clock schedule to keep the timeline honest. This adds a change detector on top:

- Each frame is compared to the last one **byte-exact** (a reference check short-circuits the common "no new frame arrived" case, so only genuine arrivals pay the compare). Exact byte equality is stronger than a hash — no collisions — and flags a change down to a single pixel, matching "something moved, even by a couple of pixels."
- The first second of every still stretch is kept (pauses read naturally), then unchanged frames are dropped until the screen moves again.
- Active stretches still play at real time: the wall-clock pacer is re-anchored whenever a dead stretch is skipped, so the gap yields no output frames rather than a catch-up burst of the next frame.

On by default. `trimStatic: false` on `screen-recording-start` keeps a faithful real-time recording (e.g. to measure how long something took on screen). `screen-recording-stop` now returns the trimmed length as `durationMs`, plus `wallClockMs` and `trimmedMs` when trimming removed anything.

## Verification

- **Unit**: 39 tests in `screen-recording.test.ts` (7 new trim cases: default-on, grace-window kept, no-trim control, resume-the-moment-the-screen-changes, byte-identical-new-frame-stays-trimmed, duration reporting, fields-omitted-when-off). Full tool-server suite green (2817). `tsc --build`, `typecheck:tests`, eslint, prettier clean.
- **E2E on a live tool-server (iOS 18.6 sim)**:
  - 23s idle-heavy session (motion → 15s idle → motion) → **1.47s** clip, `trimmedMs` 21930, with the real transitions intact (44 distinct frames — above the 30-frame grace floor, proving post-grace motion is kept). A `trimStatic:false` control of the same pattern returned the full 23s / 691 frames.
  - 8s continuous burst-motion session → **2.1s** across multiple motion bursts (not collapsed to the 1s floor).
  - **Cross-check**: ran ffmpeg's own `mpdecimate` (duplicate-frame detector) over the same untrimmed footage — it kept ≤ what this trim keeps, confirming the trim is faithful, never over-aggressive (byte-exact is deliberately more conservative than mpdecimate's near-duplicate heuristic, so smooth animation is preserved while only frozen stretches are dropped).


packages/tool-server/src/tools/screen-recording/capture.ts
```diff
@@ -450,11 +509,25 @@ export async function stopCapture(api: ScreenRecordingSessionApi): Promise<StopR
...
       startedAtMs === null ? null : (api.wallClockEndMs ?? Date.now()) - startedAtMs;
-    return { outputFile, sizeBytes: size, durationMs, ...(warning ? { warning } : {}) };
+    // durationMs is the length of the video the caller actually gets. With
+    // trimming that is shorter than the wall clock — it counts only the frames
+    // that survived (each output frame is 1/OUTPUT_FPS of a second).
+    const durationMs = trimStatic
+      ? Math.round((api.framesWritten / OUTPUT_FPS) * 1_000)
+      : wallClockMs;
+    const trimmedMs =
+      trimStatic && wallClockMs !== null ? Math.max(0, wallClockMs - durationMs!) : undefined;
```
hubgan: **Non-blocking.** `wallClockMs`/`trimmedMs` are emitted whenever trimming is *enabled* and the start stamp survived (`trimStatic && wallClockMs !== null`), not only when trimming actually removed frames. On continuously-changing content with no dead air the fields are still present, with `trimmedMs` a small rounding artifact (measured `trimmedMs: 13` on a live continuous-motion capture). That contradicts the documented contract in `screen-recording-stop`, `session-guards`, the tool descriptions, and SKILL.md ("present only when trimming actually removed frames"), which an agent may read as a "was anything trimmed?" signal. The reporting arithmetic itself is correct; only the presence condition diverges from the docs. Non-blocking.

latekvo: [[Diplomat](https://github.com/latekvo/Diplomat)]: Fixed in db8335d1. The trim-only fields are now gated on a `trimmedAnyFrames` flag that the pump sets only when a still stretch actually crosses `STATIC_GRACE_MS` and gets dropped — not on `wallClockMs !== null`, which was true for every trimStatic capture. So a continuously-changing recording (previously surfacing a phantom `trimmedMs` of a frame or two purely from the framesWritten-vs-wall-clock rounding gap) now omits both fields, matching the documented "present only when trimming removed frames" contract.

Verified against the code this points at: added a regression test where a single unchanging frame is held only for a sub-grace window (dead branch never fires) — before the fix `wallClockMs`/`trimmedMs` were present (repro: `expected 800 to be undefined`); after, both are absent. A genuinely trimmed clip still reports both, and the full tool-server suite is green (2818).

packages/tool-server/src/tools/screen-recording/capture.ts
```diff
@@ -450,11 +511,32 @@ export async function stopCapture(api: ScreenRecordingSessionApi): Promise<StopR
...
-    const durationMs =
+    // Wall-clock capture length: after the cap fires (or the encoder dies) the
+    // recording is over even if stop arrives much later.
+    const wallClockMs =
       startedAtMs === null ? null : (api.wallClockEndMs ?? Date.now()) - startedAtMs;
-    return { outputFile, sizeBytes: size, durationMs, ...(warning ? { warning } : {}) };
+    // durationMs is the length of the video the caller actually gets. With
+    // trimming that is shorter than the wall clock — it counts only the frames
+    // that survived (each output frame is 1/OUTPUT_FPS of a second).
+    const durationMs = trimStatic
```
hubgan, line 547: With trimming defaulting on, `durationMs` now reports the trimmed video length rather than wall-clock elapsed time for every recording that doesn't opt out. An existing consumer that read `durationMs` as 'how long the capture ran on screen' will silently get a shorter number unless it passes `trimStatic:false`. Flagging the contract change so downstream readers are aware.

hubgan (commented):
Notes inline.

packages/tool-server/src/tools/screen-recording/capture.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat)]: Intentional contract change — acknowledged, and confirmed no in-repo consumer actually breaks. I traced every reader of the stop result's `durationMs`: the only ones are the screen-recording tests in this PR (updated for the new semantics). Nothing in `argent-private`, the CLI, `argent-mcp`, agent code, or anywhere else reads it; the other `durationMs` occurrences across the tree are unrelated fields (gesture/network/profiler durations).

The semantics are deliberate: `durationMs` is the length of the video the caller actually receives (the registered artifact), so it stays consistent with the file in hand rather than describing a longer clip than exists. When trimming collapses dead air, `wallClockMs` (real elapsed) and `trimmedMs` (how much was cut) are emitted, so a caller that specifically wants on-screen elapsed time still has it; `trimStatic: false` restores the previous wall-clock-only behavior verbatim. This is documented at every point of use — the `ScreenRecordingStopResult` interface + tool description, `StopRecordingFile` in session-guards, the `trimStatic` start-param description, and SKILL.md. Keeping the behavior as-is.

hubgan (approved):

PR #549, opened 2026-07-22
feat(screen-recording): draw touch gestures into the recording

latekvo:
## What

Draws the touch gestures argent drives directly into the screen recording, so a captured video shows **where** each interaction landed instead of a UI reacting to an invisible finger:

- a pulse marks each **tap**
- a fading comet trail follows a **swipe / drag**
- paired markers show the two contact points of a **pinch / rotate**

On by default; pass `showTouches: false` to `screen-recording-start` for a raw, overlay-free capture.

## How

simulator-server already has a touch visualizer behind its `pointer` cargo feature (the binary argent bundles ships with it). It renders the markers into the very MJPEG stream this recording pipeline already consumes — nothing is composited host-side, and it's driven by the touch events argent already sends. This PR just turns it on for the life of a recording:

- `screen-recording-start` builds a small `PointerControl` from the resolved simulator-server handle and hands it to `startCapture`.
- The overlay is enabled right after the encoder goes live (so the first interaction is already drawn) and **restored to off** when the capture ends — stop, time-limit cap, encoder crash, or session teardown all run the same disable path, so it never outlives the recording it was enabled for.
- Two thin client helpers (`setPointerVisible`, `setPointerTrail`) POST to simulator-server's `/api/pointer`. They're best-effort: a build without the feature (or a remote sim) just yields a `warning` at stop and a normal overlay-free video, never a failed recording.

## Verification

- **Unit** — `screen-recording.test.ts`: 49 pass, incl. 10 new covering enable-at-start / disable at stop, cap, unexpected exit, and dispose; the `pointerFailed` warning; the no-control path; and the `/api/pointer` wire format (`{show}` / `{trail}`, ok/error/throw handling).
- **Static** — `tsc` (src + `tsconfig.test.json`), eslint, prettier, full workspace build all clean.
- **E2E** — drove the real tool-server → the bundled (production) simulator-server on an iOS 18.5 sim:
  - `showTouches: true` + a slow swipe → a translucent pointer bubble is rendered onto the screen at the touch point in the recording.
  - identical swipe with `showTouches: false` → paging still happens, **no bubble** — proving both the toggle and that stop restored the overlay to off (the control clip never re-enables it, yet draws nothing).

---

Stacked on #548 (static-frame trimming) → base branch `feat/screen-recording-trim`. Review after that merges.


packages/tool-server/src/tools/screen-recording/capture.ts
```diff
@@ -339,6 +371,15 @@ async function startCaptureLocked(
   registerActiveScreenRecording(api.deviceId, api.wallClockStartMs, params.timeLimitSeconds);
   startPump(api, stream);
 
+  if (params.pointer) {
+    // Arm the touch visualizer before returning, so the very first interaction
+    // is already drawn into the recording. Store the teardown first so a
+    // shutdown racing this await still restores the overlay. Best-effort: a
+    // failure only costs the touch markers, surfaced as a warning at stop.
+    api.pointerDisable = params.pointer.disable;
+    api.pointerFailed = !(await params.pointer.enable());
```
hubgan, line 430: **Non-blocking.** This `await params.pointer.enable()` is the one suspension point after the session is stamped. `enable()` sends the trail and then `setPointerVisible(true)` sequentially, so the `show:true` request goes out last. If `dispose()` (tool-server shutdown / service teardown) runs during this await, it invokes the stored `pointerDisable` (`setPointerVisible(false)`), but that request can be overtaken by enable's later `show:true`, leaving simulator-server's overlay ON after the recording is gone — it then draws touch markers into subsequent non-recording screenshots on that persistent sim-server. Storing the teardown first guarantees disable is attempted, not that it wins the ordering. The window is narrow (shutdown-only, ~two localhost round-trips on a healthy server) and disable succeeds on every non-dispose end path, so flagging non-blocking.

packages/tool-server/src/tools/screen-recording/capture.ts
```diff
@@ -507,6 +551,14 @@ export async function stopCapture(api: ScreenRecordingSessionApi): Promise<StopR
         .filter(Boolean)
         .join(" ");
     }
+    if (pointerFailed) {
```
hubgan, line 592: **Non-blocking.** `pointerFailed` is reset to false only on the stop path, not in `startCaptureLocked`'s reset block (which does reset the sibling `recordingTimedOut` / `recordingExitedUnexpectedly` / `pendingRetrieval`). So when a recording ends via the cap or an encoder crash without a stop, and a subsequent recording is started with `showTouches: false`, that new recording inherits the previous `pointerFailed` and this warning fires for a video that never requested the touch overlay. Cosmetic — a spurious warning with no functional effect, most visible against an older sim-server build where every enable fails. Non-blocking.

packages/tool-server/src/tools/screen-recording/capture.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat)]: Fixed in 2d6db0a9. Confirmed real: with `disable()` unserialized, a `dispose()` racing the `enable()` await issues `show:false` before `enable()`'s still-parked `show:true`, so simulator-server applies `show:true` last and the overlay stays on. The pointer control now lives in `makePointerControl()`, where `disable()` awaits any in-flight `enable()` before sending its `show:false` — so `show:false` is both issued and applied last, and the overlay always ends off.

Added a regression test (`issues disable's show:false only after an in-flight enable's show:true`): it parks `enable`'s `show:true` in flight, races a `disable` in, and asserts the server sees `trail → show:true → show:false`. It fails on the old unserialized `disable` (`show:false` lands first, overlay stuck on) and passes now.

packages/tool-server/src/tools/screen-recording/capture.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat)]: Fixed in 2d6db0a9. Confirmed reproducible: `assertNoActiveRecording` admits a fresh start after a cap or crash (`recordingActive` is false and no pending flag is set), and start's reset block never cleared `pointerFailed` — so a cap-ended failed-enable recording leaves it set, and a following `showTouches: false` recording warns at stop about an overlay it never requested. It is now reset in start's reset block alongside its siblings (`recordingTimedOut` / `recordingExitedUnexpectedly` / `pendingRetrieval`).

Added a regression test (`does not carry a failed enable into a later showTouches:false recording`) that drives a failed-enable recording to end via the cap, then starts an overlay-free recording and asserts no warning at stop. It fails on the old code (stale `pointerFailed`, spurious warning) and passes now.

packages/tool-server/src/tools/screen-recording/capture.ts
```diff
@@ -341,6 +378,15 @@ async function startCaptureLocked(
   registerActiveScreenRecording(api.deviceId, api.wallClockStartMs, params.timeLimitSeconds);
   startPump(api, stream);
 
+  if (params.pointer) {
+    // Arm the touch visualizer before returning, so the very first interaction
+    // is already drawn into the recording. Store the teardown first so a
+    // shutdown racing this await still restores the overlay. Best-effort: a
+    // failure only costs the touch markers, surfaced as a warning at stop.
+    api.pointerDisable = params.pointer.disable;
+    api.pointerFailed = !(await params.pointer.enable());
```
hubgan, line 430: The frame pump is started just above, but the ffmpeg `exit` handler isn't registered until after this `await params.pointer.enable()`. `waitForEncoderReady` has already removed its own exit listener by this point, so during the enable await the live ffmpeg child has no `exit` listener attached, and Node does not replay an `exit` that fired before a listener exists. If ffmpeg dies inside this window (after surviving the 800ms readiness grace), the death goes unobserved: `recordingExitedUnexpectedly`/`pendingRetrieval` are never set, so a later stop returns a truncated mp4 with no warning, and a ride to the cap misattributes the crash as a clean time-limit finalize. The window is normally a couple of localhost fetches but widens to a few seconds against a wedged sim-server. Specific to the default `showTouches:true` path (with it off, this block runs synchronously after readiness).

packages/tool-server/test/screen-recording.test.ts
```diff
@@ -695,6 +716,224 @@ describe("static-frame trimming", () => {
...
+
+    await fs.writeFile(api.outputFile!, Buffer.alloc(64, 1));
+    const outputFile = api.outputFile!;
+    const result = await stopCapture(api);
+
+    expect(result.warning).toBeUndefined();
+    await fs.rm(outputFile, { force: true });
+  });
+
+  it("does not carry a failed enable into a later showTouches:false recording", async () => {
```
hubgan: This test starts a second recording after the cap fires without calling stop first, and asserts `pointerFailed` was cleared. That is exactly the scenario #517's later admission guard (`assertNoActiveRecording` gating on `pendingRetrieval`) rejects. It passes on this branch, but once #517's guard is in the merge base it throws `SCREEN_RECORDING_ALREADY_ACTIVE`, and the `pointerFailed`-reset path it exercises becomes unreachable because that second start is now blocked. Needs reconciling with #517 as the stack is rebased.

hubgan (commented):
Notes inline.

packages/tool-server/src/tools/screen-recording/capture.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat)]: Fixed in 2598f859. Confirmed real with a deterministic repro: parking `enable()` in flight suspends the start in exactly this window, and killing the ffmpeg child there leaves `recordingExitedUnexpectedly` / `pendingRetrieval` unset — the `exit` fires with no listener (readiness removed its own `onExit`, and the handler was registered only after the await), Node never replays it, so a later stop hands back a truncated file with no warning and a ride to the cap misreads the crash as a clean time-limit finalize.

The `exit` handler is now armed right after `startPump`, before the pointer-enable await, so a crash in that window is observed. Added a regression test (`observes an encoder crash that lands during the pointer-enable await`) that parks enable, kills the child during the await, and asserts the recording flips into its unexpected-exit recovery state (`recordingExitedUnexpectedly` / `pendingRetrieval`). It fails on the old ordering (both stay false) and passes now; the full tool-server suite (2831) stays green.

packages/tool-server/test/screen-recording.test.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat)]: Fixed in d162c328. Confirmed exactly as described: I patched `assertNoActiveRecording` to #517's `pendingRetrieval`-gating form (`awaitingRetrieval = api.pendingRetrieval && api.outputFile !== null`, already on `origin/feat/screen-recording`) and the test threw `SCREEN_RECORDING_ALREADY_ACTIVE` at the second start — the cap leaves `pendingRetrieval` set, so the post-cap start is rejected and the reset path goes unreached.

Reworked it so it no longer rides that admission: `clears a stale failed-enable flag at start so a later showTouches:false recording is clean` sets the residual `pointerFailed = true` directly on an idle, startable session (the exact state a cap/crash leaves, minus the retrieval-pending gating), then starts an overlay-free recording and asserts start scrubs the flag and stop carries no spurious touch-visualizer warning. Verified passing under BOTH the current guard and #517's `pendingRetrieval`-gating guard (patched each in and ran it), and still failing if the start-block `pointerFailed = false` reset is removed — so it keeps guarding that reset without depending on the soon-to-be-forbidden path.

hubgan (approved):

PR #551, opened 2026-07-23
feat(screen-recording): save recordings to the project's .argent/recordings

latekvo:
Stacked on #549 (touch visualizer).

## What

Screen recordings now land durably under **`.argent/recordings/`** on the client, instead of the disposable temp-artifact cache — and for a remote `argent link` tool-server they land on the **client** host, not the server.

## Why

Two rough edges in the recording flow:
1. The finished video was only ever written to the OS temp-artifact cache, so it was easy to lose and not obviously part of the project.
2. Over `argent link`, the video was downloaded into that same temp cache; there was no project-scoped home for it.

## How

- New optional `saveDir` hint on the artifact wire contract (`@argent/registry` + `@argent/tools-client`).
- `screen-recording-stop` tags its video with `saveDir: ".argent/recordings"` and drops the internal `argent-` temp prefix from the filename.
- The client materializer honours `saveDir`: **copies** the file when the tool-server is co-located, **downloads** it over `/artifacts/:id` when it is remote. Either way the bytes end up in the durable directory.
- `saveDir` resolves against the client's **project root** (nearest ancestor with `.git`/`package.json`/`.argent`), falling back to the global **`~/.argent`** when run outside a project — matching how the rest of argent scopes per-project config. Adds `findProjectRoot` to `configuration-core` (a null-returning variant of `resolveProjectRoot`) to drive that choice.
- Hardened: absolute paths and `..` escapes are rejected, directory bundles excluded — a tool-server can't steer a write outside the base directory.

## Verification

- Unit: 28 cases in `argent-tools-client/test/artifacts.test.ts` (co-located copy, remote download, project-root anchoring incl. from a subdirectory, global fallback, path-traversal rejection, integrity). configuration-core suite green (55).
- E2E against a booted iOS 18.5 sim + a worktree tool-server, driving the real client materializer path:
  - in a project, co-located server → `<projectRoot>/.argent/recordings/…mp4` (valid h264)
  - in a project, forced remote download (`GET /artifacts/:id`) → same durable dir, temp cache left empty
  - outside a project (HOME redirected) → `~/.argent/recordings/…mp4`
- Root `tsc --build`, prettier, and eslint all clean.

MoQ / `sim-remote` recording support lands in a separate PR stacked on this one.


packages/argent-tools-client/src/artifacts.ts
```diff
@@ -105,6 +116,55 @@ export function artifactDir(deviceId?: string): string {
...
+  const rel = normalize(handle.saveDir);
+  if (
+    isAbsolute(rel) ||
+    rel === ".." ||
+    rel.startsWith(`..${sep}`) ||
+    rel.split(sep).includes("..")
+  ) {
+    return null;
+  }
+  const dir = join(durableBaseDir(), rel);
```
hubgan: **Blocking.** `durableSaveTarget` resolves the wire-supplied `saveDir` against `durableBaseDir()` (the client's project root) and accepts any relative, non-`..` value. It is never constrained to the `.argent/recordings` destination that the only legitimate producer (`screen-recording-stop`) sends — the client trusts whatever `saveDir` arrives on the wire. Because `materializeArtifacts` runs on every tool result, a remote or compromised `argent link` tool-server can attach `saveDir` + a chosen `filename` (sanitized to a single segment) + chosen bytes (`size: 0` skips the length check) to any response and have the client write arbitrary content anywhere under its own project tree.

Verified against this code with the built materializer from a throwaway project: `saveDir: ".git"` with `filename: "config"` overwrites `<projectRoot>/.git/config` (which leads to code execution on the client's next git command), and `saveDir: "."` writes straight into the project root (e.g. over `package.json` or a source file). The `..`/absolute-path rejection above is correct, but it only guards escaping *above* the project root, not writes *inside* it — which is where `.git` and sources live. This is the remote-server write path the PR introduces, so flagging as blocking.

latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Fixed in fab22e55.

You're right — the `..`/absolute guard only blocked escaping *above* the project root, and the base is the project root itself, so any relative non-`..` value (`.git`, `.`, `.argent`) resolved to a writable path inside the tree. I reproduced both vectors against the built materializer: `saveDir: ".git"` + `filename: "config"` + `size: 0` overwrote `<projectRoot>/.git/config` (code execution on the next git command), and `saveDir: "."` wrote into the project root.

The fix stops trusting the wire value: durable destinations are now a **client-side allowlist** (`ALLOWED_SAVE_DIRS`, currently just `.argent/recordings`). The set of directories an artifact may be persisted into is decided on the client, not by whatever the (possibly remote/compromised) tool-server sends; an unlisted `saveDir` degrades to the disposable temp cache. The absolute/`..` structural checks stay as defense in depth. Regression tests now assert `.git`/`.`/`.argent`/`.argent/flags`/… are rejected and that a hostile `.git` result leaves `.git/config` untouched — the same repro that overwrote it now passes with the file unchanged.

packages/argent-tools-client/src/artifacts.ts
```diff
@@ -222,6 +302,44 @@ export async function materializeArtifacts(
...
+                data: await readFile(saveTarget.path),
+                mimeType: value.mimeType,
+              });
+            }
+          } else {
+            const res = await fetchFn(`${ctx.toolsUrl}/artifacts/${value.id}`, {
+              headers: authHeaders,
+            });
+            if (!res.ok) return null;
+            const data = Buffer.from(await res.arrayBuffer());
```
hubgan: The durable download buffers the whole response via `arrayBuffer()` with no size cap, and the integrity check on the next line is skipped when `value.size` is 0. A linked (remote) tool-server that announces `size:0` and streams an unbounded body can drive client memory exhaustion and a persistent disk fill under `.argent/recordings/` - worse than the temp-cache equivalent because it survives cache GC. Confined to the recordings dir (no traversal), but it is attacker-influenceable over `argent link`.

packages/argent-tools-client/src/artifacts.ts
```diff
@@ -222,6 +302,44 @@ export async function materializeArtifacts(
...
+      // file is already local (co-located server), download otherwise — so an
+      // `argent link` recording ends up on the *client* host, not the server.
+      const saveTarget = durableSaveTarget(value);
+      if (saveTarget) {
+        try {
+          await mkdir(saveTarget.dir, { recursive: true });
+          if (localPath) {
+            // Already on this host — copy without buffering the whole file
+            // (recordings can be large); only re-read if it's an inline image.
+            await copyFile(localPath, saveTarget.path);
```
hubgan: Neither the co-located `copyFile` here nor the remote `writeFile` a few lines down guards against an existing destination, so a tool-server that picks a colliding `filename` silently overwrites a recording already present in `.argent/recordings/`. Confined to that directory, but an existing durable file is clobbered with no signal.

packages/tool-server/src/tools/screen-recording/screen-recording-stop.ts
```diff
@@ -64,7 +75,13 @@ Fails if no recording (running or finished-but-unretrieved) exists for the given
     // Resolve the store only after a successful stop — the "no active
     // recording" error path never needs it.
     const artifacts = requireArtifacts(ctx);
-    const video = await artifacts.register(stopped.outputFile, { mimeType: "video/mp4" });
+    const video = await artifacts.register(stopped.outputFile, {
+      mimeType: "video/mp4",
+      // Drop the internal `argent-` temp prefix so the saved file reads cleanly
+      // as `.argent/recordings/screen-recording-<device>-<ts>.mp4`.
+      filename: basename(stopped.outputFile).replace(/^argent-/, ""),
+      saveDir: RECORDINGS_DIR,
```
hubgan, line 85: With this tag, recordings now persist into `<project>/.argent/recordings/` on the client. The `argent-screen-recording` skill still describes the video only as a disposable artifact / temp `hostPath` and never mentions the durable location, so a user won't be prepared for mp4s appearing untracked in their working tree. Noting the doc gap since this PR doesn't touch the skill.

hubgan (commented):
Notes inline.

packages/argent-tools-client/src/artifacts.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Fixed in 045cd8e2.

Confirmed both halves against the built materializer: a handle with `size: 0` + `saveDir: ".argent/recordings"` had its length check skipped (`value.size > 0 && …`) while `arrayBuffer()` buffered the entire body with no cap, so a 50 KB — and by extension arbitrarily large — payload was persisted durably under `.argent/recordings/`.

The durable download now (a) requires a known, positive `size`: a real recording always has one (the store stats the finalized file at register time), so a `size: 0` handle is refused outright and nothing is written; and (b) reads through a bounded reader that rejects early when the declared `Content-Length` exceeds the cap and otherwise aborts the stream the moment the accumulated bytes pass `min(size, MAX_DURABLE_BYTES)` (2 GiB) — so an under-declared body can no longer be buffered whole. Regression tests: the `size: 0` download is dropped with nothing left on disk, and a body that declares 8 bytes but streams 200 KB is capped and resolves to null.

packages/argent-tools-client/src/artifacts.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Fixed in 045cd8e2.

Reproduced: with an existing `.argent/recordings/clip.mp4` on disk, a result tagged `saveDir: .argent/recordings` + `filename: clip.mp4` silently overwrote it via both the co-located `copyFile` and the remote `writeFile`. Durable writes are now exclusive — `copyFile(…, COPYFILE_EXCL)` and `writeFile(…, { flag: "wx" })` — and on `EEXIST` the new file lands alongside as `clip (2).mp4`, `clip (3).mp4`, … so an existing recording is never clobbered, and two concurrent materializations cannot race onto the same name either. Regression test asserts the original bytes survive untouched and the new file takes the `(2)` name.

packages/tool-server/src/tools/screen-recording/screen-recording-stop.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Fixed in 7749fe33.

Good catch — updated the `argent-screen-recording` skill rather than leaving the gap. It now documents that a finished recording is saved durably on the client under `<project>/.argent/recordings/` (or `~/.argent/recordings/` outside a project, and on the client host even for a remote `argent link` server), the `… (2).mp4` no-clobber naming, and that these files are untracked so the user should gitignore or clean them up. Step 5 no longer describes the video as only a disposable temp `hostPath`.

PR #552, opened 2026-07-23
feat(screen-recording): record sim-remote devices over the MoQ video track

latekvo:
## What

Screen recording now works on remote (`sim-remote`) iOS simulators. They used to fail the capability gate: a remote sim has no local HTTP MJPEG stream, because its screen is published as an H.264 (Annex-B) elementary stream over MoQ/WebTransport. This wires that MoQ `video` track into the same recording pipeline the local path uses.

That answers the question that prompted it - "we still have access to the video stream, so what's stopping us from using it the same way as the local one?" Nothing fundamental; the track just wasn't being consumed.

## How

- **`moq-video-stream.ts`** subscribes to the server's `simulator`/`video` track, strips hang's per-frame microsecond-timestamp VarInt off each MoQ frame, and hands the raw Annex-B access units to the capture code. It buffers frames until a consumer attaches so the leading keyframe (SPS/PPS) is never dropped, and waits for the first keyframe before proceeding. That buffer is bounded: past the cap it keeps the head (the parameter sets) and the newest GOP and drops the middle, so what replays is still decodable. A dropped session, a server that stops publishing, and a close mid-connect each fail with the cause rather than leaving the caller to time out.
- **`moq-capture.ts`** pipes those frames into `ffmpeg -f h264`, reusing the local path's encoder and adaptive-contrast watermark backend. There's no Node-side pump: frames are pushed as they arrive, and `-use_wallclock_as_timestamps` plus a constant output rate hold each picture until the next one lands. Under back-pressure it stops writing and resumes at the next keyframe rather than mid-GOP, since a dropped access unit breaks every P-frame that references it. Watermark dimensions come from the keyframe's in-band SPS via ffprobe.
- **Wiring**: `screen-recording-start` routes `ios-remote` to the MoQ path (new `appleRemote` capability on start and stop), and the session blueprint accepts the platform. `stopCapture` is host-side and source-agnostic, so it retrieves the video unchanged.
- **`moq-client.ts`** gains a small `establishMoqSimulator` helper (connection, cert pinning, lease token) shared by the screenshot client and the video stream.

## Parity note

The touch visualizer isn't available over MoQ: the control channel's protobuf has no pointer command, and a `moq+remote://` sim exposes no dialable `/api/pointer`. A requested `showTouches` therefore surfaces a clear warning at stop rather than silently doing nothing. The watermark carries over. Static-frame trimming doesn't apply, since the on-change encoder never sends the duplicate frames trimming collapses. `trimStatic` is documented as ignored here rather than silently dropped.

## Verification

End to end against a locally-built `--features moq` simulator-server driving a booted iOS 18.5 sim:

- `startMoqCapture` then `stopCapture` produces a valid watermarked H.264 mp4 (660×1434), plus the touches-unavailable warning. Watermark render confirmed by extracting a frame.
- Unit tests for the frame helpers (`stripHangTimestamp` across 1/2/4/8-byte VarInts, `isKeyframe` SPS/IDR detection), the buffer trim, the back-pressure resync, the three transport-failure paths driven through a fake MoQ session, and the duration reporting.
- Full tool-server suite green (2940 tests); root `tsc --build`, `tsc -p tsconfig.test.json`, prettier and eslint clean.

Live radon-cloud `sim-remote` verification is out of reach locally, but `openMoqVideoStream(udid)` is a thin shell (a `sim-remote moq-info` lookup) over the transport proven above.

## Timeline caveat

`-fps_mode cfr` holds a picture only up to the next packet, and this transport sends nothing while the screen is still — so a still stretch *between* two changes is reconstructed at its true length, but one that runs to the end of the recording is not: the video ends at the last change. Measured against the real ffmpeg invocation, a recording stopped ~4s after the UI settled came back as a ~1s file.

That is inherent to a pushed H.264 source (there is no frame to re-emit the way the MJPEG pump re-emits its last JPEG), so `stop` reports it instead of hiding it: `durationMs` is the length of the file, not the wall clock, and a warning names the gap. The paced local path is unaffected and still reports wall clock, which its video genuinely covers. Both `SKILL.md` copies previously stated the opposite (remote unsupported, timeline always wall-clock accurate) and are updated.


latekvo:
This has to be tested against a real sim-remote environment before being merged. Thus skipping this PR in 0.17.0, might land in 0.18.0.

latekvo:
Once #587 lands this PR has to be either re-done or moved to Radon. Keeping it open as pending feature reminder

packages/tool-server/src/tools/screen-recording/capture.ts
```diff
@@ -618,6 +650,16 @@ export async function stopCapture(api: ScreenRecordingSessionApi): Promise<StopR
       trimStatic && wallClockMs !== null && api.trimmedAnyFrames
         ? Math.max(0, wallClockMs - durationMs!)
         : undefined;
+    // Say so when the video is materially shorter than the time recorded, so a
+    // caller comparing the two does not read it as a lost or truncated capture.
+    if (lastFrameMs !== null && wallClockMs !== null && wallClockMs - lastFrameMs > 1_000) {
```
latekvo, line 655: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The "screen stopped changing" warning asserts a cause the code already knows is wrong when the MoQ stream dropped.

The new warning is appended whenever the last written frame is more than 1 s before the recording ended, with no guard on why frames stopped. `stopCapture` has already captured the answer 100 lines earlier: `const streamError = api.frameStream?.error ?? api.lastFrameStreamError ?? null;` (capture.ts:546), and appends its own accurate "frame stream … dropped … the video may freeze on its last received frame" at capture.ts:589-601. The MoQ read loop sets that error for a clean track close (`fail(new Error("simulator-server stopped publishing the video track"))`, moq-video-stream.ts:266) and for any read rejection (moq-video-stream.ts:286). When either fires, both warnings are joined into one string and the second states as fact that the screen was still and that "this transport sends nothing while it is still, so the video ends there" — when the screen may have kept changing and the transport died.

**Trigger.** Remote recording of 30 s; the MoQ video track closes or the session resets 5 s in (lease expiry, relay blip). `screen-recording-stop` returns a single `warning` containing both "The frame stream from simulator-server dropped during the recording (simulator-server stopped publishing the video track)…" and "The screen stopped changing 25s before the recording was stopped, and this transport sends nothing while it is still, so the video ends there." — the second is false and points a reader at an idle app instead of a truncated capture.

**Checked against.** capture.ts:655-662 — `if (lastFrameMs !== null && wallClockMs !== null && wallClockMs - lastFrameMs > 1_000) { warning = [warning, `The screen stopped changing … and this transport sends nothing while it is still, so the video ends there.`] … }` — no reference to `streamError`, which is in scope from capture.ts:546 `const streamError = api.frameStream?.error ?? api.lastFrameStreamError ?? null;` and already used at capture.ts:589-601. moq-video-stream.ts:260-267 — `if (!raw) { … fail(new Error("simulator-server stopped publishing the video track")); return; }`.

latekvo (commented):

packages/tool-server/src/tools/screen-recording/moq-capture.ts
```diff
@@ -0,0 +1,328 @@
...
+  api.pendingRetrieval = false;
+  api.lastExitInfo = null;
+  api.outputFile = outputFile;
+  api.logoFile = logoFile;
+  api.watermarkSkipped = watermarkSkipped;
+  // Nothing to trim: the encoder never emits the duplicate frames trimming
+  // collapses, so the stop path must not treat this as a trimmed capture.
+  // `lastFrameWrittenMs` is what keeps its duration honest instead.
+  api.trimStatic = false;
+  api.remoteTouchesUnsupported = params.showTouchesRequested;
```
latekvo, line 295: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The whole MoQ start path is unpinned — deleting its three load-bearing session stamps leaves the suite green.

`startMoqCapture` / `startMoqCaptureLocked` — the module the PR exists to add — has no test at all. Only the pure helper `createAnnexbWriter` is exercised (`test/moq-video-stream.test.ts:1451+`), imported directly from this file. The three assignments that carry the PR's headline behaviour are unpinned:

- `api.trimStatic = false;` (line 294) — what makes `stopCapture` take the `lastFrameMs` branch rather than the frame-count branch;
- `api.remoteTouchesUnsupported = params.showTouchesRequested;` (line 295) — the only writer of the flag that drives the new warning block in `capture.ts:616-623`, which is therefore never reached by any test;
- `api.lastFrameWrittenMs = Date.now();` (line 318) — the only producer of the value the PR's duration-honesty claim rests on. The new duration test (`test/screen-recording.test.ts:1168-1190`) sets `api.lastFrameWrittenMs` by hand on a **local** (MJPEG, `iosDevice`) session, so it pins `stopCapture`'s arithmetic but passes unchanged with `moq-capture.ts` deleted outright.

**Trigger.** Mutation: change `api.trimStatic = false;` to `true` at moq-capture.ts:294, or delete `api.remoteTouchesUnsupported = params.showTouchesRequested;` at 295, or delete `api.lastFrameWrittenMs = Date.now();` at 318 — each is a user-visible regression (wrong durationMs, silently missing touches warning) and the full tool-server suite still passes, because no test ever calls startMoqCapture and no test ever sets api.remoteTouchesUnsupported.

**Checked against.** moq-capture.ts:294-295,318 — `api.trimStatic = false;` / `api.remoteTouchesUnsupported = params.showTouchesRequested;` / `api.lastFrameWrittenMs = Date.now();`. capture.ts:616 — `if (remoteTouchesUnsupported) {` (only consumer, unreached). Compared against the sibling in the same test file: test/screen-recording.test.ts:860-872 `const pointer = fakePointer(false); // sim-server refused the toggle … expect(api.pointerFailed).toBe(true); … expect(result.warning).toContain("touch visualizer could not be enabled"); expect(api.pointerFailed).toBe(false); // reset for the next recording`.

latekvo:
Different approach. This PR will have to be redone

PR #555, opened 2026-07-23
fix(skills): bound sub-agent fan-out in the RN optimization skill

latekvo:
> Draft - please hold off reviewing for now.

## Why

A user reported that on their setup, Claude Code + Argent pushed Opus 4.8 into dispatching 45+ parallel sub-agents when writing mobile app features.

## What was wrong

`argent-react-native-optimization/SKILL.md` told the agent to dispatch one sub-agent per file (Phase 1, lint) and one per checklist item (Phase 2, semantic), with no cap anywhere in any shipped skill.

Phase 1 runs ESLint across the whole source root with a ruleset that fires on nearly every RN component file: `react-native/no-inline-styles`, `react-perf/jsx-no-new-function-as-prop` (every `onPress={() => ...}`), `no-color-literals`, `exhaustive-deps`. So "one per file" expands to roughly one sub-agent per source file in the project - dozens to hundreds in parallel.

This skill is the documented entry point for all performance work, and `rules/argent.md` (`alwaysApply: true`) forces skill loading before any argent tool use, so an ordinary "the app feels slow" prompt reaches it.

Each spawned agent carries its own system prompt plus argent's tool schemas (~9-32K tokens, depending on whether progressive tool loading applies), so a single prompt could burn a large share of a subscription's usage limit in minutes.

## Why only Claude Code hit it

Claude Code has a sub-agent primitive and executes "dispatch one per file" literally. Codex has no equivalent, so the same instruction degraded to inline sequential work and never blew up - exactly the asymmetry that was reported.

## Change

- Batch the findings into at most 4 sub-agents per phase.
- Say explicitly that one-per-file, one-per-finding and one-per-lint-result dispatch is not allowed.

Worst case goes from unbounded (roughly 50-150+ agents) to 8 total.

## Verification

- Measured what `eslint --fix` actually clears for this ruleset, since an earlier revision of this PR claimed autofix would do most of the work. It does not: only `react-native/no-single-element-style-arrays` has a fixer `--fix` applies (`react-hooks/exhaustive-deps` declares one but exposes it as editor suggestions). A real run over a component with inline styles, a colour literal, an index key, arrow/object/array props and an empty dep array went 13 problems -> 11, the single edit being `style={[styles.one]}` -> `style={styles.one}`. The skill and the ruleset reference now say so; the 4-agent cap, not autofix, is what bounds the work.
- `node scripts/grade-skills.mjs`: 10.0/10, all 14 skills pass.
- `prettier --check`: clean.
- Grepped all canonical shipped content under `packages/skills/`: no imperative per-unit dispatch remains, and every surviving mention is a prohibition.
- Confirmed `packages/skills/` is the single git-tracked source (`packages/argent/skills` and `packages/mcp/skills` are gitignored build copies), so this covers the whole class rather than one instance.

### DIFF
--- a/packages/skills/skills/argent-react-native-optimization/SKILL.md
+++ b/packages/skills/skills/argent-react-native-optimization/SKILL.md
@@ -15,7 +15,7 @@ description: Optimizes a React Native app by profiling first to find real bottle
 - **Profile for discovery, not only verification.** Use the profiler to find issues static analysis missed, not only to confirm fixes.
 - **One fix per cycle for architectural changes.** Mechanical batch fixes (inline styles, index keys) can be grouped — re-profile once after the batch. When the measurement involves device interaction, record it as a flow (`argent-create-flow` skill) before the first run so all subsequent cycles replay identical steps.
 - **React Compiler**: if `react-profiler-analyze` reports `reactCompilerEnabled: true`, do NOT propose `useCallback`/`useMemo`/`React.memo` unless you confirmed compiler bail-out via `react-profiler-fiber-tree` (absent `useMemoCache`).
-- **Sub-agents**: Phases 1–2 dispatch sub-agents — one per file for lint results, one per checklist item for semantic. Sub-agents CANNOT touch the device - all profiling and E2E verification must happen in the main agent.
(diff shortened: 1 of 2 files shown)


latekvo:
in the current form this PR adds arbitrary limitations, is not general enough.

PR #557, opened 2026-07-23
fix(boot-device): stop passing -crash-report-mode to emulators that reject it

latekvo:
## Problem

`boot-device` cannot launch an Android emulator on builds that do not accept
`-crash-report-mode`. The emulator rejects the unknown option and exits before
the device ever comes up:

```
$ argent run boot-device --args '{"avdName":"Pixel_6_API_34"}'
emulator binary exited with code 1 before the device booted
```

## Cause

`-crash-report-mode` is undocumented and missing from many emulator builds, so
#306 added a feature-detect gate:

```ts
const crashReportArgs = (await emulatorSupportsFlag("-crash-report-mode"))
  ? ["-crash-report-mode", "never"]
  : [];
```

That commit was a pure addition. It never removed the flag from the shared
`LAUNCH_HARDENING_ARGS` constant, which still reads:

```ts
const LAUNCH_HARDENING_ARGS = [
  "-no-boot-anim", "-netfast",
  "-crash-report-mode", "never",
  "-no-metrics",
] as const;
```

So the gate is dead code. Every launch gets the flag unconditionally through
the constant, and builds that do support it are handed it twice.

## Proof

On Android emulator 36.1.9.0, whose `-help` does not list the flag:

```
$ emulator -version
Android emulator version 36.1.9.0 (build_id 13823996)

$ emulator -help | grep -c -- -crash-report-mode
0
$ emulator -help | grep -c -- -no-metrics
1
```

`-list-avds` and an unknown `-avd` both short-circuit before option
validation, so the rejection only surfaces once a real AVD resolves. Against
one, with everything else held equal:

```
$ emulator -avd Pixel_6_API_34 -check-snapshot-loadable default_boot \
      -crash-report-mode never
INFO    | Found systemPath .../android-34/google_apis/arm64-v8a/
unknown option: -crash-report-mode
please use -help for a list of valid options       # exits, never boots

$ emulator -avd Pixel_6_API_34 -check-snapshot-loadable default_boot
INFO    |      Ok: System requirements to run avd: `Pixel_6_API_34` are met
```

The launch dies at option parsing, after the AVD resolves and before any
device work. `hardeningArgs` reaches the snapshot probe, the hot boot and the
cold boot alike, so on such a build all three spawns abort.

## Fix

- Drop `-crash-report-mode never` from `LAUNCH_HARDENING_ARGS`, leaving the
  `crashReportArgs` gate as the only source of the flag.
- Spread `crashReportArgs` into the `-check-snapshot-loadable` probe's
  `extraArgs`. The probe is itself an emulator spawn, so it gets the same
  crash-dialog suppression as the two boots; once the flag left the constant
  it would otherwise have been the only site missing it. This one is not a
  snapshot-parity requirement - unlike `-noaudio` and `-netfast`,
  `-crash-report-mode` selects crash reporting rather than qemu devices, so a
  mismatch would not invalidate the saved snapshot.
- Leave the gate where it is, after the already-running reuse fast path, so
  the `emulator -help` probe is still skipped when nothing will be spawned.
- Fix the prose that had drifted: the comment above `LAUNCH_HARDENING_ARGS`
  still listed `-crash-report-mode` among its flags and claimed they all need
  no detection, and both it and the `emulatorSupportsFlag` docstring said the
  flag is "only present in newer emulator releases (~36.x and late 35.x)" -
  which 36.1.9.0 not having it contradicts.

`bootAndroidImpl` is the only emulator launch path in the repo, and all three
of its spread sites are covered.

Making the gate load-bearing put weight on a probe that was only ever
advisory, so `emulator -help` detection is hardened alongside it:

- `emulatorSupportsFlag` memoized its verdict for the process lifetime even
  when the probe learned nothing - a transient spawn failure, a timeout, or a
  silent zero-exit run pinned "unsupported" until restart. An empty capture is
  now reported to stderr and left uncached, so the next boot retries.
- A killed probe is no verdict at all: the flag may sit in the part of the
  listing the child never transmitted, so partial output from a timeout is
  discarded rather than read as absence.
- The three emulator probe spawns (`-help`, `-list-avds`,
  `-check-snapshot-loadable`) pass `killSignal: "SIGKILL"`, matching `runAdb`.
  `execFile`'s default SIGTERM can be ignored by an emulator wedged in
  Qt/graphics init, which leaves the promise unsettled past every caller's
  deadline.

## Tests

A new `-crash-report-mode feature gate` block in
`packages/tool-server/test/boot-device-hotboot.test.ts` asserts the real
spawned argv and the probe's `extraArgs`:

- gate false: zero occurrences at the probe, hot-boot and cold-boot sites;
- gate true: exactly one occurrence at each, always followed by `never`. The
  count matters - asserting mere presence passes on the double-pass code and
  would guard nothing;
- plus a test that `-help` detection is skipped entirely on the already-running
  reuse fast path, pinning why the computation sits where it does.

`emulatorSupportsFlag` is stubbed in that file's existing
`vi.mock("../src/utils/adb")` factory, because the real one memoizes its
`-help` result per binary for the process lifetime and both branches could not
otherwise be exercised in one file. The two pre-existing launch-hardening
assertions were updated to the four remaining flags.

`packages/tool-server/test/emulator-supports-flag.test.ts` is new and covers
the detection hardening: which outcomes are memoized (a verdict read from a
non-zero exit's captured listing, and a normal zero-exit run) and which are
not (spawn failure, timeout with empty streams, zero exit with no listing,
timeout with partial output), the stderr warning, and `SIGKILL` on all three
spawns.

Discrimination check - both source changes reverted (flag back in
`LAUNCH_HARDENING_ARGS`, `crashReportArgs` dropped from the probe's
`extraArgs`), tests kept:

```
Tests  5 failed | 26 passed (31)
  x omits the flag from the snapshot probe and the hot-boot spawn when the emulator rejects it
  x omits the flag from the cold-boot spawn when the emulator rejects it
  x passes the flag exactly once to the snapshot probe and the hot-boot spawn when supported
  x passes the flag exactly once to the cold-boot spawn when supported
  x keeps exactly one flag occurrence on the probe-reject cold fallback when supported

  AssertionError: expected 1 to be +0   (x2 - the flag reaches a site that must not see it)
  AssertionError: expected 2 to be 1    (x3 - the flag arrives twice)
```

Fix restored: `Tests 31 passed (31)`.

## Verification

- `npm run build` (`tsc --build`): clean.
- `npx vitest run` in `packages/tool-server`: clean.
- `npm run typecheck:tests --workspaces --if-present`: clean. This is the CI
  gate that plain `tsc` + vitest does not cover.
- `npx prettier --check .` and `npm run lint` (`eslint . --max-warnings 0`):
  clean.
- End to end on a real emulator (36.1.9.0, the build that rejects the flag).
  The gate reads the binary correctly:

  ```
  emulatorSupportsFlag(-crash-report-mode) = false
  emulatorSupportsFlag(-no-metrics)        = true
  emulatorSupportsFlag(-no-boot-anim)      = true
  ```

  and `boot-device` against `Pixel_6_API_34` returns
  `{"platform":"android","serial":"emulator-5554","booted":true}` in 4.8 s.
  The live process argv shows the flag dropped and the four remaining
  hardening flags intact:

  ```
  qemu-system-aarch64 -avd Pixel_6_API_34 -force-snapshot-load -no-snapshot-save \
    -gpu auto -noaudio -no-boot-anim -netfast -no-metrics
  ```

  `-force-snapshot-load` means the hot-boot path ran, so the snapshot probe
  accepted its flag-free argv too.

## Docs

No docs update needed: `-crash-report-mode` is an internal launch argument
that no reference page, config key or tool parameter exposes.


latekvo (commented):

packages/tool-server/test/boot-device-hotboot.test.ts
```diff
@@ -778,3 +774,118 @@ describe("boot-device Android — hot-boot with cold-boot fallback", () => {
...
+    hasSnapshotMock.mockResolvedValue(true);
+    probeMock.mockResolvedValue({ loadable: true, reason: null });
+    mockHappyBootChain();
+
+    const tool = createBootDeviceTool(registry);
+    await tool.execute!({}, { avdName: "Pixel_7_API_34" });
+
+    expect(supportsFlagMock).toHaveBeenCalledWith(FLAG);
+
+    // The probe must carry it too: it changes emulator startup config, and the
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: New test comment gives a snapshot-parity rationale for `-crash-report-mode` that the source comment added in the same diff explicitly denies.

The diff adds two permanent comments that state opposite causal models for why `crashReportArgs` is spread into the `-check-snapshot-loadable` probe. The test says the flag "changes emulator startup config" and that probe/boot argv must agree "or a loadable snapshot gets rejected". The source comment added at boot-device.ts:1093-1096 says the exact opposite: "Unlike LAUNCH_HARDENING_ARGS this one carries no snapshot risk — it selects crash reporting, not qemu devices". The repo's own documented mechanism for probe/boot mismatch agrees with the source, not the test: checkSnapshotLoadable's comment (utils/adb.ts:730-733) says "Renderer-affecting flags (`-gpu auto` etc.) MUST match the boot spawn's argv, or the probe resolves a different renderer and rejects valid snapshots with 'different renderer configured'" — `-crash-report-mode never` is not renderer- or topology-affecting, which is precisely why the source comment carves it out from the LAUNCH_HARDENING_ARGS parity argument.

**Trigger.** A maintainer hits a snapshot-rejection bug or wants to know whether `crashReportArgs` may diverge between probe and spawn. Reading test line 840 they conclude the flag is part of the snapshot-compatibility set and must stay in lockstep; reading boot-device.ts:1094 they conclude it carries no snapshot risk at all. The two comments cannot both be right, and the false one (the test) is the one attached to the assertion that would fire.

**Checked against.** packages/tool-server/test/boot-device-hotboot.test.ts:840-841 — "// The probe must carry it too: it changes emulator startup config, and the\n// probe/boot argv have to agree or a loadable snapshot gets rejected." vs packages/tool-server/src/tools/devices/boot-device.ts:1093-1096 — "Unlike\n// LAUNCH_HARDENING_ARGS this one carries no snapshot risk — it selects crash\n// reporting, not qemu devices" ; and packages/tool-server/src/utils/adb.ts:730-733 — "Renderer-affecting flags (`-gpu auto` etc.) MUST match the boot spawn's argv, or the probe resolves a different renderer and rejects valid snapshots with \"different renderer configured\"."

latekvo: Fixed in 28782e10e — the test comment now states the real reason the probe carries the flag (the probe is itself an emulator spawn, so it gets the same crash-dialog suppression) and explicitly disclaims snapshot parity, matching the source comment and `checkSnapshotLoadable`'s renderer-parity doc. The PR body carried the same falsified claim and was corrected too.

latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Follow-up hardening pushed (d9f6b15, b3cc164, e291622, 6ca7f1b), all gated by the existing suites plus a new `emulator-supports-flag.test.ts`:

- A transient `-help` probe failure that captures no complete listing (spawn error, timeout, silent exit) was memoized as `false` for the process lifetime, silently dropping crash-dialog suppression from every later boot. Empty and killed probes now warn on stderr and stay uncached so the next boot re-probes.
- A killed probe that had captured a *partial* listing hit the same permanent-pin path one boundary over; killed runs are now verdict-less regardless of capture.
- The probe spawns kept Node's default SIGTERM kill signal, which an emulator wedged in Qt init can ignore — the awaited promise then never settles and boot-device hangs. `-help`, `-list-avds` and `-check-snapshot-loadable` now kill with SIGKILL like `runAdb` already did (mechanism reproduced: SIGTERM-immune child leaves execFile unsettled; SIGKILL settles at the timeout boundary).

Full suite: 338 files / 4081 passed, typecheck + prettier clean. One out-of-scope finding from the review pass filed separately as #945.

hubgan (approved):

PR #558, opened 2026-07-23
fix(screenshot-diff): honor an outputDir that does not exist yet on the host

latekvo:
## Problem

Passing `outputDir` to `screenshot-diff` silently did nothing when the directory didn't exist yet. It was never created, the diff images went to a temp dir instead, and nothing in the result said the requested location had been discarded.

Found during an end-to-end pass over the 0.17.0 release: the diff itself was correct, but the requested `outputDir` was never created and no diff PNG existed anywhere under it.

## Cause

`outputDir` is declared as a `kind: "probe"` file input, and `probeHostPath` for that kind is just `stat(path)`. It answers "does this path already exist on the tool-server host", nothing more.

`resolveOutputDir` read `presentOnHost: false` as "this is a remote client's directory":

```ts
if (params.outputDir && (probe === undefined || probe.presentOnHost)) {
  return params.outputDir;
}
// ...otherwise a temp dir
```

But a local agent naming a fresh output directory produces exactly the same `presentOnHost: false`, so it took the remote fallback too.

## Fix

When the probe says the path is absent, try to create it non-recursively. That succeeds only when the parent already exists on this host - true for a local agent picking a new subfolder, false for a remote client's unrelated directory tree - so the remote fallback is preserved. `EEXIST` (raced into existence between probe and create) counts as ours.

It also makes the tool self-consistent: `captureLiveInput` already did `fs.mkdir(params.outputDir, { recursive: true })` when writing live-capture intermediates into that same directory.

## Tests

- `creates and honors an outputDir that does not exist yet on this host` - asserts the diff artifact's `hostPath` is inside the requested directory and that the file is really there. Fails on main, where the artifact lands in a temp dir.
- `falls back to a temp dir when outputDir is not creatable on this host` - parent missing, i.e. the remote case. Passes before and after, pinning the behavior that has to survive.

## Verification

- `vitest run` in `packages/tool-server`: 288 files, 2916 passed, 1 skipped.
- `tsc --noEmit -p tsconfig.test.json`, `tsc --build` across the workspace, and `prettier --check` on both changed files: clean.

The comment on the `fileInputs` block still described the old fallback, so it's updated in the same commit.


packages/tool-server/src/tools/screenshot-diff/index.ts
```diff
@@ -165,12 +166,32 @@ export async function executeScreenshotDiffTool(
  * not flagged absent by the boundary probe (a remote client's local directory).
  * Everything else gets a per-call temp dir; the diff images travel back as
```
latekvo, line 172: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The first paragraph of resolveOutputDir's JSDoc still states the pre-fix rule and contradicts the paragraph added right below it.

The PR updated the stale comment on the `fileInputs` block (lines 82-89) but left the summary paragraph of `resolveOutputDir`'s own JSDoc describing the old rule. It says an outputDir is honored "when it is usable here — i.e. not flagged absent by the boundary probe" and that "Everything else gets a per-call temp dir". After this change, an outputDir that IS flagged absent by the probe is honored whenever `fs.mkdir` succeeds, and it does not get a per-call temp dir — which is exactly what the new second paragraph (lines 170-177) explains. The two paragraphs of the same comment now state incompatible rules, and the summary sentence a reader skims is the wrong one. Same class the PR fixed elsewhere in the same commit, just missed on the function being changed.

**Trigger.** A reader (or a later change) taking the JSDoc summary at face value concludes that `screenshotDiffTool` with `outputDir=/tmp/argent-diff` and `fileInputs.outputDir.presentOnHost === false` writes to a temp dir; the code at lines 184-194 writes to `/tmp/argent-diff` after creating it, as the new test at test/screenshot-diff-tool.test.ts:252 asserts.

**Checked against.** src/tools/screenshot-diff/index.ts:164-168 (unchanged): " * host. An agent-supplied outputDir is honored when it is usable here — i.e.\n * not flagged absent by the boundary probe (a remote client's local directory).\n * Everything else gets a per-call temp dir; ..." vs the same comment's added lines 173-175: "Creating the directory non-recursively separates them: it succeeds only when the parent already exists here..." and the code at 184-194. Compare the sibling comment the PR DID update at lines 85-88: "when the agent-chosen directory doesn't exist on this host, the tool creates it if its parent is already here ... and otherwise falls back to its temp default".

latekvo (commented):

PR #559, opened 2026-07-23
fix(react-profiler): stop summing inclusive durations across grouped components

latekvo:
## Problem

In a `react-profiler-analyze` report, a grouped `×N` row can claim more render time than the commit containing it. From a real profiling session on a React Native app:

```
### Commit #1 — 21.82ms 🟡 (t=15.8s)

- `Login` [React Compiler] — 1ms self, 21.56ms w/children — (mount)
- `View` ×30 — 0.85ms self, 189.26ms w/children — parent re-render
- `AnimatedComponent(View)` ×2 — 0.26ms self, 35.58ms w/children — (mount)
- `ScrollView` ×2 — 0.25ms self, 31.79ms w/children — (mount)
```

A 21.82 ms commit can't contain 189 ms of anything. Reading down that column, `View` looks like the dominant cost in the batch when it's actually 0.85 ms of work.

## Cause

`00-hot-commits.ts` groups fibers by component name within a commit, and accumulated both duration columns the same way:

```ts
existing.totalSelf += e.selfDuration ?? 0;
existing.totalActual += e.actualDuration ?? 0;   // <-- inclusive, not additive
```

`selfDuration` is exclusive, so summing it is right. `actualDuration` is inclusive - self plus the entire subtree it owns. Same-named instances are routinely nested (a `View` inside a `View` inside a `View`), so each ancestor's figure already contains its descendants'. Summing double-counts the same subtree once per level of nesting, which is how 30 `View`s reach 189 ms.

The report header already warned humans not to sum that column ("a parent's inclusive time already contains its children's time") while the code was doing exactly that when building the rows.

## Fix

Take the maximum instance instead of the sum. That's a real subtree cost, and since no fiber's inclusive duration can exceed its commit's duration, a grouped row can no longer overflow its commit. `selfDuration` keeps summing.

The header note now also spells out what a grouped row means: on a `×N` row, `self` is the total across instances and `w/children` is the single most expensive instance's subtree.

## Tests

New `test/react-profiler/hot-commit-grouping.test.ts`, built on three nested `View`s (20 / 15 / 10 ms inclusive, 1 ms self each) inside a 20 ms commit:

- `reports the largest instance's subtree, not the sum, for w/children` - fails on main with `expected 45 to be 20`.
- `never reports a grouped subtree longer than the commit that contains it` - fails on main with `expected 45 to be less than or equal to 20`.
- `keeps summing self-duration, which is exclusive and therefore additive` - passes before and after, pinning the behavior that must not change.

## Verification

- `vitest run` in `packages/tool-server`: 289 files, 2919 passed, 1 skipped.
- `tsc --noEmit -p tsconfig.test.json`, `tsc --build` across the workspace, and `prettier --check` on all three changed files: clean.

## The same bug in three other places

A sweep for other inclusive-duration aggregation found the fix above was one instance of a class, not the whole of it. Each was reproduced by running the tool, using three nested `View`s of 20/15/10ms inside a 20ms commit:

- **`profiler-commit-query` `by_component`** printed `Duration (ms) 45.0` immediately beside `Commit Total (ms) 20.0`. This is the tool the analyze report points the agent to for drill-down.
- **`profiler-commit-query` `by_time_range`** printed `` `View` ×3 45.0ms `` — the same grouped-row shape as the report, opposite meaning, no label saying which metric it is.
- **`react-profiler-renders`** accumulates `actualDuration` by name while walking the live fiber tree, so a `View` inside a `View` counts twice. Running the shipped script verbatim against that tree returns `totalActualDuration: 45`. It was headed "Total (ms)" and the table was ranked by it.

All three now report summed self time next to the largest single instance's subtree, headed so the difference is visible, and ranked by self time (the additive one). `profiler-cpu-query`'s `component_cpu` had the same shape for Hermes CPU frames — one profile node per call depth means aggregating inclusive `totalMs` by function name adds an outer frame to its own inner frame — though its ranking already used exclusive `selfMs`, so only the column was wrong.

`react-profiler-renders` separately reported its instance count as "Renders": it counts what is mounted during one tree walk, so three nested instances read as three renders with nothing having re-rendered.

## Why max, and not the true combined subtree

Summing only the outermost instances would be strictly better than a max, and is not computable at this stage. `flattenProfilingData` iterates `[fiberID, actualDuration]` and drops the ID, and the payload it reads is flat by construction — React DevTools' `CommitDataBackend` carries `fiberActualDurations` / `fiberSelfDurations` as `[id, number]` tuples with no parent, child or depth. Tree structure exists only in the `operations` stream, which argent's bootstrap discards. So two `View` rows are genuinely indistinguishable here: nested versus sibling is unrecoverable, and the largest instance is the best available answer that is a real measurement.

The consequence worth knowing: for a `×N` row this is an order statistic, so it is noisier run to run than an aggregate would be.


packages/tool-server/src/tools/profiler/query/profiler-cpu-query.ts
```diff
@@ -294,20 +294,49 @@ function renderComponentCpu(
...
+    >();
     for (const hs of hotspots) {
-      const existing = aggregated.get(hs.name);
-      if (existing) {
-        existing.selfMs += hs.selfMs;
-        existing.totalMs += hs.totalMs;
+      const seen = perWindow.get(hs.name);
+      if (seen) {
+        seen.selfMs += hs.selfMs;
+        seen.totalMs = Math.max(seen.totalMs, hs.totalMs);
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Per-window fold sums Self but maxes Total, so `component_cpu` now prints rows whose Self exceeds their own Total.

Inside one commit window the new `perWindow` fold does `seen.selfMs += hs.selfMs` but `seen.totalMs = Math.max(seen.totalMs, hs.totalMs)`. `queryCpuWindow` emits one entry per call-tree NODE (it iterates `selfHits`, keyed by nodeId, and only slices the top 50 after sorting), so two DISTINCT, non-nested nodes carrying the same `functionName` — a helper called from two different call sites, which is the ordinary case in a real Hermes profile, not just recursion — both land in the same map key. Their self times add while their inclusive times do not, so the merged row can have `selfMs > totalMs`. Before this PR both columns were summed, so `Total >= Self` always held. The comment added directly beneath (lines 323-330) states this exact shape "is impossible for a single frame and reads as broken output" and uses it as the justification for adding across windows — while the code four lines above produces it within a window. The new test's own header comment (cpu-window-aggregation.test.ts:16-18) repeats the same claim.

**Trigger.** Run `profiler-cpu-query mode=component_cpu component_name=<X>`. In one of X's commit windows, function `parse` appears as node A (5 self-hits) and node B (5 self-hits) under different parents, neither nested in the other. `queryCpuWindow` returns two entries each with selfMs=5, totalMs=5. The fold produces selfMs=10, totalMs=5, and the table prints `| `parse` | 10 | 5 | … |` — a self time twice its own inclusive time.

**Checked against.** profiler-cpu-query.ts:309-322 `for (const hs of hotspots) { const seen = perWindow.get(hs.name); if (seen) { seen.selfMs += hs.selfMs; seen.totalMs = Math.max(seen.totalMs, hs.totalMs); }` vs the comment it sits under, profiler-cpu-query.ts:326-329: "an exclusive column summed against an inclusive column maxed produces rows whose `self` exceeds their own inclusive time, which is impossible for a single frame and reads as broken output". 00-cpu-correlate.ts:139-166 confirms one entry per node, not per name: `for (const [nodeId, hits] of selfHits) { … const name = node.callFrame.functionName; … entries.push({ name, selfMs, totalMs, … }) }`.

packages/tool-server/src/utils/react-profiler/pipeline/05-render.ts
```diff
@@ -194,9 +194,12 @@ function buildFullMarkdown(
     `${compilerLine}  **Hot commits:** ${totalHotCommits} of ${input.reactCommits} total`,
```
latekvo, line 148: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: All-clear report still tells the agent `react-profiler-renders` gives render counts and to memo high-count components, which this PR made false.

`renderAllClear` still emits "Run `react-profiler-renders` for a live view of render counts — components with high counts may benefit from `React.memo`". This PR rewrote that tool so the count column is `instanceCount` (mounted fiber instances from one tree walk), and its own new description says so explicitly: "the instance count is how many exist — not how many times they re-rendered". A high count is now `View ×30` on screen, which says nothing about re-rendering, so the suggestion points the agent at a number and then prescribes `React.memo` on the basis of a reading the tool no longer supports. The same stale claim survives in three untouched skill docs the PR did not sweep (packages/skills/skills/argent-react-native-profiler/references/diagnostic-tools.md:11 "Returns render counts and durations per component", .../argent-react-native-profiler/SKILL.md:19 "Live fiber walk: render counts + durations", .../argent-react-native-optimization/SKILL.md:9 "for a live render count table").

**Trigger.** Profile a session where every React commit is under 16ms. `renderProfilingReport` takes the `input.allClear` branch (05-render.ts:80-90) and emits `renderAllClear`, whose Suggested Improvements block prints this line. The agent then runs `react-profiler-renders`, reads `Instances 30` for `View`, and applies the memo advice the report gave.

**Checked against.** 05-render.ts:167 `` `- Run \`react-profiler-renders\` for a live view of render counts — components with high counts may benefit from \`React.memo\`${…}` `` against the tool's post-PR description, react-profiler-renders.ts:124: "It is a snapshot of the tree as it stands, not a recording, so the instance count is how many exist — not how many times they re-rendered." and react-profiler-renders.ts:31-32 `results[name] = { instanceCount: 0, … }; results[name].instanceCount += 1;`

latekvo (commented):

packages/tool-server/test/react-profiler/cpu-window-aggregation.test.ts
```diff
@@ -0,0 +1,112 @@
...
+ * own inner frame's, so those cannot.
+ *
+ * Getting the first axis wrong is visible in the output: an exclusive column
+ * summed against an inclusive column maxed prints rows whose `Self` exceeds
+ * their own `Total`, which no single frame can do.
+ */
+const { renderComponentCpu } = __testables;
+
+/** One flat node: `parse` called under a root, no recursion. */
+function flatIndex(sampleTimesMs: number[]): CpuSampleIndex {
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The new CPU test cannot fail on the un-patched code — its fixture never exercises the per-window fold the PR added.

`flatIndex` builds exactly one non-root node named `parse` (its own docstring: 'One flat node: `parse` called under a root, no recursion'), so `queryCpuWindow` returns exactly one hotspot per window and the new `perWindow` map never hits its collision branch — the only behaviour the PR changed in this file. Hand-evaluating both windows: window [0,5] has 5 samples, avgIntervalMs = 5/5 = 1, so `parse` is selfMs 5 / totalMs 5; window [10,15] is identical. Pre-PR (sum, sum) gives 10/10; post-PR (max within window, sum across windows) gives 10/10. All three assertions — self <= total (10<=10), both columns ≈ single*2 (10 ≈ 5*2), total <= commit total (10<=10) — hold identically on the un-patched code. The only axis the tests pin, summing across disjoint windows, is the axis this PR did not change; `__testables` was exported from profiler-cpu-query.ts solely for this file, leaving that file's actual change with zero coverage.

**Trigger.** Revert the profiler-cpu-query.ts hunk (restore the single `for (const hs of hotspots)` loop that summed both columns into `aggregated`) and re-run `vitest run test/react-profiler/cpu-window-aggregation.test.ts`: all three tests still pass, because with one node per name per window the two code paths compute the same numbers.

**Checked against.** cpu-window-aggregation.test.ts:22-49 — `/** One flat node: `parse` called under a root, no recursion. */` with `nodeMap.set(2, {... functionName: "parse" ...})` and `sampleNodeIds: sampleTimesMs.map(() => 2)`; the file header itself says 'Within a window the same function can arrive as several call-tree nodes ... so those cannot [be added]', and no test constructs that shape. Contrast test/react-profiler/hot-commit-grouping.test.ts:28-32, whose `nestedViews` fixture does produce the collision (three `View` entries) and whose `expect(view!.actualDurationMs).toBe(20)` fails on main with 45.

packages/tool-server/src/tools/profiler/query/profiler-commit-query.ts
```diff
@@ -112,27 +112,33 @@ function renderByComponent(
...
+      // single instance is a real subtree cost and cannot exceed the commit.
+      maxSubtree: entries.reduce((m, e) => Math.max(m, e.actualDuration), 0),
+      totalSelf: entries.reduce((s, e) => s + (e.selfDuration ?? 0), 0),
       commitDuration: entries[0]!.commitDuration,
       timestamp: entries[0]!.timestamp,
       reason: formatReason(entries[0]!),
       parentName: entries[0]!.parentName ?? "—",
     }))
-    .sort((a, b) => b.totalDuration - a.totalDuration)
+    .sort((a, b) => b.maxSubtree - a.maxSubtree)
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `by_component` ranks and truncates commits by the max-instance order statistic while every sibling in this PR switched to the additive metric.

`renderByComponent` now sorts by `maxSubtree` (the largest single instance's inclusive time) before `.slice(0, topN)`. Every other grouping site touched by this PR moved its ranking onto the additive/exclusive metric: `getTopComponents` in the same file sorts by `b[1].totalSelf`, `buildHotCommitSummaries` sorts by `b[1].totalSelf`, and `react-profiler-renders` was changed to sort by `selfBaseDuration_ms`. The PR body states the rule for all of them — "All three now report summed self time next to the largest single instance's subtree, headed so the difference is visible, and ranked by self time (the additive one)" — which is false for `by_component`. The body also notes the max is "an order statistic, so it is noisier run to run than an aggregate would be", so this is the one place where the truncation set is decided by the noisy figure.

**Trigger.** `profiler-commit-query mode=by_component component_name=View` on a session where `View` rendered in more than `top_n` (default 20) commits. Commit A has one `View` instance at 50ms inclusive / 0.5ms self; commit B has 30 `View` instances totalling 40ms self with the deepest at 12ms inclusive. A is ranked above B, and if the list is truncated the commit where `View` actually did the most work is the one dropped.

**Checked against.** profiler-commit-query.ts:127 `.sort((a, b) => b.maxSubtree - a.maxSubtree)` followed by `.slice(0, topN)` on line 128, against the sibling in the same file, profiler-commit-query.ts:341 `.sort((a, b) => b[1].totalSelf - a[1].totalSelf)`, and 00-hot-commits.ts:115 `.sort((a, b) => b[1].totalSelf - a[1].totalSelf)`, and react-profiler-renders.ts:212 `.sort((a, b) => b.selfBaseDuration_ms - a.selfBaseDuration_ms)`.

packages/tool-server/src/tools/profiler/query/profiler-cpu-query.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Confirmed and fixed in 14bac0452.

Reproduced first: a fixture with two distinct `parse` nodes under different parents alternating samples inside one window printed `| `parse` | 10 | 5 |` — self twice its own inclusive time.

The fix merges same-name rows by call-tree ancestry instead of blindly maxing: `queryCpuWindow` now emits `nodeId` on each hotspot, and the per-window fold walks `childToParent` (`nodesShareCallTree`). Disjoint nodes are separate subtrees, so both columns add; nested frames keep the larger inclusive figure. Self always adds either way, since even nested frames own disjoint sample intervals.

Coverage added for both branches in cpu-window-aggregation.test.ts: the sibling fixture fails on the old max-always fold (`expected 10 to be close to 5`), a recursive `walk` fixture fails on naive summing (`expected 12 to be close to 7`). Both mutations verified killed.

packages/tool-server/src/utils/react-profiler/pipeline/05-render.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Confirmed and fixed in 14bac0452. The all-clear line no longer calls the instance count a render count or prescribe `React.memo` from it; it now says the instance column counts what is mounted — not re-renders — so memo decisions should come from commit data. The three stale skill-doc lines swept too: argent-react-native-profiler/references/diagnostic-tools.md:11, argent-react-native-profiler/SKILL.md:19, argent-react-native-optimization/SKILL.md:9.

packages/tool-server/test/react-profiler/cpu-window-aggregation.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Confirmed and fixed in 14bac0452. Reproduced exactly as described: restoring the pre-PR single summing loop left the shipped test at 3/3 green. The file now has fixtures that actually exercise the collision branch — two disjoint same-name nodes in one window (fails on max-always: total comes out 5 where both columns must read 10) and nested recursive same-name frames (fails on naive summing: total reads 12 where the outer subtree is 7).

packages/tool-server/src/tools/profiler/query/profiler-commit-query.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Confirmed and fixed in 14bac0452: `renderByComponent` now ranks and truncates by `totalSelf`, matching `getTopComponents`, `buildHotCommitSummaries` and `react-profiler-renders`. `maxSubtree` stays on the row as the reported inclusive figure; it no longer decides which commits survive the cut.

packages/tool-server/src/tools/profiler/query/profiler-cpu-query.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Follow-up: a second review pass found the 14bac0452 fold still order-dependent for 3+ same-name nodes mixing nesting AND disjointness in one window (pairwise ancestry against the first-seen row does not compose; reproduced printing `| `walk` | 10 | 13 |` inside a 10ms commit). Refined in dfd652ace: each row now tracks its set of disjoint subtree roots — a newcomer contained by a member adds nothing inclusive, one containing members absorbs them, anything else joins the union. Regression tests pin both adversarial arrival orders.

latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Fresh swarm review pass over the full diff (claims vs code, nearest twin, non-happy paths, inputs, reachability both ways, what outlives the call, plus the absence sweep) — four agents, all sibling aggregation sites swept.

**Confirmed and fixed in db346027b**, each reproduced before fixing and re-run after:

- `by_index` crashed on dump-read durations the grouped views already guard: `renderByIndex` called `.toFixed()` on raw `actualDuration`/`selfDuration` (`TypeError` on a JSON-nulled duration), and `renderCascadeTree` summed an absent `selfDuration` into `self=NaNms`. Reproduced with fixtures feeding `null` / a deleted key through the same `readCommitTree` path; both now print 0 like their siblings. Regression coverage in commit-query-null-durations.test.ts.
- `RESOLVE_FIBER_META_SCRIPT` accumulated debuggee-derived display names into a plain `{}`, so fibers named `__proto__` or `constructor` failed the membership check and were silently skipped — same class COLLECT_RENDERS_SCRIPT was hardened against. Reproduced under `node:vm` (both rows vanished pre-fix); now collects into a null-prototype object. fiber-meta-script-safety.test.ts.
- The grouping comment claimed a bare *null* prints NaN or throws at toFixed; empirically null coerces harmlessly through both reduces while an *absent* key makes them NaN. Comment now states the real mechanism, matching its own test.
- tools.mdx still introduced `react-profiler-renders` as collecting "component render counts"; row updated to mounted instances. Docs build passes.
- Two test-header comments restated the pre-change bug in past tense; now present tense. The CPU fold comment now notes the window's 50-node self-time cap so its completeness claim matches what the code does.

**Out of scope, filed:** #949 (top_functions/time_window and cpu-summary print one row per call-tree node with overlapping totals, truncating per node before any merge), #950 (deserializeCpuSampleIndex yields all-NaN figures for a v2 index missing intervalStartsMs).

Verification: vitest 16 files / 141 tests green including the three new repro-derived suites; `tsc --build packages/tool-server` and tsconfig.test.json clean; prettier clean on all changed files; `npx docusaurus build` succeeds.

latekvo:
[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Second swarm review pass over the updated branch (claims vs code, nearest twin, non-happy paths/inputs, reachability both ways, what outlives the call, absence sweep) — four agents.

**Confirmed MEDIUM ×2, both reproduced before fixing and re-run after, fixed in ca18bbcb:**

- `cascade_tree` infinitely recursed on mutually recursive components: A renders B renders A puts a cycle in the name→parentName edges and the `${name}:${depth}` dedupe key never repeats, so the walk descended until `RangeError: Maximum call stack size exceeded`. Reproduced with a three-fiber fixture through `renderCascadeTree`; now tracks the recursion path and skips names already on it. Non-cyclic chains render unchanged (regression-tested).
- The analyze pipeline itself consumed dump-read durations unguarded: `react-profiler-analyze` reads its commit tree back from disk via bare `JSON.parse`, and `JSON.stringify` drops undefined properties — so an absent `selfDuration` key reached `01-reduce`'s `sum += duration` and poisoned mean/totalRenderMs into the report as NaN. Reproduced end-to-end through `readCommitTree` → `preprocess` → `reduce` (`mean=NaN` pre-fix, `mean=1` post); guarded like every other dump-read duration in this PR.

**Confirmed LOWs fixed in the same commit:** component_cpu built commit windows from unguarded dump-read `timestamp`/`commitDuration` (zero-width windows → misleading "no CPU samples"); `react-profiler-renders` called `.toFixed` on durations crossing the debuggee JSON boundary where non-finite serializes to null (`TypeError`); the hook-missing retry path skipped the `exceptionDetails` re-check so a failing retry reported the wrong error code; `COLLECT_RENDERS_SCRIPT` had no semantic pin of the max-not-sum grouping for nested same-name fibers (coverage gap — added, fails on naive summing with `expected 3 to be 20`-style assertions... concretely: summed would read 45 vs asserted max 20).

All mutations verified killed (guards reverted → tests fail; path guard removed → RangeError returns). Full tool-server suite green (4525+ passed; one unrelated macOS platform-gate test timed out under parallel load and passes in isolation), `tsc` clean, prettier clean.

Issue #547, opened 2026-07-22
iOS sim gesture-tap reports success while touches silently fail to land; not recoverable via Argent's own tools


josh150:
## Summary

On an affected simulator, `gesture-tap` and `gesture-swipe` return `{ tapped: true }` but the synthesized touch is **never delivered** — the UI does not respond. Only HID/touch injection is affected: `launch-app`, `describe`, and `screenshot` all continue to work correctly on the same device.

The failure is **per-device**, **persists across Argent's own recovery paths** (`stop-simulator-server`, `boot-device --force`), and is cleared only by restarting the host `CoreSimulatorService` daemon. It is **not tied to a simulator OS version** (reproduced on two different iOS runtimes) and **not app-specific** (reproduced in Apple's own Settings app).

The core problem is the **false-positive success**: `tapped: true` is returned for taps that never happened, so automated flows/agents silently pass or fail with meaningless results.

## Environment

- **Argent:** 0.16.1 (global npm install, Node 22.20.0 via nvm)
- **Host OS:** macOS 27 (Darwin 27.0.0)
- **Xcode / CoreSimulator:** Xcode 27 beta
- **Simulators:** booted headless via Argent's `simulator-server` / DeviceHub (no Simulator.app windows)
- **Runtimes on which the symptom was observed:** iOS 18.6 and iOS 26.4 (the issue is not specific to either)

## Actual behavior

1. `launch-app` succeeds; `screenshot` (framebuffer) and `describe` (accessibility) return correct, current data.
2. `gesture-tap` on any control returns `{ tapped: true }`, but the screen does not change and no handler fires.
3. `gesture-swipe` behaves the same — reported as performed, but no scroll/drag occurs.

## Expected behavior

Either the touch is delivered, **or** `gesture-tap`/`gesture-swipe` reports that HID delivery failed — rather than optimistically returning `tapped: true`.

## Steps to reproduce (observed)

1. Boot a simulator and `launch-app`.
2. `gesture-tap` a known-interactive element (e.g. a tab-bar button) → returns `{ tapped: true }`, no effect.
3. Repeat in a second app, including the system **Settings** app (tapping a settings row is a no-op) — confirming it is not specific to the app under test.

**Correlation:** the affected device had earlier been driven via a **manual Xcode debug session** (debugger attach). Our working theory is that this leaves a stale per-device HID state inside `CoreSimulatorService`. Devices that were never Xcode-attached did not exhibit the problem.

## What we ruled out

[...]

PR #563, opened 2026-07-24
feat(tool-server): detect and recover wedged iOS simulator touch injection

L65FREAD:
Addresses #547.

Some iOS simulators silently wedge touch injection (typically after an Xcode debug session): `gesture-tap`/`gesture-swipe` return `{ tapped: true }` while nothing lands, and neither `stop-simulator-server` nor `boot-device --force` clears it.

- **Delivery verification** — the first touch per device session is frame-diff checked; a dropped iOS touch surfaces `verified: false` + a `warning` instead of a false success (`verify: true`/`false` to force/skip).
- **`recover-touch-injection`** — restarts the host `CoreSimulatorService` daemon and restores every previously-booted simulator.

Frame-diff is a heuristic, so it's a warning (never a hard failure) and only points at recovery once a no-change repeats.


packages/tool-server/src/utils/coresimulator-recovery.ts
```diff
@@ -0,0 +1,149 @@
...
+      });
+    }
+  };
+
+  const previouslyBooted = rebootAfter ? await listBootedDevices(exec) : [];
+
+  await run("shutdown-all", "xcrun", ["simctl", "shutdown", "all"], { tolerateFailure: true });
+  await run("killall-coresimulatorservice", "killall", [CORE_SIMULATOR_SERVICE], {
+    tolerateFailure: true,
+  });
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `shutdown-all` and `killall-coresimulatorservice` are the only two steps that actually restart the daemon, and both are hard-coded `tolerateFailure: true`, so `recoverySucceeded()` (`steps.every(s => s.ok)`) can never observe either of them fail.

On the `rebootAfter: false` path they are the only steps that run at all, which makes `recovered` a structural constant `true` — the tool reports success even when every command failed, and the `Recovery did not complete` branch of `note` is unreachable there. On the default path `recovered` only tells you whether the *target booted*, not whether the daemon was restarted: a `killall` that fails for a real reason (SIP/permissions, or the service name changing after an Xcode update) is indistinguishable from the intended "no daemon to kill" case, so the agent is told the wedge was cleared while it is still there and goes back to tapping a dead simulator. That is the same false-success shape this tool exists to eliminate, and the tool description states the opposite contract: "recovered is false when a recovery step failed (see steps)".

Reproduction — from `packages/tool-server` after `npm run build`, nothing is executed (the injectable `exec` rejects every command):

```js
import { recoverCoreSimulatorInjection, recoverySucceeded } from "./dist/utils/coresimulator-recovery.js";
const failAll = async (f, a) => { throw new Error(`${f} ${a.join(" ")}: not permitted`); };
const steps = await recoverCoreSimulatorInjection("AAAAAAAA-BBBB-CCCC-DDDD-EEEEEEEEEEEE", { rebootAfter: false, exec: failAll });
console.log(steps.map(s => `${s.step} ok=${s.ok}`), "recovered =", recoverySucceeded(steps));
```

```
[ 'shutdown-all ok=true', 'killall-coresimulatorservice ok=true' ] recovered = true
```

packages/tool-server/src/tools/simulator/recover-touch-injection.ts
```diff
@@ -0,0 +1,118 @@
...
+WARNING: this is host-wide — ALL booted simulators are briefly shut down; the target and any other previously-booted simulators are then re-booted automatically (the target is waited on, siblings finish in the background). Reach for it only when a device's taps have stopped landing; confirm the wedge first with gesture-tap { verify: true }.
+
+iOS simulators only. Returns { recovered, steps, disposedServices, note } — recovered is false when a recovery step failed (see steps).`,
+    searchHint:
+      "recover fix wedged stuck touch injection tap not landing coresimulator daemon restart reset simulator hid silent failure",
+    zodSchema,
+    capability,
+    services: () => ({}),
+    async execute(_services, params) {
+      const device = resolveDevice(params.udid);
```
latekvo, line 168: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `resolveDevice` classifies by string shape only — `device-info.ts` maps any 8-4-4-4-12 hex string to `platform: "ios", kind: "simulator"` with no `simctl` round-trip — so the capability gate accepts a UDID that names no simulator on this host, and `execute` goes straight into `recoverCoreSimulatorInjection` without ever checking the device exists.

A typo'd, stale, or copy-pasted-from-another-machine UDID therefore runs `xcrun simctl shutdown all` and `killall com.apple.CoreSimulator.CoreSimulatorService` host-wide before anything can fail: every booted simulator on the machine is shut down — including ones owned by other agents, by a second tool-server instance, or by a running Xcode debug session — and only then does `simctl boot` fail and the tool return `recovered: false`. The destructive half has already happened and there is no way to undo it. `boot-device` handles the same class of input the other way round: it resolves the UDID through `listIosSimulators()` before acting on it.

Reproduction — from `packages/tool-server` after `npm run build`; nothing is executed, the injectable `exec` only records what would run:

```js
import { resolveDevice } from "./dist/utils/device-info.js";
import { assertSupported } from "./dist/utils/capability.js";
import { createRecoverTouchInjectionTool } from "./dist/tools/simulator/recover-touch-injection.js";
import { recoverCoreSimulatorInjection } from "./dist/utils/coresimulator-recovery.js";

const BOGUS = "DEADBEEF-0000-0000-0000-000000000000";       // valid shape, no such simulator
const tool = createRecoverTouchInjectionTool({ getSnapshot: () => ({ services: new Map() }), disposeService: async () => {} });
const dev = resolveDevice(BOGUS);
console.log("resolveDevice   =", JSON.stringify(dev));
assertSupported("recover-touch-injection", tool.capability, dev);   // does not throw
console.log("capability gate = ACCEPTED");

const issued = [];
await recoverCoreSimulatorInjection(BOGUS, { exec: async (f, a) => {
  issued.push(`${f} ${a.join(" ")}`);
  if (a.includes("list")) return { stdout: JSON.stringify({ devices: { rt: [{ udid: "SOMEONE-ELSES-SIM", state: "Booted" }] } }), stderr: "" };
  if (a.includes("boot") || a.includes("bootstatus")) throw new Error("Invalid device: " + BOGUS);
  return { stdout: "", stderr: "" };
}});
issued.forEach((x, i) => console.log(`  ${i + 1}. ${x}`));
```

```
resolveDevice   = {"id":"DEADBEEF-0000-0000-0000-000000000000","platform":"ios","kind":"simulator"}
capability gate = ACCEPTED
  1. xcrun simctl list devices booted -j
  2. xcrun simctl shutdown all
  3. killall com.apple.CoreSimulator.CoreSimulatorService
  4. xcrun simctl boot DEADBEEF-0000-0000-0000-000000000000
  5. xcrun simctl boot SOMEONE-ELSES-SIM
  6. xcrun simctl bootstatus DEADBEEF-0000-0000-0000-000000000000 -b
```

Steps 2 and 3 — the host-wide part — run before anything establishes that the requested device is real.

packages/tool-server/src/tools/simulator/recover-touch-injection.ts
```diff
@@ -0,0 +1,118 @@
...
+      const failedSteps = steps.filter((s) => !s.ok).map((s) => s.step);
+      const restoredSiblings = steps.filter((s) => s.step.startsWith("boot:")).length;
+
+      const note = !recovered
+        ? `Recovery did not complete: step(s) ${failedSteps.join(", ")} failed — see steps for details. ` +
+          "The device may still be shut down; boot it with boot-device, or retry this tool."
+        : rebootAfter
+          ? "CoreSimulator daemon restarted and the target device re-booted" +
+            (restoredSiblings > 0
+              ? `; ${restoredSiblings} other previously-booted simulator(s) were also restarted (finishing in the background)`
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `restoredSiblings` counts every step whose name starts with `boot:`, which is the number of sibling boots *attempted*, not the number that succeeded. Sibling boots run with `tolerateFailure: true`, so a step that failed still carries `ok: true` and is still counted here.

When every sibling fails to come back — a realistic outcome, since `simctl boot` right after `shutdown all` commonly returns `Unable to boot device in current state: Shutting Down` — the note still tells the agent that N simulators "were also restarted (finishing in the background)" and `recovered` stays `true`. The user is told their other simulators were restored and only finds out otherwise later.

Reproduction — from `packages/tool-server` after `npm run build`:

```js
import { recoverCoreSimulatorInjection, recoverySucceeded } from "./dist/utils/coresimulator-recovery.js";
const booted = { devices: { rt: [{udid:"T",state:"Booted"},{udid:"SIB-1",state:"Booted"},{udid:"SIB-2",state:"Booted"}] } };
const steps = await recoverCoreSimulatorInjection("T", { exec: async (f, a) => {
  if (a.includes("list")) return { stdout: JSON.stringify(booted), stderr: "" };
  if (a.includes("boot") && (a.includes("SIB-1") || a.includes("SIB-2"))) throw new Error("Unable to boot device in current state: Shutting Down");
  return { stdout: "", stderr: "" };
}});
console.log(steps.filter(s => s.step.startsWith("boot:")).map(s => `${s.step} ok=${s.ok} tolerated=${s.tolerated}`));
console.log("restoredSiblings the note reports:", steps.filter(s => s.step.startsWith("boot:")).length, "| recovered =", recoverySucceeded(steps));
```

```
[ 'boot:SIB-1 ok=true tolerated=true', 'boot:SIB-2 ok=true tolerated=true' ]
restoredSiblings the note reports: 2 | recovered = true
```

packages/tool-server/src/utils/coresimulator-recovery.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+
+async function listBootedDevices(exec: ExecFn): Promise<string[]> {
+  try {
+    const { stdout } = await exec("xcrun", ["simctl", "list", "devices", "booted", "-j"]);
+    return parseBootedUdids(stdout);
+  } catch {
+    return [];
+  }
+}
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `listBootedDevices` swallows every failure of `xcrun simctl list devices booted -j` (and `parseBootedUdids` separately swallows unparseable stdout) and returns `[]` without pushing a `RecoveryStep`.

Because the snapshot runs *before* `simctl shutdown all`, a failure here means the sibling simulators are still shut down but no `boot:<udid>` step is ever issued to bring them back. Nothing in `steps` records that the snapshot failed, `recovered` is `true`, and the note reports a clean recovery. The tool description promises it "restarts every other simulator that was booted"; on this path it silently does not, and the agent gets no signal to act on. A `simctl` call that hangs against the very daemon this tool is being run to fix is exactly the condition in which the snapshot is most likely to fail — and it shares the 10s `SIMCTL_SPAWN_TIMEOUT_MS` default with the other non-boot steps.

Reproduction — from `packages/tool-server` after `npm run build`:

```js
import { recoverCoreSimulatorInjection, recoverySucceeded } from "./dist/utils/coresimulator-recovery.js";
const steps = await recoverCoreSimulatorInjection("T", { exec: async (f, a) => {
  if (a.includes("list")) throw new Error("xcrun: timed out after 10000ms");
  return { stdout: "", stderr: "" };
}});
console.log(steps.map(s => s.step), "| recovered =", recoverySucceeded(steps));
```

```
[ 'shutdown-all', 'killall-coresimulatorservice', 'boot', 'bootstatus' ] | recovered = true
```

No `boot:*` step, and no record anywhere that the snapshot was lost.

packages/tool-server/src/utils/touch-verification.ts
```diff
@@ -0,0 +1,193 @@
...
+): string | null {
+  switch (verdict) {
+    case "landed":
+      return null;
+    case "no-change":
+      return (opts.recommendRecovery ?? true)
+        ? "Touch was injected but the screen did not change — on an iOS simulator this can mean " +
+            "touch injection has wedged (it accepts touches but silently drops them). If taps have " +
+            "stopped landing, run recover-touch-injection for this device. (If this control " +
+            "legitimately has no visible effect, ignore this.)"
```
latekvo, line 145: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This warning is platform-blind, but the code path that produces it is not iOS-only. `runWithDeliveryVerification` receives only a `SimulatorServerApi`, which carries no platform information, and `gesture-tap` / `gesture-swipe` route *every* non-Chromium touch through it. So an inert touch on an Android emulator, a physical Android phone, a physical iOS device, an Apple TV, or a remote simulator produces "on an iOS simulator this can mean touch injection has wedged … run recover-touch-injection for this device".

`recover-touch-injection` declares `apple: { simulator: true }` only, so following that instruction is guaranteed to fail. Inert taps are common on Android (padding, disabled controls, an already-selected item), so this is not a corner case — and the agent is sent down a dead end instead of investigating its coordinates.

Verified end-to-end against a real `Pixel_3a_API_34` emulator on this PR's build:

```
$ curl -s -X POST $API/tools/gesture-tap -d '{"udid":"emulator-5554","x":0.05,"y":0.5,"verify":true}'
{"data":{"tapped":true,"timestampMs":...,"verified":false,
 "warning":"Touch was injected but the screen did not change — on an iOS simulator this can mean touch
  injection has wedged (it accepts touches but silently drops them). If taps have stopped landing, run
  recover-touch-injection for this device. (If this control legitimately has no visible effect, ignore this.)"}}

$ curl -s -X POST $API/tools/recover-touch-injection -d '{"udid":"emulator-5554"}'
HTTP 400  {"error":"Tool 'recover-touch-injection' is not supported on android emulator (no android support declared)."}
```

packages/tool-server/src/utils/touch-verification.ts
```diff
@@ -0,0 +1,193 @@
...
+import { sleep } from "./timing";
+
+/**
+ * Observational check that an injected touch landed. The touch transport has no
+ * delivery ack, so the only signal is whether the frame changed around the
+ * injection — a heuristic, not proof (an inert control leaves the frame
+ * unchanged too, and unrelated animation changes it), so callers surface it as
+ * a warning, never a failure. Only the first touch per simulator-server attach
+ * is auto-verified; once one lands (or proves unverifiable) later touches skip
+ * the check. A respawned server (new apiUrl) starts fresh.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: "Only the first touch per simulator-server attach is auto-verified" does not match the code, and both tool descriptions repeat the claim ("the first tap per device session is automatically delivery-verified", "The first touch per device session is automatically delivery-verified").

`recordVerdict` adds to `settledAttaches` only for `landed` and `unknown`. A `no-change` verdict deliberately leaves the attach unsettled, so on a device whose touches keep producing no visible change *every* subsequent automatic touch is checked again, with no bound. On a healthy simulator driven against a region that legitimately does not repaint, that is an endless stream of "may be wedged" warnings plus a permanent latency tax on every gesture — and the agent has no way to know the warnings are self-inflicted, because the docs told it the check only happens once.

Measured against a booted iOS 18.6 simulator running this PR's build (fresh attach, then five consecutive taps on an inert dock gap, `verify` never passed):

```
tap 1  1138ms  checked=True  verified=False  names-recover-tool=False
tap 2   349ms  checked=True  verified=False  names-recover-tool=True
tap 3   340ms  checked=True  verified=False  names-recover-tool=True
tap 4   340ms  checked=True  verified=False  names-recover-tool=True
tap 5   340ms  checked=True  verified=False  names-recover-tool=True

same tap with verify:false (no check): 79ms
```

Every touch is checked, indefinitely, at roughly 4x the unverified cost.

packages/tool-server/src/tools/gesture-tap/index.ts
```diff
@@ -72,7 +88,7 @@ export const gestureTapTool: ToolDefinition<Params, Result> = {
 Sends a Down event followed by an Up event at the same point. For Chromium, this dispatches a CDP mouse-press/release on the renderer.
 Set clickCount: 2 for a double-tap / double-click — the taps are dispatched as one gesture with proper click counting, which two separate tap calls cannot guarantee.
 Use when you need to tap a button, link, or any tappable element on the screen.
-Returns { tapped: true, timestampMs }. Fails if the simulator-server / emulator backend / Chromium CDP is not reachable for the given device.
+Returns { tapped: true, timestampMs }. On iOS/Android the first tap per device session is automatically delivery-verified (a wedged iOS simulator can accept touches but silently drop them): when a check runs the result also carries 'verified' and, if the screen never changed, a 'warning' pointing at recover-touch-injection. Pass verify:true to force the check on any tap, verify:false to skip it. Fails if the simulator-server / emulator backend / Chromium CDP is not reachable for the given device.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: "if the screen never changed, a 'warning' pointing at recover-touch-injection" is not what the automatic path does on its first check. `recommendRecovery` is `explicit || noChangeStreak(api) >= RECOVERY_HINT_STREAK`, so an automatic first no-change deliberately returns the soft text, which does not name the tool at all; only the second consecutive no-change (or an explicit `verify: true`) escalates. `gesture-swipe/index.ts` carries the same sentence.

An agent that read this description will scan the warning for `recover-touch-injection`, not find it, and conclude no recovery is indicated — on exactly the first check that the description advertises as the wedge detector. The SKILL.md prose gets this right ("the warning only points at `recover-touch-injection` once no-change **repeats**"), so the two disagree.

Observed on this PR's build, same inert tap twice on a fresh attach:

```
tap 1  verified=False  warning names recover-touch-injection: False
tap 2  verified=False  warning names recover-touch-injection: True
```

packages/skills/skills/argent-device-interact/SKILL.md
```diff
@@ -353,4 +354,7 @@ Stops on the first error (or unmet `await-ui-element` condition) and returns par
...
+  - **Detect**: the first touch per device session is delivery-verified automatically — if the screen doesn't change, the result carries `verified: false` and a `warning`. A single automatic no-change stays a soft note (many controls have no visible effect); the warning only points at `recover-touch-injection` once no-change **repeats**, or when you force a check with `gesture-tap { verify: true }` (`verify: false` skips it). Confirm a suspected wedge with an explicit `verify: true` **before** recovering.
+  - **Recover**: call **`recover-touch-injection`** with the device `udid`. It restarts the CoreSimulator daemon (`simctl shutdown all` → `killall CoreSimulatorService` → re-boot). **This is host-wide** — every booted simulator is briefly shut down, but the daemon snapshots which were running and re-boots them all automatically: the target is waited on, any siblings finish booting in the background. Then relaunch your app with `launch-app` and continue.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: "the daemon snapshots which were running and re-boots them all automatically" attributes to CoreSimulator something it does not do. The snapshot is taken by this PR's own code (`listBootedDevices` runs `simctl list devices booted -j` before the shutdown) and the restore is a per-sibling `simctl boot` issued by `recoverCoreSimulatorInjection`; launchd only respawns the daemon process itself.

The distinction matters because the sentence reads as an OS guarantee when it is best-effort behaviour of this tool with several ways to not happen: with `rebootAfter: false` no snapshot is taken and nothing is re-booted, a failed snapshot call silently yields an empty list, and each sibling boot is tolerated on failure. A reader who trusts "automatically" will not check whether their other simulators actually came back.

packages/tool-server/test/utils/touch-verification.test.ts
```diff
@@ -0,0 +1,239 @@
...
+
+describe("verifyTouchDelivery", () => {
+  it("reports 'landed' when the frame changes, running the action between captures", async () => {
+    frameQueue.push("before", "after");
+    const order: string[] = [];
+    const verdict = await verifyTouchDelivery(apiA, async () => {
+      order.push("action");
+    });
+    expect(verdict).toBe("landed");
+    expect(order).toEqual(["action"]); // the touch was injected between the two frames
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The feature's central invariant — capture "before", *then* inject, *then* wait `VERIFY_SETTLE_MS`, *then* capture "after" — is asserted by nothing. This test only checks that the action ran at all (`order` is `["action"]` no matter where the injection sits relative to the captures) and the mocked `httpScreenshot` pops from a queue, so it cannot tell an ordering violation from a correct one: moving the injection after the second capture, or deleting the settle wait entirely, leaves the suite green. That is the one thing frame-diff verification depends on for its answer to mean anything.

Two other pieces of new behaviour are equally unpinned. `onAttachClose` is stubbed to a no-op at line 22 of this file, so the per-attach reset that `notifyAttachClosed` performs (the thing that makes a respawned simulator-server re-verify) is exercised by no test in the repo. And the `pointerVisibleApis` bookkeeping added to `simulator-client.ts` — the state that makes verification stand down while a screen recording's touch overlay is on, and without which every recorded touch would falsely read as "landed" — has no test either.

packages/tool-server/src/utils/touch-verification.ts
```diff
@@ -0,0 +1,193 @@
...
+  /** Present when a check ran and the touch could not be confirmed landed. */
+  warning?: string;
+}
+
+const settledAttaches = new Set<string>();
+const noChangeStreaks = new Map<string, number>();
+const RECOVERY_HINT_STREAK = 2;
+
+onAttachClose((apiUrl) => {
+  settledAttaches.delete(apiUrl);
```
latekvo, line 39: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This reset never fires for a remote (MoQ-backed) simulator, so the per-attach state there is permanent for the life of the tool-server process.

`notifyAttachClosed` is emitted only from the WebSocket `error`/`close` handlers inside `getOrCreateWs`, and the only caller of `getOrCreateWs` is `sendCommand` — which returns early at `simulator-client.ts:109-112` whenever `api.transport` is set. An `ios-remote` device routes every touch through the MoQ transport, so no WebSocket is ever opened for it and no attach-close callback is ever delivered. Its `apiUrl` is also not a fresh port per spawn: `buildRemoteInstance` fills it with the stable stub `moq+remote://<device.id>`, so a dispose and respawn re-uses the same key.

The docstring above this file states "A respawned server (new apiUrl) starts fresh"; for a remote simulator both halves of that are untrue. After one confirmed landing, `stop-simulator-server` followed by a reconnect will not re-verify — the attach stays in `settledAttaches` and any wedge that develops afterwards goes unchecked, silently, even though the user did the one thing the docs say clears the state.

packages/tool-server/src/utils/simulator-client.ts
```diff
@@ -110,12 +138,17 @@ export function sendCommand(api: SimulatorServerApi, cmd: Record<string, unknown
...
   api: SimulatorServerApi,
   show: boolean,
   signal?: AbortSignal
 ): Promise<boolean> {
-  return pointerPost(api, { show }, signal);
+  const ok = await pointerPost(api, { show }, signal);
+  if (ok) {
+    if (show) pointerVisibleApis.add(api.apiUrl);
+    else pointerVisibleApis.delete(api.apiUrl);
+  }
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The `if (ok)` guard makes the two directions asymmetric in a way that can silently disable delivery verification for the rest of the session.

The only caller that turns the overlay off is `makePointerControl.disable()`, which passes `AbortSignal.timeout(2000)` and discards the returned boolean. When that POST is merely slow — the abort fires after the server has already applied `show: false` — `pointerPost` returns `false`, the key is never removed from `pointerVisibleApis`, and `isPointerVisible` keeps reporting `true` for an overlay that is actually off.

From then on every automatic touch on that simulator hits the `pointer-active` early return in `runWithDeliveryVerification` and returns `{}` — no `verified`, no `warning`, no state recorded. The automatic wedge detection this PR adds is off for that device with nothing anywhere indicating it, and because the only reset path is a WebSocket close, it does not recover on its own. A 2s timeout against a simulator-server that is already struggling is not a remote possibility, and a struggling simulator-server is precisely when the check matters.

latekvo (changes requested):
\[[Diplomat](https://github.com/latekvo/Diplomat)\]: Reviewed the whole diff plus an end-to-end run of the built tool-server against a booted iOS 18.6 simulator and a Pixel_3a_API_34 emulator.

The detection half works as advertised, and I verified the state machine on real hardware rather than from the tests: a real screen change yields `verified: true`, an inert touch yields `verified: false` with the soft warning, a second consecutive no-change escalates to the recovery hint, a confirmed landing stops further checks, `stop-simulator-server` resets the per-attach state so the next touch is re-checked, and two devices are tracked independently. Frame hashing is deterministic at the 0.25 verification scale (0 differing pairs over 18 back-to-back captures at 0.25 / 0.3 / 1.0 on a static screen), so the heuristic is not vacuous. Locally: 2959 tests pass, `tsc --noEmit -p tsconfig.test.json` is clean, prettier is clean, and the SpiderShield extractor picks up the new tool's description fine.

What I could not get past is the recovery half's success reporting, and one input-validation gap:

- `recovered` cannot be `false` on the `rebootAfter: false` path, and on the default path it only reflects whether the target booted — a `killall` that genuinely failed still reports a successful recovery. The tool description promises the opposite. For a tool whose entire purpose is to remove a false success, reporting one is the thing to fix first.
- The host-wide `simctl shutdown all` + `killall` run against a UDID that has only been shape-checked, so a typo shuts down every booted simulator on the machine before anything can fail.
- Two smaller reporting gaps in the same area: sibling boots that failed are still counted as "restarted", and a failed booted-device snapshot loses every sibling with nothing recorded in `steps`.

On the detection side, two paths can switch the automatic check off without saying so: the attach-close reset never fires for MoQ-backed remote simulators (no WebSocket is ever opened for them, and their `apiUrl` stub is stable across respawns), and a slow `show: false` pointer POST leaves `pointerVisibleApis` stale so every later automatic touch silently returns an empty result.

Also flagged inline: the no-change warning is platform-blind and points Android / physical-iOS / remote devices at a tool that rejects them (reproduced on a real emulator); the "only the first touch is auto-verified" claim in the module docstring and both tool descriptions does not match `recordVerdict`; the description's promise that a no-change warning names `recover-touch-injection` is not true of the first automatic check; one SKILL.md sentence credits the CoreSimulator daemon with work this tool does itself; and the capture/inject/settle ordering plus the two new per-attach and pointer-overlay state layers are untested.

Everything above is anchored inline with a reproduction. Nothing here is a design objection — the approach is sound and the E2E behaviour is good; it is the recovery tool's honesty about its own outcome that needs another pass.

packages/tool-server/src/tools/simulator/recover-touch-injection.ts
```diff
@@ -0,0 +1,118 @@
...
+
+WARNING: this is host-wide — ALL booted simulators are briefly shut down; the target and any other previously-booted simulators are then re-booted automatically (the target is waited on, siblings finish in the background). Reach for it only when a device's taps have stopped landing; confirm the wedge first with gesture-tap { verify: true }.
+
+iOS simulators only. Returns { recovered, steps, disposedServices, note } — recovered is false when a recovery step failed (see steps).`,
+    searchHint:
+      "recover fix wedged stuck touch injection tap not landing coresimulator daemon restart reset simulator hid silent failure",
+    zodSchema,
+    capability,
+    services: () => ({}),
```
latekvo, line 166: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This tool waits on a full cold boot but does not declare `longRunning`, and that flag is what stops the MCP client from timing out and retrying.

`argent-mcp/src/mcp-server.ts:222` passes `fetchTimeoutMs: meta?.longRunning ? null : FETCH_TIMEOUT_MS` (`FETCH_TIMEOUT_MS = 30_000`) into `fetchWithReconnect`, which retries `MAX_RETRIES = 4` more times on failure — and an `AbortController` firing is a failure like any other, with no error-kind check to distinguish "timed out" from "connection refused". So a call that has not answered in 30s is aborted client-side and **re-POSTed up to four more times**. Nothing cancels the server side: `execute(_services, params)` takes no ctx, and `recoverCoreSimulatorInjection` accepts no `AbortSignal`, so the first invocation keeps running while the retries pile on top of it.

Each retry re-runs `xcrun simctl shutdown all` and `killall com.apple.CoreSimulator.CoreSimulatorService` host-wide — shutting down the very simulator the previous attempt is sitting in `bootstatus -b` waiting for. Exceeding 30s is the expected case, not the edge: `BOOT_STEP_TIMEOUT_MS` is 240_000 and its own comment says these boots "can take tens of seconds to minutes". Every other tool in the repo that waits on something slow sets the flag — `flow-run`, `run-sequence`, `await-ui-element`, `await-screen-idle`, `native-profiler-stop`, `await-user-selection`, `tv-remote`. This one has the longest documented ceiling of any of them and is the only one without it.

The agent does not get the result either — it receives the `AbortError`, so it cannot tell whether the recovery worked, and its natural next move is to run the tool again.

Reproduction — drives the REAL `fetchWithReconnect` from the built `argent-mcp` dist against a server that never answers. `fetchTimeoutMs` is 1000 instead of 30000 purely so the repro finishes quickly; `maxRetries` is left at the production default:

```js
import http from "node:http";
import { fetchWithReconnect } from "./packages/argent-mcp/dist/mcp-server.js";
let posts = 0;
const srv = http.createServer((req) => { console.log(`POST #${++posts} ${req.url}`); /* never responds */ });
await new Promise((r) => srv.listen(0, "127.0.0.1", r));
try {
  await fetchWithReconnect(() => `http://127.0.0.1:${srv.address().port}/tools/recover-touch-injection`,
    async () => {}, { init: { method: "POST", body: "{}" }, fetchTimeoutMs: 1000 });
} catch (e) { console.log("client gave up:", e.name); }
console.log(`=> ${posts} POSTs for ONE tool call`);
```

```
POST #1 /tools/recover-touch-injection
POST #2 /tools/recover-touch-injection
POST #3 /tools/recover-touch-injection
POST #4 /tools/recover-touch-injection
POST #5 /tools/recover-touch-injection
client gave up: AbortError
=> 5 POSTs for ONE tool call
```

packages/tool-server/src/utils/coresimulator-recovery.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  const previouslyBooted = rebootAfter ? await listBootedDevices(exec) : [];
+
+  await run("shutdown-all", "xcrun", ["simctl", "shutdown", "all"], { tolerateFailure: true });
+  await run("killall-coresimulatorservice", "killall", [CORE_SIMULATOR_SERVICE], {
+    tolerateFailure: true,
+  });
+
+  if (rebootAfter) {
+    await run("boot", "xcrun", ["simctl", "boot", udid], { timeoutMs: BOOT_STEP_TIMEOUT_MS });
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `boot` is the only step that is not tolerated, and it fails on a state the rest of the sequence makes likely.

`shutdown-all` is tolerated and, taking no `timeoutMs`, inherits `defaultExec`'s `SIMCTL_SPAWN_TIMEOUT_MS` of 10s with `killSignal: SIGKILL` — a ceiling `simctl-config.ts` documents as sized for "cheap `simctl spawn`" calls. It is the only `simctl shutdown` in the repo carrying that ceiling (boot-device.ts:493, device-shutdown.ts:26 and :65, sim-remote.ts:82 all pass none). So a shutdown that is merely slow gets killed, records `ok: true`, and leaves the target Booted. `simctl boot` then exits non-zero with `Unable to boot device in current state: Booted`, and because this step is non-tolerated the whole recovery reports `recovered: false` — while the very next `bootstatus` step passes.

`boot-device` treats that exact message as benign (boot-device.ts:506-511 rethrows only when the message is *not* "Unable to boot device in current state: Booted"), so the two paths disagree about whether it is an error. `simctl help bootstatus` on macOS also states `-b` will "Boot the device if it isn't already booted" and that you "can safely call this before you attempt to start booting the device", so the step that follows already covers the case the failing step was meant to handle.

The consequence is worse than a wrong flag: on `recovered: false` the note tells the agent "The device may still be shut down; boot it with boot-device, or retry this tool" — steering it to re-run a host-wide `shutdown all` + `killall` that already succeeded.

Reproduction — from `packages/tool-server` after `npm run build`; nothing is executed:

```js
import { recoverCoreSimulatorInjection, recoverySucceeded } from "./dist/utils/coresimulator-recovery.js";
const UDID = "AAAAAAAA-BBBB-CCCC-DDDD-EEEEEEEEEEEE";
const steps = await recoverCoreSimulatorInjection(UDID, { exec: async (f, a) => {
  if (a.includes("list")) return { stdout: JSON.stringify({devices:{rt:[{udid:UDID,state:"Booted"}]}}), stderr: "" };
  if (a.includes("shutdown")) throw new Error("Command failed: xcrun simctl shutdown all\n");   // SIGKILLed at 10s, tolerated
  if (a[1] === "boot") throw new Error("Unable to boot device in current state: Booted");
  return { stdout: "", stderr: "" };
}});
console.log(steps.map(s => `${s.step} ok=${s.ok}`), "recovered =", recoverySucceeded(steps));
```

```
[ 'shutdown-all ok=true', 'killall-coresimulatorservice ok=true', 'boot ok=false', 'bootstatus ok=true' ] recovered = false
```

packages/tool-server/src/utils/coresimulator-recovery.ts
```diff
@@ -0,0 +1,149 @@
...
+  } catch {
+    return [];
+  }
+}
+
+/**
+ * Clear a wedged touch-injection pipeline on a *local* iOS simulator, in order:
+ *   1. `simctl list devices booted -j` — snapshot booted sims (if rebootAfter)
+ *   2. `simctl shutdown all`           — release every booted sim session
+ *   3. `killall CoreSimulatorService`  — drop stale HID state (host-wide)
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `killall CoreSimulatorService` matches no process — macOS `killall` compares the process name exactly, with no basename or substring fallback. The code is right (`CORE_SIMULATOR_SERVICE` is the fully-qualified name, and the tool's own description spells it out); it is this docstring and the shipped skill that are wrong, and the skill presents it inline as one link in an otherwise copy-pasteable chain (`argent-device-interact/SKILL.md`: "`simctl shutdown all` → `killall CoreSimulatorService` → re-boot").

Verified on a host where the daemon is demonstrably live, using `killall -s`, which reports what it *would* signal without signalling anything:

```
$ /bin/ps -Ao comm= | grep -i CoreSimulatorService | sed 's|.*/||' | sort -u
com.apple.CoreSimulator.CoreSimulatorService
$ killall -s CoreSimulatorService
No matching processes belonging to you were found        (exit 1)
$ killall -s com.apple.CoreSimulator.CoreSimulatorService
kill -term 57290                                          (exit 0)
```

Anyone following the documented sequence by hand gets "No matching processes", which reads as "there was no stale daemon" rather than "wrong name" — the opposite of what is true.

packages/tool-server/src/tools/gesture-tap/index.ts
```diff
@@ -92,29 +108,34 @@ Before tapping, determine the correct coordinates by using discovery tools — p
...
+          cmd: "touch",
+          type: "Up",
+          x: params.x,
+          y: params.y,
+          second_x: null,
+          second_y: null,
+        });
+      }
+    };
+    const check = await runWithDeliveryVerification(api, params.verify, injectTaps);
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `timestampMs` is stamped at the top of `execute` (line 106), but this call now runs an `httpScreenshot` round trip *before* the injection closure (`touch-verification.ts:106-107` captures the "before" frame, then invokes `action`). So on any checked touch — the first per attach under the automatic policy, or any `verify: true` — the reported timestamp precedes the first `Down` by a whole capture: a POST, a PNG write, a host-side read and a sha1. `gesture-swipe/index.ts:108` has the same shape.

That field is a documented, consumed anchor, not just an informational echo. `react-profiler-analyze.ts:44` tells callers to "Compute as: tapTimestampMs - startedAtEpochMs, using the timestampMs returned by tap/swipe" (repeated at :66 and :84-85), and `react-profiler/pipeline/05-render.ts:312-315` binds each React commit to the annotation with the highest `offsetMs <= commit timestamp` and prints the `(Xs prior)` delta from it. An anchor that is early by the capture latency pulls commits that fired during the capture window onto the tap and inflates every printed delta. The auto-verified first touch on a fresh attach is exactly the tap a `react-profiler-start` → tap flow annotates, and that is also when the capture is slowest — `httpScreenshot` polls up to `FIRST_FRAME_WAIT_MS` (6s) for a first frame on a freshly spawned attach.

Measured with only the transport mocked (an `httpScreenshot` costing 400ms, `sendCommand` recording `Date.now()`), running the real `gestureTapTool.execute` and the real `touch-verification` module:

```
SKEW_MS 403 verified true          # verify:true — first Down fires 403ms after the reported timestampMs
BASELINE_SKEW_MS 0                 # verify:false — the pre-PR path
```

Worth noting the convention here is split — `gesture-drag:51` and `gesture-scroll:107` also stamp at the top of `execute`, while `gesture-pinch:110` and `gesture-rotate:77` stamp at the first event — so this is about the round trip the PR inserted between the stamp and the touch, not about which convention is right.

packages/tool-server/src/utils/touch-verification.ts
```diff
@@ -0,0 +1,193 @@
...
+  const explicit = verify === true;
+  const auto = verify === undefined && shouldAutoVerify(api);
+  if (!explicit && !auto) {
+    await action();
+    return {};
+  }
+  const verdict = await verifyTouchDelivery(api, action);
+  recordVerdict(api, verdict);
+  // An automatic check during a recording can only ever say "pointer-active";
+  // stay quiet and let the first post-recording touch be the one checked.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: "An automatic check during a recording can only ever say 'pointer-active'" is not true for a recording started with `showTouches: false`.

`pointer-active` has exactly one producer — the `isPointerVisible(api)` guard at line 102 — and `pointerVisibleApis` is only ever added to inside `setPointerVisible`, whose only non-test callers are `screen-recording-start.ts:175`/`:187` via `makePointerControl`. That control is built conditionally: `const pointer = showTouches ? makePointerControl(simulator) : undefined` (screen-recording-start.ts:131-134), and `capture.ts:424-430` only calls `enable()` when it exists. `showTouches` is a first-class public param whose description is "Set false to record the raw screen with no overlay".

So during a `screen-recording-start { showTouches: false }` recording nothing registers the apiUrl, `isPointerVisible` is false, and an automatic check takes the ordinary path and can return `landed` / `no-change` / `unknown` — including emitting a wedge warning mid-recording, which the comment says cannot happen. The behaviour is fine; the comment states a guarantee the code does not make, and it is the rationale a future reader would rely on when touching this early return.

latekvo (changes requested):
\[[Diplomat](https://github.com/latekvo/Diplomat)\]: Second pass over the areas the first review did not reach — the MCP transport, the flow / run-sequence surface, error taxonomy, host-binary resolution, and a fresh adversarial re-read of both new modules. Five new findings, none overlapping the eleven already inline.

The one that changes the risk picture is the missing `longRunning` flag. Because this tool waits on a real cold boot under a 240s ceiling but is not marked long-running, the MCP client aborts it at 30s and re-POSTs it up to four more times, with nothing cancelling the server side — so one agent request can fire five overlapping host-wide `simctl shutdown all` + `killall` runs, each shutting down the simulator the previous attempt is waiting to boot. The agent gets an `AbortError` rather than the result, so its natural next step is to try again. Reproduced against the real `fetchWithReconnect` from the built `argent-mcp` dist: five POSTs for one tool call.

Alongside it, `simctl boot` is the only non-tolerated step and hard-fails on the already-Booted state that a tolerated (and 10s-SIGKILL-capped) `shutdown all` makes likely, so a recovery that actually worked reports `recovered: false` and the note steers the user into re-running the destructive path. Together with the `recovered`-always-true issue from the first review, the outcome flag is currently wrong in both directions.

The other three are smaller: `killall CoreSimulatorService` as written in the docstring and the shipped skill matches no process (the code uses the correct fully-qualified name — only the copy-pasteable prose is wrong); the verification capture now sits between `timestampMs` and the touch it is supposed to mark, and that field is the documented anchor react-profiler-analyze tells callers to compute commit annotations from; and the "can only ever say pointer-active" comment is false for a `showTouches: false` recording.

For completeness, things I chased in this pass and did **not** report because they did not survive checking: bare-name `killall`/`xcrun` resolution (all 14 argent/MCP processes running here do have `/usr/bin` on PATH, so the `ps`-style sanitized-PATH failure from #508 does not apply); `steps[].detail` losing the exit code and signal (real, but the benign cases this tool actually hits — a no-match `killall`, an ENOENT — carry their stderr straight through, so the two outcomes stay distinguishable); delivery warnings being invisible to flows (`flow-run`'s `StepReport.result` carries the full tool result, and the skill line targets live steps); an in-flight screen recording broken by the service sweep (`stop-simulator-server` behaves the same way, so it is pre-existing); missing `requires: ["xcrun"]` (sibling single-platform simulator tools do not declare it either); `parseBootedUdids`'s `state === undefined` branch (defensive parsing consistent with the function's own "[] on malformed input" contract and the `dev?.udid` guard beside it); and a WebSocket-open race on the first verified touch (connect measured at 0-4ms against a 300ms window, with 5/5 first-touch checks correct on a fresh attach).

Keeping this at changes-requested for the `longRunning` and boot-state issues. Everything else across both reviews is small and self-contained.

packages/tool-server/src/tools/simulator/recover-touch-injection.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Still present as of f3268c26bd96a3559bc6a407fe70c7e97caafa6e, and there is a degenerate case of this same path where the capability gate does not run at all. `udid: z.string()` (line 16) accepts `""`, and the gate in `http.ts` is guarded by `if (def.capability && deviceArg)` — an empty string is falsy, so `assertSupported` is skipped entirely and the platform check never happens.

Driving the real `createHttpApp` plus this tool definition from the built dist, with `execute` swapped for a recorder so nothing destructive ran:

```
REQ {"udid":"emulator-5554"}  -> HTTP 400  "not supported on android emulator"   execute reached: false
REQ {"udid":""}               -> HTTP 200  {"data":{"ok":true}}                   execute reached: true
REQ {}                        -> HTTP 400  zod: expected string, received undefined
```

With `udid: ""` the recovery then issues `xcrun simctl list devices booted -j`, `xcrun simctl shutdown all`, `killall com.apple.CoreSimulator.CoreSimulatorService` — the two host-wide steps fire before anything ever establishes that the caller named an iOS simulator, or any device at all. Noting it on this thread rather than separately because it is the same destructive path reached through the same missing check.

latekvo:
Hi @L65FREAD,

Thank you for taking your time to contribute!

Just to explain the large number of change requests: Before merging any changes into the codebase, we run all PRs through a comprehensive automated review. This step is important to us, because it ensures minimal possible amount of bugs end up in the production version of Argent.

But there's nothing to worry about, this process happens to every PR, and we encourage heavy use of AI to resolve and respond to the issues brought up in this review. 

Of course, if you feel like you don't have the time to resolve each of these issues, don't worry, just let us know and we'll resolve them for you. Really, it's no big deal :) We still encourage your feedback and critique throughout the process even if you decide we should take over the implementation. Human input is still really appreciated.

Again, thank you for your time, and I hope you contribute more in the future, both your PRs so far are genuinely very valuable to us :)

L65FREAD:
Ha @latekvo  blasted my inbox lol. the issues should be resolved. Diplomat seems cool, going to be trying it

packages/tool-server/test/tools/gesture-tap.test.ts
```diff
@@ -9,18 +9,44 @@ interface TouchCmd {
...
+    await action();
+    return mockCheck;
+  }
+);
+vi.mock("../../src/utils/touch-verification", () => ({
+  runWithDeliveryVerification: (
+    api: unknown,
+    verify: boolean | undefined,
+    action: () => Promise<void>
+  ) => runWithDeliveryVerificationMock(api, verify, action),
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The delivery-verification mock in this file forwards only `(api, verify, action)` to the spy, and every assertion checks exactly those three arguments (`toHaveBeenLastCalledWith(expect.anything(), <verify>, expect.any(Function))`). The `device` value that `gesture-tap`'s `execute` resolves and passes as the fourth argument to `runWithDeliveryVerification` is never asserted. That fourth argument is what decides whether a no-change warning names `recover-touch-injection` (iOS simulator) or instead tells the caller to re-check coordinates (Android / physical iOS) — the platform routing an earlier fix on this PR introduced. Because nothing pins the forwarding, removing the `device` argument from `execute` leaves this suite fully green (confirmed: dropping it and re-running keeps all 10 tests passing), so a regression that reintroduced the platform-blind warning — pointing an Android caller at the iOS-only, host-wide `recover-touch-injection` — would go undetected here. `gesture-swipe.test.ts` drops the fourth argument in the same way.

PR #569, opened 2026-07-24
feat(flow): CSS combinator scopes for flow selectors

hubgan:
## What

Adds the CSS combinators to flow selectors, read off frames. Any action or condition can now scope its matches by where an element sits relative to another:

```yaml
- tap:    { text: Delete, within: { id: profile-card } }          # CSS "A B" — inside that container
- tap:    { role: Switch, next: { text: Wi-Fi } }                 # CSS "A + B" — the nearest follower
- assert: { visible: { role: Button, after: { text: Danger zone } } }  # CSS "A ~ B" — any follower
- assert: { hidden: { any: true, within: { id: empty-state } } }  # CSS "*" — the universal selector
```

| Scope | CSS | Reads as |
| --- | --- | --- |
| `within: <sel>` | `A B` | the match's frame sits inside the scope element's frame |
| `after: <sel>` | `A ~ B` | the match follows the scope element in reading order |
| `next: <sel>` | `A + B` | as `after`, narrowed to the nearest follower of each anchor |

Plus `any: true` — the CSS `*` universal selector, accepted only alongside a scope and never beside `text`/`id`/`role`.

The child combinator `>` has deliberately **no analog**: flow trees flatten to leaves under one synthetic root, so parent/child structure does not survive to replay.

## How

- **Geometric, not ancestry-based.** Frames are what survives flattening, so every scope is resolved from them by `findAll` in `ui-tree-match`. The synthetic root never counts as a scope element, and every relation requires a *distinct* node (nothing scopes itself).
- **Reading order is row-band aware.** An element follows the anchor when it starts below the anchor's bottom edge, *or* shares its row band and sits entirely to its right. That band rule is what makes `{ role: Switch, next: { text: Wi-Fi } }` resolve the Wi-Fi row's own switch even though the taller switch's frame starts a couple of pixels *higher* than the label's — a raw top-y comparison gets that backwards. `followKind` is the single classification both the follower test and the nearest-pick grouping read, so they cannot disagree.
- **`next` is the union over anchors**, exactly as CSS `A + B` is: three rows on screen give three switches, one per label.
- **Scopes compose and nest**, capped at `MAX_SELECTOR_SCOPES` (6) counted across the whole relation *tree* — a size bound, not a depth bound, because three relations branch and the runner's loose-alternative expansion is exponential in the bare-string scopes it finds.
- **Flow-YAML only.** `await-ui-element`'s strict schema rejects `within`/`after`/`next`/`any`, so a scope can never be silently dropped on that path.
- Adding a fourth relation is a **compile error**: `RELATION_RESOLVERS` is a `Record` keyed by the relation union, so a new scope cannot be wired through parse/serialize/reporting and then ignored by the match engine.

## Testing

- **Unit:** full tool-server suite green — **3006 passed / 1 skipped**. New coverage in `ui-tree-match.test.ts`, `flow-utils.test.ts`, `flow-selector-scopes.test.ts` (renamed from `flow-within-selector.test.ts`, now covering all scopes) and a new `flow-skill-docs.test.ts` that parses every YAML snippet in the skill so the docs cannot rot. Each fix is mutation-checked: reverting it fails at least one test.
- **Real-device E2E** (Pixel 3a, Bluesky on Metro 8081, via a branch tool-server): three flows, 20 steps, all green — geometric ordering both ways, two `text` asserts that pin `next`'s pick *by identity* out of five candidates, empty-scope `hidden` semantics, and two observable taps (`next`-scoped and `any: true`-scoped) that navigated to the *adjacent* topic rather than the anchor's own.
- **iOS not covered.** Native-devtools does not attach on either booted simulator in this environment, so *every* flow selector fails there — an unscoped `{ text: "Video" }` fails identically. Verified environmental, not caused by this branch: the globally-installed argent reports the same, direct `SIMCTL_CHILD_DYLD_INSERT_LIBRARIES` injection also fails to connect, and it survives a forced simulator reboot on both iOS 18.3 and 26.5. Worth re-running the iOS leg once that is healthy.

## Notes for reviewers

Two deliberate deviations from CSS, both documented in the skill:

- **`next` is looser than `+`.** Where `A + B` matches nothing unless the immediate sibling is a `B`, `next` keeps looking and returns the nearest match further on. That is what makes it survive the wrapper and spacer nodes a flattened tree is full of — but a row *missing* the control you asked for resolves to the next row's rather than failing. Happy to switch to strict `+` semantics if the loud failure is preferred.
- **"follows" is not transitive.** Against a tall anchor an element can follow something that follows the anchor without following it directly, so nesting `after` scopes is not the same as chaining `~`.

Other things worth a look:

- `nearestAfter` and `afterTester` are direct scans, not indexed. An earlier sorted index was a *second* implementation of the nearest rule and the two disagreed on frames near the tolerance; it also measured 2-13x slower than the scan on realistic shapes, with a prune that provably never fired. The quadratic tail is documented with measured numbers: 1.2 ms for a single-anchor `next` at the 12k-node tree cap, ~555 ms only when the anchor selector matches essentially every node.
- A bare-string *anchor* cannot self-correct the way the loose fallback usually does — a decoy `testID` matching your anchor text wins the identifier pass whenever anything sits after it. The skill tells authors to spell sibling anchors as maps.
- `MAX_SELECTOR_SCOPES = 6` replaces the previous `MAX_SELECTOR_WITHIN_DEPTH = 10`, so a `within` chain deeper than 6 that parsed on the earlier commits of this branch now rejects. Nothing released depends on it.


packages/tool-server/src/utils/ui-tree-match.ts
```diff
@@ -210,29 +227,76 @@ function selectorTextRegex(selector: Selector): RegExp | undefined {
...
-  textRegex: RegExp | undefined,
-  acc: DescribeNode[]
-): void {
-  if (matchNodeWithRegex(node, selector, textRegex)) acc.push(node);
-  for (const child of node.children) collectMatches(child, selector, textRegex, acc);
+// `within` containment tolerance (normalized units): a hair of overhang — a
+// border, a shadow, sub-pixel rounding — must not disqualify an element that
+// visually sits in its container. Matches the magnitude of the flow runner's
+// edge tolerance (EDGE_EPS in flow-actions).
+const WITHIN_EPS = 0.005;
```
latekvo, line 285: [[Diplomat](https://github.com/latekvo/Diplomat)]: This containment tolerance has no test exercising the sub-pixel overhang band it exists for. Mutating `WITHIN_EPS` to `0` leaves the entire flow + ui-tree-match suite (173 tests) green — no fixture places an element in the `(0, WITHIN_EPS]` range across any edge, so the border/shadow/rounding cases the tolerance is meant to admit are never verified, and a later change that dropped or zeroed it would ship undetected.

packages/tool-server/src/utils/ui-tree-match.ts
```diff
@@ -210,29 +227,76 @@ function selectorTextRegex(selector: Selector): RegExp | undefined {
...
+  for (let s = selector.within; s !== undefined; s = s.within) chain.unshift(s);
+  let containers: DescribeNode[] | undefined;
+  for (const level of chain) {
+    const levelMatches = matchesOf(level);
+    const outer = containers;
+    containers = outer === undefined ? levelMatches : levelMatches.filter((n) => inSome(n, outer));
+  }
+
+  const own = matchesOf(selector);
+  return containers === undefined ? own : own.filter((n) => inSome(n, containers!));
```
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat)]: `within` resolution is quadratic in tree size when a container selector matches many non-nesting nodes: every chain level runs `levelMatches.filter(n => inSome(n, outer))` and the final line runs `own.filter(n => inSome(n, containers))`, with `inSome` scanning the whole container set per candidate. Tight id/text containers stay linear thanks to the `.some` early-out, but a broad container form the nested slot explicitly allows — e.g. `within: { role: <common role> }` — drives this into O(N²) on the flattened flow tree (capped at 12k nodes). Measured on a dense tree: ~6 ms/call at 2k nodes, ~16 ms at 4k, ~49 ms at 8k, versus ~0.4 ms unscoped — and `findAll` re-runs on every settle/poll.

latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat)]: Ran a full end-to-end sweep on this one — built the branch, ran the whole tool-server suite (2945 passed / 1 skipped), typecheck + lint + prettier clean, exercised the runner through its real entry point, and validated the geometric containment against real device-captured frames from an Android screen of identical switches (scoped resolution picks the right element; a label leaf that doesn't geometrically contain its switch correctly yields nothing — matching the docs). No bugs and no regressions surfaced across correctness, parsing/serialization, docs accuracy, and existing non-`within` behavior (proven byte-identical). Two small non-blocking notes left inline.

packages/tool-server/src/utils/ui-tree-match.ts
hubgan: Fixed in c6847e6b — added tests that place a leaf overhanging its container edge inside the (0, WITHIN_EPS] band (still resolves) and beyond it (does not resolve), plus an edge-overhang case through the grid path. Mutating WITHIN_EPS to 0 now fails these three tests.

packages/tool-server/src/utils/ui-tree-match.ts
hubgan: Fixed in c6847e6b — findAll now indexes containers in a coarse uniform grid (cell coverage padded by WITHIN_EPS so a tolerated overhang is never pruned) once the container set exceeds a small threshold, so each candidate only tests the containers in its own top-left cell; the exact frameWithin check still decides. Benchmarked 39x-130x faster on a dense tree with a broad-role container, with byte-identical results, and a grid-vs-naive fuzz over a large random tree guards correctness.

latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat)]: Ran another full end-to-end sweep on this at c6847e6b and it comes back perfectly clean. Rebuilt the branch and re-verified that both earlier notes are genuinely addressed: mutating `WITHIN_EPS` to `0` now fails the new tolerance-band tests (both the direct and the grid-indexed paths), and the grid containment is byte-identical to the naive scan across a boundary-straddling differential fuzz while dropping an ~8.8k-node scoped resolve from tens of ms to ~3ms. Beyond that I drove the real flow runner and confirmed a `within` tap lands on the correct element out of identical siblings (where the unscoped selector resolves the first), chained `a within b within c` picks the deeply-nested target, the grid path resolves correctly, and every directive (`tap`/`long-press`/`type`/`pinch`/`await`/`assert`/`when`/`scroll-to`, the loose bare-string scope, and an unresolvable scope) honors the scope and never leaks or acts on an out-of-scope lookalike. Two independent review passes across correctness, geometric containment (incl. adversarial/boundary fuzz), parsing/serialization round-trips, integration/ripple into non-flow consumers, and docs accuracy both returned clean. Nicely scoped and carefully tested — thanks for contributing!

j-piasecki (approved):
Nice idea!

latekvo:
Please adjust how this works as discussed in person before merging.

If you've already done so, then the description is really stale :v

latekvo:
[[Diplomat](https://github.com/latekvo/Diplomat)]: Ran a full end-to-end sweep on this at `b4710b75` — this time over the sibling combinators and universal selector added since the earlier `within`-only pass (`after`/CSS `~`, `next`/CSS `+`, `any`/CSS `*`). Rebuilt the branch: full tool-server suite green (**3006 passed / 1 skipped**), typecheck + lint clean, and drove the real flow runner through its `run-flow` entry point. Across two independent review passes I could not surface a bug or regression on any lens:

- **Geometry** — `after`/`next`/`within` checked against independent brute-force specs over millions of randomized trees: antisymmetry holds (an element left of / above the anchor never follows it), `next ⊆ after`, single-anchor `next` is deterministic regardless of adapter emission order, and the mutual/hairline tolerance cases resolve as documented. The grid-indexed `within` is byte-identical to the naive scan across ~90M boundary-straddling comparisons (negative / >1 / zero-area / sub-`WITHIN_EPS` overhang frames included).
- **Parser / serializer** — `parseFlow(serializeFlow(x))` round-trips for every scope shape (bare-string / map / regex, `any: true`, nesting to the cap); the `MAX_SELECTOR_SCOPES` tree-size bound is exact (6 accepted / 7 rejected, branches counted), cyclic YAML aliases terminate, and the `any` rules can't be bypassed on either the parse or serialize side.
- **Runner integration** — scopes are honored end-to-end for `tap` / `long-press` / `type` / `pinch` / `scroll-to` / `await` / `assert` / `when`, including re-resolution on every scroll settle (a scoped `scroll-to` stops only when the *in-scope* target appears, not on an out-of-scope lookalike) and the correct empty-scope `hidden`/`visible` verdicts. `loose`/`any` never leak into the match engine or `await-ui-element`'s schema, and the loose-alternative cross-product stays bounded (≤32 realizable).
- **Ripple** — the unscoped `findAll` result is byte-identical old→new on a 12k-node tree, so non-flow consumers are unaffected, and `await-ui-element` still rejects `within`/`after`/`next`/`any`. Recorder append round-trips preserve hand-authored scoped steps.
- **Docs** — every behavioral claim in the create-flow skill (the decoy-anchor caveat, `next` reaching past a missing control, "follows" non-transitivity, the `any`-ranking action-vs-condition divergence) reproduced exactly against the code.

One honest coverage note: I couldn't re-run the live-device leg this session (no device was available to me), so the on-device Android/iOS runs rest on your own reported E2E rather than a fresh reproduction here — everything above is against the real code paths with the device tree stubbed at the fetch boundary.

Carefully built and exhaustively tested. Thanks for contributing!

latekvo (approved):
Thanks.

PR #572, opened 2026-07-24
ci: add knip anti-dead-code gate

latekvo:
## What

Adds a CI gate for dead code, using [Knip](https://knip.dev).

We didn't have one. ESLint's `no-unused-vars` only sees unused locals inside a module, so whole files nobody imports, exports nobody uses, and dependencies declared but never required all slipped through. Knip covers that gap in one tool (roughly `ts-prune` + `depcheck` + `unimported` combined).

The new `.github/workflows/knip.yml` workflow ("Dead Code", job `Knip`) runs `npm run knip` on push and PR to `main`. `CONTRIBUTING.md`'s pull-request checklist gains it as a step, with the reason it has to run against a tree with no build output. Knip reads source directly, mapping each package's `dist` entry back to `src`, so the job needs no workspace build. It installs with `--ignore-scripts`: an analysis that only reads `package.json` and source gains nothing from the seven packages here that carry a native install script, six of which reach this runner (`fsevents` is darwin-only). Five try a prebuilt binary before compiling, and `dtrace-provider` always runs node-gyp but ends in `|| node suppress-error.js` — and off mac/solaris/freebsd its `binding.gyp` picks a stub target with no sources, so on ubuntu-latest node-gyp builds nothing for it and `node-pty` is the only package that reaches a compiler at all — an `optionalDependency`, whose failure npm ignores. The exposure is a bump that leaves a package with no prebuild for the runner, which is exactly what `dependabot.yml` pins `tree-sitter` against (0.25.0 ships no Node 24 prebuild and its compile breaks `npm ci`).

While adding it I checked whether any other installed checker was missing a gate. None were: ESLint (`lint.yml`), Prettier (`format.yml`), `tsc` build + `typecheck:tests` + `typecheck:scripts` + vitest (`unit-tests.yml`), the workspace-version / focused-test / merge-marker checks (`repo-hygiene.yml`) and lockfile sync (`lockfile.yml`) are all already enforced. Dead code was the only gap.

## How it's enforced

`npm run knip` is two invocations:

```
knip --exclude exports,types && knip --include exports,types,classMembers,nsExports,nsTypes --max-issues 206
```

Every issue type knip reports is an error — that is its default severity, so `knip.jsonc` sets no `rules` block at all. The pre-existing backlog is bounded by the `--max-issues` ceiling on the second pass instead of being hidden behind a non-failing severity: it cannot grow unnoticed, and the number comes down as follow-up passes pay it off. When a change legitimately adds an export nothing imports yet, the number moves up in the same commit — that bump is what review should see, rather than the contributor deleting someone else's backlog to make room. Two PRs making that identical one-line edit merge without conflicting, so `main` lands one short of the truth and its push-triggered run is what catches it. Splitting the passes also keeps a real failure out from under the backlog — the enforced classes report on their own, and the second pass only runs once the first is clean.

The second pass names `classMembers`, `nsExports` and `nsTypes` explicitly because they sit in knip's own `defaultExcludedIssueTypes`: `--exclude exports,types` never covered them and `--include exports,types` did not ask for them, so a new unused class member cost nothing and moving an export behind opaque namespace use took it out of the ceiling's scope entirely. `classMembers` has 12 findings on the unbuilt tree the ceiling counts (9 on a built one); the other two have none today. None of the three is gated at zero — all five parked classes share the one ceiling, so a new `nsExports` finding passes if something else in the backlog is retired in the same PR. Naming `classMembers` also turns knip's `isSkipLibs` off (`create-options.js:122`), which makes it search declaration files for external references and **suppresses** 44 export/type findings the same run reports without it — 57/181 becomes 44/150. That, not the 12 findings it adds, is why the ceiling is lower than the two counts alone would suggest; `--include-libs` with no `classMembers` reproduces 44/150 exactly.

The ceiling counts what CI counts, against a tree with no build output. Whether a workspace's `main`/`types` path resolves to an existing file changes the result, and not by way of anything in that file — a zero-byte `packages/registry/dist/index.js` moves the count just as a real build does, while a zero-byte `dist/zzz.js` does nothing. A built tree reports fewer issues — 43/147/9 = 199 built, against 44/150/12 = 206 unbuilt — so `npm run knip` passing after a local `tsc --build` is not evidence the job will pass, which is how the Knip job first went red on 199. The ceiling is 206 and it is exact: 206 passes, 205 fails. Remove every `dist/` before trusting a number from it.

The ceiling is one total across those classes, so it bounds the backlog rather than each class: paying off one parked export licences one new one elsewhere in the same PR. Knip has no per-class ceiling.

`treatConfigHintsAsErrors` is on, so an exemption that stops suppressing anything fails the job rather than printing a hint below the backlog.

## Config

`knip.jsonc` documents the false positives so the gate is both green and meaningful.

- **`@swmansion/argent`** is a thin dispatcher whose real payload is an esbuild bundle of the other workspaces (`packages/argent/scripts/bundle-tools.cjs`). Knip can't follow imports across that boundary, so packages consumed only by the bundled-in code read as unused; 11 are exempted by name here, though not all for that reason — the http3-quiche transport is pulled in by `@fails-components/webtransport` itself at runtime, and no first-party file imports it. They're exempted **by name**, not behind a catch-all: `ignoreDependencies` also suppresses `unlisted`, and a pattern that matches everything would take this workspace's own `src/` and `scripts/` out of that check — where an import with no matching declaration is a genuine bug rather than a bundler artifact. (It reaches `unresolved` only nominally: knip matches the list against a package name parsed out of the specifier, and an unresolvable relative path yields none, so even a catch-all leaves those reported. `ignoreUnresolved` is the option that does silence them.)
- **`tool-server`** reaches a few packages in ways static analysis can't see. The exemptions cover two distinct classes, kept apart in the config so each stated reason matches the finding it silences:
  - *declared but loaded too dynamically to detect* (would be reported as **unused dependencies**): `tree-sitter` / `tree-sitter-typescript` via dynamic `_require`, and the http3-quiche transport as a runtime companion of `@fails-components/webtransport`;
  - *imported but deliberately absent from that `package.json`* (would be reported as **unlisted dependencies**): `sharp` (kept out so the ~30 MB native binary isn't a cost every consumer pays) and `electron` (an `optionalDependency` of the published package at one call site, the Electron main-process API at the other).
- Root `scripts/`: knip resolves those from `package.json` scripts **and** from `run:` steps in `.github/workflows` via its github-actions plugin — multi-line `run: |` blocks included. What the plugin doesn't parse is a PowerShell body, so `scripts/ci/windows-chromium-e2e.mjs` — invoked at `windows-e2e.yml:176`, in the `windows-latest` job where PowerShell is the default shell — is the one entry declared by hand.

### Known limits

- `ignoreDependencies` exempts a package **name across its whole workspace**, in both directions; knip has nothing narrower. While `sharp` is listed for `tool-server`, an undeclared `import "sharp"` anywhere in that workspace passes too. The list is held to packages that genuinely cannot be declared, for exactly that reason.
- Knip analyzes JavaScript and TypeScript only (`DEFAULT_EXTENSIONS`), so **only 11 of the 30 files under `scripts/` are in the gate at all**. The other 19 — 15 `.sh` files, `e2e-full/lib/report.py`, `ci/e2e-chromium-page.html`, and e2e-full's `README.md` and `results/.gitignore` — are invisible to it however they are reached — including the whole `scripts/e2e-full/` harness (13 files), invoked from nothing outside its own directory, its only mention elsewhere being a `.prettierignore` entry.
- Reachability's widest hole is the workspace entry. Whatever `src/index.ts` re-exports is reached, so a single `export { x } from "./dead-module";` takes that module out of `files` altogether — verified against a module reported as an unused file one run earlier. The same line applied to parked exports retires them from the count without deleting anything: re-exporting `equalsCI`, `includesCI` and `Settled` from `tool-server`'s entry takes 206 to 203 and frees three ceiling slots. The ceiling bounds what knip can see, not what is dead.
- Test files are entry points, so a production module kept alive only by its own test still passes `files`. Being reached by a test says nothing about being reachable in production, which is what the check approximates, and two modules sit in exactly that gap: `tools/paste` is imported by two test files and registered by nothing, so it is unreachable over MCP and HTTP, and `utils/ios-profiler/lifecycle.ts` is a re-export barrel whose header keeps it "for source compatibility with existing imports", of which its own test is now the only one — its four production consumers all import `profiler-shared/lifecycle` directly. That single test import is exactly what keeps `files` from reporting it: repoint it and knip returns `Unused files (1)`. `knip --production` is the lens that reports the class; the repo isn't clean under it — 13 unused files (those two, seven root scripts, `packages/argent/scripts/{bundle-tools,observe,sync-readme}.cjs` and `packages/registry/tests/helpers.ts`) with the export/type backlog at 227/206 — measured with `classMembers` off, so comparable to the 57/181 above rather than to the ceiling's 44/150 — so a production-mode pass is a follow-up of its own.
- The same entry treatment hides the test trees themselves. Root `package.json`'s `test:scripts` runs `node --test`, which is what makes knip's node plugin contribute its test entries, and that plugin registers four entry patterns in **every** workspace: `**/test/**`, plus `*{.,-,_}test.<ext>`, `test-*.<ext>` and `test.<ext>` matched anywhere. So none of the **380** JS/TS files under `packages/*/test/` can be reported unused, and under `packages/*/tests/` only a name none of those three match can — of today's 15 that is `registry/tests/helpers.ts` alone, which is why it is the lone `tests/` entry in the `--production` list above. A `tests/test-utils.ts` or `tests/test.ts` would be just as invisible. A dead `zz.test.ts` planted in `packages/registry/tests/` passes; a dead `zz-helper.ts` in the same directory fails.
- `knip@5.88.1` is the last 5.x release ever published — 6.0.0 landed the next day and `latest` is now 6.x. The pin is deliberate: knip 6 removed the `classMembers` issue type outright, and the shipped script hard-errors on it (`ERROR: Invalid issue type: classMembers`, exit 2). Upgrading is therefore not a version bump — it means re-deriving the ceiling, rewriting the paragraph that explains it, and resolving the pass-1 findings knip 6 adds (six unlisted binaries: `tail`, `plutil`, `xctrace`, `netstat`, `pkill`, `pgrep`; two unresolved imports: `/usr/bin/osascript`, `/usr/libexec/PlistBuddy`). Dependabot's `npm-minor-patch` group cannot move it, since no 5.x exists above 5.88.1; the `npm-major` group will keep offering the 6.x jump.
- Knip folds `optionalDependencies` into the declared set, so importing `electron` or `node-pty` is not `unlisted` — they pass because it read them. What it never reports is an optionalDependency nothing uses, so only that direction is outside the gate, along with `argent-cli`'s runtime `req("node-pty")`, resolved through a `createRequire(import.meta.url)` in `lens-pty.ts`, which knip can't see either.

## Dead code this found

- `@types/semver` and `@argent/telemetry` from `packages/argent` — both were being suppressed by exemptions whose stated bundler-boundary reason could not apply. esbuild strips types and never resolves `@types/*`, and `packages/argent` names `semver` nowhere; `@argent/telemetry` is aliased to an absolute source path by `bundle-tools.cjs`, so the declaration bought nothing. Removing them is proved safe by building: four of the five bundles still emit and telemetry is inlined in each with no runtime `require` of it. The fifth, preview-window's main, sits behind a downloaded native asset absent from a script-free install, so it is covered by mechanism rather than observation — `bundle-tools.cjs` passes the same `alias` map to every build, and esbuild's `alias` bypasses node resolution entirely: deleting `node_modules/@argent/telemetry` outright leaves all four emitted bundles byte-identical.
Removed alongside, but **not** gate-reported: the `benchmark` npm script in `packages/argent`, whose `scripts/benchmark.cjs` does not exist. A dangling script target is a blind spot of this gate — restoring the script leaves knip at exit 0.

- Four files with no importers: the `render-hook` injected script, react-profiler's `source-map` + `snippet-extractor`, and `scripts/setup-in-workspace.cjs` (referenced nowhere, and its own usage string names a `scripts/setup-project.cjs` that doesn't exist).
- `superagent` in `tool-server`'s `test/http-streaming.test.ts` — imported for a type but declared in no manifest in the repo, resolving only through npm hoisting of supertest's own dependency. Reverting that import to `superagent` reports `Unlisted dependencies (1) … :3:32` and exits 1; `@types/supertest/index.d.ts:10` re-exports the identical type (`type Response = superagent.Response`).
- `@clack/prompts` and `@moq/lite` from `tool-server`'s `package.json` — declared, imported nowhere. (`@moq/net` is the one actually used; `@moq/lite` isn't a peer of it.)

Also deleted here, but **not** something this gate reported: `src/tools/simulator/simulator-server.ts`. It sat in the `files` blind spot above — `interaction-messages.test.ts` imported it, which is enough to count as reached. `setup-registry.ts` never registers it, so it was unreachable over MCP and HTTP, and that catalog roll-call was its only reference; the import and the catalog count (77 → 76) went with it.

## Lockfile

Adding knip moves two transitive packages: `@napi-rs/wasm-runtime` 1.1.5 → 1.1.6 and `@tybys/wasm-util` 0.10.2 → 0.10.3. Knip pulls in `@oxc-resolver/binding-wasm32-wasi`, which floors `@napi-rs/wasm-runtime` at `^1.1.6`, lifting the hoisted entry that had resolved to 1.1.5 for `@rolldown/binding-wasm32-wasi`'s `^1.1.4`, which then cascades to `@tybys/wasm-util`. No runtime impact — both are support packages for the wasm32-wasi fallback bindings, which don't get installed where a native binding exists; `npm ls` reports both extraneous on darwin-arm64. Those two are the only resolved versions that change in the diff; the rest is knip's own tree (49 packages added) and `@moq/lite` removed.

## Verification

- `npm run knip` exits 0.
- **Coverage of the enforced set.** With no `rules` block, knip's included set is 15 issue types: ten enforced at zero tolerance by the first pass, plus the five the second pass names under the ceiling. Eleven mutations cover all ten, each flipping the gate to exit 1 and back to 0 on revert:

  | class | mutation |
  | --- | --- |
  | `files` | a throwaway `.mjs` in `scripts/`; a throwaway `.ts` in `packages/argent/src` |
  | `dependencies` | a bogus declared dep in `packages/tool-server/package.json` |
  | `devDependencies` | a bogus declared devDep in the same manifest |
  | `optionalPeerDependencies` | `bytes` moved to an optional `peerDependency` there |
  | `unlisted` | an undeclared import in `packages/argent/src/cli.ts` |
  | `unresolved` | an unresolvable relative specifier in the same file |
  | `binaries` | a root `package.json` script invoking a nonexistent binary |
  | `duplicates` | `export default sleep` alongside the existing named export in `utils/timing.ts` |
  | `enumMembers` | an unused member on an enum a second module imports and uses |
  | `catalog` | a `catalog` block in the root `package.json` naming a package nothing uses |

  One nuance worth recording: `duplicates` is narrower than it looks. It reports a symbol exported both by name and as `default` (`sleep|default`); a second *named* alias of the same symbol is not a duplicate, and a freshly added symbol nothing references is not reported at all — so several naive probes read as a miss.

  `enumMembers` needed a shape the repo does not currently contain, not an unreachable one. The only exported enum here, `ServiceState`, is re-exported through `@argent/registry`'s entry, where knip counts exports as used, so an unused member added to it is **not** reported — but an enum in a non-entry module that another module imports and uses does report one (`Blue  ZzColor  …/zz-enum.ts:3:3`, exit 1; deleting the unused member returns exit 0). That is the ninth of the ten covered.

  `catalog` is reachable here despite the repo using npm workspaces: knip's catalog loader falls back to `manifest.catalog` when there is no pnpm or yarn workspace file, so a `catalog` block in the root `package.json` reports `Unused catalog entries (1)` and exits 1.
- The two new mechanisms discriminate as well: one extra unused export takes the backlog to 207 and fails the ceiling, and a stale `ignoreDependencies` entry produces `Configuration hints (1) … Remove from ignoreDependencies` and exit 1.
- `npm ci --ignore-scripts` in a clean worktree, followed by `npx knip`, is green — the job doesn't need any of the postinstalls it now skips.
- `tsc --build`, `typecheck:tests` (all workspaces), `typecheck:scripts`, `test:scripts`, `eslint . --max-warnings 0`, `prettier --check .` and the full workspace vitest suite (4711 passed, 3 skipped across 13 workspaces; 3070 of those are `@argent/tool-server` alone) are green; `package-lock.json` is untouched by the review pass.


knip.jsonc
```diff
@@ -0,0 +1,82 @@
...
+    // @swmansion/argent is a thin dispatcher whose real payload is an esbuild
+    // bundle of the OTHER workspaces (see scripts/bundle-tools.cjs). Every
+    // runtime dep it lists is consumed by that bundled-in code, which knip
+    // can't follow across the esbuild boundary — so dependency checking here
+    // is all false positives. Correctness of those deps is already enforced in
+    // the source workspaces (tool-server, installer, mcp, …). The hand-written
+    // dispatcher copied into bin/ at build time is an entry, not dead.
+    "packages/argent": {
+      "entry": ["scripts/argent-simulator-server.cjs"],
+      "ignoreDependencies": [".*"],
```
hubgan: Knip treats any value containing `*` as a regex, so this matches every package name - and the ignore list is applied to `unlisted` and `unresolved` in addition to the unused-dependency classes. Both are set to `error` below, so this silently carves an exception out of two of the eight enforced classes for this workspace.

Verified empirically: adding `import "totally-bogus-undeclared-pkg";` to `packages/argent/src/cli.ts` produces no finding and exits 0, while the identical import in `packages/tool-server/src/index.ts` is reported as an unlisted dependency and exits 1.

The comment above justifies the suppression on the grounds that dependency findings here are all false positives from the esbuild boundary. That holds for the unused direction only. An import with no matching declaration anywhere is a genuine bug class, and `packages/argent/src` (`bundled-paths.ts`, `cli.ts`, `fatal-handlers.ts`, `installer-help.ts`) plus its `scripts/` are now unchecked for it - which is exactly where an externalized package would go missing without CI noticing.

Separately, the "already enforced in the source workspaces" claim has one counterexample: `node-pty` is declared in `packages/argent/package.json` but in no source workspace, and `argent-cli` consumes it at `src/lens-pty.ts:82` via `req.resolve` without declaring it. Invisible either way today, since knip does not check `optionalDependencies`.

knip.jsonc
```diff
@@ -0,0 +1,82 @@
...
+  // here; treat it like eslint/prettier do and skip it entirely.
+  "ignoreWorkspaces": ["packages/argent-private"],
+
+  "workspaces": {
+    // Root workspace: build/dev/release/CI scripts. The ones referenced from
+    // package.json are auto-detected; the rest are invoked by GitHub workflows
+    // or by hand (knip can't see those call sites), so treat every standalone
+    // script as an entry.
+    ".": {
+      "entry": ["scripts/**/*.{cjs,mjs}"],
```
hubgan: This glob masks one script that really is dead: `scripts/setup-in-workspace.cjs` has zero references anywhere in the repo - not in `package.json`, not in any of the 16 workflows, not in docs. Narrowing the entry pattern makes knip report it immediately.

The glob itself is justified for the rest - the other 8 root scripts are each reachable only from `package.json` or a workflow that knip cannot see - so the cost is that this one file is permanently exempt from the gate the PR is adding. Excluding `.ts` is correct: there are no `.ts` scripts, and `tsconfig.scripts.json` covers only `.cjs`/`.mjs`/`.js`.

knip.jsonc
```diff
@@ -0,0 +1,82 @@
...
+        "tree-sitter-typescript",
+        // Runtime companion of the dynamically imported
+        // `@fails-components/webtransport`; loaded internally, never imported.
+        "@fails-components/webtransport-transport-http3-quiche",
+        // Lazy `require("sharp")` in chromium-server/screenshot.ts (optional —
+        // the caller degrades gracefully when it isn't installed).
+        "sharp",
+        // Provided by the Electron host at runtime; deliberately externalized
+        // from the tool-server bundle rather than declared as a dependency.
+        "electron",
```
hubgan: All six exemptions in this block are factually correct - I checked each cited call site - but three of them are categorized as something they are not. `sharp`, `electron` and `superagent` are not declared in `packages/tool-server/package.json` at all, so these entries suppress `unlisted`, not an unused dependency.

This comment is the misleading one: "deliberately externalized from the tool-server bundle rather than declared as a dependency" reads as an explanation for why the package is absent from package.json, when the finding actually being suppressed is a real unlisted import at `src/utils/preview-window.ts:148` (`require("electron")`). The entry is needed; the stated reason describes a different issue class than the one it silences, which will mislead whoever revisits this list.

knip.jsonc
```diff
@@ -0,0 +1,82 @@
...
+    },
+  },
+
+  // Severity per issue type. Mirrors the ESLint rollout in eslint.config.mjs:
+  // gate hard on the high-signal, low-noise classes now, and PARK the large
+  // pre-existing backlog of unused exports/types at "warn" (reported but
+  // non-failing) to ratchet to "error" in follow-up passes.
+  "rules": {
+    // --- Enforced (fail CI) ---
+    "files": "error",
```
hubgan: Worth recording as a known limit of this gate: knip counts test files as entry points, so production source kept alive only by its own test passes `files: "error"` even when nothing reachable at runtime imports it.

There is already a live instance. `src/tools/simulator/simulator-server.ts` is still not registered in `src/utils/setup-registry.ts`, so it is unreachable over MCP and HTTP, but the gate does not flag it because `test/interaction-messages.test.ts` imports it. I confirmed the causality by removing that import in a scratch copy, at which point knip reports the file and exits 1.

This is pre-existing rather than introduced here, and it is the same file the PR description lists as deleted. Flagging it because it is the one dead file this gate is structurally unable to see.

package.json
```diff
@@ -29,6 +30,7 @@
     "@eslint/js": "^10.0.1",
     "eslint": "^10.5.0",
     "globals": "^17.7.0",
+    "knip": "^5.88.1",
```
hubgan, line 33: Adding knip also moves two transitive packages in the lockfile, which the description does not mention: `@napi-rs/wasm-runtime` 1.1.5 to 1.1.6 and `@tybys/wasm-util` 0.10.2 to 0.10.3. Knip pulls in `@oxc-resolver/binding-wasm32-wasi`, which floors `@napi-rs/wasm-runtime` at `^1.1.6`, lifting the hoisted entry that `@rolldown/binding-wasm32-wasi` had pinned at 1.1.5, which then cascades to `@tybys/wasm-util`.

No runtime impact - both are optional wasm32-wasi fallback binaries, reported extraneous on darwin-arm64 and never installed there - and no other resolved version changes in the +734/-24. Calling it out only so the two bumps are not mistaken for unrelated churn when someone audits the lockfile diff.

hubgan (commented):
Reviewed the config and verified the deletions at runtime. All three deleted files are provably dead, both dependency removals are safe, and the gate does discriminate - I confirmed exit 1 on an injected unused file and an injected unused dependency, and exit 0 once removed.

Runtime verification, since the deleted files sat in the debugger and react-profiler paths where a static-only check would not catch a regression: I built the branch and drove the branch-built tool-server against an RN app on an iPhone 16 Pro simulator and a Pixel 3a emulator. On both, `debugger-connect`/`evaluate`/`component-tree` work, and a full `react-profiler` start-interact-stop-analyze cycle returns per-commit breakdowns (2780 fiber renders on iOS, 263 on Android). `react-profiler-component-source` still resolves locations and returns the documented 50-line snippets on both platforms, so the capability `snippet-extractor.ts` and `source-map.ts` nominally served is intact - it is served by `06-resolve/ast-index.ts` and `debugger/source-maps.ts`, neither of which this PR touches. Locally `tsc --build`, `typecheck:scripts`, `typecheck:tests`, eslint, prettier, `check:versions` and the tool-server vitest suite (3070 passed, 0 failed) are all green, and the lockfile regenerates idempotently.

One note that has no line to anchor to: the "Dead code this found" section says four source files were deleted, but the diff deletes three. `src/tools/simulator/simulator-server.ts` was deleted in 164c2e0c and then restored by the main merge in a3ab4f5d, because #582 added `test/interaction-messages.test.ts` which imports it. Restoring it was the right call, but the description still describes the earlier state, and a reviewer approving on it would believe an unregistered tool had been removed. Worth correcting before merge.

Findings inline.

knip.jsonc
latekvo: Fixed in 7808b339.

Confirmed both halves. `ignoreDependencies: [".*"]` matched every package name, and the bogus-import probe reproduced exactly as described - no finding from `packages/argent/src/cli.ts`, exit 1 from the identical import in `packages/tool-server/src/index.ts`. Replaced the pattern with the 13 package names that are actually consumed through the esbuild bundle (5 dependencies + 8 devDependencies, taken from what knip reports with the entry removed). `unlisted` and `unresolved` are live again for that workspace: an undeclared import in `packages/argent/src/cli.ts` now exits 1, and so does an unresolvable relative specifier.

The `node-pty` counterexample holds, and it is wider than one package - knip does not inspect `optionalDependencies` at all. A bogus entry added to `packages/argent`'s `optionalDependencies` is not reported, exit stays 0. So `electron` and `node-pty` are outside the gate in both directions, as is argent-cli's `createRequire("node-pty")`. Dropped the blanket "already enforced in the source workspaces" claim and recorded that limit in the workspace comment instead.

knip.jsonc
latekvo: Fixed in 7808b339. `scripts/setup-in-workspace.cjs` is deleted - zero references anywhere, and its own usage string points at a `scripts/setup-project.cjs` that does not exist.

Chasing it turned up that the glob was not needed for the rest either. Knip's github-actions plugin does read `run:` steps: with the entry removed entirely, the only two files it reports are `setup-in-workspace.cjs` and `scripts/ci/windows-chromium-e2e.mjs` - `grade-skills.mjs`, `sync-next-dist-tag.mjs`, `next-canary-version.mjs` and `extract-tools.mjs` all resolve on their own. Isolated the cause of the one miss by moving that invocation to a single-line `run:` in a scratch copy, at which point it resolves too; the real call site sits in a multi-line PowerShell block the plugin does not parse. So the entry is now just that one file, with that as the stated reason, and the other eight root scripts are covered by the gate rather than exempt from it. A throwaway `scripts/throwaway-dead.mjs` now exits 1.

knip.jsonc
latekvo: Fixed in 7808b339. Confirmed the split by dropping each group in turn: `tree-sitter`, `tree-sitter-typescript` and the http3-quiche transport come back as `Unused dependencies` against `packages/tool-server/package.json`, while `sharp`, `electron` and `superagent` come back as `Unlisted dependencies` against their call sites. The list is now split into those two groups with a header naming the class each one suppresses.

Rewrote the electron note, which was wrong twice over - not just the class, but the mechanism. The two sites need different explanations: `utils/preview-window.ts:148` resolves the executable's path from the install root, where electron is an `optionalDependency` of the published `@swmansion/argent`, and `test/fixtures/electron-smoke-app/main.js:5` is the main-process API supplied by the runtime executing the fixture. Neither is "provided by the Electron host", and the externalization the old note described is a bundler concern with no bearing on the finding. Also gave `sharp` its actual reason (keeping a ~30 MB native binary off every install) rather than just noting the graceful degradation.

knip.jsonc
latekvo: Fixed in 7808b339 - resolved by deleting the file rather than by correcting the description. `registerTool(` appears only in `setup-registry.ts` and `simulatorServerTool` is not among the 75 call sites, so the tool was unreachable over MCP and HTTP; the restore in a3ab4f5d was a merge artifact of #582 landing a test that imports it, not a decision to keep it. Dropped the import and the `definitions.set("simulator-server", …)` line from `interaction-messages.test.ts` and moved the catalog assertion 77 → 76 (verified it discriminates: it fails `expected 76 to be 77` if left at the old value). Full workspace suite green, 3070 passed / 1 skipped. That also makes the "four source files" line in the description true again.

On the structural limit: `knip --production` is the lens that reports the class, but the repo isn't clean under it yet - 14 files, mostly dev-only root scripts that leave the entry set in that mode - so a second pass is a follow-up rather than something to fold in here. Two production modules remain in the blind spot and are now named next to the `files` rule so the follow-up has a work list: `src/tools/paste/index.ts` (same shape as this one - a tool definition `setup-registry.ts` never registers, and its `cmd: "paste"` transport case has no other caller) and `src/utils/ios-profiler/lifecycle.ts` (a re-export shim kept "for source compatibility with existing imports" whose last real importer is gone). Left both alone here: `paste` overlaps in-flight work on `feat/paste-cross-platform`, and neither is something this PR introduced.

package.json
latekvo: Confirmed and documented in 7808b339 - added a Lockfile section to the description.

Verified the scope by diffing resolved versions between `origin/main` and the branch programmatically rather than by eye: exactly two entries present in both change version, `@napi-rs/wasm-runtime` 1.1.5 → 1.1.6 and `@tybys/wasm-util` 0.10.2 → 0.10.3, alongside 49 additions (knip's own tree) and one removal (`@moq/lite`). The constraint chain is as you describe - `@oxc-resolver/binding-wasm32-wasi` requires `@napi-rs/wasm-runtime@^1.1.6` where `@rolldown/binding-wasm32-wasi` only asked for `^1.1.4`.

One refinement on the no-impact half: on darwin-arm64 both packages are present in `node_modules` and `npm ls` reports them **extraneous** - `@oxc-resolver/binding-wasm32-wasi` itself is the one that isn't installed. Nothing consumes them, so the conclusion is unchanged; the description says extraneous rather than not installed.

knip.jsonc
```diff
@@ -0,0 +1,131 @@
...
+  // imports with no matching dependency). Run with `npm run knip`; enforced by
+  // .github/workflows/knip.yml.
+  //
+  // Not analyzed: build outputs and deps (git-ignored — knip skips them) and
+  // the argent-private submodule (own repo + CI), excluded below.
+  "$schema": "https://unpkg.com/knip@5/schema.json",
+
+  // The submodule ships from its own repo and has no package.json checked out
+  // here; treat it like eslint/prettier do and skip it entirely.
+  "ignoreWorkspaces": ["packages/argent-private"],
```
hubgan, line 36: Deleting this entry changes nothing today - knip runs clean without it, because `packages/argent-private` has no `package.json` checked out and the root `packages/*` glob therefore never yields a workspace there. The comment names the condition under which the entry is *unnecessary*. It only becomes load-bearing after someone runs `git submodule update --init`, which is the case actually worth recording here.

knip.jsonc
```diff
@@ -0,0 +1,131 @@
...
+
+  "workspaces": {
+    // Root workspace: build/dev/release/CI scripts. Knip reaches these on its
+    // own, both from package.json's `scripts` and from `run:` steps in
+    // .github/workflows (its github-actions plugin reads them). The one
+    // exception is the Windows E2E driver: it is invoked from a multi-line
+    // PowerShell `run:` block, which that plugin does not parse — so it is
```
hubgan: The exemption is needed, but multi-line isn't what defeats the plugin. `scripts/next-canary-version.mjs` is invoked *only* from multi-line `run: |` blocks - publish-npm.yml:231 (inside a command substitution) and :280 - and knip resolves it fine. What differs for the Windows driver is the shell: windows-e2e.yml:58 is `runs-on: windows-latest`, so that block is PowerShell, and :107 sets `shell: pwsh` explicitly on the sibling step.

knip.jsonc
```diff
@@ -0,0 +1,131 @@
...
+    // declared by hand. Everything else in scripts/ stays covered by the gate.
```
hubgan: "Everything else in scripts/ stays covered by the gate" holds only for the 15 JS files. Knip's `DEFAULT_EXTENSIONS` is JS/TS only, so 14 of the 29 files under `scripts/` are outside its scope regardless of reachability: every `.sh`, plus `e2e-full/lib/report.py` and `ci/e2e-chromium-page.html`. Notably the whole `scripts/e2e-full/` tree (11 files) is referenced from nothing in `package.json` or any workflow - exactly the dead-code class this gate exists to catch, and it is invisible to it.

knip.jsonc
```diff
@@ -0,0 +1,131 @@
...
+        // `require("electron")` twice, for two different reasons. In
+        // utils/preview-window.ts it resolves the Electron executable's path
+        // from the install root, where it is an optionalDependency of the
+        // published @swmansion/argent. In test/fixtures/electron-smoke-app it
+        // is the Electron main-process API, supplied by the runtime that
+        // executes the fixture.
+        "electron",
+        // Type-only import in test/http-streaming.test.ts, reached
+        // transitively through supertest.
+        "superagent",
```
hubgan: Because `ignoreDependencies` also suppresses `unlisted`, these three are exempt across the entire workspace, not just at the call sites described above. I added `import sharp from "sharp"` and `import { app } from "electron"` to `packages/tool-server/src/index.ts` - production source, declared nowhere - and `npm run knip` stayed at exit 0; the same file with a `lodash` import exits 1. That is the same class of bug the comment at lines 35-38 argues is real and worth catching, applied one workspace earlier.

knip.jsonc
```diff
@@ -0,0 +1,131 @@
...
+  // gate hard on the high-signal, low-noise classes now, and PARK the large
+  // pre-existing backlog of unused exports/types at "warn" (reported but
+  // non-failing) to ratchet to "error" in follow-up passes.
+  "rules": {
+    // --- Enforced (fail CI) ---
+    // Blind spot in `files`: test files are entry points, so a production
+    // module that nothing but its own test imports still counts as reached.
+    // Two modules sit in it today — tools/paste (a tool definition that
+    // setup-registry.ts never registers, so it is unreachable over MCP and
+    // HTTP) and utils/ios-profiler/lifecycle.ts (a re-export shim whose last
```
hubgan: `simulator-server.ts` was in this exact state, and the PR description lists it under "Dead code this found". Knip cannot have reported it: its only reference on main was `interaction-messages.test.ts:10,34`, and this PR hand-deletes that import and changes the count 77 -> 76 to make the file go. `knip --trace-file packages/tool-server/src/tools/paste/index.ts` shows the identical shape resolving as reached, and `paste` passes the gate today.

The deletion itself is sound - registration and docs were stripped in 5f7e7cbc back in April. But three files sat in one state and got two opposite outcomes, with no criterion here separating them. There is a real distinction the comment doesn't draw: `paste` and `ios-profiler/lifecycle.ts` each have a test that exercises behaviour (keyboard-secrets.test.ts:3, ios-instruments/lifecycle.test.ts:4), whereas simulator-server's only reference was the registry roll-call.

knip.jsonc
```diff
@@ -0,0 +1,131 @@
...
+    // real importer is gone). `knip --production` is the lens that reports the
+    // class; the repo isn't clean under it yet (dev-only root scripts leave
+    // the entry set there), so a second production-mode pass is a follow-up.
```
hubgan: `knip --production` returns 13 unused files, and 4 of the 11 beyond the two named above are not root scripts: `packages/argent/scripts/{bundle-tools,observe,sync-readme}.cjs` and `packages/registry/tests/helpers.ts`. `bundle-tools.cjs` is the same file line 28 leans on to justify the argent exemptions. Production mode also takes the export/type backlog from 57/181 to 227/206, so the deferred pass is substantially larger than "dev-only root scripts" suggests.

knip.jsonc
```diff
@@ -0,0 +1,131 @@
...
+        // executes the fixture.
+        "electron",
+        // Type-only import in test/http-streaming.test.ts, reached
+        // transitively through supertest.
+        "superagent",
+      ],
+    },
+  },
+
+  // Severity per issue type. Mirrors the ESLint rollout in eslint.config.mjs:
```
hubgan: The parallel to eslint.config.mjs doesn't hold on mechanism. That file parks its pre-existing debt at `"off"` (lines 70-86), not `"warn"` - and since `lint` runs `eslint . --max-warnings 0`, a `"warn"` there would fail CI. Shared intent, opposite behaviour, and "reported but non-failing" is specific to knip.

knip.jsonc
```diff
@@ -0,0 +1,131 @@
...
+    "dependencies": "error",
+    "devDependencies": "error",
+    "optionalPeerDependencies": "error",
+    "unlisted": "error",
+    "unresolved": "error",
+    "binaries": "error",
+    "duplicates": "error",
+
+    // --- Parked: pre-existing debt, out of scope for the initial rollout ---
+    "exports": "warn",
```
hubgan: Warn-class findings don't affect the exit code, and nothing records a baseline count or ceiling, so these two totals can grow indefinitely with the job green. That makes the promised ratchet strictly harder the longer it waits, with no signal at the point new debt lands.

knip.jsonc
```diff
@@ -0,0 +1,131 @@
...
+    "types": "warn",
+    "nsExports": "warn",
+    "nsTypes": "warn",
+    "enumMembers": "warn",
+    "classMembers": "warn",
```
hubgan: Four of the six parked classes park nothing. knip's `defaultExcludedIssueTypes` (node_modules/knip/dist/util/get-included-issue-types.js:3) excludes `classMembers`, `nsExports` and `nsTypes`, and the workflow passes no `--include`, so those three severities are never consulted - inert config. `enumMembers` *is* evaluated and has zero findings on this tree. Only `exports` (57) and `types` (181) have a backlog, so the block implies four classes of pre-existing debt that don't exist.

.github/workflows/knip.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          # Match the other tooling jobs (ESLint 10 needs a modern Node).
```
hubgan: This job never runs ESLint, and knip 5.88.1 declares `engines.node >= 18.18.0`. Matching the other tooling jobs is reasonable; the reason given for it is carried over from lint.yml:23 and doesn't apply here.

.github/workflows/knip.yml
```diff
@@ -0,0 +1,32 @@
...
+          node-version: "24"
+
+      - name: Install dependencies
+        run: npm ci
```
hubgan: `npm ci` runs the native postinstalls - tree-sitter, tree-sitter-typescript, protobufjs, the http3-quiche transport, esbuild - that knip's analysis never touches, since it only reads package.json and source. .github/dependabot.yml:25-30 documents that a tree-sitter bump can break `npm ci` on native compile, so this is a live flakiness surface for the job. lockfile.yml:27 treats the same install as script-free for its own analysis; lint/format/unit-tests carry the exposure too, so this is a repo-wide trait rather than something new here.

.github/workflows/knip.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      # Knip analyzes source directly (it maps each package's dist entry back to
+      # src), so no workspace build is required.
+      - name: Knip
+        run: npm run knip
```
hubgan, line 39: Two things pass through this step at exit 0. First, the 238 parked warn findings print on every run, so the failure signal, when it comes, arrives at the bottom of a ~240-line wall.

Second, knip's configuration hints. I confirmed they work - injecting a stale `ignoreDependencies` entry produces `Configuration hints (1) … Remove from ignoreDependencies` - but the hint prints below those 238 lines and doesn't affect the exit code. With 18 exemptions across two workspaces, several justified by a bundler indirection no test asserts, drift has no practical signal. The current config is hint-free.

knip.jsonc
```diff
@@ -0,0 +1,131 @@
...
+    // class; the repo isn't clean under it yet (dev-only root scripts leave
+    // the entry set there), so a second production-mode pass is a follow-up.
+    "files": "error",
+    "dependencies": "error",
+    "devDependencies": "error",
+    "optionalPeerDependencies": "error",
+    "unlisted": "error",
+    "unresolved": "error",
+    "binaries": "error",
+    "duplicates": "error",
```
hubgan: The verification section claims the gate "discriminates on each enforced class", but the four mutations listed cover three of these eight: `files` (twice), `unlisted` and `unresolved`. I ran two more - a bogus declared dep in packages/tool-server/package.json and an undeclared import in packages/tool-server/src - and both correctly exit 1, so `dependencies` and workspace-level `unlisted` do hold. `optionalPeerDependencies`, `binaries` and `duplicates` remain unexercised.

knip.jsonc
latekvo: Fixed in 0c6eeaab. Confirmed the entry is inert today: `packages/argent-private/` is an empty directory, so the root `packages/*` glob yields no workspace there and deleting the entry changes nothing.

The comment now names the condition that makes it load-bearing — `git submodule update --init` putting a package.json under that path — rather than the condition under which it is unnecessary.

knip.jsonc
latekvo: Fixed in 0c6eeaab. Confirmed both halves.

`scripts/next-canary-version.mjs` is invoked only from multi-line `run: |` blocks — publish-npm.yml:231 (inside a command substitution) and :280 — and knip resolves it with no entry declared, so the plugin does read those blocks.

The Windows driver's own invocation is windows-e2e.yml:176, also inside a multi-line block and with no `shell:` key on the step; what differs is the job, `runs-on: windows-latest` at :58, where PowerShell is the default shell. :107 is the sibling step that sets `shell: pwsh` explicitly and invokes no `scripts/` file. The comment names the shell now, and records that multi-line blocks parse fine.

knip.jsonc
latekvo: Fixed in 0c6eeaab. Confirmed the class; the counts came out a little wider than 14 of 29.

`scripts/` holds 30 files. 11 are in knip's scope — 9 `.mjs` + 2 `.cjs`, no `.js` or `.ts` at all — and 19 are not: 15 `.sh`, `e2e-full/lib/report.py`, `ci/e2e-chromium-page.html`, plus `e2e-full/README.md` and `e2e-full/results/.gitignore`. `DEFAULT_EXTENSIONS` (knip/dist/constants.js:14) is JS/TS only.

`scripts/e2e-full/` checks out as you describe: 12 files, and searching every tracked file for `e2e-full` or any filename inside it turns up exactly one reference from outside its own tree, `.prettierignore:14` — a formatter exclusion, not an invocation. The "everything else stays covered" claim is replaced with the scope limit, naming that harness.

knip.jsonc
latekvo: Fixed in 0c6eeaab as far as knip allows, and recorded where it doesn't.

Reproduced: with `sharp` listed, an undeclared `import sharp from "sharp"` in `packages/tool-server/src` passes at exit 0. `ignoreDependencies` is the only control knip offers here and it matches by name across the workspace, so the fix is to hold the list to packages that genuinely cannot be declared and state the scope rather than the call site.

`superagent` could be declared, so it is gone from the list. Its only site is `import type { Response }` in test/http-streaming.test.ts, and `@types/supertest` declares `type Response = superagent.Response` (index.d.ts:10) — the type comes from supertest, which is already a devDependency. `tsc --noEmit -p tsconfig.test.json` and the tool-server suite are green on the switch.

`sharp` and `electron` stay, and the comment above them now says each entry exempts the name across the whole workspace, not the call site it documents.

knip.jsonc
latekvo: Fixed in 0c6eeaab.

You are right that knip cannot have reported `simulator-server.ts`. The description no longer lists it under "Dead code this found"; it now says outright that the gate did not report it, that `interaction-messages.test.ts` importing it was enough to count as reached, and that it was deleted anyway.

On the missing criterion, the distinction is the one you drew, and it is now stated next to the blind-spot note: whether a test exercises the module or merely names it. keyboard-secrets.test.ts:224 asserts the `cmd: "paste"` that pasteTool sends, and ios-instruments/lifecycle.test.ts:4 runs the shutdown ladder through the shim; simulator-server's only reference was the catalog roll-call, which is why it goes and those two stay.

knip.jsonc
latekvo: Fixed in 0c6eeaab. Reproduced exactly.

`knip --production` reports 13 unused files: the two named modules, seven root scripts, `packages/argent/scripts/{bundle-tools,observe,sync-readme}.cjs` and `packages/registry/tests/helpers.ts`. Exports/types go 57/181 → 227/206.

Both the config comment and the description's known-limits section carry those numbers, so the follow-up is scoped against what is actually there rather than "dev-only root scripts".

knip.jsonc
latekvo: Fixed in 0c6eeaab by dropping the comparison rather than repairing it.

Confirmed the mechanism: all 13 parked rules at eslint.config.mjs:74-86 are `"off"`, and since `lint` runs `eslint . --max-warnings 0`, a `"warn"` there would fail the gate.

The knip config no longer parks anything at a lower severity either — the backlog is bounded by a `--max-issues` ceiling instead (see the thread on the ratchet) — so there is no rollout left to draw a parallel to.

knip.jsonc
latekvo: Fixed in 0c6eeaab. `npm run knip` is now two passes:

```
knip --exclude exports,types && knip --include exports,types --max-issues 238
```

The backlog classes are errors again — knip's default — and the ceiling is what holds them at today's 238 rather than a severity that never counts. Keeping it to a second pass matters: cli.js:61-63 sums the counters of every error-severity group that the run includes, so a single shared budget would have let a new dead file hide behind a backlog entry someone deleted in the same PR. Excluded groups contribute nothing, so the first pass keeps its exact 0 tolerance.

Verified it discriminates: one added unused export takes the total to 239 and exits 1; reverting returns it to 0.

knip.jsonc
latekvo: Fixed in 0c6eeaab. Confirmed at the source you cite — `defaultExcludedIssueTypes` is `['classMembers','nsExports','nsTypes']` (get-included-issue-types.js:3) and neither the workflow nor the config passes `--include`, so those three severities are never consulted. `enumMembers` is evaluated and reports nothing on this tree.

The `rules` block is gone entirely rather than trimmed: `defaultRules` maps every issue type to `'error'` (issue-initializers.js:12), so once the backlog moved to a `--max-issues` ceiling, every remaining entry restated the default. The three inert ones went with it.

.github/workflows/knip.yml
latekvo: Fixed in 0c6eeaab. `node_modules/knip/package.json` declares `"engines": { "node": ">=18.18.0" }` with `"engineStrict": true`, so the comment says that now.

The other half of the old comment was loose too: format.yml and unit-tests.yml are on Node 20, so "the other tooling jobs" was not a single number. The comment cites the jobs that actually pin 24 — lint, lockfile, repo-hygiene.

.github/workflows/knip.yml
latekvo: Fixed in 0c6eeaab — the job installs with `--ignore-scripts`.

Verified rather than assumed: in a clean worktree at this branch, `npm ci --ignore-scripts` succeeds and `npx knip` afterwards exits 0 with identical output, so nothing knip needs comes out of a postinstall. `package-lock.json` carries 9 `hasInstallScript` entries — tree-sitter, tree-sitter-typescript, tree-sitter-javascript, protobufjs, the http3-quiche transport, esbuild, node-pty, dtrace-provider, fsevents — and this job now compiles none of them.

Left lint/format/unit-tests alone: they run the code, so the exposure there is real work rather than waste.

.github/workflows/knip.yml
latekvo: Fixed in 0c6eeaab, with one narrowing on the first half.

Hints gate now: `treatConfigHintsAsErrors` is set in knip.jsonc. Reproduced your probe — a stale `lodash` entry in tool-server's `ignoreDependencies` prints `Configuration hints (1) … Remove from ignoreDependencies` and exits 1, where before it printed and exited 0.

On the 240-line wall: only `duplicates` and the hints were actually below it. Report order follows `ISSUE_TYPES` (constants.js:164), which puts `files`, `dependencies`, `devDependencies`, `optionalPeerDependencies`, `unlisted`, `binaries` and `unresolved` above `exports` and `types`; `duplicates` sits after them, and printConfigurationHints runs after the whole loop (reporters/symbols.js:25-27). Either way the split into two passes removes the wall — the enforced classes report on their own, and the backlog pass only runs once that is clean.

knip.jsonc
latekvo: Fixed in 0c6eeaab. Ran the full matrix against the new script — nine mutations, each exiting 1 and returning to 0 on revert:

| class | mutation |
| --- | --- |
| `files` | a throwaway `.mjs` in `scripts/`; a throwaway `.ts` in `packages/argent/src` |
| `dependencies` | a bogus declared dep in tool-server's manifest |
| `devDependencies` | a bogus declared devDep in the same manifest |
| `optionalPeerDependencies` | `bytes` moved to an optional `peerDependency` there |
| `unlisted` | an undeclared import in `packages/argent/src/cli.ts` |
| `unresolved` | an unresolvable relative specifier in the same file |
| `binaries` | a root `package.json` script invoking a nonexistent binary |
| `duplicates` | `export default sleep` beside the existing named export in `utils/timing.ts` |

One nuance worth having on record: `duplicates` only fires when the aliased symbol is itself referenced somewhere. Aliasing a brand-new export nobody imports reports under `exports` instead and reads as a miss — that is why the first probe came back empty. The description's verification section lists all eight classes with these mutations.

latekvo: Correcting my own reply above: "all eight enforced classes" was carried over from the `rules` block, which no longer exists. Without it, knip's included set is 12 issue types — the ten the first pass enforces at zero tolerance, plus `exports` and `types` under the ceiling. The nine mutations cover eight of the ten.

The other two have no reachable mutation here, and I'd rather state that than let the table imply coverage:

- `enumMembers` is live and works — an unused member in an isolated knip project is reported as `Unused exported enum members (1)`. It has no finding in this repo, and no mutation I can write gives it one: the only exported enum is `ServiceState` in `packages/registry/src/types.ts`, and `--trace-file` shows it re-exported through `src/index.ts`, the package entry. Knip counts entry exports as used without `--include-entry-exports`, so a member added there goes unreported (confirmed both ways in the isolated project: reported from a plain module, silent from the entry file).
- `catalog` checks pnpm catalog entries. There is no `pnpm-workspace.yaml` and no `catalog:` protocol in any manifest, so it cannot fire in an npm-workspaces repo.

The description's verification section says all of this now.

knip.jsonc
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The fix stands; the guarantee stated two lines above it does not, and this entry is the counter-example. Corrected in a2869ed7.

`treatConfigHintsAsErrors` was described as reporting "an entry that no longer suppresses anything". It does not cover this one. With `packages/argent-private` an existing empty directory, there is no hint and the gate exits 0; point an `ignoreWorkspaces` entry at a path that does not exist and knip prints `Remove from ignoreWorkspaces` and exits 1. The `ignoreDependencies` half does behave as claimed — a stale `lodash` entry hints and exits 1.

The claim is now scoped to the `ignoreDependencies` lists, with this exception named.

knip.jsonc
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Your correction was right, the mechanism claim holds, and the worked example I attached to it was wrong. Fixed in a2869ed7.

`scripts/next-canary-version.mjs` does not resolve through publish-npm.yml at all: it is imported by `scripts/next-canary-version.test.mjs:4`, which root `test:scripts` runs. Deleting `.github/workflows/publish-npm.yml` outright leaves the gate at exit 0 — so it resolves through the test chain, and "invoked from nothing else" was false on its face.

The multi-line `run: |` behaviour itself is real and still justifies the hand-declared Windows entry, so only the example went; the `entry` line the sentence exists to support is unchanged.

knip.jsonc
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Made the enumeration exact in a2869ed7.

The headline was right — 30 tracked files under `scripts/`, 11 of them JS/TS — but the list named 15 `.sh` + `report.py` + `e2e-chromium-page.html` = 17 of the 19 excluded, while reading as a full account of the blind area. It now names the remaining two as well (`e2e-full/README.md` and `e2e-full/results/.gitignore`).

knip.jsonc
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The criterion I added to answer this does not survive its own two examples. Replaced in a2869ed7.

"Whether a test exercises the module or just names it" measures test depth, not the production reachability `files` approximates — and both modules it kept fail the latter:

- `src/tools/paste/index.ts` — `pasteTool` is imported by `keyboard-secrets.test.ts` and `interaction-messages.test.ts` and by nothing else. `registerTool(` appears only in `setup-registry.ts` and never for it, and a runtime count of the built registry confirms `paste` is absent from all 73 registered ids. That is precisely the state `simulator-server` was in.
- `src/utils/ios-profiler/lifecycle.ts` — a 9-line re-export barrel whose own header keeps it "for source compatibility with existing imports". There are none left: its only importer is `test/ios-instruments/lifecycle.test.ts:4`, while all four production consumers (`native-profiler/platforms/ios.ts:16`, `screen-recording/capture.ts:7`, and the two blueprints) import `profiler-shared/lifecycle` directly. The shutdown ladder the comment credited it with lives there, not in the barrel.

Both are now described as they are and left in the deferred `--production` set, rather than exempted by a rule that cannot hold.

.github/workflows/knip.yml
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Correcting the fix in 0c6eeaab rather than the original finding — you were right that the reason was carried over from lint.yml, and the replacement was wrong too.

`engineStrict` in a **dependency's** package.json is inert: nothing in npm reads that field. arborist reads the `engine-strict` *config* (`build-ideal-tree.js:208-219`) and otherwise only logs EBADENGINE; there is no `.npmrc` in this repo, and a grep for `engine-strict|engineStrict` over the tree hit exactly one place — the comment itself. knip does not self-gate either: no `process.version`/engines check anywhere in its dist.

The floor cited also does not select 24. `>=18.18.0` is cleared by Node 20, which `format.yml:23` and `unit-tests.yml:23` both pin. The comment now rests only on the half that holds, and I re-checked the three 24s it names (`lint.yml:24`, `lockfile.yml:23`, `repo-hygiene.yml:23`).

Fixed in a2869ed7.

.github/workflows/knip.yml
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The `--ignore-scripts` fix stands; the precedent I cited for it does not. Dropped in a2869ed7.

`lockfile.yml:27` is `npm install --package-lock-only --ignore-scripts`, and `--package-lock-only` never reifies — arborist's `#reifyPackages()` returns at `reify.js:310` before anything reaches disk. So no dependency lifecycle script can run there with or without the flag. Confirmed both ways: `node_modules` is never created, and the lockfile produced with and without `--ignore-scripts` is byte-identical. Not the same reason.

knip.yml's own reason needs no precedent — all five packages it names carry `hasInstallScript` in the lockfile, and `npm ci --ignore-scripts` + `npx knip` is green in a clean worktree.

knip.jsonc
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Two corrections to my own matrix, both re-run. Fixed in a2869ed7 and the description.

`duplicates` is narrower than "fires only when the aliased symbol is referenced somewhere". Against `utils/timing.ts`: `export default sleep` (a referenced symbol) reports `sleep|default` and exits 1; `export { sleep as sleepAlias }` — same referenced symbol, but a second *named* alias — exits 0; and a fresh `zzFresh2` exported both by name and as `default` also exits 0. It wants `name|default` **and** a symbol something references.

`enumMembers` is not unreachable here either, it just needs a shape the repo does not currently contain. An enum in a non-entry module that another module imports and uses does report an unused member: `Blue  ZzColor  …/zz-enum.ts:3:3`, exit 1, returning to exit 0 when the member goes. `ServiceState` stays exempt for the reason already given. That takes the table to nine of the ten first-pass classes, leaving `catalog` as the only one with no reachable mutation.

knip.jsonc
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: One more correction to the scope sentence I added here in a2869ed7. The exception I named was right; the coverage claim around it was too narrow.

I scoped hints to "the ignoreDependencies lists below". They also cover `entry`: adding `"scripts/does-not-exist.mjs"` to the root entry gives `Configuration hints (1) … Refine entry pattern (no matches)` and exit 1, and adding an already-reached file (`scripts/grade-skills.mjs`) gives `Remove redundant entry pattern`, also exit 1. Both `entry` declarations in the file are load-bearing, so neither hints today.

Worth recording alongside it, because it has teeth: an `ignoreWorkspaces` path that does **not** exist is hinted. `rmdir packages/argent-private` and pass 1 fails with `Remove from ignoreWorkspaces`. Since `git clean -ffdx` removes submodule directories, that turns the Dead Code job red for a reason unrelated to dead code — the gap is specifically a directory that exists *without* a package.json, which is the state a plain checkout leaves it in.

Fixed in 9da0ed8d.

knip.jsonc
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Correcting the test-tree note I added to this thread's area in a2869ed7 — its `tests/` half was wrong.

I wrote that none of the 380 JS/TS files under `packages/*/test/` can be reported "while the 15 under `packages/*/tests/` can". knip's node plugin registers **four** entry patterns, not one (`plugins/node/index.js:4-9`): `**/test/**` plus three that match a `*.test.*` filename anywhere. 14 of those 15 files are named `*.test.ts`, so they are entries too. Only `packages/registry/tests/helpers.ts` — the one file there that is not itself a test — can ever be reported, which is exactly why it is the lone `tests/` entry in the `--production` list four lines below.

Reproduced both ways: a dead `zz.test.ts` planted in `packages/registry/tests/` leaves pass 1 at exit 0; a dead `zz-helper.ts` in the same directory reports `Unused files (1)` and exits 1.

Fixed in 9da0ed8d.

knip.jsonc
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Following up here because this thread is where `classMembers` came into the gate, and the reason given for the resulting ceiling was wrong.

The config said naming `classMembers` "re-attributes some findings away from `exports`/`types`". It does not — the findings are *suppressed*, and different ones. Naming it turns knip's `isSkipLibs` off (`create-options.js:122`: `isSkipLibs: !(isIncludeLibs || includedIssueTypes.classMembers)`), which makes knip search declaration files for external references. Measured on the unbuilt tree:

| run | exports | types |
| --- | --- | --- |
| `--include exports,types` | 57 | 181 |
| `--include exports,types --include-libs` | 44 | 150 |

44 findings disappear and 12 `classMembers` findings appear, so the total moves 238 → 206. `--include-libs` with no `classMembers` at all reproduces 44/150 exactly, which isolates the switch.

The same flag is what made the count build-state dependent — the pre-`classMembers` gate was not: 57/181 with `dist/` present or absent. With libs searched, a workspace's generated `.d.ts` supply references that suppress seven more findings, which is the 199-vs-206 gap the ceiling now documents.

Corrected in 9da0ed8d.

.github/workflows/knip.yml
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Follow-up on the Node version, since the replacement reason was still doing more work than it should.

"24 is what the other repo-wide checks pin (lint, lockfile, repo-hygiene)" is true of those three, but it reads as the house convention, and it isn't. Counting every workflow with a `pull_request` trigger and a `node-version`, excluding this one: **3 pin 24** (lint, lockfile, repo-hygiene) and **7 pin 20** (format, unit-tests, skill-description-quality, tool-description-quality, windows-e2e, wayland-e2e, vega-vvd-e2e).

The comment now states the pin as matching those three specifically and names the 7-on-20 rest, rather than implying a convention that runs the other way. The `engines.node >=18.18.0` half is unchanged and still correct — it is below both.

Fixed in d172edbc.

.github/workflows/knip.yml
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Third correction on this thread — the `--ignore-scripts` justification has now named the wrong package set twice.

The original said `protobufjs` was part of a "compile-failure surface". It isn't: its postinstall is pure JS that writes a version-scheme warning. I replaced it with `node-pty` and `dtrace-provider`, which do run node-gyp — and that was wrong for a different reason. Both are `optionalDependencies`, and npm ignores an optional dependency's install-script failure. Measured with a package whose install script exits non-zero:

```
REQUIRED failing install script -> npm exit=3
OPTIONAL failing install script -> npm exit=0
```

`dtrace-provider` is doubly immune — its script is `node-gyp rebuild || node suppress-error.js`. So neither can fail `npm ci` however badly its compile goes.

Six packages here compile on install; the four that are not optional — `tree-sitter`, `tree-sitter-javascript`, `tree-sitter-typescript` and the http3-quiche transport — are the actual surface, and that is what the comment says now. `esbuild` stays out (its postinstall downloads a prebuilt binary, no from-source path) and `fsevents` is `os: ["darwin"]`, so absent on ubuntu-latest.

Fixed in d172edbc.

.github/workflows/knip.yml
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Correcting my own count from two replies up: I said 3 workflows pin 24 and 7 are on 20. It is 3 and **8**.

`e2e-device-smoke.yml` is `pull_request`-triggered and has no `node-version:` of its own — all three of its jobs use `./.github/actions/start-tool-server`, whose `node-version` input defaults to `"20"` (`.github/actions/start-tool-server/action.yml:9-12`, consumed at `:21-23`, and no job passes a `with:`). A grep for `node-version:` across the workflow files cannot see it, which is how I missed it.

Full tally, excluding this job: 12 workflows trigger on `pull_request`; of the other 11, three pin 24 inline and eight are on 20. The comment now names the composite-action inheritance so the number can be re-derived.

Fixed in 9365165f.

.github/workflows/knip.yml
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Fourth correction on this thread, and the one that actually gets the mechanism right. My last reply narrowed the compile-failure surface to the four non-optional packages. That was backwards.

On `ubuntu-latest` none of those four compiles. All three tree-sitter packages ship a `linux-x64` prebuild and load it:

```
$ ls node_modules/tree-sitter/prebuilds/   # darwin-arm64 darwin-x64 linux-x64 win32-x64
$ (cd node_modules/tree-sitter && node -e 'require("node-gyp-build")(process.cwd())')
  -> loads, object
```

`node-gyp-build/bin.js:7-8` only reaches `build()` when `node-gyp-build-test` fails, and the http3-quiche transport is the same shape (prebuild-install first, cmake fallback). The two that *do* fall through to node-gyp here are the two I had excluded: `node-pty` (prebuilds are darwin/win32 only) and `dtrace-provider` (none at all).

So "optional vs required" was never the right axis. The exposure is a bump that leaves a package with no prebuild matching the runner — which is precisely the case `.github/dependabot.yml:25-28` already pins `tree-sitter` against: "0.25.0 ships no Node 24 prebuilds … the native build fails `npm install`/`npm ci`". The comment now says that, and `--ignore-scripts` takes the whole class out of a job that never runs the code.

Fixed in edc7f0d4.

PR #576, opened 2026-07-28
feat(flow): allow to run every flow in a directory from flow run

j-piasecki:
`argent flow run` now accepts a directory and runs every flow in it sequentially:

```
argent flow run .argent/flows --recursive
```

Discovery uses the same acceptance rules as `flow list` (lowercase `.yaml`, safe stem, readable file), sorted for a deterministic run order. `-r`/`--recursive` walks subdirectories too, skipping dot-directories and `node_modules` and never following directory symlinks. A directory named `foo.yaml` still batches (stat wins over extension).

Batch output is failures-only: each flow prints a `[i/n] <path>` header, its failing/errored steps — numbered exactly as a single-mode rerun would number them — and a per-flow summary; the run ends with a `PASS/FAIL — N flows: X passed, Y failed, Z skipped` verdict. `--json` prints one aggregate object with per-flow reports. Exit code reflects the batch (0 pass, 1 any failure, 2 discovery/config errors).

`--output` exports keep the single-run `<output>/<flow>/` layout, keyed by subdirectory in recursive runs so same-stem flows in different directories can't clobber each other's snapshots.

**Failure semantics:** a flow failing its steps — or one the tool-server rejects as invalid (bad YAML, unparseable step) — fails alone and the batch continues. An infra error (transport death, unclassified failure) stops the batch and counts the remaining flows skipped. To make that distinction possible, tool-error responses now carry the server's failure signal (`error_code`/`error_kind`) on the wire — on the streaming terminal error line and on JSON error bodies — and the tools client surfaces server-answered errors as a typed `ToolInvocationError`. Against a pre-signal server this degrades to stopping on any throw.

Not breaking: single-file runs behave exactly as before.


packages/argent-cli/src/flow.ts
```diff
@@ -429,6 +483,202 @@ async function isRunnableFlowFile(filePath: string): Promise<boolean> {
...
+      // typeof guard: `in` throws on a primitive wire value, and that must
+      // classify as "no report", not as an infra throw.
+      if (data && typeof data === "object" && "steps" in data) report = data;
+    } catch (err) {
+      const message = err instanceof Error ? err.message : String(err);
+      console.error(message);
+      results.push({ path: rel, status: "fail", error: message });
+      const rejectedThisFlowOnly =
+        err instanceof ToolInvocationError && err.errorKind === "validation";
+      if (!rejectedThisFlowOnly) stopped = true;
```
latekvo, line 1136: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This gate keeps the batch going only when the error is a `ToolInvocationError` with `errorKind === "validation"`; every other throw sets `stopped = true` and marks all remaining flows `not run (batch stopped)`.

A **syntactically malformed** YAML flow file never arrives here as a validation error. `parseFlow` runs the `yaml` parser before its own structural checks (`packages/tool-server/src/tools/flows/flow-utils.ts:1762`, called unguarded at `flow-run.ts:552`), so a YAML syntax error surfaces as the library's raw `YAMLParseError` — a plain error with no failure signal. It reaches the client with `errorKind` undefined, so `rejectedThisFlowOnly` is false and the whole batch aborts.

Reproduced against the real tool-server — a directory with `a-bad.yaml` (`steps: ][`) then a valid `b-ok.yaml` (`steps: []`):

```
[1/2] a-bad.yaml
[Tool:flow-execute] Unexpected flow-seq-end token in YAML stream: "]" ...
[2/2] b-ok.yaml
  · not run (batch stopped)

FAIL — 2 flows: 0 passed, 1 failed, 1 skipped
```

`b-ok.yaml` is valid and would pass, but is skipped; placing the malformed file in the middle of the list skips every flow after it as well. This contradicts the contract stated in the help text (`An invalid flow file fails alone and the batch continues`), in `argent-create-flow/SKILL.md`, in the PR description ("bad YAML ... fails alone and the batch continues"), and in the docstring directly above this function. A structurally-invalid file (valid YAML, wrong shape), a non-array `steps`, an unknown top-level key, and an unrecognized step kind all **do** continue — only an unparseable file, arguably the most common form of "bad YAML" (a stray bracket, a bad indent, a tab), stops the run. The unit test covering this path constructs the server error as `errorKind: "validation"`, so it passes without exercising the real classification of a malformed file.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat)\]: Ran a full end-to-end sweep of the directory-batch feature against the real auto-started tool-server — discovery/recursion, dot-dir & `node_modules` skipping, symlink/loop safety, exit codes (0/1/2), `--json` aggregate output, and the `error_kind` wire path from `tool-server` through `@argent/tools-client`. Almost all of it holds up. One inline note on the stop/continue behavior for unparseable flow files.

packages/argent-cli/src/flow.ts
j-piasecki: Good catch — confirmed exactly as described: `yamlParse` was the one unguarded throw in `parseFlow`, so a syntax error escaped as a raw `YAMLParseError` with no failure signal, reached the CLI with `errorKind` undefined, and the gate treated it as infra.

Fixed at the root in 27dda6bb7d8783ea8decb8fe5f980a7a8c57e531 rather than at the gate: `parseFlow` now wraps `yamlParse` and rethrows as `FailureError` with the same validation signal its structural checks use (`FLOW_FILE_INVALID`, `error_kind: "validation"`), keeping the parser's line/column detail in the message and the original error as `cause`. An unparseable file now fails alone and the batch continues; the `run:` fragment and flow-editing paths pick up proper classification too.

On the test gap: added a tool-server test that asserts the signal through `getFailureSignal` — the same mechanism the HTTP layer serializes from — so it exercises the real classification instead of constructing it. With the wrap reverted, the test fails with `error_kind: undefined`, which is precisely the hole you found.

latekvo (approved):
Amazing feature I love it ❤️

PR #578, opened 2026-07-28
feat(flow): CI-friendly failure diagnostics for argent flow

hubgan:
## Problem

A flow that fails today tells you almost nothing. The entire failure payload is one prose string:

```
  ✗  3 tap "Checkout" — no visible element matched selector text="Checkout" — if it is off-screen, add a scroll-to step before this one
```

No screenshot, no UI tree, no candidate elements, no error code, no timing, no report file. In CI that is a dead end: the device is gone, the app has moved on, and the only recourse is a re-run that often doesn't reproduce. Our own docs concede the gap - `argent-create-flow/SKILL.md` told the agent to call `screenshot`/`describe` *after* a failure, which captures a screen the failing step never saw (post-abort, post-teardown) and is impossible in CI with no LLM in the loop.

## What this does

When a step fails, the run itself captures why - a classified cause, what was actually on screen, the closest matching elements, and a screenshot at the moment of failure - and renders it to the terminal, to JUnit XML for CI, and to the MCP/agent surface.

```
Failures:

  1) tap "chooseAddAcountBtn"  @ .argent/flows/typo-tap.yaml
     selector-not-found: no visible element matched selector text="chooseAddAcountBtn"
     context:
       › Tap an id with a typo
       ✗  1 tap "chooseAddAcountBtn" (7.6s)
     candidates:
       0.93  "Sign in to account that is not listed"  Button     id=chooseAddAccountBtn  visible  at 0.50, 0.31  → {"id":"chooseAddAccountBtn"}
       0.76  "Select account"                         ViewGroup  id=chooseAccountForm    visible  at 0.50, 0.24  → {"id":"chooseAccountForm"}
     screen: 15 elements, 1080x2220
     screenshot: flow-artifacts/typo-tap/step-01-screen.png
     tree: flow-artifacts/typo-tap/step-01-tree.txt
```

The evidence was always there - the runner held the answer at the moment it failed and threw it away. `waitForCondition` discarded its matched nodes and read statistics; `waitForFrame` discarded the entire settled tree, so `tap`/`type` failures had zero evidence; `DirectiveOutcome.indeterminate` was dropped entirely by `execLeafStep`, so an unreadable tree source and a genuinely broken flow were indistinguishable.

That last one is the point for CI. **`determinacy` separates "the check failed" from "argent could not see the screen"**, which is what an operator needs to choose between re-running and editing the flow:

```
condition-dark-tail: could not evaluate the condition — the UI tree was unreadable
                     for the final 14203ms of the window
screen: 10 elements, 1080x2220 (last trusted read)
reads: 34 attempted, 0 trusted
hint: the screen went unreadable before the deadline — this is not a failed assertion;
      re-run rather than editing the flow
```

## Shape

**New** — `flow-failure.ts` (30-code taxonomy, total category map, wire types, budgets, secret scrubber), `flow-candidates.ts` (ranked suggestions + scope diagnosis), `flow-failure-report.ts` (the assembler), `argent-cli/src/flow-report.ts` (pure line builders + JUnit).

**Modified** — evidence plumbing in `flow-actions`/`flow-run`/`flow-visual`; the CLI failure block, step timings, `--reporter junit:<path>` and `exportRunArtifacts`; the MCP failure section; the SKILL.md Diagnose procedure.

Three design constraints shaped it:

- **The plumbing is additive.** Every change is a new optional parameter or field, so the abort→skip mapping, the three evidence tiers and the one-failure-per-run invariant are untouched by construction. `flow-when`, `flow-hidden-blank-reads`, `flow-abort`, `flow-progress` and `flow-composition` pass unmodified.
- **Capture may only ever degrade its own payload** - never a verdict, never a throw into the runner, never past a 5s budget. The test harness passes no `ctx` at all, so a single unguarded `requireArtifacts` would break ~20 files.
- **Wire compat both directions.** A pre-diagnostics server sends no `failure`, and every renderer is gated on presence; a new server's report renders byte-identically on `main`'s CLI (verified by diffing real output, not by reading).

Also fixes a standalone bug found while reading: the CLI's `FlowReport` had no `aborted` field though the runner has always set it, so a cancelled run printed `FAIL` with every step pass/skip and no explanation.

## Verification

Verified on a Pixel 3a against the Bluesky dev-client, through an isolated branch tool-server - not just unit tests. Every failure shape: `selector-not-found` with a ranked near-miss, `text-mismatch`, `assert-hidden-unmet`, `selector-scope-unresolved`, `snapshot-baseline-missing`, `snapshot-diff`, and `condition-dark-tail` induced by holding the Android devtools helper down mid-run.

The decisive check: **the top candidate's selector, pasted verbatim into a flow, makes the step pass.** The screenshot is confirmed to be the post-navigation screen the step actually failed on. `xmllint` accepts the JUnit; exit code is 1.

Suite: 4822 passed. Five remaining failures are pre-existing - four verified failing identically on `main`, one that passes in isolation and only trips under full-parallel load.

## Review found real bugs

Two rounds of parallel review, sixteen confirmed findings fixed. The ones worth naming:

- **Four secret leaks.** A candidate's `selectorYaml` was built from the node's *raw* strings, so a row carried a masked label beside a suggestion quoting the credential in full - straight into the model's context, the one thing `{{secret:}}` exists to prevent. `cause.message` and `expected` were never masked at all, and the selector projections were scrubbed as *serialized JSON*, so any secret containing a quote or backslash was invisible to the mask.
- **The scope diagnosis rewrote an indeterminate verdict to determinate**, turning "re-run" into "your flow is wrong" on exactly the mid-run devtools drop the tier exists for.
- **A blind read succeeds**, so it recorded no error, so a post-hoc read rendered a healthy screen next to "every read of the UI tree was empty or degraded".
- **`truncateUtf8Field` reserved one byte for a three-byte ellipsis**, overshooting its limit on every truncation.
- A timed-out capture kept running and touched the device after teardown; each failing run leaked a temp directory forever on a host-wide server; `stepLabel` was the one unclamped wire path and now feeds the failure header, where ANSI escapes could repaint the verdict above it.

## Known gaps

- **An environment fault reads as a determinate selector failure when the tree is readable but the app isn't foregrounded** (app sitting on the Expo launcher → "edit your selector"). Fixing it needs a foreground-app signal the tool-server doesn't currently have.
- **The cancelled-run suite-level `<error>` is spec-correct but invisible to dorny/test-reporter and mikepenz/action-junit-report**, which build results only from `<testcase>` - so a cancelled run shows green in the checks UI.
- **`resolveLocalFile` proves co-location by size match alone** (pre-existing in `tools-client`); this adds two call sites, widening reachability from "a snapshot that produced a diff" to "the first failing step of any kind".
- Phase 9 of the plan (YAML source lines for `@ file:14`) is deliberately deferred; every surface renders correctly without it.


j-piasecki:
Nice! I had something similar in my todos. Some ideas I also had noted were capturing the device logs and adding them as artifacts in case of a failure, and screen recording the entire flow execution and adding that as well.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -617,7 +662,19 @@ async function runLaunch(state: ExecState, app: Launch): Promise<DirectiveOutcom
...
   // it, or a cancelled gate would read as a launch that verified readiness.
   if (signal?.aborted) return ABORTED_OUTCOME;
-  if (gate) return { ok: false, reason: gate };
+  if (gate) {
+    return {
+      ok: false,
+      reason: gate.reason,
+      // The tree source never came up: an environment problem, not a broken
+      // flow. Classified as indeterminate so CI can tell retry from fix.
+      evidence: {
```
latekvo, line 671: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The blocked native-devtools precheck return ~13 lines above (`isNativeDevtoolsBlockResult("restart-app", restart)`) is the one `ok: false` in `runLaunch` that this PR left without `evidence`. Every sibling got one: `no app id declared`, `restart-app failed` (+`causeOf`), this `tree-source-not-ready` block, and all four chromium returns.

With no evidence `baseFailure` falls back to `evidence?.code ?? "unclassified"`. Checked against the built module:

```
category: tool
determinacy: determinate
isScreenless (code === "launch-failed" || category === "composition"): false
```

So an iOS launch blocked by a stale / `restart_required` / `init_failed` devtools service is reported as a determinate **tool** failure — the classification a CI operator reads as "fix the flow" rather than "repair the device and re-run".

The second-order effect is the larger one. Because `unclassified` is not screenless, `resolveScreen` falls through to a post-hoc `fetchFlowTree` against the very tree source that just answered blocked, and `captureScreenshot` runs, attaching a full-resolution capture of whatever app happens to be foregrounded — which `--output` then exports as this failure's evidence. That is the behaviour `ScreenResolution.screenless`'s own doc says it exists to stop. `restart-app` resolves the block rather than throwing, so control reaches that return on any iOS flow hitting the stale-service class.

The same hazard has a second spelling that also ships no evidence: the `tool:` step's `isNativeDevtoolsBlockResult(step.name, result)` return in `execToolStep` (~2716) reports `status: "fail"` with a reason and no `evidence` either. Its own comment names the relationship — "`launch:` already guards its own `restart-app` (see runLaunch); this is the `tool:` spelling of the same sub-tools, plus the six feature tools it never covered" — so both spellings of a resolved devtools block land as `unclassified` / `tool` / determinate, while the twin that *throws* the same condition is classified through `evidenceFromThrow`.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Reviewed `d2ed6b5` (32 files, +11852/-266) in a worktree built from the branch, driving the built `dist/` of `tool-server`, `argent-cli` and `argent-mcp` directly, so the claims below are observed renders rather than reads.

**What was run**

- The full `packages/tool-server` suite on the branch: 53 files / 1329 tests, all passing.
- Three-renderer differentials — the same `FlowStepFailure` pushed through `renderFailures` (terminal), `junitSuite` (XML) and `failureBlocks` (MCP), compared field by field. Most of the comments below come from here.
- `attachFailureDiagnostics` driven on the built module with `fetchFlowTree` and `screenshot` stubbed, to exercise the post-hoc read path without a device.
- The CLI end to end in a temp project for the reporter-artifact behaviour, including a pre-existing report file across a refused run.
- Scaling measurements on `rankCandidates` against `flowMatchAll` for the regex path.
- `packages/registry/src/artifacts.ts` on this branch against `origin/main` for the `register` signature.
- `format-tree.ts` differentially against `origin/main` over 24000 randomized trees — behaviour-identical, no finding.
- `truncateUtf8Field` at every multi-byte boundary, and the JUnit output through `xmllint` under adversarial field content — both clean, no finding.
- Mutation checks against the assertions covering the new branches.

**Shape of the comments**

Four clusters, roughly.

One payload rendering differently on the three surfaces this PR introduces — `cause`, `capture-failed`, `candidateCount`, the indeterminate-tree marker, the secret-typed artifact warning. In each case the CLI side and the MCP side disagree about the same wire field, and the doc tables shipped here describe one of the two.

The post-hoc read: `resolveScreen` falling through to a fresh `fetchFlowTree` for codes that were never meant to reach it, and what the report then asserts about a screen read after the app moved on.

Evidence that is produced and then goes nowhere — the overflow spill, the `tree` dump's failure mode, `failure.cause`.

A fourth cluster is the two skill references disagreeing with the renderer and with each other about when an agent may call `screenshot` after a replay failure.

Separately, `store.register` moved to a required-`kind` options object on `main` after this branch was cut, so both new call sites will not type-check once rebased.

One finding landed outside the diff and is filed as #961 rather than commented here: the `snapshot` `cropOn` baseline key hashes the selector's own fields but not `within`/`after`/`next`, so two scoped crops sharing a name share one baseline file.

Everything below names the concrete trigger and what it produces. Style, phrasing and unreachable-defect notes were dropped rather than filed.

packages/tool-server/src/tools/flows/flow-failure-report.ts
```diff
@@ -0,0 +1,1085 @@
...
+    // where the report says it was. Today the operator gets a message about
+    // the TARGET and goes hunting in the wrong place.
+    // Only ever REFINE a determinate selector failure. An indeterminate verdict
+    // means argent could not see the screen, so "your scope matched nothing" is
+    // a claim about a tree nobody trusts — and rewriting the code, category and
+    // determinacy together left the three self-consistent while contradicting
+    // the message beside them. `determinacy` is the one field a CI operator
+    // branches on for retry-vs-fix, and this turned "re-run" into "your flow is
+    // wrong" on exactly the mid-run devtools drop the tier exists for.
+    if (selectorBlock.unresolvedScope !== undefined && failure.determinacy === "determinate") {
```
latekvo, line 858: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This refinement is gated on `determinacy`, but never on the provenance of the tree the scope was diagnosed against. For any determinate code carrying `evidence.selector` and no `evidence.tree`, `resolveScreen` falls through to its own post-hoc `fetchFlowTree`, so `diagnoseScope` runs against a screen read *after* the step failed — the read this module elsewhere describes as an app that "may have moved on".

`snapshot-crop-empty` is the reachable case: `runSnapshot` returns `evidence: { code, selector: opts.cropOn, expected }` and drops the tree its `TreeReadSink` already filled, and `cropOn` goes through `parseSelector`, so it can carry `within`/`after`/`next`.

Driving `attachFailureDiagnostics` on the built module with `fetchFlowTree` stubbed to a post-hoc screen that no longer holds the scope container:

```
code       : selector-scope-unresolved
category   : selector
determinacy: determinate
message    : snapshot "receipt": the cropOn region rounds to zero pixels
hint       : the within scope matched no element, so the target was never searched for — fix the scope selector first
screen     : available after-failure
candidates : 2 ["{\"id\":\"summaryTotal\"}","{\"text\":\"Total due\"}"]
```

A crop bug is filed and grouped as a selector-scope authoring error, with a hint that contradicts the message beside it. And because the rewrite mutates `failure.category` before the candidates guard below, `wantsCandidates` — whose comment names `snapshot-crop-empty` as exactly the case category is the honest test for — now passes, so the block also offers two "did you mean" selectors for a failure where nothing was mis-selected.

packages/tool-server/src/tools/flows/flow-candidates.ts
```diff
@@ -0,0 +1,633 @@
...
+ * A regex locator is tested against `subtreeText` here even though
+ * `matchNodeWithRegex` deliberately restricts itself to the node's own
+ * label/value. The engine's exclusion protects RANKING BETWEEN MATCHES — every
+ * unshielded ancestor of a text leaf would match and dilute exact-beats-
+ * substring. Nothing matched here (that is why a report is being written), and
+ * on a flattened tree the hoisted text is the only place the string lives, so
+ * excluding it would hide the very element the pattern was written for.
+ */
+function scoreRegex(node: DescribeNode, regex: RegExp): Scored | undefined {
+  for (const field of [node.label, node.value, node.subtreeText]) {
```
latekvo, line 197: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `scoreRegex` tests the pattern against `node.subtreeText`, which `matchNodeWithRegex` deliberately excludes. On a flattened tree `subtreeText` is a container's whole hoisted text, so ranking feeds the author's regex strings orders of magnitude longer than anything the match engine ever hands it — and this walk is synchronous, so the `Promise.race` behind `FLOW_DIAGNOSTICS_BUDGET_MS` cannot preempt it (the same property the dedupe loop's comment calls out). `similarity()` bounds its inputs with `EDIT_DISTANCE_MAX_CHARS` for this reason; `scoreRegex` has no bound.

Measured against the built dist — one Android-shaped tree of container rows carrying hoisted `subtreeText`, a selector of `text: { matches: "^(\w+\s?)+$" }` — "the row is words separated by spaces":

```
3 rows x 4 words: flowMatchAll 1ms | rankCandidates 4ms
3 rows x 5 words: flowMatchAll 0ms | rankCandidates 54ms
3 rows x 6 words: flowMatchAll 0ms | rankCandidates 891ms
3 rows x 7 words: flowMatchAll 0ms | rankCandidates 14071ms
```

Matching stays at 0ms because label/value are short; ranking grows ~16x per added word of hoisted text. At 30 rows x 8 words the same call had not returned after 10 minutes. The tool-server is host-wide and single-threaded, so the failing step blocks every other agent's request on the box for that whole time — and the failure block it was assembling never arrives.

packages/argent-cli/src/flow.ts
```diff
@@ -1538,6 +2083,31 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
...
+   *
+   * Returning through `exitAfterFlush` without this left CI worse off than a
+   * wrong file: the publisher picks up whichever `junit.xml` sits at that path,
+   * which is the PREVIOUS run's, so a red build reports the last green one's
+   * results. The directory path already synthesises a suite for exactly this
+   * (`rejectedFlowReport`, whose comment calls a `failures="0"` file for a run
+   * that exited 1 "the one thing a CI artifact must never do"); the two paths
+   * simply disagreed.
+   */
+  const reportRejection = async (message: string): Promise<void> => {
```
latekvo, line 2099: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `reportRejection` is wired into two exits — the `callTool` throw and the non-report result. Every refusal that happens *before* the tool call returns through `fail(..., 2)` with `args.reporter` already parsed, and writes nothing: flow file not found / unreadable, the `..` segment, the trailing separator, the extension and stem complaints, the case-mismatched basename, the remote-routing refusal, and on the directory path `Could not read flow directory`, `No flows found` and the batch's own routing refusal.

Reproduced against the built CLI in a temp project, with a previous run's report already sitting at the reporter path:

```
$ cat junit.xml
<testsuites name="argent flow" tests="4" failures="0" errors="0" skipped="0" time="1.000"/>

$ node -e '… flow(["run","checkout","--reporter","junit:junit.xml"])'
Flow file not found: …/.argent/flows/checkout.yaml
No flow named "checkout" is saved there — run `argent flow list` to see the saved flows.
[exit 2]

$ cat junit.xml
<testsuites name="argent flow" tests="4" failures="0" errors="0" skipped="0" time="1.000"/>
```

Byte-identical. The path is `path.resolve`d and taken literally, which is the fixed location the help text's own example (`--reporter junit:flow-artifacts/junit.xml`) points CI at, so the publisher picks up the last green run's `failures="0"` for a job that exited 2 and ran nothing — the outcome this docstring, and `rejectedFlowReport`'s, both name as the thing a CI artifact must never do.

packages/argent-cli/src/flow-report.ts
```diff
@@ -0,0 +1,1035 @@
...
+      if (measured !== undefined) sysout.push(`durationMs: ${measured}`);
+      if (f) sysout.push(`code: ${f.code}`);
+      out.push(`      <system-out>${xmlEscape(sysout.join("\n"))}</system-out>`);
+    }
+    out.push("    </testcase>");
+  }
+
+  const narration = steps
+    .filter((s) => s.kind === "echo")
+    .map((s) => wireText(s.message))
```
latekvo, line 958: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The suite narration filters on `kind === "echo"` and never looks at `status`. The tool-server pushes un-run steps — echoes included — as `status: "skip"` carrying their `message` (`reportBlockSkipped`, used for both `when block skipped` and `run aborted`), so a skipped echo lands in `<system-out>` byte-identical to one that ran.

`renderEchoLine` draws the distinction deliberately; its comment says a skipped echo "must not print identically to one that ran".

Same report through both renderers on the built dist, a run cancelled after step 1:

```
TERMINAL
  › Signing in
  ✓  1 tap Sign in (0.9s)
  · › Order placed — run aborted
  ·  2 tap Place order — run aborted

JUNIT
    <testcase classname="checkout" name="02 tap Place order">
      <skipped message="run aborted"/>
    </testcase>
    <system-out>Signing in
Order placed</system-out>
```

The XML a CI job publishes says the flow narrated "Order placed" for a run that never reached it. `MIXED_STEPS` pins this shape — its last step is `{ kind: "echo", status: "skip", message: "Done" }` and the assertion expects `<system-out>Opening the cart\nDone</system-out>` for a report whose step 5 errored.

packages/argent-cli/src/flow-report.ts
```diff
@@ -0,0 +1,1035 @@
...
+  // drives no behaviour, so a newer server's vocabulary degrades to prose
+  // rather than to a blank line.
+  const code = wireText(f.code, 64) ?? "unclassified";
+  const step = (f.step ?? {}) as Record<string, unknown>;
+  const category = wireText(f.category, 32);
+  const indeterminate =
+    f.determinacy === "indeterminate" ||
+    (f.determinacy === undefined && /^(condition-|when-guard-|tree-source-)/.test(code));
+
+  const out: NormalizedFailure = {
```
latekvo, line 416: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `normalizeFailure` never reads `f.cause`, and `NormalizedFailure` has no slot for it, so neither the terminal block nor `junitDetailLines` can print it. The MCP twin does (`cause: <code>: <message>`).

`causeOf` attaches the registry `error_code` on every `launch-failed` return in `runLaunch` and the chromium launches, and on every `directive-threw` / `tree-source-unavailable` from `evidenceFromThrow`. `flow-failure.ts`'s module doc states the contract: "its registry code is preserved separately on `FlowStepFailure.cause`, so e.g. a `NATIVE_DEVTOOLS_NOT_CONNECTED` under a flow step surfaces both spellings."

Same payload through all three renderers on the built dist (`code: "directive-threw"`, `cause: { code: "NATIVE_DEVTOOLS_NOT_CONNECTED", message: "helper socket closed" }`):

```
CLI    directive-threw: await-ui-element failed: native devtools is not connected
       screen: unavailable — read-failed: no tree
JUNIT  <error type="directive-threw" message="await-ui-element failed: …">screen: unavailable — read-failed: no tree</error>
       <system-out>status: error / durationMs: 900 / code: directive-threw</system-out>
MCP    cause: NATIVE_DEVTOOLS_NOT_CONNECTED: helper socket closed
```

`cause.code` is not recoverable from anything else on the wire — `<failure type=>` carries the flow taxonomy code and `message` carries only prose. So the registry code, the one machine-greppable identity of the underlying failure, is absent from both CI-facing surfaces while an agent reading the same run over MCP is told it outright. The strategy table this PR ships for `tool-ui-wait-unmet` / `tool-step-failed` / `directive-threw` / `step-kind-unsupported` / `unclassified` says to "Read `message` and `cause` (the underlying error code)".

packages/argent-mcp/src/content.ts
```diff
@@ -279,15 +394,431 @@ function stepIndent(depth: unknown): string {
...
+      const size = isRecord(screen.size) ? screen.size : undefined;
+      const w = wireNumber(size?.width);
+      const h = wireNumber(size?.height);
+      if (w !== undefined && h !== undefined && w > 0 && h > 0) {
+        bits.push(`${Math.round(w)}x${Math.round(h)}`);
+      }
+      if (screen.capturedAt === "after-failure") {
+        bits.push("captured AFTER the failure — the app may have moved on");
+      } else if (screen.capturedAt === "at-failure") {
+        bits.push("captured at the failure");
```
latekvo, line 713: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This branch keys off `capturedAt` alone. The producer sets `capturedAt: "at-failure"` for every screen built from `evidence.tree`, and on the indeterminate tiers that tree is by definition the last read the runner *trusted*, not the state at the deadline — `conditionEvidence` attaches `lastTrustedTree` for `condition-dark-tail`, `condition-hidden-unconfirmable` and `when-guard-indeterminate` alike, and `resolveScreen` checks `evidence?.tree` before its indeterminate branch.

The CLI twin handles it: `normalizeScreen` suppresses the `capturedAt` note when indeterminate, appends `(last trusted read)`, and adds a `reads:` line. No field of `failure.timing` is referenced anywhere in content.ts.

Same payload through both, on the built dist (`condition-dark-tail`, `capturedAt: "at-failure"`, `attempts: 34`, `trustedAttempts: 3`, `darkTailMs: 14203`):

```
CLI  screen: 10 elements, 1080x2220 (last trusted read)
     reads: 34 attempted, 3 trusted, last trusted 14.2s before the deadline
MCP  screen: 10 elements, 1080x2220, captured at the failure, via android-devtools
```

The agent surface asserts a 14-second-stale element list was captured at the moment of failure, and neither number that quantifies the staleness reaches it. The doc table this PR ships describes the MCP behaviour as "(no marker — read `determinacy` instead)"; it is not a missing marker but the opposite one.

packages/argent-mcp/src/content.ts
```diff
@@ -334,10 +886,39 @@ export async function flowRunToMcpContent(
...
+    // and inline the annotated diff image when the assertion failed (once per
+    // run, not once per failing step).
     if (isRecord(step.artifacts)) {
-      blocks.push(...(await stepArtifactBlocks(step.artifacts, step.status, ctx, step.depth)));
+      // A snapshot registers `current` (and `diff`) itself, independently of
+      // the failure diagnostics — so the producer declining `failure.screenshot`
+      // on a secret run removes only a POINTER. Inlining either image here
+      // would hand the agent the same screen as pixels and defeat the omission
+      // whose own text says "a capture of this screen could reveal it".
+      const secretTyped =
```
latekvo, line 897: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `stepArtifactBlocks` correctly declines to inline the image when the run typed a `{{secret:…}}`, but still prints the `baseline:` / `current:` / `diff:` paths — and the failure block below then says the screenshot is withheld because a capture "could reveal" the credential. The CLI closed exactly this with `secretArtifactWarning`, whose docstring says: "Saying 'a capture of this screen could reveal it' beside those paths, with nothing else, reads as a protection that was not applied." There is no counterpart in content.ts.

Same payload through both, on the built dist (`snapshot-diff`, `screenshotOmitted: "secret-typed"`, artifact handles for all three roles):

```
CLI  baseline: /tmp/ev/cart-baseline.png
     current: /tmp/ev/cart-current.png
     diff: /tmp/ev/cart-diff.png
     screenshot: (omitted — a secret was typed onto this device, and a capture of this screen could reveal it)
     warning: the baseline/current/diff images above are of that same screen — review them before publishing as a CI artifact

MCP  baseline: /tmp/ev/cart-baseline.png
     current: /tmp/ev/cart-current.png
     diff: /tmp/ev/cart-diff.png
     screenshot: omitted — a {{secret:…}} value was typed onto this device … read the `tree` file below, whose text is masked.
     (no warning line)
```

The one surface whose reader is an autonomous agent holding a file-read tool is the one that gets the assurance without the caveat — and the `snapshot-diff` row of the strategy table this PR ships tells that agent to "**Review the diff image**". The device latch is never cleared, so this also reaches flows later in a directory run that typed nothing themselves.

packages/argent-mcp/src/content.ts
```diff
@@ -279,15 +394,431 @@ function stepIndent(depth: unknown): string {
...
+/**
+ * What goes in the `screenshot:` slot when the producer declined the capture,
+ * keyed by its `data.screenshotOmitted`. Each ends with what the agent should
+ * do INSTEAD — the failure mode this replaces is an agent reading a missing
+ * image as an oversight and calling `screenshot` itself.
+ */
+const SCREENSHOT_OMISSION_NOTE: Record<string, string> = {
+  "secret-typed":
+    "omitted — a {{secret:…}} value was typed onto this device and a capture of this screen could reveal it. Do NOT call `screenshot` here; read the `tree` file below, whose text is masked.",
+  "no-screen":
```
latekvo, line 604: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This map has two keys; the producer emits three. `flow-failure-report.ts` stamps `screenshotOmitted: "capture-failed"` whenever `attempted.value` was set and the `screenshot` invoke came back without a handle, and its comment gives the reason: "a capture that was ATTEMPTED and came back empty … printed no `screenshot:` line at all … it is the difference between 'argent chose not to' and 'the device could not'." The CLI map has all three.

Same payload through all three renderers on the built dist (`data: { screenshotOmitted: "capture-failed" }`, screen available):

```
CLI    screenshot: (unavailable — the device did not return an image)
JUNIT  screenshot: (unavailable — the device did not return an image)
MCP    (no screenshot line of any kind)
```

The comment above the lookup attributes the silence to "a newer server's vocabulary", but this value ships in the same PR. On the agent surface a failed capture becomes indistinguishable from a tool-server that never had the feature — and the Diagnose procedure this PR ships forbids calling `screenshot` unless the block says `screen: unavailable`, which it does not here, because the tree read succeeded.

packages/tool-server/src/tools/flows/flow-failure.ts
```diff
@@ -0,0 +1,659 @@
...
+ * ever drove.
+ */
+const secretTypedDevices = new Set<string>();
+
+/** Latch a device as having had a credential typed onto its screen. */
+export function markDeviceTypedSecret(deviceId: string): void {
+  secretTypedDevices.add(deviceId);
+}
+
+/** Whether anything this process ran has typed a credential onto `deviceId`. */
```
latekvo, line 545: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This JSDoc says "Whether anything this process ran has typed a credential onto `deviceId`", and the block above says "Devices this PROCESS has typed a `{{secret:…}}` value onto". `markDeviceTypedSecret` has exactly one call site in the repo:

```
$ /usr/bin/grep -rn "markDeviceTypedSecret(" packages/tool-server/src
packages/tool-server/src/tools/flows/flow-failure.ts:541:export function markDeviceTypedSecret(deviceId: string): void {
packages/tool-server/src/tools/flows/flow-run.ts:2514:    if (state.device) markDeviceTypedSecret(state.device.id);
```

The two tools that actually resolve and type credentials — `tools/keyboard/index.ts:228` and `tools/paste/index.ts:83`, both calling `resolveSecretPlaceholders` in `execute` — never mark. So a credential typed by a bare `keyboard-type` / `paste` call, or one carried inside a `run-sequence`, leaves the latch clear; a later `flow-execute` whose own steps contain no `{{secret:` then sees `typedSecret(env)` false and takes the full-resolution capture, which `--output` exports and the MCP renderer inlines into the model's context.

The docblock's own reason for keying on the device — "a credential typed by one flow is still rendered on the device when the next flow runs against it" — holds identically for a credential typed by a bare tool call, which is the case the wording claims to cover.

The reachable sequence: an agent (or a `run-sequence`) calls `keyboard` with `{ text: "{{secret:APP_PASSWORD}}" }`, the app echoes it into a non-secure field — the case `createSecretScrubber`'s header is written for — and no latch is set. A later `flow-execute` against that device fails a step whose own YAML contains no `{{secret:`, so `typedSecret(env)` is false, `captureScreenshot` takes the full-resolution image, and it is both inlined into the model's context and written to the CI artifact directory by `--output`. Every string in that same report is masked correctly, which is what makes the image the only unmasked copy.

packages/skills/skills/argent-create-flow/references/flow-yaml.md
```diff
@@ -192,15 +192,17 @@ A `run:` target is a YAML path resolved against the directory of the flow file c
...
+
+`--reporter junit:<path>` additionally writes a JUnit XML file: one `<testsuite>` per flow, one `<testcase>` per step, and the failure block's slots in the `<failure>` body — so a CI checks UI attributes the failure to the step that produced it. A directory run writes every flow it ran into that one file. Repeat the flag for several outputs; `--reporter default` names the terminal output, which is always on.
```
latekvo, line 205: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This paragraph and the rewritten `--output` sentence above it are the only place the new behaviour is written down. `packages/docs/docs/reference/flow-yaml.mdx` carries the published `argent flow run` option table and is untouched — `git diff origin/main...HEAD --stat` lists no file under `packages/docs`:

```
| `--update-baselines` | Write the snapshot baselines instead of comparing them
| `--output <dir>`     | Write the failed baseline, current and diff images to `<dir>/<flow>/`, for a CI artifact upload
| `-r, --recursive`    | …
| `--json`             | Print the raw JSON report
```

There is no `--reporter` row, and the `--output` row no longer describes what the flag writes now that `step-NN-screen.png` and `step-NN-tree.txt` land there too. `packages/docs/docs/features/flows.mdx` still presents the whole CI story as "prints the result of each step and exits with a status code".

The repo CLAUDE.md asks for the matching `packages/docs/docs/reference/` page in the same PR, and for the description to say so when no docs update is needed; the description does not mention `packages/docs`. The skill reference and the published docs now state different capabilities for the same command, and the flag whose whole purpose is CI adoption is undiscoverable on docs.swmansion.com.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -249,6 +262,22 @@ export interface StepReport {
...
+   * screenshot and element dump captured at the moment of failure. Present on
+   * at most ONE step per run (the runner hard-stops at the first non-passing
+   * leaf) and absent from every passing or skipped report, so an unchanged run
+   * stays byte-identical. `failure.message` duplicates {@link reason}
+   * verbatim, so a renderer that ignores this field prints exactly what it
+   * printed before.
+   */
+  failure?: FlowStepFailure;
+  /**
+   * Wall-clock duration of the step. Omitted on skips (which did no work), so
```
latekvo, line 277: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This says "Omitted on skips (which did no work)". Two `status: "skip"` pushes in this file stamp it anyway — the two when-guard markers, whose `durationMs:` lines sit at 2205 and 2237, both carrying `Date.now() - guardStartedAt`, i.e. the cost of evaluating the guard on a block that by definition did not run.

`renderStepLine` appends the timing suffix on `durationMs` alone, with no reference to `status`, so a skipped `when:` block prints a duration on the terminal step line. Its docstring states this field's contract the other way round — the suffix "appears only when `durationMs` is present, so a pre-timing tool-server renders byte-identically to before" — and that contract holds only if a skip never carries one.

packages/argent-cli/src/flow.ts
```diff
@@ -13,12 +13,122 @@ import {
...
+    matchCount?: number;
+    visibleMatchCount?: number;
+    element?: FlowFailureNode;
+    text?: string;
+    ownText?: string;
+    mismatchPercentage?: number;
+  };
+  screen?: FlowFailureScreen;
+  candidates?: FlowFailureCandidate[];
+  candidateCount?: number;
```
latekvo, line 81: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `candidateCount` is declared on the CLI's wire type and never read again — the one field of this interface with no consumer:

```
$ /usr/bin/grep -rn "candidateCount" packages/argent-cli/src packages/argent-mcp/src
packages/argent-cli/src/flow.ts:81:  candidateCount?: number;
packages/argent-mcp/src/content.ts:304:  candidateCount?: number;
packages/argent-mcp/src/content.ts:684:      const total = wireNumber(failure.candidateCount);
```

`normalizeFailure` does not carry it onto `NormalizedFailure`, so neither the terminal block nor `junitDetailLines` can reach it. The producer sets it to the pre-truncation total precisely so the two numbers can differ — `rankCandidates` caps the emitted list at `FLOW_FAILURE_CANDIDATE_LIMIT` (5) while `failure.candidateCount = ranked.total` keeps the full count.

Same payload through all three renderers on the built dist (`candidateCount: 23`, 5 candidates emitted):

```
CLI    did you mean: {"id":"submitBtn"} · {"text":"Submit"} · … (5 shown)
JUNIT  did you mean: … (5 shown)
MCP    did you mean: … (23 candidates considered)
```

On the two CI surfaces "5 shown" reads as the whole set — the signal that 18 further elements scored above threshold, and that the selector is ambiguous rather than merely wrong, is dropped.

packages/argent-cli/src/flow.ts
```diff
@@ -292,13 +447,27 @@ export function renderEchoLine(s: StepReport): string | undefined {
...
   const where = s.flow && s.flow !== topFlow ? ` [${s.flow}]` : "";
-  const what = s.tool ?? s.target;
-  const label = what ? `${s.kind} ${what}` : s.kind;
-  const reason = s.reason ? ` — ${s.reason}` : "";
+  const label = stepLabel(s);
+  const timing = s.durationMs === undefined ? "" : ` ${formatDuration(s.durationMs)}`;
+  const reason = s.reason && !opts.omitReason ? ` — ${s.reason}` : "";
   const glyph = s.status === "pass" && s.warning ? "⚠" : STATUS_GLYPH[s.status];
-  return `  ${glyph} ${String(n).padStart(2)} ${stepIndent(s.depth)}${label}${where}${reason}`;
+  return `  ${glyph} ${String(n).padStart(2)} ${stepIndent(s.depth)}${label}${where}${timing}${reason}`;
```
latekvo, line 470: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `renderStepLine` interpolates `s.reason` raw. Every other string this module puts on a terminal line goes through a clamp — `stepLabel` truncates to `STEP_LABEL_LIMIT` and strips control characters, and the failure block's fields go through `wireText`.

`reason` is assembled from `error.message` on the tool-server side, so a native error text carrying a raw ESC or a newline reaches the line directly. Rendered on the built CLI with `reason` = a 300-char message containing `ESC[2J` and `\n`:

```
✗  1 tap Sign in — the request failed<ESC>[2J
continues here… (0.9s)
```

The clear-screen sequence executes, and the wrap pushes the `(0.9s)` and the following steps' `✓`/`✗` marks out of column alignment for the rest of the run. `stepLabel`'s own comment gives the reason the same treatment is applied to the label.

packages/tool-server/src/tools/flows/flow-failure.ts
```diff
@@ -0,0 +1,659 @@
...
+ * operator sees when they inspect the screen themselves cannot disagree.
+ *
+ * That means `shouldEmit`, not `hasContent`. `hasContent` is only half of
+ * describe's gate; the other half is the ROLE term, which is what keeps
+ * unlabeled `AXButton`/`AXStaticText`/`AXImage`/`AXTextField` nodes — the
+ * icon-only controls it exists for — on screen. Counting with `hasContent`
+ * alone made the report undercount every screen carrying one, against the
+ * `describe` it tells the operator to compare against.
+ */
+export function isActionableNode(node: DescribeNode, source?: DescribeSource): boolean {
```
latekvo, line 512: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: "mirrors the actionability filter `describe` applies, so the failure block never lists an element the agent could not have targeted" — `describe`'s filter and this one are not the same predicate.

`isActionableNode` here tests `role`/`focusable`/`editable` on the projected node. `fetchTree`'s trimming additionally drops nodes on geometry (zero-size and offscreen frames) and depth, and `fetchFlowTree` — the source this module actually walks — applies neither. So an element clipped out of the agent's `describe` view is still listed as a "did you mean" candidate, and the flow's own matcher will not find it either.

Worth stating what the filter actually is, since the sentence is the contract a reader would rely on when deciding whether a candidate is targetable.

packages/skills/skills/argent-create-flow/references/reliability-and-recovery.md
```diff
@@ -117,17 +117,53 @@ Classify before editing:
...
+| `… (captured after the failure)` | `…, captured AFTER the failure — the app may have moved on` | the step held no tree, so the runner read one straight after it failed    |
+| `… (last trusted read)`          | (no marker — read `determinacy` instead)                    | an indeterminate failure — the screen went unreadable before the deadline |
+| `unavailable — <reason>`         | `unavailable (<reason>) — <detail> [<hint>]`                | there is no element list at all; the reason says why                      |
+
+`screenshot` and `tree` are optional slots, not guarantees. Most of the indeterminate family carries no `tree` — `condition-dark-tail` is the exception: it attaches the last trusted read, so it renders a full element list. And no `screenshot` is captured once a `{{secret:…}}` value has been typed onto the device: pixels are never masked, so the capture is declined for that device from then on, not just for the run that typed it. A failed `snapshot:` still carries its own `baseline`/`current`/`diff` images on the step's artifact lines, and `--output` exports them — review those before publishing them as CI artifacts.
```
latekvo, line 133: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: "Most of the indeterminate family carries no `tree` — `condition-dark-tail` is the exception: it attaches the last trusted read, so it renders a full element list."

`conditionEvidence` attaches the tree unconditionally whenever one exists — the spread is `...(lastTrustedTree !== undefined ? { tree: lastTrustedTree, source, screenSize, readAt } : {})`, outside any per-code branch. So `condition-hidden-unconfirmable` (the `!lastReadTrusted` return) carries a full tree on exactly the same terms, and `when-guard-indeterminate` re-codes that same probe evidence rather than stripping it. All three are in `INDETERMINATE_CODES`, and `normalizeScreen` marks all three `(last trusted read)`.

`condition-never-readable` is the only member of the family that genuinely carries no tree, and only because `lastTrustedReadAt === undefined` returns before the tree is ever attached. A reader who takes the stated exclusivity literally treats a `condition-hidden-unconfirmable` element list as absent when it is present — and, reading the table row two lines above, as current when it is not.

packages/argent-cli/test/flow-report.test.ts
```diff
@@ -0,0 +1,607 @@
...
+        skipped: 0,
+        errored: 0,
+        steps: [],
+      }),
+    ]) {
+      expect(XMLValidator.validate(xml)).toBe(true);
+      for (const suite of parseSuites(xml)) {
+        // The whole content model in one assertion: nothing but the four legal
+        // children, and the counters agree with the elements below them.
+        expect(
```
latekvo, line 531: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The comment directly above states what this assertion is for — "The whole content model in one assertion: nothing but the four legal children, and the counters agree with the elements below them." `expect.arrayContaining([])` asserts neither half: an empty expected array is satisfied by any array, including `[]`, so the sorted key list is compared against nothing.

Confirmed by mutation on the built dist: emitting no children at all from the suite body leaves this case green, and the surrounding cases green too. Nothing else in the file pins the child set, so the content model the comment describes is unpinned — an illegal fifth child, or a missing `<system-out>`, passes here.

packages/tool-server/src/tools/flows/flow-failure.ts
```diff
@@ -0,0 +1,659 @@
...
+  lastTrustedReadAt?: number;
+  darkTailMs?: number;
+}
+
+export interface FlowStepFailure {
+  version: 1;
+  code: FlowFailureCode;
+  category: FlowFailureCategory;
+  /**
+   * The tier the runner used to drop. "the check could not be evaluated" is
```
latekvo, line 319: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: "The tier the runner used to drop." — `determinacy` is not a tier the runner selects; it is derived from the code by `determinacyOf` at report-assembly time, after the runner has already returned.

Same class as three comments in `flow-actions.ts` (1765, 1806, 1808) that still name `lastMatches`, a local this PR deletes, and the `--output` sentence in the skill reference that reads as a change note. Permanent prose is read by someone who never saw the diff; these describe a shape that no longer exists or a decision point that never did.

packages/tool-server/src/tools/flows/flow-failure-report.ts
```diff
@@ -0,0 +1,1085 @@
...
+      ].filter((b) => b !== undefined);
+      return bits.join("  ");
+    });
+    const dir = await failureEvidenceDir();
+    sweepEvidenceDir(dir);
+    // The stem repeats across runs, so the filename carries a unique segment;
+    // the client materializes by the DECLARED filename, which stays stable.
+    const file = path.join(dir, `${stem}-tree-${randomUUID()}.txt`);
+    await fs.writeFile(file, `${lines.join("\n")}\n`, "utf8");
+    return await store.register(file, { mimeType: "text/plain", filename: `${stem}-tree.txt` });
```
latekvo, line 739: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Both `store.register` calls in this file use the two-argument form. `main` no longer has it — #564 (`5d76ed4d6`) collapsed it to a single options object with a required `kind`:

```
$ git show origin/main:packages/registry/src/artifacts.ts | grep -n "async register"
160:  async register(opts: RegisterArtifactOptions): Promise<ArtifactHandle> {

$ git show origin/main:packages/registry/src/artifacts.ts | sed -n '117,122p'
export interface RegisterArtifactOptions {
  /** Absolute path of the file or directory on the tool-server host. */
  hostPath: string;
  /** Semantic category of the artifact, distinct from its MIME type. */
  kind: ArtifactKind;
```

Against this branch's own copy it is `register(hostPath: string, opts?: RegisterArtifactOptions)` with no `kind` field at all, so the branch builds in isolation and both call sites — this one and the overflow spill at 969 — fail `tsc --build` on arity and on the missing required property once rebased.

Past the mechanical fix, these would be the only artifacts in the repo registered without the field main's own module docblock (artifacts.ts:26) calls load-bearing: "MIME type tells a consumer how to read the bytes; kind tells it what the artifact represents." A tree dump and a JSON payload spill are exactly the two whose bytes say least about what they are.

packages/tool-server/src/tools/flows/flow-failure-report.ts
```diff
@@ -0,0 +1,1085 @@
...
+      // Swept here as well as in `registerTreeDump`: the failure shapes that
+      // produce ONLY a spill (any `screen: unavailable`) write into this same
+      // directory, so hanging the sweep off the tree dump alone let those
+      // accrete until some later failure happened to carry a tree.
+      sweepEvidenceDir(dir);
+      const file = path.join(dir, `${stem}-failure-${randomUUID()}.json`);
+      await fs.writeFile(file, JSON.stringify(failure, null, 2), "utf8");
+      trimmed.overflow = {
+        omittedBytes,
+        artifact: await store.register(file, {
```
latekvo, line 969: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `overflow` is written here and at `trimmed.overflow = { omittedBytes }` (1037), and no renderer on any surface reads it:

```
$ /usr/bin/grep -rn "overflow\|omittedBytes" packages/argent-cli/src packages/argent-mcp/src
(no matches)

$ /usr/bin/grep -rn "overflow" packages/tool-server/src/tools/flows/flow-failure.ts
350:  overflow?: { omittedBytes: number; artifact?: ArtifactHandle };
```

So when a failure exceeds `FLOW_FAILURE_BYTE_LIMIT` the block is shed down to fit, the shed bytes are counted, the full payload is written out and registered as an artifact — and the terminal block, the JUnit body and the MCP block all say nothing about any of it. The reader sees `screen: unavailable — omitted-for-size` (or simply fewer elements) with no indication that a complete copy exists.

`--output` exports the `screenshot` / `tree` / snapshot handles; this handle is on `failure.overflow`, which `exportRunArtifacts` does not walk, so the spill does not reach the CI artifact directory either. On a remote `argent link` tool-server the file is on the server host and the handle is the only route to it, which makes it unreachable rather than merely unmentioned.

packages/argent-cli/src/flow-report.ts
```diff
@@ -0,0 +1,1035 @@
...
+ * `<failure>` vs `<error>` split, which argent already draws (`fail` = an
+ * assertion said no; `error` = the machinery broke).
+ *
+ * `echo` steps are narration, not steps: they are excluded from the counters
+ * (as the runner's own `summarize` excludes them), so emitting them as
+ * testcases would make `tests=` disagree with the testcase count. Their text
+ * joins the suite-level `<system-out>` instead.
+ */
+function junitSuite(report: FlowReport, meta: JUnitMeta): { lines: string[]; totals: SuiteTotals } {
+  const flow = wireText(report.flow, 128) ?? "flow";
```
latekvo, line 809: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Suite and testcase identity are both `report.flow` — `<testsuite name="${xmlEscape(flow)}">` here, and `["classname", flow]` on all three testcase paths (873, 891, 912). `report.flow` is the basename stem:

```
packages/tool-server/src/tools/flows/flow-run.ts:1090
  params.flow_path === undefined ? undefined : path.basename(params.flow_path, ".yaml");
```

A recursive directory run is the case that puts several flows in one document, and same-stem flows in sibling platform directories are the ordinary layout — `.argent/flows/ios/login.yaml` and `.argent/flows/android/login.yaml` emit `<testsuite name="login">` twice, with `classname="login"` and identical step names inside.

Consumers that key on classname+name (the Jenkins JUnit plugin, `EnricoMi/publish-unit-test-result-action`, CircleCI, Buildkite) merge or drop one of the pair, so a passing `login/01 tap Sign In` and a failing one collapse to one result and the annotation points at an ambiguous flow. `stepLabel` disambiguates the same collision on the terminal side by appending the flow name when it differs from the top flow; the XML identity has no equivalent.

packages/tool-server/src/tools/flows/flow-failure-report.ts
```diff
@@ -0,0 +1,1085 @@
...
+ *
+ * 0700 because the contents are screen dumps and whole failure payloads: masked
+ * for the secrets THIS run resolved, but still the app's UI text, on a path
+ * every user of the host can otherwise read.
+ */
+const EVIDENCE_TTL_MS = 60 * 60 * 1000;
+
+let evidenceDir: Promise<string> | undefined;
+function failureEvidenceDir(): Promise<string> {
+  evidenceDir ??= (async () => {
```
latekvo, line 124: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `evidenceDir ??=` caches the promise, not the directory — so a rejection is cached with the same permanence as a success. The `mkdir` is the only thing that can reject here, and every caller catches and returns `undefined`, so nothing retries and nothing reports.

If the first failing step of the process runs while `os.tmpdir()` is unwritable — a full disk, a `TMPDIR` pointing at a non-directory, or the shared `argent-flow-failure` path already owned by another user on a multi-user host — that rejected promise is what every later failure in the process gets, indefinitely, after the condition clears.

The docblock immediately above frames the host as exactly the one where this matters: "a host-wide, never-exiting tool-server". What is silently lost for the process lifetime is `failure.tree` (the element dump the skill reference tells the agent to `Read`) and the over-budget spill, on every flow the server runs afterwards, with no line on any surface saying why.

packages/tool-server/src/tools/flows/flow-failure-report.ts
```diff
@@ -0,0 +1,1085 @@
...
+  // holds here too, and it is the difference between "argent chose not to" and
+  // "the device could not".
+  //
+  // `baseFailure` may already have stamped "secret-typed", which outranks both.
+  else if (failure.data?.screenshotOmitted === undefined) {
+    const reason =
+      screenless === true ? "no-screen" : attempted.value ? "capture-failed" : undefined;
+    if (reason !== undefined) failure.data = { ...failure.data, screenshotOmitted: reason };
+  }
+  if (tree !== undefined) {
```
latekvo, line 921: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: There is no `treeOmitted` counterpart to `screenshotOmitted` anywhere in the repo:

```
$ /usr/bin/grep -rn "treeOmitted" packages
(no matches)
```

The dump write and registration are fail-soft — `failureEvidenceDir` rejecting, the write failing, or the budget expiring between the screen read and the dump all return `undefined` — and the `tree:` line then simply does not appear. The screenshot slot beside it treats its own absence as information worth spelling out in three distinct ways, on the stated grounds that "a silently absent screenshot reads as a broken capture".

The result is a block that reports an element count with no dump to read and no reason given, where a reader cannot tell a fail-soft I/O error from a screen argent chose not to dump.

packages/argent-cli/src/flow-report.ts
```diff
@@ -0,0 +1,1035 @@
...
+  // No image, and the producer said why. Rendered in words, since a silently
+  // absent screenshot reads as a broken capture — and the two reasons call for
+  // opposite reader behaviour: "do not take one yourself" vs "there was never
+  // one to take".
+  else {
+    const omitted = wireText(
+      (f.data as Record<string, unknown> | undefined)?.screenshotOmitted,
+      32
+    );
+    if (omitted !== undefined && omitted in SCREENSHOT_OMISSION_NOTE) {
```
latekvo, line 500: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `omitted in SCREENSHOT_OMISSION_NOTE` traverses the prototype chain, so `Object.prototype` keys pass the guard and the lookup returns the inherited function. A `failure.data.screenshotOmitted` of `toString`, `constructor`, `valueOf` or `hasOwnProperty` renders, in the terminal block and inside the JUnit `<failure>` body a CI reader sees:

```
screenshot: function toString() { [native code] }
```

There is no `Object.hasOwn` guard on this lookup, nor on the MCP twin at content.ts:774.

This module's own docblock is why it is worth pinning rather than shrugging at: "a `failure` object arrives from a possibly-remote `argent link` tool-server, so it is untrusted input. It is normalized ONCE, here … clamped, truncated, and coerced". This is the one branch in the normalizer whose stated job is to degrade safely on vocabulary it does not recognise, and on this class of value it does the opposite of degrading safely.

packages/skills/skills/argent-create-flow/references/reliability-and-recovery.md
```diff
@@ -117,17 +117,53 @@ Classify before editing:
...
-4. Classify the cause: selector, screen, missing element, readiness, stale data, optional interstitial, or product behavior.
-5. State the diagnosis in one sentence before correcting it.
+2. **Read the failure block.** The `candidates` are the highest-value part: a selector fix is usually "paste this candidate's YAML into the step". Verify a candidate before committing to it by tapping its `at x, y` centre.
+3. **`Read` the `tree:` file** when the candidates don't settle it. It is the full element list from the failing read — the flow's own full-hierarchy tree, not the trimmed one `describe` returns — so an element missing from a `describe` dump is still in there.
+4. **Do not call `screenshot` / `describe` to see "where the app is now" while the block still answers your question.** The run hard-stopped at the failing step, and by the time you read the report it has also torn down — so a fresh capture shows a **later screen** than the one that failed. Call them only when the report carries **no** failure block (an older tool-server), when the block says `screen: unavailable`, or when you have interacted with the device since the run and genuinely need its current state.
```
latekvo, line 140: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Step 4's exception list authorizes the call on exactly the failures the renderer tells the agent not to make it: "Call them only when the report carries **no** failure block (an older tool-server), **when the block says `screen: unavailable`**, or when you have interacted with the device since the run".

`screen: unavailable` is precisely what a screenless failure produces — `isScreenless` covers `launch-failed` and the whole `composition` category, and `resolveScreen` returns `{ state: "unavailable", reason: "no-screen" }` for them, as it does for a device-free run.

For those same failures the MCP block renders the `no-screen` note, which says the opposite in as many words:

```
omitted — this step failed before it reached the device (a launch that never started, or a flow-
composition error), so no screen belongs to it. A `screenshot` here would show an unrelated app;
fix the flow file instead.
```

Both surfaces ship in this PR and both are read by the same agent. An agent following the numbered procedure calls `screenshot` on every launch and composition failure and diagnoses from whatever app is foregrounded — the outcome the note beside it exists to prevent, and the same wrong-screen class the rest of this section is written against.

packages/argent-mcp/src/content.ts
```diff
@@ -279,15 +394,431 @@ function stepIndent(depth: unknown): string {
...
+
+/**
+ * What goes in the `screenshot:` slot when the producer declined the capture,
+ * keyed by its `data.screenshotOmitted`. Each ends with what the agent should
+ * do INSTEAD — the failure mode this replaces is an agent reading a missing
+ * image as an oversight and calling `screenshot` itself.
+ */
+const SCREENSHOT_OMISSION_NOTE: Record<string, string> = {
+  "secret-typed":
+    "omitted — a {{secret:…}} value was typed onto this device and a capture of this screen could reveal it. Do NOT call `screenshot` here; read the `tree` file below, whose text is masked.",
```
latekvo, line 603: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The note ends by redirecting the agent to a file the block does not always contain: "Do NOT call `screenshot` here; read the `tree` file below, whose text is masked." The `tree:` line is emitted only when a tree artifact resolved (780), and the two conditions are independent.

`screenshotOmitted: "secret-typed"` is stamped in `baseFailure` from the device latch alone:

```
packages/tool-server/src/tools/flows/flow-failure-report.ts:370
    ...(typedSecret ? { data: { screenshotOmitted: "secret-typed" } } : {}),
```

— before any screen is resolved and regardless of whether one ever is. So a `launch-failed`, a `tree-source-unavailable` or a `condition-never-readable` on a device an earlier flow latched renders this note with no `tree:` line anywhere in the block, and no `screen:` element list either.

The note is the entire substitute for the withheld screenshot, and it is the block's only remaining instruction, so the agent is left pointing at a file that was never written — with the one alternative it is explicitly forbidden from using.

packages/argent-mcp/src/content.ts
```diff
@@ -279,15 +394,431 @@ function stepIndent(depth: unknown): string {
...
+  if (identifier) parts.push(`id=${identifier}`);
+  const visibility = frameVisibility(node.frame);
+  if (visibility) parts.push(visibility);
+  const centre = frameCentre(node.frame);
+  if (centre) parts.push(`at ${centre}`);
+  const meta = [wireText(candidate.basis, 40), wireText(candidate.note, 60)]
+    .filter((v): v is string => v !== undefined)
+    .join(", ");
+  if (meta) parts.push(`(${meta})`);
+  const yaml = wireText(candidate.selectorYaml, 120);
```
latekvo, line 580: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `selectorYaml` is the block's paste-able output — flow-failure.ts:292 documents it as "Paste-able straight into the flow file", flow-report.ts calls it "a selector the operator pastes straight back into the flow", and the skill procedure shipped here tells the agent a selector fix is usually "paste this candidate's YAML into the step". Both renderers send it out through the prose sanitizer: `wireText(candidate.selectorYaml, 120)` here, `wireText(c.selectorYaml, 200)` at flow-report.ts:292.

`wireText` is written for prose and does two things to it:

```
const flat = value.replace(/[\s\u0000-\u001F\u007F\u2028\u2029]+/g, " ").trim();
return flat.length <= limit ? flat : `${flat.slice(0, Math.max(0, limit - 1))}…`;
```

The quiet one is the whitespace collapse. `\s` matches U+00A0, and `deriveSelector` only trims the ends, so an internal run of spaces or a non-breaking space survives to the wire and is rewritten on the way out. Running that exact expression:

```
{"text":"Add\u00A0to cart"}  ->  {"text":"Add to cart"}   changed: true
{"text":"Add  to cart"}      ->  {"text":"Add to cart"}   changed: true
```

Both results are valid YAML and paste cleanly, and neither matches the node they were derived from — `includesCI` compares against the label that still holds the NBSP. `resolvesToNode` verified the suggestion against `entry.suggestion.derived`, but a different string is what reaches the reader, so the guarantee its docblock states ("An LLM pastes the suggestion and taps the wrong element, which is worse than offering no suggestion at all") does not cover the value actually rendered.

The loud one is truncation: `suggestSelector` bounds nothing, so a derived text past ~110 characters crosses the 120-char cap here and is cut mid-token with a `…` appended, yielding an unterminated string that `parseFlow` rejects — with nothing in the row saying it was shortened. Chromium's describe caps `accessibleName` at exactly 200, so a full-length name is always past the CLI's own 200-char limit too.

packages/tool-server/src/tools/flows/flow-failure-report.ts
```diff
@@ -0,0 +1,1085 @@
...
+
+const GESTURE_KINDS = new Set(["tap", "long-press", "type", "pinch", "rotate"]);
+function isGestureKind(kind: string): boolean {
+  return GESTURE_KINDS.has(kind);
+}
+
+/**
+ * Keep the serialized payload under {@link FLOW_FAILURE_BYTE_LIMIT}.
+ *
+ * The element list is the only unbounded part worth dropping — code, message,
```
latekvo, line 937: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: "The element list is the only unbounded part worth dropping — code, message, selector, observation, candidates and the artifact handles are what make the report actionable, and together they are small."

`trimToBudget`, the first thing this function calls, sheds four things in sequence, three of them named above as kept: the screen (→ `omitted-for-size`), then `actual.invisibleMatches` — the `observation` — then `candidates` (→ `[]`), then `selector.fields` and `selector.alternatives`. Its own comment states the reason the single-cut model was abandoned: "Dropping only the element list left a payload eight times over budget whenever the weight was somewhere else, while `overflow.omittedBytes` cheerfully reported the few hundred bytes that had actually gone."

So the one place stating the budget stage's contract describes the behaviour the body one screen below was written to replace. The fields at stake are the report's headline outputs — `invisibleMatches` is what both renderers put in their `match:` slot, and `candidates` is what the skill procedure calls "the highest-value part" — and a reader who trusts this docblock will conclude an empty `candidates` array means nothing scored, rather than that the budget dropped them.

packages/tool-server/src/tools/flows/flow-candidates.ts
```diff
@@ -0,0 +1,633 @@
...
+  }
+
+  const limit = Math.max(0, opts.limit ?? FLOW_FAILURE_CANDIDATE_LIMIT);
+  const candidates: FlowFailureCandidate[] = [];
+  // Rectangles already spoken for by an emitted row that carries NO usable
+  // suggestion. Such a row says only "this element, this rectangle", so a
+  // second one for the same rectangle repeats it — the noise `dedupeKey`
+  // exists to remove, which it cannot see until the suggestion has been
+  // verified.
+  const bareFrames = new Set<string>();
```
latekvo, line 556: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `bareFrames` is only ever written inside the `offer === undefined` branch, so it suppresses a bare row after a bare row and never a bare row after an *offered* one. A second row at a rectangle an offered row already described carries the same frame, the same label and the same score with no selector beside it, which is exactly the "says only 'this element, this rectangle'" repetition the set is introduced to remove — the earlier row simply happens to have a suggestion.

The shape that produces it is the one `dedupeKey`'s docblock calls routine: a flattened container and its text leaf at the same frame with the same label. The container is identified, so `{identifier}` verifies and is offered; the leaf's `{text}` lands on a different element, `resolvesToNode` refuses it, and the leaf is emitted bare at the container's rectangle.

Driven against the built `flow-candidates.js` with three leaves — `row-3` + label "Save" at one frame, a second "Save" at that same frame, and a third "Save" elsewhere — and selector `{ text: "Save", role: "AXNoSuchRole" }`:

```
total: 3
{"frame":{"x":0.4,"y":0.8,...},"label":"Save","score":0.8,"yaml":"{\"text\":\"Save\"}"}
{"frame":{"x":0.1,"y":0.3,...},"label":"Save","id":"row-3","score":0.8,"yaml":"{\"id\":\"row-3\"}"}
{"frame":{"x":0.1,"y":0.3,...},"label":"Save","score":0.8,"yaml":null}
```

Rows 2 and 3 are the same rectangle, label and score, and row 3 offers nothing. It costs one of the five `FLOW_FAILURE_CANDIDATE_LIMIT` slots, so a real suggestion is displaced whenever more than five candidates rank above the threshold.

packages/tool-server/src/tools/flows/flow-visual.ts
```diff
@@ -281,6 +296,11 @@ export async function runSnapshot(
           snapshotKey,
           // No crop exists to show, so attach the full capture (already registered).
           artifacts: { current: shot.image },
+          evidence: {
```
latekvo, line 299: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This PR gives four of `runSnapshot`'s six `status: "fail"` returns an `evidence` block (299, 338, 393, 426). The cross-app baseline collision at 254 is left without one, though it is a snapshot failure in the same function with a `snapshotKey` already in hand:

```
  const priorApp = opts.seenKeys.get(snapshotKey);
  if (priorApp !== undefined && priorApp !== opts.appIdentity) {
    return {
      status: "fail",
      reason: `snapshot "..." was already captured in this run from a different app ...`,
      snapshotKey,
    };
```

With no evidence it takes `evidence?.code ?? "unclassified"`, so a snapshot-naming problem is filed as `code: unclassified`, `category: tool`, `determinacy: determinate` — the catch-all bucket — while the taxonomy this PR ships has a `snapshot` category its four siblings use.

`expected` stays empty too — it is filled from `evidence.expected` or `isGestureKind`, and `GESTURE_KINDS` excludes `snapshot` — so the block carries neither the `snapshotKey` nor the `maxMismatch` its siblings do.

And because this return alone supplies no `artifacts.current`, `captureScreenshot`'s reuse branch does not fire and it takes a fresh full-resolution capture of whichever app is now foregrounded, for a failure decided from two baseline keys that never looked at the screen at all.

The trigger is the case the comment above the branch describes: two `snapshot:` steps of the same name either side of a `launch:` that switches app — the shape `flow-snapshot-collision.test.ts` already exercises.

packages/tool-server/test/flows/flow-failure-budget.test.ts
```diff
@@ -0,0 +1,718 @@
...
+    currentFetch = () => ({ tree: screen([]), source: "native-devtools", hint: huge });
+    await writeFlow("huge-hint", {
+      executionPrerequisite: "",
+      steps: [{ kind: "assert", condition: "exists", selector: { text: "Nothing Here" } }],
+    });
+
+    const failure = singleFailure(await run("huge-hint"));
+
+    expect(failure.hint).toBeDefined();
+    expect(Buffer.byteLength(failure.hint!, "utf8")).toBeLessThanOrEqual(
```
latekvo, line 229: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This assertion cannot fail. The fixture plants a 50,000-character adapter hint, but its step is `kind: "assert"`, which `flow-actions.ts` dispatches to `waitForCondition(env, step, DEFAULT_ASSERT_TIMEOUT_MS)`. With every read blind, `lastTrustedReadAt` stays undefined and the early return supplies its own literal hint rather than the adapter's:

```
evidence: conditionEvidence("condition-never-readable", {
  hint:
    "argent never got a trustworthy look at the screen, so this is not a failed " +
    "assertion — check the tree source (devtools connection / app state) and re-run " +
    "rather than editing the flow",
}),
```

That string is 182 characters / **184 bytes**, permanently under the 256-byte `FLOW_FAILURE_FIELD_BYTE_LIMIT` this line asserts against. Remove the cap entirely and the test still passes; the `huge` fixture never reaches `failure.hint`.

The comment above the fixture states the intended path — "A BLIND read is the shape that carries adapter prose: an empty tree plus the source's own hint, which `degraded()` copies straight onto `evidence.hint`" — so the test is pinning a route the `assert` step does not take. The cap matters because `trimToBudget` sheds `screen`, `actual.invisibleMatches`, `candidates` and the selector projections but never `hint`, which makes the field cap the only bound on it.

packages/argent-cli/src/flow.ts
```diff
@@ -13,12 +13,122 @@ import {
...
+export type FlowFailureScreen =
+  | {
+      state: "available";
+      source?: string;
+      capturedAt?: "at-failure" | "after-failure";
+      ageMs?: number;
+      elementCount?: number;
+      elements?: FlowFailureNode[];
+      truncated?: true;
+      /** Device size in points. Absent from a server that does not report it. */
```
latekvo, line 127: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: "Device size in points" is a unit claim the field does not carry, and the two other declarations of the same wire field deliberately avoid making it:

```
tool-server/src/tools/flows/flow-failure.ts:370
  /** Screen size {@link tree}'s frames were normalized against, when the source reports it. */

argent-mcp/src/content.ts:255-259
  /**
   * Screen size the frames were normalized against, in the tree source's
   * native units. Absent from a source (or a tool-server) that does not
   * report it, so every use is presence-gated.
   */
```

The MCP re-declaration of the same field says "native units" in as many words, and the producer's is unit-neutral. On Android the value comes from `devtools.getScreenSize()` (flow-android-tree.ts:241), which is uiautomator device pixels — so on that platform the field is definitively not points, and the three declarations of one wire field disagree.

The number is rendered verbatim into the terminal failure block and the JUnit body, and it is the denominator every normalized frame in the report was divided by. A reader who takes this docblock at face value converts or compares against a points figure, and the `at x, y` centres derived from those frames stop lining up.

packages/argent-mcp/src/content.ts
```diff
@@ -279,15 +394,431 @@ function stepIndent(depth: unknown): string {
...
+  const cx = box.x + box.width / 2;
+  const cy = box.y + box.height / 2;
+  return cx < 0 || cx > 1 || cy < 0 || cy > 1 ? "off-screen" : "visible";
+}
+
+/**
+ * The check the step was making, for every arm the producer emits — not just
+ * `condition` + `text`. Gating the slot on `expected.text` alone meant the
+ * `scroll`, `snapshot`, `gesture` and text-less `condition` shapes rendered
+ * nothing, so `condition: "visible"`, `timeoutMs` and `maxMismatch` never
```
latekvo, line 508: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The docblock lists three things the widened slot now delivers — "so `condition: \"visible\"`, `timeoutMs` and `maxMismatch` never reached the agent at all". `condition` and `maxMismatch` do land; `timeoutMs` has no render path in this file:

```
$ /usr/bin/grep -n "timeoutMs" packages/argent-mcp/src/content.ts
279:    timeoutMs?: number;
508: * nothing, so `condition: "visible"`, `timeoutMs` and `maxMismatch` never
```

Line 279 is the local type declaration and 508 is this comment — `expectedLine`'s `condition` arm reads `expected.text` and `expected.condition` and never touches it.

The producer always sets it: `conditionEvidence` builds `expected: { kind: "condition", condition, ...text, timeoutMs }` unconditionally, and `timeoutMs` is non-optional on `FlowFailureExpectation`. So on an `await: { visible: "Checkout", timeout: 15000 }` that times out, the agent is shown the condition and not the window it was given — the one number that says whether the fix is a longer timeout or a different selector. The CLI's `normalizeExpected`, which this comment says it mirrors, is the surface the claim is true of.

packages/skills/skills/argent-test-ui-flow/SKILL.md
```diff
@@ -108,7 +108,7 @@ Steps:
...
 - If a permission dialog or modal is visible: re-run `describe` first. Stay in screenshot-driven navigation only when the overlay is not exposed reliably, then switch back to `describe` / `debugger-component-tree` as soon as it is dismissed.
 - If tap fails twice at same coordinates: stop, re-discover, report if element not found.
-- If a **saved flow** fails during `flow-execute` replay (as opposed to live test steps above): follow `argent-create-flow`'s [Diagnose a replay failure](../argent-create-flow/references/reliability-and-recovery.md#diagnose-a-replay-failure) — classify the failure, inspect the actual screen, repair the smallest justified unit, then replay the full flow.
+- If a **saved flow** fails during `flow-execute` replay (as opposed to live test steps above): follow `argent-create-flow`'s [Diagnose a replay failure](../argent-create-flow/references/reliability-and-recovery.md#diagnose-a-replay-failure) — classify the failure from the run's `Failures:` block, repair the smallest justified unit, then replay the full flow. The block already carries the screen that failed; do not re-inspect the device for it.
```
latekvo, line 111: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This line drops "inspect the actual screen" and replaces it with "The block already carries the screen that failed; do not re-inspect the device for it." The prohibition is unconditional, and the premise is false for two shapes the block routinely produces.

There is often no screen in the block at all. `resolveScreen` returns `{ state: "unavailable" }` for `launch-failed`, every `composition` code, a device-free run, and any recorded tree-source error — `tree-source-unavailable` and `condition-never-readable` among them. The rendered line is `screen: unavailable (...)` and there is nothing to diagnose from.

When a screen *is* present it is frequently not the one that failed: anything without `evidence.tree` gets a post-hoc `fetchFlowTree`, stamped `capturedAt: "after-failure"`, and the indeterminate tiers carry the last *trusted* read, which the CLI marks `(last trusted read)`.

The sibling reference this line points at draws exactly that carve-out one section down — "Call them only when the report carries **no** failure block (an older tool-server), when the block says `screen: unavailable`, or when you have interacted with the device since the run" — so after this change the two skills give an agent conflicting instructions for the same failure, and the one an agent reads first tells it to diagnose an environment fault with no screen data at all.

PR #583, opened 2026-07-28
fix(describe): stop ordering a simulator reboot on every externally-booted read

### LINKED ISSUE #521: describe repeats a 'MUST call boot-device with force=true' order on every call for externally booted simulators
Every `describe` call on a simulator that argent didn't boot appends this to the response:

> This simulator was not booted through argent — system dialogs and native modals may not appear. You MUST call boot-device with force=true now to restart the simulator and apply full accessibility settings before continuing.

Two problems with it, from running argent as a CLI driven by coding agents (Claude Code) against an Expo dev workflow where the simulator is already up with a dev client, Metro session and staged app state:

1. An agent that takes the instruction literally force-reboots the simulator and destroys the session the developer deliberately booted. "You MUST ... now" is phrased as an order, and agents follow orders. In our runs everything the agent needed (describe, gestures, screenshots, the Metro debugger tools) worked fine without the reboot, so the strong wording buys nothing and risks a lot. We ended up writing a rule into our agent docs telling it to ignore this hint, which is a smell.
2. It repeats on every single `describe`, so a long interaction session pays the ~50 token tax dozens of times and the actual caveat (native modals may not appear) drowns in repetition.

What I'd hope for: say it once per device per server lifetime, drop the MUST/now phrasing in favor of stating the actual limitation ("system dialogs and native modals may not appear; boot through argent if you need them"), and maybe a flag or env var to suppress it entirely for setups that knowingly attach to an externally booted simulator.

Happy to PR the wording change if you'd take it. Otherwise, argent has been a clear upgrade for us over the accessibility-dump MCP tools we used before, the normalized tap coordinates in `describe` especially.

v0.15.0, macOS, iOS simulator, Metro/Expo dev client.

latekvo:
Fixes #521.

## The report holds, and is worse than described

Reproduced against an unmodified `origin/main` tool-server driving a simulator booted the way a developer boots one (plain `xcrun simctl boot`, no argent involvement), iOS 18.6:

```
===== describe call #1 =====   ===== #2 =====   ===== #3 =====
hint: "This simulator was not booted through argent — system dialogs and native
modals may not appear. You MUST call boot-device with force=true now to restart
the simulator and apply full accessibility settings before continuing."
```

Identical on every call, as reported. The part that makes it more than a wording nit is what came back *alongside* it — a complete, working tree:

```
  AXStaticText "19:51"  (0.139, 0.023, 0.102, 0.023)
  AXGroup "Maps" value="Widget"  (0.082, 0.096, 0.395, 0.206)
  ...
```

And the caveat's own claim does not hold either. Triggering a real system permission alert on that same externally-booted sim (`simctl privacy reset location com.apple.Maps` + launch):

```
  AXGroup "Allow “Maps” to use your location?"  (0.193, 0.367, 0.614, 0.296)
  AXButton "Allow Once"  (0.193, 0.524, 0.614, 0.046)
  AXButton "Allow While Using App"  (0.193, 0.570, 0.614, 0.046)
  AXButton "Don’t Allow"  (0.193, 0.617, 0.614, 0.046)
```

The system dialog reads fine, buttons and all. So an agent obeying "You MUST ... now" restarts the simulator — killing the Metro session, dev client and staged app state — to fix nothing.

## Root cause

`AXServiceApi.degraded` is `!entitlementBypassActive`, read from `IgnoreAXServerEntitlements`. That pref is only writable pre-boot, so `boot-device` is the only thing that ever sets it and *any* externally-booted sim reads as degraded for its whole life. The pref bypasses the AX-server entitlement check introduced in iOS 26.5 — on earlier runtimes it is irrelevant to whether the read works. The flag describes how the sim was booted, never whether describe succeeded, and the hint conflated the two.

## Change

Resolve the caveat against what the accessibility read produced:

- **Blind read** — the reboot has something to gain, so the actionable guidance stays, on every call, and now names what the reboot costs.
- **Populated read** — the read is its own proof no reboot is needed. States the limitation instead of ordering anything, once per device per server lifetime.

The per-device gate is applied at the `describe` tool boundary only. `await-ui-element`, the flow runner and `await-screen-idle` poll `describeIos` directly, treat a present `hint` on an empty tree as "this emptiness is untrustworthy" (`isBlindRead`), and fold it into one terminal timeout note — they keep every copy, so a `hidden` wait still cannot resolve off a blind read.

The native-devtools fallback can fill a tree the accessibility read left empty, and that split has to key on the accessibility read rather than on the tree that comes back. Its elements are the injected app's own view hierarchy: no SpringBoard chrome, no system dialog, and no evidence the accessibility subsystem works.

Found doing exactly that on a fresh iOS 18.6 sim. Describing within seconds of `simctl boot` leaves the accessibility read blind for the rest of that boot (3/3; it survives a tool-server restart, and only `boot-device force=true` recovers it). With the Maps location prompt on screen in that state, the fallback returned the app's own 20 elements with the dialog absent — and resolving against *that* tree picked the standing note, which the per-device gate then deleted, so from the second call the agent had a tree silently missing a modal blocking the screen and no caveat at all. The note it briefly carried says "everything else reads normally", which nothing there did. Keying on the accessibility read makes both wordings true of the state that produces them.

I skipped the suppression env var / flag the issue floated as a "maybe" — once the caveat is a single non-directive line per device, there is little left to suppress, and it seemed better not to add config surface for it. Happy to add it if you'd rather have the switch.

## Verification

E2E against the branch build, same externally-booted sim:

| | elements | hint |
|---|---|---|
| describe #1 | 14 | standing caveat, no MUST/now |
| describe #2-4 | 14 | *(none)* |
| after `simctl shutdown` (blind) ×2 | 0 | blind caveat, both times |
| after `boot-device force=true` ×2 | 28 | *(none)* — `IgnoreAXServerEntitlements` = 1 |

Unit: 6 new tests in `describe-tool.test.ts`. Each was mutation-checked — restoring the MUST wording, removing the dedup, collapsing the blind/standing split, keying the gate globally instead of per device, and re-resolving the caveat against the native-devtools tree each fail exactly one test and nothing else.

Full gate: `vitest run` 4715 passed / 1 skipped, `tsc --build`, `typecheck:tests`, `eslint --max-warnings 0`, `prettier --check`, `test:scripts` 59 passed.


packages/tool-server/src/tools/describe/platforms/ios/index.ts
```diff
@@ -13,15 +13,86 @@ import { DescribeTreeData, parseDescribeResult, type DescribeNode } from "../../
...
+const bootCaveatToldDevices = new Set<string>();
+
+export function __resetBootCaveatStateForTests(): void {
+  bootCaveatToldDevices.clear();
+}
+
+/**
+ * Drop DEGRADED_STANDING_HINT from a result whose device has already been told
+ * it. Applied only at the `describe` tool boundary, the one surface that repeats
+ * the caveat to the agent call after call. Internal callers (await-ui-element,
```
latekvo, line 62: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: JSDoc justifies the tool-boundary-only gate with behaviour three of the four named internal callers do not have.

The new JSDoc on `withBootCaveatOncePerDevice` reads: "Internal callers (await-ui-element, flows, await-screen-idle, preview) keep every copy: they gate blind-read detection on `hint` being present and fold it into a single terminal note, so they never pay per poll for it." This is the stated safety argument for putting the dedup only at the tool boundary, and it holds for exactly one of the four listed callers. - **flows** never receive the caveat at all. `fetchFlowTree` routes iOS to `queryFullHierarchyTree`, not `describeIos`, and the repo has a dedicated regression test pinning that (`flow-tree-no-fallback.test.ts` mocks `describeIos` to throw "describeIos must not be reached by a flow tree fetch" and asserts `expect(describeIos).not.toHaveBeenCalled()`). The `data.hint` blindness gate in `flow-actions.ts:1018` reads the full-hierarchy source's hint, never `DEGRADED_BLIND_HINT`/`DEGRADED_STANDING_HINT`.

**Trigger.** A maintainer deciding whether the per-device suppression is safe follows the JSDoc, opens `await-screen-idle` to find the promised "single terminal note", and finds neither a `hint` read nor a note field; the same reader concludes iOS flow runs are handed `DEGRADED_*_HINT` copies when the flow tree source is deliberately built never to produce them, and a test exists specifically to keep it that way.

**Checked against.** ios/index.ts:64-67 — " * the caveat to the agent call after call. Internal callers (await-ui-element,\n * flows, await-screen-idle, preview) keep every copy: they gate blind-read\n * detection on `hint` being present and fold it into a single terminal note, so\n * they never pay per poll for it."  ||  flows/flow-tree.ts:39-41 — `if (device.platform === "ios") { return queryFullHierarchyTree(registry, device); }`; test/flows/flow-tree-no-fallback.test.ts:12-17 — `const describeIos = vi.fn(async (): Promise<DescribeTreeData> => { throw new Error("describeIos must not be reached by a flow tree fetch"); })` and :63 `expect(describeIos).not.toHaveBeenCalled();`  ||  await-screen-idle/index.ts:62- …

packages/tool-server/src/tools/describe/platforms/ios/index.ts
```diff
@@ -13,15 +13,86 @@ import { DescribeTreeData, parseDescribeResult, type DescribeNode } from "../../
 import { adaptAXDescribeToDescribeResult } from "./ios-ax-adapter";
 import { adaptNativeDescribeToDescribeResult } from "./ios-native-adapter";
 
-const DEGRADED_HINT =
-  "This simulator was not booted through argent — system dialogs and native modals may not appear. You MUST call boot-device with force=true now to restart the simulator and apply full accessibility settings before continuing.";
+// `degraded` means the pre-boot accessibility prefs were never written — the one
+// thing boot-device does that an external `xcrun simctl boot` (Xcode, `expo
+// run:ios`, a developer's own simulator) cannot. It describes how the simulator
+// was booted, not how the read went: `IgnoreAXServerEntitlements` only bypasses
+// a check that exists on iOS 26.5+, so on earlier runtimes an externally-booted
```
latekvo, line 20: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: New comment infers "earlier runtimes serve the complete tree" from one pref, ignoring the three others boot-device writes.

The new header comment reasons that because `IgnoreAXServerEntitlements` "only bypasses a check that exists on iOS 26.5+", it follows that "on earlier runtimes an externally-booted sim serves the complete tree, system alerts included". `degraded` is `!entitlementBypassActive`, i.e. a proxy for "boot-device's pre-boot write never ran" — and that write sets four prefs, not one. `ax-prefs.ts` documents, with its own reproduction, that two of the other three independently blank the tree on any runtime: without `AccessibilityEnabled` + `ApplicationAccessibilityEnabled`, SpringBoard never spawns `AccessibilityUIServer` and "describe returns an empty ROOT even though the entitlement check passes". The PR body reports exactly that outcome on an earlier runtime — a fresh iOS 18.6 sim whose AX read stayed blind 3/3 for the whole boot and recovered only via `boot-device force=true`. So the generalisation over "earlier runtimes" is contradicted both by a sibling module's documented repro and by this PR's own field evidence; only the parenthetical single observation is supported.

**Trigger.** A maintainer reading this comment to reason about a blind read on an externally-booted iOS 18.x sim concludes the AX subsystem must be fine on that runtime and looks elsewhere for the cause, when the three unwritten prefs the comment omits are the documented mechanism.

**Checked against.** ios/index.ts:19-22 — "`IgnoreAXServerEntitlements` only bypasses a check that exists on iOS 26.5+, so on earlier runtimes an externally-booted sim serves the complete tree, system alerts included"  ||  utils/ax-prefs.ts:78-82 — "`AccessibilityEnabled` + `ApplicationAccessibilityEnabled` gate the AT subsystem bootstrap. Without them SB never spawns `AccessibilityUIServer` and describe returns an empty ROOT even though the entitlement check passes (reproduced on a wiped iPhone 17e: AccessibilityUIServer active count = 0 without these two…)"; :97-104 shows all four keys written pre-boot  ||  blueprints/ax-service.ts:429 — `degraded: !entitlementBypassActive` (reads only `IgnoreAXServerEntitleme …

latekvo (commented):

packages/tool-server/src/tools/describe/platforms/ios/index.ts
```diff
@@ -13,15 +13,86 @@ import { DescribeTreeData, parseDescribeResult, type DescribeNode } from "../../
...
+  "and ends anything running on it.";
+
+// Emitted once per device per server lifetime by `withBootCaveatOncePerDevice`:
+// it holds for every describe against this simulator until the sim is booted
+// through argent, and a session makes dozens of describe calls.
+const DEGRADED_STANDING_HINT =
+  "This simulator was not booted through argent, so system dialogs and native modals may not appear " +
+  "in this tree; everything else reads normally. If something you expect is missing, boot-device with " +
+  "force=true reboots the simulator with the full accessibility settings — at the cost of restarting " +
+  "it and everything running on it.";
```
latekvo:
```suggestion
  "force=true reboots the simulator with the full accessibility settings";
```

packages/tool-server/src/tools/describe/platforms/ios/index.ts
```diff
@@ -13,15 +13,86 @@ import { DescribeTreeData, parseDescribeResult, type DescribeNode } from "../../
...
+// Restarting the simulator costs the developer whatever is running on it — a
+// Metro session, a dev client, staged app state — so the caveat splits on what
+// the accessibility read actually produced. Only a read that came back blind
+// has anything to gain from the reboot; a populated accessibility read is its
+// own proof the reboot is not needed, and says so without ordering one.
+const DEGRADED_BLIND_HINT =
+  "The accessibility read returned no elements, and this simulator was not booted through argent, " +
+  "so the pre-boot accessibility settings were never applied. Unless the screen is genuinely blank, " +
+  "call boot-device with force=true to reboot it with those settings — this restarts the simulator " +
+  "and ends anything running on it.";
```
latekvo:
```suggestion
  "The accessibility read returned no elements, and this simulator was not booted through argent, " +
  "so the pre-boot accessibility settings were never applied. Unless the screen is genuinely blank, " +
  "call boot-device with force=true to reboot it with those settings — this restarts the simulator";
```

PR #587, opened 2026-07-29
feat(screen-recording): record through simulator-server instead of host ffmpeg

latekvo:
Moves screen recording off the host `ffmpeg` pipeline and onto simulator-server's own recorder.

> **Blocked on** software-mansion/radon#155, which adds the endpoint and enables it in the `argent` build variant. Until a release carries it, every simulator-server answers 404 and this branch behaves exactly like `main` — that fallback is deliberate and covered below, but the win only lands once #155 ships.

## Why

simulator-server already encodes every frame of the device screen for its live stream, with the touch overlay drawn in. Recording subscribed to that MJPEG stream, decoded it on the host, paced it in Node and re-encoded it through an ffmpeg child. So a recording cost two encodes with a lossy JPEG round-trip in between, and required an external binary the user had to install.

It now asks simulator-server to record instead: it paces to 30fps, trims static stretches, stamps the watermark and muxes the mp4 from the frames it already holds.

## What stays identical

Everything the tools present. Same admission guards (double start, start-after-cap, concurrent stop), same per-call reminder note, same `outputFile` returned by start, same `durationMs`/`wallClockMs`/`trimmedMs`, same warnings, same `.argent/recordings` artifact with the `argent-` prefix stripped. The touch visualizer is still argent's to toggle for the life of a recording, since it is server-global state.

Two details worth calling out:

- **The video is copied out of simulator-server's session directory** rather than referenced there. That directory is wiped when the server exits, and the artifact is materialized by the client afterwards — possibly downloaded over `argent link` much later.
- **The time limit is enforced by the server**, and a host timer mirrors it purely as bookkeeping: it flips the reminder to "ended, still to retrieve" and makes a start-after-cap fail instead of discarding the finished video. It must not stop the recording — that is `screen-recording-stop`'s job, whenever it arrives.

## The fallback

The endpoint exists only where simulator-server can encode: the software-encoder builds compile the routes out, and no simulator-server predating radon#155 has them either — including the one argent currently pins. Those answer 404 and `startCapture` runs the existing host pipeline unchanged. That path is untouched here, and `ffmpeg` is required exactly there.

Which one you get is a property of the build, not of the host OS, so the tool description and SKILL.md say that rather than naming platforms. They would otherwise be wrong today in the most visible direction: the shipped macOS simulator-server has no recording route, so macOS falls back and *does* need ffmpeg.

Detecting the absent route needed care: it is answered by the router's unmatched-route fallback with an **empty body**, which the shared `simulatorPost` helper would surface as "non-JSON response" — indistinguishable from a server in a bad state. `recordingPost` tests the status *and* the emptiness. A 404 carrying a body came from a handler, so the route exists and refused the command; falling back there would start a second capture over a recording that is already running. Failed recording commands are the other case: HTTP 200 carrying an `error` field.

## Found in review

Each reproduced before it was changed, and each pinned by a test that fails without the fix.

- **A start suspended at `pointer.enable()` resumed past a dispose that had already stopped the server's recording**, reported `{status: "recording"}` for it, and armed a cap timer no later stop or dispose could clear. The start request's own await was guarded; the pointer await, the one suspension point after the session is stamped, was not.
- **`serverStop` stayed set across the stop request**, so a dispose landing mid-stop issued a second, concurrent stop for the recording that call was already finalizing — and that stop takes down the simulator-server whose session directory the copy reads from.
- **A failed copy lost the video silently.** It threw a raw filesystem error naming two temp paths, with no failure signal, while the only finished video sat in simulator-server's session directory waiting to be deleted with it. It now names that path while it is still fetchable, classified like the empty-output failure two calls down.
- **The recording commands had no request timeout.** A simulator-server that accepts the connection and never answers held `startPending`/`stopPending` — and so every other recording call on the device — for undici's 300 s header timeout. Measured at 301011 ms per call, twice per recording; the pointer toggles five lines away were already bounded at 2 s.
- **The 404 fallback keyed on the status alone** while its own comment described an empty body. Verified against the shipped macOS simulator-server, which has no recording route: `POST /api/recording/start` answers 404 with `content-length: 0`.
- **`res.json().catch(() => null)` reported a body that never finished arriving, and any non-JSON reply, as "simulator-server rejected the recording command"** — sending the caller after a refusal reason that was never sent. Split out the way `simulatorPost` already splits it.
- **The ffmpeg prose was keyed on the host OS** while the code keys on the route being present, and the two disagree today (above).

## Verification

3095 tool-server tests pass (295 files), plus tsc for `src` and `tsconfig.test.json`, eslint and prettier.

25 new tests in `screen-recording.test.ts` (62 → 87). Each of these mutations turns the suite red, so none of them is passing on incidental behaviour:

| mutation | tests failing |
|---|---|
| server path never chosen | 10 |
| stop dispatches to the host path | 8 |
| dispose leaves the server recording running | 1 |
| post-dispose start race guard removed | 1 |
| post-pointer dispose guard removed | 1 |
| `serverStop` left set across the stop await | 1 |
| copy failure throws a plain `Error` | 1 |
| recording commands sent without a timeout | 1 |
| any 404 falls back | 1 |
| any empty-bodied non-2xx falls back | 1 |
| back to `res.json().catch(() => null)` | 2 |
| trim fields always reported | 1 |
| server's copy of the video not released | 1 |
| cap timer stops the recording early | 1 |
| ambiguous start reply treated as success | 1 |

**End-to-end on an iOS 18.6 simulator**, tool-server driving a real simulator-server (native framebuffer 1179×2556):

*Server path* (`--features argent` build of radon#155): start → appearance flip → `gesture-tap` → 5s static → stop.
- `durationMs: 2500`, `wallClockMs: 9908`, `trimmedMs: 7408` — trimming engaged
- mp4: h264, **1178**×2556 (even-cropped), `r_frame_rate=30/1`, **75 frames / duration 2.500000** — exactly the reported `durationMs`
- **zero mentions of ffmpeg in the tool-server log**; no encoder was spawned
- watermark legible in an extracted frame, contrast-matched to the UI underneath
- the tap's translucent bubble is present in the video (frame 31), so the touch overlay still lands
- simulator-server's media directory is **empty** afterwards — the video was moved out, not left behind
- the artifact carries `saveDir: ".argent/recordings"` and the prefix-stripped filename

*Fallback path* (today's shipped `simulator-server-argent-macos`, `strings … api/recording` = 0): same tool calls, start fell through the 404 to the host pipeline → h264 1178×2556, 30/1, 176 frames, `duration=5.866667` matching `durationMs: 5867`.

## Summary by CodeRabbit

- **New Features**
  - Screen recordings can now be created and finalized by the simulator server when supported.
  - Builds without server-side recording continue using local encoding.
  - Time limits, watermarking, and static-content trimming work across both recording paths.
  - Improved recovery for interrupted recordings, including safer stop retries and clearer handling when video files are unavailable.

- **Documentation**
  - Updated guidance explains recording paths, requirements, timing, and recovery behavior.


packages/tool-server/src/tools/screen-recording/server-capture.ts
```diff
@@ -0,0 +1,275 @@
...
+    const empty = await fs
+      .stat(outputFile)
+      .then((s) => s.size === 0)
+      .catch(() => false);
+    if (empty) await fs.rm(outputFile, { force: true }).catch(() => {});
+    throw err;
+  } finally {
+    // Always return the session to a startable state: a failed hand-off must not
+    // wedge the next start behind "already active". simulator-server dropped its
+    // recording when `stop` returned, so there is nothing a retry could recover.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: A stop whose `stop()` call throws orphans the live server-side recording and drops the only handle that could stop it.

`stopServerCapture` hands ownership over before the await (`api.serverStop = null` at line 171) and its `finally` re-nulls `serverStop`, `outputFile`, `pendingRetrieval` and clears the reminder unconditionally — including when `await stop()` itself threw. The justifying comment ("simulator-server dropped its recording when `stop` returned, so there is nothing a retry could recover") is only true when `stop` RETURNED. When it throws, simulator-server is still recording and argent has thrown away every handle to it: `dispose()` skips it (its guard is `if (state.serverStop)` — screen-recording-session.ts:253, added by this PR precisely because "a recording running inside simulator-server outlives this process"), a retried `screen-recording-stop` gets "No active screen recording" (`recordingActive` and `pendingRetrieval` are both false), the per-call NOTE is gone so the agent is never told, and a new `screen-recording-start` is refused hard rather than falling back — `recordingPost` throws on the HTTP 200 `{error: "a recording is already running"}` reply this PR's own test asserts (scree …

**Trigger.** Server-side recording is live; `screen-recording-stop` is called against a simulator-server that accepts the connection and never answers (exactly the case `RECORDING_STOP_TIMEOUT_MS` was added for) — or answers 500/non-JSON mid-mux. `stopServerRecording` rejects at the 60 s AbortSignal, the FailureError propagates, and the `finally` clears `serverStop`. The server keeps recording; dispose will not stop it, stop cannot be retried, and every later `screen-recording-start` on that device fails with "a recording is already running".

**Checked against.** server-capture.ts:171 `api.serverStop = null;` (before the try), 178 `const result = await stop();`, 257-265 `} finally {  // Always return the session to a startable state ... simulator-server dropped its\n    // recording when `stop` returned, so there is nothing a retry could recover.\n    await disablePointer(api); ... api.serverStop = null;` — vs the reason the dispose branch exists, screen-recording-session.ts:250-256 `// A recording running inside simulator-server outlives this process, and\n// it accumulates frames until something stops it — so end it here even\n// though the video is being abandoned.\nif (state.serverStop) { ...

packages/tool-server/src/blueprints/screen-recording-session.ts
```diff
@@ -233,6 +247,15 @@ export const screenRecordingSessionBlueprint: ServiceBlueprint<
           await disable().catch(() => {});
         }
 
+        // A recording running inside simulator-server outlives this process, and
+        // it accumulates frames until something stops it — so end it here even
+        // though the video is being abandoned.
+        if (state.serverStop) {
+          const stop = state.serverStop;
+          state.serverStop = null;
+          await stop().catch(() => {});
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: dispose awaits the full 60 s stop budget for a video it discards, contradicting its own documented 1.5 s salvage policy.

The constants block 130 lines above states the dispose contract: "an in-flight recording is being abandoned, so the video is a best-effort salvage rather than something a caller is waiting on ... give that one short grace before SIGKILL. A caller that wants the file calls `screen-recording-stop`, which has its own (longer) finalize contract" — DISPOSE_FINALIZE_GRACE_MS 1_500 + DISPOSE_REAP_MS 1_000. The new branch instead invokes exactly that longer contract: `state.serverStop` is `() => stopServerRecording(simulator, AbortSignal.timeout(RECORDING_STOP_TIMEOUT_MS))` with RECORDING_STOP_TIMEOUT_MS = 60_000, awaited inline, on top of the ~2 s `pointerDisable` above it. So a dispose can block ~62 s per device to obtain a video it then reports as unrecoverable ("nothing was written to ... — re-record rather than looking for it", line 311).

**Trigger.** A server-side recording is live when an agent calls `stop-all-simulator-servers` at session end (ScreenRecordingSession is inserted into the registry before SimulatorServer, so it is disposed first and the request reaches a live server). If that simulator-server is slow to mux or wedged, the sweep — and on shutdown, the whole registry teardown — stalls up to 60 s on this one device, versus the 2.5 s the same dispose budgets for the host ffmpeg child.

**Checked against.** screen-recording-session.ts:117-125 `// ... the video is a best-effort salvage rather than something a caller is waiting on ... give that one short grace before SIGKILL. A caller that wants the file calls `screen-recording-stop`, which has its own (longer) finalize contract.\nconst DISPOSE_FINALIZE_GRACE_MS = 1_500;\nconst DISPOSE_REAP_MS = 1_000;` vs line 256 `await stop().catch(() => {});`; screen-recording-start.ts:42 `const RECORDING_STOP_TIMEOUT_MS = 60_000;` and 230 `stop: () => stopServerRecording(simulator, AbortSignal.timeout(RECORDING_STOP_TIMEOUT_MS))`; stop-all-simulator-servers.ts:102-110 `// ...

packages/tool-server/src/blueprints/screen-recording-session.ts
```diff
@@ -66,8 +68,15 @@ export interface ScreenRecordingSessionApi {
...
-  /** Host path ffmpeg is writing the video to. */
+  /** Host path the finished video lands at. */
   outputFile: string | null;
+  /**
+   * Finalizes a recording simulator-server is running and hands back the muxed
+   * video. Set only while a server-side capture is live, so it doubles as the
+   * marker for which side owns the recording: with it set there is no capture
+   * child, no frame stream and no pump, and stop goes to the server.
+   */
+  serverStop: (() => Promise<ServerRecordingResult>) | null;
```
latekvo, line 81: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The two permanent docs that enumerate what a ScreenRecordingSession owns and how it is captured were left describing only the ffmpeg path.

`serverStop` adds a resource the session owns that no previous one resembled: a recording running inside another process that outlives the tool-server (the new dispose branch exists precisely because of that). Two pieces of permanent prose whose job is to enumerate exactly this were not updated and are now wrong. `device-services.ts` documents each entry of DEVICE_OWNED_NAMESPACES by the resources it owns and states that the list is "about what `stopped` names"; its ScreenRecordingSession bullet still reads "owns an ffmpeg child, an MJPEG frame stream, and the touch-visualizer overlay it enabled on the device" — for a server-path recording there is no ffmpeg child and no MJPEG stream, and the one resource that matters most for a teardown (a recording inside simulator-server) is unlisted.

**Trigger.** A maintainer reading DEVICE_OWNED_NAMESPACES to answer "what does a scoped stop-all-simulator-servers actually reap on this device?" is told an ffmpeg child and a frame stream, and would not learn that the entry also ends a recording living inside simulator-server — the one resource that survives the tool-server process if the entry is ever dropped from the list.

**Checked against.** New field: screen-recording-session.ts:73-79 `/** Finalizes a recording simulator-server is running and hands back the muxed video. Set only while a server-side capture is live... */ serverStop: (() => Promise<ServerRecordingResult>) | null;` and its dispose branch at 250-257. Stale siblings: packages/tool-server/src/tools/simulator/device-services.ts:108-109 ` * - `ScreenRecordingSession` owns an ffmpeg child, an MJPEG frame stream, and\n *   the touch-visualizer overlay it enabled on the device.`; packages/registry/src/failure-codes.ts:214-216 `// screen-recording-start / screen-recording-stop.

packages/tool-server/src/tools/screen-recording/capture.ts
```diff
@@ -522,6 +524,10 @@ export async function stopCapture(api: ScreenRecordingSessionApi): Promise<StopR
   // finalizes (see assertStoppableSession / assertNoActiveRecording).
```
latekvo, line 537: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The post-`pointer.enable()` dispose guard added to the server path is missing on the host path, which has the identical suspension point.

The PR identifies and fixes this exact bug for server captures — server-capture.ts:117-121 `// Enabling is the one suspension point left after the session is stamped, so it needs the same guard the start request got ... a session that resumes here would report a live recording dispose has already stopped, and arm a cap timer no later call can clear` — but leaves the twin in `startCaptureLocked` unguarded. On the host path the same await is followed straight by `api.recordingTimeout = setTimeout(...)` and `return { status: "recording", ... }`, with no `assertNotDisposed`. Result: `screen-recording-start` reports success for a capture dispose already SIGKILLed, and arms a timer up to 600 s that no later stop or dispose can clear (a later stop throws on the reaped breadcrumb; dispose has already run). The timer is inert thanks to its `api.captureProcess !== child` ownership guard, so the reachable damage is the false `{status: "recording"}` plus a pending handle.

**Trigger.** Host-path recording with `showTouches` on (the default); `stop-all-simulator-servers` or process shutdown disposes the ScreenRecordingSession while the start is suspended at `await params.pointer.enable()`. dispose kills the ffmpeg child (`pendingChild.kill("SIGKILL")` — `startCapture`'s finally has not run yet) and clears the reminder; the start then resumes, arms the cap timer and returns `{status: "recording", outputFile}` for a recording that no longer exists.

**Checked against.** capture.ts:432-442 `if (params.pointer) { ... api.pointerFailed = !(await params.pointer.enable()); }\n\n  api.recordingTimeout = setTimeout(() => {` — no guard between them; compare server-capture.ts:115-121 `api.pointerFailed = !(await params.pointer.enable());\n\n    // Enabling is the one suspension point left after the session is stamped ...\n    if (api.disposed) assertNotDisposed(api, "screen_recording_start");`.

latekvo (commented):

latekvo:
Heads-up on a wire change landing in the sim-server side, software-mansion/radon#155, from review there: `/api/recording/stop` is now keyed by an id.

    POST /api/recording/start {}   -> {"status":"ok","id":"18ccb2ca653cf560"}
    POST /api/recording/stop {"id":"18ccb2ca653cf560"} -> { path, url, size_bytes, ... }

The reason is the one this client's own comments describe: a sim-server is shared, only one recording runs at a time, and `start` already refuses a second one — so a bare `stop` from another process on the machine ended and collected a recording it never started. It happened to the reviewer during review.

For `recordingPost` this means `stop` needs the id `start` returned carried through, and the start response is now read for `id` rather than only checked for `status`. The two failure shapes it keys on are unchanged and deliberately so:

- an empty-bodied 404 still means "this build has no recording API" (the routes stay compiled out on `swenc`; I tried answering `200 {"error"}` there and backed it out precisely because it would break this fallback)
- a `stop` without an id answers `200 {"error":"stop needs the id that start returned"}`, not the extractor's 422, so an un-updated client gets something actionable rather than a parse failure

packages/tool-server/src/tools/screen-recording/server-capture.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Confirmed real — reproduced before touching anything: with `stop()` rejecting (the timeout case the 60 s budget exists for), the `finally` nulled `serverStop`, so dispose skipped it, a retried `screen-recording-stop` was refused with "No active screen recording", and every later start died on simulator-server's own "already running" guard.

Fixed in 357d2296: the failure of the stop *request* is now distinguished from a failure of the hand-off (`finalizeRequestFailed` in `stopServerCapture`). That one case keeps the recovery handles — `serverStop` restored, `pendingRetrieval` set, reminder flipped to "ended, still to retrieve" instead of cleared — so both recovery paths work: a retried stop finalizes, dispose ends the recording. Every other failure (copy out, missing/empty output) keeps the old full reset, since there the server really has dropped the recording.

Pinned by "keeps the recording recoverable when the stop request itself fails" (asserts `api.serverStop === server.stop` after a rejected stop, then recovers through a retried stop) and "still ends the recording on dispose after its stop request failed"; both were red before the fix.

packages/tool-server/src/blueprints/screen-recording-session.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Confirmed real — reproduced first: a dispose against a simulator-server whose salvage `stop()` never settles made the teardown unresolvable past the 5 s mark (the test parks the stop and advances virtual time; pre-fix `disposed` never flipped), i.e. up to the 60 s abort for a video dispose then reports as unrecoverable.

Fixed in 357d2296: the salvage stop in `dispose` now races `DISPOSE_SERVER_STOP_MS = 1_500` — the same short grace the ffmpeg finalize path gets, and the policy the constants block documents. The request itself keeps running under its own abort bound; only the teardown's wait on it is bounded.

Pinned by "bounds dispose's salvage stop at the short grace, not the stop tool's budget", which times out on the unfixed code.

packages/tool-server/src/blueprints/screen-recording-session.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Non-issue on the current branch: the stale prose this points at no longer exists. main's #931 (b2321144, carried by the merge in 331e50b2) rewrote device-services.ts and dropped the per-namespace ownership bullet list entirely — including the "owns an ffmpeg child, an MJPEG frame stream…" ScreenRecordingSession line (`git diff 2ffbb5f6 HEAD -- …/device-services.ts` shows it removed). The file now names the namespace only as a member of `DEVICE_OWNED_NAMESPACES`, with no per-resource claims left to go stale.

packages/tool-server/src/tools/screen-recording/capture.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Confirmed real — reproduced first on the host path: parking `pointer.enable()`, disposing once the start was suspended there, then releasing it returned `{status: "recording", timeLimitSeconds: 180}` for an encoder dispose had already finalized, with a fresh cap timer armed that no later stop or dispose could clear.

Fixed in 357d2296: `startCaptureLocked` now checks `if (api.disposed) assertNotDisposed(api, "screen_recording_start")` right after the enable await, mirroring server-capture.ts.

Pinned by "does not report success for a host capture dispose stopped during pointer arming" (touch visualizer suite), which resolved with success instead of rejecting on the unfixed code.

packages/tool-server/src/tools/screen-recording/server-capture.ts
```diff
@@ -0,0 +1,313 @@
...
+  try {
+    if (api.recordingActive) {
+      api.recordingActive = false;
+      api.wallClockEndMs = Date.now();
+    }
+    let result: ServerRecordingResult;
+    try {
+      result = await stop();
+    } catch (err) {
+      finalizeRequestFailed = true;
```
hubgan: Every rejection from `stop()` is classified here as a failed *request*, but `stopServerRecording` / `recordingPost` also reject for replies simulator-server definitely sent — an empty-bodied 404, a body with no `path`/`duration_ms`, and `!res.ok || !body || body.error` (e.g. HTTP 200 `{"error": "unknown recording id"}`) — and for a connection refused against a simulator-server that is gone. For all of those the recording provably cannot be recovered, yet the recovery branch below parks the session at `pendingRetrieval` with `serverStop` restored on the premise that a retry might still get the video.

`pendingRetrieval` has no other exit: `assertNoActiveRecording` blocks every later start, the session declares no `recoverable`, and the restored closure still holds the old `SimulatorServerApi` and a recording id that no longer exists.

Reproduced end to end on a real iOS simulator, tool-server built from this branch driving a simulator-server that serves `/api/recording/*`:

1. `screen-recording-start {udid}` — server path taken, no ffmpeg spawned.
2. simulator-server dies (I killed the process; an HTTP 200 `{"error": ...}` stop reply gives the same result).
3. An ordinary `gesture-tap` transparently respawns simulator-server and returns 200 — the blueprint is `recoverable` on connection-refused, so this needs no teardown call from the agent.
4. `screen-recording-stop {udid}` → `SIMULATOR_NETWORK_CONNECTION_REFUSED` against the dead port. Retried: identical, indefinitely.
5. `screen-recording-start {udid}` → `SCREEN_RECORDING_ALREADY_ACTIVE`: "…Call `screen-recording-stop` to hand it over before starting a new recording" — an instruction that can never succeed.
6. `stop-simulator-server {udid}`, which the stop failure itself recommends, does not clear it (it disposes `SimulatorServer` but not `ScreenRecordingSession`). Only `stop-all-simulator-servers` does, and neither message mentions it.

Blast radius while wedged, all observed:
- Screen recording is unusable on that device for the tool-server's lifetime.
- The reminder entry is never cleared, so every tool result for every agent sharing this tool-server carries `NOTE: The screen recording on device <udid> already ended … but its video has not been retrieved yet` forever.
- With the server still alive, the finally deliberately skips `disablePointer`, so the server-global touch visualizer stays armed — I counted the pointer traffic: `trail` + `show:true` at start and no `show:false` after the failed stop, so every later screenshot and recording on that device shows touch markers.

The same run on the host fallback path returns the video with a "frame stream dropped" warning and leaves the session immediately startable, so this is a narrowing specific to the new path. It is gated behind radon#155 shipping the route, but an ordinary simulator-server crash is enough to trigger it once that lands.

One more thing the wedged state gets wrong: `assertNoActiveRecording` has no case for it, so it falls to its last arm and every refused start says the previous recording "already ended (time limit or unexpected exit)" — neither of which happened, and it may not have ended at all.

packages/tool-server/src/blueprints/screen-recording-session.ts
```diff
@@ -196,37 +221,58 @@ export const screenRecordingSessionBlueprint: ServiceBlueprint<
...
+        // Whether this dispose is destroying an unretrieved capture, decided
+        // BEFORE the teardown below clears the flags it is read from. Both
+        // states owe the caller a video: one is still encoding, the other
+        // finished and is waiting to be handed over.
         const hadUnretrievedCapture =
           state.recordingActive || state.startPending || state.pendingRetrieval;
         const abandonedOutput = state.outputFile;
+        // Which side held the video, read before the teardown below hands
+        // `serverStop` over — the two paths owe the caller opposite stories.
+        const serverCapture = state.serverStop !== null;
```
hubgan: `stopServerCapture` hands `serverStop` over synchronously (server-capture.ts:176) before its first suspension point, so a dispose landing during `await stop()` — a window up to `RECORDING_STOP_TIMEOUT_MS`, 60s — reads `null` here and classifies a server capture as a host one. The breadcrumb it then records is the ffmpeg salvage story, which points the caller at a host path the server path never writes.

Reproduced: start on the server path with `timeLimitSeconds: 2`; let the cap fire, so `pendingRetrieval` keeps `hadUnretrievedCapture` true; call `screen-recording-stop` against a simulator-server that does not answer, and run `stop-all-simulator-servers {devices:[udid]}` while it is in flight. The retried stop answers:

"…ffmpeg was given a moment to finalize the container first, so the video captured up to that point is usually playable at /var/folders/…/argent-screen-recording-<udid>-1788345242353.mp4 — check it before re-recording."

That file does not exist and never did. The test "does not point a reaped server recording at a host file that was never written" only covers a dispose with no stop in flight, so it passes.

hubgan (commented):
Reviewed the diff plus the code it transitively affects, driving a tool-server built from this branch against a real iOS simulator and a real Android emulator.

Because no shipped simulator-server has `/api/recording/*` yet, I stood one in: a wrapper on `ARGENT_SIMULATOR_SERVER_DIR` that spawns the real binary, proxies everything to it (including the `/ws` upgrade) and serves the recording routes itself. So both paths ran for real — real HTTP, real devices, real files — rather than through mocks. I checked the fallback sniff against the shipped binary on both backends first: `POST /api/recording/start` answers 404 with `content-length: 0` on `simulator-server ios` and on `simulator-server android` alike, so the 404 detection is correct for everything that ships today.

What passed: the fallback path records and stops correctly on iOS and Android; the server path takes no ffmpeg at all on either; the admission guards (double start, concurrent stop, start-after-cap) behave identically on both paths; tvOS and Vega are rejected as intended; `npx vitest run` in `packages/tool-server` is 4534 passed / 1 failed, the failure being `lens-tools-platform-gate` timing out at 5s and passing at 30s, unrelated to this PR; prettier, eslint and both tsconfigs clean.

The findings are inline. Three are broken flows I reproduced end to end and would want fixed before this path goes live; the rest are smaller, and the documentation ones include a docs page the PR leaves contradicting its own tool description.

Not covered: a physical Android device (none available here; it goes through the same code and the same simulator-server router, only a different subcommand).

packages/tool-server/src/blueprints/screen-recording-session.ts
```diff
@@ -196,37 +221,58 @@ export const screenRecordingSessionBlueprint: ServiceBlueprint<
...
         state.disposed = true;
-        // Decided BEFORE the teardown below clears the flags it reads: a
-        // capture still encoding, a start mid-flight and one waiting to be
-        // handed over all owe the caller a video.
+        // Whether this dispose is destroying an unretrieved capture, decided
+        // BEFORE the teardown below clears the flags it is read from. Both
+        // states owe the caller a video: one is still encoding, the other
+        // finished and is waiting to be handed over.
         const hadUnretrievedCapture =
           state.recordingActive || state.startPending || state.pendingRetrieval;
```
hubgan: `stopPending` is missing from this set. `stopServerCapture` clears `recordingActive` synchronously (server-capture.ts:185) before awaiting `stop()`, so for the whole of that await — again up to 60s — `stopPending` is the only flag still true and no breadcrumb is recorded at all. On the server path the video exists only inside the simulator-server the sweep is killing.

Reproduced twice, deterministically: start on the server path, call `screen-recording-stop` against a simulator-server that does not answer, and run `stop-all-simulator-servers {devices:[udid]}` while it is in flight. The in-flight stop fails with a network error, and the retry answers:

"No active screen recording on device <udid>. Call `screen-recording-start` first."

That is the "you never started one" answer reaped-sessions.ts exists to prevent, for a recording that did run and whose only copy the teardown just destroyed.

packages/tool-server/src/utils/simulator-client.ts
```diff
@@ -140,6 +140,213 @@ async function pointerPost(
...
+        error_kind: "not_found",
+      }
+    );
+  }
+  return {
+    path: body.path,
+    sizeBytes: body.size_bytes ?? 0,
+    durationMs: body.duration_ms,
+    wallClockMs: body.wall_clock_ms ?? body.duration_ms,
+    trimmedMs: body.trimmed_ms ?? null,
```
hubgan, line 444: `0 ?? null` is `0`, so a server reporting `trimmed_ms: 0` when trimming collapsed nothing passes `stopServerCapture`'s `result.trimmedMs !== null` check and the tool returns the trim fields for a recording where no trimming applied. Reproduced end to end: a stop reply carrying `trimmed_ms: 0` came back as `{durationMs: 3029, wallClockMs: 3029, trimmedMs: 0}`. That contradicts the comment directly above the spread in server-capture.ts, the stop tool's own contract ("Present only when trimming actually removed frames") and SKILL.md. The host path guards exactly this with `trimmedAnyFrames`.

Line 263 has the matching problem: `wall_clock_ms ?? duration_ms` means a reply carrying `trimmed_ms` but no `wall_clock_ms` reports `wallClockMs === durationMs` alongside a non-zero `trimmedMs`, which cannot both be true.

packages/tool-server/src/utils/simulator-client.ts
```diff
@@ -140,6 +140,213 @@ async function pointerPost(
...
+  if (body === null) {
+    // `serverStop` is only ever stamped by a start that the route answered, so
+    // the same server answering 404 now means it stopped serving that route
+    // mid-recording. Never observed; handled rather than falling back, because
+    // there is no video to hand over either way.
+    throw new FailureError(
+      `screen-recording-stop failed: simulator-server no longer exposes a recording endpoint, ` +
+        `so the recording in progress cannot be finalized.`,
+      {
+        error_code: FAILURE_CODES.SCREEN_RECORDING_OUTPUT_MISSING,
```
hubgan, line 418: `SCREEN_RECORDING_OUTPUT_MISSING` with `error_kind: "not_found"` means "the video file is absent or empty" everywhere else in this repo (session-guards.ts:166,176; server-capture.ts:226). This throw is about a simulator-server build that does not expose the route at all, which is a capability problem, not a missing file — and 130 lines further down this same file the equivalent case is classified as `unsupported` (`PASTE_CLIPBOARD_UNSUPPORTED`). So an agent sees `SCREEN_RECORDING_OUTPUT_MISSING` for a build limitation, and telemetry cannot separate it from genuine missing-video failures.

It also matters beyond the label: the kind is the only thing that could let `stopServerCapture` tell "nothing to retry" from "may still be recording", which is the classification behind the wedge flagged on server-capture.ts. The sibling throw at line 252 additionally omits the `failure_command: "simulator_server"` this one carries.

packages/tool-server/src/utils/simulator-client.ts
```diff
@@ -140,6 +140,213 @@ async function pointerPost(
...
+          error_code: FAILURE_CODES.SIMULATOR_NON_JSON_RESPONSE,
+          failure_stage: "simulator_server_parse_response",
+          failure_area: "tool_server",
+          error_kind: "network",
+          network_failure: "invalid_response",
+        }
+      );
+    }
+  }
+  if (!res.ok || !body || body.error) {
```
hubgan, line 552: An HTTP 200 with a completely empty body skips the JSON-parse branch above (guarded by `raw.trim() !== ""`), lands on `!body`, and is reported as "simulator-server rejected the recording command (HTTP 200)". The server rejected nothing, and the message drops the actionable "restart the simulator-server" hint the non-JSON branch carries. At stop it also feeds the `finalizeRequestFailed` path. The PR's own test "says the reply was unreadable rather than blaming the server for rejecting" argues against this shape for a malformed body; an empty one still gets it.

packages/tool-server/src/tools/screen-recording/screen-recording-start.ts
```diff
@@ -19,6 +26,20 @@ const MAX_TIME_LIMIT_SECONDS = 600;
...
+ * finalizes the mp4 container before it answers. Without a signal they fall to
+ * undici's 300s header timeout, which would hold `startPending`/`stopPending`
+ * — and so every other recording call on the device — for five minutes per
+ * command against a sim-server that accepts the connection and never replies.
+ *
+ * Starting is a bare "begin encoding" command, so it gets the smaller budget;
+ * finalizing scales with the video, which frames-already-encoded muxing still
+ * keeps far inside a minute.
+ */
+const RECORDING_START_TIMEOUT_MS = 10_000;
```
hubgan, line 41: Adding this pre-flight POST makes the host fallback conditional on the simulator-server HTTP API answering, which it never used to be. Frames come from `streamUrl`, and the only other `/api` call on the start path is the pointer toggle, which `pointerPost` deliberately swallows into `false` ("so a recording is never lost to a pointer toggle"). A `/api/recording/start` that hangs now costs the whole recording.

Measured on a real simulator with ffmpeg installed and the frame stream healthy: the start failed after 10s with `SIMULATOR_NETWORK_TIMEOUT` and no ffmpeg was spawned. The session does recover (the next start succeeds), so this loses one recording rather than wedging, but it is a fallback that used to exist and no longer does.

packages/tool-server/src/utils/reaped-sessions.ts
```diff
@@ -62,11 +77,18 @@ export function takeReapedSession(
...
- * `Registry._teardown` calls a blueprint's `dispose()` with no caller, so the
- * disposer that left the breadcrumb cannot see which teardown triggered it —
- * hence a message naming the family rather than asserting one member.
+ * The disposer that leaves a breadcrumb cannot see who triggered it — a
+ * blueprint's `dispose()` is called by `Registry._teardown`, with no caller — so
+ * the message names the family rather than asserting one member.
+ * `stop-all-simulator-servers` is the common one and is named first, but it is
+ * not the only one: `stop-simulator-server` on Chromium cascades into the
+ * debugger through `ChromiumCdp` (its documented behaviour), and
+ * `react-profiler-start { force: true }` disposes the debugger and the profiler
```
hubgan: `params.force` in react-profiler-start.ts is read at exactly one place, line 281 (`canTakeOverSilently`), and its only effect is the in-page `STOP_FOR_TAKEOVER_SCRIPT` evaluate at line 293, which requires the debugger alive. The two `disposeAndWait()` call sites, lines 158 and 163, run before that check and are gated on an unhealthy node or a dead CDP connection, never on `force`. So this cause cannot produce the breadcrumb the comment attributes to it. The same attribution is repeated in the agent-visible string at line 99 (pre-existing there).

Two other claims in the rewritten header: "disposes every device-owned service" holds only for the unscoped sweep — a scoped call disposes only the named devices' services; and "the three that hold captured output" misses `ChromiumJsRuntimeDebugger`, which is device-owned, holds a console-log file and records the same `js-runtime-debugger` kind.

packages/tool-server/src/utils/simulator-client.ts
```diff
@@ -140,6 +140,213 @@ async function pointerPost(
...
+}
+
+/**
+ * POST a recording command, returning null when the route does not exist.
+ *
+ * Deliberately not `simulatorPost`: a build without the recording feature has
+ * no route at all, so its answer comes from the router's unmatched-route
+ * fallback — 404 with an empty body, which that helper would surface as
+ * "non-JSON response", indistinguishable from a server in a bad state and the
+ * one answer callers must be able to act on. Every failure a recording handler
```
hubgan: "Every failure a recording handler reports for itself comes back as HTTP 200 carrying an `error` field" is stated as established fact about a handler that exists in no simulator-server build in this repo — `strings` over the bundled binaries finds no `api/recording` — so it is an assertion about unmerged radon#155. It is also in tension with the branch it justifies at line 334, which exists precisely to handle a non-2xx carrying a body.

Worth noting alongside it: the natural axum spelling of an unknown recording id, `StatusCode::NOT_FOUND` with no body, is indistinguishable from "no route" under this sniff. At stop that turns into the `finalizeRequestFailed` wedge flagged on server-capture.ts.

packages/tool-server/src/blueprints/screen-recording-session.ts
```diff
@@ -100,14 +114,23 @@ export interface ScreenRecordingSessionApi {
...
 const DISPOSE_FINALIZE_GRACE_MS = 1_500;
 const DISPOSE_REAP_MS = 1_000;
+/**
+ * Same policy for a server-side recording: the salvage stop dispose issues to
+ * end a recording running inside simulator-server gets this much time, not the
+ * stop tool's own (much longer) finalize contract — the video here is being
+ * abandoned, so a wedged or slow-to-mux simulator-server must not stall a
+ * teardown sweep (or process shutdown) to obtain one.
+ */
+const DISPOSE_SERVER_STOP_MS = 1_500;
```
hubgan: The constant bounds how long dispose *waits*, not how long the stop gets. The request still carries `AbortSignal.timeout(RECORDING_STOP_TIMEOUT_MS)` = 60s, created inside the closure in screen-recording-start.ts, and keeps running after the race resolves — dispose simply stops waiting on it. The "must not stall a teardown sweep" rationale is right; "gets this much time" is not, and it is the mechanism behind the salvage message below asserting the recording ended when the timeout branch cannot know that.

packages/skills/skills/argent-screen-recording/SKILL.md
```diff
@@ -27,7 +27,7 @@ A recording does not stop itself before its `timeLimitSeconds` cap, so a forgott
...
 3. Set the end-of-recording reminder described in §2 — this step is not optional.
 4. Drive the interaction to capture: gestures, navigation, typing (`argent-device-interact`). Prefer `run-sequence` for tight multi-step interactions so tool-call latency does not pad the video.
-5. Call `screen-recording-stop` with the same `udid`. It returns `{ video, durationMs, wallClockMs?, trimmedMs?, warning? }`; `video` is an artifact whose resolved path points at the durably-saved file (see below). The video is already final when stop returns (the watermark is stamped during capture, not in a second pass), so stop takes well under a second.
+5. Call `screen-recording-stop` with the same `udid`. It returns `{ video, durationMs, wallClockMs?, trimmedMs?, warning? }`; `video` is an artifact whose resolved path points at the durably-saved file (see below). The video is final when stop returns. On builds where simulator-server does the recording, stop also asks it to finalize the mp4 and copies the file out, so stop can take longer for long recordings; the host-ffmpeg fallback finalizes during capture and stops quickly.
```
hubgan: "the host-ffmpeg fallback finalizes during capture" is not what happens. ffmpeg *encodes* during capture; the container trailer is written when `stopCapture` closes ffmpeg's stdin (`finalizeCapture`), and stop then waits on the child for up to `FINALIZE_WAIT_MS` (20s) before SIGINT and a further 5s before SIGKILL. The pre-PR sentence was accurate about what does happen during capture ("the watermark is stamped during capture, not in a second pass").

Two things this skill now needs and does not say: that a stop failing with the finalize-request error is meant to be retried — the recovery designed into server-capture.ts is unreachable if the agent does not know to retry — and that a failed copy-out leaves the only finished video inside simulator-server, retrievable only until that server exits.

packages/skills/skills/argent-screen-recording/SKILL.md
```diff
@@ -44,5 +44,5 @@ A recording does not stop itself before its `timeLimitSeconds` cap, so a forgott
...
 - **Android**: records at the device's native resolution; secure screens (DRM, some password fields) come out black.
 - **Unsupported**: tvOS simulators, physical iPhones, Chromium apps, Vega/Fire TV, and remote (`remote:`-prefixed) simulators — none of them expose a readable frame stream. For a single still frame use `screenshot`; for a replayable interaction script use `argent-create-flow` instead of a video.
-- **ffmpeg is required**: it is the encoder, so `screen-recording-start` fails up front with an install hint if it is missing (`brew install ffmpeg`). It is resolved from `PATH` plus the usual Homebrew prefixes.
+- **ffmpeg is needed unless simulator-server encodes for itself**: builds that expose a recording endpoint record and mux the video inside the server, and nothing extra is installed. Every other build — and every simulator-server predating the endpoint — is encoded host-side instead, and there `screen-recording-start` fails up front with an install hint if `ffmpeg` is missing, resolving it from `PATH` plus the usual Homebrew prefixes. Which one you get is not something to predict from the platform: install `ffmpeg` when a start actually asks for it.
```
hubgan: This bullet was corrected here, but `packages/docs/docs/features/screen-recording.mdx:48` still reads "Argent uses `ffmpeg` to encode the video. Install `ffmpeg` before the first recording." — which now contradicts the shipped tool description ("install it when that is what comes back rather than up front"). It is the only ffmpeg mention in the docs tree, and the repo CLAUDE.md requires the docs page to move with a user-facing capability change; the PR description does not state that no docs update was needed.

`packages/registry/src/failure-codes.ts:216` is stale for the same reason: "iOS and Android share one capture path (simulator-server frames into ffmpeg)". There are two capture paths now and the preferred one has no ffmpeg.

packages/tool-server/src/tools/screen-recording/capture.ts
```diff
@@ -618,7 +668,8 @@ export async function stopCapture(api: ScreenRecordingSessionApi): Promise<StopR
   } finally {
     // Always return the session to a startable state — a failed stat must not
     // wedge the next start behind "already active". The video is host-side, so
-    // there is nothing a retried stop could recover.
+    // unlike a device-side capture there is nothing a retried stop could
```
hubgan: "unlike a device-side capture" names a path that does not exist in this codebase. The contrasting path this PR actually introduces does the opposite of what the sentence implies: server-capture.ts deliberately restores `serverStop` and sets `pendingRetrieval` so that a retried stop *can* recover.

packages/tool-server/src/tools/screen-recording/screen-recording-start.ts
```diff
@@ -184,3 +206,24 @@ export function makePointerControl(simulator: SimulatorServerApi): PointerContro
...
+ */
+export function makeServerRecordingControl(simulator: SimulatorServerApi): ServerRecordingControl {
+  return {
+    async start(opts) {
+      const id = await startServerRecording(
+        simulator,
+        opts,
+        AbortSignal.timeout(RECORDING_START_TIMEOUT_MS)
+      );
+      if (id === null) return null;
```
hubgan, line 224: Nothing joins this to the tool: replacing `if (id === null) return null` with a throw leaves all 107 tests green. The 404-to-host-fallback is the path every shipped build takes today, and it is only ever covered through a hand-rolled fake whose `start` already returns null (`fakeServer({supported: false})`), so `startServerRecording`'s 404 handling and this null-forwarding are never exercised together.

Other branches that survive mutation with the suite green: `stopServerRecording`'s 404 (returning a fabricated result instead of throwing), the `!body` operand at simulator-client.ts:334 (which is what stops a second concurrent capture), the `duration_ms` half of the missing-path guard, the cap-timer ownership guard at server-capture.ts:137, both `toSimulatorNetworkError` wrappings in `recordingPost`, and the two timeout constants (setting both to 1 keeps the suite green — the test only asserts `signal instanceof AbortSignal`).

packages/tool-server/src/tools/screen-recording/server-capture.ts
```diff
@@ -0,0 +1,313 @@
...
+  const outputFile = api.outputFile!;
+  const stop = api.serverStop!;
+  const endedEarly = api.recordingTimedOut;
+  const pointerFailed = api.pointerFailed;
+  // Hand ownership over before the await: dispose decides whether to stop the
+  // server's recording by looking at `serverStop`, and leaving it set across
+  // this suspension lets a shutdown issue a second, concurrent stop for the
+  // recording this call is already finalizing. That second stop also kills the
+  // simulator-server whose session directory the copy below reads from.
+  api.serverStop = null;
```
hubgan, line 186: Neither of the two windows this handover opens is covered. No test disposes the session while `stop()` is in flight, on either side of the cap — which is why the two breadcrumb defects flagged on screen-recording-session.ts pass the suite. The dispose grace itself is also only pinned loosely: `DISPOSE_SERVER_STOP_MS = 4_900` still passes (the test advances 5000ms), so the assertion is really "under 5s" rather than 1.5s, and no test has the salvage `stop()` reject inside the race.

packages/tool-server/src/tools/screen-recording/server-capture.ts
```diff
@@ -0,0 +1,313 @@
...
+  // handed over by `screen-recording-stop`, whenever that arrives.
+  api.recordingTimeout = setTimeout(() => {
+    api.recordingTimeout = null;
+    // Ownership guard: a newer capture may have stamped the session already.
+    if (api.serverStop !== stop) return;
+    api.recordingTimedOut = true;
+    api.recordingActive = false;
+    api.wallClockEndMs = Date.now();
+    api.pendingRetrieval = true;
+    markScreenRecordingFinalized(api.deviceId, `it hit its ${params.timeLimitSeconds}s time limit`);
```
hubgan, line 152: Nothing on this path ever notices that simulator-server — and with it the recording and the video — is gone. `recordingExitedUnexpectedly` has no writer here at all (it is written only by the host ffmpeg exit handler), so the session keeps reporting a healthy recording, and then this timer upgrades the claim.

Reproduced on a real simulator, killing simulator-server 3s into a 20s recording so the video is destroyed with it:

- t=5s: "NOTE: A screen recording is still running on device <udid> (started 8s ago, auto-stops after 20s)…"
- t=23s: "NOTE: The screen recording on device <udid> already ended (it hit its 20s time limit) but its video has not been retrieved yet — call `screen-recording-stop` with { "udid": … } to get the file."

The second note is an affirmative claim that a retrievable video exists, made about a file destroyed twenty seconds earlier, and it is attached to every tool result for every agent on the tool-server. The stop it instructs then fails, and the session wedges.

The host path detects the same event: the MJPEG socket errors, `stopPump` stashes it, and stop returns the file plus "The frame stream from simulator-server dropped during the recording…; the video may freeze on its last received frame." Nothing equivalent exists here.

packages/tool-server/src/utils/simulator-client.ts
```diff
@@ -140,6 +140,213 @@ async function pointerPost(
...
+  if (body.status !== "ok" || typeof body.id !== "string") {
+    // Neither the success reply nor a recognized rejection, so whether a
+    // recording is now running is unknown — and a success without an id is no
+    // better, since nothing could ever stop it. Fail instead of falling back: a
+    // second capture over one that did start would record the screen twice and
+    // strand the server's copy with nothing left to stop it.
+    throw new FailureError(
+      `screen-recording-start failed: simulator-server answered the recording command with ` +
+        `${JSON.stringify(body).slice(0, 200)} instead of a status. A recording an earlier ` +
+        `start attempt began may still be running inside simulator-server — for example one ` +
```
hubgan, line 375: This branch knows the right thing to say when a start's reply cannot be interpreted — that a recording may now be running inside simulator-server and that retrying after its own time limit should succeed. The far more likely way to lose that reply, the `AbortSignal.timeout(RECORDING_START_TIMEOUT_MS)` firing or the connection dropping, goes through `recordingPost`'s `toSimulatorNetworkError` instead and carries none of it.

Reproduced with a simulator-server that accepts the start, begins recording, and answers after the 10s budget:

1. `screen-recording-start {udid, timeLimitSeconds: 600}` → "screen-recording-start timed out — simulator-server at … did not respond in time. The simulator may be unresponsive." The recording is in fact running server-side; the host session is left completely clean, with no reminder entry.
2. Retried start → "simulator-server rejected the recording command (HTTP 200: a recording is already running)" — a bare rejection with no guidance.
3. `screen-recording-stop` → no active recording.

The device is then unrecordable for up to the server's own `time_limit_secs` (600s here) and nothing tells the caller why, or that waiting it out is the fix — even though the branch three lines above knows to say exactly that.

packages/tool-server/src/tools/screen-recording/screen-recording-start.ts
```diff
@@ -77,7 +98,7 @@ By default every tap, swipe, drag, pinch and rotate is drawn into the video as a
 The recording keeps running across other tool calls (every result carries a reminder) until \`screen-recording-stop\` is called or timeLimitSeconds elapses — immediately after starting, set yourself a reminder/wakeup for the expected end of the recording so it is never left running.
 Use when the user wants a video of an interaction, animation, or app behavior — for a single still frame use \`screenshot\` instead.
 Returns { status: "recording", timeLimitSeconds, outputFile } — the video is retrieved later by \`screen-recording-stop\`, not by reading outputFile directly.
-Fails if a recording is already running on the device, the device is not booted, ffmpeg is not installed, or the platform cannot be recorded (tvOS, Chromium, Vega and remote simulators are unsupported).`,
+Fails if a recording is already running on the device, the device is not booted, or the platform cannot be recorded (tvOS, Chromium, Vega and remote simulators are unsupported). Where simulator-server cannot record for itself the video is encoded locally instead, and \`ffmpeg\` must be installed — the failure names it, so install it when that is what comes back rather than up front.`,
```
hubgan: I'm not sure if we should instruct agent to install ffmpeg. I think they should ask the user

packages/tool-server/src/tools/screen-recording/server-capture.ts
```diff
@@ -0,0 +1,313 @@
...
+          error_code: FAILURE_CODES.SCREEN_RECORDING_OUTPUT_MISSING,
+          failure_stage: "screen_recording_stop",
+          failure_area: "tool_server",
+          error_kind: "not_found",
+          failure_command: "simulator_server",
+        },
+        { cause: err instanceof Error ? err : new Error(String(err)) }
+      );
+    }
+    await fs.rm(result.path, { force: true }).catch(() => {});
```
hubgan, line 251: I think this can be replaced with `fs.rename` which will do the same work, but faster

packages/tool-server/src/tools/screen-recording/server-capture.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Confirmed and fixed in 5357ee64d. `stopServerCapture` classified every `stop()` rejection as a failed *request* and kept the recovery handles, so a connection-refused (server gone), an HTTP-200 `{error}`, an empty 404 or a no-path reply all parked the session at `pendingRetrieval` forever - the wedge you reproduced. `finalizeRequestFailed` now keys on `getFailureSignal(err)?.network_failure === "timeout"`: only a timeout leaves the recording's fate unresolved and a retry worth trying. Every other rejection resets the session to startable (serverStop cleared, reminder cleared, pointer disabled), so a stop against a gone server returns the connection error and the next start is admitted. Pinned by "resets the session to startable when the stop fails against a gone server" (a connection-refused stop → fully startable), red before the fix.

packages/tool-server/src/blueprints/screen-recording-session.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Confirmed and fixed in 79a32d9fb. Because stop hands `serverStop` over before its await, a dispose mid-stop read it null and took the ffmpeg salvage branch, naming a host file the server path never wrote. A stable `serverCapture` flag - set at the server start and surviving the hand-over - now decides which salvage story dispose tells. Pinned by "records the server salvage breadcrumb when disposed mid-stop" (asserts the message says "nothing was written to" and NOT "ffmpeg"), red before the fix.

packages/tool-server/src/blueprints/screen-recording-session.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Confirmed and fixed in 79a32d9fb. `hadUnretrievedCapture` now includes `stopPending`, so a dispose during an in-flight stop - where stop has already cleared `recordingActive` and no cap set `pendingRetrieval` - still records the breadcrumb instead of leaving the retry to hit "no active recording". Same test as above: without the fix it answered "No active screen recording ... Call screen-recording-start first" for a recording whose only copy the teardown had just destroyed.

packages/tool-server/src/utils/simulator-client.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Confirmed and fixed in 5357ee64d. Gated the trim fields on `result.trimmedMs !== null && result.trimmedMs > 0`, matching the host path's `trimmedAnyFrames`, so `trimmed_ms: 0` (trimming on, nothing collapsed) no longer surfaces them. Pinned by "omits the trim fields when trimming ran but collapsed nothing". On the `wall_clock_ms ?? duration_ms` half: it only reaches the output when `trimmedMs > 0`, and a real server that sends `trimmed_ms > 0` always sends `wall_clock_ms` with it, so the contradiction needs a malformed reply no build produces - I left the default as the harmless fallback for the now-unreported path rather than guard an impossible shape.

packages/tool-server/src/utils/simulator-client.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in 79e0cdc71. The 404-at-stop "route vanished" throw is now `error_kind: "unsupported"` (a capability problem, like the clipboard's missing-endpoint case), not `not_found` - which does mean "the video file is absent" everywhere else - and its sibling no-path throw gains the `failure_command: "simulator_server"` the 404 one already carried. I kept the `SCREEN_RECORDING_OUTPUT_MISSING` code rather than mint a new one: the branch is provably unreachable (`serverStop` is only stamped by a start the route answered) and the caller-facing outcome is still "no video". Worth noting the kind no longer feeds the wedge either - after 5357ee64d, stopServerCapture keys recovery on `network_failure`, not `error_kind`. Pinned by "fails a stop against a route that vanished, classified as unsupported".

packages/tool-server/src/utils/simulator-client.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Confirmed and fixed in 79e0cdc71. An empty-bodied 2xx now throws the "empty response ... Restart the simulator-server" error (the `SIMULATOR_NON_JSON_RESPONSE` shape), since a present route always answers 2xx with a JSON body. An empty non-2xx still falls through to the "HTTP <status>" rejection, so an empty 500 keeps its meaning and stays distinct from the empty-404 missing route. Pinned by "reports an empty 2xx body as unreadable, not as a rejection" and "still rejects an empty non-2xx body as an HTTP failure".

packages/tool-server/src/tools/screen-recording/screen-recording-start.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: You're right that the server-recording probe is now on every start's critical path and a hung `/api/recording/start` costs the recording. I addressed the actionable half in 7a4a95fea: a start whose reply is lost (timeout / generic network error) now carries the "a recording ... may still be running inside simulator-server ... retry after its time limit" guidance, so the caller knows why the device is unrecordable and how to recover.

I deliberately did NOT restore a host fallback on that timeout. If the probe timed out, the server may already have begun recording, and falling back to the host encoder there would record the screen twice and strand the server's copy - exactly the double-capture `startServerRecording` fails-instead-of-falls-back to avoid (the fallback is chosen only by an empty 404, never by an error or timeout). Failing with actionable guidance is the safe resolution; a host recording would also be blocked on the *next* start anyway, since its own probe would get the server's "already running".

packages/tool-server/src/utils/reaped-sessions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: All three confirmed and fixed in 261e836bd.

1. The `force` attribution was wrong: `react-profiler-start`'s two `disposeAndWait` sites are gated on an unhealthy node / dead CDP connection, and `params.force` only drives the in-app `STOP_FOR_TAKEOVER_SCRIPT` (react-profiler-start.ts:293), which disposes no service. The prose now says a takeover disposes a *stale* debugger, and that `force` itself disposes nothing.
2. "disposes every device-owned service" is now qualified - an unscoped sweep reaps every device's services, a scoped call only the named devices'.
3. The enumeration now names the `JsRuntimeDebugger`/`ChromiumJsRuntimeDebugger` pair, since `ChromiumJsRuntimeDebugger` is a fourth device-owned service holding its own console-log file under the same `js-runtime-debugger` kind.

packages/tool-server/src/utils/simulator-client.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in 79e0cdc71. Reworded so the comment describes the wire contract radon#155 upholds - a present route answers 2xx with a JSON body (success or in-band `{error}`) - rather than asserting it as observed fact about an unmerged build. It now also records the ambiguity you name: a handler answering an unknown id with a bare 404 reads as "no route", which is harmless here (no video to hand over either way, and the 404→fallback only runs at start, where a route-present server would have answered 2xx).

packages/tool-server/src/blueprints/screen-recording-session.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in 261e836bd. The comment now says the constant bounds how long dispose *waits* on the salvage stop, not how long the stop gets - the request keeps its own `RECORDING_STOP_TIMEOUT_MS` abort and may outlive the race; dispose simply stops waiting.

packages/skills/skills/argent-screen-recording/SKILL.md
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in 261e836bd. §3 step 5 no longer says the host fallback "finalizes during capture": it encodes every frame during capture and only closes/flushes the container at stop, returning in well under a second (waiting on ffmpeg up to ~20s before forcing it). Added a step 7 on reading a failed stop - retry a timeout / hand-over failure (the reminder stays, pointing you back), retry a copy-out failure promptly (the video is inside simulator-server only until it exits), and start fresh on any other failure (the device has reset).

packages/skills/skills/argent-screen-recording/SKILL.md
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in 261e836bd. `packages/docs/docs/features/screen-recording.mdx` now says ffmpeg is needed only on the host-encoding build and the failure names it (matching the tool description), not "install before the first recording". `failure-codes.ts` no longer describes the single "simulator-server frames into ffmpeg" path - it names both. `npx docusaurus build` and `npm run format` are clean.

packages/tool-server/src/tools/screen-recording/capture.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in 261e836bd. Dropped the non-existent "device-side capture" contrast; the comment now states the host stop already finalized its file so a retry recovers nothing, and points at the server path as the real contrast - it *does* keep a recording recoverable when its finalize request times out.

packages/tool-server/src/tools/screen-recording/screen-recording-start.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in 0edb1b38f. Added discriminating tests for the branches you listed: the `makeServerRecordingControl` null-forwarding every shipped build relies on ("makeServerRecordingControl.start returns null so a 404 build falls back host-side" - the fallback test drove a fake whose start already returned null, so this bridge was never exercised), both halves of the stop path/duration guard, `stopServerRecording`'s 404, and the empty-2xx-vs-500 split (which also pins the `!body` operand and both `toSimulatorNetworkError` wrappings).

Two I left, with reasons: the timeout constants can't be pinned to exact values because `AbortSignal.timeout` isn't driven by the suite's fake timers (verified) and exporting the consts would trip the zero knip ceiling - the existing test pins that both commands ARE bounded, which is the wedge-prevention property. And the cap-timer ownership guard is belt-and-suspenders (the timer is always cleared before a new capture stamps), so I left it rather than stage a contrived stale-timer.

packages/tool-server/src/tools/screen-recording/server-capture.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Confirmed and fixed in 79a32d9fb. Added "records the server salvage breadcrumb when disposed mid-stop", which disposes while `stop()` is in flight and covers both handover windows (it is what pins the two breadcrumb fixes above). Tightened "bounds dispose's salvage stop at the short grace" to pin the 1.5s bound - still waiting at 1.4s, resolved by 1.6s (it passed at 4.9s before) - and added "does not hang the teardown when the salvage stop itself rejects".

packages/tool-server/src/tools/screen-recording/server-capture.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Confirmed and fixed in c2a5949e6. The finalized note now branches on which path recorded it: the server-side variant says the video is inside simulator-server and not yet retrieved and points at copying it out "while that server is still up", so it never claims a host file that does not exist. Combined with 5357ee64d, following that guidance against a gone server resets the session to startable rather than wedging. Pinned by "does not claim a server-side finalized video exists when the server may be gone".

packages/tool-server/src/utils/simulator-client.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Confirmed and fixed in 7a4a95fea. `recordingPost` now appends "A recording an earlier start attempt began may still be running inside simulator-server ... retry after its time limit" for a start; `toSimulatorNetworkError` adds that hint only to the timeout and generic-network messages, never to connection-refused/reset (where nothing began). Pinned by "warns a timed-out start may be recording server-side" and its refused counterpart.

packages/tool-server/src/tools/screen-recording/screen-recording-start.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Good call - changed in 261e836bd. The tool description and SKILL no longer command the agent to install ffmpeg up front; they name it as a requirement the failed start surfaces, leaving the agent to install it or ask the user only when a recording actually asks for it. The user-facing docs page still tells the reader to install it.

packages/tool-server/src/tools/screen-recording/server-capture.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: I looked at this and kept the copy. `fs.rename` fails with EXDEV across filesystems and does not fall back to a copy the way `mv` does - simulator-server's session directory is not guaranteed to share a mount with argent's temp dir (a remote `argent link` server makes them different hosts entirely). The copy-then-drop is also deliberately fail-safe: a failed removal of the source leaves a harmless duplicate, whereas a failed rename would lose the only copy. Videos are small enough that the copy is not a real cost. Leaving it as a copy.

packages/tool-server/src/utils/simulator-client.ts
```diff
@@ -317,6 +317,254 @@ async function pointerPost(
...
+        error_kind: "not_found",
+        failure_command: "simulator_server",
+      }
+    );
+  }
+  return {
+    path: body.path,
+    sizeBytes: body.size_bytes ?? 0,
+    durationMs: body.duration_ms,
+    wallClockMs: body.wall_clock_ms ?? body.duration_ms,
```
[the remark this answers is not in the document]
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Confirmed and fixed in 93c38c2a3. You are right that `wall_clock_ms ?? duration_ms` fabricates wallClockMs === durationMs, and paired with trimmed_ms > 0 that is a contradiction. The map now preserves the absence (`typeof body.wall_clock_ms === "number" ? body.wall_clock_ms : null`), wallClockMs is `number | null`, and stopServerCapture reports the trim pair only when `result.wallClockMs !== null` as well - so a reply that omits the wall clock drops the trim fields rather than emitting a faked one. Pinned by two discriminating tests: "maps an absent wall_clock_ms to null rather than the video length" (wire) and "omits the trim fields when the server reports a trim but no wall clock" (end to end); both fail on the old `?? duration_ms`.

packages/docs/docs/features/screen-recording.mdx
```diff
@@ -45,7 +45,7 @@ The files stay in the working tree. Add the directory to `.gitignore` when you d
 
 - Argent records iOS simulators, Android emulators, and physical Android devices. Argent does not record tvOS simulators, physical iPhones, Chromium apps, Vega devices, or remote simulators.
 - One recording per device at a time. Different devices can record at the same time.
-- Argent uses `ffmpeg` to encode the video. Install `ffmpeg` before the first recording.
+- Where simulator-server can record for itself, no extra tool is needed. On every other build Argent encodes the video with `ffmpeg` instead, and the failed recording names it — install `ffmpeg` when a recording asks for it rather than up front. Which one applies is a property of the build, not the platform.
```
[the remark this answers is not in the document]
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in cdf1d495b. Split the bullet into short imperative sentences under the 25-word limit, keeping the build-dependent meaning: "Some simulator-server builds record video without an extra tool. Every other build encodes with `ffmpeg` instead, and a failed recording names it. Install `ffmpeg` only when a recording asks for it, not up front. The build, not the platform, decides which applies." `npx docusaurus build` and `npm run format` are clean.

PR #589, opened 2026-07-30
feat: support screen-sourced flow selectors

makhataibar:
## What changed

- add an explicit `source: app | screen` field to flow selector maps
- keep `app` as the default full-app hierarchy with no fallback
- resolve `screen` selectors against the screen-wide accessibility tree with no app-hierarchy fallback
- inherit a selector's source through relational scopes and reject conflicts
- preserve the source in YAML round-trips, diagnostics, reports, and snapshot crop identities
- route selector-driven actions and conditions through the selected hierarchy

## Why

On iOS, the full `UIView` hierarchy intentionally exposes app test IDs but cannot see system UI such as keyboard keys and the Share sheet. `describe` can see those elements through the screen accessibility hierarchy. Flows need an explicit, deterministic way to choose that tree without weakening the existing no-fallback behavior.

## Validation

- TDD red/green coverage for parsing, serialization, inheritance conflicts, screen-tree selection, semantic tapping, visible/hidden waits, reporting, and the existing app-tree failure contract
- `npm test -w @argent/tool-server` — 296 files, 3,077 passed, 1 skipped
- `npm run lint`
- `npm run build`

Live iOS recording/replay evidence will be added before this PR is marked ready.


makhataibar:
Closing this draft because the diblas Personal Test Guide l ive E2E path has moved to Maestro. Maestro’s screen-wide accessibility model completed the accepted proof and the atomic-listener replay, so this PR should not present Maestro results as live Argent selector evidence. The selector implementation branch and local worktrees are being preserved as forensic material; no source history is being discarded.

PR #590, opened 2026-07-30
feat(screenshot): save screenshots to the project's .argent/screenshots

latekvo:
## Human written:

Argent's CLI exposed `--out` flag on `screenshot` while the MCP command had no such option.
Thus all MCP-originated screenshots used to end up in tmp, while all videos used to end up in `.argent`.

This PR fixes that by adding `out` param to MCP.

---

Slop:

<details>

## The symptom

> Screenshots save to a temporary location that disappears. `screenshot` writes to a temp directory that goes away with the session. `screen-recording-stop` writes to the durable `.argent/recordings/`. There is a risk of losing screenshots if the session crashes between taking one and copying it somewhere.

(QA, Bluesky, physical Android phone.)

Confirmed: `screenshot` registered its PNG with `requireArtifacts(ctx).register(path, { mimeType: "image/png" })` and no `saveDir`, so the client materialized it under `tmpdir()/argent-artifacts/<project>/<session>/<device>/` — disposable scratch on a session-scoped path. It hits hardest exactly where a screenshot matters most: `includeImageInContext: false` exists to capture a full-resolution baseline for `screenshot-diff`, and that file is only useful if it is still there later.

**The durable mechanism already existed** — the `saveDir` hint on the artifact wire contract, added in #551 for recordings. Screenshots simply were not using it. So this PR is mostly about *who* gets it, not about building anything.

## What now happens

`screenshot` tags its PNG with `saveDir: ".argent/screenshots"` on all four capture paths (simulator-server, Chromium CDP, tvOS `simctl`, Vega), `.argent/screenshots` joins the client's `ALLOWED_SAVE_DIRS` allowlist, and the file lands on the **client** at `<project>/.argent/screenshots/` (or `~/.argent/screenshots/` outside a project) — including for a remote `argent link` server.

**One canonical filename**, `screenshot-<device>-<timestamp>.png`, mirroring `screen-recording-<device>-<timestamp>.mp4`. The backends name their temp files inconsistently and the durable directory accumulates across sessions, so the name has to say something:

| capture path | backend's name |
| --- | --- |
| simulator-server (iOS/Android) | `821081000-1785417279821.png` (verified live) |
| Chromium | `argent-screenshot-<deviceId>-<stem>.png` |
| tvOS | `argent-tv-screenshot-<udid8>-<hrtime>.png` |
| Vega | `vega-screenshot-<hrtime>.png` |

Renaming subsumes the `argent-` prefix strip that `screen-recording-stop` does.

## The volume problem — and where I drew the line

The MCP layer appends a screenshot after most tool calls (`mcp-server.ts` → `callTool("screenshot", { udid })`, gated by `AUTO_SCREENSHOT_TOOLS`). A naive tag dumps **hundreds of PNGs per session into the user's working tree** — worse than the bug being fixed. `screenshot` also has a second internal caller I found while auditing: `flow-visual.ts` invokes it for a `snapshot` step's comparison capture and reuses the returned handle as its `current` artifact.

**The distinction is made wherever it is actually visible**, which puts the two cases in different layers:

- **MCP auto-screenshot → client-side.** It arrives at the tool-server as *the same* `POST /tools/screenshot` an agent's own call does; the tool-server genuinely cannot tell them apart. Only the layer that synthesized the call knows, so `MaterializeContext.transient` declines persistence for that one round-trip. The tool-server's contract stays unconditional, and no internal-only parameter appears on a public tool schema an agent reads. The construction lives in `autoScreenshotContext()` in `auto-screenshot.ts` — the module whose stated job is "all tunables live here so they can be tested in isolation" — so the wiring is pinned by a test rather than being an inline literal.
- **Flow snapshot capture → tool-server-side.** Here flow-visual *does* know it is a sub-invocation, so it registers the compared image as its own `current` artifact and the hint never rides along. Fixing this client-side would have been action-at-a-distance: a third materialize site for flow artifacts would silently reintroduce the leak. A snapshot already persists what matters deliberately — an adopted baseline under `__baselines__/`, failure artifacts through the CLI's `--output` — and that export materializes these very handles and copies them, so a durable hint would have written the same PNG twice, in two places.

No retention/pruning of the project directory: nothing writes files it shouldn't, so there is nothing to prune. I don't think a bounded retention is worth adding on top — a screenshot only lands there when someone asked for one, which is the same footing as a recording.

## A regression this would have shipped: Lens previews

The Lens preview route's `previewImage` allowlist was `[os.tmpdir(), process.cwd(), "/tmp"]`. The Lens workflow's whole premise is passing a fresh screenshot's path straight to `propose_variant`, so moving screenshots out of temp breaks it whenever the durable dir is not under cwd:

- cwd is a **subdirectory** of the project (monorepo `apps/mobile`, project root above) → `<projectRoot>/.argent/screenshots/…` is under none of the three roots → 404 → **"No preview"**;
- client outside a project → `~/.argent/screenshots/…` → same.

The route now admits a durable screenshot by the shape of its own realpath — a file sitting immediately inside a directory named `.argent/screenshots` — rather than by a root derived from this process's cwd. Deriving it from cwd answers the wrong question a third way, which review caught: see below.

## Sibling artifact producers — audited

- **`screenshot-diff`** (`diffPath`, `contextDiffPath`) — left in temp deliberately. It already has an agent-controllable `outputDir` for durable output, and the context diff is rendered inline once, like an auto-screenshot. A second durable home would conflate the two mechanisms.
- **Flow snapshot `baseline` / `diff`** — left in temp; baselines are already durable under `__baselines__/` and failures export via `--output`.
- **Profiler `.trace` bundles** — cannot be tagged at all: `durableSaveTarget` excludes `archive` directory bundles by design. `native-profiler-analyze`'s report file is a different artifact class; `.argent/screenshots` is the wrong home and `.argent/profiles` is a separate change.

## Verification

Gates from the worktree root, all clean: `npx tsc --build`, `npx prettier --check .`, `npx eslint . --max-warnings 0`, and `typecheck:tests` in tool-server / tools-client / mcp (CI runs it separately). Suites: **tool-server 3082**, **tools-client 186**, **mcp 81**, **cli 276**, **registry 87** — all passing. No `package.json`/lockfile change, so no lockfile regen.

**Mutation checks** — every one of these reverts turns a test red, so nothing here is pinned by accident:

| reverted | result |
| --- | --- |
| `saveDir` tag in `screenshot` | 1 failed |
| `ctx.transient` honored in the materializer | 4 failed |
| `.argent/screenshots` from the client allowlist | 4 failed |
| `transient: true` in `autoScreenshotContext` | 1 failed |
| flow-visual re-registration | 2 failed |
| durable roots in `previewImageRoots` | 2 failed (unit) + 2 failed (HTTP route) |

One of these caught a **vacuous test of mine**: the first version of the `/variant-image` route test put its fixtures under `tmpdir()`, which is *itself* an allowed root, so it passed against the very regression it existed to catch. Fixtures now live beside the package (outside tmpdir, `/tmp`, and cwd), and the mutation goes red.

**E2E through the real client materializer** — worktree tool-server on `:3199`, live iOS 18.6 sim `39B9C547-…`, real captures driven over `POST /tools/screenshot`, then the real `materializeArtifacts` from the built `@argent/tools-client`:

| case | landed | temp cache |
| --- | --- | --- |
| co-located, in a project | `<proj>/.argent/screenshots/screenshot-39B9C547-…-1785418952951.png` (354×767 png, 238 935 B) | EMPTY |
| forced remote `GET /artifacts/:id` (gate missed) | `<proj>/.argent/screenshots/screenshot-39B9C547-…-1785418324439.png`, 237 623 B = announced size | EMPTY |
| outside a project (HOME redirected) | `<home>/.argent/screenshots/screenshot-39B9C547-…png` | EMPTY |
| auto-screenshot shape (`transient: true`) | `<cache>/case4-project-635689/20260730-133204-12116/39B9C547-…/screenshot-….png` | `.argent/` in project: **`[]`** |

`sips` reports every file as a real `png` and I opened the remote-downloaded one — a genuine capture of the sim's Settings screen. The inline image still renders in the transient case; declining persistence does not suppress the image block.

## Docs

#551 updated the recording SKILL when it moved recordings, so the equivalent prose here:

- `argent-lens/SKILL.md` claimed `previewImage` is "served from the OS temp dir / cwd" — falsified by this change (and load-bearing, per the regression above), now lists `.argent/screenshots/`.
- `argent-device-interact/SKILL.md` gains a "where the file lands" note, including that the automatic post-action screenshot is *not* saved there.
- `argent-screen-recording/SKILL.md` said recordings are durable "unlike most argent artifacts" — screenshots are durable now too, so it names visual-diff images as the scratch contrast instead.
- The `screenshot` tool description states where the PNG is saved.

## Interaction with #564 (`feat/artifact-kind`)

#564 changes `ArtifactStore.register` to a single `{ hostPath, kind, … }` options object and tags every producer, touching the same registration call sites in `screenshot/index.ts`, `flow-visual.ts` and `screen-recording-stop.ts`. Textual conflicts are small (my screenshot registrations are funnelled through one `registerScreenshot` helper, so it is one call site to reconcile, not four), and the two changes are orthogonal in substance: `kind` says *what an artifact is*, `saveDir` says *where it should live*.

Longer term a semantic `kind` is plausibly the more natural carrier for this: `kind: "screenshot"` → durable is a mapping the client could own outright, which would collapse `ALLOWED_SAVE_DIRS` into a kind→directory table and remove the server's say in the destination entirely. I deliberately did not build toward that here — it needs #564 to land first, and `transient` would still be required either way, since it distinguishes *requests*, not artifact kinds.

## Found in review

Each reproduced before it was changed, and each pinned by a test that fails without the fix.

- **A durable write that could not happen destroyed the screenshot.** Every failure inside the durable block returned `null`, and tagging `screenshot` routed all four capture paths through it. A `.argent/screenshots` occupied by a regular file, a read-only project root, a symlinked `.argent` (dotfile-managed, indistinguishable from the hostile case the guard was written for), or a full disk turned every explicit `screenshot` into `{"image": null}` — no image, no path, no error, and indistinguishable from a broken device. `argent run screenshot --out X` exited 1 having written nothing, blaming a missing `url`. Before this branch a screenshot always landed in the temp cache and always worked.

  The two failures are told apart now. A **destination** problem degrades to the disposable cache: nothing was written into the durable tree — which is exactly what the symlink guard is refusing — so the fallback gives up no safety and the caller still gets its image. An **untrustworthy artifact** (a size that cannot bound the download, a refused request, a truncated body) still returns `null`, because the cache download is deliberately uncapped for ordinary artifacts and retrying there would hand back the unbounded read the size check exists to prevent. Splitting these was not optional: collapsing them turned three existing security tests red.

- **A write that failed part-way left the bytes it managed at the canonical name.** Measured under a file-size limit: a 1024-byte corpse held `screenshot-SIM-1-dl.png` permanently and displaced the next healthy capture to `… (2).png`. The failed candidate is removed before the error propagates. Same class for `.argent/recordings`, where the files are far larger.

- **The Lens allowlist was derived from the *tool-server's* cwd, while the durable path is chosen by the *client's*.** The tool-server is a machine-global daemon keyed by install: `ensureToolsServer` reuses one across projects precisely because it "may be healthy and serving another project's session", and it inherits the cwd of whichever session spawned it. A client working in a second project saved into a directory the daemon had never heard of → 404 → "No preview". On `main` this could not happen, because the materialized path was always under `os.tmpdir()`, allowed regardless of cwd. It is the same failure class this PR set out to fix for monorepo subdirectories, one level further out.

- **`argent run screenshot --out X` wrote the PNG twice**, once where the caller asked and once into the project unasked — the same double-write the flow snapshot capture is withheld from the durable tag to avoid. Now transient, gated on the same condition as the `--out` write itself, so `screen-recording-stop`'s mp4 (the other durable producer, and not an image) still persists.

- **A `saveDir` with a trailing separator** missed the allowlist and degraded to scratch with nothing to show for it. No producer trips it today, but the allowlist's own comment invites future entries and the failure is a silent no-op.

- **`.gitignore` did not cover `.argent/`**, which driving argent at this repo fills. Unanchored, because the client resolves the project root to the nearest ancestor with a `package.json` — in this monorepo, the individual package rather than the repo root.

- **`argent-lens/SKILL.md` told the agent a `/tmp` path would not be served.** `/tmp` has been an allowed root throughout, and the same file's next paragraph now says so.

### Test gaps, each confirmed by mutation

- **Three of the four capture paths' `saveDir` tag was unpinned** — every existing case used udid `"ABC"`, which classifies as android and reaches simulator-server, so dropping the tag from the Chromium, tvOS or Vega path left the full suite green. All four now go red individually.
- **The auto-screenshot's `transient` opt-out was pinned as a helper, never as the wiring.** Reverting the call site to a plain inline context left `argent-mcp` 81/81 green. The render moved into `auto-screenshot.ts` — the module whose stated job is "all tunables live here so they can be tested in isolation" — so the context is no longer assemblable beside it, and the test drives the function `mcp-server.ts` actually calls.
- **The preview containment guard had no negative case with the right shape.** Added: a sibling `.argent/screenshots-evil`, a `not.argent/screenshots`, a subdirectory below the screenshots dir, `.argent/recordings`, and a symlink borrowing the durable directory's name.
- **One test asserted a device id the system cannot produce.** Chromium ids are `chromium-cdp-<port>` and already filename-safe; `"chromium:9222/page"` classifies as android. The shapes that genuinely need the sanitizer are `remote:<udid>` and adb wireless serials, which are what it exercises now.

## Left out / worth knowing

- A `tool: screenshot` step in a flow persists under MCP (an explicit request by the flow author) but not under `argent flow run`, which never materializes tool-step results — a pre-existing asymmetry in what the CLI renders, not something this PR introduces.
- Two captures of one device in the same millisecond would collide; the client's existing exclusive-write logic saves the second as `… (2).png` rather than overwriting.
- Four test files that replaced `@argent/configuration-core` wholesale with a one-export factory now spread `importOriginal()`, because `preview.ts` legitimately uses two more of its exports.
- One flake seen once in a full tool-server run: `artifacts.test.ts > streams a registered file with its content type`, a supertest-based test that is self-contained (own tmpdir, own ephemeral app) and independent of everything here. Three subsequent full runs were clean.

</details>


packages/argent-tools-client/src/artifacts.ts
```diff
@@ -299,7 +312,10 @@ export function durableSaveTarget(
   // that would reject the whole `materializeArtifacts` and lose every sibling
   // artifact, instead of this one handle degrading to the temp cache.
   if (typeof handle.saveDir !== "string" || !handle.saveDir || handle.archive) return null;
-  const rel = normalize(handle.saveDir);
+  // `normalize` keeps a trailing separator, so `.argent/screenshots/` would miss
+  // the allowlist and degrade to scratch with nothing to show for it. The two
+  // spellings name the same directory; strip it so they compare equal.
+  const rel = normalize(handle.saveDir).replace(new RegExp(`\\${sep}+$`), "");
```
latekvo, line 308: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The trailing-separator normalization this PR adds is pinned by no test.

`durableSaveTarget` strips a trailing separator so that `.argent/screenshots/` and `.argent/screenshots` compare equal against the allowlist, and the comment above it states the consequence of not doing so: the spelling with the separator misses the allowlist and degrades to scratch. That behaviour has no coverage. Deleting the `.replace(...)` call leaves both the tools-client and tool-server suites green, so the branch is not held by anything.

**Trigger.** Remove `.replace(new RegExp(`\\${sep}+$`), "")` from durableSaveTarget and run the tools-client and tool-server suites: both stay green. Every `saveDir` literal in either suite is already separator-free (`.argent/recordings`, `.argent/screenshots`, `/etc`, `../escape`, `.git`), so no case reaches the normalization.

**Checked against.** packages/argent-tools-client/src/artifacts.ts:315-318 and its comment, against `grep -rn 'saveDir' packages/argent-tools-client/test packages/tool-server/test` — no literal in either suite carries a trailing separator. The PR body's mutation table lists six reverts; this branch is not among them.

latekvo (commented):

The observation below was confirmed by reading the cited code paths directly, against the pre-PR file and the rest of the suite.

packages/argent-tools-client/src/artifacts.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Confirmed and fixed in fb05fe5b.

**Reproduction.** Applied your exact mutation — deleted `.replace(new RegExp(`\\${sep}+$`), "")` from `durableSaveTarget` — and ran the full tools-client suite: **14 files / 205 tests, all green** with the fix gone. Your grep result holds: every `saveDir` literal in both suites is separator-free (`.argent/recordings`, `.argent/screenshots`, `/etc`, `../escape`, `.git`), so no case ever reached the normalization.

**Fix.** Added a pinning case to the `durableSaveTarget` suite: each allowed directory (`.argent/screenshots`, `.argent/recordings`) spelled with one and two trailing separators must resolve to the target identical to the plain spelling. Verified by mutation: red without the `.replace(...)` (`1 failed | 59 passed`), green with it (60/60). Gates re-run clean after the fix: `tsc --build`, `prettier --check` (both files), root `eslint --max-warnings 0`, `typecheck:tests` in tools-client.

latekvo:
Bad approach. Will rewrite to fix the MCP-CLI disparity in a minimally invasive way.

latekvo:
Superseded by #1015, which adds an `out` path to the `screenshot` tool and keeps the temp directory as the default.

This PR made `.argent/screenshots` durable for every explicit capture, and most of its weight was spent keeping that default from filling the working tree: the client-side allowlist, the `transient` opt-out for the MCP auto-screenshot, the carve-out for flow-visual's internal capture, the Lens preview-route regression, the `.gitignore` entry. Naming the file is what the agent wanted, and #1015 does it with one parameter. The non-happy-path findings from the review here (a failed durable write must not cost the capture; a part-written file must not hold the canonical name) carry over as behaviour in #1015.

PR #595, opened 2026-07-30
fix(e2e): stop the harness reporting checks it never made

latekvo:
Found while running the full E2E harness against the packed 0.17.0 bundle for the release check. All of these are harness bugs, not product bugs.

## The core problem

The harness reported results it had not actually measured.

`assert_reject` looked for zod issues on **stdout**, but the CLI prints them on stderr and leaves stdout empty on a rejection. Both arms of the check called `pass()`, so the miss was invisible: the zod-path argument every case passes in was decorative, and a CLI that simply exited 1 for every input would have scored the same 88/88. **All 88 validation cases asserted nothing beyond "exit code was non-zero."**

Once that check could fail, it exposed the rest:

- **awk has no `\b` word boundary.** `/\bnumber\b/` matches nothing, so every flag kind resolved to `unknown`, `model_number_flags` was always empty, and **the bad-type matrix never emitted a single case** - while `_build_args` handed the string `"x"` to every required numeric/boolean/object field.
- The describe parser matched `--[a-zA-Z0-9-]+`, **truncating every snake_case flag at the underscore** (`--device_id` → `device`), and kept the `-json` display suffix (`--selector-json`) that is not part of the schema key. Both produced payloads carrying keys no schema knows.
- `missing-required` asserted `invalid_type` for enum flags, which zod reports as `invalid_value`.

Parsed kinds went from **270 unknown / 21 enum / 1 array** to 131 string, 68 number, 28 boolean, 25 integer, 21 enum, 16 array, 3 object, 2 unknown (the last two are `any`, which is correct).

## Teardown and the exit code

`run-e2e.sh` had no `trap`. An unbound variable inside a phase is fatal to the whole shell - `run_phase || warn` does not catch it - so an abort skipped the cleanup phase, the report, and `rm -rf $E2E_WORK`, leaking a ~500MB sandbox plus a tool-server, Electron, Metro and a fixture http server. Report generation, teardown and the exit code now live in an EXIT trap, with a completion flag: without it an aborted run exits **0** on "0 failures", which is verified below by mutation.

Teardown runs *before* the sandbox is removed - `$E2E_HOME` holds the detached tool-server's pid file and that server is started with no idle timeout, so deleting the sandbox first strands a process `argent server stop` can no longer find.

`90-cleanup.sh` gated `stop-all-simulator-servers` and `stop-metro` on `ARGENT_TOOLS_URL`, which the harness never sets (`ensure_server` deliberately does not pin it), so **neither ever ran** and every run left a simulator-server and Metro alive - while `teardown-complete` passed regardless.

## Other findings fixed

- **`--android-avd` could never work**: `boot-device` was asked for 840s inside `run_tool`'s 120s timeout, and the booted serial stayed in a `local`, so the RN tier blanket-skipped all 24 debugger/profiler tools for want of a device that was right there. Nothing shut the emulator down either - the tool-server only reaps devices Lens booted, not ones `boot-device` started.
- **The validation tier was byte-identical green against a dead server.** `argent run` does no client-side validation, so every call failed with a transport error that no one distinguished from a rejection. It refuses to run instead, and no longer counts `timeout(1)`'s 124-127 as a rejection.
- **The post-uninstall restore** hardcoded `--omit=optional` and the sandbox prefix - dropping electron on a full run (the chromium tier then skipped itself as "electron not installed"), and under `--system` restoring the machine's real global `argent` into the sandbox instead of putting it back.
- **The Android teardown claimed to leave an injected device alone** and then issued the same `stop-simulator-server` call as the self-booted branch, discarded the result, and recorded a skip saying it had not - tearing down the allocator's device service immediately before the RN tier drove that same serial.
- **The chromium port-owner fallback** matched `:$port` as a substring and took the first pid anywhere in `ss -ltnp` output, so cleaning up `:1234` would kill a process listening on `:12345`.
- **The RN tier's `stop-metro`** sat after every early return (a failed `debugger-connect` left Metro running) and ran unconditionally, killing a developer's own Metro on :8081 that the harness never started.
- **The chromium display gate** accepted an installed `xvfb-run` as a substitute for `DISPLAY`, but nothing wraps the Electron spawn in it - so a headless box with it installed cleared the gate and failed `boot-device`, turning the gate red for exactly the environment it exists to skip.
- A phase file with a syntax error silently re-ran the **previous** phase under the new phase's name; an unknown `--phase` selected nothing and exited 0; `-h` printed 7 lines of source code; a value-taking flag with no value died on an unbound `$2`; two runs in the same second shared one results file.
- `flow-add-step` declares its own `args` property, which makes `--args` that field's value rather than the whole-payload escape hatch, so `run_tool` cannot address its fields. Skipped with that reason rather than failing on a payload the tool never received.

## Verification

Against `swmansion-argent-0.17.0.tgz`, `introspection,validation`:

| | before | after |
|---|---|---|
| totals | 177 pass / 0 fail / 5 skip | 221 pass / 0 fail / 6 skip |
| rejection cases naming the field they test | **0 of 88** | **130 of 131** |
| bad-type cases emitted | **0** | **43** |

- `assert_reject` checked against right-field, wrong-field, wrong-code, no-path and unreachable-binary inputs. Every one of those cases passes on the un-fixed code.
- Abort path: injected an unbound variable into a phase - cleanup phase runs, report written, sandbox removed, exit 1, `PARTIAL` reported. With the completion flag mutated to always-set, the same abort exits **0**, which is what pins it.
- awk `\b`, the `-json` suffix, the underscore truncation, the `ss` port collision and the `timeout` 124-127 contract were each reproduced standalone before being fixed.

## Found by running against a live Android emulator

The android tier executed end-to-end for the first time: **23 pass / 1 fail / 10 skip**, including the `screenshot-diff` pair that the original unbound-variable bug aborted on. Three further defects only surfaced by running:

- **Nine `pass`/`fail`/`skip` calls passed their message as the third argument**, which `_record()` takes as the *case name* — so the detail column was empty and the case string changed every run. The chromium boot failure was the visible cost: it recorded `case=" "` with **no diagnostic whatsoever**.
- **A call killed by `timeout(1)` writes nothing**, so `RT_OUT` is empty exactly when the operator most needs to know what happened. A hung `boot-device` was recorded as a failure with a blank detail. All nine ad-hoc detail formatters now share `rt_detail()`, which names the timeout (124) and a command `timeout` could not run (125-127).
- **The chromium display gate tested that `$DISPLAY` was non-empty, not that it was usable.** A display-manager session switch on this box left `$DISPLAY` pointing at a server the process holds no cookie for; Electron then bound its CDP port and never answered on it (verified: the TCP connection is accepted, the HTTP response never arrives), so the tier hard-failed instead of skipping.

Moving `profiler-stack-query` and `profiler-combined-report` into the native block also closed a hole their removal had opened: both were listed in `RN_TOOLS` and in the failure-arm skips but had no case on the success path, so a successful `react-profiler-start` left them with no pass, fail or skip at all.

## Not covered here

- The **postinstall assertion** is left alone - #556 already fixes it, and this branch does not touch it. Until that merges the install phase still records that one failure.
- The **RN tier still has not executed.** The emulator itself is fine — it cold-boots in ~24-30s and ran the whole android tier — but it dies from CPU starvation (`detected a hanging thread 'QEMU2 main loop'`) whenever this machine's load average climbs, which it does to 90-120 on 16 cores from concurrent work. Four attempts: it survived the android tier at load ~10, and was starved each time by a competing `npm install` or by Metro's own bundler startup. The four RN profiler fixes are verified against the tools' source contracts (the throw sites and their required params), not against a live run.
- The **chromium tier still has not executed** past `boot-device`, for the display reason above.
- Known remaining, deliberately left: `assert_ok` cannot fail for calls whose result is a constant `true`; the cookie/storage set→get pair never verifies the round-trip value; a tier-level skip writes one row so `report.py`'s coverage matrix silently drops the tools that tier would have recorded; `report.py` renders an empty or partial log as a clean green report.


latekvo (commented):

scripts/e2e-full/lib/common.sh
```diff
@@ -161,39 +182,65 @@ assert_true() { # phase tool case json-args jq-filter
...
 }
 
-# The tool call must FAIL (rc != 0). If `path` is given, and stdout is a zod
-# issue array, at least one issue must match that path (proves the specific
-# field was rejected, not just an unrelated error).
+# The zod issue array a rejected call produces, or empty if the output holds
+# none. The CLI prints issues on STDERR — stdout carries only successful results
+# and is empty on a rejection — but both streams are searched so a tool that
+# reports on stdout is still checked rather than waved through.
+zod_issues() {
```
latekvo, line 189: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: zod_issues() can never match the current CLI, so every validation case still passes as "(unstructured)".

`zod_issues` accepts a stream only if it parses as a bare top-level JSON array (`type=="array" and length>0 and all(.[]; has("path") and has("code"))`). The CLI stopped emitting that shape in #648 (`d08c3a3c fix(cli): name the missing flag instead of printing a Zod dump`, 2026-08-03), which is already an ancestor of this PR's base 4975ff49. Today every rejection funnels through `failValidation` in packages/argent-cli/src/run.ts:228: without `--json` it writes prose (`console.error(`Error: ${summary}\n`)`, run.ts:245) and, with `--json`, an *object* `{error, missing, issues}` — the issue array only ever appears nested under `.issues`, never as a top-level array. `run_tool` (common.sh:69) never passes `--json`. So `zod_issues` returns 1 for every generated case and `assert_reject` takes the `pass "$phase" "$tool" "$case" "rc=$RT_RC (unstructured)"` arm at line 228 — the exact "asserted nothing beyond exit code was non-zero" behaviour the PR exists to remove. This also makes the new enum/`invalid_value` code selection in 20-validation.sh:99-101 dead.

**Trigger.** `bash scripts/e2e-full/run-e2e.sh --phase introspection,validation` against a tgz packed from this branch. For `bad-enum:button` on `button`, `argent run button --args '{"udid":"x","button":"__not_a_valid_enum__"}'` → tool-server 400 with the zod issue array → `describeServerValidationFailure` recognises it → stderr is `Error: --button Invalid option: ...`, stdout is the help block. Neither parses as a JSON array, so `zod_issues` fails and the case is recorded `pass ... rc=2 (unstructured)` instead of checking that the issue names `button`.

**Checked against.** common.sh:185-192 `# The zod issue array a rejected call produces... The CLI prints issues on STDERR — stdout carries only successful results and is empty on a rejection` / `if printf '%s' "$s" | jq -e 'type=="array" and length>0 ...'`. Contradicted by packages/argent-cli/src/run.ts:244-247 `} else { console.error(`Error: ${summary}\n`); printToolHelp(meta); }` (printToolHelp uses console.log → stdout) and by the CLI's own test, packages/argent-cli/test/run-required-flags.test.ts:174-176: `expect(stderr()).toContain("Error: --x Too big: expected <=1"); expect(stderr()).not.toContain('"code"'); expect(stdout()).toContain("argent run gesture-tap [flags]");`.

scripts/e2e-full/phases/90-cleanup.sh
```diff
@@ -16,9 +37,16 @@ run_phase() {
...
   # own description and the skills tell agents to do.
-  if [ -n "${ARGENT_TOOLS_URL:-}" ]; then
-    run_tool stop-all-simulator-servers '{}' >/dev/null 2>&1 && pass "$P" stop-all-simulator-servers teardown || skip "$P" stop-all-simulator-servers teardown "no server/none running"
+  #
+  # That same sandbox discovery is why the gate is `server_running` rather than
+  # ARGENT_TOOLS_URL: the harness never pins that variable, so keying off it
+  # skips both teardowns and leaves a simulator-server and Metro alive.
+  if server_running; then
+    run_tool stop-all-simulator-servers '{}' >/dev/null 2>&1 && pass "$P" stop-all-simulator-servers teardown || skip "$P" stop-all-simulator-servers teardown "none running"
     run_tool stop-metro '{}' >/dev/null 2>&1 && pass "$P" stop-metro teardown || skip "$P" stop-metro teardown "no metro"
```
latekvo, line 46: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Cleanup now kills whatever listens on :8081 — a developer's Metro the harness never started.

The PR replaces the `[ -n "${ARGENT_TOOLS_URL:-}" ]` gate with `server_running`. As the PR body and the new comment both state, `ARGENT_TOOLS_URL` is never set by the harness, so this branch previously never executed. It now executes on every run where a tool-server is reachable, and `run_tool stop-metro '{}'` sends *no* port, so the tool falls back to its schema default of 8081 (`.default(8081)`, stop-metro.ts:91) and SIGTERMs every process listening there on the real host (`lsof -ti tcp:8081 -sTCP:LISTEN`, stop-metro.ts:74) — the sandboxed HOME gives no protection, the lookup is host-wide. The run then records `pass cleanup stop-metro teardown`. This directly undoes the protection the same PR adds in the RN tier and contradicts its comment, and silently falsifies the README's untouched Isolation section ("safe to run on a shared box"). The same missing `port` applies to `_rn_stop_metro`'s `assert_ok "$P" stop-metro stop "{}"` (50-rn-bluesky.sh:77): with the documented `E2E_METRO_PORT=8082` override the tier starts Metro on 8082 and then stops the one on 8081.

**Trigger.** Developer has Metro serving :8081 for an unrelated project. Run `bash scripts/e2e-full/run-e2e.sh --skip-install --phase introspection` (the rn tier is not even selected). Introspection leaves a server up (`ensure_server`, 10-introspection.sh:113); the EXIT trap runs 90-cleanup; `server_running` is true; `run_tool stop-metro '{}'` resolves port 8081 and kills the developer's Metro, recorded as a pass.

**Checked against.** 90-cleanup.sh:44-46 `if server_running; then ... run_tool stop-metro '{}' >/dev/null 2>&1 && pass "$P" stop-metro teardown || skip ...`. packages/tool-server/src/tools/simulator/stop-metro.ts:85-93 `port: z.number().int().min(1).max(65535).default(8081)` and :74 `execFileSync("lsof", ["-ti", `tcp:${port}`, "-sTCP:LISTEN"])`. Contradicted by 50-rn-bluesky.sh:70-72 `# Stop only a Metro this phase started. A developer's own Metro on :8081 is not ours to kill`. Pre-PR gate: `git show 4975ff49:scripts/e2e-full/lib/common.sh | grep ARGENT_TOOLS_URL` shows only `# ... we do NOT pin a port or ARGENT_TOOLS_URL`, i.e. the variable is never exported anywhere under scripts/e2e-full.

scripts/e2e-full/phases/50-rn-bluesky.sh
```diff
@@ -55,30 +56,58 @@ run_phase() {
...
   fi
-  if _metro_ready "$MPORT"; then pass "$P" metro ready; else fail "$P" metro ready "Metro not up on :$MPORT"; _skip_all "Metro unavailable"; return 0; fi
+  # Stop only a Metro this phase started. A developer's own Metro on :8081 is
+  # not ours to kill, and every early return past this point has to run it or
+  # the one we did start outlives the run.
+  _rn_stop_metro() {
+    if [ "$STARTED_METRO" -ne 1 ]; then
+      skip "$P" stop-metro stop "Metro was already running on :$MPORT; left as found"
+    elif _metro_ready "$MPORT"; then
+      assert_ok "$P" stop-metro stop "{}"
```
latekvo, line 77: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: _rn_stop_metro stops port 8081 regardless of E2E_METRO_PORT, so a non-default run leaks its own Metro and kills someone else's.

The tier starts Metro on `$MPORT` (`E2E_METRO_PORT`, default 8081) at line 64 and checks readiness with `_metro_ready "$MPORT"` at line 76, but then calls `assert_ok "$P" stop-metro stop "{}"` with an empty payload. stop-metro's schema defaults `port` to 8081 (stop-metro.ts:86-92), so with E2E_METRO_PORT=8082 the call SIGTERMs whatever holds :8081 — a process this tier never started, which is precisely what the new comment at lines 70-72 says must not happen — while the Metro the tier did start on :8082 survives the run (90-cleanup's E2E_METRO_PID kill is the only remaining reaper, and it is skipped whenever the harness aborts before the trap can reach it with ARGENT_BIN set). assert_ok only checks rc, and stop-metro returns `{stopped:false}` with exit 0 when it finds nothing (stop-metro.ts:114-116), so the case records pass for a teardown that did nothing.

**Trigger.** `E2E_METRO_PORT=8082 bash scripts/e2e-full/run-e2e.sh --phase rn --android-serial <serial>` — the documented env override at 50-rn-bluesky.sh:19. The tier starts Metro on 8082, `_rn_stop_metro` calls stop-metro with no port, 8081 is targeted, and `pass rn stop-metro stop` is recorded.

**Checked against.** 50-rn-bluesky.sh:29 `local MPORT="${E2E_METRO_PORT:-8081}"`; :64 `( cd "$RN_DIR" && exec $expo start --dev-client --port "$MPORT" )`; :76-77 `elif _metro_ready "$MPORT"; then assert_ok "$P" stop-metro stop "{}"`. Sibling call sites that do parameterise the port do not exist in the harness; stop-metro.ts:106 description: "Stop the Metro bundler process listening on a given port (default 8081)".

scripts/e2e-full/phases/30-android.sh
```diff
@@ -105,12 +118,16 @@ run_phase() {
...
-  if [ -z "${E2E_ANDROID_SERIAL:-}" ]; then
-    assert_ok "$P" stop-simulator-server stop "$U"
+  # Nothing is reaped here, on either branch. An injected device is the caller's
+  # (on a shared machine, the allocator's). One this phase booted is still in
+  # use: the RN tier drives the same serial straight after this, so both the
+  # per-device service and the emulator have to outlive this phase. 90-cleanup
+  # owns that reap, and it runs from run-e2e.sh's EXIT trap, so an aborted run
+  # reaps the AVD too — which ending the phase here would not.
+  if [ "$BOOTED_HERE" -eq 1 ]; then
+    export E2E_ANDROID_REAP_SERIAL="$DEV"
```
latekvo, line 128: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: A self-booted AVD leaks on an abort during the android tier — E2E_ANDROID_REAP_SERIAL is only exported at the end of the phase.

`export E2E_ANDROID_REAP_SERIAL="$DEV"` is the last statement of run_phase, ~70 lines and 30+ tool calls after boot-device succeeded at line 47-56. 90-cleanup's reap block is entirely gated on that variable (90-cleanup.sh:12), so any exit from the harness between the successful boot and the end of the android tier leaves the emulator running with nothing to shut it down — the tool-server only reaps devices Lens booted, as the PR's own comment at 90-cleanup.sh:18-20 states. This falsifies both the README line this PR adds and the comment at 30-android.sh:124-126 ("90-cleanup owns that reap, and it runs from run-e2e.sh's EXIT trap, so an aborted run reaps the AVD too"). The one-line fix — exporting it beside `BOOTED_HERE=1` at line 51 — would close the whole window.

**Trigger.** `bash scripts/e2e-full/run-e2e.sh --android-avd Pixel_9a` and press Ctrl-C during the android tier (screenshots + 30 gesture/app-lifecycle calls). `trap 'exit 130' INT` -> finish() -> run_one cleanup, where `${E2E_ANDROID_REAP_SERIAL:-}` is empty, so neither `stop-simulator-server` nor `adb emu kill` runs and the emulator the harness booted stays up. Same for a `set -u` abort anywhere in the tier body — the exact failure class this PR exists to fix.

**Checked against.** 30-android.sh:49-56 `if [ "$RT_RC" -eq 0 ] && ... ; then DEV=...; BOOTED_HERE=1; export E2E_ANDROID_SERIAL="$DEV"; pass ...` (no reap export here) vs :127-129 `if [ "$BOOTED_HERE" -eq 1 ]; then export E2E_ANDROID_REAP_SERIAL="$DEV"; skip ...`. 90-cleanup.sh:12 `if [ -n "${E2E_ANDROID_REAP_SERIAL:-}" ]; then`. README.md: "the emulator is shut down in the cleanup phase — after every tier that uses it, and on an aborted run too."

scripts/e2e-full/phases/30-android.sh
```diff
@@ -27,24 +27,35 @@ run_phase() {
...
-      pass "$P" boot-device "booted $DEV"
+      BOOTED_HERE=1
+      # Publish it: the RN tier resolves its device from E2E_ANDROID_SERIAL, so
+      # without this `--android-avd` drives this tier fully and then blanket-skips
+      # all 24 debugger/profiler tools for want of a device that is right there.
+      export E2E_ANDROID_SERIAL="$DEV"
+      pass "$P" boot-device boot "booted $DEV"
     else
-      fail "$P" boot-device "$(printf '%s' "$RT_OUT" | tr '\n' ' ' | cut -c1-160)"; return 0
+      fail "$P" boot-device boot "$(rt_detail 160)"; return 0
```
latekvo, line 58: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: A boot-device timeout is reported as "timed out after 120s" when the call was actually given 900s.

The phase raises TOOL_TIMEOUT to `840000/1000 + 60` = 900 for the boot-device call, then restores it to the previous value (120) on line 48 — before line 58 evaluates `$(rt_detail 160)`. timeout(1) kills the CLI before it writes anything, so RT_OUT is empty and rt_detail falls into the `124)` arm, which prints `"timed out after %ss, no output" "$TOOL_TIMEOUT"` — reading the already-restored 120. The operator is told the boot was killed at 2 minutes when it was actually killed at 15, which points the investigation straight back at the run_tool/boot-device timeout mismatch this hunk exists to fix. Contrast assert_reject's sibling message (common.sh:217), which interpolates TOOL_TIMEOUT while it is still the value the call ran under.

**Trigger.** `bash scripts/e2e-full/run-e2e.sh --android-avd <avd>` with an AVD that does not reach a booted state within 900s (the documented starvation failure mode on a loaded box). run_tool returns RT_RC=124 with empty RT_OUT; the report records `[android] boot-device — boot: timed out after 120s, no output`.

**Checked against.** 30-android.sh:45-48 `local BOOT_MS=840000 prev_timeout="$TOOL_TIMEOUT"; TOOL_TIMEOUT=$(( BOOT_MS / 1000 + 60 )); run_tool boot-device "..."; TOOL_TIMEOUT="$prev_timeout"` then :58 `fail "$P" boot-device boot "$(rt_detail 160)"`. common.sh:131-132 `case "$RT_RC" in 124) printf 'timed out after %ss, no output' "$TOOL_TIMEOUT" ;;`.

scripts/e2e-full/phases/20-validation.sh
```diff
@@ -57,7 +57,16 @@ PY
 
 run_phase() {
   local P=validation
-  ensure_server || warn "no private server; validation still runs (client-side schema)"
+  # Every verdict here comes from the server's schema: `argent run` does no
+  # client-side validation, it posts the payload. With no server every call
```
latekvo, line 61: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The comment justifying the new hard failure claims `argent run` does no client-side validation, which run.ts contradicts.

The comment states "Every verdict here comes from the server's schema: `argent run` does no client-side validation, it posts the payload." run.ts:332-335 does exactly that validation locally: `const missing = findMissingRequired(payload, schema); if (missing.length > 0) { await failValidation({ missing, invalid: [], rawIssues: null }, "cli_run_required_flags"); }` — the call never leaves the process. Since `missing-required` is the case generated for every tool with required flags (20-validation.sh:101), the largest single group of cases in this tier is decided entirely client-side, which is the opposite of what the comment tells the next reader. run-validation.ts:5-7 documents the same split explicitly ("a required flag was not passed. Knowable here, from the tool's published schema, so it is caught before the call goes out").

**Trigger.** Any reader of 20-validation.sh trying to understand why the tier hard-fails without a server, or trying to extend it — e.g. deciding whether a new missing-required case needs the server at all. `argent run gesture-tap --args '{}'` with no tool-server reachable still produces a schema-driven "missing required flags --udid, --x, --y" rejection from the CLI alone.

**Checked against.** 20-validation.sh:60-65 comment vs packages/argent-cli/src/run.ts:329-335 `// Checked against the merged payload, so a required field supplied through --args ... const missing = findMissingRequired(payload, schema); if (missing.length > 0) { await failValidation(...) }` and run-validation.ts:44-58 `export function findMissingRequired(payload, schema)`.

scripts/e2e-full/phases/50-rn-bluesky.sh
```diff
@@ -115,31 +144,51 @@ run_phase() {
...
+  # run, so calling them earlier fails on every healthy build.
   run_tool native-profiler-start "{\"device_id\":\"$DEV\",\"app_process\":\"$PKG\"}"
   if [ "$RT_RC" -eq 0 ]; then
     pass "$P" native-profiler-start np-start
     run_tool gesture-swipe "{\"udid\":\"$DEV\",\"fromX\":0.5,\"fromY\":0.7,\"toX\":0.5,\"toY\":0.3}" >/dev/null 2>&1; sleep 3
     assert_ok "$P" native-profiler-stop    np-stop   "{\"device_id\":\"$DEV\"}"
     assert_ok "$P" native-profiler-analyze np-analyze "{\"device_id\":\"$DEV\"}"
-    assert_ok "$P" profiler-load load-native "{\"mode\":\"load_native\",\"device_id\":\"$DEV\",\"app_process\":\"$PKG\"}"
+    assert_ok "$P" profiler-stack-query stacks "{$D,\"mode\":\"thread_breakdown\"}"
+    assert_ok "$P" profiler-combined-report combined "{$D}"
```
latekvo, line 174: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: profiler-combined-report was moved into the native block only, so a react-profiler failure no longer reaches it.

profiler-combined-report requires BOTH profilers, so moving it wholesale into the native block fixes one direction and opens the other. The tool gates on the native trace and then on React data — profiler-combined-report.ts:138-145 throws "No React commit data. Run react-profiler-analyze first." when getCachedProfilerPaths has no commitsPath. Trigger: react-profiler-start fails (rc != 0) while native-profiler-start succeeds. The react failure arm at 50-rn-bluesky.sh:155-157 no longer lists profiler-combined-report among its skips (pre-PR it did), so the native arm still runs assert_ok at :174; the call throws, is recorded as a fail, and run-e2e.sh:171 exits 1 — even though the react-profiler-start failure itself was recorded only as a skip, making this the single red in a run that is otherwise a clean environmental skip. profiler-stack-query is unaffected (it never reads React data), so the placement is correct for it; only combined-report needs the conjunction of both blocks succeeding, plus a skip in the react failure arm.

PR #598, opened 2026-07-30
chore: add server.json for the official MCP registry

latekvo:
Adds the two files the official MCP registry (`registry.modelcontextprotocol.io`) needs in order to list argent. Argent is not in that registry today (`?search=io.github.software-mansion` returns 0 servers).

The registry is metadata-only — it points at the npm package rather than hosting anything.

### `server.json`

Describes the server and how a client starts it: `npx @swmansion/argent mcp`. The `mcp` positional is the dispatcher case in `packages/argent/src/cli.ts:137`, and it is the same subcommand `argent init` writes into every client config — `packages/argent-installer/src/mcp-configs.ts:130-141` spells it four ways (`node <path> mcp`, `yarn argent mcp`, `npx --no-install argent mcp`, and the global `{ command: MCP_BINARY_NAME, args: ["mcp"] }`, where `MCP_BINARY_NAME` is `"argent"`). None of them is literally the npx-by-package-name form in `server.json`, which is the form the registry expects a client to construct from `identifier` + `runtimeHint` — so it is verified below by running it, not by matching a string.

`repository.id` is the GitHub repo id (`1164885639`); the schema recommends it so a registry can detect a delete-and-recreate of the repo.

### `mcpName` in `packages/argent/package.json`

The ownership proof. The registry fetches the npm package and requires `mcpName` to equal the server name in `server.json` before it will accept a publish under `io.github.software-mansion`. It is the field the registry's own quickstart names for npm packages. Nothing in the publish path drops it: `scripts/pack-mcp.cjs` shells out to `npm pack` without rewriting the manifest, and the canary path's `scripts/next-canary-version.mjs --write` round-trips the JSON and only sets `version`.

### Namespace

`io.github.software-mansion/argent` — GitHub OAuth auth. The registry grants an org namespace only to org **Owners**; plain membership is no longer enough. If no Owner is available to run the publish, the alternative is DNS auth via an apex TXT record on `swmansion.com`, which changes the name to `com.swmansion/argent` (and the `mcpName` value with it).

### Version sync

`server.json` sits at the repo root, outside `packages/*`, so a release bump would have left its two `0.18.0`s untouched — and the registry rejects a publish whose `packages[].version` names a version npm has never seen. `scripts/check-workspace-versions.mjs` (already run by the Static checks job) also holds `server.json` to the version in `packages/argent/package.json` — the package it names — and pins the two cross-file links a publish depends on: `name` == `mcpName`, `identifier` == the published package name. Both halves are reported in one run, so a half-finished bump doesn't cost a CI round-trip per file.

### Not in this PR

- **Publishing.** Nothing has been published; this is just the manifest. Publishing is `mcp-publisher login github && mcp-publisher publish`, and it can only succeed **after** a release whose npm tarball carries `mcpName` — i.e. not against `0.18.0`, which is `latest` (published 2026-07-30) and predates `mcpName` — `curl registry.npmjs.org/@swmansion%2Fargent/0.18.0 | jq .mcpName` → `null`. The first publishable release is the next bump.
- **`github.com/mcp`.** A separate, manual step. Onboarding a new server there is hand-curated: per [github/github-mcp-server#1257](https://github.com/github/github-mcp-server/discussions/1257) (trent-j, answering for GitHub — “a few clarifications from our side”, “our product team” — 19 May 2026; GitHub's API reports `authorAssociation: NONE` for them on that repo) "onboarding a new server is still a manual curation process today. Once a server has been onboarded, newly published versions from the OSS registry should sync from there." The nomination channel is `partnerships@github.com`, from GitHub's [registry how-to post](https://github.blog/ai-and-ml/generative-ai/how-to-find-install-and-manage-mcp-servers-with-the-github-mcp-registry/) (24 Oct 2025) — that address does not appear anywhere in the discussion thread.

### Verification

- `mcp-publisher validate` (v1.8.0, which posts the file to the live registry) → `server.json is valid`. Mutation-checked what that actually covers: removing `transport` fails it and removing `$schema` is a 422, but a foreign namespace (`io.github.anthropics/argent`), a nonexistent npm identifier and a `99.0.0` package version all still pass — so it checks shape only, and the ownership/existence checks happen at publish time. Also validated offline against the pinned `2025-12-11` schema.
- Ran the entrypoint the file declares. `npx -y @swmansion/argent@latest mcp` fed an `initialize` over stdio answered `{"protocolVersion":"2024-11-05","capabilities":{"tools":{}},"serverInfo":{"name":"argent","version":"0.18.0"},"instructions":"…"}`, with stderr empty.
- Mutation-checked the new version guard: drifting `server.json` `version`, drifting `packages[0].version`, deleting `mcpName`, and typo-ing `identifier` each fail `check-workspace-versions.mjs` with the specific line; the pre-existing packages/* drift check still fires on its own.
- Asserted across the two files: `mcpName` == `server.json` `name`, `identifier` == package `name`, all three versions equal, description 89 chars (registry limit is 100).
- `prettier --check` and `eslint` clean on both changed files.
- Mutation-checked the guard as a whole — 23 mutants, 2 survivors. Killed: reverting the self-invocation guard to a raw `argv[1]` compare and making it never match; `exit(1)` → `exit(0)` on the failure path; dropping either `failed = true`; exiting early on `packages/*` drift instead of collecting; `packages.slice(0, 1)`; hardcoding `packages[0]` in either message; dropping the no-packages-array guard, the non-object-entry guard, the non-object-JSON guard, and the presence checks for `version`, `name` and `mcpName`; `continue` → `break` on the unparseable-manifest and version-less arms of the `packages/*` scan, and in the entry loop; dropping the on-message read, the `git checkout` hint, and the per-version name accumulator. The two survivors are both scan-order concerns no single run can observe: the `packages/*` readdir sort (there so a scan that stops early drops the same packages on every machine rather than a filesystem-dependent set), and the `!entry.isDirectory()` skip, which no fixture reaches because every fixture puts only directories under `packages/`.
- Reproduced the symlink no-op on the real repo: with `packages/registry` at `9.9.9` and `server.json` at `0.17.0`, `node /tmp/argent-alias/scripts/check-workspace-versions.mjs` now exits 1 and prints both problems, where before it exited 0 with no output at all.
- Closed the vacuous-pass class in `serverJsonMismatches`: all four comparisons it makes were bare `!==`, so a field absent from *both* sides compared equal and passed. A `packages/argent/package.json` with no `version` printed `All workspace packages and server.json are at undefined.` and exited 0; with no `name`, a `server.json` naming nothing installable was reported in sync. `version` and `name` now require presence, as `mcpName` already did.
- `readTrackedJson` guarantees an object, not just parseable bytes — a `server.json` holding `null` used to reach the field reads and throw, and `"0.18.0"` or `[]` produced mismatch lines that never said the file was the wrong shape.
- Full CI parity locally: `test:scripts` 86/86, `tsc -p tsconfig.scripts.json`, `eslint --max-warnings 0`, `prettier --check`.

### DIFF
--- a/scripts/check-workspace-versions.mjs
+++ b/scripts/check-workspace-versions.mjs
@@ -32,7 +32,14 @@ const serverJsonPath = join(repoRoot, "server.json");
 export function serverJsonMismatches(publishedVersion, argentPkg, server) {
   const mismatches = [];
 
-  if (server.version !== publishedVersion) {
+  // Without a version on the manifest there is no target to compare against, and
+  // every comparison below would pass by matching undefined against undefined —
+  // so its absence is a mismatch in its own right, exactly like mcpName's.
+  if (!publishedVersion) {
+    mismatches.push(
+      "packages/argent/package.json has no version — nothing for server.json to be checked against"
+    );
+  } else if (server.version !== publishedVersion) {
     mismatches.push(
       `server.json version is ${server.version}, packages/argent/package.json is ${publishedVersion}`
     );
@@ -141,8 +148,8 @@ function main() {
     failed = true;
   }
 
-  // An empty packages/ never gets past this read, so the version below is always
-  // a real one.
+  // An empty packages/ never gets past this read. It only proves the file parses,
+  // though — a manifest missing its version is caught as a mismatch below.
   const argentPkg = readTrackedJson(argentManifestPath);
   const server = readTrackedJson(serverJsonPath);
 
--- a/scripts/check-workspace-versions.mjs
+++ b/scripts/check-workspace-versions.mjs
@@ -155,7 +155,10 @@ function main() {
     } catch {
       continue; // directory without a package.json
     }
-    if (!manifest.version) continue;
+    // `?.` because JSON.parse succeeds on a file holding `null`, which the catch
+    // above never sees — the same shape readTrackedJson rejects for the two
+    // manifests it reads.
+    if (!manifest?.version) continue;
     const names = byVersion.get(manifest.version) ?? [];
     names.push(manifest.name ?? entry.name);
     byVersion.set(manifest.version, names);
--- a/scripts/check-workspace-versions.mjs
+++ b/scripts/check-workspace-versions.mjs
@@ -119,13 +119,9 @@ function main() {
     byVersion.set(manifest.version, names);
   }
 
-  // An empty packages/ never gets past this read, so the version below is always
-  // a real one.
-  const argentPkg = readTrackedJson(argentManifestPath);
-  const server = readTrackedJson(serverJsonPath);
-
   let failed = false;
 
+  // Reported before the reads below, so a corrupt server.json can't swallow it.
   if (byVersion.size > 1) {
     console.error("Workspace package versions are out of sync:");
     for (const [version, names] of [...byVersion].sort()) {
@@ -136,6 +132,11 @@ function main() {
     );
     failed = true;
   }
+
+  // An empty packages/ never gets past this read, so the version below is always
+  // a real one.
+  const argentPkg = readTrackedJson(argentManifestPath);
+  const server = readTrackedJson(serverJsonPath);
 
   // server.json names the npm coordinates of packages/argent, so that manifest
   // is what it has to agree with. Reported even when packages/* disagree, so a
--- a/scripts/check-workspace-versions.mjs
+++ b/scripts/check-workspace-versions.mjs
@@ -109,18 +109,36 @@ function readTrackedJson(path) {
     console.error(`\nIt is tracked in git — restore it with \`git checkout -- ${shown}\`.`);
     process.exit(1);
   }
+  let parsed;
   try {
-    return JSON.parse(raw);
+    parsed = JSON.parse(raw);
   } catch (error) {
     console.error(`${shown} is not valid JSON: ${error.message}`);
     console.error(`\nIt is tracked in git — restore it with \`git checkout -- ${shown}\`.`);
     process.exit(1);
   }
+
+  // JSON.parse returns whatever the file holds, so a scalar, null or an array
+  // parses cleanly and then either throws or compares equal to undefined on
+  // every field read below. Both callers want an object or nothing.
+  if (parsed === null || typeof parsed !== "object" || Array.isArray(parsed)) {
+    const shape = parsed === null ? "null" : Array.isArray(parsed) ? "an array" : typeof parsed;
+    console.error(`${shown} is ${shape}, not a JSON object`);
+    console.error(`\nIt is tracked in git — restore it with \`git checkout -- ${shown}\`.`);
+    process.exit(1);
+  }
+  return parsed;
 }
 
 function main() {
   const byVersion = new Map(); // version -> package names
-  for (const entry of readdirSync(packagesDir, { withFileTypes: true })) {
+  // Sorted so the scan visits packages/* in the same order on every machine —
+  // readdir order is filesystem-dependent, and a scan that stops early would
+  // otherwise drop a different set of packages each run.
+  const entries = readdirSync(packagesDir, { withFileTypes: true }).sort((a, b) =>
+    a.name < b.name ? -1 : a.name > b.name ? 1 : 0
+  );
+  for (const entry of entries) {
     if (!entry.isDirectory()) continue;
     let manifest;
     try {
--- a/scripts/check-workspace-versions.mjs
+++ b/scripts/check-workspace-versions.mjs
@@ -10,74 +10,112 @@
 // rejects one whose `name` doesn't match the npm package's `mcpName`.
 import { readFileSync, readdirSync } from "node:fs";
 import { dirname, join } from "node:path";
-import { fileURLToPath } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
+import { argv as processArgv } from "node:process";
 
 const repoRoot = join(dirname(fileURLToPath(import.meta.url)), "..");
 const packagesDir = join(repoRoot, "packages");
 
-const byVersion = new Map(); // version -> package names
-for (const entry of readdirSync(packagesDir, { withFileTypes: true })) {
-  if (!entry.isDirectory()) continue;
-  let manifest;
-  try {
-    manifest = JSON.parse(readFileSync(join(packagesDir, entry.name, "package.json"), "utf8"));
-  } catch {
-    continue; // directory without a package.json
+/**
+ * Everything server.json has to keep in step with the workspace, as one
+ * human-readable line per problem. Pure, separated from the fs reads so it can
+ * be unit-tested.
+ * @param {string | undefined} workspaceVersion the single version every packages/* manifest carries
+ * @param {{ name?: string, mcpName?: string }} argentPkg packages/argent/package.json
+ * @param {{ name?: string, version?: string, packages?: { identifier?: string, version?: string }[] }} server server.json
+ * @returns {string[]} empty when in sync
+ */
+export function serverJsonMismatches(workspaceVersion, argentPkg, server) {
+  const mismatches = [];
+
+  if (server.version !== workspaceVersion) {
+    mismatches.push(`server.json version is ${server.version}, workspace is ${workspaceVersion}`);
   }
-  if (!manifest.version) continue;
-  const names = byVersion.get(manifest.version) ?? [];
-  names.push(manifest.name ?? entry.name);
-  byVersion.set(manifest.version, names);
-}
 
-if (byVersion.size > 1) {
-  console.error("Workspace package versions are out of sync:");
-  for (const [version, names] of [...byVersion].sort()) {
-    console.error(`  ${version}: ${names.sort().join(", ")}`);
-  }
-  console.error(
-    "\nEvery package under packages/* must share one version. Bump the outliers to match."
-  );
-  process.exit(1);
-}
-
-const [workspaceVersion] = byVersion.keys();
-
-const argentPkg = JSON.parse(readFileSync(join(packagesDir, "argent", "package.json"), "utf8"));
-const server = JSON.parse(readFileSync(join(repoRoot, "server.json"), "utf8"));
-
-const mismatches = [];
-if (server.version !== workspaceVersion) {
-  mismatches.push(`server.json version is ${server.version}, workspace is ${workspaceVersion}`);
-}
-if (server.name !== argentPkg.mcpName) {
-  mismatches.push(
-    `server.json name (${server.name}) != packages/argent/package.json mcpName (${argentPkg.mcpName})`
-  );
-}
-for (const [i, pkg] of (server.packages ?? []).entries()) {
-  if (pkg.version !== workspaceVersion) {
+  // mcpName on the published npm tarball is what proves to the registry that the
+  // io.github.software-mansion namespace is ours, so its absence is as fatal as a
+  // mismatch — and an absent one equals an absent server.json name by ===.
+  if (!argentPkg.mcpName) {
     mismatches.push(
-      `server.json packages[${i}].version is ${pkg.version}, workspace is ${workspaceVersion}`
+      "packages/argent/package.json has no mcpName — the registry reads it off the published tarball to prove the namespace"
+    );
+  } else if (server.name !== argentPkg.mcpName) {
+    mismatches.push(
+      `server.json name (${server.name}) != packages/argent/package.json mcpName (${argentPkg.mcpName})`
     );
   }
-  if (pkg.identifier !== argentPkg.name) {
+
+  const packages = server.packages ?? [];
+  if (packages.length === 0) {
     mismatches.push(
-      `server.json packages[${i}].identifier (${pkg.identifier}) != published package name (${argentPkg.name})`
+      "server.json lists no packages — the registry entry would name nothing to install"
     );
   }
+  for (const [i, pkg] of packages.entries()) {
+    if (pkg.version !== workspaceVersion) {
+      mismatches.push(
+        `server.json packages[${i}].version is ${pkg.version}, workspace is ${workspaceVersion}`
+      );
+    }
+    if (pkg.identifier !== argentPkg.name) {
+      mismatches.push(
+        `server.json packages[${i}].identifier (${pkg.identifier}) != published package name (${argentPkg.name})`
+      );
+    }
+  }
+
+  return mismatches;
 }
 
-if (mismatches.length > 0) {
-  console.error("server.json is out of sync with the workspace:");
-  for (const line of mismatches) console.error(`  ${line}`);
-  console.error(
-    "\nserver.json is not under packages/*, so a release bump misses it — keep it in step\n" +
-      "with packages/argent/package.json by hand."
-  );
-  process.exit(1);
+function main() {
+  const byVersion = new Map(); // version -> package names
+  for (const entry of readdirSync(packagesDir, { withFileTypes: true })) {
+    if (!entry.isDirectory()) continue;
+    let manifest;
+    try {
+      manifest = JSON.parse(readFileSync(join(packagesDir, entry.name, "package.json"), "utf8"));
+    } catch {
+      continue; // directory without a package.json
+    }
+    if (!manifest.version) continue;
+    const names = byVersion.get(manifest.version) ?? [];
+    names.push(manifest.name ?? entry.name);
+    byVersion.set(manifest.version, names);
+  }
+
+  if (byVersion.size > 1) {
+    console.error("Workspace package versions are out of sync:");
+    for (const [version, names] of [...byVersion].sort()) {
+      console.error(`  ${version}: ${names.sort().join(", ")}`);
+    }
+    console.error(
+      "\nEvery package under packages/* must share one version. Bump the outliers to match."
+    );
+    process.exit(1);
+  }
+
+  const [workspaceVersion] = byVersion.keys();
+
+  const argentPkg = JSON.parse(readFileSync(join(packagesDir, "argent", "package.json"), "utf8"));
+  const server = JSON.parse(readFileSync(join(repoRoot, "server.json"), "utf8"));
+
+  const mismatches = serverJsonMismatches(workspaceVersion, argentPkg, server);
+  if (mismatches.length > 0) {
+    console.error("server.json is out of sync with the workspace:");
+    for (const line of mismatches) console.error(`  ${line}`);
+    console.error(
+      "\nserver.json is not under packages/*, so a release bump misses it — keep it in step\n" +
+        "with packages/argent/package.json by hand."
+    );
+    process.exit(1);
+  }
+
+  // An empty packages/ can't reach here: reading packages/argent/package.json above
+  // throws first, so workspaceVersion is always a real version by this point.
+  console.log(`All workspace packages and server.json are at ${workspaceVersion}.`);
 }
 
-// An empty packages/ can't reach here: reading packages/argent/package.json above
-// throws first, so workspaceVersion is always a real version by this point.
-console.log(`All workspace packages and server.json are at ${workspaceVersion}.`);
+// Run only when invoked directly, not when imported by the test.
+if (processArgv[1] && import.meta.url === pathToFileURL(processArgv[1]).href) {
+  main();
+}
--- a/scripts/check-workspace-versions.mjs
+++ b/scripts/check-workspace-versions.mjs
@@ -3,34 +3,39 @@
 // version. Keeps the monorepo's lockstep versioning from silently drifting when
 // a release bump misses a package.
 //
-// The root server.json (the MCP registry manifest) is held to that same version
-// and to the npm coordinates in packages/argent/package.json. It is checked here
-// because it is the one such file outside packages/*, so a bump that sweeps the
-// workspace leaves it behind — and the registry then rejects the publish, either
-// because `packages[].version` names a version npm has never seen or because
-// `name` doesn't match the tarball's `mcpName`.
-import { readFileSync, readdirSync } from "node:fs";
-import { dirname, join } from "node:path";
-import { fileURLToPath, pathToFileURL } from "node:url";
+// The root server.json (the MCP registry manifest) is held to the version and
+// the npm coordinates in packages/argent/package.json, the package it points at.
+// It is checked here because it is the one such file outside packages/*, so a
+// bump that sweeps the workspace leaves it behind — and the registry then
+// rejects the publish, either because `packages[].version` names a version npm
+// has never seen or because `name` doesn't match the tarball's `mcpName`.
+import { readFileSync, readdirSync, realpathSync } from "node:fs";
+import { dirname, join, relative } from "node:path";
+import { fileURLToPath } from "node:url";
 import { argv as processArgv } from "node:process";
 
-const repoRoot = join(dirname(fileURLToPath(import.meta.url)), "..");
+const selfPath = realpathSync(fileURLToPath(import.meta.url));
+const repoRoot = join(dirname(selfPath), "..");
 const packagesDir = join(repoRoot, "packages");
+const argentManifestPath = join(packagesDir, "argent", "package.json");
+const serverJsonPath = join(repoRoot, "server.json");
 
 /**
- * Everything server.json has to keep in step with the workspace, as one
- * human-readable line per problem. Pure, separated from the fs reads so it can
- * be unit-tested.
- * @param {string | undefined} workspaceVersion the single version every packages/* manifest carries
+ * Everything server.json has to keep in step with the npm package it points at,
+ * as one human-readable line per problem. Pure, separated from the fs reads so
+ * it can be unit-tested.
+ * @param {string | undefined} publishedVersion the version packages/argent/package.json carries
  * @param {{ name?: string, mcpName?: string }} argentPkg packages/argent/package.json
  * @param {{ name?: string, version?: string, packages?: { identifier?: string, version?: string }[] }} server server.json
  * @returns {string[]} empty when in sync
  */
-export function serverJsonMismatches(workspaceVersion, argentPkg, server) {
+export function serverJsonMismatches(publishedVersion, argentPkg, server) {
   const mismatches = [];
 
-  if (server.version !== workspaceVersion) {
-    mismatches.push(`server.json version is ${server.version}, workspace is ${workspaceVersion}`);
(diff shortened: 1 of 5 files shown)


scripts/check-workspace-versions.mjs
```diff
@@ -2,38 +2,121 @@
...
+  // An empty packages/ can't reach here: reading packages/argent/package.json above
+  // throws first, so workspaceVersion is always a real version by this point.
+  console.log(`All workspace packages and server.json are at ${workspaceVersion}.`);
 }
 
-console.error("Workspace package versions are out of sync:");
-for (const [version, names] of [...byVersion].sort()) {
-  console.error(`  ${version}: ${names.sort().join(", ")}`);
+// Run only when invoked directly, not when imported by the test.
+if (processArgv[1] && import.meta.url === pathToFileURL(processArgv[1]).href) {
```
hubgan: **LOW** — This guard silently disables the whole check when `argv[1]` is an absolute path that traverses a symlink. Node realpaths the ESM main module before setting `import.meta.url`, but leaves `process.argv[1]` exactly as typed, so the two strings differ, `main()` never runs, and the process exits **0 with no output at all** — indistinguishable from a pass.

Reproduced with genuine drift present (`packages/registry` bumped to `9.9.9`):

```console
$ node scripts/check-workspace-versions.mjs          # relative
Workspace package versions are out of sync:
  0.18.0: @argent/archive, ... @swmansion/argent
  9.9.9: @argent/registry
exit=1

$ ln -sfn "$PWD" /tmp/argent-alias
$ node /tmp/argent-alias/scripts/check-workspace-versions.mjs
exit=0          # no output whatsoever
```

It is a regression: `main`'s version ran at import time, and through the same symlink with the same drift it still exits 1.

**Why this is LOW and not higher** — I went looking for a way it could actually bite and could not find one. `process.cwd()` is always the *physical* path, so a relative `argv[1]` is structurally immune, and both shipped call sites use the relative form (`package.json:26`, `repo-hygiene.yml:42`). Confirmed from a symlinked cwd: `node scripts/check-workspace-versions.mjs`, `npm run check:versions` and `node --run check:versions` all still exit 1. And `main()` demonstrably runs on the real runner — this PR's own Static checks job logged `All workspace packages and server.json are at 0.18.0.` Repo paths containing spaces, `#`, `+`, `&` and non-ASCII were all fine; only symlinks do it.

Still worth a one-line hardening since the failure mode is invisible. This fix is tested — it closes the symlink hole while keeping the relative path working and the 9 tests importing cleanly:

```js
import { realpathSync } from "node:fs";
// ...
if (processArgv[1] && import.meta.url === pathToFileURL(realpathSync(processArgv[1])).href) {
```

(The same pattern in `extract-tools.mjs`, `next-canary-version.mjs` and `sync-next-dist-tag.mjs` has the same flaw, but those are producers — a skipped `main` there yields an empty string and fails loudly. This is the only one where the no-op looks like success.)

scripts/check-workspace-versions.test.mjs
```diff
@@ -0,0 +1,111 @@
...
+    ARGENT_PKG,
+    server((s) => {
+      s.version = "0.17.0";
+      s.name = "com.swmansion/argent";
+      s.packages[0].version = "0.17.0";
+      s.packages[0].identifier = "argent";
+    })
+  );
+  assert.equal(problems.length, 4);
+});
```
hubgan, line 220: **LOW** — The suite covers the pure `serverJsonMismatches()` well (8 of 16 mutants killed, including removing either per-entry check, swapping the message operands, and comparing against `server.version` instead of the workspace version). But it never touches `main()`, which is the part CI actually runs — so a fail-open typo there passes every check in the repo.

Exhibit, with real `server.json` drift present and a single character changed in the `server.json` mismatch branch (`process.exit(1)` → `process.exit(0)`):

```console
$ node scripts/check-workspace-versions.mjs
server.json is out of sync with the workspace:
  server.json version is 0.17.0, workspace is 0.18.0
  server.json packages[0].version is 0.17.0, workspace is 0.18.0
SCRIPT exit=0        <-- CI Static checks goes GREEN on a drifted tree

$ node --test scripts/check-workspace-versions.test.mjs   ->  9 pass, 0 fail
$ npx eslint scripts/check-workspace-versions.mjs --max-warnings 0   ->  exit=0
$ npx tsc --noEmit -p tsconfig.scripts.json                          ->  exit=0
```

The script correctly diagnoses the problem and then reports success, and nothing anywhere catches it. One test that spawns the real script against a fixture tree and asserts on exit code would pin this — and would also have caught the symlink issue above.

**Second, related gap (also LOW):** every fixture has exactly one `packages[]` entry, so `i` is only ever `0` and the loop is never exercised. Three mutants survive as a result — restricting validation to `packages.slice(0, 1)`, and hardcoding the index in either message. With the `slice(0, 1)` mutant, a `server.json` whose second entry names `@attacker/pkg@6.6.6` returns `[]` and CI passes.

To be clear, **the shipped code is correct here** — I ran the real function with that two-entry input and it reports both problems with the right index:

```console
[ "server.json packages[1].version is 6.6.6, workspace is 0.18.0",
  "server.json packages[1].identifier (@attacker/pkg) != published package name (@swmansion/argent)" ]
```

It is purely untested, and becomes live the moment a second distribution channel is added.

hubgan (commented):
Reviewed with a swarm of agents under a hard-reproduction bar: a finding only counted if it was driven for real — real script runs against real mutated files, real npm/registry/GitHub API calls, and real MCP tool calls against a booted iOS Simulator. Code-reading arguments, mocks and stubs were rejected. Everything unreproducible was discarded, and three claims that came back from the swarm were dropped or downgraded after I re-ran them myself.

**Verdict: no HIGH, no MEDIUM. Six LOW findings, all in the guard's test coverage and error ergonomics — nothing wrong with what the PR actually ships.** The manifest is correct and the entrypoint it declares genuinely works.

### Verified working (all executed, not read)

- **`server.json` is schema-valid** against the live pinned `2025-12-11` schema (fetched, HTTP 200, 22090 bytes; ajv + ajv-formats, zero errors). `description` is 89 chars vs `maxLength: 100`.
- **External facts all check out**: `repository.id` `1164885639` matches `api.github.com/repos/software-mansion/argent`; `websiteUrl` returns 200; the `io.github.software-mansion` namespace is free on the live registry (`count: 0`).
- **The declared entrypoint works end-to-end on a real simulator.** Booted a dedicated iPhone 17 Pro / iOS 26.4, started the server exactly as the manifest declares, and drove real apps: `launch-app com.apple.Preferences` → `describe` → `gesture-tap` at coordinates computed from the AX frame → `describe`, with the navigation independently confirmed by `screenshot-diff` (8.02% pixel change, "General" moving, "About" appearing). Also `open-url` into Safari. A 22-step soak (two levels deep, rotate, home, relaunch, `await-ui-element`) ran clean with no crash or stream desync.
- **stdout is pure JSON-RPC in 7 scenarios**, including cold `npx` installs — the telemetry notice, the update-available notice and npm's own deprecation warnings all go to stderr. No strict MCP client gets a corrupted stream. Malformed input on stdin did not desync framing.
- **`mcpName` survives packaging.** Real `npm pack` → the tarball's `package/package.json` carries `"mcpName": "io.github.software-mansion/argent"`. The canary path (`next-canary-version.mjs --write`) round-trips it with key order intact and stays prettier-clean. `server.json` itself is correctly absent from the tarball.
- **`mcpName` is inert at runtime.** A/B'd two worktrees identical except for that key: `tools/list` byte-identical (148572 bytes), normalized run diff 0 lines, installer config generation byte-identical (3973 bytes, zero `mcpName` leakage into generated configs), and the version gate respawns correctly on a real version change while ignoring the added key in both directions.
- **No regressions.** Workspace suite failures were proved pre-existing by A/B against a manifest byte-identical to `main` (identical counts both ways). Full CI parity re-run under Node 20 and 24: `test:scripts` 68/68, lint, prettier, typecheck all clean.
- **The guard's core job works, and self-heals.** A full `0.19.0` bump including `server.json` passes; forgetting `server.json` fails with the exact line. That is the mechanism that makes the registry publish work at the next release.

### Dropped after re-running (recording these so they don't get re-raised)

- Two agents rated **"the registry publish will be rejected"** MEDIUM/HIGH. The npm facts are real — `0.18.0` is `latest` and its published tarball has no `mcpName` — but the PR explicitly scopes publishing out and states this precondition, and I verified the design self-heals at the next release. Not a defect; only the stale `0.17.0` reference in the description remains (noted inline).
- An `identifier`-vs-`registryType` concern was dropped outright: it required inventing an `mcpb` entry that does not exist.
- My own first exhibit for the coverage gap (`if (false)`) was **wrong** — the repo's eslint catches it via `no-constant-condition`. Replaced with a mutant that survives every check (inline below).

### Also LOW, not worth inline comments

- A half-finished release bump costs two CI round-trips: the `packages/*` check `process.exit(1)`s before `server.json` is ever read, so you fix the packages, push, and only then learn `server.json` also drifted. Verified by running both rounds.
- The new script is *stricter* than `main` on a malformed `packages/argent/package.json` — `main` silently `continue`d past it, HEAD exits 1. Fail-closed rather than fail-open, which reads as deliberate.

Nice PR — the verification write-up in the description held up on every point I re-tested except the stale version reference.

scripts/check-workspace-versions.mjs
```diff
@@ -2,38 +2,121 @@
...
+    console.error(
+      "\nEvery package under packages/* must share one version. Bump the outliers to match."
+    );
+    process.exit(1);
+  }
+
+  const [workspaceVersion] = byVersion.keys();
+
+  const argentPkg = JSON.parse(readFileSync(join(packagesDir, "argent", "package.json"), "utf8"));
+  const server = JSON.parse(readFileSync(join(repoRoot, "server.json"), "utf8"));
```
hubgan: **LOW** (ergonomics only) — A missing or malformed `server.json` escapes as a raw Node stack trace instead of the carefully-written guidance a few lines below. All fail-closed at exit 1, so nothing ships wrong; the cost is a stack dump in the CI log instead of a sentence.

```console
$ rm server.json && node scripts/check-workspace-versions.mjs
Error: ENOENT: no such file or directory, open '.../server.json'
exit=1

$ echo '{ not json' > server.json && node scripts/check-workspace-versions.mjs
SyntaxError: Expected property name or '}' in JSON at position 2
exit=1

$ # packages as an object instead of an array
TypeError: packages.entries is not a function ...   (serverJsonMismatches, line 55)
exit=1
```

Only reachable by hand-corrupting a tracked file, so this is a nit — but wrapping the read in the same try/catch style used for the `packages/*` manifests would keep the failure on-message.

server.json
```diff
@@ -0,0 +1,31 @@
...
+    "source": "github",
+    "id": "1164885639"
+  },
+  "websiteUrl": "https://argent.swmansion.com",
+  "version": "0.18.0",
+  "packages": [
+    {
+      "registryType": "npm",
+      "identifier": "@swmansion/argent",
+      "version": "0.18.0",
```
hubgan, line 17: **LOW / informational** — no change requested here; flagging one stale detail in the PR description.

The description says publishing "can only succeed after a release whose npm tarball carries `mcpName` — i.e. not against the currently published `0.17.0`". `0.17.0` is no longer current: `0.18.0` was published `2026-07-30T22:49:46Z` and is now `latest`, and it *also* predates `mcpName`.

```console
$ curl -sS https://registry.npmjs.org/@swmansion%2Fargent/0.18.0 | jq '{version, mcpName}'
{ "version": "0.18.0", "mcpName": null }

$ npm view @swmansion/argent dist-tags
{ latest: '0.18.0', next: '0.17.1-next.5' }
```

The conclusion is unchanged and the sequencing still works — I verified the whole chain rather than assuming it. A full `0.19.0` bump including this file passes the guard; forgetting this file fails it with the exact line; and a real `npm pack` at `0.19.0` produces a tarball carrying `"mcpName": "io.github.software-mansion/argent"`. So the publish succeeds once `0.19.0` is on npm.

Worth correcting the version in the description so whoever runs `mcp-publisher publish` doesn't read it as "0.18.0 is the one that works" — that attempt would fail with a confusing *"missing required `mcpName` field"* pointing at a `package.json` that visibly contains the field.

scripts/check-workspace-versions.mjs
latekvo: Fixed in 49c0b1cc65e489bacad86c8efcbbc142d5c38257.

Realpath'd **both** sides rather than just `argv[1]`, so the inverse hole closes too — under `--preserve-symlinks-main` `import.meta.url` is the symlinked path and realpathing only `argv[1]` would skip `main()` the other way round:

```js
const selfPath = realpathSync(fileURLToPath(import.meta.url));
// ...
if (processArgv[1] && realpathSync(processArgv[1]) === selfPath) main();
```

Your repro, re-run on the branch with `packages/registry` at `9.9.9` and `server.json` at `0.17.0`, through `/tmp/argent-alias` → exit 1 with both problems printed, where it previously exited 0 silently. Pinned by `[script] a run through a symlinked path still catches drift`, which is the only test that dies when the guard is reverted to the raw compare.

Left `extract-tools.mjs`, `next-canary-version.mjs` and `sync-next-dist-tag.mjs` alone, for the reason you gave: they are producers whose skipped `main` yields empty output and fails loudly downstream, so the pattern there is not a silent pass.

scripts/check-workspace-versions.test.mjs
latekvo: Fixed in 49c0b1cc65e489bacad86c8efcbbc142d5c38257.

Added a spawned layer that runs the real script against a throwaway repo (a copy of the script, two `packages/*` manifests, a `server.json`) and asserts on exit codes — six tests covering the in-sync pass, `server.json` drift alone, a half-finished bump, the symlinked invocation, and the two malformed-file paths.

Your `exit(1)` → `exit(0)` mutant is now killed by three of them. Ran the whole set through a mutation harness — 10 mutants, 0 survivors, each killed only by the test that targets it:

```
guard-raw-compare (pre-fix symlink hole)   KILLED by: [script] a run through a symlinked path still catches drift
fail-open exit(0)                          KILLED by: [script] server.json drift alone exits 1; +2
packages loop slice(0,1)                   KILLED by: every packages[] entry is validated, not just the first
hardcode index in version msg              KILLED by: every packages[] entry is validated, not just the first
hardcode index in identifier msg           KILLED by: every packages[] entry is validated, not just the first
...
SURVIVORS: none
```

The second gap is covered by a two-entry fixture asserting the exact pair of lines your run produced, so `slice(0, 1)` and both hardcoded-`[0]` mutants now fail. `test:scripts` is 75/75.

scripts/check-workspace-versions.mjs
latekvo: Fixed in 49c0b1cc65e489bacad86c8efcbbc142d5c38257.

`packages/argent/package.json` and `server.json` now go through a `readTrackedJson` helper that reports both the missing and the unparseable case as a sentence:

```console
$ rm server.json && node scripts/check-workspace-versions.mjs
Cannot read server.json: ENOENT: no such file or directory, open '.../server.json'

It is tracked in git — restore it with \`git checkout -- server.json\`.
exit=1

$ echo '{ not json' > server.json && node scripts/check-workspace-versions.mjs
server.json is not valid JSON: Expected property name or '}' in JSON at position 2

It is tracked in git — restore it with \`git checkout -- server.json\`.
exit=1
```

The third case you found was not a read failure — an object-shaped `packages` reached `packages.entries()` inside the pure function. That is now folded into the existing "nothing to install" mismatch (`!Array.isArray(packages) || packages.length === 0`) rather than escaping as a TypeError. Both spawned tests assert `doesNotMatch(stderr, /^\s+at /m)`, so a regression to a raw stack fails rather than just reading worse.

server.json
latekvo: Fixed in 49c0b1cc65e489bacad86c8efcbbc142d5c38257 (description only — no code change).

Corrected to `0.18.0`, and re-verified rather than taking the numbers across:

```console
$ npm view @swmansion/argent dist-tags
{ next: '0.18.1-next.0', latest: '0.18.0' }

$ curl -sS https://registry.npmjs.org/@swmansion%2Fargent/0.18.0 | jq '{version, mcpName}'
{ "version": "0.18.0", "mcpName": null }
```

Also refreshed the entrypoint transcript in the same section, which was a real run but against a `@latest` that has since moved — it now answers `"version":"0.18.0"` with stderr empty.

scripts/check-workspace-versions.mjs
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Confirmed and fixed in 49c0b1cc.

Reproduced independently against `5a23b061` before changing anything, with genuine drift (`packages/registry` at `9.9.9`):

```console
$ node scripts/check-workspace-versions.mjs                      # relative
Workspace package versions are out of sync:
  0.18.0: @argent/archive, ... @swmansion/argent
  9.9.9: @argent/registry
exit=1

$ ln -sfn "$PWD" ../.pr598-alias
$ node /home/user/dev/argent-worktrees/.pr598-alias/scripts/check-workspace-versions.mjs
exit=0          # no output whatsoever
```

Same tree, same symlink, same drift at 49c0b1cc → `exit=1` with both lines printed.

The fix goes slightly further than the suggestion: it realpaths *both* sides (`selfPath` is the realpath of `import.meta.url`, compared against `realpathSync(processArgv[1])`) rather than only `argv[1]`. That also holds under `--preserve-symlinks`, where `import.meta.url` is itself the symlinked spelling — verified, since realpathing only `argv[1]` would reintroduce the mismatch in the opposite direction:

```console
$ node --preserve-symlinks --preserve-symlinks-main /…/.pr598-alias/scripts/check-workspace-versions.mjs
Workspace package versions are out of sync:
  ...
exit=1
```

It is pinned by `[script] a run through a symlinked path still catches drift`: reverting the guard to the raw `import.meta.url === pathToFileURL(argv[1]).href` compare fails that test and only that test.

scripts/check-workspace-versions.test.mjs
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Both gaps confirmed and fixed in 49c0b1cc.

**`main()` untested.** Reproduced your exhibit on `5a23b061` — real `server.json` drift plus the one-character fail-open — and it behaved exactly as you described: the script printed the full diagnosis and exited 0, with all 9 tests, `eslint` and `tsc -p tsconfig.scripts.json` green.

Seven spawned-script tests now run the real thing against a throwaway repo (a copy of the script, two `packages/*` manifests and a `server.json`) and assert on exit codes. The equivalent mutant at HEAD is killed three times over:

```console
$ # `if (failed) process.exit(1)` -> `process.exit(0)`
$ node --test scripts/check-workspace-versions.test.mjs
✖ [script] server.json drift alone exits 1
✖ [script] a half-finished bump reports packages/* and server.json in one run
✖ [script] a run through a symlinked path still catches drift
ℹ pass 15  ℹ fail 3
```

**The `packages[]` loop only ever seeing index 0.** Also fixed — `every packages[] entry is validated, not just the first` pushes a second entry and asserts on the full message list, so it pins the index as well as the iteration. Both mutants you named are now killed by it and nothing else:

```console
$ # packages.slice(0, 1)
✖ every packages[] entry is validated, not just the first     ℹ pass 17  ℹ fail 1

$ # `packages[${i}].identifier` -> `packages[0].identifier`
✖ every packages[] entry is validated, not just the first     ℹ pass 17  ℹ fail 1
```

Worth noting your two-entry probe is now the literal fixture: the test asserts exactly the pair of lines you got back by hand, `@attacker/pkg@6.6.6` included.

scripts/check-workspace-versions.mjs
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Confirmed, fixed in 49c0b1cc — and re-running your three cases turned up a fourth that the fix missed, now fixed in 47e7e72b.

All three of yours reproduce on `5a23b061` exactly as posted, and all three are on-message at HEAD:

```console
$ rm server.json && node scripts/check-workspace-versions.mjs
Cannot read server.json: ENOENT: no such file or directory, open '…/server.json'

It is tracked in git — restore it with `git checkout -- server.json`.
exit=1

$ echo '{ not json' > server.json && node scripts/check-workspace-versions.mjs
server.json is not valid JSON: Expected property name or '}' in JSON at position 2 (line 1 column 3)
…
exit=1

$ # "packages" as an object
server.json is out of sync with packages/argent/package.json:
  server.json has no packages array — the registry entry would name nothing to install
exit=1
```

**The one that was still open.** The array check guards the *container*, not the entries, so a hand-edit that empties an entry rather than the array walked straight into the loop:

```console
$ echo '{"name":"io.github.software-mansion/argent","version":"0.18.0","packages":[null]}' > server.json
$ node scripts/check-workspace-versions.mjs
file:///…/scripts/check-workspace-versions.mjs:65
    if (pkg.version !== publishedVersion) {
            ^
TypeError: Cannot read properties of null (reading 'version')
    at serverJsonMismatches (…:65:13)
exit=1
```

That is the same escape the array check exists to prevent, and the comment above it ("so they get one message rather than a TypeError out of the loop") reads as covering it. 47e7e72b reports the entry's type and continues:

```console
$ node scripts/check-workspace-versions.mjs      # packages: [null]
  server.json packages[0] is null, not an object naming a registry and identifier
exit=1

$ node scripts/check-workspace-versions.mjs      # packages: ["@swmansion/argent"]
  server.json packages[0] is string, not an object naming a registry and identifier
exit=1
```

Pinned by `a packages[] entry that is not an object is caught`, which fails when the guard is removed and passes otherwise. `test:scripts` 77/77, `eslint`, `tsc` and `prettier --check` clean.

server.json
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Confirmed, and the description now names `0.18.0`.

Re-checked against live npm rather than taking the numbers over — they had moved again since your comment, `next` included:

```console
$ npm view @swmansion/argent dist-tags
{ next: '0.18.1-next.0', latest: '0.18.0' }

$ curl -sS https://registry.npmjs.org/@swmansion%2Fargent/0.18.0 | jq .mcpName
null
$ curl -sS https://registry.npmjs.org/@swmansion%2Fargent/0.17.0 | jq .mcpName
null

# publish times
0.17.0: 2026-07-23T22:23:07.221Z
0.18.0: 2026-07-30T22:49:45.770Z
```

So the conclusion holds and the sequencing is unchanged — the first publishable release is the next bump, whatever its number.

While re-verifying I also re-ran the entrypoint claim in the same section, since `@latest` now resolves to a different tarball than when it was written. It still answers, and stdout is still pure JSON-RPC with npm's notices on stderr:

```console
$ printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"initialize",…}' | npx -y @swmansion/argent@latest mcp
{"result":{"protocolVersion":"2024-11-05","capabilities":{"tools":{}},
 "serverInfo":{"name":"argent","version":"0.18.0"},"instructions":"…"},"jsonrpc":"2.0","id":1}
```

The other cross-file assertions in that section still hold at HEAD: `repository.id` `1164885639` matches `api.github.com/repos/software-mansion/argent`, `description` is 89 chars against the schema's `maxLength: 100`, and `cli.ts:137` is still the `case "mcp":` dispatcher.

hubgan (approved):

latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat)\]: Review continued past the merge; recording what landed here and what did not.

**All four inline findings are resolved and re-verified against the merged tree** — each was reproduced independently first, then re-run after the fix (details on the individual threads). The last of them turned up a fourth case the fix had missed (`"packages": [null]` still throwing), fixed in 47e7e72b.

Three further defects were found and fixed before the merge, all of the same shape — a comparison that passed *because* a field was missing:

- `packages/argent/package.json` with no `version` printed `All workspace packages and server.json are at undefined.` and exited **0**. This was a regression from 5a23b061, which caught it: switching the comparison target from the workspace version to `argentPkg.version` removed the only thing that noticed. Fixed in e467bc9c.
- With no `name` on the manifest and no `identifier` on the entry, `pkg.identifier !== argentPkg.name` is `undefined !== undefined`, so a `server.json` naming nothing installable was reported in sync. Fixed in 32fc4c1f.
- A `server.json` holding `null` reached the field reads and threw; `"0.18.0"` and `[]` produced mismatch lines that never said the file was the wrong shape. Fixed in e09ce2be.

And the guard's *original* job was pinned by nothing: every `[script]` test reaching a non-zero exit did so through a `server.json` problem, so deleting `failed = true` from the `packages/*` branch left the whole suite, `eslint` and `tsc` green while a really drifted workspace printed its full report and exited **0**. Also fixed in e09ce2be, along with three more unexercised properties (stepping over a manifest-less `packages/*` directory — `packages/argent-private` is exactly that; grouping every package sharing a version onto one line; continuing past a bad `packages[]` entry).

**Not in this PR — opened as #608**, since it merged while these were still being verified:

- The scan's `catch { continue; }` swallows *every* read/parse failure, so a manifest that is present but unparseable leaves the lockstep comparison and the run affirms a version it never checked. A BOM is the reachable form: `prettier --check` reports the file clean, `npm install` parses it, and `repo-hygiene.yml` runs the guard on a bare checkout with no `npm ci`. Pre-existing on the merge base, but `readTrackedJson` now hardens two tracked JSON files while the other sixteen manifests keep the silent skip.
- Three mismatch lines report a problem in `packages/argent/package.json`, but the trailer tells the reader to edit `server.json` to match it — the opposite of the fix.
- `typeof [] === "object"`, so an array `packages[]` entry bypasses the shape guard.

**Description corrected** on two counts that no longer held: `test:scripts` is 86, not 85, and the mutation survey is 23 mutants with 2 survivors (the readdir sort and the `!entry.isDirectory()` skip, both scan-order properties no single run can observe) rather than 22 with 1. The `trent-j` attribution is softened — the quote and its date are exact, but GitHub's API reports `authorAssociation: NONE` for them on `github/github-mcp-server`, so "a collaborator on that repo" is not supported.

**Left alone, named rather than fixed:** the same entrypoint-guard idiom in `extract-tools.mjs`, `next-canary-version.mjs` and `sync-next-dist-tag.mjs`. All shipped call sites are relative, so the trigger is dormant — but the claim that those three fail loudly does not hold for `sync-next-dist-tag.mjs`: it is the terminal step of its own workflow, nothing re-reads the dist-tag, so a skipped `main()` there shows a green "Sync 'next' dist-tag" run in which `next` was never re-pointed. Verified by running it against a fake `npm` on `PATH` (the real registry was never contacted). Out of scope for a `server.json` PR; worth its own change.

PR #605, opened 2026-07-31
fix(chromium): make Electron apps testable without window focus

filip131311:
## Problem

Field report: *"electron needs focus to be tested."* Analysis confirmed three focus/visibility dependencies in the Chromium/Electron surface:

1. **Hidden window → throttled renderer.** A minimized/occluded window's compositor throttles: mouse-input acks stall **~5 s per event** (they wait on hit-testing), wheel scrolls hang (`gesture-scroll` already refused), `visibilityState` flips to `hidden`.
2. **Unfocused window → focus-gated app logic breaks.** `document.hasFocus()` is `false` before any input arrives; focus/blur-dependent UI misbehaves.
3. Focus-gated web APIs (clipboard already worked around it locally with `Page.bringToFront`).

## Fixes

**1. CDP focus emulation on every page session** (`primePageSession`, né `enableCoreDomains`): `Emulation.setFocusEmulationEnabled(true)`, best-effort, re-applied automatically after tab switches (Playwright precedent). Live A/B on the Electron testbed: identical boot→unfocus→probe sequence gives `hasFocus:false` on main, `true` on this branch. Bonus verified by A/B (kill server → same minimized window flips to `hidden`): in current Chromium/Electron, emulation also pins reported visibility to `visible` and keeps input unthrottled while genuinely minimized — minimized tap latency **~5 s → ~65 ms**, wheel scroll works.
   *Trade-off (same as Playwright ships):* while attached, the app can observe neither real blur nor real hidden, so pause-on-hidden behavior is unobservable during a test.

**2. Anti-throttling switches for every argent-booted Electron app** (`--disable-background-timer-throttling --disable-backgrounding-occluded-windows --disable-renderer-backgrounding`). Sessions are created lazily and die with the idle-exiting tool-server while the app outlives it — the flags cover that detached lifecycle plus emulation-less runtimes. Also covers flow-booted apps via the shared `bootElectronApp`.

**3. Hidden-window fail-fast extended from gesture-scroll to gesture-tap + gesture-drag** via shared `assertChromiumWindowVisible`, under a new dedicated `CHROMIUM_WINDOW_HIDDEN` failure code (scroll migrated; `failure_stage` strings unchanged so telemetry stays joinable — analytics keyed on `CHROMIUM_INPUT_INVALID` + `chromium_scroll_window_hidden` should switch to the new code). With emulation active the guard is a backstop that fires only where emulation couldn't be applied — exactly where the stall is real. Keyboard/button stay unguarded on purpose: measured live, key events skip hit-testing and stay fast on hidden windows.

## Verification

- Unit: 3075 tool-server + 87 registry tests green, `typecheck:tests` green, new tests for emulation priming, boot args, and the guard (all three tools).
- Live matrix on the Electron testbed (dev tool-server): focused baseline; unfocused (`hasFocus` true pre-input, typing lands); minimized booted app (tap 69 ms, scroll works, typing + slider drag land, screenshot fine); minimized **external** un-flagged app on :9222 (tap 63 ms, scroll 460 ms — emulation alone suffices while attached).
- Plan + implementation went through three adversarial review rounds (scope, CDP semantics, failure-code taxonomy/telemetry compat, doc accuracy).


packages/tool-server/test/chromium-window-hidden-guard.test.ts
```diff
@@ -0,0 +1,108 @@
...
+      )
+    );
+    expect(getFailureSignal(err)?.error_code).toBe(FAILURE_CODES.CHROMIUM_WINDOW_HIDDEN);
+    // The stage string is unchanged so old and new telemetry stay joinable.
+    expect(getFailureSignal(err)?.failure_stage).toBe("chromium_scroll_window_hidden");
+  });
+
+  it("proceeds when the visibility probe rejects (mid-navigation, bare test fakes)", async () => {
+    // A failed probe proves nothing about visibility; only an explicit
+    // "hidden" refuses. This also covers api objects with no `cdp` at all.
```
latekvo, line 94: This comment says the test "also covers api objects with no `cdp` at all", but it doesn't. The test starts from `fakeChromiumApi()` (which *does* define `cdp`) and replaces `cdp.send` with a rejecting mock, so it only exercises the rejected-promise path. An api with no `cdp` fails through a different mechanism: a TypeError raised while reading `.send` off `undefined`, not a rejected send.

That path is currently exercised only incidentally, by the fakes in `chromium-drag.test.ts:15` and `tools/gesture-tap.test.ts:41`, neither of which defines `cdp`. If either fake ever gains a `cdp` key, the no-`cdp` path silently loses its only coverage while this comment still claims it is covered here.

Checked by reading both sibling fakes and running the four chromium gesture test files against the merge commit.

packages/tool-server/src/utils/chromium-visibility.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Key events skip hit-testing and stay fast on hidden windows, so the
+ * keyboard/button tools deliberately do NOT use this guard. It is a backstop:
+ * while a session is attached, primePageSession's focus emulation pins the
+ * renderer's reported visibility to "visible" and keeps input fast even for a
+ * genuinely minimized window, so the probe only ever reads "hidden" on
+ * sessions where emulation could not be applied (unsupported runtime, failed
+ * priming) — exactly the world where the stall is real and un-minimizing is
+ * the fix.
+ *
+ * The entire probe (including the property access on `chromium`) stays inside
```
latekvo, line 23: The stated rationale doesn't match the mechanism it describes. For an api that lacks `cdp`, `chromium.cdp` evaluates to `undefined` without throwing; the TypeError comes from reading `.send` off that `undefined`, and that read is inside the try regardless of where `chromium.cdp` itself is evaluated.

Verified rather than reasoned: hoisting `const client = chromium.cdp;` above the try and calling `client.send(...)` inside it leaves all 16 tests across chromium-window-hidden-guard / chromium-drag / chromium-scroll / tools-gesture-tap green.

The placement would only matter if `chromium` itself were null or undefined, which is not the case the comment names and not something any registry path produces - the blueprint always sets `api.cdp = server.cdp`.

packages/tool-server/src/utils/chromium-visibility.ts
```diff
@@ -0,0 +1,53 @@
...
+ * the try: test fakes and mid-navigation contexts may lack `cdp`, and neither
+ * says anything about visibility.
+ */
+export async function assertChromiumWindowVisible(
```
latekvo, line 27: `action` and `failureStage` are adjacent parameters of the same type, and every call site passes a strictly redundant pair: `("tap", "chromium_tap_window_hidden")`, `("drag", "chromium_drag_window_hidden")`, `("scroll", "chromium_scroll_window_hidden")`. The stage is always `chromium_<action>_window_hidden`.

Since both are plain `string`, transposing the two arguments or mistyping the stage compiles cleanly and reads fine in review; the only symptom is telemetry landing in a bucket nobody queries.

That carries more weight here than it normally would: this change deliberately froze `chromium_scroll_window_hidden` so pre- and post-migration rows stay joinable, so a silent drift in one of these hand-written strings breaks precisely the property the failure-code migration was designed to preserve.

packages/tool-server/src/tools/gesture-scroll/index.ts
```diff
@@ -119,7 +84,7 @@ Returns { scrolled: true, timestampMs }. Fails if the Chromium CDP session is no
   async execute(services, params) {
     const timestampMs = Date.now();
     const chromium = services.chromium as ChromiumCdpApi;
-    await assertWindowVisible(chromium);
+    await assertChromiumWindowVisible(chromium, "scroll", "chromium_scroll_window_hidden");
```
latekvo, line 87: With scroll, tap and drag now refusing through this shared helper, the comment at `packages/tool-server/src/tools/flows/flow-run.ts:517` no longer describes the code: it still tells the reader "gesture-scroll fails fast on that case itself", naming scroll as the only tool with that behaviour. A reader arriving at `frontChromiumPage` will under-estimate which flow steps can now refuse up front.

The rest of that comment does hold. I checked the `Page.bringToFront` claim on an Electron testbed: against a genuinely minimized window it flips `document.hasFocus()` to true but leaves `document.visibilityState` at `hidden`, so it really cannot unhide a minimized window. Only the tool list is stale.

packages/tool-server/src/utils/chromium-visibility.ts
```diff
@@ -0,0 +1,53 @@
...
+ * A hidden window (minimized, fully occluded, or on another workspace) halts
+ * the renderer's mouse-input pipeline: every mouse dispatch waits on
+ * compositor hit-testing, which a throttled window services at ~5s per event
+ * — a drag's ~60fps interpolated moves would stall for minutes, and an agent
+ * loop would keep "working" blind because describe stays fast. Probe
+ * `document.visibilityState` and refuse up front with a fix the caller can
+ * act on. Only an explicit "hidden" refuses — a failed or empty read proves
+ * nothing, and the gesture itself will surface a real transport error.
+ *
+ * Key events skip hit-testing and stay fast on hidden windows, so the
```
latekvo, line 14: This exemption is correct, but nothing enforces it.

I measured it against a genuinely minimized Electron window with the renderer throttled: `Input.dispatchKeyEvent` returned in 1-14 ms, while ten consecutive `Input.dispatchMouseEvent` moves cost 5002-5005 ms each. So keeping keyboard/button out of the guard is the right call, and the reasoning in this comment is sound.

The problem is that the invariant lives only in prose. Adding the guard to `keyboard` or `button` would make them refuse input that demonstrably works on a hidden window, and the suite would stay green. I mutation-tested this against the merge commit (13 mutations): the guard's *presence* on tap/drag/scroll is well pinned - removing it from any of the three, inverting the condition, neutering the comparison, reverting the failure code, or reordering guard-vs-dispatch were all caught - but its deliberate *absence* on keyboard/button is not pinned at all.

latekvo (commented):
Post-merge review. Findings verified against the merge commit (11f384d3): I reproduced the premise on an Electron testbed (control, minimized: `visibilityState: hidden`, mouse-move dispatch 5002-5005 ms x10, timers throttled to 1 Hz), A/B'd both fixes, and mutation-tested the new suite. The three fixes hold up - focus emulation restores `hasFocus`/`visibility` and un-throttles input whether primed before or after the window is minimized, it survives a cross-process navigation, and the anti-throttling switches independently keep a booted app responsive. The comments below are the smaller things left over.

PR #608, opened 2026-07-31
fix(scripts): stop the version guard passing a workspace it failed to read

latekvo:
Follow-up to #598, which merged while these were still being verified. Three defects in the version guard it added, plus the test gaps that let them through.

### An unreadable `packages/*` manifest was silently dropped

The scan's `catch { continue; }` swallowed every `readFileSync` + `JSON.parse` failure, not just a missing file, so a manifest that is *present but unparseable* left the lockstep comparison entirely — and the run then affirmed a version it never checked:

```console
$ node scripts/check-workspace-versions.mjs      # BOM on a drifted packages/registry manifest
All workspace packages and server.json are at 0.18.0.
exit=0
```

Control: identical tree with the BOM removed → `exit=1`, `0.17.0: @argent/registry`.

A BOM is the trigger that matters. A trailing comma is caught upstream (`npm ci` fails `EJSONPARSE`, and `prettier --check .` flags the file), and `chmod 000` is a workstation-only state since `actions/checkout` writes 644. A BOM is invisible to both — `prettier --check` reports the file clean and `npm install` parses it — and `repo-hygiene.yml` runs the guard on a bare checkout with no `npm ci` in that job, so nothing in the pipeline sees it.

Now only `ENOENT` is skipped, which is what the comment (`// directory without a package.json`) ever meant to cover; anything else fails naming the file. This half is pre-existing — the same `catch` sits on the merge base and fails open there identically — but #598 added `readTrackedJson` precisely so a missing/malformed tracked JSON file fails on-message, and applied it to two files while the other sixteen manifests kept the silent skip.

The same fail-open had a second, subtler window: a manifest whose *parse succeeds* but returns a non-object (`null`, a bare scalar, or an array). `JSON.parse` accepts all three, so none of them ever reached the `catch`, and the `manifest?.version` read then skipped the package as if it carried no version — a present-but-unusable manifest still leaving the comparison and affirming a version it never read. That shape is exactly what `readTrackedJson`'s shape check rejects for `server.json` and `packages/argent`, so the scan now rejects it the same way (on-message, naming the file) with tests for the `null` and array forms.

### The failure report told you to edit the wrong file

Three mismatch lines report a problem in `packages/argent/package.json` (no `version`, no `name`, no `mcpName`), but the trailer read "server.json is not under `packages/*` … edit it by hand to match `packages/argent/package.json`" — the opposite of the fix:

```console
$ node scripts/check-workspace-versions.mjs
server.json is out of sync with packages/argent/package.json:
  packages/argent/package.json has no mcpName — the registry reads it off the published tarball to prove the namespace

server.json is not under packages/*, so a workspace bump leaves it behind — edit it
by hand to match packages/argent/package.json.
```

Every mismatch line already begins with the file it concerns, so the trailer now says that instead of naming one file for all of them.

### An array `packages[]` entry bypassed the shape guard

`typeof [] === "object"`, so `["@swmansion/argent", "0.18.0"]` fell through to the field comparisons and came out as two `undefined` mismatches rather than the wrong shape it is.

### Test gaps

Each of the following was mutated and survived the whole suite plus `eslint` and `tsc` before this change, and is killed by exactly the test added for it now:

| mutation | was |
| --- | --- |
| `catch` swallows every manifest read error | 27/27 green |
| `readTrackedJson(argentManifestPath)` → raw `JSON.parse` | 27/27 green |
| drop `names.sort()` in the drift report | 27/27 green |
| `manifest.name ?? entry.name` → always the manifest name (drops the `entry.name` fallback) | 27/27 green |
| version-less manifest joins the version map | 27/27 green |
| either failure report's guidance trailer deleted | 27/27 green |
| drop the array arm of the entry shape guard | 27/27 green |

`readTrackedJson`'s malformed-JSON arm is unreachable from the `packages/argent` side — the scan reads the same file first and stops there — so the two tests added for that call site cover the arms that are reachable (missing, and not-an-object).

The `names.sort()` gap was a fixture artifact: the only two-package fixture used directories `alpha`/`argent` holding `@argent/alpha`/`@swmansion/argent`, already in sorted order, so the sort was a no-op. The fixture now puts `@zzz/scanned-first` in directory `aaa`, so scan order and sorted order disagree.

Same story on the name-fallback row: the `manifest.name ?? entry.name` *fallback* arm (a versioned manifest carrying no `name`) had no fixture at all, so writing it as `manifest.name` (dropping the fallback → prints `0.16.0: undefined`) survived the whole suite. The reverse mutation — `entry.name` only — was never the gap: the merge base's fixture in directory `registry` named `@argent/registry` already asserts the manifest name, so `@argent/registry`-expecting lines fail under it. The new fixture (a versioned manifest with no name) is what pins the fallback arm.

### Verification

- `test:scripts` 91/91, `eslint --max-warnings 0`, `tsc -p tsconfig.scripts.json`, `prettier --check` — all clean on top of `main` at eea6c27c.
- Each repro above re-run against the fix; the BOM case additionally re-run against the merge base to confirm the pre-existing fail-open.


scripts/check-workspace-versions.test.mjs
```diff
@@ -303,21 +318,60 @@ test("[script] server.json drift alone exits 1", (t) => {
...
   assert.match(result.stderr, /Workspace package versions are out of sync/);
   assert.match(result.stderr, /0\.17\.0: @argent\/registry/);
+  assert.match(result.stderr, /Bump the outliers to match\./);
   // server.json agrees with packages/argent, so nothing else could have failed it.
   assert.doesNotMatch(result.stderr, /server\.json is out of sync/);
 });
 
+// The manifest is what names a package in the report; the directory name is only
+// the fallback for one that does not name itself.
+test("[script] a versioned manifest with no name is reported by its directory", (t) => {
```
latekvo, line 341: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The PR body's mutation table claims a coverage gap that four pre-existing assertions already killed.

The PR body justifies this new test with the table row `| manifest.name ?? entry.name -> always the directory name | 27/27 green |`. That mutation could not have been green: the pre-PR test file already contained four assertions that require the manifest's `name`, not the directory name, for the fixture package in directory `registry` named `@argent/registry`. Under `names.push(entry.name)` the report line becomes `  0.17.0: registry`, which fails all four. The test added here is still worth having — it is the first fixture with a versioned manifest carrying no `name`, so it is what pins the `?? entry.name` fallback arm (mutant `names.push(manifest.name)` prints `0.16.0: undefined` and is killed only by this test) — but the mutation the body credits it with was already covered, so the table overstates the pre-PR gap for that row.

**Trigger.** Apply the body's stated mutation (`names.push(entry.name)`) to the merge base eea6c27c and run `node --test scripts/check-workspace-versions.test.mjs`: the three `assert.match(result.stderr, /0\.17\.0: @argent\/registry/)` assertions and the `/@argent\/registry/` one fail, so the suite was not 27/27 green as claimed.

**Checked against.** `git show eea6c27c:scripts/check-workspace-versions.test.mjs | grep -n '0\.17\.0:'` → lines 316, 331, 343: `assert.match(result.stderr, /0\.17\.0: @argent\/registry/);` and line 387: `assert.match(result.stderr, /@argent\/registry/);`, against the fixture at line 264 `name: "@argent/registry"` written into directory `registry`. The report line is built at scripts/check-workspace-versions.mjs:176 `names.push(manifest.name ?? entry.name);`.

latekvo (commented):

scripts/check-workspace-versions.test.mjs
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: Agreed and fixed. Reproduced it first: applying `names.push(entry.name)` (always the directory name) to the merge-base script fails 5 pre-existing tests (lines 316/331/343/387 expecting `@argent/registry`), so that mutation was never the uncovered gap. The row's `27/27 green` claim was wrong. PR body updated: the row now names the mutation this test actually pins — dropping the `?? entry.name` fallback (`manifest.name ?? entry.name` → `manifest.name`, which yields `0.16.0: undefined` and is killed only by the new fixture asserting `/0\.16\.0: unnamed/`) — and I added a note below the table explaining both directions for anyone re-auditing. The test itself is untouched; it was the PR description that overstated the gap.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: Second full pass over the updated branch (commits 3887aaa6c, e1540255b, 042f13dc0) — cross-checked by two independent subagents across the review moves + mutation pass. **No HIGH or MEDIUM finding remains.** The two defects I introduced-and-fixed are verified: (1) a `packages/*/package.json` that parses to a non-object (null/scalar/array) was silently skipped and passed a drifted workspace — now fails on-message (repro before/after, exit 0→1); (2) a mutation dropping the scalar arm of that guard survived the pre-fix suite — the three scan-shape assertions are now a loop covering null/string/array, matching the other two guard sites, and the mutation is killed (31 fail under it). 92/92 `test:scripts`, eslint `--max-warnings 0`, prettier, and `tsc -p tsconfig.scripts.json` all clean.

Two LOW findings to record, one of which this PR introduces:

**1. LOW (introduced by this PR; reporting-completeness regression, still fails closed):** an unreadable `packages/*` manifest now aborts inside the scan loop (`process.exit(1)` at scripts/check-workspace-versions.mjs:158-170) *before* the drift aggregate is printed. On the merge base the same tree printed the `0.17.0: @argent/registry` drift summary *and* skipped the bad manifest (old `catch { continue }`). Reproduced: `packages/a_broken/package.json` = `{ oops` + `packages/registry` at `0.17.0` → today only "Cannot read packages/a_broken/package.json: … ", the drift report is hidden; merge base printed both. Still exits 1, so CI stays red — a dev fixes the manifest and reruns and *then* sees the drift. This slightly undercuts the PR's stated "surface every remaining edit in one run" goal and its own comment (":195 "so a corrupt server.json can't swallow it"). Worth a follow-up (collect the manifest error and report drift before exiting), but because it fails closed it's LOW, not a blocker for this PR.

**2. LOW (pre-existing, out of scope):** a symlinked `packages/*` directory is silently excluded (`isDirectory()` is false for a symlink at scripts/check-workspace-versions.mjs:154), so a drifted real package reachable only via a symlink under `packages/` is skipped and the run passes. Reproduced on both merge base and HEAD. Identical on the merge base, pre-dates this PR, and needs someone to deliberately symlink a package dir — recording it as a limitation, not a regression.

Issue #606, opened 2026-07-31
flow replay: a raw `tool: flow-execute` step always reports PASS, whatever the composed flow did


hubgan:
## Summary

A recorded `tool: flow-execute` step replays as `status: "pass"` regardless of what the composed flow actually did. The enclosing run then reports a top-level `ok: true` for a composition that **failed every step**, or that **ran no steps at all**.

Found while verifying #574. **It is not caused by that PR** — it reproduces byte-identically on `main` (`eea6c27c`), and the relevant code is untouched by it. Filing separately as agreed on that PR.

## Where

- `packages/tool-server/src/tools/flows/flow-run.ts` — `case "tool"`: only `isUnmetUiWaitResult` is inspected; every other non-throwing result becomes `status: "pass"`. `flow-execute` returns `{ ok: false, ... }` or `{ notice: ... }` **without throwing**, so neither is noticed.
- The step is produced by the "kept the raw flow-execute step" fallback in `packages/tool-server/src/tools/flows/flow-add-step.ts` — reached when the target is not a sibling of the recording (cross-project composition) and, in client/remote persist mode, always.

`run:` composition does not have this problem; only the raw `tool:` form does.

## Reproduction

Real iOS simulator (iPhone 16, `D72CC5CA-FF01-4976-8211-50713576D3DA`), real app `com.apple.Preferences`, tool-server driven over HTTP.

```bash
mkdir -p projA projB/.argent/flows
cat > projB/.argent/flows/b-prereq.yaml <<'YAML'
executionPrerequisite: Settings must be open on the root Settings page
steps:
  - echo: this never runs
YAML

t() { curl -s -X POST "http://127.0.0.1:$PORT/tools/$1" -H 'content-type: application/json' -d "$2"; }

t flow-start-recording '{"name":"raw-prereq","project_root":"'$PWD'/projA"}'
# projB/b-prereq is not a sibling of projA -> the raw flow-execute step is kept
t flow-add-step '{"name":"raw-prereq","project_root":"'$PWD'/projA","command":"flow-execute","args":"{\"name\":\"b-prereq\",\"project_root\":\"'$PWD'/projB\",\"device\":\"'$UDID'\"}"}'
t flow-finish-recording '{"name":"raw-prereq","project_root":"'$PWD'/projA"}'
t flow-execute '{"name":"raw-prereq","project_root":"'$PWD'/projA","device":"'$UDID'"}'
```

### Observed

Recorded YAML:

```yaml
steps:
[...]

PR #610, opened 2026-07-31
fix(debugger): cut debugger tool fail rate — structured not-connected results, classified CDP failures, narrow recovery

filip131311:
## Why

PostHog (30d) shows the debugger tool family is argent's least reliable surface:

| tool | fail rate | dominant causes |
|---|---|---|
| `debugger-status` | **37%** | METRO_NO_TARGETS / SERVICE_INITIALIZATION_FAILED / zod |
| `debugger-connect` | **24%** | SERVICE_INITIALIZATION_FAILED (~7k, >1k users), METRO_NO_TARGETS |
| `debugger-log-registry` | 41.7% raw / 4.2% excl. storms | one user retry-looped it **12,742×** on NO_TARGETS |

Three design issues drive this: (1) "app/Metro not ready yet" — an expected precondition — is reported as a tool *error*, which both inflates fail rates and invites agent retry storms; (2) every CDP init/lifecycle fault lands in telemetry as the catch-all `REGISTRY_SERVICE_INITIALIZATION_FAILED` with `error_kind=unknown`; (3) ~600 users/30d hit zod validation failures with zero visibility into which parameter was wrong.

## What

**1. Classified failures** (`d71efa43`) — six former plain-`Error` sites in `cdp-client.ts` / `target-selection.ts` now carry dedicated `DEBUGGER_CDP_*` / `DEBUGGER_TARGET_DEVICE_MISMATCH` signals (messages byte-identical). The registry's two causeless "terminating window" `ServiceInitializationError`s get their own `REGISTRY_SERVICE_TERMINATING` code.

**2. Narrow recovery** (`ff9c1955`) — `recoverable()` on both debugger blueprints, scoped to windows where retry is provably safe: Metro recovers only `DEBUGGER_CDP_NOT_CONNECTED` (request never left the host; a detected socket death already self-heals for the *next* call via the terminated cascade — recovering `CONNECTION_CLOSED` there would be dead code, see the blueprint comment). Chromium also recovers `CONNECTION_CLOSED` for the tab-switch reconnect window. `REQUEST_TIMEOUT` stays unrecoverable (double-execution risk). A reachability test proves the retry actually fires for a RUNNING node.

**3. Structured not-connected results** (`48b91ffd`) — `debugger-status` and `debugger-log-registry` become factories that return `{ status: "not_connected", reason, detail, guidance }` for classified preconditions (`metro_not_running | no_app_connected | device_mismatch | cdp_unreachable | stale_connection | reconnecting`) and still throw on unexpected faults. `detail` preserves the original error text for string-matching agents. `debugger-status` gains a socket-state gate (Metro: dispose stale node; Chromium: report transient reconnect, no dispose — the client belongs to the ChromiumCdp dependency). log-registry keeps **no** gate on purpose: captured logs are readable over a dead socket, and disposal would destroy the post-crash log file. Flows/run-sequence map a not_connected result to a **failed step** (mirrors `isUnmetUiWaitResult`) so recorded connectivity gates don't silently pass. `debugger-evaluate` re-codes agent-expression throws as `DEBUGGER_EVALUATE_EXPRESSION_THREW`.

**4. Telemetry + guidance + docs** (`7602a70d`) — `tool:fail.invalid_params` (schema-declared names only; unknown keys → literal `"unrecognized_keys"`; capped emit-side at 16), new `debugger:tool_outcome` event (coded enum; joins `tool:invoke`/`complete` via `tool_invocation_id` — `ai_client`/`duration` live on the joined row), anti-retry-loop guidance appended to the Metro discovery errors (first sentences verbatim — skills match on them), six skills updated, Telemetry.md bumped to v1.03 with the new coded diagnostics disclosed.

## Dashboard migration note

`debugger-status` / `debugger-log-registry` precondition failures stop emitting `tool:fail` — track them via `debugger:tool_outcome` (`outcome != 'connected'`). Per-AI-client breakdowns of those outcomes need a join on `tool_invocation_id` against `tool:complete`.

## Testing

- Full suites green: tool-server 3,142, telemetry 290, registry 87. New coverage: recoverable truth tables (incl. non-Error inputs), status not-connected matrix (Metro/Chromium/terminating-window/stale/dispose-side-effect/fall-through-rethrow), log-registry rows, zod `invalid_params` derivation + emission-level guard, flow gate, evaluate rewrap, sanitize-level enum checks.
- Live-tested against a booted iOS simulator + RN app and an Electron target (details in PR comments).

Plan hardened through 3 adversarial review rounds (46 confirmed findings addressed) before implementation.


packages/tool-server/src/tools/debugger/not-connected.ts
```diff
@@ -0,0 +1,156 @@
...
+  stale_connection:
+    "The cached debugger connection went stale; it has been discarded. Restart the app " +
+    "(restart-app) if it is not running, then call debugger-connect — the next call " +
+    "reconnects fresh.",
+  reconnecting:
+    "The debugger connection is being re-established (the previous one was torn down or a " +
+    "tab switch is in progress). Wait a moment and retry once.",
+};
+
+const NOT_CONNECTED_CODE_MAP: Record<string, DebuggerNotConnectedReason> = {
```
latekvo, line 67: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `NOT_CONNECTED_CODE_MAP` omits two failure codes that are reachable while the service is being
resolved, so `debugger-status` / `debugger-log-registry` still return HTTP 500 in states the new
description promises they cannot: *"Never fails when the runtime is simply unreachable — it
returns a discriminated result instead"* (`debugger-status.ts:51`), echoed in
`argent-metro-debugger/SKILL.md:35`.

**`CHROMIUM_CDP_INVALID_RESPONSE`** — a non-CDP HTTP server on a Chromium debug port. The Metro
arm treats the identical situation as a structured result, and this PR even added guidance text
for it ("Something else is listening on this port — it did not answer like Metro"). Ran both arms
against one plain `python3 -m http.server` on port 9300, via the tool-server built from this
branch:

```
POST /tools/debugger-status {"device_id":"emulator-5554","port":9300}
  -> 200 {"status":"not_connected","reason":"metro_not_running", ...}
POST /tools/debugger-status {"device_id":"chromium-cdp-9300"}
  -> 500 {"error":"... [ChromiumCdp:chromium-cdp-9300] Chromium CDP discovery: GET
          http://127.0.0.1:9300/json/version failed (HTTP 404)"}
POST /tools/debugger-log-registry {"device_id":"chromium-cdp-9300"}
  -> 500 (same)
```
`CHROMIUM_CDP_INVALID_RESPONSE` is the only one of `cdp-session.ts`'s three discovery codes not in
the map; the other two are on lines 63-64.

**`DEBUGGER_CDP_REQUEST_TIMEOUT`** — raised by the four un-`catch`ed sends in the blueprint factory
(`js-runtime-debugger.ts:226-231`) when Metro lists a target whose JS runtime is frozen but whose
socket is still open. Against a mock Metro serving one target whose WebSocket accepts and never
replies:

```
POST /tools/debugger-status {"device_id":"emulator-5554","port":19500}
  -> 500 {"error":"... CDP request Runtime.enable (id=3) timed out"}   after 30s
```
No `reason`, no `guidance`, 30 s of dead time — in the state a diagnostic tool is most likely to
be reached for. The `failure-scenarios.md` row this PR adds for that code describes the opposite
observation (*"`debugger-status` may still report `"connected"` in this state"*), which holds only
when the hang starts after a successful connect, not before it.

packages/tool-server/src/utils/debugger/discovery.ts
```diff
@@ -55,7 +58,9 @@ export async function discoverMetro(port: number): Promise<MetroInfo> {
   const statusText = await statusRes.text();
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Only the first probe is wrapped: the `try`/`catch` at lines 38-53 converts a bare
`TypeError: fetch failed` into a classified `DEBUGGER_METRO_NOT_RUNNING`, and its own comment says
an unclassified one "would surface as an opaque 500". The two later network reads have no such
guard — `await statusRes.text()` on this line, and `await fetch(/json/list)` on line 82 — so a
connection-level failure after the `/status` probe escapes as a plain `Error` with no
`FailureSignal`. `classifyNotConnected` then returns `undefined` and both "never fails" tools
rethrow.

Reproduced against the tool-server built from this branch, with a mock Metro that answers
`/status` like Metro and then destroys the socket on `/json/list`:

```
POST /tools/debugger-status       {"device_id":"emulator-5554","port":19200}
  -> 500 {"error":"[Tool:debugger-status] Service dependency failed:
          [JsRuntimeDebugger:19200:emulator-5554] fetch failed — caused by: other side closed"}
POST /tools/debugger-log-registry {"device_id":"emulator-5554","port":19200}
  -> 500 (same)
```
No `reason`, no `guidance` — the caller gets exactly the opaque failure the guarded sibling exists
to prevent, for the ordinary case of Metro going down between the two probes.

packages/tool-server/src/blueprints/chromium-js-runtime-debugger.ts
```diff
@@ -140,6 +142,26 @@ export const chromiumJsRuntimeDebuggerBlueprint: ServiceBlueprint<JsRuntimeDebug
...
+  // CDPClient.reconnect() rejects in-flight requests with CONNECTION_CLOSED (and
+  // late sends with NOT_CONNECTED) while suppressing the `disconnected` event,
+  // and the same client object comes back alive on the new tab. Retrying re-runs
+  // the tool against the (re)opening socket — side effects on the discarded tab
+  // are moot. Best-effort: an immediate retry can still land before the new
+  // handshake completes and re-fail with a classified NOT_CONNECTED. A genuinely
+  // dead Chromium socket instead fires ChromiumCdp's own terminated event, whose
+  // teardown cascades into this dependent — the node leaves RUNNING and the next
+  // call re-resolves fresh, no recovery involved.
+  recoverable(error: unknown): boolean {
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This `recoverable()` makes `Registry._recoverFailedServices` call `disposeService()` on the
ChromiumJsRuntimeDebugger node (`registry.ts:255-266`). That blueprint's dispose calls
`logWriter.close()` (line 276), and `LogFileWriter.close()` ends in `fs.unlinkSync(this.filePath)`
(`log-file-writer.ts:231`) — so the Chromium session's whole captured console log is deleted from
disk, and the re-resolve mints a new empty `argent-logs-<port>-<ts>.log`.

The dispose runs unconditionally *before* the retry, so the log is destroyed whether or not the
retry then succeeds. It is not offset by a cascade either: this is precisely the window where
nothing is dying — `CDPClient.reconnect()` (`cdp-client.ts:209`) strips the old socket's listeners
so no `disconnected`/`terminated` fires, and both ChromiumCdp and the wrapper stay RUNNING, which
is why the self-heal engages at all.

Reproduced with two real CDP page-target WebSocket servers, a real `CDPClient` on a real socket,
the real blueprint and the real `debugger-evaluate` tool, driven through `registry.invokeTool`,
with the same `cdp.reconnect(...)` call `chromium-tabs` makes (`chromium-server/tabs.ts:139`):

```
BEFORE file: /home/user/.argent/tmp/argent-logs-37933-1785753275947.log  entries: 1
tool outcome: FAILED: [Tool:debugger-evaluate] CDP not connected
log file still on disk? false
JSD state: RUNNING

CONTROL (same window, `recoverable` deleted from the blueprint):
CONTROL tool outcome: [Tool:debugger-evaluate] CDP not connected
CONTROL log file on disk? true      (still contains the captured entry)
```
In that run the retry healed nothing and the log file was gone. Reachable from every debugger tool
that still declares `services: (params) => ({ debugger: debuggerServiceRef(params) })` —
`debugger-evaluate`, `debugger-connect`, the `view-network-*` tools.

This also cuts against the reasoning elsewhere in the PR: `debugger-log-registry` is deliberately
given no socket gate because "disposing the stale service would close the LogFileWriter —
destroying exactly the post-crash logs the caller came for" (`debugger-log-registry.ts:56-59`).

packages/tool-server/src/tools/debugger/debugger-status.ts
```diff
@@ -12,49 +19,101 @@ const zodSchema = z.object({
...
+          // remove the node first; that end state is what we wanted, so a
+          // failed dispose is absorbed.
+          const ref = debuggerServiceRef(params);
+          await registry.disposeService(typeof ref === "string" ? ref : ref.urn).catch(() => {});
+          const result = buildNotConnected(
+            "stale_connection",
+            new Error("Cached debugger connection is no longer open"),
+            params
+          );
+          trackDebuggerOutcome("debugger-status", "stale_connection", params, ctx);
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `disposeService` on line 88 runs the blueprint dispose, whose first statement is
`forgetDeviceAlias(api.logicalDeviceId)` (`js-runtime-debugger.ts:308`). `trackDebuggerOutcome`
then derives `platform` from `canonicalDeviceId(params.device_id)` — reading the alias this same
call has already deleted.

So when the caller passes the forwarded Metro `logicalDeviceId` (which
`argent-metro-debugger/SKILL.md` tells agents to do, and which this tool's own `device_id`
description accepts), the 40-hex handle falls back to shape classification and an iOS session is
reported as Android. That is the exact misattribution `test/debugger/track-outcome-platform.test.ts`
was added to prevent; it calls `trackDebuggerOutcome` directly, so it does not cover this branch.

Observed on the tool-server built from this branch, same iOS UDID and same returned
`logicalDeviceId` throughout (mock Metro, socket put into a half-closed state to reach the gate):

```
debugger:tool_outcome {outcome: connected,        platform: ios}
debugger:tool_outcome {outcome: stale_connection, platform: android}
```

packages/tool-server/src/tools/debugger/not-connected.ts
```diff
@@ -0,0 +1,156 @@
...
+  ctx: ToolContext | undefined
+): void {
+  let platform;
+  try {
+    // Classify the id the caller CONNECTED with, not the raw param: a forwarded
+    // Metro logicalDeviceId (an opaque hex handle) fails the iOS-UDID shape
+    // test and would misreport every iOS Metro session as "android". The alias
+    // map (learned at connect) rewrites it back to the UDID/serial; ids with no
+    // learned alias pass through unchanged, keeping the old behavior.
+    const deviceId = canonicalDeviceId(params.device_id);
```
latekvo, line 152: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `platform` here is derived from the alias-rewritten id, but `http.ts`'s `inferPlatform`
(`http.ts:153-160`), which feeds `tool:invoke` / `tool:complete` / `tool:fail`, classifies the raw
`device` argument. The JSDoc on this event states that `tool_invocation_id` "joins 1:1 against the
`tool:invoke` / `tool:complete` pair" (`events.ts:200`), so a dashboard doing that join gets two
different platforms for one call.

Captured from the tool-server built from this branch — `debugger-connect` with an iOS UDID, then
`debugger-status` with the `logicalDeviceId` it returned:

```
tool:invoke            tool_invocation_id=a874f8a0-c177-4dc2-99b8-8fe9fb8ef1de  platform=android
debugger:tool_outcome  tool_invocation_id=a874f8a0-c177-4dc2-99b8-8fe9fb8ef1de  platform=ios
tool:complete          tool_invocation_id=a874f8a0-c177-4dc2-99b8-8fe9fb8ef1de  platform=android
```
The same skew is on the `tool:*` events of every other debugger tool that takes a forwarded
`logicalDeviceId` (`debugger-evaluate`, `-component-tree`, `-inspect-element`, `-reload-metro`),
which matters for a PR whose stated goal is a per-platform debugger failure-rate dashboard.

packages/tool-server/src/tools/debugger/not-connected.ts
```diff
@@ -0,0 +1,156 @@
...
+    "until Metro is started. Start Metro (e.g. `npx react-native start` or `npx expo start`) " +
+    "or ask the user, wait for it to report ready, then retry once.",
+  no_app_connected:
+    "Metro is running but no app is attached. Do not retry immediately — launch or restart " +
+    "the RN app on the target device (launch-app / restart-app), wait a few seconds for the " +
+    "bundle to load, then retry once.",
+  device_mismatch:
+    "The device_id does not match any debugger target on this Metro. Re-target with the " +
+    "logicalDeviceId listed in the detail message, or give the device its own Metro port.",
+  cdp_unreachable:
```
latekvo, line 50: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `cdp_unreachable` is the reason both Chromium codes map to (lines 63-64), but its guidance tells
the agent to "Verify the app is running (launch-app)". On Chromium `launch-app` cannot start
anything — its own handler documents itself as a no-op ("no concept of installing or launching a
separate bundle inside one Chromium instance", `launch-app/platforms/chromium.ts`) — and it
declares `chromium: chromiumCdpRef(device)` as a service (`launch-app/index.ts:95`), so it
re-resolves the very service that just failed.

Following the guidance therefore produces a second, guaranteed `tool:fail` carrying the same
underlying error. Both calls against the tool-server built from this branch:

```
POST /tools/debugger-status {"device_id":"chromium-cdp-59999"}
  -> 200 reason: cdp_unreachable
     guidance: "The runtime's CDP endpoint could not be reached. Verify the app is running
                (launch-app), then call debugger-connect and retry once."
POST /tools/launch-app {"udid":"chromium-cdp-59999","bundleId":"com.example.app"}
  -> 500 {"error":"[Tool:launch-app] Service dependency failed: [ChromiumCdp:chromium-cdp-59999]
          Chromium CDP discovery: GET http://127.0.0.1:59999/json/version could not connect ..."}
```
An agent obeying the guidance manufactures the failure this PR set out to remove.

packages/tool-server/src/tools/debugger/debugger-status.ts
```diff
@@ -12,49 +19,101 @@ const zodSchema = z.object({
...
+              "reconnecting",
+              new Error("CDP socket is reconnecting (tab switch in progress)"),
+              params
+            );
+            trackDebuggerOutcome("debugger-status", "reconnecting", params, ctx);
+            return result;
+          }
+          // Metro path owns its CDPClient: discard the stale node so the next
+          // call reconnects fresh. Tradeoff acknowledged: dispose closes the
+          // LogFileWriter, so any logs surviving in the suppressed-close window
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Two statements in this comment do not hold for the branch they annotate.

"any logs surviving in the **suppressed-close window**" — the only close suppression in the
codebase is `CDPClient.reconnect()` (`cdp-client.ts:209`, documented as re-pointing the socket
"WITHOUT emitting `disconnected`"), and `grep -rn "\.reconnect(" packages/tool-server/src` returns
exactly one caller: `chromium-server/tabs.ts:139`, a Chromium tab switch. On the Metro path
nothing suppresses a close; the state this branch actually fires in is the one the sibling comment
this PR adds at `js-runtime-debugger.ts:139-144` names precisely — "the WebSocket is CLOSING but
the close event has not dispatched yet".

"read debugger-log-registry BEFORE status if post-crash logs matter" — that ordering does not
preserve the logs. `debugger-log-registry` returns a `file` path plus counts and 20 truncated
clusters, and both its own description ("call this first for an overview, then read the returned
file for details") and `argent-metro-debugger/SKILL.md` §5 ("1. Call `debugger-log-registry` …
2. Search the file using `Grep` or `Read`") send the agent back to that path afterwards. The
dispose on line 88 unlinks it (`log-file-writer.ts:231`). Verified over HTTP against the
tool-server built from this branch:

```
POST /tools/debugger-connect      -> ok
POST /tools/debugger-log-registry -> file: /home/user/.argent/tmp/argent-logs-8800-...log   (exists)
<half-close the CDP socket>
POST /tools/debugger-status       -> 200 reason: "stale_connection"
$ ls /home/user/.argent/tmp/argent-logs-8800-...log   ->  No such file or directory
```

packages/telemetry/src/sanitize.ts
```diff
@@ -234,9 +239,19 @@ export const ALLOWED: ValidatorMap = {
     tool_invocation_id: UUID,
     platform: PLATFORM,
     duration_ms: DURATION_MS,
+    // Schema-declared parameter names only (emit side filters against the tool's
+    // own zod shape and caps at 16 before this gate); the array validator also
+    // voids anything longer or with a non-identifier element.
+    invalid_params: arrayOf(matches(/^[a-z][a-z0-9_]{0,63}$/i, 64), 16),
```
latekvo, line 245: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Neither half of this validator is pinned by a test, and `arrayOf` is all-or-nothing — one failing
element or one element over the cap voids the entire property (`sanitize.ts:40-52`), silently, with
no error anywhere.

Both mutations survive the whole telemetry suite:

```
# drop the /i flag:  matches(/^[a-z][a-z0-9_]{0,63}$/i, 64) -> matches(/^[a-z][a-z0-9_]{0,63}$/, 64)
$ cd packages/telemetry && npx vitest run --reporter=dot
  Test Files  16 passed (16)   Tests  290 passed (290)

# shrink the cap below the emit-side cap:  arrayOf(..., 16) -> arrayOf(..., 8)
  Test Files  16 passed (16)   Tests  290 passed (290)
```
The new `tool:fail invalid_params` block only ever feeds lowercase snake_case names
(`["device_id","port","unrecognized_keys"]`) and array lengths 3 and 17, so neither the
case-insensitivity nor the 16 boundary is exercised.

The `/i` flag is load-bearing: loading every registered tool from
`createRegistry()` and testing its declared zod keys gives 101 of 294 keys containing an uppercase
letter, across 34 of the 72 tools with schemas (`boot-device`, `launch-app`, `screenshot`,
`gesture-tap`, `chromium-cookies`, …). A single camelCase failing param — `bundleId` on
`launch-app` — would void the whole property for that call.

packages/tool-server/src/tools/debugger/not-connected.ts
```diff
@@ -0,0 +1,156 @@
...
+};
+
+const NOT_CONNECTED_CODE_MAP: Record<string, DebuggerNotConnectedReason> = {
+  [FAILURE_CODES.DEBUGGER_METRO_NOT_RUNNING]: "metro_not_running",
+  [FAILURE_CODES.DEBUGGER_METRO_NO_TARGETS]: "no_app_connected",
+  [FAILURE_CODES.DEBUGGER_TARGET_DEVICE_MISMATCH]: "device_mismatch",
+  [FAILURE_CODES.DEBUGGER_CDP_CONNECT_FAILED]: "cdp_unreachable",
+  [FAILURE_CODES.DEBUGGER_CDP_SOCKET_CLOSED_BEFORE_OPEN]: "cdp_unreachable",
+  [FAILURE_CODES.DEBUGGER_CDP_NOT_CONNECTED]: "cdp_unreachable",
+  [FAILURE_CODES.DEBUGGER_CDP_CONNECTION_CLOSED]: "cdp_unreachable",
```
latekvo, line 74: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Half of this map is unpinned: deleting any one of five entries turns a structured `not_connected`
result back into a thrown tool failure — the regression this PR exists to prevent — with nothing
going red.

```
$ cd packages/tool-server
$ npx vitest run test/debugger test/flows test/metro test/utils --reporter=dot
  BASELINE                                       Tests  898 passed (898)
  del DEBUGGER_TARGET_DEVICE_MISMATCH            Tests  898 passed (898)
  del DEBUGGER_CDP_SOCKET_CLOSED_BEFORE_OPEN     Tests  898 passed (898)
  del DEBUGGER_CDP_NOT_CONNECTED                 Tests  898 passed (898)
  del DEBUGGER_CDP_CONNECTION_CLOSED             Tests  898 passed (898)
  del CHROMIUM_CDP_NO_PAGE_TARGET                Tests  898 passed (898)

  del DEBUGGER_METRO_NOT_RUNNING                 Tests  2 failed | 896 passed
  del DEBUGGER_METRO_NO_TARGETS                  Tests  1 failed | 897 passed
  del DEBUGGER_CDP_CONNECT_FAILED                Tests  1 failed | 897 passed
  del CHROMIUM_CDP_UNREACHABLE                   Tests  1 failed | 897 passed
  del REGISTRY_SERVICE_TERMINATING               Tests  2 failed | 896 passed
```
`CHROMIUM_CDP_NO_PAGE_TARGET` was re-run four times (focused and full suite) to rule out flake; the
full suite stays at exactly the 6 pre-existing `boot-device-hotboot` failures.

`device_mismatch` is one of the six reasons the tool description enumerates and has its own row in
`failure-scenarios.md`, so its mapping is part of the documented contract.

packages/registry/src/registry.ts
```diff
@@ -369,7 +373,9 @@ export class Registry {
           /* ignore */
         }
         node.initPromise = null;
-        throw new ServiceInitializationError(urn, "Service was terminated during initialization");
+        throw new ServiceInitializationError(urn, "Service was terminated during initialization", {
+          cause: terminatingSignalCause("Service was terminated during initialization"),
```
latekvo, line 381: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This is the one of the two new `terminatingSignalCause` sites that nothing exercises. Removing the
`cause:` here leaves the whole registry suite green:

```
$ cd packages/registry && npx vitest run --reporter=dot
  Test Files  6 passed (6)   Tests  87 passed (87)
```
`grep -rn "terminated during initialization" packages` finds no test anywhere — the only non-source
hit is `tool-server/src/tools/profiler/react/RFC-profiler-reconnect.md`, which documents this exact
state as an observed production failure ("the `_initialize` guard throws 'Service was terminated
during initialization'" when a CDP drop cascades while a dependent factory is still running).

Without the cause the error falls back to the catch-all `REGISTRY_SERVICE_INITIALIZATION_FAILED`,
which is not in `NOT_CONNECTED_CODE_MAP`, so the reason would silently change from `reconnecting`
to a thrown tool failure. The sibling site at line 331 is covered — test `(h)` in
`debugger-status-not-connected.test.ts` holds a node in TERMINATING and resolves it, which reaches
`registry.ts:328`, not this branch.

packages/tool-server/src/tools/debugger/debugger-log-registry.ts
```diff
@@ -20,35 +27,53 @@ const zodSchema = z.object({
...
+When the debugger cannot be reached, this tool does not fail: it returns { status: "not_connected", reason, detail, guidance } with NO log file — follow the guidance (do not retry in a loop, and do not try to read a log file from this state). A "connected" result's stats may come from a session whose socket has since died — use debugger-status, not this tool, to judge debugger health.`,
+    zodSchema,
+    capability: DEBUGGER_TOOL_CAPABILITY,
+    // Resolved manually in execute so a not-connected precondition becomes a
+    // structured result instead of a service-resolution tool failure.
+    services: () => ({}),
+    async execute(_services, params, ctx) {
+      try {
+        const api = await resolveDebuggerService(registry, params);
+        // Unlike debugger-status, no socket-state gate here: captured logs are
```
latekvo, line 65: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This deliberate asymmetry — the reason `debugger-log-registry` is safe to call over a dead socket
while `debugger-status` is not — is not pinned by anything. Adding a `debugger-status`-style socket
gate plus dispose right here leaves the suite green:

```
$ cd packages/tool-server
$ npx vitest run test/debugger test/flows test/metro test/utils --reporter=dot
  BASELINE                                Tests  898 passed (898)
  + socket gate & disposeService here     Tests  898 passed (898)
```
The consequence the comment describes is real and worse than "closes the writer":
`LogFileWriter.close()` ends in `fs.unlinkSync` (`log-file-writer.ts:231`), so the mutated version
deletes the on-disk log and the `not_connected` result also omits `file` — the caller loses the
stats and any route back to the data. Nothing would catch that regression.

packages/telemetry/src/events.ts
```diff
@@ -161,6 +161,50 @@ export interface ToolFailProps extends FailureTelemetryProps, AiTelemetryProps {
...
+export type DebuggerNotConnectedReason = (typeof DEBUGGER_NOT_CONNECTED_REASONS)[number];
+
+export const DEBUGGER_TOOL_OUTCOMES = ["connected", ...DEBUGGER_NOT_CONNECTED_REASONS] as const;
+export type DebuggerToolOutcome = (typeof DEBUGGER_TOOL_OUTCOMES)[number];
+
+/**
+ * Emitted by debugger-status / debugger-log-registry: exactly once whenever the
+ * tool returns a result, never on a thrown failure (unclassified faults, zod
+ * rejects, and capability gates still emit tool:fail only).
+ * Compensates for those tools no longer emitting tool:fail on not-connected
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: "Compensates for those tools **no longer** emitting `tool:fail` on not-connected preconditions" is
transition narration in permanent JSDoc: "no longer" points at a prior behaviour that will not
exist anywhere in the tree once this diff is history, and this JSDoc is the only documentation of
the event's semantics (`Telemetry.md` describes the disclosure, not the event). The preceding
sentence in the same block states the same fact in state terms and reads correctly.

Same class, same PR, also permanent: `packages/tool-server/test/metro/target-selection.test.ts:126`
— "// signal unblinds telemetry (previously a plain Error → unclassified SIE)".

Checked against the rest of the diff so this is not a blanket complaint: the other "no longer"
occurrences added here (`debugger-status.ts:65`, `:91`, and the `failure-scenarios.md`
stale-connection row) describe a socket's runtime state and are fine.

packages/skills/skills/argent-react-native-app-workflow/SKILL.md
```diff
@@ -85,7 +85,7 @@ To kill a Metro process, use the `stop-metro` tool (requires user confirmation).
 - **App must point at the same host/port as the running Metro.** Default: same machine, port 8081.
 - **iOS Simulator:** By default uses localhost; no extra config needed for same-machine Metro.
 
-**Verify Metro is reachable:** use the `debugger-status` tool.
+**Verify Metro is reachable:** use the `debugger-status` tool. Any result other than `reason: "metro_not_running"` means Metro answered (`reason: "no_app_connected"` means the app has not attached yet — see §2.3).
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: "Any result other than `reason: "metro_not_running"` means Metro answered" is false for two of the
six reasons this PR defines. `stale_connection` and `reconnecting` are both produced without any
request reaching Metro: `stale_connection` comes from the socket-state gate on a cached service
(`debugger-status.ts:88-95`) and `reconnecting` from `REGISTRY_SERVICE_TERMINATING`, i.e. a resolve
racing a teardown — neither performs Metro discovery.

Both are raised before any Metro discovery runs — `reconnecting` comes from the
`node.state === TERMINATING` guard at `registry.ts:328`, which rejects ahead of the factory.
Observed against the tool-server built from this branch by firing four concurrent
`debugger-status` calls at a device whose CDP socket had been half-closed:

```
{"status":"not_connected","reason":"reconnecting","detail":"... Service is currently terminating"}
{"status":"not_connected","reason":"reconnecting", ...}
{"status":"not_connected","reason":"reconnecting", ...}
{"status":"not_connected","reason":"stale_connection","detail":"Cached debugger connection is no longer open"}
```
An agent applying the rule in §2.2 would conclude Metro is reachable and skip starting it.

packages/tool-server/src/tools/debugger/not-connected.ts
```diff
@@ -0,0 +1,156 @@
...
+    const deviceId = canonicalDeviceId(params.device_id);
+    platform = deviceId ? classifyDeviceForTelemetry(deviceId) : undefined;
+  } catch {
+    platform = undefined;
+  }
+  track("debugger:tool_outcome", {
+    tool,
+    outcome,
+    ...(platform ? { platform } : {}),
+    ...(ctx?.toolInvocationId ? { tool_invocation_id: ctx.toolInvocationId } : {}),
```
latekvo, line 161: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `ctx.toolInvocationId` is only populated when the caller passes `options.toolInvocationId` into
`registry.invokeTool`. On the sub-invoke path `invokeSubTool` forwards it only on its attribution
branch (`utils/sub-invoke.ts:36-40`), which needs `ctx.recordChildInvocation` — and `http.ts` sets
that only when `extractInvocationMeta` returns non-null. For `flow-execute` it cannot: the tool
declares no `capability` (so the platform branch is skipped) and its device parameter is named
`device`, which `extractDeviceArg` does not read (it looks only at `udid` / `device_id`). The only
remaining source is the `x-argent-ai-client` header — which the shipped `@argent/tools-client`
`callTool` never sends, and that is the function `argent flow run` uses.

So inside a flow step the event is emitted without the join key, while its own `tool:invoke` /
`tool:complete` pair carries one. Both variables held constant except the header, against the
tool-server built from this branch (flow with a single `debugger-status` step):

```
no x-argent-ai-client:
  tool:invoke            tool_invocation_id=b7df2173-677a-410a-8600-5b42ee11d5f1
  debugger:tool_outcome  tool_invocation_id=None
  tool:complete          tool_invocation_id=b7df2173-677a-410a-8600-5b42ee11d5f1

x-argent-ai-client: claude_code
  tool:invoke            tool_invocation_id=1d6d4c1d-472f-45ca-8b5e-fffa5748b522
  debugger:tool_outcome  tool_invocation_id=1d6d4c1d-472f-45ca-8b5e-fffa5748b522
  tool:complete          tool_invocation_id=1d6d4c1d-472f-45ca-8b5e-fffa5748b522
```
The event deliberately carries neither `duration_ms` nor `ai_client` of its own, so an orphaned row
cannot be attributed at all — and since these tools no longer emit `tool:fail` for preconditions, a
not-connected step in a CLI flow run shows up as a bare `tool:complete` plus a floating outcome row.
This is the join the JSDoc here and at `events.ts:200` both promise is 1:1. A direct
`POST /tools/debugger-status` always carries the id, so the gap is specific to the sub-invoke path.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -1123,6 +1124,14 @@ async function execLeafStep(
             reason: `await-ui-element condition not met${note ? `: ${note}` : ""}`,
           };
         }
+        if (isDebuggerNotConnectedResult(step.name, result)) {
+          return {
+            ...base,
+            status: "fail",
+            tool: step.name,
+            reason: `debugger not connected (${result.reason}): ${result.guidance}`,
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This branch keeps `reason` (the per-reason `guidance`) but drops `result`, so `detail` never reaches
the flow report. `GUIDANCE.device_mismatch` explicitly forwards the agent to it — "Re-target with
the logicalDeviceId **listed in the detail message**" — and the detail message is the only place the
available `logicalDeviceId`s are named (`target-selection.ts:94-104`). A flow whose gate fails that
way tells the agent to read a field the report does not contain.

It is also a step backwards from the pre-PR behaviour for every reason. Before this PR these tools
threw, so the step took the `catch` below and reported `status: "error"` with `reason: errMsg(err)`
— the full text, including the `got:` fragment that identifies what actually answered the port. Now
only the generic per-reason string survives. Observed on the tool-server built from this branch, a
one-step flow gated on `debugger-status` against a dead port:

```
"status": "fail",
"tool": "debugger-status",
"reason": "debugger not connected (metro_not_running): Metro is not running on this port. Do not
           retry in a loop — the result will not change until Metro is started. Start Metro ..."
```
No `detail`, no `result`, so nothing says which port answered or what it said. (The
`await-ui-element` branch above drops the same fields, but its `note` is already carried into
`reason`, so no guidance of its own points at a dropped field.)

packages/skills/skills/argent-tv-interact/SKILL.md
```diff
@@ -61,7 +61,7 @@ Needs a Debug build + Metro running. argent only _connects_ to Metro — start M
 
 ## Debugging the JS runtime (Vega)
 
-Once that same Debug build + Metro setup is in place, the JS-runtime tools work on a Vega VVD: `debugger-connect`, `debugger-status`, `debugger-evaluate`, `debugger-log-registry` (console logs), `view-network-logs`, and `view-network-request-details`. See the `argent-metro-debugger` skill.
+Once that same Debug build + Metro setup is in place, the JS-runtime tools work on a Vega VVD: `debugger-connect`, `debugger-status`, `debugger-evaluate`, `debugger-log-registry` (console logs), `view-network-logs`, and `view-network-request-details`. Verify with `debugger-status`: it returns a status result rather than an error when not connected — `status: "connected"` means the setup works; `status: "not_connected"` carries a `reason` and `guidance` (e.g. `metro_not_running` → Metro/port-forwarding not up, `no_app_connected` → the Debug app hasn't attached). See the `argent-metro-debugger` skill.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: "`metro_not_running` → Metro/port-forwarding not up" attributes the port-forward to a reason it
cannot produce. `discoverMetro` probes `http://localhost:${port}/status` from the **tool-server
host** (`discovery.ts:40`), which on a Vega setup is the same machine running Metro; the only host
rewrite in the blueprint is gated on `classifyDevice(deviceId) === "ios-remote"`
(`js-runtime-debugger.ts:199`). `vega device start-port-forwarding` is the device→host hop, so
whether it is up has no bearing on that loopback probe.

With the forward down, Metro answers `packager-status:running`, the VVD never attaches, `/json/list`
is empty, and the tool returns `no_app_connected`. Confirmed against the tool-server built from this
branch (mock Metro with an empty target list vs. a dead port):

```
port with Metro but no targets -> {"reason":"no_app_connected", ...}
dead port                      -> {"reason":"metro_not_running", ...}
```
So the reason a Vega agent actually gets for a broken forward is the one this sentence attributes
solely to "the Debug app hasn't attached", and that reason's `guidance` says only "launch or restart
the RN app on the target device (launch-app / restart-app)" — it never mentions
`start-port-forwarding`. An agent following both will restart the app repeatedly instead.

packages/tool-server/src/tools/debugger/not-connected.ts
```diff
@@ -0,0 +1,156 @@
...
+ * Flow integration: a not_connected result is a successful tool return, but a
+ * recorded flow step that used these tools as a connectivity gate must not
+ * silently green-pass on it. Mirrors isUnmetUiWaitResult for await-ui-element.
+ */
+export function isDebuggerNotConnectedResult(
+  toolId: string,
+  result: unknown
+): result is DebuggerNotConnectedResult {
+  return (
+    (toolId === "debugger-status" || toolId === "debugger-log-registry") &&
```
latekvo, line 190: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Only the `debugger-status` arm of this predicate is exercised. Narrowing it to
`toolId === "debugger-status"` — which makes a `debugger-log-registry` connectivity gate silently
green-pass in a flow, the exact behaviour the JSDoc above says must not happen — leaves the suite
green:

```
$ cd packages/tool-server
$ npx vitest run test/debugger test/flows test/metro test/utils --reporter=dot
  BASELINE                          Tests  898 passed (898)
  arm narrowed to debugger-status   Tests  898 passed (898)
```
`test/flows/flow-debugger-gate.test.ts` has three cases, all driving `debugger-status`; its third
case pins the negative (an unrelated tool returning `status` must not trip the mapping), but nothing
pins the second tool id.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat)\]: Reviewed at 9b8fd27440686230cb788aba7ee78733d4ce9b59. Ran the tool-server built from this branch end to end over HTTP (mock Metro + CDP targets, a mock Chromium debug port, and half-closed/hung-runtime sockets), plus the tool-server, telemetry and registry suites and targeted mutation runs. Findings are inline. Note that `packages/tool-server/test/boot-device-hotboot.test.ts` fails 6 tests on this Linux host both on this branch and on `main`, so it is unrelated to this PR.

filip131311:
Rebased onto main (0.19.0 — the flow-run conflict was against #606's nested-orchestrator mapping; both mappings now coexist) and addressed all 17 findings from the Diplomat review at 9b8fd27, in four commits:

**`review: classify every reachable resolve failure…`**
- `CHROMIUM_CDP_INVALID_RESPONSE` → `cdp_unreachable` and `DEBUGGER_CDP_REQUEST_TIMEOUT` → new reason **`runtime_unresponsive`** (own guidance: each retry waits out the full CDP timeout, so looping is expensive; distinct from the post-connect hang where status legitimately says `connected` — `failure-scenarios.md` row rewritten to cover both shapes).
- `discoverMetro`'s two later network reads (`/status` body, `/json/list` fetch) now classify a mid-sequence Metro death as `DEBUGGER_METRO_NOT_RUNNING` instead of escaping as an opaque 500.
- `cdp_unreachable` / `runtime_unresponsive` guidance is platform-aware: the Chromium variant no longer points at `launch-app` (documented no-op that re-resolves the failing service) — it says to relaunch with `--remote-debugging-port` / `boot-device` + `electronAppPath`.
- The stale-connection outcome is tracked **before** the dispose that forgets the alias, so its platform no longer flips to `android`.
- Every `NOT_CONNECTED_CODE_MAP` row is now individually pinned (`not-connected-map.test.ts`); deleting any entry goes red.

**`review: drop the Chromium wrapper's recoverable()`**
You were right that this was worse than useless: recovery disposed the wrapper, whose dispose unlinks the captured console log, and the same-client reconnect means the cached node heals for the next call without any dispose. Removed entirely; the truth-table test pins the absence, and a new log-registry test pins the dead-socket asymmetry (no gate, no dispose, stats + on-disk file survive).

**`review: make the tool_outcome join key and platform dimension hold on every dispatch path`**
- `registry.invokeTool` now puts the id it minted on `ctx.toolInvocationId`, so a flow/CLI sub-invocation's `debugger:tool_outcome` carries the same id as its `tool:invoke`/`tool:complete` pair (previously `None` without the attribution header).
- `http.ts` platform inference canonicalizes a forwarded `logicalDeviceId` through the device alias before classifying — `tool:invoke`/`complete`/`fail` and `debugger:tool_outcome` now report one platform for one call.
- Sanitize-side `invalid_params` validator pinned (camelCase names via the `/i` flag, the 16-element boundary both sides), and the factory-guard `terminatingSignalCause` at `registry.ts` now has a test proving the `REGISTRY_SERVICE_TERMINATING` code survives.

**`review: carry detail into flow gate reports…`**
- A failed debugger gate step now includes `detail` in `reason` and carries the full structured `result` (so `device_mismatch`'s \"listed in the detail message\" pointer resolves, and the `got:` fragment survives).
- The `debugger-log-registry` arm of the flow-gate predicate is pinned by test.
- Doc truth fixes: RN-workflow §2.2 no longer claims any non-`metro_not_running` result means Metro answered (`stale_connection`/`reconnecting` fire before any probe); tv-interact maps a broken Vega port-forward to `no_app_connected` (the `/status` probe is host-local) and says to check `start-port-forwarding` before relaunching; the two \"no longer\"/\"previously\" transition narrations are reworded in state terms; the debugger-status stale-branch comment now states plainly that the dispose deletes the log file (no false \"read log-registry first\" workaround).

**Testing**
- Full suites green: tool-server **3460** (316 files, 1 pre-existing skip), telemetry **295**, registry **90**.
- Live against the branch tool-server (:3010, iPhone 17 Pro sim + my-app Expo + electron-test-app):
  - your two 500 repros now return structured results: plain HTTP server on a chromium id → `cdp_unreachable` (`detail: …HTTP 404`, no `(launch-app)` in guidance); frozen-runtime mock (WS accepts, never replies) → `runtime_unresponsive` after the CDP timeout;
  - mock Metro that dies between `/status` and `/json/list` → `metro_not_running`;
  - connected/`no_app_connected`/kill-cascade matrix unchanged; forwarded `logicalDeviceId` now yields `platform: ios` on **both** `tool:invoke` and `debugger:tool_outcome` (was the `android`/`ios` skew);
  - flow with a dead-port `debugger-status` gate, no AI-client header: step fails with `detail` + full `result` in the report, and the nested outcome event carries the same `tool_invocation_id` as its `tool:invoke`/`complete` pair;
  - Electron: `window.open` second tab → `chromium-tabs select` (real `cdp.reconnect`) → captured log file **survives** with its entries and `log-registry` still points at it; killed app → `cdp_unreachable` with the Chromium-corrected guidance;
  - `launch-app` with a bad `bundleId` → `tool:fail` carrying `invalid_params: [\"bundleId\"]` (camelCase through the sanitizer).

packages/tool-server/src/tools/debugger/not-connected.ts
filip131311: Fixed: `CHROMIUM_CDP_INVALID_RESPONSE` → `cdp_unreachable`, and `DEBUGGER_CDP_REQUEST_TIMEOUT` → new reason `runtime_unresponsive` (own guidance — each retry waits out the full CDP timeout; the timeout error message itself now explains the frozen-runtime state). The failure-scenarios row now covers both the pre-connect and post-connect shapes. Live-verified both of your repros: plain HTTP server on a chromium id → structured `cdp_unreachable` with the HTTP 404 detail; accept-and-never-reply WS → `runtime_unresponsive` after the timeout.

packages/tool-server/src/utils/debugger/discovery.ts
filip131311: Fixed: both later reads (`/status` body, `/json/list` fetch) now classify as `DEBUGGER_METRO_NOT_RUNNING` with the same first-sentence contract and no-loop guidance (stages `debugger_discover_metro_status_body` / `debugger_discover_metro_list`). Tests cover a body stream that dies and a rejected list fetch; reproduced your mock live — now a structured `metro_not_running`.

packages/tool-server/src/blueprints/chromium-js-runtime-debugger.ts
filip131311: Fixed by removing `recoverable()` from this blueprint entirely — you're right it destroyed the log while healing nothing (the reconnect re-points the same client object, so the cached node heals for the next call without any dispose; the failing call surfaces a classified error debugger-status maps to `reconnecting`). The blueprint comment now documents the deliberate absence, the truth-table test pins it, and a live tab-switch reconnect on the Electron testbed kept the captured log file with its entries.

packages/tool-server/src/tools/debugger/debugger-status.ts
filip131311: Fixed: the outcome is now tracked BEFORE the dispose that forgets the alias (comment in the stale branch explains why the order is load-bearing). Test (l) in debugger-status-not-connected.test.ts pins it end-to-end — connect with a UDID against a mock Metro advertising a logicalDeviceId, forward the logical id into the stale gate, and assert the outcome still says `platform: ios`.

packages/tool-server/src/tools/debugger/not-connected.ts
filip131311: Fixed: `http.ts` `inferPlatform` now canonicalizes through the same device-alias map before classifying, so `tool:invoke`/`complete`/`fail` and `debugger:tool_outcome` report one platform for one `tool_invocation_id` (covers every debugger tool that accepts a forwarded logicalDeviceId). Pinned in http-platform-alias.test.ts; live-verified: connect with a UDID, call with the returned logicalDeviceId → `ios` on both event families.

packages/tool-server/src/tools/debugger/not-connected.ts
filip131311: Fixed: `cdp_unreachable` (and the new `runtime_unresponsive`) guidance is platform-aware — the Chromium variant never mentions launch-app and points at a relaunch with `--remote-debugging-port` (boot-device + electronAppPath for Electron). Pinned in not-connected-map.test.ts; live-verified against a killed Electron app.

packages/tool-server/src/tools/debugger/debugger-status.ts
filip131311: Fixed: the comment now names the real window (WebSocket CLOSING, close not yet dispatched) and states plainly that the dispose deletes the captured log file — including a path a prior debugger-log-registry call returned — with no false read-ordering workaround.

packages/telemetry/src/sanitize.ts
filip131311: Fixed: both halves are now pinned in sanitize.test.ts — camelCase names (`bundleId`, `electronAppPath`) must pass (kills the /i-flag mutation), exactly 16 names must pass and 17 must void the property (kills the cap mutation), plus non-identifier and >64-char elements voiding the whole array. Also verified live end-to-end: a bad `bundleId` on launch-app emitted `invalid_params: ["bundleId"]`.

packages/tool-server/src/tools/debugger/not-connected.ts
filip131311: Fixed: every row of NOT_CONNECTED_CODE_MAP (including the two newly added codes) is individually pinned in not-connected-map.test.ts, plus the unmapped-code and plain-Error rethrow paths — deleting any single entry now goes red.

packages/registry/src/registry.ts
filip131311: Fixed: a registry test now holds a factory open, disposes the node mid-init, and asserts the rejection carries the `REGISTRY_SERVICE_TERMINATING` cause — removing this `cause:` goes red.

packages/tool-server/src/tools/debugger/debugger-log-registry.ts
filip131311: Fixed: pinned in log-registry-not-connected.test.ts — after a connect, the socket is forced dead and the next call must still return the connected stats with the same `file`, the file must still exist on disk, and the factory must not have re-run (no dispose). A status-style gate here now fails loudly.

packages/telemetry/src/events.ts
filip131311: Fixed: the JSDoc sentence is reworded in state terms ("Not-connected preconditions emit no tool:fail — this event is where they are counted"), and the target-selection test comment likewise drops the "previously a plain Error" narration.

packages/skills/skills/argent-react-native-app-workflow/SKILL.md
filip131311: Fixed: §2.2 no longer claims any non-metro_not_running result means Metro answered — it now names the two Metro-probe reasons and defers every other reason to the result's `guidance` (kept to one sentence to avoid bloating the skill).

packages/tool-server/src/tools/debugger/not-connected.ts
filip131311: Fixed at the registry level: `invokeTool` now exposes the id it minted on `ctx.toolInvocationId`, so the event carries the join key on every dispatch path — attributed or not. Pinned in registry.test.ts (minted id matches the toolInvoked event; caller-provided id passes through); live-verified with a flow-execute'd debugger-status step and no ai-client header: outcome id == invoke/complete id.

packages/tool-server/src/tools/flows/flow-run.ts
filip131311: Fixed: the gate step's `reason` now embeds `detail` (so the device_mismatch pointer and the `got:` fragment survive) and the step carries the full structured `result` like a passing step would. Pinned in flow-debugger-gate.test.ts; live-verified in a flow report.

packages/skills/skills/argent-tv-interact/SKILL.md
filip131311: Fixed: the Vega note now maps a broken forward to `no_app_connected` (the /status probe is host-local, so metro_not_running can't name it) and tells the agent to check `vega device start-port-forwarding` before relaunching — the one Vega-specific fact the generic guidance can't carry.

packages/tool-server/src/tools/debugger/not-connected.ts
filip131311: Fixed: flow-debugger-gate.test.ts now drives a `debugger-log-registry` gate and asserts the fail mapping — narrowing the predicate to debugger-status alone goes red.

PR #611, opened 2026-07-31
fix(chromium): correct the hidden-window guard's docs and pin its exemptions

latekvo:
Follow-ups from a post-merge review of #605. Each one is a comment thread on that PR.

No behaviour changes to what the guard does — the three emitted `failure_stage` strings and the `CHROMIUM_WINDOW_HIDDEN` code are byte-for-byte identical, so telemetry stays joinable.

## Changes

**Derive `failure_stage` from `action`.** It was a second free-form `string` parameter sitting next to `action`, and all three call sites passed a strictly redundant pair — `("tap", "chromium_tap_window_hidden")` and so on. Two adjacent same-typed params meant transposing them or mistyping the stage compiled cleanly and read fine in review. That carries extra weight here because #605 deliberately froze `chromium_scroll_window_hidden` to keep pre- and post-migration rows joinable. `action` is now `"tap" | "drag" | "scroll"`, so a typo is a compile error, and the docblock spells out all three literals so they stay greppable in `src`.

**Pin the keyboard exemption.** Key events skip hit-testing and stay fast on a hidden window, so `keyboard` is deliberately unguarded. That was correct but lived only in a comment: adding the guard to `keyboard` would have made it refuse input that demonstrably works, and the suite would have stayed green. Now covered by `keyboard-chromium-unguarded.test.ts`.

**Pin the `button` exemption too.** The same file now asserts that `assertSupported` rejects a Chromium device for `button`. The claim that `button` can never reach the guard rested entirely on a capability literal no test read — adding `chromium: { app: true }` to it reddened nothing.

**Correct the same comment's claim about `button` — three times.** The first pass replaced the wrong rationale (that `button` skips the guard because key events bypass hit-testing) with the real one: its capability omits chromium. The replacement was itself wrong about where Chromium's button surface lives. `sendButton` is reached only from `handleWsCommand`, i.e. the chromium-server **WebSocket** channel — the HTTP router exposes screenshot / clipboard / fps / navigate / reload / history / viewport / stream and no button at all. It also over-claimed: `sendButton` emulates `Back` alone as an Alt+Left chord and throws for every other button, so "chromium hardware buttons exist" contradicted `button/index.ts` ("Chromium apps have no hardware buttons"). The wording now agrees with both neighbours. A third pass dropped the last false note in it: the lockout test said the gate stops `button` "before it can reach the guard", but `button` has no chromium branch and never imports the guard — there is no guard on its path to reach.

**Give the keyboard measurement its missing condition.** `Input.dispatchKeyEvent` at 1–14 ms against 5002–5005 ms per mouse move, and "focus emulation keeps input fast even for a genuinely minimized window", describe the same scenario with opposite outcomes — those numbers can only come from a window where focus emulation was not in effect. The docblock says so, and names focus emulation alone: `primePageSession` applies it to every session at connect, so the probe reads "hidden" only where the runtime refused it.

The anti-throttling launch flags are not part of that condition and do not narrow the guard's reach. A window carrying all three, minimized without emulation, still reads "hidden" and still costs ~5 s per `Input.dispatchMouseEvent` — Electron 42.5.0 5001–5005 ms and Chrome 152 5002–5010 ms, 3 trials each — so an argent-spawned app is no less exposed than an externally launched one. `boot-electron.ts`'s `ANTI_THROTTLING_ARGS` docblock, where that claim originated, is corrected to match, along with its test.

**Two mechanisms that were stated unconditionally.** `Page.bringToFront` un-minimizes on Chrome 152 (`windowState` minimized → normal, `visibilityState` hidden → visible, mouse dispatch 5002–5006 ms → 16–25 ms, 3/3) and not on Electron 42.5.0 (returns ok, stays `"hidden"`, stays 5001–5005 ms, 3/3), so `flow-run`'s comment calls the outcome runtime-dependent rather than denying it. And the capability gate that keeps `button` off a Chromium device is not universal: `assertSupported` runs at the HTTP edge and in run-sequence's pre-flight, but `Registry.invokeTool` skips it, so a flow step dies resolving `services()` with `SimulatorServer.factory does not support platform "chromium"` rather than an `UnsupportedOperationError`.

**Replace the try-boundary rationale.** It said the property access had to stay inside the `try` because fakes "may lack `cdp`" — but for such an api `chromium.cdp` yields `undefined` without throwing; the TypeError comes from reading `.send`, which is inside the `try` wherever `chromium.cdp` is read. Verified: hoisting `const client = chromium.cdp;` above the `try` leaves every relevant test green. The placement would only matter if `chromium` itself were nullish, which no registry path produces.

**Cover the no-`cdp` probe path directly.** The existing test claimed to cover "api objects with no `cdp` at all" but used a rejecting `cdp.send` — a different failure mode. The real shape was exercised only incidentally, by bare fakes in `chromium-drag.test.ts` and `tools/gesture-tap.test.ts`.

**Cover the other half of "a failed or empty read proves nothing".** Only the throw paths were pinned. Flipping the comparison to `value !== "visible"` left the entire suite green, so a probe that resolves without a usable value — `Runtime.evaluate` against a torn-down document — would have started refusing tap/drag/scroll on a fully visible window.

**Pin the action word in the user-facing message.** Hardcoding `Cannot tap:` for all three tools also stayed green, so a scroll refusal reading "Cannot tap" could have shipped unnoticed.

**Update `flow-run`'s comment.** It still named gesture-scroll as the only tool that fails fast on a hidden window; tap and drag have since #605. It also over-promised "fail fast": on the flow path `frontChromiumPage` has just primed focus emulation, so the three tools proceed. They carry the guard; whether it fires depends on the mitigations.

## Verification

- Full gate on the branch: `tsc --build`, `typecheck:tests`, `eslint --max-warnings 0`, `prettier --check`, and the whole tool-server suite sharded — **360 files / 4504 passed, 1 skipped, 0 failed**.
- Every empirical claim in the comments was re-measured on a live CDP testbed rather than carried over: an Electron 42.5.0 app and Chrome 152, each launched with and without the three `ANTI_THROTTLING_ARGS` plus `--remote-debugging-port`, minimized, probed for `document.visibilityState` and per-event `Input.dispatchMouseEvent` / `Input.dispatchKeyEvent` latency before any focus emulation, then after `Page.bringToFront`, then after `Emulation.setFocusEmulationEnabled`. Runs were sequential, one browser at a time, 3 trials per condition.
- The `button`-on-chromium flow path was run, not just read: `registry.invokeTool("button", { udid: "chromium-cdp-9222", button: "home" })` with the simulator-server blueprint registered throws `ToolExecutionError … SimulatorServer.factory does not support platform "chromium"`.
- The measurements behind the keyboard exemption were taken on a real minimized Electron window (v42.5.0): `Input.dispatchKeyEvent` 1–14 ms, against 5002–5005 ms for each of ten consecutive `Input.dispatchMouseEvent` moves, with `Runtime.evaluate` staying at 1–3 ms.
- Mutation-tested every pin rather than assuming it bites. Each of these fails the suite: adding the guard to `keyboard/chromium`; adding `chromium` to `button`'s capability; making the probe rethrow instead of swallowing; drifting the derived stage template; collapsing the stage to one action for all three tools; flipping the comparison to `value !== "visible"`; and hardcoding the action word in the message.

- No docs update needed: comments only, no user-facing behaviour or tool-surface change.

## Summary by CodeRabbit

- **Documentation**
  - Clarified how window minimization, focus emulation, and anti-throttling behavior affect Chromium interactions.

- **Bug Fixes**
  - Standardized hidden-window handling and error reporting for tap, drag, and scroll gestures.
  - Preserved keyboard input on hidden Chromium windows.

- **Tests**
  - Expanded coverage for hidden-window gestures, keyboard input, missing browser capabilities, and varied visibility responses.


latekvo (commented):

packages/tool-server/test/chromium-window-hidden-guard.test.ts
```diff
@@ -89,9 +89,9 @@ describe("hidden-window guard on chromium mouse tools", () => {
     expect(getFailureSignal(err)?.failure_stage).toBe("chromium_scroll_window_hidden");
```
latekvo, line 94: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The changed test file still carries the exact `button` claim this PR set out to correct.

The PR's stated purpose includes "Correct the same comment's claim about `button`" — and it does so in src/utils/chromium-visibility.ts:17-20 ("The `button` tool never reaches it either, for a different reason: its capability omits chromium entirely") and in the new test/keyboard-chromium-unguarded.test.ts:16-18. But the header comment of the very file this PR edits still states the refuted rationale: `button` is said to "skip the guard because key events bypass hit-testing". `button` has no chromium capability at all (packages/tool-server/src/tools/button/index.ts:49-53 declares only `apple`, `appleRemote`, `android`), so it can never reach a Chromium device and therefore never reaches the guard for any hit-testing-related reason; packages/tool-server/src/chromium-server/input.ts:151 says the same ("the registered `button` tool is apple+android only"). This is now the only remaining copy of the wrong explanation in the repo (`grep -rn "keyboard/button"` returns this line and nothing else), and it sits three lines above the assertions the PR touched.

**Trigger.** A maintainer reading chromium-window-hidden-guard.test.ts — the canonical test for this guard — is told `button` is exempt because key events are fast on a hidden window. Acting on that (e.g. adding `chromium: { app: true }` to the button tool's capability, believing the exemption rationale already covers it) routes button presses to a path that the guard's own docs say is exempt for a reason that does not apply, while the actual chromium button surface (chromium-server/input.ts sendButton) only supports `Back` and throws a plain Error for every other button.

**Checked against.** post-PR packages/tool-server/test/chromium-window-hidden-guard.test.ts:9-11 — "// (measured live on an Electron testbed). tap, drag and scroll refuse up\n// front with an actionable error; keyboard/button skip the guard because key\n// events bypass hit-testing and stay fast on hidden windows." (identical to the pre-PR text at 67a4481:packages/tool-server/test/chromium-window-hidden-guard.test.ts:9-11), against the corrected sibling prose the PR added in packages/tool-server/src/utils/chromium-visibility.ts:17-20 — "The `button` tool never reaches it either, for a different reason: its capability omits chromium entirely" — and packages/tool-server/src/tools/button/index.ts:49-53 — "const cap …

latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]\: Fixed in e57dc768f. The header comment in chromium-window-hidden-guard.test.ts now reads: "keyboard skips the guard because key events bypass hit-testing and stay fast on hidden windows, and button never reaches it - its capability omits chromium." Verified against button/index.ts, whose capability is apple/appleRemote/android only (no chromium), so the grouped keyboard/button claim is gone and the wording now matches the corrected chromium-visibility.ts body.

latekvo:
[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Full review pass on this branch — claims-vs-code, reachability both ways, absence pass, and mutation checks on every pin. Two gaps found, both fixed and re-verified in 3afa2f63:

1. **Scroll refusal ordering was unpinned.** The gesture-scroll refusal case asserted the error code and stage string but never that the tool refused *before* dispatching — its tap/drag siblings both do. Mutation check: making scroll send one wheel event ahead of the guard went red only after adding `expect(api.server.sendWheel).not.toHaveBeenCalled()`; it passed green before. Case also retitled to match its siblings.

2. **The keyboard exemption stopped below the platform impl.** Both cases drove `makeChromiumImpl(registry).handler(...)` directly; a guard added at the wrapper level (`createKeyboardTool().execute`, before `dispatchByPlatform`) would have shipped green since existing e2e keyboard tests use non-hidden fakes. Added a case driving `tool.execute` through the same hidden-window registry. Mutation check: a guard injected into the wrapper's execute (resolving via the closed-over registry) now fails exactly that case while the impl-level cases stay green — proving the new test closes the gap the old ones missed.

Dismissed without action: the chromium-server WS/HTTP `touch`/`wheel` commands (`handleWsCommand` → `sendTouch`/`sendWheel`) dispatch over CDP with no guard. Screened: nothing agent-reachable reaches them — no registered tool routes through that channel; it is mounted under `/chromium-server/:deviceId` for the human-operated preview remote control, where the operator sees the window state themselves. The guard's declared scope is the autonomous agent loop.

Everything else checked out: all three `failure_stage` strings byte-identical under the derived template and no external consumers beyond tests; every docblock claim traced to code (ANTI_THROTTLING_ARGS applied at the sole spawn path, `Emulation.setFocusEmulationEnabled` in primePageSession, Back-only WS button, 10s CDP client timeout); no tautological assertions; no unguarded agent-facing path to Chromium mouse input.

Note: `test/lens-tools-platform-gate.test.ts` intermittently hits its 5s timeout in full-suite runs — reproduced identically on the pristine tree, passes in isolation. Pre-existing flake, unrelated.

packages/tool-server/src/utils/chromium-visibility.ts
```diff
@@ -5,22 +5,41 @@ import type { ChromiumCdpApi } from "../blueprints/chromium-cdp";
...
- * The `chromium.cdp` access stays inside the try: test fakes and
- * mid-navigation contexts may lack `cdp`, which says nothing about visibility.
+ * It is a backstop, because two mitigations already cover the stall. Apps
+ * argent spawns carry ANTI_THROTTLING_ARGS, which keep the compositor awake and
+ * stop `visibilityState` flipping to "hidden" for as long as the app runs;
+ * independently, while a CDP session is attached, primePageSession's focus
+ * emulation pins reported visibility to "visible". The probe therefore reads
+ * "hidden" only where both are absent — an externally launched target on a
+ * runtime where emulation could not be applied — exactly where the stall is
+ * real and un-minimizing is the fix.
```
hubgan: This two-mitigation account does not survive contact with a device.

I spawned an Electron app (42.5.0) with exactly the three `ANTI_THROTTLING_ARGS` that `boot-electron.ts` applies, plus `--remote-debugging-port`, minimized it, and probed the session before any focus emulation was applied: `document.visibilityState` reads `"hidden"`, and each `Input.dispatchMouseEvent` takes 5011-5262 ms. The same app launched *without* the flags behaves identically (`"hidden"`, 5248-7931 ms), and the same A/B on Chrome 151 is also identical (5002-5003 ms vs 5014-5302 ms). The flags therefore neither keep the compositor awake for mouse hit-testing nor stop `visibilityState` flipping to `"hidden"`. There is one mitigation here, not two.

That makes the conclusion wrong in both halves. "only where both are absent" is not the condition, because flag presence is irrelevant. And "an externally launched target" is not required: the app in that measurement is one argent spawned, carrying those very flags. Running this branch's `assertChromiumWindowVisible` against that live session refuses all three actions with the correct stage strings.

The mitigation that does carry the weight is the other one. Applying `Emulation.setFocusEmulationEnabled` to that same minimized window flips the probe to `"visible"` and mouse dispatch back to 5-31 ms, and `createChromiumServer` applies it to every session at connect regardless of how the app was launched, so launch origin never enters into it.

What concerns me about leaving this in place is the use a future reader will make of it: it reads as a licence to drop the guard as unreachable on argent-spawned apps, and it is reachable there. Worth noting the same claim exists upstream in `boot-electron.ts` (not touched by this PR), which appears to be where it was picked up.

packages/tool-server/src/utils/chromium-visibility.ts
```diff
@@ -5,22 +5,41 @@ import type { ChromiumCdpApi } from "../blueprints/chromium-cdp";
...
+ * loop keeps working blind.
  *
- * Only an explicit "hidden" refuses — a failed or empty read proves nothing.
- * primePageSession's focus emulation pins reported visibility to "visible"
- * while a session is attached, so the probe reads "hidden" only where emulation
- * could not be applied — exactly where the stall is real and un-minimizing is
- * the fix.
+ * Key events bypass hit-testing and stay fast on a hidden window — measured on
+ * a minimized Electron window carrying neither mitigation below,
+ * Input.dispatchKeyEvent 1–14ms against 5002–5005ms per mouse move — so the
```
hubgan, line 12: "carrying neither mitigation below" presents the absence of the anti-throttling flags as a controlled condition of the measurement, which implies a contrast that is not there. On a minimized Electron window with those flags installed I measured `Input.dispatchKeyEvent` at 1-62 ms against mouse moves at 5014-5042 ms, essentially the numbers quoted here. The flags are not part of what produced the contrast; only the absence of focus emulation is.

The key-versus-mouse split itself, which is the load-bearing claim, reproduces exactly.

hubgan (commented):
Reviewed the diff and drove the guard against real Chromium targets: Electron 42.5.0 spawned with the exact `ANTI_THROTTLING_ARGS`, and Chrome 151 launched externally, each probed visible / minimized / occluded, with and without focus emulation.

The code change itself is clean. All three call sites are converted, the derived stage strings are byte-identical to the ones telemetry is keyed on (confirmed end to end against a live minimized window), the merge with current `main` type-checks, and every new test pin genuinely bites under mutation. The keyboard exemption is correct on a device: on a minimized window where `gesture-tap` refuses, `keyboard` typed "hi" into the focused field in 40 ms and the value landed.

The findings below are all about the accuracy of the new prose, which is this PR's own subject. None of them block: no behaviour changes, and the guard does not misfire on argent's real path.

packages/tool-server/test/chromium-window-hidden-guard.test.ts
```diff
@@ -7,12 +7,13 @@ import { gestureScrollTool } from "../src/tools/gesture-scroll";
...
-// visibility to "visible" while a session is attached, so the probe reads
-// "hidden" only on sessions where emulation could not be applied — exactly
-// where the stall is real.
+// front with an actionable error; keyboard skips the guard because key events
+// bypass hit-testing and stay fast on hidden windows, and button never reaches
+// it — its capability omits chromium. In practice the guard is a backstop:
+// argent-spawned apps carry anti-throttling flags, and focus emulation pins
+// reported visibility to "visible" on every session that could apply it, so the
+// probe reads "hidden" only on an externally launched target whose runtime
+// could not — exactly where the stall is real.
```
hubgan: Same account as the one on `assertChromiumWindowVisible`, and it carries the same two problems: the anti-throttling flags are named as a mitigation for this stall when a minimized window carrying them still reads `"hidden"` and still stalls at ~5 s per mouse event, and "only on an externally launched target" is contradicted by an argent-spawned app in exactly that state.

Worth flagging separately from the docblock because of where it sits: the fakes in this file hard-code `visibility`, so nothing in the suite can catch this header drifting from the runtime it describes. That is presumably how it drifted.

packages/tool-server/test/keyboard-chromium-unguarded.test.ts
```diff
@@ -0,0 +1,100 @@
...
+import { createKeyboardTool } from "../src/tools/keyboard";
+import { buttonTool } from "../src/tools/button";
+import { assertSupported, UnsupportedOperationError } from "../src/utils/capability";
+
+// The mouse tools (gesture-tap/-drag/-scroll) refuse up front on a hidden
+// Chromium window because every mouse dispatch waits on compositor hit-testing.
+// Key events skip hit-testing and stay fast on the same window — measured on a
+// minimized Electron window carrying neither of the mitigations described on
+// assertChromiumWindowVisible, Input.dispatchKeyEvent returned in 1-14ms while
+// ten consecutive mouse moves cost 5002-5005ms each — so `keyboard` is
```
hubgan: This one is framed as methodology, which makes the inaccuracy costlier than the same wording in the docblock. "carrying neither of the mitigations described on assertChromiumWindowVisible" tells a reader the flag-free state was a necessary condition of the observation. It was not: I reproduced 1-62 ms key events against 5014-5042 ms mouse moves on a minimized Electron window that *did* carry the anti-throttling flags.

The practical cost is that someone who later measures against an argent-spawned app, which always carries the flags, would expect these numbers not to hold and could conclude the exemption no longer applies. The only mitigation whose absence matters is focus emulation.

packages/tool-server/test/keyboard-chromium-unguarded.test.ts
```diff
@@ -0,0 +1,100 @@
...
+// deliberately NOT guarded. These tests are what hold it that way: guarding
+// `keyboard` would make it refuse input that demonstrably works.
+//
+// The sibling `button` tool is exempt for a different reason, pinned below: it
+// declares no chromium capability, so the gate rejects a Chromium device and
+// `execute` never runs — there is no guard on its path to reach. A Chromium app
```
hubgan: "the gate rejects a Chromium device and `execute` never runs" is stated unconditionally, but the capability gate is not universal. `assertSupported` runs at the HTTP edge and in run-sequence's explicit pre-flight; `Registry.invokeTool` does not call it, and `button` uses a hand-written `execute` rather than `dispatchByPlatform`, so it never picks the gate up that way either.

On the flow path, which dispatches through `invokeSubTool` to `registry.invokeTool`, a `tool: button` step against a Chromium device is not rejected by the gate at all. It dies earlier, during `services()` resolution, with `SimulatorServer.factory does not support platform "chromium"` rather than the clean `UnsupportedOperationError` this comment promises. The in-tree comment at `run-sequence/index.ts:182` already describes this exact scenario, naming this exact tool and platform.

The conclusion the paragraph is actually making, that there is no guard on `button`'s path to reach, holds regardless. It is the stated mechanism that overreaches, which is the same class of thing the rest of this PR is removing.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -1197,7 +1197,10 @@ Pass exactly one flow source: name for a saved flow under project_root, or flow_
       // throttle rendering — wheel-event acks (scroll steps) stall on a throttled
       // compositor. Covers the instance the run starts on; a launch that boots
       // one fronts it itself. Best-effort: bringToFront can focus a page but
-      // cannot unhide a minimized window (gesture-scroll fails fast on that).
+      // cannot unhide a minimized window — resolving the session applies focus
```
hubgan: "cannot unhide a minimized window" is platform-split, and the sentence states it unconditionally.

On Electron it holds: `Page.bringToFront` returns ok, the window stays minimized, `visibilityState` stays `"hidden"`, and a mouse move still costs 5093 ms. On Chrome 151 it is false, 3 trials out of 3: `windowState` goes `minimized` to `normal`, `visibilityState` goes `"hidden"` to `"visible"`, and mouse dispatch drops from ~5010 ms to 0-2 ms.

A Chromium-family browser on `--remote-debugging-port` is a first-class `platform: "chromium"` device, so this covers half of what the branch it annotates will actually meet. It also undersells `frontChromiumPage`: on a Chrome target that call really does resolve the condition the guard exists for.

The wording predates this PR, but the clause was rewritten here and a new claim was appended to it, so it is now load-bearing. The final clause of the same comment, conditioning purely on whether emulation could apply, is the one accurate statement of the guard's scope anywhere in the diff.

packages/tool-server/src/utils/chromium-visibility.ts
hubgan: Correcting myself on one sentence above: "There is one mitigation here, not two" over-generalises, and I should not have written it.

Everything I measured was on a **minimized** window, and for that state the finding stands unchanged: with all three flags applied, minimized, no emulation, the probe reads `"hidden"` and mouse dispatch sits at ~5 s (Electron 5011-5262 ms, Chrome 5014-5302 ms), indistinguishable from the same runtimes without the flags. So the docblock's unconditional "for as long as the app runs", and the conclusion that only an externally launched target can read hidden, are still contradicted by an argent-spawned app carrying those flags.

What I did not establish is the **occlusion** half. `--disable-backgrounding-occluded-windows` is aimed precisely at that case, and there is reason to think it does prevent an occluded (non-minimized) window from going hidden. I could not test it cleanly on this machine - my occlusion runs were contaminated, with the supposedly-unoccluded reference already reading `"hidden"` - so I am not asserting a result either way.

If the flags do cover occlusion, then the accurate statement is narrower than mine: the flags are a real mitigation for one of the two states this guard names and do nothing for the other, so the claim is over-broad rather than simply false. That distinction matters for whatever wording replaces it, which is why I would rather flag it than leave my sentence standing.

hubgan (approved):
Code comments are 🔥

packages/tool-server/src/utils/chromium-visibility.ts
latekvo: Fixed in ac842bb1e9beadf0c2f0458c4464ed0ff7c73fb0 - the two-mitigation paragraph is gone. Reachability now hangs on focus emulation alone: `primePageSession` applies it to every session at connect and `createChromiumServer` is the only path that builds a `ChromiumCdpApi`, so the probe reads "hidden" only where the runtime refused it. The flags appear once, as the trap they are: "Launch origin does not narrow that: an app carrying all three ANTI_THROTTLING_ARGS, minimized without emulation, still reads `"hidden"` and still costs ~5s per mouse event (measured on Electron 42 and Chrome 152)."

I reproduced that rather than taking it on report. Electron 42.5.0 with all three flags, window minimized (`win.isMinimized()` logged true), 3/3 trials: `"hidden"`, `Input.dispatchMouseEvent` 5001-5005 ms, `Input.dispatchKeyEvent` 0-3 ms. Chrome 152 with the flags, minimized via `Browser.setWindowBounds`, 3/3: `"hidden"`, 5002-5010 ms. `Emulation.setFocusEmulationEnabled` on the same minimized window flips both to `"visible"` with 0-17 ms mouse dispatch on either runtime.

Per your follow-up, nothing in the replacement asserts an occlusion result in either direction.

The upstream copy is fixed too. `ANTI_THROTTLING_ARGS` in `boot-electron.ts` no longer claims to keep an app responsive while minimized, and "Externally launched CDP targets are unaffected" is dropped; it now states what the switches disable, then "They do not reach minimization: a window carrying all three still reads `"hidden"` and still costs ~5s per Input.dispatchMouseEvent once minimized (measured on Electron 42 and Chrome 152)." The same claim in the anti-throttling case of `boot-electron-spawn-error.test.ts` is corrected as well.

One loose end I am flagging rather than acting on: my Electron A/B did not match yours. *Without* the flags my minimized Electron 42.5.0 window read `"visible"` with 1-24 ms mouse dispatch, 3/3 - so on my testbed the flags are what made it read hidden and stall. The Chrome 152 A/B showed no such inversion (identical with and without). I have no mechanism for it and it contradicts your numbers, so no prose depends on it either way; it is one more reason the wording now claims nothing about what the flags do or do not fix.

packages/tool-server/src/utils/chromium-visibility.ts
latekvo: Fixed in ac842bb1e9beadf0c2f0458c4464ed0ff7c73fb0 - "carrying neither mitigation below" is now "with no focus emulation applied", which is the only condition that produced the contrast.

Confirmed on my own testbed before rewriting: minimized Electron 42.5.0 *with* all three anti-throttling flags, `Input.dispatchKeyEvent` 0-3 ms against mouse moves at 5001-5005 ms (3 trials); Chrome 152 with the flags, 0-3 ms against 5002-5010 ms (3 trials). The key-versus-mouse split reproduces exactly as you found, and the flags have nothing to do with it.

The quoted 1-14ms / 5002-5005ms are left as they stand - they sit inside every range measured on either runtime.

packages/tool-server/test/chromium-window-hidden-guard.test.ts
latekvo: Fixed in ac842bb1e9beadf0c2f0458c4464ed0ff7c73fb0 - same correction as the docblock. The header now reads "focus emulation, applied to every session at connect, pins reported visibility to `"visible"`, so the probe reads `"hidden"` only where the runtime refused it", with no flags clause and no externally-launched clause.

On the testability point, you are right and it cannot be papered over: the fakes hand the guard a hard-coded string, so no assertion in this file can notice a real minimized window behaving differently. Rather than invent a test that would only pin the fake, the header says so outright - "The fakes below hard-code `visibility`, so what they pin is the refusal, not the reading that provokes it."

The structural half that reachability actually turns on is pinned, and elsewhere: `chromium-cdp-blueprint.test.ts` drives `chromiumCdpBlueprint.factory` against a fake CDP server and asserts `Emulation.setFocusEmulationEnabled` was sent, and `createChromiumServer` is the only path that builds a `ChromiumCdpApi`, so every api the guard can be handed has been through `primePageSession`.

packages/tool-server/test/keyboard-chromium-unguarded.test.ts
latekvo: Fixed in ac842bb1e9beadf0c2f0458c4464ed0ff7c73fb0 - "carrying neither of the mitigations described on assertChromiumWindowVisible" is now "with no focus emulation applied", so the header no longer presents the flag-free state as a condition of the measurement.

Checked the same way as the docblock claim, on a window that *did* carry the flags: minimized Electron 42.5.0 with all three, key dispatch 0-3 ms against mouse moves 5001-5005 ms, 3 trials; Chrome 152 with the flags, 0-3 ms against 5002-5010 ms, 3 trials. Someone re-measuring against an argent-spawned app will now expect exactly what they get.

packages/tool-server/test/keyboard-chromium-unguarded.test.ts
latekvo: Fixed in ac842bb1e9beadf0c2f0458c4464ed0ff7c73fb0 - the paragraph keeps its conclusion and drops the mechanism: "it declares no chromium capability, so there is no guard on its path to reach. What enforces that varies: `assertSupported` runs at the HTTP edge and in run-sequence's pre-flight, while a flow reaches `Registry.invokeTool`, which skips it and fails earlier still, resolving `services()` against the simulator-server factory."

Checked against source and then run. `Registry.invokeTool` calls `definition.services(...)` before `execute` and never calls `assertSupported`; `button` has a hand-written `execute`, so it does not pick the gate up through `dispatchByPlatform` (which is where that call lives); its `services()` returns `simulatorServerRef(device)` for anything non-Android. A throwaway test driving `registry.invokeTool("button", { udid: "chromium-cdp-9222", button: "home" })` with the simulator-server blueprint registered produced exactly what you described:

```
ToolExecutionError :: [Tool:button] Service dependency failed: [SimulatorServer:chromium-cdp-9222] SimulatorServer.factory does not support platform "chromium". Use the platform-specific service blueprint instead.
```

not an `UnsupportedOperationError`. The test below is unchanged: it asserts `assertSupported` throws, which is accurate for the path it covers.

packages/tool-server/src/tools/flows/flow-run.ts
latekvo: Fixed in ac842bb1e9beadf0c2f0458c4464ed0ff7c73fb0 - the clause is now "whether bringToFront un-minimizes is runtime-dependent (measured: Chrome restores the window and unthrottles input, Electron leaves it minimized and hidden)". The final clause, which you called the one accurate statement of the guard's scope, is kept as it was.

Reproduced the split before rewriting. Chrome 152, minimized via `Browser.setWindowBounds`, 3/3 trials: `Page.bringToFront` takes `windowState` minimized to normal, `visibilityState` hidden to visible, and mouse dispatch from 5002-5006 ms to 16-25 ms. Electron 42.5.0 carrying the anti-throttling flags, minimized, 3/3: `bringToFront` returns ok, `visibilityState` stays `"hidden"`, mouse stays 5001-5005 ms.

PR #613, opened 2026-07-31
fix(tool-server): derive hidden-window failure_stage, pin keyboard guard exemption

filip131311:
Follow-up to #605, addressing the two actionable findings from the post-merge review.

## 1. Derive `failure_stage` instead of hand-writing it per call site

`assertChromiumWindowVisible` took adjacent `(action, failureStage)` string params where every call site passed a strictly redundant pair — the stage is always `chromium_<action>_window_hidden`. A typo or transposed argument compiled cleanly and read fine in review; the only symptom would be telemetry landing in an unqueried bucket, breaking exactly the pre/post-migration joinability #605 froze `chromium_scroll_window_hidden` to preserve.

The helper now takes a typed `"tap" | "drag" | "scroll"` action and derives the stage internally, so a drifting stage string can no longer compile. The existing guard tests assert the exact per-tool stage strings at the tool boundary, which pins the derivation (including the frozen `chromium_scroll_window_hidden`).

## 2. Pin the keyboard exemption with a test

The keyboard/button exemption (key events skip compositor hit-testing and stay fast on hidden windows) was enforced only in prose — mutation testing in the review showed adding the guard to those tools would keep the suite green. New test: an api whose visibility probe **would** report `"hidden"` must still type, and the probe must never even be consulted. `button` has no chromium branch at all, so keyboard is the only tool to pin.

**Mutation-verified:** temporarily adding `assertChromiumWindowVisible` to the keyboard chromium impl fails the new test; reverting restores green.

## Verification

- Unit: 3087 tool-server + 87 registry tests green, `typecheck:tests` green, lint + prettier clean on changed files.
- Live (dev tool-server from this branch, Electron testbed, genuinely minimized windows):
  - keyboard types in **~37 ms**, tap lands in **~55–73 ms** on both an argent-booted (flagged) app and an external un-flagged app on `:9222` — no guard false-positives.
  - Confirmed the guard cannot fire while a tool-server session is attached (its focus emulation pins reported visibility even after an external session disables emulation and the window is re-minimized) — consistent with #605's backstop design; the refusal path and derived stage strings are pinned by the unit tests.

### DIFF
--- a/packages/tool-server/src/tools/gesture-drag/index.ts
+++ b/packages/tool-server/src/tools/gesture-drag/index.ts
@@ -61,7 +61,7 @@ Returns { dragged: true, timestampMs }. Fails if the Chromium CDP session is not
     // A drag interpolates ~60fps mouse moves; on a hidden (throttled) window
     // each one stalls on compositor hit-testing (~5s), turning one drag into
     // minutes of wall clock — refuse up front like gesture-scroll.
-    await assertChromiumWindowVisible(chromium, "drag", "chromium_drag_window_hidden");
+    await assertChromiumWindowVisible(chromium, "drag");
     const vp = chromium.getViewport();
     const startPx = { x: params.fromX * vp.width, y: params.fromY * vp.height };
     const endPx = { x: params.toX * vp.width, y: params.toY * vp.height };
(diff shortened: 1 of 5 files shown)


latekvo:
Possibly a duplicate of #611

PR #652, opened 2026-07-31
fix(deps): name the platform a package-manager command belongs to

### LINKED ISSUE #641: Linux: device-resolution errors tell the user to `brew install` Android platform-tools
## Symptom

On Linux, calling a device tool without adb available returns:

> `Android SDK Platform Tools not found. Install with 'brew install --cask android-platform-tools'…`

Homebrew is not the relevant package manager there.

## Otherwise good

The error is a single line, exits 1, has no stack trace, and does mention `$ANDROID_HOME` — so only the
remediation clause needs to be platform-aware.

## Suggested fix

Pick the hint by host platform (apt/dnf/pacman or the SDK-manager path on Linux, brew on macOS), or drop
the package-manager specifics and point at `$ANDROID_HOME` plus the official platform-tools download.

---

filip131311:
Fixes #641.

Reproduced on a real Linux VM (Ubuntu aarch64, published `0.18.0`, no `adb`):

```
$ argent run boot-device --avdName Pixel_9
Android SDK Platform Tools not found. Install with `brew install --cask android-platform-tools` …
```

## What changed, and why not the obvious thing

My first instinct — and the issue's second suggestion — was to delete the package-manager specifics
entirely. Two things killed that:

- **`sdkmanager` ships *inside* the SDK command-line tools.** A reader with no SDK cannot run it, so
  it can't be the headline replacement. It's kept, but qualified with "if you already have the SDK
  command-line tools".
- **Removing brew would have been a real regression for macOS.** `adb` gates every USB-phone and
  already-running-emulator path (`launch-app`, `keyboard`, `settings-permissions`, `tv-remote`,
  `reinstall-app`, `open-url` on Android) — none of which need the emulator. For those,
  `brew install --cask android-platform-tools` genuinely *is* a complete one-command fix. Dropping it
  would have traded Linux's wrong advice for macOS's missing advice.

So the rule is **labelled examples, no host branching**: every hint leads with a route that works
anywhere (SDK Manager, standalone download), and any package-manager command names its platform.
No `process.platform` read, no mechanism, nothing new to test per host — and Windows is fixed too,
which a darwin/non-darwin split would have missed.

## Also in scope

- **`ffmpeg` (`capture.ts`)** had the identical defect and is the *more* reachable one on Linux, where
  ffmpeg is the encoder for Android recording. Same treatment. It additionally warns that the build
  needs **libx264**: `capture.ts` hardcodes `-c:v libx264` with no fallback, and Fedora's default
  `ffmpeg-free` package omits it — so that install "succeeds" and then dies with an exit code and no
  mention of a codec.
- **`xcrun`** now leads with the disqualifier. `classifyDevice` is purely shape-based, so any
  8-4-4-4-12 udid routes to the iOS branch on *any* host; a Linux reader was being told to run
  `xcode-select --install` before being told iOS needs a macOS host. (Making every iOS tool fail
  up front on a non-macOS host is the real fix for that class — it's a behaviour change with its own
  failure code, so I left it out; happy to file it.)
- **One SDK root added.** Following the `sdkmanager` route left the tool *still* failing: the
  Homebrew command-line-tools cask roots the SDK at `/opt/homebrew/share/android-commandlinetools`,
  which wasn't probed. A remediation string that doesn't unblock isn't fixed.
- **`skills/argent-screen-recording/SKILL.md`** quoted the old `brew install ffmpeg` — updated so the
  shipped skill and the runtime error agree.

Not touched: `watermark.ts`'s `/opt/homebrew/bin/ffmpeg` and the existing Android roots. Those are
*search paths* — probing a Homebrew prefix on a machine without Homebrew simply misses.

## Verification

**On the Linux VM:** the exact repro now returns the labelled hint. Then, satisfying it, the dep gate
**advances to the next dependency** (`emulator`) instead of repeating — i.e. following the hint
actually unblocks `adb`, which is the only thing that makes a remediation string correct.

One honest caveat from that exercise: Google's standalone Linux platform-tools zip is **x86-64 only**,
so on this aarch64 VM the downloaded binary can't execute. That's a limitation of the upstream
download, not of this change, and the SDK-Manager route is unaffected — but it means the
"download-and-run" leg was verified by resolution, not by executing Google's binary.

**Tests:** all existing assertions pass **unchanged** — keeping the labelled brew cask means the
`/android-platform-tools/` assertions in `check-deps.test.ts` and `http-dep-gate.test.ts` never needed
relaxing. Added a guard that states the actual invariant rather than banning one word: every hint that
names `brew`/`apt`/`dnf`/`pacman` must also name the matching platform, every Android hint must offer
a platform-agnostic route and `$ANDROID_HOME`, and the `xcrun` hint must mention the macOS-host
requirement *before* `xcode-select`. Banning the string `brew` would have passed `apt install adb` —
the same bug — while failing the correct fix.

tool-server 3092 passed.

---

> **#680 is stacked on this branch** — it rewrites the same ffmpeg message and SKILL bullet, so it was rebased on top and absorbed the per-platform hints added here. Merge this first.

---

**Update 2026-09-17:** merged `origin/main` (no rebase). One conflict in `check-deps.ts`: #1173 had added `android.sdkRoot` to the `adb`/`emulator` hints and #931 shortened the comment above them. The platform-labelled `adb` hint now keeps main's `point $ANDROID_HOME (or the android.sdkRoot config key) at the SDK root … <root>/platform-tools/adb` wording. The problem is still live on main: its `adb` hint is still an unlabelled `brew install --cask android-platform-tools`, and the ffmpeg hint is still `brew install ffmpeg`. Checked again against a tool-server built from this branch, with `adb` hidden (clean env, fake `HOME`, no `ANDROID_HOME`): `launch-app` and `boot-device` on main return the unlabelled brew hint, and on this branch they return the labelled one. The tool-server suite passes (one `http-dep-gate` case failed once under load with an HTTP parse error, then passed when run alone). `typecheck:tests`, knip and prettier are clean.


packages/tool-server/src/utils/check-deps.ts
```diff
@@ -34,11 +34,16 @@ const cache = new Map<ToolDependency, CacheEntry>();
...
+    "Xcode command-line tools not found — iOS/tvOS simulators need a macOS host with them installed. On macOS, run `xcode-select --install` (or install Xcode from the App Store) and retry.",
   "adb":
-    "Android SDK Platform Tools not found. Install with `brew install --cask android-platform-tools` or via Android Studio → SDK Manager. If installed, ensure `adb` is on PATH or set `$ANDROID_HOME` to the SDK root (the resolver checks `$ANDROID_HOME/platform-tools/adb`). Only required for Android devices and emulators.",
+    "Android SDK Platform Tools not found. Install `platform-tools` via Android Studio → SDK Manager, the standalone download at https://developer.android.com/tools/releases/platform-tools, `sdkmanager 'platform-tools'` if you already have the SDK command-line tools, or `brew install --cask android-platform-tools` on macOS. If installed, ensure `adb` is on PATH or set `$ANDROID_HOME` to the SDK root (the resolver checks `$ANDROID_HOME/platform-tools/adb`). Only required for Android devices and emulators.",
```
latekvo, line 46: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: On Linux aarch64 — the host issue #641 was reproduced on — none of these routes yields a usable adb. The standalone download and the SDK command-line tools both ship only x86-64 Linux binaries (the Homebrew cask definitions carry a single Linux zip with no arm64 variant — the arm64_linux and x86_64_linux sha256 entries are identical — and the PR body notes the downloaded binary "can't execute" on that VM); Android Studio's Linux build is x86-64-only too, and brew is labelled macOS-only. An agent following this hint on that host is therefore left with no working command, even though a Linux distro route exists (the one the sibling ffmpeg hint in this same PR adds for ffmpeg). The "following the hint actually unblocks adb" check was by file resolution only, not by running the binary.

latekvo, line 46: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: `sdkmanager 'platform-tools'` is written in POSIX shell syntax: on Windows cmd.exe (the default shell for most Windows automation) single quotes are literal characters, so running this verbatim passes the token `'platform-tools'` (quotes included) to sdkmanager, which finds no such package and errors. That is the same class of defect this PR is fixing — a remediation command that only works on the author's platform — and it sits against the "Windows is fixed too" claim. PowerShell treats single quotes as delimiters and the SDK Manager / standalone routes do work on Windows, so the consequence is one broken route rather than a full blockage; the `sdkmanager 'emulator'` on the next hint has the same shape.

packages/tool-server/src/utils/check-deps.ts
```diff
@@ -34,11 +34,16 @@ const cache = new Map<ToolDependency, CacheEntry>();
...
   "emulator":
     "Android Emulator not found. Install via Android Studio → SDK Manager → Emulator, or `sdkmanager 'emulator'`. If installed, ensure `emulator` is on PATH or set `$ANDROID_HOME` to the SDK root (the resolver checks `$ANDROID_HOME/emulator/emulator`). Only required to launch new Android emulators via `boot-device`.",
```
latekvo, line 48: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: The `sdkmanager` route here is unqualified, while the adb hint directly above qualifies it with "if you already have the SDK command-line tools". On a host that has adb but not the SDK command-line tools — e.g. after `brew install --cask android-platform-tools`, which puts adb on PATH but ships no sdkmanager — an agent following this hint runs `sdkmanager 'emulator'` and gets `command not found`, the same failure mode this PR is meant to eliminate.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: Ran the full sweep: claims vs code, nearest twin, non-happy paths, inputs, reachability both ways, and what outlives the call, plus the absence pass (sibling/prose/symmetry and a mutation of the new constants and hint labels). Verified the two new SDK roots against the live Homebrew cask definitions and each remediation string against the resolver it feeds; ran check-deps.test.ts and android-binary.test.ts (23/23) and mutated the new root strings to confirm nothing pins them. Five findings, all inline below.

packages/tool-server/src/utils/android-binary.ts
```diff
@@ -135,7 +135,12 @@ function defaultAndroidRoots(): string[] {
     // System-wide locations (Linux package managers, Homebrew on macOS).
     "/opt/android-sdk",
     "/usr/lib/android-sdk", // Debian/Ubuntu `android-sdk` apt package
-    "/usr/local/share/android-sdk", // Homebrew cask
+    "/usr/local/share/android-sdk", // Homebrew cask (Intel prefix)
+    // Where `brew install --cask android-commandlinetools` roots the SDK, so an
+    // `sdkmanager`-installed platform-tools is found without setting ANDROID_HOME
+    // — otherwise following the missing-adb hint still leaves the tool failing.
+    "/opt/homebrew/share/android-commandlinetools",
```
latekvo, line 142: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: These two root strings are what make following the sdkmanager route in the adb hint actually unblock, but no test pins them. android-binary.test.ts exercises only the HOME-derived roots and the env-var roots; a typo or regression in either literal leaves the suite green (verified by breaking both strings and re-running the tests — 23/23 still pass), so the unblocking this PR adds would silently stop working with no test to flag it.

packages/tool-server/src/tools/screen-recording/capture.ts
```diff
@@ -292,7 +292,7 @@ async function startCaptureLocked(
   const ffmpeg = await resolveFfmpeg();
   if (!ffmpeg) {
     throw new FailureError(
-      "`ffmpeg` was not found on PATH. Install it (e.g. `brew install ffmpeg`) to record the screen.",
+      "`ffmpeg` was not found on PATH. Install it with your system package manager (`brew install ffmpeg` on macOS, `apt install ffmpeg` on Debian/Ubuntu; on Fedora use RPM Fusion's `ffmpeg`, since the default `ffmpeg-free` build has no libx264) or see https://ffmpeg.org/download.html, then retry.",
```
latekvo, line 295: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: This libx264 / RPM Fusion guidance lives only in the "ffmpeg not found" branch, so the Fedora reader it targets never sees it: a user who already has ffmpeg-free installed (the default Fedora package) has ffmpeg on PATH, resolveFfmpeg() succeeds, and the failure they actually hit is the encode-time "Unknown encoder 'libx264'" exit, which carries no pointer to RPM Fusion. An agent seeing that is likely to retry `dnf install ffmpeg`, which reinstalls ffmpeg-free — the same libx264-less build — and loops. The SKILL.md bullet documents the requirement, but the runtime error the agent acts on does not.

PR #657, opened 2026-07-31
fix(installer): only rewrite the skills store the updated install owns

### LINKED ISSUE #637: argent update --local also rewrites the global skills store
## Symptom

`argent update --local` inside a project reports:

```
+ project: synced 15
+ global:  synced 15
```

and the global skills store's mtimes confirm it was rewritten.

## Why it matters

It was harmless in this run only because the content was identical to the tarball's. With a global
install at a different version than the project-local one — precisely the case `--local` exists to
support — a project-scoped update would silently rewrite the skills every other project on the machine
uses.

## Expected

`--local` touches project scope only; global skills are updated by `argent update --global`.

## Related

The dangling-symlink half of the same subsystem (`uninstall --local` leaving `.claude/skills/` symlinks
pointing at deleted `.agents/skills/*`) is noted in the uninstall issue. Also related historically:
#195, #308.

---

filip131311:
## Update 2026-09-17: merged with main

- Merged `origin/main` (no rebase). Two conflicts, both with the comment trim from #931: the `install-runner.ts` init-triggered refresh and the `refreshArgentSkills` header in `skills.ts`. I kept the scoped calls and main's shorter comments. The installer files are the same in v0.25.1 and main, so the bug is still there on main.
- Replaced the hand-rolled `compareVersions` with `semver`, which the installer already depends on. The old helper ignored prerelease tags.
- No docs change: no CLI flag, config key or tool changed.

Live check with a packed tgz. HOME, npm prefix and PATH all point into a sandbox, and a stub `npx skills` edits the lock files the way the real one does. The project is local-mode. The global store holds a foreign skill, the bundled argent skills at `v0.99.0`, and `argent-future-skill@v0.99.0`. The project store holds a foreign skill and `argent-retired-skill@v0.24.0`. The control is npm `0.25.1`, whose installer is the same as main's:

| run | main (0.25.1) | this branch |
|---|---|---|
| `update --local --yes` | also syncs `-g`, re-pins the global store to v0.25.1, and **deletes `argent-future-skill`** | project only (sync + prune `argent-retired-skill`); global store unchanged; prints the "Left the machine-wide skills alone" hint |
| `update --global --yes` (global store newer) | re-syncs and prunes the project lock; **deletes `argent-future-skill`** | syncs global only; no prune; project lock unchanged |
| `update --global --yes` (global store at v0.25.0 with `argent-retired-global`) | prunes | still prunes `argent-retired-global`, so an owned store still refreshes |

Foreign skills survived in every run. Checks: installer vitest 587/587, `typecheck:tests`, `npm run build`, eslint, prettier, knip.

Fixes #637. See [my triage comment there](https://github.com/software-mansion/argent/issues/637#issuecomment-5147964114)
for the full findings — this PR fixes two defects, and the reported one is the milder.

## 1. The reported bug

`refreshArgentSkills(projectRoot)` took no scope and looped both specs, so `argent update --local`
rewrote the machine-wide store. Reproduced against the shipped code with a sandboxed `HOME` and a
stub `npx`:

```
scopes acted on: project(synced=15), global(synced=15)
npx … skills add …#v0.18.1 --skill * -y
npx … skills add …#v0.18.1 --skill * -y -g     ← global, from a --local update
```

Callers now name the stores their update owns, and the scope is **required** — a call site that
forgets one is a compile error rather than a silent write to a shared store. (Both existing call
sites did forget; the compiler caught them.)

This also closes the **mirror** case: `argent update --global` no longer rewrites a project's
*committed* `skills-lock.json` when that project runs a local install.

One trap worth naming: "is there a local install" is the wrong predicate. `chooseScope` returns
`local` for non-interactive **global**-mode installs too, and `init-skills.ts` only passes `-g` for
an explicitly global scope — so a project lock can legitimately track the global binary. The rule
keys on `resolveInstallMode`.

## 2. The worse bug the fix had to also close

Narrowing scopes alone would have closed this issue while leaving a **deletion** path open.

Obsolete skills are computed as the difference against the **running** package's bundled set, and the
global spec's removal carries `-g`. So an install older than the one that filled a store classifies
everything added since as orphaned and deletes it machine-wide. The skew is real — `v0.15.0` ships 13
skills, current ships 15. Reproduced:

```
npx --force skills remove -y -g argent-screen-recording argent-settings-permissions
```

And scope-narrowing does **not** stop it: any run whose targets include `global` still prunes from the
running package's set — including plain `argent update --yes` on a machine with both installs, and the
tool-server's own `update-argent`, which pins `--global` while being served by the local install.

So pruning a store is now gated on **ownership**: allowed only when the store was last written by a
version no newer than ours (the lock records a `ref` per entry). An unreadable or unversioned lock
means we cannot tell, so we leave it alone. Syncing stays unconditional — it is additive and
recoverable; deletion is not.

## Deliberate behaviour changes

- A machine with only a local install stops refreshing a leftover global lock. That is correct — no
  binary owns that store — but it is a change, so the command now says so once rather than letting it
  drift silently: *"Left the machine-wide skills alone — they track the global install."*
- Telemetry `scope_count` legitimately drops from 2 to 1 for local updates. The field still means
  "scopes this refresh acted on", so I left the event schema alone.

## Repair for anyone already affected

A re-pinned global store self-heals on the next run that includes the global target **and** is
executed by a package at least as new as the store's tracked set — i.e. `argent update --global` from
the global binary. It does *not* heal when run by the older local install, which is what caused it.
Worth a line in the release note.

## Verification

The original sandboxed repro, re-run against the fix:

```
update --local                 -> project            (no -g invocation at all)
update --global (global-mode)  -> project, global    (-g appears only here)
```

and the deletion path, against a store written by a newer install:

```
pruned: []      removals issued: 0
```

installer suite 548 passed. Existing scope tests were passing partly by accident — they touched one
store only because the other lock happened not to exist — so each now states the scope it exercises,
and the new containment tests stage **both** stores so acting on the wrong one is visible.

## Not in this PR

- **Version provenance.** `buildArgentSkillsSource(getInstalledVersion())` uses the *running* package,
  so `argent update --local` invoked from the global binary still pins the project lock to the
  global's version. The fix is a per-target version, but an unreadable version falls back to an
  absolute `SKILLS_DIR` path, which would write a machine-specific path into a committed lock — the
  portability problem #208 fixed. Needs its own decision.
- `uninstall.ts` computes orphans from the running package the same way, so it can strand newer skills.
  Same root cause, different command.

### DIFF
--- a/packages/argent-installer/src/skills.ts
+++ b/packages/argent-installer/src/skills.ts
@@ -3,13 +3,16 @@ import * as p from "@clack/prompts";
 import pc from "picocolors";
 import { track } from "@argent/telemetry";
 import { FAILURE_CODES, type FailureSignal } from "@argent/registry";
+import type { InstallMode } from "./utils.js";
 import {
   buildArgentSkillsSource,
   getGlobalSkillLockPath,
   getInstalledVersion,
   getProjectSkillLockPath,
+  compareVersions,
   listArgentSkillsInLock,
   listBundledSkills,
+  lockedArgentSkillVersion,
   withNpmForce,
   SKILLS_DIR,
 } from "./utils.js";
@@ -42,7 +45,12 @@ interface ScopeSpec {
   removeArgs: string[];
 }
 
-function getScopeSpecs(projectRoot: string): ScopeSpec[] {
+function getScopeSpecs(projectRoot: string, scopes: readonly SkillScope[]): ScopeSpec[] {
+  const wanted = new Set(scopes);
+  return allScopeSpecs(projectRoot).filter((spec) => wanted.has(spec.scope));
+}
+
+function allScopeSpecs(projectRoot: string): ScopeSpec[] {
   return [
     {
       scope: "project",
@@ -59,11 +67,61 @@ function getScopeSpecs(projectRoot: string): ScopeSpec[] {
   ];
 }
 
-// Re-syncs bundled argent skills into every scope that already tracks at
-// least one argent-owned skill, and prunes any argent-prefixed entry that is
-// no longer part of the bundled set. Safe to call from any command — a scope
-// with nothing tracked is skipped, so this never creates a
+/**
+ * Which skill stores a command may rewrite, given the installs it updated.
+ *
+ * A store is owned by exactly one install: the global lock always belongs to the
+ * binary on PATH, and a project's lock belongs to whichever install serves that
+ * project. Only the install that moved may rewrite the store that tracks it —
+ * otherwise a project-scoped update re-pins the machine-wide skills every other
+ * project uses.
+ *
+ * Note the project scope is not "the local install's scope": a non-interactive
+ * global install also writes a project lock, and that lock tracks the global
+ * binary. `installMode` is what distinguishes them.
+ */
+export function skillScopesForTargets(
+  targets: readonly InstallMode[],
+  installMode: InstallMode
+): SkillScope[] {
+  const scopes = new Set<SkillScope>();
+  if (targets.includes("local")) scopes.add("project");
+  if (targets.includes("global")) {
+    scopes.add("global");
+    // A global bump moves the binary this project runs, so its lock follows.
+    if (installMode === "global") scopes.add("project");
+  }
+  return (["project", "global"] as const).filter((scope) => scopes.has(scope));
+}
+
+/**
+ * Whether this package may decide which of a store's argent skills are obsolete.
+ *
+ * The obsolete set is the difference against THIS package's bundled skills, so a
+ * package older than the one that filled the store would classify everything
+ * added since as orphaned and delete it. Judging is allowed only when the store
+ * was last written by a version no newer than ours; an unreadable or unversioned
+ * lock means we cannot tell, so we do not prune.
+ */
+function mayPruneScope(lockPath: string): boolean {
+  const lockedVersion = lockedArgentSkillVersion(lockPath);
+  if (lockedVersion === null) return false;
+  const running = getInstalledVersion();
+  if (!running) return false;
+  return compareVersions(running, lockedVersion) >= 0;
+}
+
+// Re-syncs bundled argent skills into the caller's chosen scopes, and prunes
+// argent-prefixed entries that are no longer part of the bundled set.
+//
+// Two filters apply, and both matter. The caller names the scopes, because only
+// the install that moved may rewrite the store that tracks it. A named scope is
+// then skipped if it tracks no argent skill, so this never creates a
 // `skills-lock.json` in an unrelated working directory.
+//
+// Pruning has a third condition — see `mayPruneScope`. Sync is idempotent and
+// additive, so re-syncing a store from a slightly different version is
+// recoverable; a prune deletes, and is not.
 //
 // Sync prefers the GitHub-pinned source (`<repo>/packages/skills/skills#v<ver>`)
 // so the lockfile entry stays portable across machines. If the
@@ -75,7 +133,11 @@ function getScopeSpecs(projectRoot: string): ScopeSpec[] {
 // previous-version lock written from SKILLS_DIR would never refresh. `skills
 // add` rewrites the entry from the source we pass, which is exactly the
 // behavior we want after `npm i -g @swmansion/argent@new`.
-export function refreshArgentSkills(projectRoot: string): SkillScopeResult[] {
+export function refreshArgentSkills(opts: {
+  projectRoot: string;
+  scopes: readonly SkillScope[];
+}): SkillScopeResult[] {
+  const { projectRoot, scopes } = opts;
   const bundled = new Set(listBundledSkills());
   // An empty bundled set means THIS package's skills dir is unreadable — a
   // pruned pnpm store dir mid-update, a broken install — never "argent ships
@@ -92,7 +154,7 @@ export function refreshArgentSkills(projectRoot: string): SkillScopeResult[] {
     cwd: string;
   };
 
-  for (const spec of getScopeSpecs(projectRoot)) {
+  for (const spec of getScopeSpecs(projectRoot, scopes)) {
     const tracked = listArgentSkillsInLock(spec.lockPath);
     if (tracked.length === 0) continue;
 
@@ -123,7 +185,7 @@ export function refreshArgentSkills(projectRoot: string): SkillScopeResult[] {
       }
     }
 
-    if (orphaned.length > 0) {
+    if (orphaned.length > 0 && mayPruneScope(spec.lockPath)) {
       try {
         execFileSync("npx", withNpmForce([...spec.removeArgs, ...orphaned]), execOpts);
         result.pruned = orphaned;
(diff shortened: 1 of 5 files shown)


packages/argent-installer/src/update.ts
```diff
@@ -810,7 +810,15 @@ export async function update(args: string[]): Promise<void> {
...
       }
 
-      reportSkillRefresh(projectRoot, "installer_update_skills_refresh");
+      // Scopes follow the installs this run actually updated, not the project's
+      // mode alone — unlike the MCP config refresh above, which encodes which
+      // binary the project should launch. A skills lock records which binary's
+      // bundled skills it is pinned to, so only that binary moving may rewrite it.
+      reportSkillRefresh({
+        projectRoot,
+        scopes: skillScopesForTargets(targets, installMode),
```
latekvo, line 819: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: `scopes` here is derived from `targets` — the installs the run *attempted* — while the per-target `outcomes` collected above (lines 635-645) feed only the two early-exit gates. A partial run passes both gates (652 requires no `updated` at all; 661 likewise), so any target whose update failed, was declined, or no-op'd still contributes its scope here, and its store is re-synced even though that install never moved.

Reproduced both ways against a harness driving the real `update()` flow (modeled on update-decline.test.ts, child_process mocked, global store staged with refs at v0.20.0 and the global binary pinned at 0.20.0):

- `argent update --yes` in a local-mode project with both installs present, the global package-manager command failing with EACCES, the local install landing: the run reaches this line with `global` still in scope, and `skills add …#v0.18.1 … -g` re-pins every bundled entry of the machine-wide store to v0.18.1 while the PATH binary remains 0.20.0 — final lock refs `['v0.18.1', 'v0.20.0']` instead of all `v0.20.0`.
- Interactive variant: both targets selected, global confirm declined, local accepted — identical result.
- A `noop` needs no failure at all: `update --version <older>` no-ops the newer global target and still includes it in the scopes.

Impact: the machine-wide store's refs and fetched content get re-pinned to a version its owning binary does not run — the exact cross-install rewrite this PR's description and the comment two lines above ("only that binary moving may rewrite it") state is impossible — and the re-pinned refs then falsify the ownership gate's input for every later run (the lock claims a version the global binary never had). EACCES on a root-owned npm prefix and nvm/prefix-split "reported success but version didn't land" are both ordinary failure shapes, and `npx @swmansion/argent@latest update` widens the version skew further since the sync source is the running package's version.

packages/argent-installer/src/utils.ts
```diff
@@ -145,6 +145,47 @@ export function listArgentSkillsInLock(lockPath: string): string[] {
...
+  } catch {
+    return null;
+  }
+}
+
+/** Numeric-segment comparison of two `x.y.z` versions. Prerelease tags are ignored. */
+export function compareVersions(a: string, b: string): number {
+  const parse = (v: string) =>
+    v
+      .split("-")[0]!
```
latekvo, line 179: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: Dropping everything after the first `-` makes a same-base prerelease compare *equal* to its stable: `compareVersions("0.19.1-next.9", "0.19.1")` returns 0. `mayPruneScope` (skills.ts:111) therefore lets a running prerelease prune a store written by its own stable release, although the gate's documented rule is "the store was last written by a version no newer than ours" and semver orders the prerelease below its release (`semver.gt("0.19.1", "0.19.1-next.8")` → true).

Reproduced against the built dist with a fake running package at version `0.19.1-next.9` whose bundle lacks one store entry, and a global store written by stable 0.19.1 (both entries `ref: "v0.19.1"`):

```
RESULT: [{"scope":"global","synced":1,"syncError":null,"pruned":["argent-stable-only-skill"],…}]
skills remove -y -g argent-stable-only-skill
```

The deletion the gate exists to prevent — a package older than the store's owner removing skills that owner still ships — goes through. The repo publishes `-next.N` versions (git tags v0.17.1-next.1…, npm 0.19.1-next.8…), so a prerelease running over a store filled by its stable is a real shape, not a hypothetical input. This file's existing `isNewerVersion` (semver-based, invalid inputs rejected) is the nearest twin and does not produce this verdict.

packages/argent-installer/src/skills.ts
```diff
@@ -123,7 +185,7 @@ export function refreshArgentSkills(projectRoot: string): SkillScopeResult[] {
       }
     }
 
-    if (orphaned.length > 0) {
+    if (orphaned.length > 0 && mayPruneScope(spec.lockPath)) {
```
latekvo, line 188: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: The gate reads the lock *after* the sync above has rewritten it, and the skills CLI's own lock readers reject any lock whose numeric `version` field is missing or below the current schema (skills 1.5.22, `readSkillLock`: `if (typeof parsed.version !== "number" || !parsed.skills) return createEmptyLockFile()`; `readLocalLock` has the identical check for project locks). `skills add` against such a lock therefore rewrites the file with *only the just-added entries* — and by the time `mayPruneScope` runs, the refs it judges are the ones the sync just wrote.

Reproduced end-to-end with the real skills CLI 1.5.22 and the real pinned GitHub source (no stubs): a global lock with **no `version` field**, two argent entries pinned `ref: "v0.20.0"`, and a running package at 0.19.0 whose bundle lacks one of them. Per the ownership rule this store (v0.20.0) must be safe from a 0.19.0 package:

```
RESULT: [{"scope":"global","synced":1,"syncError":null,"pruned":["argent-from-the-future"],…}]
lock version: 3 | entries: 15
argent-from-the-future still in lock? false
argent-from-the-future still on disk? false
```

The sync wiped the v0.20.0 refs from the lock before the gate read them, so the gate saw only freshly-written v0.19.0 refs and allowed the machine-wide removal — skill gone from both the lock and `~/.agents/skills/`. Reachable for any lock the CLI considers malformed: hand-edited, merge-resolved (the project lock is a committed file), or written by an older schema version. Independent of the prune, the same rewrite silently drops every tracking entry the CLI did not reinstall.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: Reviewed at b9d1f7fc against an `npm install --include=dev` + full `tsc --build` worktree, installer suite 548/548 green. What I ran: claims-vs-code on every new comment/JSDoc/message, nearest-twin comparison (uninstall.ts / init-skills.ts / topology.ts / the file's own semver-based isNewerVersion), non-happy paths through the real compiled dist with stubbed and real `npx` (skills CLI 1.5.22 installed from npm, real GitHub clone of the pinned source), hostile-lock fuzzing (39 shapes — all fail-closed), mutation testing of every new branch and both call sites, and end-to-end harnesses driving the real `update()` flow. Four findings below, each anchored where it bites and each reproduced before writing: requested-vs-actual targets in the update wiring, a prerelease tie in the new version comparator that opens the prune gate, the sync step falsifying the prune gate's input on locks the skills CLI rejects as malformed, and the call-site wiring surviving mutation (including the exact regression this PR fixes).

packages/argent-installer/test/skills.test.ts
```diff
@@ -286,3 +297,113 @@ describe("formatSkillRefreshSummary", () => {
     expect(summary).toContain("permission denied");
   });
 });
+
+describe("scope containment", () => {
```
latekvo, line 301: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: The containment and ownership tests below pin `refreshArgentSkills` in isolation, but nothing pins the scopes the commands actually pass: no suite references `reportSkillRefresh` or either call site (update.ts:817, install-runner.ts:378), even though update-decline.test.ts executes the real `update()` flow through the former. Mutation runs against the full installer suite (548 tests, all green at the PR head):

- update.ts passing `skillScopesForTargets(["global", "local"], installMode)` unconditionally — i.e. reintroducing exactly the rewrite this PR fixes — leaves the suite green (verified by applying the mutation and re-running `npx vitest run`: 548 passed).
- update.ts passing `scopes: []` (refresh silently disabled) — green.
- install-runner.ts claiming both scopes on the global bump — green.
- Dropping the `!running` guard in `mayPruneScope`, and selecting the smallest ref instead of the newest in `lockedArgentSkillVersion` (every fixture uses uniform refs) — both green.

The reported bug (#637) could regress, or the refresh could be switched off entirely, without a single test failing; the two new call sites ship without any test asserting which store a real command touches.

PR #560, opened 2026-07-23
fix(native-devtools): derive restart_required from the running process instead of asserting it

latekvo:
this is urgent 

---

## Problem

Every `native-*` tool can get stuck telling the agent to do something that will never work.

I hit this during an end-to-end pass over 0.17.0. After a mid-session simulator reboot, `native-devtools-status` reported:

```json
{ "envSetup": true, "appRunning": true, "connected": false,
  "requiresRestart": true, "nextLaunchWillBeInjected": true, "injectable": true }
```

and every `native-*` tool returned `restart_required` with "Call restart-app then retry." Following that advice never helped, and the message offered nothing else. `flow-execute` was blocked too, since its `launch` step needs the connection, and flow recording silently degraded a portable `tap: { selector }` to raw coordinates when it couldn't reach native devtools.

The app was injected the whole time. All three dylibs were mapped into the process and the simulator log showed a successful dial on every launch, while the tool-server held the listening end of that exact socket and still reported `connected: false`. `argent server stop && argent server start` cleared it instantly.

## The real defect is the state, not the wording

`requiresAppRestart` returned `true` for every app without a live connection:

```ts
async requiresAppRestart(bundleId) {
  if (connections.has(bundleId)) return false;
  await reverifyEnv();
  return true;   // <- asserted, never measured
}
```

That is not a measurement. `restart_required` was a constant wearing a state's clothes, and the advice built on it stays literally true forever, so an agent obeying it restarts the app indefinitely.

Softening the message would have left that intact. This PR measures the state instead.

## Fix — derive it from the running process

`restart-app` relaunches into the simulator's **current** launchd environment, so it can only ever help a process launched under different terms than a fresh launch would get. That is answerable: read the process. `launchctl list` already gives its pid; `ps eww -p <pid> -o etime=,command=` gives the environment it was exec'd with and its age. Simulator apps are ordinary host processes owned by the same user, so both are readable.

`appConnectionState` (replacing `requiresAppRestart`) returns:

| state | meaning | remedy |
| --- | --- | --- |
| `connected` | devtools are live | — |
| `not_running` | no process to read | `launch-app` |
| `stale_process` | no bootstrap dylib, a previous run's endpoint, or older than this service's listener | `restart-app` |
| `unregistered` | injected against **this** endpoint, started after the listener came up, still silent | restart the **tool-server** |
| `indeterminate` | process could not be read (ios-remote, unreadable `ps`) | `restart-app`, with the loop warning |

[...]

PR #677, opened 2026-08-01
fix(flow): let a flow launch an app that can never be instrumented

### LINKED ISSUE #623: flow `launch` step requires native devtools, so a coordinate-only flow can never launch a non-injectable app (com.apple.*)
## Summary

A flow whose `launch` step targets an app that can **never** be injected (an Apple system app — library
validation blocks the dylib) always errors, even when the rest of the flow only taps coordinates and
needs no view hierarchy at all:

```
[1] ✗ launch — could not connect to native devtools for com.apple.Preferences.
    Re-run to relaunch the app and retry. If it keeps failing, a stale or duplicate
    argent server may be holding the devtools connection — restart the argent server and try again.
FAIL — 0 passed, 0 failed, 1 errored, 2 skipped
```

`native-devtools-status` correctly reports `injectable: false` for such apps, so the information needed to
short-circuit this is already available — but `launch` still waits for a devtools connection and then
suggests remedies (re-run, restart the server) that can never succeed for this app.

## Impact

System apps can't be driven from flows at all, even for pure coordinate/echo/await sequences. The error
also reads as retryable, which invites exactly the restart→retry loop that #453 set out to eliminate.

## Suggested fix

- Skip (or soft-fail with a note) the devtools wait when `injectable: false`, letting the launch proceed
  for flows that don't need the hierarchy.
- Distinguish "this app cannot be injected" (terminal) from "the devtools connection was lost" (needs a
  tool-server restart) in the error text.

## Verified not a flow-runner regression

With the `launch` step removed, a coordinate-based flow replays green:
`PASS — 3 passed, 0 failed, 0 errored, 0 skipped` (echo → await visible → tap → await hidden; the target
was really dismissed).

## Related

#453 (closed — non-injectable apps never reaching a terminal status), #561 (the devtools service dropping
connections; that issue already lists this same `launch` failure as a downstream symptom), #519 (flows
false-passing `hidden` against a non-injectable app).

---

filip131311:
Fixes #623.

## Reproduced

```
[0] launch  error  could not connect to native devtools for com.apple.Preferences.
                   Re-run to relaunch the app and retry… restart the argent server…
[1] echo    skip
[2] tap     skip          ok=false errored=1 skipped=2
```

Control — the identical flow **without** the launch step passes (`ok=true`). So the runner was fine; the gate alone made system apps undriveable, and it gave advice that can never work for them.

The signal already existed: `isInjectableBundleId` is used by `native-devtools-status` (whose docs call `injectable: false` **terminal — do NOT restart/retry**) and by `describe`. The flow gate never consulted it, so it burned the full **8 s** `NATIVE_READY_TIMEOUT_MS` waiting for a connection that cannot occur.

After:

```
[0] launch  pass   ⚠ com.apple.Preferences is an Apple system app… coordinate steps work;
                     selector-based steps cannot resolve for this app.
[1] echo    pass
[2] tap     pass          ok=true, ~3s total
```

## Removing the wait exposes a race the wait was hiding — so this closes it too

This is the part I'd want a reviewer to look at hardest.

Selector steps reach `resolveNativeTargetApp`, which auto-targets whatever app is connected. `chooseFrontmostConnectedApp` has a **weak tier** accepting `applicationState === "inactive"` — exactly what an app reports while it backgrounds. Today the first tree read lands ~9.5 s after launch (8 s gate + 1.5 s settle), comfortably past that window. **Without the wait it lands at ~1.5 s, inside it.**

So a selector step could resolve against a *different* app's tree. Most conditions then fail noisily, but `hidden`-shaped ones (`assert: { hidden: X }`, `await: { hidden: X }`, a `when: { hidden: X }` guard reporting a green skip) would **pass**. The fix records that the launched app cannot be instrumented and fails tree reads immediately with a terminal reason, which closes it.

Without that guard the later path was also *worse* than the gate it replaced: `NATIVE_TARGET_NO_CONNECTED_APPS` says "**Launch or restart the app first**", after a 3 s settle.

```
[1] await  fail  `com.apple.Preferences` is an Apple system app, so … selector-based steps cannot
                 resolve. This is terminal — relaunching or restarting the argent server will not
                 change it. Target this screen by coordinate (`tap: { x, y }`) instead.
```

## Design choices

**Guard on tree reads, not step kinds** — `settleTree` + `waitForCondition`. That covers `tap`/`long-press`/`await`/`assert`/`scroll-to`/`snapshot` in two places, and leaves selector-less `pinch`/`rotate` working (they degrade to a default aspect and never read a tree). Guarding by step kind would have missed `scroll-to`, which calls `settleTree` directly.

**No pre-flight "does this flow use selectors" scan.** `launch` can appear anywhere, including inside a nested fragment, and `run:` fragments are read lazily at execution time — so "does this flow contain selectors" is ill-posed (which steps? after this launch? under a `when:` that may not be entered?).

**Pass, not fail.** The launch genuinely succeeded — the issue's own control proves the rest of such a flow runs. Failing it would re-introduce the bug for coordinate flows.

**No 5th status.** The CLI maps status with no fallback, so an unknown value renders `undefined`. Used the existing `warning` field instead: ⚠ replaces the pass glyph, the text prints under the step, and it's counted in the summary. Both renderers already handle it as legacy wire-compat, so an old CLI renders it correctly — those comments are updated, since it's now actually produced.

## Also fixed

The create-flow skill's recording walkthrough used `com.apple.Preferences` **with selector steps** (`tap: { text: General }`). That could never have been captured (selector capture reads the same tree) or replayed. Switched to a third-party app.

## Not fixed, deliberately

A *fragment* with no `launch` step, run while a system app is frontmost, never sets the flag and still gets the raw message. That's the issue's own control case.

## Checks

- 3088 tests pass; 2 new, both failing against the pre-fix source. **All 542 existing flow tests pass unmodified**, including `flow-composition.test.ts:213`, which pins the retry-worded failure for a genuinely injectable app.
- CLI (277) and MCP (78) renderer suites pass.
- One unrelated flake on a first full run that passed on re-run.
- Skills gate 10.0; extract-tools 46/46; prettier, eslint, both typechecks clean; lock untouched.

### DIFF
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -52,12 +52,22 @@ export interface ActionEnv {
   ctx?: ToolContext;
   device: DeviceInfo;
   signal?: AbortSignal;
+  /**
+   * Bundle id of an app the run launched that can never carry the
+   * view-hierarchy instrumentation. Set means every tree read is doomed, so
+   * they fail immediately with a terminal reason instead of polling a source
+   * that will never appear — and, more importantly, instead of auto-targeting
+   * whichever other app happens to still be connected.
+   */
+  nonInjectableApp?: string;
 }
 
 /** Outcome of a selector directive: ok, or a machine-readable reason it failed. */
 export interface DirectiveOutcome {
   ok: boolean;
   reason?: string;
+  /** Caveat carried by a step that still succeeded. */
+  warning?: string;
   /** The run was cancelled mid-step — reported as a skip, not a step failure. */
   aborted?: boolean;
   /**
@@ -314,7 +324,24 @@ function flowSelectorToFrame(tree: DescribeNode, sel: FlowSelector): DescribeFra
  * convert the outage into a misleading "element not found" downstream. The
  * throw lands in the step's structured report via `execLeafStep`'s catch.
  */
+/**
+ * Why no selector can resolve against an app that cannot be instrumented.
+ * Deliberately not the native-devtools recovery text, which tells the caller to
+ * use `describe`/`screenshot` — the answer for a flow is coordinate steps.
+ */
+export function nonInjectableTreeReason(bundleId: string): string {
+  return (
+    `\`${bundleId}\` is an Apple system app, so argent's view-hierarchy instrumentation can never ` +
+    `be injected into it and selector-based steps cannot resolve. This is terminal — relaunching or ` +
+    `restarting the argent server will not change it. Target this screen by coordinate ` +
+    `(\`tap: { x, y }\`) instead.`
+  );
+}
+
 export async function settleTree(env: ActionEnv): Promise<DescribeNode | undefined> {
+  if (env.nonInjectableApp) {
+    throw new Error(nonInjectableTreeReason(env.nonInjectableApp));
+  }
   const deadline = Date.now() + SETTLE_TIMEOUT_MS;
   let prevFp: string | undefined;
   let prevTree: DescribeNode | undefined;
@@ -992,6 +1019,12 @@ async function waitForCondition(
   },
   timeoutMs: number
 ): Promise<DirectiveOutcome> {
+  // Not routed through settleTree, so it needs its own guard — and a `hidden`
+  // wait is the one condition that would otherwise resolve TRUE off an
+  // unreadable screen.
+  if (env.nonInjectableApp) {
+    return { ok: false, reason: nonInjectableTreeReason(env.nonInjectableApp) };
+  }
   const deadline = Date.now() + timeoutMs;
 
   let lastMatches: ReturnType<typeof findAll> = [];
--- a/packages/tool-server/src/tools/flows/flow-run.ts
+++ b/packages/tool-server/src/tools/flows/flow-run.ts
@@ -41,7 +41,11 @@ import {
   type ActionEnv,
   type DirectiveOutcome,
 } from "./flow-actions";
-import { nativeDevtoolsRef, type NativeDevtoolsApi } from "../../blueprints/native-devtools";
+import {
+  nativeDevtoolsRef,
+  isInjectableBundleId,
+  type NativeDevtoolsApi,
+} from "../../blueprints/native-devtools";
 import { androidDevtoolsRef, type AndroidDevtoolsApi } from "../../blueprints/android-devtools";
 import {
   chromiumCdpRef,
@@ -111,6 +115,12 @@ export interface StepReport {
    * percentage, baseline written/updated).
    */
   reason?: string;
+  /**
+   * A caveat about a step that still passed — currently, launching an app that
+   * can never carry the view-hierarchy instrumentation. Renderers show it as a
+   * `⚠` in place of the pass glyph and count it in the summary.
+   */
+  warning?: string;
   /** Underlying tool id for `tool` steps. */
   tool?: string;
   /** Tool result for `tool` steps. */
@@ -282,6 +292,10 @@ async function treeSourceGate(
   signal?: AbortSignal
 ): Promise<string | null> {
   if (device.platform === "ios" && !signal?.aborted) {
+    // An app that can never be injected will never connect, so waiting the full
+    // timeout only delays advice that cannot work. The flow can still run every
+    // step that does not read the view hierarchy.
+    if (!isInjectableBundleId(bundleId)) return null;
     const connected = await waitForNativeDevtools(registry, device, bundleId, signal);
     if (!connected && !signal?.aborted) {
       return (
@@ -396,11 +410,25 @@ async function runLaunch(state: ExecState, app: Launch): Promise<DirectiveOutcom
     return { ok: false, reason: `restart-app failed: ${errMsg(err)}` };
   }
   if (!(await sleepOrAbort(POST_LAUNCH_SETTLE_MS, signal))) return ABORTED_OUTCOME;
+  // Recorded on every launch, so a later injectable launch clears it.
+  state.nonInjectableApp =
+    device.platform === "ios" && !isInjectableBundleId(bundleId) ? bundleId : undefined;
+
   const gate = await treeSourceGate(registry, device, bundleId, signal);
   // The gate returns null (ready) on abort — check the signal before trusting
   // it, or a cancelled gate would read as a launch that verified readiness.
   if (signal?.aborted) return ABORTED_OUTCOME;
   if (gate) return { ok: false, reason: gate };
+  if (state.nonInjectableApp) {
+    return {
+      ok: true,
+      warning:
+        `${bundleId} is an Apple system app: it is a platform binary with library validation, so ` +
+        `argent's view-hierarchy instrumentation can never be injected into it. The app launched — ` +
+        `coordinate steps (\`tap: { x, y }\`), \`wait\` and \`snapshot\` work; selector-based steps ` +
+        `cannot resolve for this app.`,
+    };
+  }
   return { ok: true };
 }
 
@@ -1052,7 +1080,12 @@ async function execLeafStep(
       // A run cancelled mid-launch is a skip (matching the pre-step guard and
       // the directives), never a step failure — the app did nothing wrong.
       if (r.aborted) return { ...base, status: "skip", reason: r.reason };
-      return { ...base, status: r.ok ? "pass" : "error", reason: r.reason };
+      return {
+        ...base,
+        status: r.ok ? "pass" : "error",
+        reason: r.reason,
+        ...(r.warning ? { warning: r.warning } : {}),
+      };
     }
 
     case "tap":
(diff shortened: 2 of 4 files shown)


latekvo:
This is likely a duplicate of #560, although the very same bug seems fixed using a different approach.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -314,7 +324,24 @@ function flowSelectorToFrame(tree: DescribeNode, sel: FlowSelector): DescribeFra
...
+  return (
+    `\`${bundleId}\` is an Apple system app, so argent's view-hierarchy instrumentation can never ` +
+    `be injected into it and selector-based steps cannot resolve. This is terminal — relaunching or ` +
+    `restarting the argent server will not change it. Target this screen by coordinate ` +
+    `(\`tap: { x, y }\`) instead.`
+  );
+}
+
 export async function settleTree(env: ActionEnv): Promise<DescribeNode | undefined> {
+  if (env.nonInjectableApp) {
```
latekvo, line 342: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This derives a terminal verdict about the view hierarchy from `bundleId.startsWith("com.apple.")` rather than from whether the hierarchy read succeeds, and where those two disagree it turns a passing flow into a failing one.

With `com.apple.Preferences` in `listConnectedBundleIds()`, `isConnected()` returning true, `applicationState: "active"` and a `getFullHierarchy` payload containing `Settings`, the step `assert: { visible: "Settings" }` after `launch: com.apple.Preferences`:

```
main @3fff176a  ->  launch:pass, assert:pass   ok=true
this branch     ->  launch:pass, assert:fail   ok=false
```

with the reason `"... can never be injected into it and selector-based steps cannot resolve. This is terminal — relaunching or restarting the argent server will not change it."` — emitted while the read that message describes as impossible had already returned a tree containing `Settings`. Same for `tap: { text: Search }`.

That state is not hypothetical: it is what was measured on an iOS 18.5 simulator during the review of #560 — `lsof` showing both `libArgentInjectionBootstrap.dylib` and `libNativeDevtoolsIos.dylib` mapped into the running process, the process holding a live unix peer of the tool-server's own `/tmp/argent-nd-<udid>.sock`, `native-devtools-status` returning `"connected": true` and `"injectable": false` **in the same response**, and a selector fragment passing 4/4. #453 recorded `connected: false` on iOS 26.5 and #623 was filed from a 26.5 matrix, so the evidence splits by runtime; the message states one runtime's reading as a universal.

The clearest form of it is a pair that differs by one line of YAML. Same app, same connection, same three steps, same tree:

```
no `launch` step   ->  3/3 pass on BOTH branches   (the guard never arms)
`launch` prefixed  ->  assert:fail, tap:skip, assert:skip   on this branch only
```

A second consequence: for a system app that genuinely never connects, this branch emits **byte-identical** output to the connected case above, so the two are no longer distinguishable from the report. On the guards-removed build the unreadable case still fails, but says `"could not read the UI tree: No native-devtools-connected apps are available for auto-targeting."`

On isolating it — deleting only the two `env.nonInjectableApp` guards, keeping the `flow-run.ts:298` gate skip and the launch warning, turns the whole file green at 16/16. The gate skip is not implicated.

On the 8 s the description cites as the cost being removed: `waitForNativeDevtools` tests `api.isConnected(bundleId)` at the top of its loop, before the first sleep, and `isConnected` is a synchronous `Map.has`. A connected app therefore already returns in ~0 ms on `main`. Measured, launching a connected system app takes **1507 ms on this branch and 1508 ms on `main`** — the wait being removed never fires for the case that regresses. The full 8 s elapses only when the app never connects (9522 ms), and there `main` errored the launch, so no selector step ran.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat)\]: Reviewed against `bce41535`, comparing behaviour with `origin/main` (`3fff176a`). Evidence is one identical test file run on both, driven through the real `createRunFlowTool`: the native-devtools **service** is faked (the pattern in `native-target-app.test.ts`), not `fetchFlowTree`, so the whole chain runs — `settleTree → fetchFlowTree → queryFullHierarchyTree → resolveNativeTargetApp → queryViewHierarchy → adaptFullHierarchy`.

**`main`: 16/16 passed. This branch: 4 failed | 12 passed.**

Two things I want to say plainly before the findings. The launch-step fix is right — a flow that launches a system app should not have its launch step errored, and the control in #623 proves the rest of such a flow runs. And the race this PR raises in its description is **real**: I reproduced it, and on a build with the gate skip kept but the two guards removed, `launch` of a non-connecting system app followed by `assert hidden` does flip to a false pass. That case is genuine and the guard does close it.

The findings below are about the guard's derivation, not about whether the problem is worth solving.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -992,6 +1019,12 @@ async function waitForCondition(
   },
   timeoutMs: number
 ): Promise<DirectiveOutcome> {
+  // Not routed through settleTree, so it needs its own guard — and a `hidden`
+  // wait is the one condition that would otherwise resolve TRUE off an
+  // unreadable screen.
+  if (env.nonInjectableApp) {
+    return { ok: false, reason: nonInjectableTreeReason(env.nonInjectableApp) };
```
latekvo, line 1026: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This early return omits `indeterminate`, so a `when:` guard reads an unreadable tree as a plainly-false condition and green-skips the block.

This one does not depend on the injectability question at all — it fires when the premise is entirely correct, on an app that genuinely never connects.

`probeWhenCondition`'s own docstring states the contract: *"`indeterminate` distinguishes an unreadable tree (the caller errors — unknown is not false) from a plainly unmet condition (the caller skips)."* `execWhenStep` branches on exactly that flag, and its docstring adds that *"silently skipping would let a broken tree source turn every guarded dismissal into a green no-op."* Every other unreadable-tree exit in this same function sets it (the `lastTrustedReadAt === undefined` arm, and both `!lastReadTrusted` arms).

Driving `launch: com.apple.Preferences` then `when: { hidden: "Onboarding" }` wrapping an `echo`, against a registry whose `resolveService` throws:

```
this branch  ->  launch:pass, when:skip, echo:skip    ok=TRUE
                 when reason: 'condition not met (hidden text="Onboarding") — block skipped (1 step)'
main         ->  launch:error, when:skip              ok=false
```

The run reports green and the report asserts the element was not hidden, which nothing observed. The sibling `await: { hidden: ... }` on the identical state correctly reports `fail`, so the same condition yields two different verdicts depending only on which directive asked. That is the shape of #519, reached through the `when:` door.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -396,11 +410,25 @@ async function runLaunch(state: ExecState, app: Launch): Promise<DirectiveOutcom
     return { ok: false, reason: `restart-app failed: ${errMsg(err)}` };
   }
   if (!(await sleepOrAbort(POST_LAUNCH_SETTLE_MS, signal))) return ABORTED_OUTCOME;
+  // Recorded on every launch, so a later injectable launch clears it.
+  state.nonInjectableApp =
+    device.platform === "ios" && !isInjectableBundleId(bundleId) ? bundleId : undefined;
```
latekvo, line 415: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The guard is keyed on the **launched** bundle id, but the read it guards resolves a **different** quantity: `flow-ios-tree.ts:287` calls `resolveNativeTargetApp(nativeApi, undefined)`, which auto-targets from the connected list and never consults `state.nonInjectableApp`. Two consequences fall out of the mismatch.

**It is bypassable.** `tool: { name: launch-app, args: { bundleId: com.apple.Preferences } }` followed by the same `assert` never sets the flag:

```
tool:pass, assert:pass    on BOTH branches
```

Same app, same launch, same tree — the guard is silent because the launch did not go through the `launch:` directive. So the terminal claim holds or not depending on which spelling of "launch this app" the flow used.

**It fires for an app that was never the read target.** Launching `com.apple.Preferences` while a connected, `active`, injectable `com.example.myapp` is what auto-target would resolve: this branch blocks that read with `"com.apple.Preferences ... can never be injected"`, naming an app that is not the one the read would have used.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -396,11 +410,25 @@ async function runLaunch(state: ExecState, app: Launch): Promise<DirectiveOutcom
...
   // it, or a cancelled gate would read as a launch that verified readiness.
   if (signal?.aborted) return ABORTED_OUTCOME;
   if (gate) return { ok: false, reason: gate };
+  if (state.nonInjectableApp) {
+    return {
+      ok: true,
+      warning:
+        `${bundleId} is an Apple system app: it is a platform binary with library validation, so ` +
+        `argent's view-hierarchy instrumentation can never be injected into it. The app launched — ` +
+        `coordinate steps (\`tap: { x, y }\`), \`wait\` and \`snapshot\` work; selector-based steps ` +
```
latekvo, line 428: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This warning tells the flow author that `snapshot` works for such an app, and the guard added in this PR contradicts it.

`snapshot: { cropOn }` routes through `waitForFrame` into `settleTree`, which is the `flow-actions.ts:342` throw. On `launch: com.apple.Preferences` + `snapshot: { name: row, cropOn: { text: Settings } }`:

```
"kind": "snapshot", "status": "error",
"reason": "`com.apple.Preferences` is an Apple system app, ... This is terminal ..."
```

On `main` the same step gets past the tree read. A plain `snapshot` with no `cropOn` does work, because `flow-visual.ts` swallows that throw — so the sentence is true for one form of the directive and false for the other. The PR description's design-choices paragraph lists `snapshot` as covered by the guard, which is the opposite of what this line tells the author.

filip131311:
eehhh well fix it as part of the #560

Issue #678, opened 2026-08-01
uninstall --local leaves the .zed/settings.json it created and an empty .kiro/


filip131311:
`argent uninstall --local` leaves behind the `.zed/settings.json` it created (holding only argent's own stanza) and an empty `.kiro/`.

Split out of #622, which is about the destructive *ordering* of uninstall. These two are cleanup-completeness bugs on a different path, with no failure involved — worth fixing separately.

### Reproduced on 0.18.0 (Linux, Lima VM)

Workspace created empty, so every dot-directory below was created by `argent init` itself:

```
mkdir local-repro && cd local-repro
npm init -y
argent init -y --local
argent uninstall -y --local          # succeeds, exit 0
```

Leftovers:

```
.kiro  .zed  node_modules  package-lock.json  package.json

.kiro/                     ← empty; created by init, never removed
.zed/settings.json         ← created by init, still contains ONLY argent's stanza:
   {
     "agent": { "tool_permissions": { "default": "confirm" } }
   }
```

argent created that file and that directory and leaves both behind containing nothing but argent configuration. A user who never used Zed now has a `.zed/` in their repo after installing and uninstalling argent.

### The global path gets this right

The same test in global mode ends at a bare `./package.json` — every directory argent created is removed, including the now-empty parents. So the writers themselves are fine; the `--local` teardown is the gap.

### Correctness note for whoever fixes it

The cleanup must only remove a settings file argent **created**. If the user already had `.zed/settings.json` with their own keys, argent injects into it and must then remove only its own stanza and leave the file. I verified that direction separately: with a pre-existing `.zed/settings.json` containing `{"theme": "One Dark"}`, init left it untouched and uninstall left it untouched — so the injection only happens in some conditions, and the fix needs to distinguish "file is entirely ours" from "we added a stanza to theirs".

### Not reproducible, for the record

The third adjacent claim in #622 — dangling skill symlinks in `.claude/skills/` after `uninstall --local` — does **not** reproduce on 0.18.0. In both global and local mode skills materialize as real directories under `.agents/skills`, `find . -xtype l` is empty before and after, and `.claude/` is removed entirely. Either already fixed or specific to an older layout.

PR #679, opened 2026-08-01
fix(installer): refuse an uninstall that cannot finish, before it destroys anything

### LINKED ISSUE #622: argent uninstall removes workspace configs before `npm uninstall -g` fails unelevated, leaving a half-uninstalled install
## Symptom

On the ordinary Linux install shape (`sudo npm i -g @swmansion/argent`, default `/usr/local` prefix),
`argent uninstall -y`:

1. removes **all** workspace/editor configs first, then
2. runs `npm uninstall -g` **unelevated** → `EACCES` and ~35 lines of raw `npm ERR!` stack, ending
   `■ global uninstall failed`

Result: configs gone, package still installed, and no hint that sudo is what's missing. Re-running under
sudo completes cleanly — so only the diagnosis is absent, but the intermediate state is broken.

## Suggested fix

- Probe writability of the global prefix **before** touching any config, and abort early with
  "re-run with sudo" if it isn't writable.
- Or uninstall the package first and only remove configs once that succeeded.
- Either way, don't surface a raw npm stack — map EACCES to the actionable one-liner.

## Smaller adjacent findings in the same command

- `argent uninstall --local` leaves **dangling skill symlinks** in `.claude/skills/` pointing at the
  deleted `.agents/skills/*`.
- Uninstall correctly strips MCP entries and allowlists, but leaves an argent-injected
  `agent.tool_permissions` stanza in `.zed/settings.json`, plus empty `.kiro/` and `.zed/` dirs.

---

filip131311:
Fixes #622.

`argent uninstall` prunes the workspace first and removes the package last. On the layout the issue reports — `sudo npm i -g @swmansion/argent`, npm prefix `/usr/local`, root-owned `node_modules` — the removal dies on EACCES after the pruning is already done.

## Reproduced

On a Linux VM matching the report exactly:

```
◇  Pruned Argent Content ──────────────────────────────────╮
│  + Removed 15 Argent skill entries from .agents/skills    │
│  + Removed 15 Argent skills from skills-lock.json         │
╰──────────────────────────────────────────────────────────╯
●  Running: npm uninstall -g @swmansion/argent
npm ERR! code EACCES
npm ERR! syscall rename
   … 30 more lines of arborist/reify internals …
■  global uninstall failed: Error: Command failed: npm uninstall -g @swmansion/argent

workspace:  ./package.json          ← every skill + lockfile gone
package:    /usr/local/bin/argent   ← still installed, still runs
```

**And it exits 0.** A failed, destructive, half-finished uninstall reports success to the shell, so `argent uninstall -y && …` carries on as though the package were gone. That is not in the issue; I found it while reproducing. It reads as an oversight rather than a contract — the sibling tool-server failure in the same function *throws* (`uninstall.ts:709`), and nothing in this repo chains `argent uninstall`.

## The fix

Ask the environment before touching anything. npm doesn't unlink the package directory, it **renames it aside within its parent** (`@swmansion/argent` → `@swmansion/.argent-p3dt2fHx`), so one `access(W_OK)` on that parent decides the outcome. Measured on the reported layout: `false` unelevated, `true` as root.

The probe is biased hard toward silence. Windows (where `access(W_OK)` reflects only the read-only attribute and would answer "writable" for a directory you cannot touch), any non-npm global manager, a linked install, and every inconclusive stat all return `"unknown"`, which callers treat exactly like `"writable"`. **A wrong `"blocked"` refuses an uninstall that works, which is worse than the bug being fixed.**

It also derives the install path *logically* from the prefix instead of realpathing the bin. Under `npm link` the realpath lands in the source checkout, and probing that checkout's parent — a directory npm never renames — would produce a false block whenever the checkout happened to be read-only.

Blocking applies only where the removal is already consented to (`-y`, `--global`, or the coexistence prompt). A bare interactive run still asks, still defaults to no, and pruning-while-keeping-the-package remains a supported outcome — the prompt just says up front that the removal needs elevation.

### Why not simply reorder

Removing the package first is the obvious fix and it is unsafe. The prune reads `SKILLS_DIR` / `RULES_DIR` / `AGENTS_DIR` off the **running** package at prune time (`getBundledSkillNames` → `[]` on a missing dir, `uninstall.ts:182`; `removeBundledContent` → empty, `:112-114`). When the running copy *is* the global install, npm renames that tree away and the entire prune becomes a silent no-op that orphans every installed skill, rule and agent. Making the reorder safe needs snapshotting the manifests and a signature change — a much larger, riskier change than this.

### Two things that fall out of the same bug

- **Exit code.** A failed removal now exits 1, and a failed global no longer silently skips a local removal the user also asked for.
- **Classification without parsing.** `execShellCommandSync` inherits stdio, so npm's stderr goes to the terminal and the caught error carries only `"Command failed: …"` — there is nothing to regex. A removal that fails anyway is classified by **re-running the probe**. That also means no stdio change was needed, so `update` is untouched and npm's live output is unchanged.

### The remedy line

`sudo HOME="$HOME" argent uninstall --global`, deliberately **not** `sudo -E`. Ubuntu 25.10 ships **sudo-rs**, which does not implement `-E` at all:

```
$ sudo -E bash -c 'echo HOME=$HOME'
warning: preserving the entire environment is not supported, `-E` is ignored
HOME=/root
```

A reset `HOME` would make the global-scope cleanup clean root's home and silently miss the user's own `~/.claude`, `~/.cursor` and friends. The `sudo VAR=value` form assigns the variable directly and survives `env_reset` on both sudo implementations. Windows gets an Administrator-terminal hint instead.

Re-running **argent** rather than npm directly matters too: removing the package by hand strands every MCP entry, skill and rule pointing at a binary that is gone, with argent no longer around to clean them up.

## Verified live, end to end

Same VM, same workspace, fix grafted into the installed package:

```
■  Cannot remove the global @swmansion/argent package:
   /usr/local/lib/node_modules/@swmansion is not writable by this user,
   so the removal would fail partway through.
●  Re-run it with elevated permissions: sudo HOME="$HOME" argent uninstall --global
●  Nothing was changed by this run: argent is still installed, so its
   configuration was left in place.
└  @swmansion/argent was not removed.
EXIT CODE: 1

workspace after:  all 16 skills still present   ← the whole point
```

Then the suggested command, from that same state:

```
$ sudo HOME="$HOME" argent uninstall --global -y
◆  Removed global package.
└  argent has been removed.
EXIT: 0

workspace:  ./package.json only        package: gone from PATH
```

## Tests

`uninstall-permissions.test.ts` covers the probe (blocked / writable / linked-install / not-on-PATH / missing dir / non-npm manager / Windows). `uninstall-preflight.test.ts` covers the flow: workspace byte-identical and no package-manager call on a blocked run, its own telemetry code, the remedy text, fail-open on `"unknown"` and `"writable"`, `--local` unaffected by an unwritable global prefix, exit 1 on failure, and permission-vs-other-failure classification.

Both halves are mutation-verified — disabling the gate fails 4 tests, reverting the exit code fails 4 more.

Two existing-suite changes were required: `uninstall.test.ts:148` drives exactly the path that now exits 1 (it becomes the regression guard for the reported bug), and its probe is pinned so the suite's result cannot depend on whether the machine running it happens to have a root-owned `/usr/local/lib/node_modules/@swmansion` — green on nvm, red on a stock Linux box.

Also fixes the e2e global-uninstall phase, which asserted the config was gone regardless of whether the uninstall actually ran (`00-install.sh:165`) and would hard-fail when the uninstall correctly refuses.

## Not in scope

- `update` has the same unmapped-EACCES gap. Not destructive in the same way (nothing is pruned first), so it belongs in its own change.
- The two adjacent cleanup findings from the issue are split out as #678. The third — dangling `.claude/skills` symlinks — does **not** reproduce on 0.18.0 (skills are real directories under `.agents/skills`), so I did not claim it.
- `uninstall` trusts the exit code, unlike install/update which decide from disk. A post-removal `isGloballyInstalled()` check would catch "npm exited 0 but the package is still there". Left alone here.

## Update 2026-09-17: merged with main

- Merged `origin/main`. Conflicts were in `uninstall.ts` and `utils.ts`. I kept main's knip cleanup (`BundledContentRemoval` and the new topology types are no longer exported) and put the preflight back on top of main's scope rules for local and committable installs. Line numbers quoted above are from before the merge.
- The "configuration was already removed" warning now also shows when only MCP entries were removed, not just pruned skills.
- Docs: the `argent uninstall` row in `reference/cli.mdx` now describes the refusal and exit code 1.

Re-verified with packed tarballs of main and this branch. Each run used its own `HOME`, npm prefix and project, with the prefix's `@swmansion` dir set to `a-w`:

| run | exit | workspace + home | package |
|---|---|---|---|
| main, `uninstall -y --global`, blocked | **0** | 18 skills and 3 MCP configs deleted | still installed |
| branch, same | 1 | byte-identical | still installed |
| branch, healthy prefix | 0 | cleaned | removed |
| branch, global + local both installed, `--global --local`, blocked | 1 | local removed, global-scope home config kept | global kept |

Checks: build, installer vitest (590 passing), `typecheck:tests`, eslint, knip, prettier, `docusaurus build`.


packages/argent-installer/src/topology.ts
```diff
@@ -85,6 +86,81 @@ export function getGloballyInstalledPackageRoot(): string | null {
...
+
+  const parentDir = path.dirname(logicalPkgDir);
+  try {
+    fs.accessSync(parentDir, fs.constants.W_OK);
+    return { verdict: "writable", parentDir };
+  } catch (err) {
+    const code = (err as NodeJS.ErrnoException).code;
+    // Only a permission error is evidence. ENOENT and friends mean the layout
+    // is not what we assumed, which is a reason to stay quiet, not to block.
+    return code === "EACCES" || code === "EPERM"
```
latekvo, line 158: On a read-only mount `access(W_OK)` answers `EROFS`, not `EACCES`/`EPERM`, so this returns `UNKNOWN_REMOVAL` — which callers treat exactly like "writable", and the uninstall proceeds into the failure this preflight exists to prevent.

Measured on the machine I ran this on (macOS 15, sealed read-only system volume, uid 501):

```
$ node -e 'const fs=require("fs"); for (const d of ["/","/usr","/System"]) { try { fs.accessSync(d, fs.constants.W_OK); console.log(d,"writable") } catch(e){ console.log(d, e.code) } }'
/ -> EROFS
/usr -> EPERM
/System -> EPERM
$ mount | grep ' on / '
/dev/disk3s1s1 on / (apfs, sealed, local, read-only, journaled)
```

The reachable case is a Nix-managed toolchain: npm derives its global prefix from the node binary, so under Nix `<prefix>/lib/node_modules/@swmansion` is inside the store, which NixOS and nix-darwin mount read-only. #817 hit the same class from the install side and blocks on `EACCES`/`EPERM`/`EROFS` for that reason.

packages/argent-installer/src/topology.ts
```diff
@@ -85,6 +86,81 @@ export function getGloballyInstalledPackageRoot(): string | null {
...
+  // "writable" for a directory the user cannot touch. Never guess there.
+  if (process.platform === "win32") return UNKNOWN_REMOVAL;
+
+  // The rename-in-parent mechanic above is npm's. pnpm/yarn/bun globals live in
+  // their own stores and mutate different paths, so a reading taken here would
+  // not describe the command we are about to run. detectPackageManager() is the
+  // same function that BUILDS that command, so probe and command always agree.
+  if (detectPackageManager() !== "npm") return UNKNOWN_REMOVAL;
+
+  if (process.getuid?.() === 0) return { verdict: "writable", parentDir: null };
```
latekvo, line 129: The root short-circuit reports "writable" without probing anything, which is not true on the read-only-mount case above: a store path stays unwritable for uid 0, and `sudo argent uninstall --global` would take the destructive path before failing.

`sudo` is also what `elevatedRerunHint()` tells the user to run after a blocked verdict, so this is the branch that re-runs land on.

Issue #609, opened 2026-07-31
Rotated (landscape) iOS simulators: screenshots come out sideways and tap coordinates use the portrait-native frame space


scriptease:
## Summary

On a rotated (landscape) iOS simulator, `screenshot` returns a sideways image unless the caller passes a matching `rotation` — and normalized touch coordinates are interpreted in the same portrait-native space, so taps computed from an upright screenshot land in the wrong place. Both symptoms share one root cause: the bundled `simulator-server` is orientation-unaware at its HTTP/WS boundary.

## Reproduction (argent 0.18.0, macOS 26.5, iOS 26.5 iPad (A16) simulator, landscape)

1. Boot an iPad simulator and rotate it to LandscapeLeft (via the `rotate` tool or Simulator.app). The simulator itself renders correctly.
2. `POST /api/screenshot {}` against the simulator-server → returns a **1640×2360 portrait-framed PNG with the landscape content lying sideways**. `xcrun simctl io <udid> screenshot` of the same device returns a correct upright 2360×1640 image.
3. Sending the server its own `{"cmd":"rotate","direction":"LandscapeLeft"}` WS command does **not** change what `/api/screenshot` returns afterwards — the server keeps no orientation state for captures. Only an explicit `"rotation"` in the screenshot request corrects the image, and an MCP agent has no way to know the current orientation to pass it (there is no orientation query endpoint; the binary is closed-source, so this can't be fixed downstream).
4. Touch coordinates have the same problem in the opposite direction: with the sim in landscape, tapping a UI element at its upright-landscape normalized position does nothing, while tapping the position the element occupies **in the sideways portrait frame** hits it. So the moment an agent uses `rotation` to get a readable screenshot (which every agent needs — multimodal models misread sideways screenshots), screenshot space and tap space diverge, and coordinate-based taps silently miss. We suspect this is behind a class of "taps do nothing" reports on landscape-only apps.

Interestingly the stream protocol already carries the needed data — the video packet header has `display_orientation` / `display_orientation_correction` fields (that's presumably how Radon IDE renders the mirror upright) — it just isn't applied to `/api/screenshot` or the touch path, and isn't exposed to API consumers.

## Suggested fixes

Any of these would resolve it, in descending order of completeness:

1. simulator-server applies its known `display_orientation` to `/api/screenshot` output and maps incoming normalized touch coordinates from upright-orientation space to the native framebuffer — API consumers then never need to know about orientation.
[...]

PR #693, opened 2026-08-01
fix(android): capture the screen the way it is actually oriented

filip131311:
Part of #609.

## What was wrong

On a rotated Android device `screenshot` returned a portrait-framed image with the content lying sideways. `describe` frames and gesture coordinates were **already upright** — uiautomator measures against the rotated display — so the capture was the one surface out of step. An agent reading that image saw geometry that disagreed with every coordinate it was given.

Measured on a landscape Pixel_9 (API 36, `mCurrentOrientation=1`):

| | space |
|---|---|
| `describe` | upright — `"Search Settings"` centre `(0.530, 0.234)` |
| taps | upright — tapping describe's `"Display & touch"` centre opens it |
| `screenshot` (default) | **portrait-native, sideways** |

## The fix

simulator-server already applies `rotation` correctly; nothing was telling it which one. The rotation is now queried from the device per capture, so `screenshot`, `screenshot-diff`'s live captures and the auto-attached "Screen after action" screenshot all land in `describe`/tap space.

**Nothing is tracked.** A cached orientation could only be staler than the ~40 ms probe, and serving a stale one is exactly the failure being fixed. When the rotation cannot be read the request is left byte-identical to before rather than guessing — a wrong orientation is a confidently wrong image, which is worse than a sideways one.

**iOS is deliberately untouched.** There the whole surface is consistently portrait-native, so rotating only the capture would *break* an agreement rather than restore one. The iOS half of #609 needs a different change and is not in this PR.

### Also fixed: the legacy uiautomator describe path

Same root cause, worse symptom. `wm size` reports the **unrotated** size — on the landscape Pixel_9 it still answered `Physical size: 1080x2424`. Dividing rotated bounds by that does not merely squash frames: `isVisibleRect` prunes nodes whose left edge is past the screen width, so **the right-hand half of a landscape screen vanished from the tree**. The dump carries `<hierarchy rotation="N">`, so the divisor is oriented from bytes already in hand — no extra round-trip on the path that runs precisely when the helper could not be reached.

## On the rotation-name table

Pinned by measurement, never by what the names mean — the reporter of #609 found the same names are 180° inverted on iOS. `adb exec-out screencap` is rotation-aware, so it is ground truth; each candidate was scored as mean absolute difference against it both as-is and rotated 180°, so an inverted mapping could not pass:

```
rotation 1  LandscapeLeft       2.19  (vs 6.54 for its 180° twin)
rotation 3  LandscapeRight      2.19  (vs 6.34)
rotation 2  PortraitUpsideDown  2.67  (vs 13.13)
rotation 0  Portrait            identity
```

This table is the **inverse** of simulator-server's own `90° → LandscapeRight` convention, because it compensates for a rotation that layer applies while decoding the scrcpy stream. That makes it a property of a particular simulator-server build rather than of Android — which is why a table test alone is not enough, and the delivered PNG's aspect is checked against the rotation requested, falling back to an unrotated capture if they disagree.

## Verified live

Branch tool-server against a real emulator:

- landscape, default capture → **2424×1080 upright**, MAD 2.19 vs ground truth (6.54 vs its 180° twin)
- cropping `describe`'s frame for the toolbar out of that default capture shows exactly the "Display & touch" toolbar — the two spaces agree
- rotation 0 → 1080×2424, MAD 2.82, unchanged
- rotated **iPad** → still 2064×2752 sideways, i.e. iOS provably untouched

Plus 27 new unit tests, including one that asserts the legacy path *drops* the right-hand node before the fix and keeps it after. Full suite 3125 passing; lint, prettier, `typecheck:tests` and the tool-description gate all clean.

## Not in scope

- The iOS half of #609.
- `rotate` on Android silently no-ops when auto-rotate is enabled (it returned `{orientation:"LandscapeLeft"}` while `dumpsys` still read `cur=1080x2424`). Filing separately.
- `screen-recording` still records sideways — same root cause, different pipeline. Worth noting a session's screenshots are now upright while its video is not.
- Existing visual baselines captured on a rotated Android device will stop matching, since the baseline key embeds capture dimensions and those now swap. This fails loudly as a missing baseline rather than silently comparing the wrong thing — and those baselines were sideways captures of an upright UI.

---

> **#673 and #658 are both stacked on this branch** — each conflicts with this one (`screenshot/index.ts` and `screenshot-diff/index.ts` respectively) but not with each other, so both were rebased onto this branch as siblings rather than chained. Merge this first; GitHub then retargets both to `main`.

### DIFF
--- a/packages/tool-server/src/tools/describe/platforms/android/index.ts
+++ b/packages/tool-server/src/tools/describe/platforms/android/index.ts
@@ -3,7 +3,11 @@ import type { Registry, ToolDependency } from "@argent/registry";
 import type { DescribeTreeData } from "../../contract";
 import { adbExecOutBinary, isAndroidTv } from "../../../../utils/adb";
 import { resolveDevice } from "../../../../utils/device-info";
-import { getAndroidScreenSize } from "../../../../utils/android-screen";
+import {
+  getAndroidScreenSize,
+  orientScreenSize,
+  parseDumpRotation,
+} from "../../../../utils/android-screen";
 import { parseUiAutomatorDump } from "./uiautomator-parser";
 import {
   androidDevtoolsRef,
@@ -113,6 +117,12 @@ export async function describeAndroid(
       }
     );
   }
-  const tree = parseUiAutomatorDump(raw, size.width, size.height);
+  // `wm size` is not rotation-aware, but the dump says which rotation it was
+  // taken at. Orienting the divisor here is what keeps a rotated device's frames
+  // in the same upright space the android-devtools path already produces — and
+  // stops the right-hand half of a landscape screen being pruned away as
+  // off-screen (#609).
+  const oriented = orientScreenSize(size, parseDumpRotation(raw));
+  const tree = parseUiAutomatorDump(raw, oriented.width, oriented.height);
   return { tree, source: "uiautomator", hint };
 }
--- a/packages/tool-server/src/tools/screenshot-diff/index.ts
+++ b/packages/tool-server/src/tools/screenshot-diff/index.ts
@@ -5,6 +5,7 @@ import path from "path";
 import { z } from "zod";
 import { FAILURE_CODES, FailureError } from "@argent/registry";
 import type {
+  DeviceInfo,
   FileInputSpec,
   ServiceRef,
   ToolContext,
@@ -14,6 +15,7 @@ import type {
 import { simulatorServerRef, type SimulatorServerApi } from "../../blueprints/simulator-server";
 import { resolveDevice } from "../../utils/device-info";
 import { httpScreenshot } from "../../utils/simulator-client";
+import { captureScreenshotUpright } from "../../utils/rotation-aware-capture";
 import { requireArtifacts, type ArtifactHandle } from "../../artifacts";
 import { diffPngFiles } from "./screenshot-diff";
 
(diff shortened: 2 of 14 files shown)


filip131311:
Live QA on a tgz built from 2447c48 merged with main (iOS 26.5, Pixel_9 API 36 emulators, tvOS 26.5, Electron/CDP):

- Android default capture upright at surface rotation 0/1/2/3 (dims match `screencap`, MAD 2.3–2.9 vs ground truth, 10–17 vs the 180° twin); explicit `rotation` still wins; legacy uiautomator `describe` path returns the same frames as the devtools path on a landscape device (right half no longer pruned); screenshot-diff `captureCurrent` vs upright baseline 0 %; MCP auto-screenshot upright.
- ~90-call regression sweep over the unrelated tool surface (gestures, keyboard, flows, await-*, run-sequence, permissions, profilers, debugger, chromium-*, tvOS, scoped stop-simulator-server) — no regressions; iOS/tvOS/Chromium captures byte-for-byte the old path.
- CI-equivalent suite on the merge commit green (4386 tool-server tests, lint, knip 175/215).
- Cost (A/B vs a main-only build, median of 30 × 4 rounds): +~1.2 ms per Android screenshot with the android-devtools helper running, +~9.5 ms before anything has started it (adb-socket probe), 0 with explicit `rotation`, 0 on iOS.

Known, pre-existing, not introduced here: on a rotated Android device the gesture tools still take normalized coords in the portrait frame, so `describe` frames ≠ tap coords in landscape (identical on main). Follow-up: rotate gesture input by the same surface rotation.

PR #699, opened 2026-08-01
fix(cli): let a boolean flag take the value you typed

### LINKED ISSUE #586: Boolean flags silently ignore a space-separated value (--includeChildren false stays true)
I ran into this on 0.15.0. I called

```
argent run native-find-views --udid <udid> --bundleId <app> --className UIButton --includeChildren false --includeAncestors false
```

and got matches with full ancestor chains and children anyway, which at first read as the className filter being broken because the output was huge.

The filters are fine though. The problem is the flag parser: `--includeChildren false` doesn't error and doesn't set the flag to false, the token gets swallowed and the flag keeps its default of true. The `=` form works:

```
--includeChildren=false --includeAncestors=false   # works, no ancestors/children keys in the output
--includeChildren false --includeAncestors false   # silently true, both keys present
```

Since `--help` prints the flags as plain `--includeChildren    boolean` with no hint that `=` is required, the space form looks legal and nothing tells you it was ignored. Accepting the space-separated value would be nicest, but even erroring on the stray token would have saved me a while of chasing a filter bug that didn't exist. I only tested these two flags on native-find-views but I'd guess it applies to every boolean flag.

filip131311:
Fixes #586.

## The bug

`--includeChildren false` left the flag **true** and silently dropped the word, so the tool ran with the opposite of what was asked. The reporter spent a while chasing a filter bug that did not exist — the output was simply the unfiltered kind.

Reproduced on 0.18.1, device-free, with a discriminator that shows the flag's value directly:

```
$ argent run screenshot-diff --udid U --baselinePath A.png --currentPath B.png --captureCurrent=false
{ "summary": "Screenshot diff summary … pixel_mismatch: 1.79% …"      <-- false, diff runs

$ argent run screenshot-diff --udid U --baselinePath A.png --currentPath B.png --captureCurrent false
[Tool:screenshot-diff] Provide either currentPath or captureCurrent, not both.   <-- stayed TRUE
```

That message is the whole problem in one line: the user typed `false`, and the tool refuses on the grounds that they asked to capture.

## Why the old behaviour existed, and why it does not need to

The parser deliberately declined to look ahead:

> `// Bare boolean flag: true. Don't consume next argv to avoid surprising behavior when a positional follows.`

That protects a case which does not exist. `parseFlags` has exactly one caller, and **`positional` is never read anywhere in the repo** — every bare token is collected and dropped, for booleans and for every other type. There is no positional to steal. The module header had documented the space form as legal all along:

```
//   --name value          (string / number / integer / boolean-with-value)
```

This matters because several tool booleans **default to true**, so `false` is the only way to say what you mean: `screenshot.includeImageInContext`, `network-request.includeBody`, `debugger-component-tree.onScreenOnly`, `debugger-inspect-element.resolveSourceMaps`.

## What changed

A boolean flag consumes a following token when it is exactly the word `true`/`false`, case-insensitively. `1`/`0` and everything else are left alone — a bare number after a switch is more likely an unrelated argument than a value, and guessing at an ambiguous token is what caused this bug. `--flag -- false` still keeps the word positional.

Three consequences are handled here rather than left to bite later:

- **`argent run` now reports arguments it ignored.** This is what stops `--flag 0` and `--flag yes` being the same silent inversion, in a syntax users are now being taught works. stderr, so `--json` stays parseable.
- **`--no-flag true` / `--no-flag false` now name the positive form** instead of quietly meaning false. One is a contradiction, the other a double negative.
- **`--flag=True` used to throw** while `--flag True` would now work — the inline coercion was case-sensitive and did not trim. Both paths now share one helper, so a single word cannot get two answers one function apart.

The **flow** and **lens** parsers take switches that accept no value at all, and would have silently ignored the syntax their users just learned elsewhere. They now refuse it — in the flow parser that also stops `false` being swallowed as the flow name, since its first bare token is the flow to run.

`--help` gains **one legend line** for tools with a boolean field, rather than widening the flag column on every field of every tool:

```
  Booleans: --flag or --flag true sets true; --flag false, --flag=false, or --no-flag sets false.
```

It deliberately does not begin with `--`. `scripts/e2e-full/lib/discover-tools.sh` reads any such line in the `Flags:` section as a flag row and takes the first `--token` as its name, so a legend starting with a flag would inject a phantom flag into every tool model it builds. **There is a test pinning that shape**, verified live: running the harness's awk over the new help output yields exactly the real flags and no legend row.

## Verified live

```
--captureCurrent false  -> diff runs (was: "Provide either currentPath or captureCurrent, not both.")
--captureCurrent=false  -> unchanged
--captureCurrent 0      -> "Note: ignoring unused argument(s): 0. Pass values as --flag <value> or --flag=<value>."
screenshot --help       -> legend present; harness awk parses only the real flags
```

25 new tests; 8 confirmed to fail before and pass after. The rest pin what deliberately did *not* change: bare flag still true, a following flag or non-boolean word left alone, `1`/`0` still positional in the space form, `--` escape, string and unknown fields untouched, `--no-flag` semantics, boolean arrays, last-write-wins. The parser had **zero** boolean coverage before — which is why this survived. Full CLI suite 298 passing; lint, prettier and `typecheck:tests` clean.

## Scope and neighbours

Only `flag-parser.ts` (the schema-driven parser for `argent run`) changes semantics. The other CLI parsers take fixed, default-off switches where `false` has no coherent meaning; three of them already reject a stray token outright, and the two that did not now do. `flow.ts`'s deliberate rejection of `--flag=value` is untouched.

**#648** also edits `flag-parser.ts` (rewrites `renderFlagName`, inserts helpers at the same anchor). Behaviourally compatible — it keeps boolean → bare `--name` — so this is a textual conflict only, and either order works.

## Found while here, not fixed

`scripts/e2e-full/lib/discover-tools.sh:60-64` detects flag kinds with `\bboolean\b` in awk, where `\b` is **backspace**, not a word boundary — so it never matches and every non-enum flag is typed `unknown`. That makes the `bad-type` validation case downstream unreachable. Pre-existing and unrelated to this fix; filing separately.

---

> **Stacked on #648** (`filip/run-cli-validation-messages`). Both rewrite the same region of `packages/argent-cli/src/flag-parser.ts` — #648 factors out `flagNameFor`/`isJsonField` and rewrites `renderFlagName`, this one adds `booleanWord` beside them and a boolean legend to `formatSchemaUsage`. Merging them independently conflicts; this branch is rebased on #648, so review only the top commit. GitHub retargets it to `main` when #648 merges.

### DIFF
--- a/packages/argent-cli/src/flag-parser.ts
+++ b/packages/argent-cli/src/flag-parser.ts
@@ -5,7 +5,7 @@
 //   --name value          (string / number / integer)
 //   --name=value
 //   --name                (boolean: true)
-//   --name true|false     (boolean: explicit value — only these two words are consumed)
+//   --name true|false     (boolean: explicit value — also 1|0; only these are consumed)
 //   --no-name             (boolean: false)
 //   --name a --name b     (array of scalars)
 //   --name-json '<json>'  (arbitrary nested object/array — escape hatch)
@@ -65,14 +65,18 @@ export function flagNameFor(name: string, prop: JsonSchema | undefined): string 
 }
 
 /**
- * `true`/`false` as a standalone word, case-insensitive and whitespace-tolerant.
- * `undefined` for anything else, including `1`/`0` — a bare number following a
- * switch is far more likely to be an unrelated argument than a value.
+ * A token that can only have been meant as a boolean value: `true`/`false` or
+ * `1`/`0`, case-insensitive and whitespace-tolerant. `undefined` for anything
+ * else, so an ambiguous token is left alone rather than guessed at.
+ *
+ * The single source of truth for every form — the `--flag <value>` lookahead,
+ * `--flag=<value>`, boolean array items, and the `--no-flag` contradiction
+ * guard — so the same word cannot mean different things one call site apart.
  */
-function booleanWord(raw: string): boolean | undefined {
+function booleanLiteral(raw: string): boolean | undefined {
   const value = raw.trim().toLowerCase();
-  if (value === "true") return true;
-  if (value === "false") return false;
+  if (value === "true" || value === "1") return true;
+  if (value === "false" || value === "0") return false;
   return undefined;
 }
 
@@ -95,16 +99,11 @@ function coerceScalar(raw: string, type: string | undefined, field: string): unk
     return n;
   }
   if (type === "boolean") {
-    // Shares booleanWord with the bare-token lookahead, so `--flag=True` and
+    // Shares booleanLiteral with the bare-token lookahead, so `--flag=True` and
     // `--flag True` cannot disagree about the same word.
-    const word = booleanWord(raw);
-    if (word !== undefined) return word;
-    // 1/0 only where the value is unambiguously attached (`--flag=1`, array
-    // items); the lookahead stays narrower on purpose.
-    const value = raw.trim();
-    if (value === "1") return true;
-    if (value === "0") return false;
-    throw new FlagParseException(`--${field} expected true/false, got "${raw}"`);
+    const value = booleanLiteral(raw);
+    if (value !== undefined) return value;
+    throw new FlagParseException(`--${field} expected true/false (or 1/0), got "${raw}"`);
   }
   // string or unknown: pass through
   return raw;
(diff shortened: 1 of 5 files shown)


filip131311:
Added `1`/`0` to the space form as a second commit (aacb1eb), per review.

`--flag 0` previously set the flag **true** and dropped the `0` — the same silent inversion as `--flag false`, which is the bug this PR exists to fix. There was no reason for 1/0 to be treated as more ambiguous: `argent run` takes no positionals after the tool name, and a value belonging to any other flag is consumed by that flag, so a bare `1`/`0` following a boolean switch has no other possible meaning. The old comment about it being "more likely an unrelated argument" described an argument that was being discarded anyway.

`booleanWord` became `booleanLiteral` (the name no longer fit) and is now the single reader for all four sites — the lookahead, `--flag=<value>`, boolean array items, and the `--no-flag` contradiction guard. That closes the split where `--flag=1` worked but `--flag 1` did not.

Unchanged on purpose: `--flag notabool` is still left alone and still warned about — genuinely ambiguous, which `1`/`0` is not. `--flag -- 0` still keeps `0` positional.

Verified through the built CLI, not just the unit tests:

```
--captureCurrent 0          -> false   positional=[]
--captureCurrent 1          -> true    positional=[]
--captureCurrent false      -> false   positional=[]
--captureCurrent            -> true    positional=[]
--captureCurrent=0          -> false   positional=[]
--captureCurrent notabool   -> true    positional=["notabool"]
--no-captureCurrent         -> false   positional=[]
--captureCurrent -- 0       -> true    positional=["0"]
```

Help legend now reads:

```
  Booleans: --flag, --flag true, or --flag 1 sets true; --flag false, --flag 0, --flag=false, or --no-flag sets false.
```

Still does not start with `--` after its indent, which `scripts/e2e-full/lib/discover-tools.sh` requires — pinned by a test.

PR #704, opened 2026-08-03
fix(e2e): make the debugger assertions fail when the debugger is not connected

latekvo:
Found while reviewing #610 and set aside as out of scope for that review: the weaker of the two assertions is a live defect on `main` today, the other becomes one when #610 lands.

## The defect

`assert_ok` (`lib/common.sh:122-130`) records PASS on `RT_RC == 0` and never reads `RT_JSON`. Both tools on lines 84 and 86 answer **HTTP 200 while reporting that the debugger is not connected**, so the two assertions that exist to check the debugger chain pass over a dead session.

**Line 84 is broken on main today.** Half-close the CDP socket server-side (the socket lands in CLOSING with no `close` event, so no terminated cascade) and the cached service still resolves:

```
ground truth: {"connected":false}

=== shipped assertions (lines 84/86) ===
  ✓ debugger-status status
  ✓ debugger-log-registry logs
=== patched assertions ===
  ✗ debugger-status [status] expected .connected to be true, got 'false'
  ✓ debugger-log-registry logs
```

**Line 86 breaks once #610 lands**, which turns the unreachable-runtime 500 into a 200 `not_connected` payload. Same two lines against a build of that branch with nothing listening on the Metro port:

```
ground truth: {"status":"not_connected","connected":false,"reason":"metro_not_running"}

=== shipped ===   ✓ debugger-status status   ✓ debugger-log-registry logs
=== patched ===   ✗ debugger-status [status] expected .connected to be true, got 'false'
                  ✗ debugger-log-registry [logs] expected (.status // "connected")=connected got 'not_connected'
```

## Why these two jq paths

Measured against live tool-servers built from both `main` and #610, in both the connected and the not-connected state:

| | main connected | main dead socket | #610 connected | #610 not-connected |
|---|---|---|---|---|
| `debugger-status` | `.connected=true` | `.connected=false` | `.connected=true` | `.connected=false` |
| `debugger-log-registry` | **no `.connected`** | - | **no `.connected`**, `.status="connected"` | `.status="not_connected"` |

`.connected` is absent on log-registry's success path in *both* builds, so `assert_true '.connected'` would fail a healthy run - that is why the second line uses `(.status // "connected")`, which is `"connected"` on main (field absent) and on #610's success, and `"not_connected"` on its failure. Both assertions therefore work on main and keep working after #610.

No new helper: `assert_true` / `assert_field` already exist and the harness already gates `launch-app '.launched'`, `restart-app '.restarted'`, `gesture-tap '.tapped'` the same way. `debugger-status`'s `.connected` was the one flag of that shape left bare.

## Verification

Repro is a real tool-server per build plus the mock Metro/CDP target, driving the assertion lines lifted verbatim out of the patched phase file. The unpatched `assert_ok` passes in **both** states - that is the mutation control. Timing note for anyone re-running it: the half-closed socket self-heals after ~30s when the `ws` close timeout fires, so steps 5 and 6 have to run back to back.

## Deliberately not in this PR

The same sweep found ~20 sibling `assert_ok` sites whose tool reports failure inside a 200 payload - `await-ui-element` `{success:false}`, `await-screen-idle` `{settled:false}`, `run-sequence` `{completed:0}` with the error in `steps[]`, `stop-metro` (whose entire body is inside a `try/catch` returning `{stopped:false}`, so `assert_ok` on it can never fail), `chromium-cookies set` `{set:false}`, `react-profiler-status` `{session_status:"no_react_runtime"}`, and the string-returning tools that answer `"No network traffic captured..."` at 200. Those verdicts are read off each tool's result contract, not driven live - confirming them needs the Android and Chromium tiers - so they are left for #595, which is already rewriting this harness with the same intent.

One adjacent bug worth recording while it is fresh: `50-rn-bluesky.sh:96` extracts `RID` with `jq '(.requests // .entries // [])[0].requestId'` from `view-network-logs`, which returns a formatted **string**, not an object. `RID` is therefore always empty, line 98 is dead and line 100's skip is always taken.

Merges cleanly with #595 (test-merged: `Automatic merge went well`; that PR rewrites the `local D=` line above but touches neither assertion).


scripts/e2e-full/phases/50-rn-bluesky.sh
```diff
@@ -81,9 +81,14 @@ run_phase() {
   local D="\"device_id\":\"$LID\",\"port\":$MPORT"
 
   # --- debugger chain -------------------------------------------------------
-  assert_ok    "$P" debugger-status status "{$D}"
+  # Both of these answer 200 while reporting that the debugger is NOT connected,
+  # so the exit code alone says only that the server replied. debugger-status
+  # carries `connected`; debugger-log-registry does not, and reports the same
+  # state through `status` (absent on a healthy call, "not_connected" when the
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Added comment says debugger-log-registry's `status` is absent on a healthy call, but on main it is `"connected"`.

The new comment describes a `debugger-log-registry` response shape that matches neither build. Its first half (`status` "absent on a healthy call") only holds at this PR's base commit 3fff176a, where the tool returns `LogStats` with no `status` field at all — but in that same build there is no `"not_connected"` payload either (the tool resolves its debugger through `services: (params) => ({ debugger: debuggerServiceRef(params) })`, so an unreachable runtime is a service-resolution failure, i.e. a non-200). Its second half (`"not_connected" when the runtime is unreachable`) only holds after #610, which is already merged into origin/main (`git merge-base --is-ancestor 4975ff49 origin/main` succeeds) — and there the success path explicitly sets `status: "connected"`, so `status` is NOT absent on a healthy call. The comment is the only stated rationale for the `// "connected"` default in the jq filter on line 91, and it tells a reader the opposite of which branch of that default actually fires once this lands on main (the default becomes unreachable).

**Trigger.** After this PR merges into main (which contains #610), a maintainer reads line 86-88, greps `packages/tool-server/src/tools/debugger/debugger-log-registry.ts`, and finds `status: "connected" as const` on the success path — contradicting "absent on a healthy call". Conversely, at the PR head there is no code path that ever produces `status: "not_connected"` from that tool, so the second half of the parenthetical points at something the reader cannot find either.

**Checked against.** Added prose, scripts/e2e-full/phases/50-rn-bluesky.sh:86-88: `# carries \`connected\`; debugger-log-registry does not, and reports the same` / `# state through \`status\` (absent on a healthy call, "not_connected" when the` / `# runtime is unreachable).` — Compared against `git show origin/main:packages/tool-server/src/tools/debugger/debugger-log-registry.ts`, success path: `const response: LogRegistryResponse = { status: "connected" as const, ...stats, clusters, ... }` and its interface `interface LogRegistryResponse extends LogStats { status: "connected"; ... }`.

latekvo (commented):

scripts/e2e-full/phases/50-rn-bluesky.sh
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: Fixed in b4dca80a3. You are right — post-#610 the success path always returns `{ status: "connected", ... }`, so `status` is not "absent on a healthy call"; the only build where that field was missing is the pre-#610 base, which is exactly why the `// "connected"` default exists. Rewrote the comment to state the actual shape ("connected" on a healthy call, "not_connected" when the runtime is unreachable) and to explain that the default covers the pre-#610 field-absent shape.

latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: Follow-up in 77cd12c5a. On a closer read, the log-registry assertion is a no-op on this (pre-#610) base — the tool returns LogStats with no `status` field, so `(.status // "connected")` always yields "connected". It only becomes a real check once #610 (already on main) is in the tree. That is not a defect requiring code: the phase still fails correctly on a dead session because debugger-status (the `.connected` gate) runs immediately before and catches even a half-closed socket; and this PR merges into main, so the landed log-registry line does get the `status` field. Refined the comment to say debugger-status is the dead-session gate and the log-registry line is inert until #610 — rather than implying both lines fail a dead session equally.

PR #709, opened 2026-08-03
test(chromium): stop electron boot tests depending on host CDP ports

latekvo:
## Problem

`boot-electron-spawn-error.test.ts` drove `bootElectronApp` against hardcoded CDP ports 19222-19226 and relied on nothing answering there.

## It is not load, and it is not the timeout

I chased the obvious hypothesis first — vitest's 5000 ms default colliding with a built-in wait — and it is wrong. Measured at `origin/main` before any edit: 15/15 isolated runs, 10/10 concurrent copies, 3/3 full-suite runs at load avg 16, and 10/10 under sustained load at 19-20. **38/38, a 0% spontaneous failure rate.** A later hostile re-run added 15/15 more under 24 CPU burners on 14 cores (load 15→27).

## The real trigger, and it is deterministic

Anything on the host answering the hardcoded CDP ports. Squatting them with an HTTP server that serves `/json/version` fails the file every time. At `origin/main`:

```
× rejects with a clear, actionable message when spawn emits ENOENT
AssertionError: promise resolved "{ platform: 'chromium', …(5) }" instead of rejecting
+   "id": "chromium-cdp-19223",
Tests  1 failed | 8 passed (9)
```

9/9 squatted runs on `main` failed, always the same tests; 0/20 failed with the ports free.

**Mechanism.** When a real endpoint answers, `waitForCdpReady` wins `bootElectronApp`'s `Promise.race` in ~1 ms, long before the test's 10 ms `setTimeout` fires its synthetic error. The boot then completes and *resolves*, so the rejection the test asserts never arrives.

The source is correct throughout. The test bound a global shared resource it did not own. `chromium-cdp-19222` is argent's canonical device id (`src/utils/chromium-discovery.ts:9`) and `e2e-device-smoke.yml:139,173` boots a real Electron on exactly that port — a developer with an Electron app open, or a second agent on the same machine, is enough.

## Fix

One `UNREACHABLE_CDP_PORT` constant pinned to port 1, which `ensureCdpReachable` can never reach: it dials through Node's `fetch`, which refuses port 1 as a WHATWG **"bad port"** before opening a socket. Measured with `net.connect` hooked, counting sockets per call:

| port | result | sockets opened |
|---|---|---|
| **1** | `fetch failed` → cause `bad port` | **0** |
| 2 | `connect ECONNREFUSED 127.0.0.1:2` | 2 |
| 22 | `bad port` | 0 |
| 1023 | `connect ECONNREFUSED 127.0.0.1:1023` | 2 |

Identical on Node 20 (what `unit-tests.yml` runs), 22, 24 and 26. A `MockAgent` explicitly intercepting `http://127.0.0.1:1/json/version` still gets `bad port` — the check precedes dispatch, so nothing can answer it.

Privilege is *not* the mechanism and would not be sufficient: 2 and 1023 are equally privileged yet dial normally, and Windows has no privileged range at all. The JSDoc says so, so the constant cannot be "tidied" below 1024 and quietly reinstate the flake. Not a timeout widening.

## Pinning the ports-file guard

The `vi.mock("../src/utils/chromium-discovery", …)` that stops a completed boot persisting its port into the developer's real `~/.argent/chromium-cdp-ports.json` is **already on `main`** — it landed with #714, and against the current merge base it is context here, not an addition.

What was missing is any test that notices when it stops working. This PR adds that: the file scopes `HOME` with #714's own `scopeTempHome()` helper and asserts, after the successful boot, that no `~/.argent/chromium-cdp-ports.json` exists. Deleting the mock, or replacing it with a call-through `vi.fn(actual.trackChromiumPort)` variant, previously left the suite green at 9/9 while a real ephemeral port was written; both now fail.

## Sibling fixed

`chromium-discovery.test.ts` asserted port `43211` was unreachable, but that sits inside Linux's ephemeral range (`ip_local_port_range` → `32768 60999`), so a `listen(0)` fake CDP server — including this file's own — can legitimately be handed it. It now shares one named `UNREACHABLE_PORT`, replacing that literal and a second bare `1` already in the file.

**Left alone, verified safe:** `boot-electron-kill.test.ts`, `flows/flow-chromium-boot.test.ts`, `list-devices.test.ts` — each owns its port via `listen(0)` or fully mocks the launcher. The fixed ports `43210`, `43212`, `43213`, `54321`, `54322` (all in `chromium-discovery.test.ts`) are only ever tracked or untracked, never network-probed.

## Verification

- Repro above reproduced at `origin/main`; on this branch, 20 runs with all five of 19222-19226 squatted, plus `--pool=threads`, `--sequence.shuffle`, `--no-file-parallelism` and `--maxWorkers=1` variants: **80 runs, 0 failures**.
- **Mutations, each of which passed before:** delete the module mock, replace it with a call-through variant, swap `CHROMIUM_ELECTRON_CDP_TIMEOUT` for another code at the throw site, and point the catch-path test at a nonexistent `appPath` — all four now fail. The catch-path test pins the failure code rather than the message, matching `boot-electron-ready-race.test.ts:136`; that code had no assertion anywhere in the repo before.
- Full `packages/tool-server` under a redirected `HOME` creates no `.argent` directory at all.
- `eslint` 0, `prettier --check` 0, `tsc --build` 0, `tsc -p tsconfig.test.json` 0 — exit codes read directly.


packages/tool-server/test/boot-electron-spawn-error.test.ts
```diff
@@ -17,6 +17,24 @@ import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
 
+/**
+ * CDP port for the boots below that must never reach a live endpoint.
+ *
+ * Every test that drives a spawn-error / early-exit / no-pid path needs
+ * `waitForCdpReady` to keep failing, so the readiness probe cannot win
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The new port rationale covers the no-pid path, which never reaches the readiness probe it describes.

The JSDoc on UNREACHABLE_CDP_PORT states that *every* test driving a spawn-error / early-exit / **no-pid** path needs `waitForCdpReady` to keep failing so the readiness probe cannot win `bootElectronApp`'s `Promise.race`. That holds for the spawn-error and early-exit paths, but not for the no-pid one: `bootElectronApp` throws synchronously inside `if (!child.pid)` before the `Promise.race` is ever constructed, and it removes its own `'error'` listener on the way out. For that third path the port value is irrelevant, so the stated reason does not apply to the case it names.

**Trigger.** A reader following this rationale to add a fourth no-pid case concludes the CDP port choice is load-bearing for it. It is not: the no-pid branch returns before any readiness probe is started, so the case would behave identically on a reachable port.

**Checked against.** packages/tool-server/test/boot-electron-spawn-error.test.ts:19-35 (the JSDoc, added by this PR) against packages/tool-server/src/tools/devices/boot-electron.ts:339-357, where `if (!child.pid)` calls `child.removeListener("error", spawnErrorListener)` and throws a FailureError before the Promise.race further down the function.

latekvo (commented):

The observation below was confirmed by reading the cited code paths directly, against the pre-PR file and the rest of the suite.

packages/tool-server/test/boot-electron-spawn-error.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Qwen3.8 Max\]: Fixed in 49019d26e56e40b989214504507fd6e3dd4939c0. You're right: the no-pid branch throws synchronously in `boot-electron.ts` (`if (!child.pid)` → `throw new FailureError(...)`) before the `Promise.race` / readiness probe is ever constructed, so the port value is not load-bearing for that path. I reworded the JSDoc so only the spawn-error and early-exit paths are claimed to need the unreachable port for the race, and noted the no-pid tests keep the constant for host-independence rather than a race reason. Verified at the new head: `boot-electron-spawn-error.test.ts` 9/9 and `chromium-discovery.test.ts` 11/11 pass; prettier clean.

latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Re-verified independently at the current head `ad259244b0277eb4e6a248d44dcaafba6eca262f`, not just taking the earlier reply at its word.

What I checked: `packages/tool-server/src/tools/devices/boot-electron.ts` — `if (!child.pid)` is at line 430, it calls `child.removeListener("error", spawnErrorListener)` at 437 and throws at 439; `detachBootListeners` is defined at 492 and the `Promise.race([waitForCdpReady(...), earlyExit, spawnError])` at 500-505. The no-pid throw therefore returns 60+ lines before the race is constructed, exactly as reported.

The JSDoc at `boot-electron-spawn-error.test.ts:23-30` now scopes the race rationale to the spawn-error and readiness-failure tests and states outright that the no-pid tests do not depend on it, keeping the constant only for host-independence. The misdirection this thread named is gone. Tests re-run at this head: `boot-electron-spawn-error.test.ts` + `chromium-discovery.test.ts` → 20/20 pass.

Resolving.

PR #713, opened 2026-08-03
fix(tools-client): make the server identity guard survive a clipped `ps` and a sanitized PATH

latekvo:
## Problem

`processCommandMatches` reads a tool-server's command line to confirm a pid is one of ours before signalling it — the guard that stops kill-before-respawn from killing an unrelated process that inherited a recycled pid.

It read that command line with:

```ts
execFileSync("ps", ["-p", String(pid), "-o", "command="], …)
```

Two things are wrong with that call, and they fail the same way.

**1. procps-ng truncates `ps -o command=` to `$COLUMNS`.** When the recorded bundle path does not fit that width, the command line comes back clipped, never matches its own marker, and the function returns `false`.

**2. A bare `"ps"` resolves through `PATH`.** An MCP server launched from a GUI / launchd context inherits a sanitized `PATH` that omits `/bin`, where the spawn `ENOENT`s — and the `catch` turns that into the same `false`, on *every* call rather than only a long path.

`false` is the fail-safe answer — *don't kill something you can't identify* — which is right when `ps` is genuinely missing or unsupported. In both cases above it is wrong, and it fails in the expensive direction: the kill is skipped, a second server spawns, and **the live one is orphaned**.

Measured on Linux (procps-ng 4.0.4, stdout piped, same pid):

```
COLUMNS=<unset>  plain_len=157  ww_len=157
COLUMNS=40       plain_len=40   ww_len=157
COLUMNS=80       plain_len=80   ww_len=157
```

and, replicating the guard's own `execFileSync` with `PATH="/usr/local/bin"`:

```
bare "ps"  → throws ENOENT
"/bin/ps"  → OK
```

## The twin already had both

`packages/tool-server/src/utils/vega-process.ts` pins the width flag in both its arg sets:

```ts
export const PS_ARGS = ["-A", "-ww", "-o", "command="] as const;
export const PS_ARGS_WITH_PID = ["-A", "-ww", "-o", "pid=,command="] as const;
```

…and resolves the binary absolutely:

```ts
export const PS_BIN = ["/bin/ps", "/usr/bin/ps"].find((p) => existsSync(p)) ?? "ps";
```

That `PS_BIN` pin landed in #508, which fixed the identical sanitized-`PATH` failure across all four `ps` spawn sites in `tool-server` (a VVD silently listed twice). `launcher.ts` lives in a different package and was the one site left behind, so this PR takes both halves. `grep '"ps"'` over the tracked sources now returns no bare-`ps` production call site anywhere in the repo.

## Why no existing test caught it

Six tests across `launcher-sweep`, `launcher-duplicate-spawn` and `launcher-version-gate` do fail without `-ww` — but only when `COLUMNS` happens to be narrow in the environment running them. Nothing sets it, so on a normal-width terminal or in CI the whole suite is green either way. Nothing at all covered the `PATH` half.

## The tests

Four, all in `launcher-sweep.test.ts`. Each spawns a real `node <bundle> start` keepalive, writes a state record, and runs `sweepDeadStateFiles()`.

**`terminates a live server whose bundle path is longer than the terminal is wide`** narrows `COLUMNS` to 40 for the duration (the `execFileSync` child inherits it) and restores the ambient value afterwards, distinguishing "was unset" from "was set to something".

It **skips where `ps` ignores `$COLUMNS`**. BSD `ps` clips only to a *terminal* width, and the guard hands it `stdio: ["ignore", "pipe", "ignore"]` — never a terminal — so on macOS this assertion holds against the un-hardened guard too. `psClipsToColumns()` probes that with the guard's own binary and options against the live pid, rather than reading `process.platform`, so the test says "this platform cannot observe the bug" instead of passing vacuously. On CI (`unit-tests.yml` → `ubuntu-latest`) it runs: the job log shows `test/launcher-sweep.test.ts (10 tests)`, none skipped.

Its fixture is also sized past **132**, not just past `COLUMNS`. procps's single `-w` widens to `max($COLUMNS, 132)` rather than lifting the limit, so a shorter command line cannot tell `-ww` from a one-character regression to `-w` — straight back to the orphaning bug, green. The test asserts `` `node ${bundle} start`.length > PS_SINGLE_W_FLOOR ``, so a future fixture rename that shortened the path fails rather than quietly making it vacuous.

**`terminates a live server when PATH omits every directory holding \`ps\``** points `PATH` at a directory that does not exist for the duration and restores it the same way. It discriminates on every platform, macOS included.

**Two look-alike tests** cover the guard's structural match, which is what `-ww` makes load-bearing: command lines the width limit would have cut off before they could collide now reach the regex whole, and neither half of it was pinned — both could be deleted with the suite green.

- `does not signal a process that merely mentions the bundle path` — a live process carrying the recorded path in its argv without running it. Red when the trailing `` ` start` `` requirement is dropped, which would SIGTERM a user's `tail -f <bundlePath>`.
- `does not signal a different install whose bundle path ends with ours` — a second install nested under a longer prefix. Red when the leading `(?:^|\s)` boundary is dropped, which would SIGTERM a live server another session is using.

Both record a bundle path that is **not** on disk and assert it, because `sweepDeadStateFiles` short-circuits on one that is and would never reach the guard. That assertion is the pin: with a shared fixture path and only declaration order keeping them apart, swapping the two blocks made the first pass against a guard that had lost the exact behaviour it names.

## Verification

Six mutations, each redding **exactly one** test and leaving the other nine green — which is why each test had to be added. Linux runs are `node:24-bookworm`, procps-ng 4.0.2 (baseline there: `10 passed (10)`).

| mutation | red test | where |
|---|---|---|
| drop `-ww` | `…longer than the terminal is wide` | Linux |
| `-ww` → `-w` | `…longer than the terminal is wide` | Linux |
| `PS_BIN` → bare `"ps"` | `…PATH omits every directory holding \`ps\`` | Linux + macOS |
| drop the regex's `(?:^\|\s)` anchor | `…install whose bundle path ends with ours` | macOS, in **both** test orders |
| drop the regex's `` ` start` `` tail | `…merely mentions the bundle path` | macOS, in **both** test orders |
| ps `timeout: 2_000` → `1` | 3 tests (macOS) / 4 (Linux) | both |

- Unmutated: Linux `10 passed (10)` — nothing skipped, so CI gates all of it; macOS `9 passed | 1 skipped (10)`.
- The probe was checked against the environment it exists for: with `PATH` holding only node's directory (no `ps`, `/bin/ps` present), the suite is `9 passed | 1 skipped`; resolving the probe through `PATH` instead of `PS_BIN` there gives `Error: spawnSync ps ENOENT`, `1 failed`.
- Package suite: `12 files, 194 passed | 1 skipped (195)` on macOS.
- `eslint --max-warnings 0` 0, `prettier --check` 0, `tsc --build` 0, `tsc -p tsconfig.test.json --noEmit` 0, `npm run knip` 0 — exit codes read directly, after `npm run build` (the type-aware lint rules report spurious errors against an unbuilt workspace).

## Note on the tests' timeout

`waitForExit`'s window exceeds vitest's 5 s default, so the tests carry an explicit `20_000`. That is what makes a mutation fail as `expected false to be true` — an unterminated server — rather than as an uninformative bare timeout.

## Found alongside, filed separately

Out of this PR's scope, all reproduced: #832 (on Windows, `ensureToolsServer`'s kill-before-respawn is the one of three call sites without the `win32` guard, so the wedged server is never killed before the respawn), #833 (`LOCK_WAIT_TIMEOUT_MS`'s budget comment counts one `ps` where the critical section runs three, so its "worst case ≈ 26s" is not a bound), and #834 (`stop-metro` swallows a missing `lsof` into `stopped: false`, reporting "port is free" when it could not look — the same PATH-resolved-silent-failure class this PR fixes for `ps`).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of the correct background server process, including long command lines and shell-wrapped processes.
  - Prevented unrelated or similarly named processes from being stopped during cleanup.
  - Improved recovery when an automatically started server becomes unresponsive or cannot be verified.
  - Added more reliable handling for restricted environments, unusual installation paths, and processes with similar names.
  - Improved diagnostics when process verification encounters command-line inspection errors.


packages/argent-tools-client/src/launcher.ts
```diff
@@ -727,7 +739,14 @@ function processCommandMatches(pid: number, marker: string | undefined): boolean
     // so a bundle path containing spaces still matches its own ps output.
     const escaped = marker.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
     return new RegExp(`(?:^|\\s)${escaped} start(?:\\s|$)`).test(cmd);
-  } catch {
+  } catch (err) {
+    // Fail safe — don't kill what we can't identify — but say why: a fallback
+    // `"ps"` under a sanitized PATH ENOENTs here forever, and an undiagnosed
+    // miss orphans every live server silently. Same visibility tool-server's
+    // vega-process probes give their callers.
+    process.stderr.write(
```
hubgan, line 690: The `stdio` array on the `ps` call discards the child's stderr, so this line can never carry `ps`'s own complaint. Measured: with stderr ignored a non-zero exit stringifies to just `Error: Command failed: /bin/ps -ww -p 1 -o command=`; with stderr piped the identical case reads `Error: Command failed: ... \nps: invalid option -- 'w'`. A flag rejection is precisely the new failure mode that adding `-ww` risks, and it is the one failure this message cannot name. The vega-process probes the comment points at use promisified `execFile`, whose rejection message does append the child's stderr - that is where their diagnostic value comes from. `ENOENT` and `ETIMEDOUT` still surface, because those live on the Error object rather than in the child's output.

The parenthetical also asserts a cause that is wrong for every failure other than a mid-check exit, and systematically wrong on Windows. Line 943 is the one call site without the `win32` guard used at 562 and 686; there `PS_BIN` falls back to bare `"ps"` and always `ENOENT`s, while `isProcessAlive` proved the pid alive on the line immediately before. The MCP health monitor drives that path through `reconnect()` every 30s while a server is unreachable, so the line repeats indefinitely while blaming an exit that did not happen.

Separately, "same visibility" is not quite the relationship: vega-process runs inside the tool-server child, which this file spawns with its stderr pointed at `~/.argent/tool-server.log`, whereas this write lands on the host process's own stderr - the user's terminal for `argent run`/`tools`/`flow`/`update`/`uninstall`.

packages/argent-tools-client/src/launcher.ts
```diff
@@ -727,7 +739,14 @@ function processCommandMatches(pid: number, marker: string | undefined): boolean
     // so a bundle path containing spaces still matches its own ps output.
     const escaped = marker.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
     return new RegExp(`(?:^|\\s)${escaped} start(?:\\s|$)`).test(cmd);
```
hubgan: Now that the width limit is gone this match sees whole command lines, and `(?:^|\s)` is satisfied by any interior space - so the comment above it ("an unrelated process merely mentioning the path never matches") claims more than the pattern enforces. What it actually excludes is a mention that is not followed by ` start`. Checked against a live process: a `/bin/sh -c 'node <bundlePath> start >/dev/null 2>&1; sleep 30'` wrapper is reported as our tool-server, and would be truncated out of the match at a narrow width.

The two look-alike tests added in this PR pin the shapes that still fail - a mention with no trailing `start`, and a longer prefix whose path ends with ours - but not the mid-argv shape, which is the one the widened output newly exposes. Reaching it needs pid reuse landing on such a process, so this is about the comment and the coverage rather than about a likely kill.

packages/argent-tools-client/src/launcher.ts
```diff
@@ -714,7 +721,12 @@ export async function killToolServerForInstallDir(packageDir: string): Promise<n
...
   try {
-    const cmd = execFileSync("ps", ["-p", String(pid), "-o", "command="], {
+    // `-ww` disables ps's width truncation. Without it procps-ng clips the
+    // command to $COLUMNS, so a bundle path longer than that never matches
+    // its own marker, this returns false, and the kill-before-respawn below
+    // is skipped — orphaning the live server. Same flag tool-server's
+    // vega-process PS_ARGS uses.
+    const cmd = execFileSync(PS_BIN, ["-ww", "-p", String(pid), "-o", "command="], {
       encoding: "utf8",
       timeout: 2_000,
```
hubgan: No `maxBuffer` here, so the call inherits Node's 1 MiB default, while the vega-process calls this change is modelled on set 16 MiB explicitly. On Linux `ARG_MAX` is commonly 2 MiB, so a recycled pid sitting on a process with a very large argv comes back as `ENOBUFS`. The outcome is still fail-safe, but it lands in the catch below and is then reported as a pid that possibly just exited.

packages/argent-tools-client/test/launcher-sweep.test.ts
```diff
@@ -51,25 +51,45 @@ function writeRecord(bundlePath: string, pid: number): string {
...
 }
 
+/** Does this platform's `ps` clip `-o command=` to $COLUMNS? procps-ng does;
+ * BSD `ps` clips only to a terminal width, and the guard never hands `ps` a
+ * terminal — so the width regression is unobservable on macOS. Probed against a
+ * live pid through the guard's own binary and options rather than assumed from
+ * `process.platform`. */
+function psClipsToColumns(pid: number, columns: number): boolean {
+  const width = (widen: string[]): number =>
+    execFileSync(launcher.PS_BIN, [...widen, "-p", String(pid), "-o", "command="], {
```
hubgan: The probe exists so the suite can say "this platform cannot observe the bug" instead of passing vacuously, but `execFileSync` raises on a non-zero exit, on `ENOENT` and on timeout - so on a host whose `ps` rejects these flags the width test errors out rather than skipping, which is the opposite of the intent. Confirmed on `node:20-alpine`: `PS_BIN` resolves to busybox's `/bin/ps`, and `-ww -p 1 -o command=` exits 1 with `unrecognized option: w`. The guard being modelled treats that same failure as "unidentifiable" and returns false. The bite is limited, because on such a host the pre-existing sweep tests fail anyway.

Worth noting too that the probe hand-copies the guard's argv and stdio while importing only `PS_BIN`, so if the guard's invocation changes the probe silently stops representing it and the test either skips with no signal or mis-detects.

hubgan (commented):
Reviewed the diff and the code it transitively affects (all four `processCommandMatches` call sites, `terminatePid`, and the sibling launcher tests).

Verified on both platforms rather than by reading: macOS (BSD `ps`) and a Linux container (procps-ng 4.0.2/4.0.4). The suite is 11/11 on Linux with nothing skipped, 10 passed / 1 skipped on macOS, and stable over 5 consecutive Linux runs with no leaked keepalives. Six mutations - drop `-ww`, `-ww` to `-w`, `PS_BIN` to bare `"ps"`, drop the leading `(?:^|\s)` anchor, drop the trailing `` ` start` ``, drop the metacharacter escape - each red exactly one test and leave the other ten green, so the tests really do pin what they claim to. I also checked the Alpine case, since it is the one layout where preferring `/bin/ps` could have picked busybox over procps: `apk add procps` replaces `/bin/ps` itself, so the ordering is fine there.

Nothing blocking. The comments below are all minor.

packages/argent-tools-client/test/launcher-sweep.test.ts
```diff
@@ -117,6 +137,128 @@ describe("sweepDeadStateFiles", () => {
...
+    const marker = join(TEST_HOME, "gone-install/dist/tool-server.cjs");
+    const watcher = join(TEST_HOME, "watcher.cjs");
+    expect(existsSync(marker)).toBe(false);
+    // Carries the recorded path in its argv, but is not running it: the guard
+    // requires a following `start`, so this must survive untouched.
+    const child = await spawnFakeServer(watcher, { args: [marker] });
+    const file = writeRecord(marker, child.pid!);
+    try {
+      await launcher.sweepDeadStateFiles();
+      expect(await waitForExit(child, 1_000)).toBe(false);
```
hubgan: `sweepDeadStateFiles` unlinks the record before it terminates anything, so the `existsSync(file)` assertion on the next line already entails that no signal was sent - this wait cannot fail unless that assertion fails too. Between this test and its neighbour that is roughly 2.06s of pure sleeping in a file whose entire run is 2.28s on macOS and 2.23s on Linux.

hubgan (approved):

packages/argent-tools-client/src/launcher.ts
latekvo: Fixed in cb9c8c2f4c8b268337b565e1a1b39985bcf94035 - all three parts.

The read is now `readProcessCommandLine`, which pipes ps's stderr. Driven through it with a stub `ps` that exits 1 after writing `ps: invalid option -- 'w'`: the thrown error carries that line, while with `stdio[2]: "ignore"` restored the same case reads only `Error: Command failed: ps -ww -p 61591 -o command=`. Pinned as `carries ps's own complaint into the error it throws` in `test/launcher-ps-guard.test.ts`.

The parenthetical is gone, and so is the Windows hole behind it: the platform fallback moved into the guard itself (`couldBeOurToolServer` answers true where there is no `ps` to ask), so the three call sites share one rule instead of two of them carrying a `win32` local and the third carrying none. The new test spawns a wedged `unhealthy` auto-spawned server, fakes `process.platform` to `win32` with a `ps` on PATH that returns nothing, and asserts `ensureToolsServer` retires it - dropping the `win32` branch leaves the server alive (`expected true to be false`), which is the 30s-repeat loop you described.

The `same visibility` sentence is gone; the message says only what it can.

packages/argent-tools-client/src/launcher.ts
latekvo: Fixed in cb9c8c2f4c8b268337b565e1a1b39985bcf94035.

The comment states what the pattern enforces: the path at an argument boundary followed by `start`, so a mention that is not being run misses, while a command line embedding the pair mid-argv - a `sh -c` wrapper - matches. `signals a shell wrapper whose script embeds our launch line mid-argv` pins that shape against a live `/bin/sh -c 'echo ready; node <marker> start >/dev/null 2>&1; sleep 30'`.

Left the rule as it is rather than anchoring: requiring the path at argv[1] would stop identifying any launch that puts the bundle later, and an unidentified live server is orphaned - the direction every fallback in this guard leans away from. Confirmed the new test is the discriminator for that choice: mutating the pattern to `^\S+ ${escaped} start(?:\s|$)` fails it alone (1 failed | 10 passed | 1 skipped), while the blunter `^${escaped}` mutation fails five.

packages/argent-tools-client/src/launcher.ts
latekvo: Fixed in cb9c8c2f4c8b268337b565e1a1b39985bcf94035 - `maxBuffer: 16 * 1024 * 1024`, the same ceiling as the vega-process probes.

Reproduced through the guard's own reader with a stub `ps` printing a 2 MiB command line: unchanged it raises `Error: spawnSync ps ENOBUFS` from `readProcessCommandLine`, with the ceiling it returns all 2097176 characters (`reads a command line larger than Node's 1 MiB exec-output default`). It needs a stub because macOS cannot produce the input - `getconf ARG_MAX` is 1048576 and a 1,040,000-byte argv already fails to exec, so the ENOBUFS window only opens where argv may exceed 1 MiB.

packages/argent-tools-client/test/launcher-sweep.test.ts
latekvo: Fixed in cb9c8c2f4c8b268337b565e1a1b39985bcf94035.

`psClipsToColumns` calls `launcher.readProcessCommandLine(pid)` for the widened arm - the guard's own binary, flags, timeout, `maxBuffer` and stdio, so a change to the invocation reaches the probe - and `(pid, [])` for the truncating control, with both inside one try/catch that returns false. A `ps` that rejects the flags therefore skips the test rather than erroring it, matching the guard's own `unidentifiable` verdict. `PS_BIN` is no longer exported; the reader is what the test imports.

packages/argent-tools-client/test/launcher-sweep.test.ts
latekvo: Fixed in cb9c8c2f4c8b268337b565e1a1b39985bcf94035 - both waits removed (this one and its neighbour in `does not signal a different install whose bundle path ends with ours`).

The surviving `existsSync(file)` assertion carries the whole claim, since a signal is only ever sent after the record is unlinked; the two tests also lose their 20s timeouts, which were sized for the wait. Measured on macOS: `vitest run test/launcher-sweep.test.ts` goes from 2.31s (tests 2.24s) at 08004bf51 to 0.58s (tests 235ms), with one more test in the file.

packages/argent-tools-client/src/launcher.ts
latekvo: Follow-up in 9b811aa67398b1b43bdb1d6b9356fe4ff85a889c: the wrapper fixture as first pushed would have been green here and red on the Linux runner. dash - ubuntu's `/bin/sh` - execs the last command of a `-c` script in place, so the process `ps` reports becomes `sleep 30` with none of the wrapper's argv; macOS `/bin/sh` (bash) keeps the shell, which is why it passed locally. Measured with `/bin/dash -c 'echo ready; sleep 5'` -> `ps -ww -p <pid> -o command=` prints `sleep 5`, and with a trailing `:` prints the full `-c` line. The script now ends in a no-op, emits `ready` after the failing `node` so the shape is settled before the sweep looks, and asserts the command line contains `<marker> start` before sweeping. Ran the test against `/bin/dash` directly: 3/3 pass with the no-op, 3/3 fail on that precondition without it.

packages/argent-tools-client/test/launcher-ps-guard.test.ts
```diff
@@ -0,0 +1,158 @@
...
+  stubDir = mkdtempSync(join(tmpdir(), "argent-ps-stub-"));
+  writeFileSync(join(stubDir, "ps"), PS_STUB, "utf8");
+  chmodSync(join(stubDir, "ps"), 0o755);
+  bundlePath = join(stubDir, "tool-server.cjs");
+  copyFileSync(FIXTURE_BUNDLE, bundlePath);
+  // os.homedir() — which STATE_DIR is built from — reads USERPROFILE on Windows
+  // and HOME elsewhere, so pin both or the redirect is inert there and these
+  // tests operate on the real ~/.argent.
+  process.env.HOME = stubDir;
+  process.env.USERPROFILE = stubDir;
```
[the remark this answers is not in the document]
latekvo: Fixed in 19542e161465a906eef835d523baa19a7487c2ee.

Confirmed against the file first: `beforeAll` set `HOME`, `USERPROFILE` and `PATH`, and `afterAll` restored only `PATH` before `rmSync(stubDir)`. Reproduced rather than assumed - a probe file asserting `existsSync(homedir())` placed after this suite:

```
npx vitest run test/launcher-ps-guard.test.ts <probe> --no-isolate --no-file-parallelism
```

read `homedir()` as the removed `…/argent-ps-stub-YHKyUj`, `exists=false`. With the restore it passes, and deleting the restore loop again makes it fail - so the fix is pinned by the repro in both directions rather than asserted. Under the shipped config the leak is unobservable (`isolate` defaults to true, so each file gets its own fork); `--no-isolate` is what exposes it.

All three variables are now captured in `beforeAll` and restored-or-deleted in `afterAll` ahead of the `rmSync`, matching `test/artifacts.test.ts`. I audited the rest of what the suite mutates: `ARGENT_PS_STUB` is deleted in `afterEach`, `FAKE_MODE` in a `finally`, `process.platform` is put back from its saved descriptor in a `finally`, spawned pids are SIGKILLed in `afterEach`, and the suite never changes cwd.

Same class in the neighbours: `test/launcher-sweep.test.ts` has it too and is fixed in the same commit (it is the other file this PR touches). Six more - `launcher-state`, `launcher-spawn`, `launcher-version-gate`, `launcher-duplicate-spawn`, `launcher-exit`, `link-config` - carry the identical asymmetry and each fails the same probe; left out of this PR since they are untouched by it and filed as software-mansion/argent#1028 with the repro, including the suggestion to lift the four lines into `test/setup/` rather than keep eight copies.

PR #585, opened 2026-07-29
feat(flow): boot a Chromium instance per launch step

j-piasecki:
A `launch:` on Chromium now starts its own Electron instance, so nested e2e flows and mid-flow relaunches work as they do on iOS/Android. Previously the runner booted one app per run and rejected any second chromium launch.

A chromium device id _is_ its CDP port, so a launch that starts an app from scratch has to boot one and move the run onto it: `state.device` became reassignable (steps already read it per call), and `state.owned` tracks what the runner booted, torn down in reverse at run end. An instance the runner did not boot — pinned via `--device`, or auto-detected — is attached to, never killed. Relaunching the same app kills its instance and waits for the process to actually exit, or the replacement loses a race against the app's single-instance lock.

The runner needs a device before step 1, so `resolveRunDevice` follows a leading `run:` chain to find the launch the run _begins_ with: a fragment composing a chromium e2e flow now boots instead of failing with "No booted chromium device found". A launch declaring no app id for the run's platform stays an error — the run's platform is fixed start to finish.

Also fixes a pre-existing teardown leak: the spawned handle is the npm `electron` node wrapper, not Electron, so signalling it left the app and its helpers running (reparented to launchd, still in the dock). Teardown now sweeps the child's process group.

**Behavior change:** a fragment leading with `run:` into a chromium e2e flow now boots a fresh instance where it previously attached to a running one; pass `--device` to keep attaching.

## Tests

Cover the boot/attach/settle three-way, teardown ordering and the group sweep, device rebinding after a nested launch, same-app retire ordering, cancellation mid-boot, both "no chromium app declared" guards, and the leading-`run:` chain (multi-hop, cyclic, no-launch). The guard, exit-wait and sweep tests are mutation-checked.

PR #714, opened 2026-08-03
test: stop tests binding machine-wide resources they do not own

latekvo:
## Why

Three flaky or failing tests were diagnosed separately this week, and all three turned out to be one defect wearing different clothes: **a test binding a machine-wide resource it does not own**, so a second concurrent run — another checkout, a CI matrix cell, a developer's own app — perturbs it. Those three are fixed in #707, #709 and #711.

Each of those was found by chasing a symptom, and each audit covered only its own resource type. This is the systematic sweep: all **448 test files across 13 packages**, plus the 4 `node --test` scripts. (The sweep itself ran over the 400 files that existed when it started; merging `main` brought the count to 448, and every category below was re-run over the full set — 25 files assign `env.HOME` and none omit `USERPROFILE`, one test file calls `pkill`/`pgrep` (`vega-cli-timeout.test.ts`, scoped out to #711 below), and no `listen()` names a fixed port.)

## Findings

| Category | Files checked | Real findings |
|---|---|---|
| Network endpoints (bind / probe / unix socket) | 448 | 4 |
| Filesystem — shared temp namespace (`os.tmpdir()` and literal `/tmp`) | 448 | 8 |
| Filesystem — real `$HOME` / `~/.argent` | 448 | 4 (28 files) |
| Ambient environment | 448 + all `src` readers | 12 |
| Device / emulator / SDK state | 448 | 2 |
| Process table (`pgrep` / `pkill` / fixed-pattern sweeps) | 448 | **0** |
| In-process globals leaking across files | 13 packages × 13 modes | **0** |

Every finding was reproduced with a concrete instrument before being touched, and the instrument re-run after. Representative:

- **`stop-tools.test.ts`** ran the real `stop-metro` against port 59999, which `lsof`s and **SIGTERMs whatever it finds**. With a squatter there: `16 passed` → `1 failed`, and the squatter logged `received SIGTERM #1`. Now port 1.
- **`bind-failure-telemetry` / `lens-relay-telemetry`** were gated on the real `~/.argent/flags.json`. Enabling the documented `tool-server-event-log` flag: `4 passed` → **`4 failed`**, and records were appended to the developer's real event log.
- **LogFileWriter (12 files)** — its constructor `mkdir -p`s `os.homedir()/.argent/tmp`. Instrument: run each of the 16 test files that reference a writer, or a blueprint that builds one, under a scratch `HOME`/`USERPROFILE`, then test `$HOME/.argent/tmp` for existence. 4 created it; the 12 that actually construct a writer now all call `scopeTempHome`, and none creates it.
- **`screen-recording.test.ts`** — the device ids are fixtures and `vi.useFakeTimers()` seeds its clock from the real time, so only the millisecond a run starts in separates two runs' `.mp4` paths, and each test deletes the path it derived. Self-concurrent: **2 of 3 rounds failed** → 0 of 4. Scoping `TMPDIR` removes the window rather than narrowing it.
- **`DO_NOT_TRACK=1`** → telemetry `282 passed` → **`13 failed`**. **`npm_config_user_agent=pnpm|yarn|bun`** → installer `535 passed` → 4/5/3 failed. **19 tests required a real `adb`** — PATH shadow + empty HOME: `30 passed` → **`19 failed`**.
- A fixed `/tmp/argent-nd-<udid8>.sock` **destroyed a concurrent holder's live socket** (proven by inode) in **3** files that run the native-devtools factory. Two of them also fail outright: `native-devtools-app-state` lost **3 of 3** rounds of two concurrent runs on `expected 'unregistered' to be 'connected'`. All three now derive the UDID from `process.pid`.
- `flow-visual`, `react-profiler/dump`, `file-inputs`, `http-upload` — shared `/tmp` paths and prefix scans, each reproduced with a squatter. In `http-upload` the collision surfaced *as a timeout*, inside `vi.waitFor`. (`flow-chromium-boot` was fixed on this branch too, but main's #585 rewrote that file afterwards and the merge kept main's copy, so it is byte-identical to `origin/main` here and carries no fix.)

## Clean, with the evidence

- **Process table** — an `LD_PRELOAD` exec interposer over a full run (507 execs) plus all 12 other suites. Excluding `vega-cli-timeout.test.ts`, which is #711's and does sweep the table (`pkill -f '^sleep 6910[0-9]$'` at line 134): no `pgrep`/`pkill`/`killall`/fixed-pattern sweep anywhere, and **no test signals a pid it did not spawn**.
- **Real device state** — **zero** exec attempts of real `adb` / `simctl` / `emulator` / `avdmanager` / `vega`; no test asserts a device list. That list was too narrow: `workspace-reader-integration` spawned the real `yarn` / `pnpm` / `bun` / `pod` / `eas` / `expo`, which is the second finding in this row.
- **Cross-file in-process leakage** — structurally impossible: vitest 4.1.9 resolves `pool: "forks"` / `isolate: true`, confirmed by a scratch experiment (distinct pids, no env leak, module `Map` = 1,1,1). 120 default-mode runs across 12 packages produced 1 failure.
- **`scripts/`** — 4 files, 86 tests: `mkdtemp` throughout, no ports, no sweeps, writes nothing.

## Verification

Each fix was mutated back and killed by a **named** test under its instrument — e.g. `× reads a plain absolute path`, `× returns null when only the extensionless binary exists`, `× pins both names os.homedir() consults, at the same directory`. Restores verified with `git diff` / `git status`, never a substring grep.

**Pinned by an instrument only:** the socket fix for `ios-only-blueprint-gate`, the `~/.argent/tmp` and `~/.argent/flags.json` fixes, the `/tmp` fixture cleanup, the `stop-metro` port change (the test asserts `stopped: false` / `pids: []`, which hold for either port whenever nothing is listening) and the `workspace-reader` `PATH` pin (only `node` is asserted; the other seven probes are not) are not pinned by any suite assertion — reverting them still passes, and only the external squatter / scratch-`HOME` instruments catch them. They are pollution fixes. The `native-devtools-app-state` socket fix is the exception in the other direction: reverting it turns two concurrent runs red.

Gates, unpiped and read by exit code: `eslint .` 0, `prettier --check .` 0, `tsc --build` 0, `typecheck:tests` 0.
Suite, run as six shards: tool-server **4287 passed, 1 skipped, 0 failed**; argent-cli 462; telemetry 306; argent-tools-client 191; configuration-core 122; argent-mcp 80; argent-installer 555. All other packages and `test:scripts` green, identical to baseline.

`lens-tools-platform-gate.test.ts` times out at the 5s default under full-suite load and passes at `--testTimeout=30000` (24.7s, almost all of it transform). It predates this branch and is untouched by it.

## Scope

This branch is **test-only** — no `src/` file is touched. Both production defects the sweep turned up are separated out, so each can be reviewed as the user-facing fix it is: `ps` without `-ww` orphaning live tool-servers is #713, and #852 is `ensureToolsServer` respawning on every call when `ARGENT_HOST` is exported — `buildToolsServerEnv` passes the inherited value to the child while the state records the literal `127.0.0.1` the health check then probes. The second surfaced here as eight stranded `fake-tool-server` processes, and pinning the environment in this branch removes that signal, which is why it is filed rather than merely worked around. A third, outside this branch entirely, is #853: ten `tool-server` test servers omit the host on `listen(0)`, binding every interface where twenty-two sibling sites pass `127.0.0.1`. It is hardening, not a live exposure — `createHttpApp`'s Host allow-list answers a LAN-addressed request with 403, and only a raw client forging `Host: 127.0.0.1` reaches the stub registry behind it, inside the ~100 ms one `it()` holds the socket. It is also not a collision risk: ephemeral ports are kernel-assigned and never reissued while in use, which is why the fixed-port audit above is unaffected. All ten predate this branch.

Left untouched on purpose: `boot-electron-spawn-error.test.ts`, `chromium-discovery.test.ts` and `vega-cli-timeout.test.ts` — owned by the still-open #709 and #711, and all three still carrying their defects on this base. (#707 has since merged into this base, so `boot-device-hotboot.test.ts` and `vitest.config.ts` carry its fix rather than a defect; `test/setup/clear-argent-env.ts` is its work, and this branch relies on it.)

**Not fixed, deliberately** — an adjacent class (test-*order* dependence, not machine-global resources), reproducible if wanted: `link-config.test.ts` (deterministic: `-t "returns null when required field"` → 4 failed), `network/network-integration.test.ts` and `android-perfetto/dispatch.test.ts` (fail under `--sequence.shuffle` seeds).

**Also observed, not addressed:** one green tool-server run leaves **56** entries behind in a scratch `TMPDIR` — **30** in the `argent-screenshot-diff*` family, 4 `argent-events-*`, 3 `argent-flow-crop-*`, 3 `argent-tar-upload-*`, 2 `argent-flow-diff-*`, 2 `argent-file-input-*`, and 12 assorted singletons (`ast-index-*`, `component-source-*`, `argent-profiler-cwd`, `argent-chromium-media`, …). Measured per run against a scoped temp dir, so it ranks the contributors rather than counting one machine's accumulated history — screenshot-diff, not flow-crop, is the dominant one. A separate leak. That instrument is blind to a *literal* path, which is how it missed the two `/tmp/native-profiler-*-report.md` writers the fourth pass found. Re-run across all 13 packages, `argent-mcp` is the only one outside `tool-server` that leaves anything — two entries per run, filed as #854.

## Second pass

A review of this branch re-ran every category above with its own instrument rather than reading the diff, and found the sweep had pinned four resources in one file while leaving a sibling that does the same thing live. Fixed in `4289e3d8f` and `8164955b4`; the table and bullets above are the corrected totals.

- **`tool-server-event-log`** was mocked off in 2 of the 4 files that call the real `start()`. With the flag on, `startup-telemetry` and `tool-fail-invalid-params-emission` failed 3 tests in `attachRegistryEventLogger` and truncated the real `~/.argent/tool-server-events.jsonl` to 0 bytes first. `ARGENT_EVENT_LOG` cannot redirect that: `test/setup/clear-argent-env.ts` strips every `ARGENT_*` var before the module graph loads.
- **`scopeTempHome`** reached 8 files; 4 more built a real `LogFileWriter` through the JS-runtime-debugger blueprint.
- **`os.homedir()` on Windows** reads `USERPROFILE`, not `HOME`. `android-binary-windows` and `adb-resolve-avd-path` pinned only `HOME`, and both run on a real windows-latest host. Proven by running the resolver under a `node:os` mock that reproduces the Windows lookup: with `HOME` alone the stock Studio SDK under the profile resolves; with both pinned it misses. `update-checker` had the same gap.
- **`getConsentState`** honours a falsy `ARGENT_TELEMETRY` at the same precedence as `DO_NOT_TRACK`, ahead of the config override the suite writes. Only `DO_NOT_TRACK` was pinned: `ARGENT_TELEMETRY=0` gave `17 failed | 6 passed`.
- **`workspace-reader-integration`** let the version probe spawn the real package managers off `PATH`; their shims self-install, so the file pulled **~44 MB** into `~/.cache/node/corepack` and `~/Library/Caches/eas-cli`. `PATH` is now a directory holding only `node`, the one version the test asserts — 44 MB → 4 KB.
- **`argent-mcp`** snapshotted `ARGENT_AUTO_SCREENSHOT_DELAY_MS` without clearing it, and that package has no `ARGENT_*`-stripping setup file, so 3 tests asserted the developer's exported value.

Smaller items in the same commits: the `/tmp` socket fixtures now remove the directories they create per run; `http-upload`'s pre-scan snapshot is gone, since scoping `TMPDIR` to a fresh directory made it permanently empty and its filter inert, as its two siblings already did; and two `detectProjectPackageManager` assertions that enumerated the entire return union are pinned to a single value.

The confirmation pass also caught a defect the first round of fixes introduced: pinning `HOME`/`USERPROFILE` to the fixture root in `android-binary-windows` let `<home>/Android/Sdk` shadow the `%LOCALAPPDATA%` root the two positive tests are named after, so deleting `roots.push(join(localAppData, …))` from the resolver left all three green. Home has its own directory in `6aee3ee7a`, and that mutation now fails both tests.

A third pass, running the nearest-twin move repo-wide, found the `os.homedir()`-on-Windows gap in one more package: eight `argent-tools-client` test files redirect only `HOME`, and four of them recursively delete `STATE_DIR` — two before each test, two after — which on Windows is the developer's real `%USERPROFILE%\.argent`. Fixed in `139447d4a`, together with the same gap in `avd-snapshot`. Nothing runs that package on Windows CI today, so no current run fires it.

## Fourth pass

A fourth pass audited the three fix commits and re-ran the sweep with two instruments the earlier ones did not have. Both holes were in the *measurement*, not the diff — which is why three passes missed what they found. Fixed in `a67c43f8d`.

- **`os.tmpdir()` does not read the same variable on every platform.** POSIX takes `TMPDIR`, `TMP`, `TEMP` in that order; Windows takes `TEMP`, then `TMP`, and never looks at `TMPDIR`. The five pins this branch added set `TMPDIR` alone, so on Windows production still materialized into the machine-wide `%TEMP%` while the test scanned a scratch directory nothing ever wrote to — and `file-inputs`, `http-upload` and `flow-visual` each end in a leak assertion over that listing, which then passes whatever the code does. Against `origin/main`, which diffed a before/after listing of the real tmpdir, that is a regression rather than a wash. `redirectTmpdir` (`test/helpers/tmpdir-env.ts`) sets all three and returns the restore; the **7** files that scope a tmpdir use it, including the two `android-perfetto` ones that predate this branch. `test/tmpdir-env.test.ts` pins both lookup chains — reverting the helper to `TMPDIR` alone fails two of its three tests. All three leak assertions were re-confirmed live on POSIX by mutating the production cleanup each one covers.
- **A hardcoded absolute path is invisible to a `TMPDIR`-scoped instrument.** `session-correctness` and `ios-instruments/analyze-freshness` passed a literal `/tmp/native-profiler-{x,ios}.trace` as the session's trace file, and the renderer writes its report beside it (`deriveReportPath` = `dirname(traceFile)/<base>-report.md`). Every run left `native-profiler-x-report.md` and `native-profiler-ios-report.md` in the shared `/tmp`; nothing removed them, and two concurrent runs write the same names. Reproduced with `TMPDIR` pointed at a scratch dir — the files still landed in `/tmp`, which is exactly why the `os.tmpdir()` row and the leftover census below both read clean here. Both tests now own the directory.
- **A prefix-scanned env var is invisible to a grep for variable names.** `secretSources` enumerates `process.env` by the `ARGENT_SECRET_` prefix and ranks it *above* every file the test writes, so `secrets-command.test.ts` — which already pins `HOME`, `USERPROFILE` and cwd — still read the developer's shell. `ARGENT_SECRET_FOO=bar`, argent's own documented way to supply a secret: `6 passed` → **`3 failed | 3 passed`**, deterministic. `argent-cli` has no `ARGENT_*`-stripping setup file, so the prefix is cleared per test. Its sibling `configuration-core/tests/secrets.test.ts` was already immune — it threads an explicit `env: NO_ENV` into every call.
- **`android-binary` and `vega-cli-fallback`** were the last two files redirecting `HOME` without `USERPROFILE` — the former the direct twin of `android-binary-windows`, in the same directory, driving the same resolver. Under a `node:os` mock reproducing the Windows lookup with an SDK planted in the profile, it loses **5 of 11** tests. No test file in the repo redirects `HOME` alone any more. (`6fbb66ce1` later skipped both of this file's describes on win32 — its fixtures are extensionless and the win32 resolver only takes `.exe` — so its own `USERPROFILE` pin is now symmetry with the sibling files that do run there, not a live guard.)

Two smaller items in the same commit. `native-devtools-socket-bind`'s ENOENT fixture still rooted at `os.tmpdir()` while the comment three screens above explains why the others do not: with a 142-character ambient `$TMPDIR` that path exceeds `sun_path` and `listen` rejects it for *that* reason, which the assertions cannot tell from the missing-directory one they were written for. And `<home>\AppData\Local\Android\Sdk` — the resolver's fallback for a GUI-spawned server that never inherited `%LOCALAPPDATA%` — could be deleted with the whole suite green; it now has a test, and deleting either win32 root fails exactly the tests named for it.

## Fifth pass

A fifth pass took the platform-branch shape the fourth pass exposed and ran it as its own move: *for every function that branches on `process.platform`, does the test's guard cover every arm, or only the one the author's machine takes?* Fixed in `e08b466b6`.

- **`init-stale-config.test.ts` deletes the developer's real VS Code MCP config on Windows.** The file mocks `os.homedir()` and says it does so "so hidden-scope probes (`~/.claude.json`, VS Code user profile) never touch the real home directory". `vscodeUserDirs()` branches `darwin → homedir()/Library/Application Support`, `win32 → %APPDATA%`, `else → homedir()/.config` — the win32 arm never calls `homedir()`, so the mock is inert there, and the test's own `userMcpJsonPath()` mirrors the same branch. Both user-profile tests then write a one-entry fixture over `%APPDATA%\Code\User\mcp.json` with a whole-file `writeFileSync`, and assert that `remove()` deleted it. Reproduced by forcing `process.platform` to `win32` with a two-server `mcp.json` planted at a probe `%APPDATA%`: **`3 passed`, and the file and its directory were gone.** With `APPDATA` pinned, the whole file passes under the same instrument — 31 tests — and the planted config is byte-identical.
- **`argent-tools-client` hands its ambient `ARGENT_*` to a real spawned child.** The launcher spawns an actual tool-server, and `buildToolsServerEnv` overrides only the keys it is given — everything else rides in on `...baseEnv`. `ARGENT_HOST` is the one that bites: the fixture binds the host it names while the launcher records and health-checks `127.0.0.1`, so reuse misses and a second server is spawned. With six overrides exported: **`10 failed | 181 passed`, and eight stray `fake-tool-server` processes still holding TCP listen sockets after the run.** With the prefix stripped: `191 passed`, none surviving. The package now carries the same setup file `tool-server` has, for the same reason — and it is the second of the two packages whose missing `ARGENT_*` strip this review found.

## Later passes

A sixth pass audited the fix commits themselves and bounded the SDK probe (`6fbb66ce1`); a seventh ran the moves against the diff one more time and turned up the two below, plus the `listen(0)` defect filed as #853.

- **The `PATH` pin made `workspace-reader-integration` POSIX-only without the guard its siblings got.** The shim it puts on `PATH` is an extensionless `node`, and `symlink()` to a file needs an elevated process or Developer Mode on Windows, so `beforeAll` throws `EPERM` there before any assertion runs; a copy in place of the symlink would not rescue it either, since a bare command name resolves through `PATHEXT`. This sweep already skips `android-binary` and `native-devtools-socket-bind` on win32 for exactly this class, and the Windows job runs a curated list of seven `tool-server` files that excludes this one, so nothing was catching it. Fixed in `095d75bd8` — verified POSIX `2 passed`, and under a forced win32 platform `2 skipped` with the top-level `beforeAll` never entered.
- **A rationale that rested on a case the file excludes.** `android-binary`'s header justified pinning `USERPROFILE` by what `os.homedir()` reads on Windows, two sentences before noting that both describes skip on win32. Fixed in `f37b3605e`; the variable stays in the snapshot list, the justification no longer claims a platform this file never reaches.

A seventh pass also re-ran the sweep against a hostile machine: every feature flag on in a scratch `~/.argent/flags.json`, no `adb` on `PATH`, and a `TMPDIR` long enough to overrun the 104-byte `sun_path` limit. It found one resource this branch had not closed.

- **A red `launcher-version-gate` run strands a real tool-server.** `ensureToolsServer` spawns a fixture server, but its pid only reached the reaper after three or four assertions on the state it returned; any of those throwing left the pid unrecorded, and the same `afterEach` then deleted `STATE_DIR`, the only other record of what to kill. **Eight such processes were alive on the development machine**, spread from 14:33 to 18:23, every one from an `argent-version-gate-test-*` fixture and none from any other family — each still holding a TCP listen socket with its fixture directory long gone. Forcing one assertion to fail took the `tool-server.cjs start` count from 10 to 11. The nearest twin, `launcher-duplicate-spawn`, guards this twice: it records the pid straight after the null check and sets `FAKE_TTL_MS` so an escaped server self-exits. Fixed in `26c2c60ec` with both, the TTL also covering the window between the spawn and the state read where no pid exists yet — the same forced failure now leaves **zero** strays.

An eighth pass audited the seventh's own fixes, and found the version-gate fix had been applied to one file when the defect was a property of four.

- **Three more files strand a real tool-server on a red run.** `launcher-spawn` and `launcher-exit` set `FAKE_TTL_MS` nowhere; `launcher-duplicate-spawn` — the file `26c2c60ec` cites as the model — armed it inside one test's `try/finally`, so eight of its nine spawns ran without it and the `finally` disarmed it for the remainder of that test. Reproduced per file by forcing a failure before the pid is recorded: a detached node process survives, reparented to init, holding a listening socket, with its `TEST_HOME` removed and its state record already cleared. Fixed in `3bb69ded6` by moving the net to a file-wide `beforeEach` in all three. Verified end to end — the child carries `FAKE_TTL_MS=60000`, and a stranded server spawned at 21:28:36 exited by itself at 21:29:36 with nothing signalling it; three consecutive green runs stay at 191 passed.
- **A 29th file writes the real `~/.argent`.** `boot-electron-spawn-error`'s success-path test completes a boot, and a completed boot calls `trackChromiumPort`, persisting to `os.homedir()/.argent/chromium-cdp-ports.json`. The real file on the development machine had accumulated **106 dead ephemeral ports**. Its two siblings in the same directory already mock `chromium-discovery` for exactly this reason; this was the third, with no mock. Fixed in `3306b6f40`.
- **`scopeTempHome` was itself unpinned.** Deleting its `USERPROFILE` line — the Windows half, the same property the sibling helper's test pins — left all twelve consumers green at 48 passed. `3306b6f40` adds the missing test; it kills that mutation plus dropping the `HOME` pin and replacing the per-test `mkdtemp` with one shared directory.

Two findings from that pass were refuted on verification and are recorded rather than acted on: the tool-server tmpdir leftovers (mechanism real, but they are the separate leak this body already scopes out) and a claim that `26c2c60ec`'s message overstated its twin.

Four comments were also corrected against their code: `deriveReportPath` joins `<base>-report.md` *inside* the trace's directory rather than appending to the directory's name, which is what makes the `rm(traceDir)` added in `a67c43f8d` cover it; two `android-perfetto` comments still named `TMPDIR` as the mechanism after the switch to `redirectTmpdir`; and `android-binary`'s header claimed the home pin lets its two positive tests resolve "on either platform", though its fixtures are extensionless and the win32 resolver only accepts `.exe`.


packages/tool-server/test/bind-failure-telemetry.test.ts
```diff
@@ -84,6 +84,14 @@ vi.mock("@argent/registry", async (importOriginal) => {
   const actual = await importOriginal<typeof import("@argent/registry")>();
   return { ...actual, attachRegistryLogger: vi.fn() };
 });
+// start() gates its event log on the `tool-server-event-log` flag, which the
+// real reader resolves from the developer's own ~/.argent/flags.json. Left
+// live, a developer who enables that documented flag fails these tests and
+// gets unit-test records appended to their real event log. Pin it off.
+vi.mock("@argent/configuration-core", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("@argent/configuration-core")>();
+  return { ...actual, isFlagEnabled: vi.fn(() => false) };
```
latekvo, line 95: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The tool-server-event-log flag is pinned off in only 2 of the 4 tests that run the real start().

Exactly four test files import and call the real `start()` from `../src/index` (grep 'src/index' over packages/tool-server/test): bind-failure-telemetry.test.ts, lens-relay-telemetry.test.ts, startup-telemetry.test.ts, tool-fail-invalid-params-emission.test.ts. The PR adds the `vi.mock("@argent/configuration-core", ... isFlagEnabled: vi.fn(() => false))` guard to the first two only. The other two carry the identical shape — the same `registryMock = vi.hoisted(() => ({ dispose: vi.fn()... }))` with no `.events`, the same `vi.mock("../src/utils/setup-registry", () => ({ createRegistry: vi.fn(() => registryMock) }))` — and neither mocks @argent/configuration-core nor scopes HOME (grep for HOME/homedir/configuration-core in both returns nothing). So with the flag on, src/index.ts:185 takes the true branch and both files hit the real reader, exactly the defect this hunk fixes.

**Trigger.** `argent enable tool-server-event-log` (or a `<repo>/.argent/flags.json` with it set — isFlagEnabled reads project scope first, flags.ts:221), then run the tool-server suite. startup-telemetry.test.ts and tool-fail-invalid-params-emission.test.ts throw `Cannot read properties of undefined (reading 'on')` out of start(), and the developer's ~/.argent/tool-server-events.jsonl is truncated to empty first.

**Checked against.** Fixed file, bind-failure-telemetry.test.ts:91-94: `vi.mock("@argent/configuration-core", async (importOriginal) => { const actual = await importOriginal<...>(); return { ...actual, isFlagEnabled: vi.fn(() => false) }; });`. Unfixed sibling startup-telemetry.test.ts:15-17 `const registryMock = vi.hoisted(() => ({ dispose: vi.fn().mockResolvedValue(undefined) }));` and :83 `const { start } = await import("../src/index");` — `grep -n 'configuration-core' packages/tool-server/test/startup-telemetry.test.ts packages/tool-server/test/tool-fail-invalid-params-emission.test.ts` returns nothing.

packages/tool-server/test/helpers/temp-home.ts
```diff
@@ -0,0 +1,38 @@
...
+import { afterEach, beforeEach } from "vitest";
+
+/**
+ * Point HOME — and USERPROFILE, which `os.homedir()` reads on Windows — at a
+ * fresh temp directory around every test in the calling file, then remove it.
+ *
+ * Whatever reaches `os.homedir()` then resolves inside a directory this run
+ * owns. LogFileWriter is the case that matters here: its constructor
+ * mkdir -p's `os.homedir()/.argent/tmp`, so a suite that builds one — directly,
+ * or through the JS-runtime-debugger blueprints — otherwise creates that
```
latekvo, line 13: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The LogFileWriter home-scoping sweep misses four test files that build a real writer, so a full suite run still creates ~/.argent/tmp in the developer's real home.

The new helper is applied to 8 files, and its docblock states the rule it enforces: "LogFileWriter ... its constructor mkdir -p's `os.homedir()/.argent/tmp`, so a suite that builds one — directly, or through the JS-runtime-debugger blueprints — otherwise creates that directory in the developer's real home." Four more test files build one through exactly that blueprint and were left unpinned:

- packages/tool-server/test/metro/teardown-log-history.test.ts — its own header says it "Drives the real Registry -> JsRuntimeDebugger -> debugger-log-registry path against a mock Metro", and line 102 does `const api = await registry.resolveService<JsRuntimeDebuggerApi>(urn)` then `api.logWriter.write({...})`. This is a direct sibling of metro/integration.test.ts, metro/multi-device.test.ts and metro/vega-target-routing.test.ts, which all got `scopeTempHome`.

**Trigger.** `rm -rf ~/.argent/tmp && cd packages/tool-server && npx vitest run test/metro/teardown-log-history.test.ts` — the blueprint factory runs `new LogFileWriter(port)`, whose constructor does `fs.mkdirSync(path.join(os.homedir(), ".argent", "tmp"), {recursive:true})` against the real HOME, so ~/.argent/tmp is recreated (and argent-logs-<port>-<ts>.log written into it) exactly as before the PR. Same for the three debugger/ files.

**Checked against.** temp-home.ts:11-14 — "LogFileWriter is the case that matters here: its constructor mkdir -p's `os.homedir()/.argent/tmp`, so a suite that builds one — directly, or through the JS-runtime-debugger blueprints — otherwise creates that directory in the developer's real home."
src/blueprints/js-runtime-debugger.ts:245 — `const logWriter = new LogFileWriter(port);`
src/utils/debugger/log-file-writer.ts:77-78 — `const dir = path.join(os.homedir(), ".argent", "tmp"); fs.mkdirSync(dir, { recursive: true });`
test/metro/teardown-log-history.test.ts:101-110 — `const api = await registry.resolveService<JsRuntimeDebuggerApi>(urn); ...

packages/tool-server/test/android-binary-windows.test.ts
```diff
@@ -40,6 +40,11 @@ describe("resolveAndroidBinary on Windows", () => {
     delete process.env.ANDROID_HOME;
     delete process.env.ANDROID_SDK_ROOT;
     tmpRoot = await mkdtemp(join(tmpdir(), "argent-android-win-"));
+    // The win32 resolver also derives a root from the home directory
+    // (%USERPROFILE%\AppData\Local\Android\Sdk). Point HOME at the same empty
+    // temp root so a real Studio install under the developer's home cannot
+    // satisfy a lookup these tests expect to fail.
+    process.env.HOME = tmpRoot;
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Windows-resolver test pins HOME but not USERPROFILE, so the added guard is inert on the one platform it emulates — and the comment promises otherwise.

The added comment claims: "The win32 resolver also derives a root from the home directory (%USERPROFILE%\\AppData\\Local\\Android\\Sdk). Point HOME at the same empty temp root so a real Studio install under the developer's home cannot satisfy a lookup these tests expect to fail." The resolver reads `homedir()` (android-binary.ts:3 `import { homedir } from "node:os"`, line 126 `const home = homedir()`), and `os.homedir()` reads USERPROFILE on Windows, not HOME. Setting only HOME therefore does nothing on Windows — and this file is deliberately run on a real windows-latest host (.github/workflows/windows-e2e.yml lists `test/android-binary-windows.test.ts` under "Resolver unit tests on Windows (tool-server)"). The PR's own new helper gets this right in the same diff: temp-home.ts:7 "Point HOME — and USERPROFILE, which `os.homedir()` reads on Windows", and sets both. The file's ENV_KEYS array already contains HOME only, so USERPROFILE is neither snapshotted nor pinned.

**Trigger.** A Windows contributor (or any Windows runner image) with Android Studio's default SDK at %USERPROFILE%\\AppData\\Local\\Android\\Sdk runs `npx vitest run test/android-binary-windows.test.ts`. beforeEach deletes ANDROID_HOME/ANDROID_SDK_ROOT and redirects LOCALAPPDATA to tmpRoot, but `defaultAndroidRoots()` still pushes `join(homedir(), "AppData","Local","Android","Sdk")` = the real profile path, so the test `returns null when only the extensionless binary exists (Windows can't exec it)` resolves a real adb.exe and fails on `expect(await resolveAndroidBinary("adb")).toBeNull()`.

**Checked against.** test/android-binary-windows.test.ts:43-47 — comment + `process.env.HOME = tmpRoot;`; :13 `const ENV_KEYS = ["PATH", "ANDROID_HOME", "ANDROID_SDK_ROOT", "HOME", "LOCALAPPDATA"] as const;` (no USERPROFILE)
src/utils/android-binary.ts:126 `const home = homedir();` and :145-146 `if (process.platform === "win32") { ... roots.push(join(home, "AppData", "Local", "Android", "Sdk")); }`
test/helpers/temp-home.ts:7 and :26-28 — `process.env.HOME = active; process.env.USERPROFILE = active;`
.github/workflows/windows-e2e.yml:81-88 — `runs-on: windows-latest` ... `npx vitest run test/command-on-path.test.ts test/android-binary-windows.test.ts test/adb-resolve-avd-path.test.ts ...`

packages/tool-server/test/metro/integration.test.ts
```diff
@@ -6,6 +6,12 @@ import { jsRuntimeDebuggerBlueprint } from "../../src/blueprints/js-runtime-debu
 import { debuggerConnectTool } from "../../src/tools/debugger/debugger-connect";
 import { createDebuggerStatusTool } from "../../src/tools/debugger/debugger-status";
 import { debuggerEvaluateTool } from "../../src/tools/debugger/debugger-evaluate";
+import { scopeTempHome } from "../helpers/temp-home";
+
+// The JS-runtime-debugger / network blueprints build a real LogFileWriter,
+// whose constructor mkdir -p's os.homedir()/.argent/tmp. Keep that out of the
+// developer's real home.
+scopeTempHome("argent-metro-integration-home-");
```
latekvo, line 14: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: scopeTempHome skips sibling files in the same directory that also build a real LogFileWriter.

The LogFileWriter sweep is incomplete on this branch's head: `packages/tool-server/test/metro/teardown-log-history.test.ts` sits in the same directory as the three metro files that received `scopeTempHome`, drives the identical path, and still runs against the developer's real `$HOME`. It registers the real `jsRuntimeDebuggerBlueprint` (line 81) and resolves the service through `debugger-connect`, so `src/blueprints/js-runtime-debugger.ts:245` runs `new LogFileWriter(port)`, whose constructor `mkdirSync`s `os.homedir()/.argent/tmp` and opens `argent-logs-<port>-<ts>.log` there; `connectAndCapture` then writes up to 60 real entries into it. Trigger: `rm -rf ~/.argent/tmp && npx vitest run test/metro/teardown-log-history.test.ts` inside `packages/tool-server` recreates the directory in the real home - exactly the outcome the PR description reports as no longer happening after the fix ("rm -rf ~/.argent/tmp then a full suite run: recreated. After: not recreated"), so that evidence and the "LogFileWriter (8 files)" count are both false as merged.

packages/telemetry/test/index.test.ts
```diff
@@ -60,7 +60,14 @@ vi.mock("posthog-node", () => {
 describe("telemetry public surface", () => {
   const { tmp } = scopeHome();
 
+  // DO_NOT_TRACK is a consortium-standard opt-out any developer may export.
+  // Left ambient it disables consent for the whole suite, so these tests would
+  // assert what the developer's shell says rather than what markEnabled() does.
+  let restoreOptOut: () => void;
+
   beforeEach(() => {
+    restoreOptOut = snapshotEnv(["DO_NOT_TRACK"]);
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The consent pin covers DO_NOT_TRACK but not ARGENT_TELEMETRY, which the same function checks.

packages/telemetry/test/index.test.ts:69-70 pins only DO_NOT_TRACK, but getConsentState checks two env opt-outs at equal precedence: DO_NOT_TRACK (packages/telemetry/src/consent.ts:105) and then ARGENT_TELEMETRY whenever its value is one of 0/false/no/off (parseFalsy, consent.ts:92-96 and 112-118). Both short-circuit before readConfigOverride() at consent.ts:129, so neither can be overridden by the config flag this suite's beforeEach markEnabled() writes. With `ARGENT_TELEMETRY=0` exported in the developer's shell, track() returns at the `if (!isEnabled()) return;` guard (src/index.ts:192), posthogMock.instances stays empty, and the ~13 consent-dependent tests in this file die on `posthogMock.instances[0]!` being undefined - identical to the DO_NOT_TRACK=1 failure this hunk fixes ("282 passed -> 13 failed" in the PR body). Nothing else clears it: packages/telemetry/vitest.config.ts declares no setupFiles or env, there is no repo-root vitest config, test/helpers.ts touches only HOME/USERPROFILE plus caller-named keys, and no workflow sets it.

latekvo (commented):

packages/tool-server/test/bind-failure-telemetry.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 4289e3d8f.

Reproduced first. Project-scoped `packages/tool-server/.argent/flags.json` with `tool-server-event-log: true`, then the four files that call the real `start()`:

```
Test Files  2 failed | 2 passed (4)
     Tests  3 failed | 4 passed (7)
TypeError: Cannot read properties of undefined (reading 'on')
 ❯ attachRegistryEventLogger src/event-log.ts:111:19
 ❯ start src/index.ts:197:5
 ❯ test/tool-fail-invalid-params-emission.test.ts:87:5
```

The two failures are exactly `startup-telemetry.test.ts` and `tool-fail-invalid-params-emission.test.ts`; `bind-failure-telemetry` and `lens-relay-telemetry` pass.

One correction to the trigger, in the direction of worse: `ARGENT_EVENT_LOG` cannot redirect the write away from the real home. `vitest.config.ts` loads `test/setup/clear-argent-env.ts`, which deletes every `ARGENT_*` key before the module graph is imported, so `createToolServerEventLog` always falls through to `homedir()/.argent/tool-server-events.jsonl`. I ran the repro with `ARGENT_EVENT_LOG` pointed at a scratch path and my real log was truncated anyway — `-rw-r--r-- 0 bytes 17:20`, the minute of the run (`event-log.ts:42` is an unconditional `writeFileSync(filePath, "")`).

The same `vi.mock("@argent/configuration-core", …)` guard is now on both files. Re-run with the flag on: 4 files / 7 tests passed, and a `sentinel` line written into `~/.argent/tool-server-events.jsonl` before the run was still there after it. Flag off: 4 files / 7 tests passed, unchanged.

packages/tool-server/test/helpers/temp-home.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 4289e3d8f — the class was real, though not the four files named.

I did not take the file list on trust; I swept it with an instrument. Every test file referencing `LogFileWriter` or a blueprint that builds one (16 files) was run under `HOME=$probe USERPROFILE=$probe`, then `$probe/.argent/tmp` was tested for existence — a file carrying `scopeTempHome` writes into its own mkdtemp instead, so the probe discriminates cleanly. Result on 0bf93788c:

```
test/debugger/debugger-recovery-reachability.test.ts   POLLUTES-HOME
test/debugger/debugger-status-not-connected.test.ts    POLLUTES-HOME
test/debugger/log-registry-not-connected.test.ts       POLLUTES-HOME
test/metro/teardown-log-history.test.ts                POLLUTES-HOME
(the other 12, including all three debugger/ files that received the helper)  clean
```

So `teardown-log-history.test.ts` is confirmed, but "the three debugger/ files" in this comment refers to `log-file-writer.test.ts`, `log-file-writer-level.test.ts` and `log-registry.test.ts`, which the PR did pin — those three probe clean. The three that actually leak are `debugger-recovery-reachability`, `debugger-status-not-connected` and `log-registry-not-connected`. Count coincides, files do not.

`scopeTempHome` is now on all four. Re-probed: all four `clean`, and green — 8, 2, 13 and 6 tests passed respectively.

packages/tool-server/test/android-binary-windows.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 4289e3d8f, and extended to the sibling in the same diff.

Verified the mechanism rather than asserting it. Since `os.homedir()` is native and ignores `USERPROFILE` on darwin, I ran the resolver under a `vi.mock("node:os")` that reproduces the Windows lookup (`USERPROFILE ?? …`), with a stock Studio layout planted at `$USERPROFILE/AppData/Local/Android/Sdk/platform-tools/adb.exe`:

```
✓ PRE-FIX (HOME only): the real profile SDK leaks in      → resolveAndroidBinary("adb") !== null
✓ POST-FIX (HOME + USERPROFILE): lookup misses as intended → resolveAndroidBinary("adb") === null
```

The pre-fix arm is what `returns null when only the extensionless binary exists` would hit on a Windows host. `USERPROFILE` is now pinned alongside `HOME` and added to `ENV_KEYS` so it is restored too.

`adb-resolve-avd-path.test.ts` carries the identical defect and is in the same windows-e2e job (`.github/workflows/windows-e2e.yml:91`): this PR added a `HOME`-only pin there for the `os.homedir()/.android/avd` default root, and `src/utils/adb.ts:774` states the reason it cannot work — "os.homedir() (not $HOME) so the default root resolves on Windows too, where it's %USERPROFILE%\\.android\\avd and $HOME is typically unset". Both files now pin the pair; the comments name both variables and why. 6 passed / 1 skipped.

packages/tool-server/test/metro/integration.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 4289e3d8f. Same finding as the thread on `test/helpers/temp-home.ts:13`; the detail is there.

Short version: confirmed by probing all 16 `LogFileWriter`-reaching test files under a scratch `HOME`/`USERPROFILE` and checking for `$probe/.argent/tmp`. Four leaked — `metro/teardown-log-history`, `debugger/debugger-recovery-reachability`, `debugger/debugger-status-not-connected`, `debugger/log-registry-not-connected`. All four now call `scopeTempHome`, all four re-probe clean and stay green.

packages/telemetry/test/index.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 4289e3d8f.

Reproduced in `packages/telemetry`:

```
ARGENT_TELEMETRY=0     → Tests  17 failed | 6 passed (23)
ARGENT_TELEMETRY=false → Tests  17 failed | 6 passed (23)
clean                  → Tests  23 passed (23)
```

17 rather than the ~13 estimated, and `index.test.ts` is the only file in the package that breaks — the full suite goes 306 passed → `1 failed | 16 passed`. `DO_NOT_TRACK=1` already passes 306/306 on the branch, so this was the one remaining half of the pair.

The pin is now a `CONSENT_ENV_KEYS` list covering both, snapshotted and cleared together. Re-run: 23 passed under `ARGENT_TELEMETRY=0`, `ARGENT_TELEMETRY=false`, `DO_NOT_TRACK=1` and clean.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Review of `test/isolate-global-resources` at `3306b6f40`, closing out the five earlier threads and eight further passes. Everything found was fixed on the branch rather than left as a comment, so this is the record of what was checked and what moved.

**Scope check.** 63 files, `+1019 / -240`, all under `test/`. `git diff --name-only origin/main...HEAD -- 'packages/*/src'` is empty at every commit on the branch, so no production behaviour is in this diff.

**Earlier threads.** All five are fixed in `4289e3d8f` and answered in place. The `temp-home` and `metro/integration` ones were the same finding at two sites; the fix generalised to all twelve files that call `scopeTempHome`.

**Found and fixed here** (each reproduced before the change and re-run after):

- `095d75bd8` — `workspace-reader-integration` pinned `PATH` at a directory of extensionless shims. On Windows `beforeAll` throws `EPERM` at `symlink()`, and a copy would not rescue it either, since a bare command name resolves only through `PATHEXT`. Skipped at the `describe`, which keeps `beforeAll` from running at all.
- `26c2c60ec`, `3bb69ded6` — a failing assertion between spawn and the `spawnedPids.push` left a real tool-server running while the same hook deleted the record that could find it. Forcing a failure took the `tool-server.cjs start` count 10 → 11. Fixed at all four launcher files, not the one where it was found: `launcher-spawn`, `launcher-exit` and `launcher-duplicate-spawn` had no net at all, and the last armed one for 1 of 9 spawns. Proof the net fires: a child reparented to init, listening on 127.0.0.1:52300, spawned 21:28:36 and exited on its own at 21:29:36 with nothing signalling it. 60s outlasts the longest test here (30s).
- `3306b6f40` — `boot-electron-spawn-error` completes a boot, and `persistPorts` writes `os.homedir()/.argent/chromium-cdp-ports.json`; the real file on this machine held 106 dead ephemeral ports. Two sibling files already carried the mock. Same commit adds `temp-home.test.ts`, because deleting the `USERPROFILE` line from `scopeTempHome` left all 48 tests green — the helper was inert on Windows and nothing said so.
- `f37b3605e`, `a02af22e4` — three comments asserting things the code does not do: a `USERPROFILE` rationale resting on a Windows case its own file excludes, an `fs` mock described as covering three SDK roots when deleting any of them leaves the file green, and a claim that clearing the `ARGENT_` prefix guards four reads when only `ARGENT_HOST` changes an outcome.
- `6f558e669` reverts `df4f9c6f8`. The vega sentinel collision is real and reproducible, but #711 already fixes it and fixes it better — it carries the entropy in the fraction of the sleep duration under a 600s cap, where mine spelled the sentinel as the duration and would have leaked `sleep <pid>0X` processes for days.

**Verified.** 6003 passed / 1 skipped / 0 failed across all 12 suites (tool-server sharded 6×), 0 stray `fake-tool-server` and 0 stray `tool-server.cjs` after the run, CI 7/7 green at `3306b6f40` including the Windows job. Guard tests were checked by mutation, not by reading: removing `USERPROFILE` → 2 failed, sharing one home path across tests → 1 failed, restoring → green.

**Out of scope, filed separately:** #852 (`ensureToolsServer` respawn), #853 (`listen(0)` omits host — hardening; the Host allow-list already blocks browsers, a raw socket forging `Host: 127.0.0.1` still gets 200), #854 (`argent-mcp` temp leftovers), #858 (323 accumulated chromium screenshot PNGs).

No HIGH or MEDIUM finding survives at `3306b6f40`.

PR #715, opened 2026-08-04
perf(flow): stop the recorder returning the whole flow file per step

hubgan:
## What changes

`flow-add-step` and `flow-add-echo` returned the **entire growing YAML** on every call. Across measured recording sessions that was the majority of all tool-result text, and it grows quadratically with step count.

- Both tools now return `stepCount` — how many steps the flow has — instead of `flowFile`.
- `flow-add-step` additionally returns `recorded`: the one summary line that was appended.
- The full file still comes back once, from `flow-finish-recording`.

## Why

The cost is not just tokens. With the whole file echoed per step, context pressure during a long walkthrough visibly changes the artifact being produced — checks get trimmed from a test to buy back room. Returning a line per step removes that pressure without removing any information the author needs *at that moment*.

## Supporting change

`summarizeStep` is lifted out of `flow-finish-recording`'s per-file loop so the recorder can render its one line exactly the way the final summary does — one spelling, not two that can drift.

That line now also carries `times`, `duration` and `delayMs`. They change what replays, and since the recorder no longer returns the YAML, this line is the author's only per-step view of what landed in the file.

## Compatibility

This is a **breaking change to the tool result shape**: `flowFile` is gone from `flow-add-step` and `flow-add-echo`. `savedTo` is unchanged and is still the only field naming the destination in remote-client mode. `flow-finish-recording`'s result **shape** is unchanged, but its `summary` strings are not: taps now render `×N`, long-presses `for Nms`, and tool steps `(after Nms)`, so the two surfaces spell a step the same way. Nothing outside the agent reads `summary`, so that is not a break.

## Tests

Recorder tests now assert against the flow **as persisted on disk** rather than a returned string, which is the surface that actually matters.


j-piasecki (commented):
Adversarial review of the diff at `2fa6897`. Four things worth acting on, all small — nothing here blocks. Two are load-bearing (a comment whose stated invariant this PR falsifies, and a coverage gap on `stepCount`); two are nits.

One point on the PR description rather than the code: **Compatibility says "`flow-finish-recording` is unchanged"**, and its result *shape* is — but its `summary` strings are not. Taps now render `×N`, long-presses `for Nms`, tool steps `(after Nms)`. The PR's own `finish-recording's summary carries the same delay/times spellings as recorded` test pins the new output. Nothing outside the agent consumes `summary`, so it isn't a break — the section just reads broader than it is.

### Checked and found nothing, for the record

- **No consumer of `flowFile` from `flow-add-step`/`flow-add-echo` exists** outside the tool-server tests this PR updates — grepped `.ts`/`.tsx`/`.md`/`.json` across `packages/`. The breaking change is contained.
- **Telemetry is an allowlist** (`ALLOWED` in `packages/telemetry/src/sanitize.ts`), so the new `recorded` field is not emitted. Swapping a whole-YAML field for a per-step summary line cannot widen what leaves the machine.
- **`recorded` renders the in-memory step while the file holds the serialize→parse round-trip of it.** That is only sound because parse/serialize are exact inverses, and the one place the in-memory copy can hold something the file cannot — `times: 1` — is explicitly handled and tested. The round-trip is also pinned directly (`result.recorded === summarizeStep(parseFlow(onDisk)…)`) in five places, across both host and client mode.
- **The takeover scenario** in the skill's "Pick a name unique to your task" bullet does not get worse: an agent whose key was stolen now sees `stepCount: 1` where it expected 6, which is a sharper signal than the truncated YAML it used to get.
- **`savedTo: null`, as the SKILL.md commit documents, is real** — `applyClientFileDirectives` in `packages/argent-tools-client/src/file-inputs.ts` resolves a rejected path or a failed write to `null`.
- `yarn ts-check` clean; `flow-tools`, `flow-remote-recording`, `flow-record-tap`, `flow-concurrent-recording` and `interaction-messages` all green.

Not filed inline: `flow-add-step` now imports from `flow-finish-recording` for a pure renderer. No cycle today, and moving `summarizeStep` + `selectorLabel` + `textConditionLabel` + `renderToolArgs` into their own module is more churn than this PR warrants — noting it only because the dependency direction (recorder → finisher) is the one that will read oddly to the next person in this file.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
```diff
@@ -163,60 +164,86 @@ function renderToolArgs(args: unknown): string {
...
-        return `${n}. tool: ${step.name} ${renderToolArgs(step.args)}`;
+  return flow.steps.map((step, i) => summarizeStep(step, i + 1));
+}
+
+/**
+ * One recorded step, rendered the way the flow FILE spells it. Shared with the
+ * recorder, which echoes just the line it appended instead of the whole
+ * growing file.
+ */
+export function summarizeStep(step: FlowStep, n: number): string {
```
j-piasecki, line 206: Exporting `summarizeStep` falsifies the reachability argument in `renderToolArgs`'s doc comment, ~25 lines above (lines 152–157):

> No reachable input is undefined today: every caller comes through {@link summarizeSteps}, which is only ever handed `parseFlow` output, and `fromYamlStep` normalises a missing/`null` `args:` to `{}` on the way through.

After this PR `summarizeStep` is public and `flow-add-step` calls it with a step it built in memory — one that went through neither `parseFlow` nor `fromYamlStep`.

The guard is still *correct*: `flow-add-step` builds `args` as `stripDeviceKeys(params.args ? JSON.parse(params.args) : {})`, which cannot be `undefined`. But the recorded reason is exactly what a future reader would use to decide whether the `try`/`catch` and the `${}` interpolation are still load-bearing, and that reason no longer holds — the second caller reaches the same `default:` arm by a path the comment says does not exist.

Worth re-stating as "every caller hands it either `parseFlow` output or a step the recorder built", with the recorder's own `args` normalisation as the second half of the argument.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -2659,14 +2659,18 @@ function assertSessionStillLive(session: RecordingSession, step: FlowStep): void
...
     session.lastTouchedSeq = touch();
     if (session.persist === "host") {
       const flowFile = await appendStep(session.filePath, step);
       session.flow = parseFlow(flowFile);
-      return { flowFile, savedTo: session.filePath };
+      // Count inside the lock, off the just-refreshed `session.flow`: a caller
+      // reading `session.flow.steps.length` after this returns would be racing
+      // a concurrent same-key append, which can reassign `session.flow` between
+      // the release here and that read.
+      return { savedTo: session.filePath, stepCount: session.flow.steps.length };
```
j-piasecki, line 3235: `stepCount` is now the number every `recorded` line opens with, and nothing in the suite pins it to the **re-read file** rather than to the number of appends.

The two only diverge when the file is hand-edited mid-recording — a documented workflow, and one this PR expands the docs for in three places. `test/flows/flow-concurrent-recording.test.ts:808` already builds exactly that scenario (two echoes → edit the `.yaml` down to one → append a third) and asserts the markers on disk, but not the count. Every `stepCount` assertion added in this PR appends without editing, so appends == file length in all of them.

Concretely: swap this for a session-local `++session.appendCount` and the whole suite still passes, while an author who deleted a step mid-recording is told "3." for what is line 2 of their file. That is the one surface where the count is not merely informational — it numbers the author's only per-step view.

Two lines on that existing test close it (`addEcho` already returns the result):

```ts
const third = await addEcho(root, "alpha", "s3");
expect(third.stepCount).toBe(2);
```

packages/tool-server/src/tools/flows/flow-finish-recording.ts
```diff
@@ -163,60 +164,86 @@ function renderToolArgs(args: unknown): string {
   }
 }
 
+/** The pre-step sleep a replay performs, when the step carries one. */
+function delayLabel(step: FlowStep): string {
```
j-piasecki: Nit. `delayLabel` is declared over the whole `FlowStep` union, and that is the only reason it needs the `as` cast — which is also what stops the compiler from checking the field. Its single call site has already narrowed to the `tool` arm, where `step.name` and `step.args` are read directly on the same line.

Either narrow the parameter:

```ts
function delayLabel(step: Extract<FlowStep, { kind: "tool" }>): string {
  return step.delayMs !== undefined ? ` (after ${step.delayMs}ms)` : "";
}
```

or drop the helper — `${step.delayMs !== undefined ? ` (after ${step.delayMs}ms)` : ""}` typechecks as-is on that arm, since the exhaustive cases leave `default:` contributing `never` and the arm narrows to the `tool` member. I compiled the direct access against the package tsconfig to confirm.

No behaviour change either way; `renders a tool step's pre-step delay` covers the output.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
```diff
@@ -163,60 +164,86 @@ function renderToolArgs(args: unknown): string {
...
+    case "launch":
+      return `${n}. launch: ${typeof step.app === "string" ? step.app : JSON.stringify(step.app)}`;
+    case "run":
+      return `${n}. run: ${step.flow}`;
+    case "tap":
+    case "long-press": {
+      // `times` (tap) and `duration` (long-press) change what replays, so a
+      // summary line that drops them misdescribes the file, and since the
+      // recorder stopped returning the YAML per step this line is the author's
+      // only per-step view of what was appended. Neither kind carries a
```
j-piasecki: Nit, comment accuracy. Half this justification does not apply to half the branch it is attached to.

`long-press` has no recorder path — `flow-add-step` only ever builds `tap`, `launch`, `run` and `tool` steps — so a `long-press` line never reaches an author through `recorded`, only through `flow-finish-recording`'s `summary`, which still returns `flowFile` beside it. This PR's own test says as much: *"long-press steps have no live recorder path, so this is the only coverage of that rendering."*

The first clause ("change what replays, so a summary line that drops them misdescribes the file") justifies `duration` on its own and needs no help. Suggest scoping "since the recorder stopped returning the YAML per step this line is the author's only per-step view" to `times`.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
hubgan: Fixed in 4d67e9cd — the comment now states both paths: the finisher's `parseFlow` output, and the step the recorder builds, whose `args` is a fresh spread out of `stripDeviceKeys` and so an object either way.

packages/tool-server/src/tools/flows/flow-utils.ts
hubgan: Fixed in 9ca58dda — added to that test. Confirmed the gap first: swapping the count for a session-local `++appendCount` left 917/918 green, and this assertion was the only one that caught it (`expected 3 to be 2`).

packages/tool-server/src/tools/flows/flow-finish-recording.ts
hubgan: Fixed in 6ff747da — narrowed to `Extract<FlowStep, { kind: "tool" }>`, cast gone. Kept the runtime `typeof`: `fromYamlStep` copies `delayMs` across unvalidated, so a hand-edited `delayMs: soon` reaches the renderer as a string and `!== undefined` prints `(after soonms)`. Pinned that in a test, since nothing covered it.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
hubgan: Fixed in b81a9df1 — the "author's only per-step view" half is now scoped to `times`, with `long-press`'s reachability (finish-recording's `summary` only, which still returns `flowFile`) stated next to it.

j-piasecki (approved):

packages/tool-server/src/tools/flows/flow-finish-recording.ts
```diff
@@ -163,60 +167,95 @@ function renderToolArgs(args: unknown): string {
...
+/**
+ * The pre-step sleep a replay performs, when the step carries one. Narrowed to
+ * the one arm that has a `delayMs` — over the whole union the field could only
+ * be read through a cast, which is also what would stop the compiler checking
+ * it. The runtime `typeof` still earns its keep: `fromYamlStep` copies `delayMs`
+ * across unvalidated, so a hand-edited `delayMs: soon` arrives here as a string
+ * and must render nothing rather than `(after soonms)`.
+ */
+function delayLabel(step: Extract<FlowStep, { kind: "tool" }>): string {
+  return typeof step.delayMs === "number" ? ` (after ${step.delayMs}ms)` : "";
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `typeof step.delayMs === "number"` and the test the *runner* applies to the same field are not the same test, so the line this produces can be wrong in both directions.

The runner gates on truthiness and then hands the raw value to `setTimeout`: `flow-run.ts:2167` is `if (step.delayMs && !(await sleepOrAbort(step.delayMs, signal)))`, and `sleepOrAbort` (`utils/timing.ts:8`) passes `ms` straight through. So a truthy non-number still produces a real sleep by JS coercion, while `NaN` — which *is* a `number` — produces none. `fromYamlStep` copies `delayMs` across unvalidated and `validateFlow` does not check it, so both survive a parse; a quoted numeric is an ordinary slip in the hand-edit workflow the SKILL.md bullet in this PR expands on.

I replayed four one-step flows (`tool: list-devices`) through `flow-execute` against a tool-server built from d185472e, timing each, then rendered the same four files through `summarizeStep`:

| flow file | replay wall time | rendered line |
| --- | --- | --- |
| no `delayMs` | 581 ms | `1. tool: list-devices {}` |
| `delayMs: 2000` | 2577 ms | `1. tool: list-devices {} (after 2000ms)` |
| `delayMs: "2000"` | 2582 ms | `1. tool: list-devices {}` |
| `delayMs: .nan` | 575 ms | `1. tool: list-devices {} (after NaNms)` |

The quoted case sleeps two full seconds on every replay and the summary says nothing about it. The `.nan` case reports a delay that never happens — a statement that could not be made before this PR, since the `tool:` arm previously rendered no delay at all.

The doc comment above says a hand-edited non-number "must render nothing rather than `(after soonms)`", and the runtime `typeof` was kept in 6ff747da on that basis. That holds only if such a value is inert to the runner, and it is not — `"soon"` happens to coerce to `NaN` and sleep ~0, but `"2000"` does not.

Reachable only through `flow-finish-recording`'s `summary`: `flow-add-step`'s `recorded` renders the step the recorder built, and `params.delayMs` is a zod-validated integer there.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
```diff
@@ -163,60 +167,95 @@ function renderToolArgs(args: unknown): string {
...
+      // per-step view of what was appended. `long-press` has no recorder path,
+      // so it only ever reaches an author through flow-finish-recording's
+      // `summary`, which still returns `flowFile` beside it. Neither kind
+      // carries a `delayMs` (only `tool` steps do), so no delayLabel here.
+      const target = step.selector ? selectorLabel(step.selector) : `(${step.x}, ${step.y})`;
+      // Only ×2..×10 is renderable: `times: 1` is the default and never lands in
+      // the file (parseTapTimes normalizes it to absent), so rendering `×1` for
+      // a stray in-memory `times: 1` would describe a file that can't exist.
+      const times =
+        step.kind === "tap" && step.times !== undefined && step.times > 1 ? ` ×${step.times}` : "";
```
latekvo, line 237: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The rendered count is pinned only by its presence and by the `> 1` boundary, never by its value.

Replacing `×${step.times}` with a constant `×2` leaves the whole `packages/tool-server` suite green — 919/919, `tsc --build` exit 0 — in a separate worktree at this commit. Every assertion that renders a count uses `times: 2`: `flow-tools.test.ts:495` (`recorded`), `:546` (the finish `summary`), `:2728` (direct), plus the negative `times: 1` case at `:2733`. The only `times: 3` in the suite is `flow-tap.test.ts:81`, which asserts parse/serialize and never a rendered line, so nothing observes a count other than 2.

That range is reachable: `gesture-tap` accepts `clickCount` up to 10, `flow-add-step` records it as `times`, and `parseTapTimes` admits 2..10 — the comment two lines above says as much ("Only ×2..×10 is renderable"). Under that mutation a recorded triple-tap renders `×2` on both surfaces, and per the comment at 202-209 the `recorded` line is the author's only per-step view of what was appended, so nothing else would show the wrong count.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
```diff
@@ -163,60 +167,95 @@ function renderToolArgs(args: unknown): string {
...
+      const held =
+        step.kind === "long-press" && step.duration !== undefined ? ` for ${step.duration}ms` : "";
+      return `${n}. ${step.kind}: ${target}${times}${held}`;
     }
-  });
+    case "type":
+      return `${n}. type: ${selectorLabel(step.into)} ← "${step.text}"`;
```
latekvo, line 243: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `type.submit` and `await.timeout` are dropped while `times`, `duration` and `delayMs` are now rendered, and the argument stated at 202-203 covers all five equally: "`times` (tap) and `duration` (long-press) change what replays, so a summary line that drops them misdescribes the file."

Both fields survive the parser and neither arm reads them, so a step renders identically with and without. Running `parseFlow` + `summarizeStep` at d185472e:

```
type submit:false    step:   {"kind":"type","into":{"identifier":"pw"},"text":"hunter2","submit":false}
                     render: 1. type: {"id":"pw"} ← "hunter2"

await timeout:15000  step:   {"kind":"await","condition":"visible","selector":{"identifier":"home"},"timeout":15000}
                     render: 1. await: visible {"id":"home"}
```

`submit: true` normalises away, so `submit: false` — the case that suppresses the Enter press — is exactly the one that reaches here. `long-press.duration` was added under this rule despite having no recorder path at all, so recorder-reachability is not the line being drawn between them. Neither field appears in the new `summarizeStep rendering` block.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -451,8 +456,9 @@ export function createFlowAddStepTool(
...
-If a step was recorded by mistake, edit the .yaml file directly to remove it.`,
+    description: `Execute a tool call and record it as a step in the flow named by \`name\` + \`project_root\` (the recording must already be open — see flow-start-recording). Use when recording a flow and you want to run and capture each action. A coordinate \`gesture-tap\` is recorded as a portable \`tap: { selector }\` step when the tapped element has stable text/identifier (otherwise coordinates are kept with a warning); a \`restart-app\` is recorded as a \`launch\` step (record one FIRST to make the flow a self-contained e2e flow; restart-app has no chromium support, so a chromium flow records as a fragment — add the \`launch: { chromium: <app path> }\` line to the YAML afterward, deleting the executionPrerequisite line if one was recorded: a flow that starts with a launch must not declare it).
+Returns { message, toolResult, stepCount, recorded, savedTo } on success. If it fails an error is returned and nothing is recorded.
+If a step was recorded by mistake, edit the .yaml to remove it.`,
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This sentence, and its twin in `flow-start-recording`'s description ("If a recorded step turns out to be wrong, you can edit the .yaml file directly to remove or reorder steps"), are unqualified — while the SKILL.md bullet rewritten in this PR now states the same edit is discarded against a remote client.

In client mode `appendStepToFlow` never re-reads the client's file: it pushes onto the in-memory `session.flow` and returns `savedTo: clientFileDirective(session.filePath, serializeFlow(session.flow))`, a whole-file overwrite that still contains the step the author deleted. The client applies it and the tool reports success, so the removal is silently undone with nothing in the result indicating it.

67598723 ("qualify the mid-recording edit advice everywhere it is given") added that caveat to `flow-start-recording`'s description for exactly this reason; d185472e removed it again while keeping SKILL.md's half, so the skill and the two tool descriptions now answer the same question in opposite ways. `flow-finish-recording`'s wording is scoped to "afterwards" and is unaffected.

packages/skills/skills/argent-create-flow/SKILL.md
```diff
@@ -137,15 +137,15 @@ The standalone command uses only the auto-started local tool server. It is unava
 | `flow-read-prerequisite` | Read a flow's execution prerequisite without running it (same `name`/`flow_path` sources)                                                               |
 | `flow-execute`           | Replay a flow — a saved one by `name`, or any flow YAML by absolute `flow_path`                                                                         |
 
-Every tool during recording returns the current flow file contents, so you can track what has been recorded. Rules:
+**Rules:**
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The sentence replaced here was the only place the skill told the author what the recording tools return — "Every tool during recording returns the current flow file contents, so you can track what has been recorded." After the replacement the file names no result field of any recording tool: `stepCount`, `flowFile` and `savedTo` occur zero times in SKILL.md, and `recorded` appears only as the English past participle.

The Recording section still walks start → add-step → add-echo → finish, and step 5 tells the author to read the saved `.yaml` and convert raw `tool:` steps, but nothing points at the field that now carries the per-step view. That matters most for the rewrites: this PR's own description says `recorded` exists because "since the recorder no longer returns the YAML, this line is the author's only per-step view of what landed in the file", and a coordinate tap stored as a selector, or a `restart-app` stored as a `launch`, shows up nowhere else per step.

aa1550b6 wrote a replacement paragraph naming `recorded`/`stepCount`/`flowFile`/`savedTo`; d185472e removed it. Neither CI gate covers the gap — `grade-skills.mjs` scores structure only (this skill still reports 10.0), and the tool-description gate keys on the literal `Returns { … }` substring, which is still present.

packages/skills/skills/argent-create-flow/SKILL.md
```diff
@@ -137,15 +137,15 @@ The standalone command uses only the auto-started local tool server. It is unava
...
-- **Mistakes can be edited out.** Edit the `.yaml` file directly to remove or reorder steps.
+- **Mistakes can be edited out.** Edit the `.yaml` file directly to remove or reorder steps. Mid-recording, this holds only when the tool-server shares your filesystem (the usual local case): the recorder re-reads the file before each append, so your edit is kept — provided it still parses, since the next step re-validates the whole file and fails on a botched one instead of quietly keeping it. Against a **remote client** the in-memory copy is authoritative and the next append overwrites the file, so edit after `flow-finish-recording`. Editing after the recording is finished is always safe.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Two things this bullet names are narrower than the behaviour behind them.

**Remote client — "the next append overwrites the file".** `flow-finish-recording` overwrites too, with no append involved. Its client branch serializes the in-memory flow and returns it as a client-file directive (`flow-finish-recording.ts:102-104`, `flowFile = serializeFlow(session.flow); savedTo = clientFileDirective(filePath, flowFile)`), which the client writes over the edited file. A reader who has already made the edit and reads that only an append is dangerous can reasonably conclude that finishing instead is safe. The actionable half — edit after `flow-finish-recording` — is correct, since the session is cleared by then; it is the stated reason that does not cover the finish path.

**Host mode — "your edit is kept — provided it still parses".** Parsing is the only failure mode named, but the append rewrites the file from the parsed model (`appendStep` is `parseFlow` → mutate → `serializeFlow` → write), so anything `parseFlow` does not model is dropped even when the edit parsed cleanly. Comments are the visible case. Against a tool-server built from d185472e, host mode:

```
$ cat .argent/flows/cm.yaml
# Checkout smoke flow - keep the waits, they gate on the spinner
steps:
  # opens the cart drawer
  - echo: one

$ curl -XPOST .../tools/flow-add-echo -d '{"name":"cm","project_root":"…","message":"two"}'
{"data":{"message":"Echo added to \"cm\" flow","stepCount":2,"savedTo":"…/cm.yaml"}}

$ cat .argent/flows/cm.yaml
steps:
  - echo: one
  - echo: two
```

Both comments are gone and the call reports plain success. The two operations the bullet names — remove and reorder steps — do survive, so this is about the scope of "your edit"; annotating a flow while recording is a plausible reading of it. The rewrite predates this PR, but `flow-add-echo`/`flow-add-step` used to return `flowFile`, i.e. the post-append comment-free YAML, so the loss showed up in the very next result.

latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Correcting one clause in the host-mode half above: I wrote that the loss "showed up in the very next result" pre-PR. That is stronger than the truth. `appendStep` returns `serializeFlow(flow)` (flow-utils.ts:2884 at 4c30ae25), so the old `flowFile` carried the *post*-append YAML — already comment-free, and byte-indistinguishable from a file that never had a comment. It flagged nothing; it merely put the file's contents in front of you, so an author who remembered their own annotation could notice the absence.

The finding itself is unaffected: the comments are still dropped, the call still reports plain success, and the bullet still names only "provided it still parses" as the way a mid-recording host-mode edit can fail to survive. Note also that the rewrite predates this PR and that `flow-start-recording` and `flow-finish-recording` do both still return `flowFile` — so this is about the bullet's stated scope, not about a signal this PR took away.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
hubgan: Fixed in 1abef919 — `delayLabel` now gates on truthiness the way the runner does, then renders what `setTimeout` will actually wait, dropping anything it floors to an immediate tick. Reproduced your table first against a server built from d185472e, then re-ran it on the fix: `"2000"` sleeps ~2s and now renders `(after 2000ms)`; `.nan` sleeps none and now renders nothing.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
hubgan: Fixed in 7f73215b — added `times: 3` and `times: 10` cases. Confirmed the gap: with `×${step.times}` replaced by a constant `×2`, the new case is the only failure in the block (`expected ... ×2 to be ... ×3`), the other eight still pass.

packages/tool-server/src/tools/flows/flow-add-step.ts
hubgan: Fixed in 877b025c — restored the caveat on both descriptions, scoped to "every write" rather than "the next append", since `flow-finish-recording` serializes the same in-memory copy. Verified both come back over `GET /tools` carrying it.

packages/skills/skills/argent-create-flow/SKILL.md
hubgan: Fixed in 76edb703 — the skill names `recorded`, `stepCount`, `flowFile` and `savedTo` again, and says what `recorded` is for (the tap/restart-app rewrites appear nowhere else per step), without re-explaining the rest.

packages/skills/skills/argent-create-flow/SKILL.md
hubgan: Fixed in d9b0249a — the remote-client clause now says any write overwrites, naming `flow-finish-recording` alongside the append; the host-mode clause now says only what the parser models survives. Reproduced the comment drop against a server built from this branch: both YAML comments gone after one `flow-add-echo`, plain success reported.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
hubgan: Not fixed, and I think out of scope here. The behaviour is unchanged by this PR — the `type` and `await` arms are byte-identical to 4c30ae25, and both fields render identically with and without on base. What this PR added was the rule, stated more broadly than it is applied. 000fee58 records the boundary instead: neither kind is recorder-built, so both reach an author only through the finish `summary`, beside the `flowFile` that spells them out. Rendering them is a fair follow-up, just not one this per-step view opened.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
hubgan: Resolved in 1abef919

packages/tool-server/src/tools/flows/flow-finish-recording.ts
hubgan: Resolved in 7f73215b

packages/tool-server/src/tools/flows/flow-add-step.ts
hubgan: Resolved in 877b025c

packages/skills/skills/argent-create-flow/SKILL.md
hubgan: Resolved in 76edb703

packages/skills/skills/argent-create-flow/SKILL.md
hubgan: Resolved in d9b0249a

PR #716, opened 2026-08-04
perf(mcp): cap the auto-screenshot settle wait at three seconds

hubgan:
## What changes

`run-sequence`'s auto-screenshot settle cap drops from **15s to 3s**, and a test pins the invariant across the whole table.

## Why

The cap was being read as "time for the sequence to finish". It is not — a nested sequence's own steps have already settled by the time the tool returns. It only bounds the case where a screen **never** settles.

So on any screen with something permanently in motion (a spinner, a looping animation, a video), every `run-sequence` call sat out the full 15 seconds and then took a screenshot that was going to be a blur regardless. One measured session paid that 33 times.

3s is past every real transition.

## Scope

`packages/argent-mcp` only — no dependency on anything else in flight. The new test asserts *no* tool in `AUTO_SCREENSHOT_DELAY_MS_BY_TOOL` exceeds 3s, so a future entry cannot quietly reintroduce a long dead wait.


packages/argent-mcp/test/auto-screenshot.test.ts
```diff
@@ -153,6 +153,32 @@ describe("getAutoScreenshotDelayMs", () => {
...
+  // An in-app-transition cap only bounds a screen that never settles, so a
+  // generous value there buys nothing but dead wait — this guards against a
+  // future entry reintroducing one (run-sequence's old 15s was exactly that).
+  // The app-launch tools are exempt: launch-app/restart-app/open-url start an
+  // app (or, for a web open-url, the browser), a real cold start that can
+  // legitimately exceed 3s, so their cap is safety headroom, not the in-app dead
+  // wait this guard targets. (The effective runtime delay can still be raised
+  // above the cap via the ARGENT_AUTO_SCREENSHOT_DELAY_MS floor, covered
+  // separately below.)
+  const APP_LAUNCH_TOOLS = new Set(["launch-app", "restart-app", "open-url"]);
```
latekvo, line 165: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: The PR description states the new test asserts *no* tool in `AUTO_SCREENSHOT_DELAY_MS_BY_TOOL` exceeds 3s, "so a future entry cannot quietly reintroduce a long dead wait". At this head the assertion exempts `launch-app`/`restart-app`/`open-url` (this set), and the test comment above argues why an app-launch cap can legitimately exceed 3s. So the blanket guarantee the description promises does not hold for the app-launch family: a future entry there could be added above 3s without this test noticing. It holds today only because the three exempted values happen to sit at or below 3000. The description predates the exemption commit and overstates the guard's coverage.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: Reviewed the PR's true surface (its own 4 commits: 2 files, 34 lines) at head bb97035f3.

Moves covered: claims vs code (the new settle-cap comment checked against the run-sequence implementation - ALLOWED_TOOLS contains only in-app interactions, DEFAULT_INTER_STEP_DELAY_MS = 100, per-step delayMs override; all accurate), nearest twin (standalone gesture/button tools already settle at 1500ms caps, run-sequence now gets 2x their budget; app-launch caps unchanged and exempt), non-happy paths (the ARGENT_AUTO_SCREENSHOT_DELAY_MS floor can still raise the effective runtime delay; the new tests are env-immune by asserting the map constant directly), reachability (the cap is the await-screen-idle timeoutMs upper bound at mcp-server.ts:332, so the change is live), and what outlives the call (nothing accumulates - this only shortens a wait).

Absence pass: the file's suite passes 34/34 at head. Mutating run-sequence back to 15000 (the exact regression this PR fixes) fails both new tests; mutating it to 2000 (in-range drift) fails only the value pin - each test is load-bearing. No stale 15000 copy survives in source (dist/ is untracked). One finding below.

packages/argent-mcp/test/auto-screenshot.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Still present as of bb97035f30fb64b7dd31c558d26e1ade7db1bae4.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Reviewed the PR's net surface vs main (2 files, +34/-1 — the branch's other commits are content-identical to what main already carries via the squash-merged #574, so GitHub's 70-file listing is a stale merge-base artifact).

Moves covered: claims vs code (the new cap comment checked against packages/tool-server/src/tools/run-sequence/index.ts — ALLOWED_TOOLS holds only in-app interactions, tv-remote and await-ui-element included, no launch-app/open-url; DEFAULT_INTER_STEP_DELAY_MS = 100 with a per-step delayMs override), nearest twin (run-sequence at 3000 sits at 2x the standalone interaction caps; ARGENT_AUTO_SCREENSHOT_DELAY_MS still floors the effective runtime delay above any cap), non-happy paths (if await-screen-idle fails, the fallback sleeps the full budget — now bounded at 3s; consumer passes the value as timeoutMs at mcp-server.ts:326), reachability both ways (run-sequence is in AUTO_SCREENSHOT_TOOLS and normalizeToolName resolves the mcp__argent__ prefix before the map lookup), and what outlives the call (nothing accumulates; the change only shortens a wait).

Absence pass: no stale 15s prose survives in docs or skills; nothing else consumes AUTO_SCREENSHOT_DELAY_MS_BY_TOOL. Ran the file's suite at head: 34/34 pass. Mutating run-sequence back to 15000 fails both new tests; drifting it to 2000 fails only the value pin — each test is load-bearing. The one finding stands on the existing thread below; nothing further surfaced.

PR #717, opened 2026-08-04
fix(flow): re-probe a recorded wait against the tree the runner reads

hubgan:
> Stacked on #715.

## The problem

`await-ui-element` evaluates against the tree it reads — the **accessibility hierarchy** on iOS and Android, the **CDP DOM** on Chromium, the **toolkit page source** on Vega. The `await:` / `assert:` directive that the polish pass converts a recorded wait into is evaluated against the **runner's** tree instead.

They overlap, but outside Vega neither contains the other — an id present in one can be absent from the other, and on iOS even the role vocabularies are disjoint. Both directions have been hit in practice: an id visible to the runner but not the recorder, and text-field values visible to the recorder but not the runner.

The consequence is that a check can pass live and fail once converted. That makes the core promise — *each step is executed live so you verify it works before it's recorded* — untrue at exactly the point it is relied on.

## What changes

After a recorded `await-ui-element` succeeds, the recorder re-runs the same condition against the tree the runner resolves directives against, and reports the result on the step it just wrote.

**It warns; it does not refuse.** What gets written is a raw `tool: await-ui-element` step, and at replay that tool reads the same tree it just passed against — so "this would fail every run" was never true of the form actually recorded. What the probe genuinely reports is whether the **polish-time conversion** is safe, which is why the warning says to act on it *then* rather than re-record now (re-recording would duplicate the step).

Because polish begins after `flow-finish-recording`, each verdict is also carried there: it is repeated on its own step's `summary` line, and `message` counts how many steps carry one. A warning raised at step 7 of a 40-step recording is otherwise long out of reach by the time it is actionable.

## Three outcomes, kept distinct

| Runner tree says | Reported as |
| --- | --- |
| condition holds | nothing — step recorded clean |
| condition does not hold | recorded + warning naming the causes to rule out before converting |
| tree unreadable, or too slow to answer | recorded + warning that the conversion is **UNKNOWN, not known-bad** |

The middle row is deliberately not "the conversion will fail". The identical verdict comes back when the screen simply moved on between the live wait and the re-probe, and on `text` checks when the selector matches several elements and the two sides elect different ones — so the consequence stays conditional on the cause, and the message names the causes in the order worth ruling out.

The third row matters: nothing was compared, so calling it a divergence would send the author off to rewrite a selector that may be perfectly good. It is also the injection-free case the flow skills explicitly sanction, so refusing there would block a supported way of working. On iOS the quoted reason ends "provide bundleId explicitly" — advice aimed at the native tools. The warning corrects it, because no directive carries a bundleId and forwarding one would make the probe answer a question the runner never asks.

## Platform-specific explanation

The *why* differs per platform, and stating the iOS story everywhere makes the message wrong where someone is trying to act on it. iOS gets AX-vs-native-hierarchy; Chromium gets "the flow tree keeps only addressable nodes"; Android gets the importance trim; Vega gets "the trees agree, so the screen changed". The named tool for reading the runner's side follows the same rule — pointing an Android author at `native-find-views` points at a tool they cannot call.

The remedy is per **condition** as well as per platform: on `hidden` a determinate verdict means the runner's tree still *has* the element, so "retarget at an id the full hierarchy carries" is the opposite of what helps.

## A failed live wait is reported by cause

`await-ui-element` returns `{ success: false }` for three different things: a condition read and found false, a tree source that never answered, and a cancelled wait. Only the first judges the condition, so only the first is narrated as "the wait never held" and prescribes re-recording or deleting the step. The other two observed nothing and are reported as UNKNOWN — the same unknown-vs-known-bad distinction the probe draws for the runner's tree.

A wait nested inside a recorded `run-sequence` gets neither warning: that tool reports its own shape, which carries no `success` key. The skill says so, and says to read `toolResult` there.


packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -67,6 +73,462 @@ function fallbackSourceWarning(source: DescribeSource, platform: string): string
...
+  }
+  // Determinate: the trees really were compared and really do disagree, so this
+  // is the one warning that may explain the divergence and name how to read the
+  // runner's side.
+  return {
+    warning:
+      `recorded, but this condition does NOT hold against the tree the runner resolves ` +
+      `directives against (${cappedReason(outcome.reason ?? "no match")}). As the raw ` +
+      `\`tool: ${AWAIT_UI_ELEMENT_TOOL_ID}\` step it replays fine — it reads the same tree it ` +
+      `just passed against — but an \`assert:\` conversion WILL fail (it reads that tree on ` +
```
j-piasecki: **The warning asserts a consequence with certainty, then names a cause under which that certainty doesn't hold.**

Assembled, the determinate message reads:

> …but an `assert:` conversion **WILL fail** (it reads that tree on the same short grace this probe just used), and an `await:` will too unless … \[spelling clause\] … *A screen that changed between the live wait and this re-probe reads the same way, so rule that out first.*

On the screen-changed branch the conversion does **not** fail. The probe's verdict is about the device *now*; a converted `assert:` runs at replay immediately after the wait it was converted from, against a screen the wait has just gated into the right state. A toast that expired between the live wait and the re-probe yields this exact verdict with the two trees in perfect agreement — and its `assert:` conversion passes.

So "WILL fail" is only warranted on the tree-divergence branch of the disjunction the message itself opens two sentences later. And "rule that out first" only makes sense as an instruction if ruling it out is what settles whether the conversion fails — as written, the message has already decided.

This looks like a leftover from 571869a8 (*"stop the cross-tree warning asserting one cause it cannot know"*). That commit did the right thing to the **cause** — `SCREEN_MAY_HAVE_MOVED` is now appended on every platform — but the **consequence** upstream of it stayed absolute. The `await:` clause on the next line inherits the same certainty.

Conditioning it would fix it, e.g. `…an \`assert:\` conversion will fail the same way, and an \`await:\` too unless <awaitStillNeeds> within its longer timeout — unless the screen simply moved on.` Two assertions in `flow-record-cross-tree.test.ts` pin the current string and would move with it: `"an \`assert:\` conversion WILL fail"` (iOS case) and `"an \`await:\` will too unless the element reaches that tree"`.

Worth noting this is the one platform-independent sentence in the message — Vega is where it's most plainly wrong, since `treeDivergenceFor`'s Vega arm states outright that a disagreement there *means* the screen changed, and this sentence still tells that author the conversion WILL fail.

packages/skills/skills/argent-create-flow/SKILL.md
```diff
@@ -66,7 +66,7 @@ One way `next` is deliberately **looser than CSS `+`**: where `A + B` matches no
...
...iews on iOS and system chrome on Android — but so do `describe` and the raw tool, so those are not divergences.) The traffic runs the other way too: the flow tree walks 12000 nodes where the agent-facing `describe` stops at 5000, so on a dense screen it holds elements `describe` never reached. **Neither tree contains the other**, which is why recording an `await-ui-element` step re-probes the condition against the flow tree and warns you when the two disagree (see Recording). And when several elements match — including wrappers whose native text aggregates descendant content — the action directives (`tap`, `type`, `scroll-to`) pick the **most specific** match: an exact text/identifier match beats a substring hit (for a regex matcher, a pattern consuming the element's whole text counts as exact), then the smallest frame wins. (A universal `any: true` selector has no field to be exact about, so its matches rank by reading order instead — the first element in the scope, which is the element a condition reads too. Where two matches share a top-left corner, an action breaks the tie toward the smaller, more specific one and a condition does not, so those two can name different elements.)
```
j-piasecki: **"It keeps only views a selector can address _and_ that are actually on screen" is false on Vega**, and it's stated here as an unqualified property of the flow tree.

`projectVegaNode` returns `skip: false` for every node, and its own comment says so:

> Every parsed node is emitted as a leaf … Zero-area (scrolled-off) nodes stay too — `exists` deliberately accepts them

Nothing is dropped there for being off screen or unaddressable. This is the same overstatement the PR already removed from the runtime warning — `treeDivergenceFor`'s Vega arm now says the flow tree "drops no element", and `flow-record-cross-tree.test.ts` asserts the Vega warning does *not* contain `"different projections of the screen"`. The skill's leading clause still asserts it.

The per-platform list immediately after covers Chromium / iOS / Android only, so scoping the sentence to those three is enough: *"On Chromium, iOS and Android it keeps only views a selector can address and that are actually on screen, and what that costs you…"*. Vega's absence from the list already reads correctly as "no cost" — it's the absolute in front of it that contradicts.

(Two smaller things in the same paragraph I checked and am **not** flagging: the 12000-vs-5000 walk limit holds on Chromium *and* Android — `FLOW_WALK_LIMITS` and `FLOW_MAX_NODES` both raise the helper's 5000 default — so the unqualified phrasing is fair on the platforms where a walk limit exists at all. And the iOS / Android / Chromium divergences listed here each reproduce through the real adapters in the new suite.)

packages/tool-server/test/flows/flow-record-cross-tree.test.ts
```diff
@@ -0,0 +1,1106 @@
...
+ * unconditional prefix of EVERY message, warnings included — so a regression
+ * that nags on every correctly-recorded wait would sail through. Split the
+ * prefix off and require the remainder to be absent instead.
+ */
+/**
+ * The probe's own reason, as the determinate warning quotes it back — the only
+ * part of the message carrying screen content, and so the only part the cap
+ * governs. Asserting on the whole warning instead measures the fixed prose
+ * around it, which moves whenever the explanation is reworded.
+ */
```
j-piasecki: **Two JSDoc blocks stack on `echoedReasonOf`, and the first one documents a different function.**

The block at 206–213 describes splitting the `Step added to "<name>" flow` prefix off and requiring the remainder to be absent — that is `warningOf` (line 230), which is now undocumented. `echoedReasonOf` has its own block at 214–219.

Reads like `echoedReasonOf` was inserted above `warningOf` (in 571869a8, which added it) and landed between `warningOf`'s doc and its body. Moving the first block down to sit on `warningOf` restores both.

Worth doing rather than leaving: the stranded comment carries the reason the helper exists at all (asserting `message` contains "Step added" proves nothing, since that prefix is unconditional) — which is exactly the rationale a future edit to `warningOf` needs and will not find there.

j-piasecki (commented):
Adversarial review of the cross-tree re-probe. Verified locally on `27503f4`: `tsc --build` clean, `prettier --check` clean, `flow-record-cross-tree.test.ts` 31/31 green, `scripts/extract-tools.test.mjs` 46/46 green.

Three findings below. Everything else I chased came back clean and is not reported: `settleWithin` already consumes the abandoned promise (no unhandled rejection on the timeout path); `resolveDevice` really has no throw path, so the no-try/catch comment holds; `AbortSignal.any` is already established in `flow-pixels`; the `giveUp` abort genuinely prevents the post-deadline `finalPoll` read; `identifier` is a parse-only alias in flow YAML, so "copy the recorded `selector:` map through unchanged" is sound; and the iOS merge / Android resource-id shield / Chromium `[password]` divergences all reproduce through the real adapters.

I also dropped two candidates after checking them, since neither survives as a defect:

- **Flakiness in `keeps the tail of an over-long reason`.** `probeStartedAt` is captured before `startRecording`, so the fixture's 900 ms trusted-read window is shortened by however long the setup takes; past ~300 ms the dark tail exceeds `CONDITION_DARK_TAIL_TOLERANCE_MS` (600 ms) and the verdict flips to indeterminate, failing the test. But I measured the actual gap at 1–4 ms over five runs — a ~100x margin. Worth knowing the threshold exists; not worth changing.
- **Suite wall-clock (~24 s for the file).** Each determinate-warning case burns the full 1 s assert grace, and `probeWhenCondition` hardcodes `DEFAULT_ASSERT_TIMEOUT_MS` with no seam to shorten it. Real cost, but the only fixes are a production seam or fake timers over an async poll loop — more risk than the seconds are worth.

packages/skills/skills/argent-create-flow/SKILL.md
```diff
@@ -181,7 +181,7 @@ Record an `await-ui-element` step to **gate** the next step on a screen transiti
...
...less the step passed `textMatch: equals`, while the directive has no default and forces you to pick — so a recorded step with no `textMatch` converts to `contains:`, and only `textMatch: equals` converts to `equals:`. Picking the wrong one fails on the very screen the recorder approved: a wait for `Total` expecting `$5.00` on a `Total: $5.00` row converts to `{ text: { in: { text: Total }, contains: "$5.00" } }`, where the `equals` spelling fails with `its text was "Total: $5.00"`. Converting also moves the wait off the tree the raw tool reads and onto the flow tree, and neither contains the other (see Selectors) — so check `message` on the recorded step first. If the recorder warned that the condition does NOT hold against the runner's tree, converting is exactly what would break it: retarget the directive at something the flow tree does carry (usually the `id`) and prove it by running the flow, or leave that one step raw. Otherwise convert — keeping a step raw for any other reason costs you the auto-wait and the full `testID` coverage, so the only standing exceptions are a custom `pollIntervalMs`/`bundleId` the directive can't express and a warned step you haven't retargeted yet.
```
j-piasecki: **On Vega this tells the author to do the one thing the warning explicitly tells them not to.**

Recorded through the real Vega adapter (`udid: amazon-…`), `message` ends:

> …Both read the same automation-toolkit page source, and the flow tree only re-shapes it — it drops no element and its text hoist can only add matches — so on this platform **a disagreement means the SCREEN changed** between the live wait and this re-probe, not that the two trees differ. `describe` reads the same source the runner does, so **re-run the wait rather than re-recording the selector**

This bullet reads the same warning the other way: *"converting is exactly what would break it: retarget the directive at something the flow tree does carry (usually the `id`)"*. `flow-record-cross-tree.test.ts:849-851` pins the runtime side against that exact wording:

```js
expect(warning).not.toContain("retarget the DIRECTIVE");
expect(warning).not.toContain("re-record with a selector");
expect(warning).toContain("re-run the wait");
```

The absolute also outruns the other three platforms as of 3a629291d: the message now splits the consequence across two causes — *"if the trees really do differ over this element…"* / *"if the SCREEN simply moved on since the live wait, this verdict is no evidence against either"* — and "converting is exactly what would break it" states the first as if it were the only one.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -457,7 +934,8 @@ export function createFlowAddStepTool(registry: Registry): ToolDefinition<
...
+A recorded \`await-ui-element\` is re-probed against the tree the RUNNER resolves \`await:\`/\`assert:\` directives against, which is NOT the tree the live call read; when the two disagree the step is still recorded and \`message\` carries a warning saying the conversion would fail. The probe judges the selector exactly as recorded, so write the conversion in the strict map spelling (\`{ visible: { text: Continue } }\`, copying the step's \`selector:\`) — the bare-string spelling (\`{ visible: Continue }\`) re-parses as a loose selector that resolves identifier-first and falls back to text, which is a different check. \`message\` also warns when the live wait never held — that tool reports an unmet condition by returning \`{ success: false }\` rather than failing, so the step is recorded and will stop the run at replay.
```
j-piasecki: **The description promises a verdict the message it describes stopped giving two commits ago.**

Description: *"when the two disagree the step is still recorded and `message` carries a warning saying the conversion would fail."*

What `message` actually carries, dumped from the real path:

> What conversion costs you depends on WHY the two disagree: if the trees really do differ over this element, an `assert:` conversion fails the same way (it reads that tree on the same short grace this probe just used), and an `await:` does too unless … ; **if the SCREEN simply moved on since the live wait, this verdict is no evidence against either** — at replay the directive runs where that wait ran, not a moment after it.

3a629291d and bbbce187b rewrote that sentence and its assertions; this description string is from 528d8fe05 and did not move with them. `flow-record-cross-tree.test.ts:589` and `:843` now assert the message does **not** contain "WILL fail", while the description still tells its reader a warning means the conversion fails.

j-piasecki (commented):
Re-review after the previous round. All three earlier comments are addressed: the determinate consequence is now split per cause in `probeAgainstRunnerTree` (and `flow-record-cross-tree.test.ts` asserts the message no longer contains "WILL fail"); `SKILL.md`'s flow-tree drop rule is scoped to Chromium/iOS/Android; the stranded JSDoc block is back on `warningOf`, with `echoedReasonOf` keeping its own.

Verified locally on `bbbce187b`: `tsc --build` clean, `prettier --check` clean on all four files, `flow-record-cross-tree.test.ts` 31/31, `scripts/extract-tools.test.mjs` 46/46.

Claims I chased and cleared:

- **"As the raw `tool:` step it replays fine."** `flow-run.ts:2149-2171` runs a `kind: "tool"` step through `invokeSubTool`, so a recorded `await-ui-element` reads its own describe tree at replay. Holds.
- **The Chromium walk-limit numbers.** `DEFAULT_WALK_LIMITS.maxNodes` is 5000 (`describe/platforms/chromium.ts:53`), `FLOW_WALK_LIMITS.maxNodes` is 12000 (`flow-chromium-tree.ts:94`). Both directions of that clause hold.
- **`PROBE_BUDGET_MS`'s Android claim.** `getHierarchy` is the one method that gets `LONG_RPC_TIMEOUT_MS` = 15s (`android-devtools-client.ts:26,121`). Holds.
- **The probe's cost on the clean path.** `records the step, with no warning, when both trees agree` runs in 16 ms — `waitForCondition` returns on the first poll, so only a condition that does not hold burns the 1s assert grace.
- **`typeof selector !== "object"` silently skipping a real wait.** `selectorSchema` is a refined object schema (`ui-tree-match.ts:70`), so a recorded `await-ui-element` can never carry a string selector. Guard is dead by construction, not a gap.
- **Flake risk in the two ~4s tests.** `gives each platform a distinct remedy` and `gives up on a tree read that outruns the probe budget` measured 4020-4032 ms and 4057-4062 ms over five runs, and both carry explicit `15_000` timeouts — nothing near vitest's 5s default. (The file's other tests top out at ~1013 ms, and the suite already has 7755 ms tests elsewhere.)

Two comments, both on doc surfaces the two conditioning commits did not reach.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -67,6 +73,477 @@ function fallbackSourceWarning(source: DescribeSource, platform: string): string
...
+ * message false exactly where the author is trying to act on it.
+ *
+ * Each story names what the two projections do differently WITHOUT asserting
+ * which side ended up missing the element: on Chromium both directions are
+ * reachable, and a message that only ever describes the runner dropping
+ * something sends an author to fix the wrong end.
+ */
+function treeDivergenceFor(udid: unknown): string {
+  const platform = platformOf(udid);
+  if (platform === "ios") {
```
latekvo, line 210: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: **The iOS arm of `treeDivergenceFor` is the one platform arm no test pins** - and iOS is the default platform of nearly every fixture in the new file.

Mutation-tested at `bbbce187` against `test/flows/flow-record-cross-tree.test.ts` (baseline 31/31, ~24 s), one edit at a time, file restored and re-verified between each:

| edit | result |
| --- | --- |
| iOS arm body -> `return "";` | **31 passed (31)** |
| Vega arm body -> `return "";` | 1 failed \| 30 passed |
| Android arm body -> `return "";` | 1 failed \| 30 passed |

Android and Chromium are pinned by `expect(warning).toContain("first. No read-only tool exposes the ...")` (lines 684-686 and 731-733), which holds only because `SCREEN_MAY_HAVE_MOVED` ends "...rule that out first." and line 543 joins the divergence clause to the reader clause with a space. Vega is pinned at line 836 by `toContain("the SCREEN changed between the live wait and this re-probe")`. iOS gets no positive assertion on any of it: `it("iOS: warns when the AX tree's merged label exists on no single view")` (line 558) reaches only strings from `runnerSideReadClause` and the fixed prose in `probeAgainstRunnerTree`. The two places the iOS text is named - lines 373 and 904 - are `expect(warning).not.toContain("neither contains the other")`, which pin its **absence** on the unmet-wait and indeterminate branches. And `it("gives each platform a distinct remedy")` (line 875) does read the whole warning, but asserts only `new Set(...).size === 4`, which the four already-distinct `runnerSideReadClause` texts satisfy on their own.

What that leaves unguarded is specifically the `SCREEN_MAY_HAVE_MOVED` append on line 204: dropping it from the iOS arm alone ships green. Every iOS divergence warning would then assert the two trees differ with nothing telling the author that a screen which simply moved on returns the identical verdict - while the consequence sentence upstream still says "if the SCREEN simply moved on since the live wait, this verdict is no evidence against either", conditioning itself on a cause its own platform clause no longer raises. That per-platform admission of the screen-moved cause is the premise the conditional consequence rests on, and on the platform where it matters most it is held in place by nothing.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -67,6 +73,477 @@ function fallbackSourceWarning(source: DescribeSource, platform: string): string
...
+} as const;
+
+/**
+ * The clause naming how to read the tree the RUNNER resolves against — or, on
+ * iOS, Android and Chromium, that no read-only tool does.
+ *
+ * Android's runner tree is the full accessibility hierarchy, and Android
+ * `describe` returns the TRIMMED interactables tree the recorder already read.
+ * Chromium's runner tree keeps only addressable nodes that are on screen, yet
+ * `describe` returns the FULL DOM the recorder read — a superset that still
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: **This block's Chromium premise is false, and the same file retracts it 160 lines further down.**

Here: "`describe` returns the FULL DOM the recorder read - a superset that still shows the very nodes the runner drops".

At lines 261-265, inside `treeDivergenceFor`'s chromium arm: "The agent-facing walk stops at DEFAULT_WALK_LIMITS (5000 nodes); FLOW_WALK_LIMITS raises it to 12000. Past 5000 it is the RECORDER's tree that is short, and `describe` cannot show the element at all - which also made "`describe` returns the full DOM the recorder read" false exactly when it mattered." That comment quotes this sentence verbatim and calls it false.

Reproduced on a live Chromium target (Electron, branch-built tool-server at `bbbce187`): a page of 9000 id-bearing divs with a viewport-pinned `<button id="marker" aria-label="Marker Element">` last in document order. `describe` returned 4997 `n` nodes and no `Marker Element`. Recording `await-ui-element {condition: "hidden", selector: {identifier: "marker"}}` then returned `success: true` live - the recorder genuinely could not see it - and the probe answered `an element matching id="marker" was still visible`. The runner's tree holds a node `describe` cannot show, which is the opposite of the superset relation asserted here.

The consequence is scoped to this file rather than to a user: this block is the stated rationale for how the Chromium clause is worded, and the runtime clause it justifies was deliberately made two-directional ("it both lists nodes the runner drops and omits nodes the runner keeps"). A maintainer who takes the one-directional premise stated here as given has the argument for collapsing that clause back.

packages/skills/skills/argent-create-flow/SKILL.md
```diff
@@ -169,9 +169,9 @@ command: "flow-execute"
...
...ixed `delayMs`. If its condition is not met before the timeout, replay **stops at that step** (the steps after it assume the transition happened). See the `await-ui-element` section of `argent-device-interact` for the full condition/selector reference. The live call sees only what `describe` reads — on iOS and Android that is the trimmed accessibility tree, so if it can't find an identifier you know exists, gate on visible text to get the step recorded, then retarget the identifier in the `await:` form during polish (the directive resolves the full hierarchy — see Selectors); don't conclude the testID is unusable in the flow. The recorder covers the opposite case for you: after the wait runs live it **re-probes the same condition against the flow tree** and warns in `message` when the check would not survive conversion to `await:`/`assert:`. The step is still recorded — a raw `tool:` step replays against the tree it just passed — so treat that warning as a polish-time instruction, not a failure. A wait that **did not hold live** is warned about too, and that one is a real problem: it is recorded anyway (see _Every step runs live_) and will stop the run at that step on every replay.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: **"warns in `message` when the check would not survive conversion" collapses the third outcome - the one that warns in order to say the opposite.**

The PR description keeps three outcomes apart on purpose, and its own table's third row is "tree unreadable -> recorded + warning that the conversion is **UNKNOWN, not known-bad**". `flow-add-step.ts:541-546` emits exactly that: "...so it passed against the tree `await-ui-element` reads and nothing else. Whether it would convert to `await:`/`assert:` is UNKNOWN, not known-bad - re-probe once that tree source is back before trusting the conversion". The code comment two lines above states the reasoning: nothing was compared, so claiming the trees differ "would send the author to rewrite a selector that may be perfectly good".

This sentence is the skill's definition of what a warning on the step means, and it admits only the determinate reading. The indeterminate triggers are reachable and named by the PR's own `PROBE_BUDGET_MS` block at `flow-add-step.ts:340-367`: an iOS recording whose full-hierarchy service is unavailable (`flow-tree.ts` deliberately refuses to degrade to a fallback), an Android `getHierarchy` failure, or any tree read that outruns the 4000 ms budget. An author who hits one gets a warning whose own text says the answer is unknown, having been told here that a warning means the check would not survive - and the distinction the three-outcome design exists to preserve is exactly the one lost at that point.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat)\]: Ran the full sweep at `bbbce187`.

**End-to-end, on a real device.** Branch-built tool-server driven over HTTP against a live Chromium (Electron) target, recording through the real `flow-add-step` path. All three outcomes fire correctly: a wait both trees agree on records clean; an off-viewport `exists` and a `{role: input, text: Passphrase}` wait on a password field each produce the determinate warning; an unmet wait produces the unmet-wait warning with `fetchFlowTree` never called. I then replayed both forms with `flow-execute` to test the warning's two predictions - the raw `tool:` steps pass 2/2, and both `assert:` conversions fail with exactly the reason the probe quoted. The two remedies the Chromium clause prescribes also pass when run: `scroll-to` before the check for a zero-height frame, and `id`/`role` for a password field.

The reverse divergence direction reproduces too: on a 9000-node page with a viewport-pinned `<button id="marker">` last in document order, `describe` stopped at 4997 nodes without it while the flow tree carried it, so a `hidden` wait passed live and the probe warned - with the `hidden`-specific `awaitStillNeeds` clause ("unless the element LEAVES that tree").

**Suite.** `tsc --build` clean; `flow-record-cross-tree.test.ts` 31/31. I mutation-tested it in an isolated worktree: `PROBE_BUDGET_MS` (both directions), `MAX_PROBE_REASON_CHARS`/`PROBE_REASON_TAIL_CHARS`, deleting `giveUp.abort()`, flattening `awaitStillNeeds`'s per-condition arms, swapping the ios/android arms of `runnerSideReadClause`, short-circuiting the indeterminate branch, dropping the `isUnmetUiWaitResult` gate, and removing `cappedReason` are each caught by a failing test. Three comments below.

Claims I chased and am **not** reporting:

- **A wait nested in a recorded `run-sequence` gets no `message` warning.** Reproduced E2E - recorded silently, and at replay the flow fails with `run-sequence stopped at await-ui-element after 0 of 1 steps`. But `toolResult` comes back `{completed: 0, total: 1, steps: [{tool, error}]}`, the skill twice instructs the author to read exactly that, and `flow-nested-outcome.ts:12-21` states the tool-scoped-branches doctrine deliberately. The rule this PR rewrote ("only steps that RETURNED successfully") describes the case correctly.
- **The probe's abandoned tree read holding the Android devtools client.** The mechanism is real - I measured 5006 ms of head-of-line blocking against the real `connectAndroidDevtoolsClient` with a fake helper socket, since `chain.then(send, send)` serialises every RPC and the in-flight `getHierarchy` is uncancellable. But `captureTapSelector` already awaited `fetchFlowTree` on that same chain unbounded at `a39e06fe`, and the only >4s read measured anywhere in the repo is attributed to Chromium, whose CDP transport has no request chain.
- **`cappedReason` returning a longer string than its input** for reasons of 201-218 chars. The dropped-character count is correct and the band is cosmetic.
- **The 12000-vs-5000 sentence at SKILL.md:71 being unqualified** where iOS and Vega have no node cap. Android's `describe` does cap at 5000 - `getHierarchy`'s `maxNodes ?? 5000` default in `blueprints/android-devtools.ts:340` - so the claim holds on both platforms that have a walk limit at all.
- **`ios-remote` reaching `UNSUPPORTED_PLATFORM`.** `classifyDevice` emits exactly the five platforms the JSDoc names, and `platformMatrix` maps `ios-remote` to `capability.appleRemote`, which `await-ui-element` does not declare - so `assertSupported` throws while the step is still executing live. The unreachability claim holds.

packages/tool-server/src/tools/flows/flow-add-step.ts
hubgan: Resolved in 3a629291 and bbbce187.

packages/skills/skills/argent-create-flow/SKILL.md
hubgan: Resolved in 8d7b0414.

packages/tool-server/test/flows/flow-record-cross-tree.test.ts
hubgan: Resolved in 3fc1816e.

packages/skills/skills/argent-create-flow/SKILL.md
hubgan: Resolved in 016c8756, with a follow-up in ed77fde6 for the same absolutism left in the bullet's closing exception.

packages/tool-server/src/tools/flows/flow-add-step.ts
hubgan: Resolved in 9f374d4a.

packages/tool-server/src/tools/flows/flow-add-step.ts
hubgan: Resolved in cbd752f5.

packages/tool-server/src/tools/flows/flow-add-step.ts
hubgan: Resolved in f73036d8.

packages/skills/skills/argent-create-flow/SKILL.md
hubgan: Resolved in 1a88c424.

packages/tool-server/src/tools/await-ui-element/index.ts
```diff
@@ -49,6 +49,41 @@ export function isUnmetUiWaitResult(tool: string, result: unknown): boolean {
...
+ * Telling an author to re-record or delete a step on the strength of a cause
+ * that observed nothing sends them to rewrite something that may be fine.
+ */
+export type UnmetUiWaitCause = "unmet" | "unreadable" | "cancelled";
+
+export function unmetUiWaitCause(result: unknown): UnmetUiWaitCause {
+  const note = (result as { note?: unknown } | null)?.note;
+  if (typeof note !== "string") return "unmet";
+  if (note === WAIT_CANCELLED_NOTE) return "cancelled";
+  if (note.startsWith(TREE_FETCH_FAILED_NOTE_PREFIX) || note === HIDDEN_UNREADABLE_NOTE) {
```
j-piasecki: **`note === HIDDEN_UNREADABLE_NOTE` never matches the note the tool builds for the case it is written for**, so the AX-down `hidden` wait classifies as `unmet`.

`timeoutNote` returns `appendDiagnostics(base, lastData)` (index.ts:255), which appends the boot hint / restart prompt whenever `lastData` carries one — and an empty-tree-plus-`hint` read is exactly what `isBlindRead`'s first clause fires on. The fetch-error arm survives because it returns before `appendDiagnostics`; this one does not.

Running the real tool with a failing AX service, `condition: "hidden"`:

```
note:  "could not confirm the element is hidden — the UI tree was empty or unreadable at timeout (This simulator was not booted through argent — system dialogs and native modals may not appear. You MUST call boot-device with force=true now…)"
cause: unmet
```

Feeding that note through the recorder:

```
recorded, but the wait itself never held — `await-ui-element` reports an unmet condition by
returning success:false instead of failing… so re-record it once the condition can actually
hold, or delete the step from the .yaml — in host (local) mode…
```

`await-ui-element.test.ts:624` already pins that this note carries the boot hint (`expect(result.note).toMatch(/boot-device/i)`). `flow-record-cross-tree.test.ts:489` feeds the bare constant instead, so the `unreadable` arm stays green.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -548,6 +1274,14 @@ If a step was recorded by mistake, edit the .yaml to remove it — against a rem
 
       const { savedTo, stepCount } = await appendStepToFlow(session, step);
 
+      // Keep the probe's verdict for `flow-finish-recording`. It is a
+      // polish-time answer, and polish starts after the recording closes — by
+      // which point this `message` is many tool results back.
+      session.stepsAppended = stepCount;
```
j-piasecki: **`stepsAppended` is maintained here only, but `flow-add-echo` is a second appender** — one echo after the last recorded wait desyncs it and `flow-finish-recording` drops every verdict.

`flow-insert-echo.ts:38` calls `appendStepToFlow` and never touches `session.stepsAppended`.

Recording a diverging wait, then one `flow-add-echo`, then finishing:

```
summary[0]: 1. tool: await-ui-element {"condition":"visible","selector":{"text":"Continue"}}
message:    Finished recording "echodrop" flow (2 steps)
```

No `warning:` line and no count, though the wait's own `message` carried the full determinate verdict. Without the echo the same recording produces `warning:` on `summary[0]` and `1 step carries a cross-tree warning` — the assertions at flow-record-cross-tree.test.ts:1515.

`flow-start-recording`'s description tells the author to "Use flow-add-echo to add labels".

packages/tool-server/src/tools/flows/flow-finish-recording.ts
```diff
@@ -53,6 +54,60 @@ const zodSchema = z.object({
...
+/**
+ * The verdicts still anchored to the steps they judged, or none when a
+ * mid-recording edit changed the flow's length (see {@link attachStepWarnings}).
+ */
+function anchoredWarnings(
+  session: RecordingSession,
+  steps: number
+): Map<number, string> | undefined {
+  if (session.stepWarnings === undefined) return undefined;
+  return session.stepsAppended === steps ? session.stepWarnings : undefined;
```
j-piasecki: **The length equality only catches an edit that is the last thing before the finish.** A delete followed by another append puts the length back, so nothing is dropped and a verdict lands on a step it never judged.

Probe: step 1 diverges and carries the verdict, step 2 agrees, the author deletes step 1 from the .yaml, then records a third wait that also agrees, then finishes.

```
summary[0]: 1. tool: await-ui-element {"condition":"visible","selector":{"text":"Continue"}}
   warning: recorded, but this condition does NOT hold against the tree the runner resolves
            directives against (no element matched selector text="Continue")…
message:    Finished recording "relen" flow (2 steps) — 1 step carries a cross-tree warning…
```

Both surviving steps agreed with the runner's tree. `stepsAppended` is 2 and `flow.steps.length` is 2, so `anchoredWarnings` keeps the map.

Delete-then-re-record is the sequence `UNMET_WAIT_WARNING` itself prescribes: "delete the step from the .yaml — in host (local) mode, where the recorder re-reads the file before each append".

j-piasecki (commented):
Verified and cleared:

- `describe` stops at 5000 nodes, the runner at 12000 — `DEFAULT_WALK_LIMITS` (chromium.ts:53), `FLOW_WALK_LIMITS` (flow-chromium-tree.ts:94).
- The Chromium divergence text — `projectChromiumNode` keeps a node only when `onScreen && addressable`, and redacts a password leaf's label to `[password]` (flow-chromium-tree.ts:37-52).
- The Vega story ("drops no element, and its hoist can only add matches") — `projectVegaNode` returns `skip: false` with a leaf for every node (flow-vega-tree.ts:38-49).
- "At equal depth" on iOS — `native-full-hierarchy` documents `maxDepth` default 8; `queryFullHierarchyTree` sends 40 (flow-ios-tree.ts:301).
- `TEXT_TIE_CLAUSE`'s enumeration-order claim — `findAll` pushes a node before recursing (ui-tree-match.ts:529-532); `flattenHoisting` emits after its child loop (flow-tree-flatten.ts:157-161).
- Probe budget arithmetic — `DEFAULT_ASSERT_TIMEOUT_MS` is 1000 (flow-actions.ts:232) and `LONG_RPC_TIMEOUT_MS` is 15000 (android-devtools-client.ts:26).
- `cappedReason` never emits over 200 chars, and never longer than its input — brute-forced over input lengths 1..200000.
- The abandoned probe raises no unhandled rejection on the timeout path — `settleWithin` attaches its handlers before the race (timing.ts:52-55).
- `AbortSignal.any` is available on the supported floor (`node >= 20.12.0`) and already used at flow-pixels.ts:399.
- `crossTreeWarning` can only ever land on the `tool:` branch of the step switch, so no earlier arm can drop it.
- Flows suite on this head: 44 files, 1094 tests, 0 failures, 72.7s.

The comments are on the verdict-anchoring scheme and the unmet-cause classifier.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: Ran the full review sweep at the current head 9c52f3a7a1abe342b96583ed02044cb4cd6bd809: claims vs code, nearest twin, non-happy paths, inputs, reachability, what-outlives-the-call, and the four absence moves (sibling/prose/symmetry/mutation) across flow-add-step.ts, await-ui-element/index.ts, flow-finish-recording.ts, flow-utils.ts, flow-actions.ts, and the skill docs. Ran the full flow and await-ui-element suites (1114 + 61 tests green at head) and a 13-mutation absence pass over the cross-tree probe logic (no unpinned mutation). Both previously-flagged anchoring bugs (echo as a second appender, delete-then-re-append defeating the length check) and the HIDDEN_UNREADABLE_NOTE note-matching issue are resolved at this head. The reviewed code returned no High or Medium findings; one Low noted below. Thanks for the thorough work.

packages/tool-server/src/tools/await-ui-element/index.ts
```diff
@@ -129,6 +182,81 @@ interface WaitResult {
...
+  condition: Params["condition"],
+  lastTrustedReadAt: number | undefined,
+  finalRead: FinalRead,
+  pollIntervalMs: number
+): UnmetUiWaitCause {
+  if (lastTrustedReadAt === undefined) return "unreadable";
+  if (finalRead === "trusted") return "unmet";
+  if (finalRead === "untrusted" && condition === "hidden") return "unreadable";
+  const darkTailMs = Date.now() - lastTrustedReadAt;
+  return darkTailMs > DARK_TAIL_TOLERANCE_INTERVALS * pollIntervalMs ? "unreadable" : "unmet";
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: **The dark-tail tolerance scales with the caller's `pollIntervalMs`, so a source that answers once and then hangs can still be narrated as a definite `unmet` verdict.**

Line 259 multiplies the fixed `DARK_TAIL_TOLERANCE_INTERVALS` (2) by the user-supplied `pollIntervalMs`, and the schema (lines 152-158) admits up to 5000ms - so the tolerance reaches 10000ms. Trace a wait where the tree source returns one usable read at t0 and then stops answering: `lastTrustedReadAt` stays ~t0, the final read attempt is `unsettled`, and with `timeoutMs` up to a few seconds `darkTailMs` (= deadline - t0) stays within 2\u00d7pollIntervalMs, so `timeoutCause` returns `unmet` - a definite "condition was false" verdict - for a window whose screen was not observed for the entire remaining wait.

Concretely: `pollIntervalMs: 5000` with `timeoutMs: 6000` classifies ~6s of unbroken silence as `unmet`. The flow-actions twin is not exposed to this: `CONDITION_DARK_TAIL_TOLERANCE_MS = POLL_INTERVAL_MS (300) * 2` (flow-actions.ts:142,248) is fixed at 600ms. The `unmet` cause is the one that licenses an author to re-record or delete the step (`UNMET_WAIT_WARNING` in flow-add-step), which is exactly the harm the classifier's own JSDoc (lines 71-73) says it exists to prevent: telling someone to rewrite a check on evidence nobody saw.

packages/skills/skills/argent-create-flow/references/live-authoring.md
```diff
@@ -130,11 +130,27 @@ For every retained raw gesture, add an echo and a recorded result check.
...
+- `unmet`: The tree was readable, but the condition was false. Restore the expected state or correct the selector or timeout. Then, record the check again.
+- `unreadable`: The wait ended without a trustworthy read. Restore the tree source and record the check again.
+- `cancelled`: The caller stopped the wait. Record the check again.
 
-The live tool and flow runner use different trees. A live wait can pass while its converted directive cannot resolve. Replay every conversion. Keep the raw tool only when its `pollIntervalMs` or `bundleId` is required. Live tools use `identifier`. Flow YAML uses `id`.
+Delete the old failed step after `flow-finish-recording`. Only `unmet` disproves the condition.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: The new blanket line “Delete the old failed step after `flow-finish-recording`” follows all three cause bullets, so it reads as applying to `unreadable` and `cancelled` as well as `unmet` — but the recorder's own warning for the unreadable case says the opposite: `UNREADABLE_WAIT_WARNING` in flow-add-step.ts ends “Do not delete the step on this warning alone”, and the previous revision of this skill carried exactly that scoping (“Do not delete it on this warning” under the no-trustworthy-read bullet). The two instructions now reach the same author in the same tool result: the step's `message` says keep, the skill says delete. In the permanent no-instrumentation case (`com.apple.*` system apps — the mode this same section of reliability-and-recovery.md describes, where the re-record path in the bullet above can never succeed), following the skill deletes the only copy of a gate that replays fine as written; on any `unreadable` wait it deletes a recorded step whose verdict is knowingly UNKNOWN. The qualifier “Only `unmet` disproves the condition” states which cause carries a verdict; it does not scope the deletion back to the `unmet` bullet.

packages/skills/skills/argent-create-flow/references/live-authoring.md
```diff
@@ -130,11 +130,27 @@ For every retained raw gesture, add an echo and a recorded result check.
...
+
+A stale `hidden` whose selector matches nothing replays as a silent pass — the unfalsifiable gate that [Record absence in three steps](#record-absence-in-three-steps) exists to prevent. Never proceed as though the gate passed. See the `await-ui-element` section of `argent-device-interact` for the full live condition and selector reference.
+
+A wait inside `run-sequence` gets no recorder warning. Inspect the nested result. Any `success: false` fails the sequence during replay.
+
+The live tool and flow runner use [different trees](flow-yaml.md#the-runner-tree-is-not-the-discovery-tree). After a successful wait, the recorder checks the same condition on the runner tree:
+
+- No warning: The condition holds on both trees.
+- Mismatch: Rule out a changed screen first. For `text`, also rule out multiple matches. Otherwise, use a runner-tree selector and replay.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: For a `text` divergence, this line reverses the one cause-ordering the probe's own message gives the author. The determinate warning in `probeAgainstRunnerTree` appends the text-tie clause — which explicitly reads “Check FIRST whether the selector matches more than one element” (flow-add-step.ts `textTieClause`) — ahead of the tree stories in the assembled message, with a code comment stating the clause is placed there deliberately. The previous revision of this bullet said the same order: “On a `text` check, first ask whether the selector matches more than one element… Next, rule out a changed screen…”. As rewritten, a reader on a text mismatch is told to rule the changed screen out before the multiple-match cause — the one cause with its own remedy (narrow the selector until it resolves a single node) and the one a longer `await:` timeout cannot fix. Non-`text` mismatches are unaffected; for `text` it is the reverse of what the same step's message orders.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: Fresh sweep at the current head. One docs-only commit (9787c9728, "chore: simplify bloated skills") sits between this sweep and the last reviewed head (9c52f3a). Ran the review moves on that delta: claims-vs-code for every new sentence in the rewritten live-authoring / flow-yaml / reliability sections against the code at this head, nearest-twin across the five skill files, and the absence pass (sibling that has it, prose promise, symmetry, mutation of the new prose). The two findings above were each confirmed by two independent verifiers against the exact code paths named in the comments. Everything else in the rewrite holds word-for-word against the head code and its tests. No repo files were touched.

packages/tool-server/src/tools/await-ui-element/index.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: Still present as of 9787c972814c3fb83ba67bfbaffdbe1029f4386b — the head moved only by the docs commit `9787c9728`; `await-ui-element/index.ts:259` still multiplies `DARK_TAIL_TOLERANCE_INTERVALS` (2) by the caller-supplied `pollIntervalMs` with the same schema (`max 5000`, index.ts:152-156), so the once-then-hung source still classifies as a determinate `unmet` verdict.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -3206,12 +3267,78 @@ function assertSessionStillLive(session: RecordingSession, step: FlowStep): void
...
+ * from parsing the file after the previous append, and `now` from parsing it
+ * after this one — so absent an edit the two parse byte-identical prefixes and
+ * a structural comparison is exact. That is what makes `JSON.stringify` safe
+ * here where it would not be against a raw in-memory step: no key-order or
+ * normalization difference can exist between two parses of the same bytes.
+ */
+function samePrefix(now: FlowStep[], before: FlowStep[], n: number): boolean {
+  if (now.length < n || before.length < n) return false;
+  for (let i = 0; i < n; i += 1) {
+    if (JSON.stringify(now[i]) !== JSON.stringify(before[i])) return false;
```
j-piasecki: **A cyclic `args` alias makes an already-written append report failure.**

`parseFlow` accepts cyclic YAML aliases in tool args, but this `JSON.stringify` throws. In a real-tool probe I recorded an unmet wait (populating `stepWarnings`), hand-edited its args to `&a { self: *a }`, and called `flow-add-echo`: the call returned `TypeError: Converting circular structure to JSON`, while rereading the YAML showed 2 steps—the echo had already been persisted by `appendStep` before `samePrefix` ran. Retrying the reported failure therefore duplicates the stored step; `flow-add-step` also invokes the live device action before reaching this path. The identical probe at merge base `4a113177` passed 1/1 and returned normally.

j-piasecki (commented):
Checked and cleared:

- Confirmed the head/base/diff and traced the successful-wait re-probe through the real per-platform flow adapters and shared selector evaluator.
- Verified timeout, hung-read, and cancellation handling through `pollDescribeTree`/`settleWithin` and the MCP long-running transport path.
- Traced `cause` through `await-ui-element`, `run-sequence`/`flow-run`, recorder warning persistence, host/client finish, edit anchoring, and concurrent append/finish paths.
- Ran the four affected suites (295 tests), ESLint, Prettier, and `git diff --check`. The test typecheck's TS2538 failure in `src/file-inputs.ts` reproduces unchanged at the merge base.

packages/tool-server/src/tools/await-ui-element/index.ts
hubgan: Resolved in e96f0b8b7.

packages/tool-server/src/tools/flows/flow-add-step.ts
hubgan: Resolved in 53484f26e.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
hubgan: Resolved in 5c218df66. 58d786564 caught the renumbering but not this probe: both survivors render identically, so the content check matched the shifted alignment too.

packages/tool-server/src/tools/await-ui-element/index.ts
hubgan: Resolved in 73b3ac87b.

packages/skills/skills/argent-create-flow/references/live-authoring.md
hubgan: Resolved in 2be9b057e.

packages/skills/skills/argent-create-flow/references/live-authoring.md
hubgan: Resolved in 2be9b057e.

packages/tool-server/src/tools/flows/flow-utils.ts
hubgan: Resolved in 325a3cff1.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -3356,12 +3417,154 @@ function assertSessionStillLive(session: RecordingSession, step: FlowStep): void
...
+ * inside a run are no more invisible than one. What stays out of reach is an
+ * edit that leaves the length alone: a reorder of two identical steps has no
+ * witness at all, in this or any content comparison.
+ */
+function anchorHolds(now: FlowStep[], before: FlowStep[], n: number): boolean {
+  if (!sameStepRun(now, before, n, 0, 0)) return false;
+  // A deletion ahead of the prefix slides `before` forward relative to `now`;
+  // an insertion slides `now` forward relative to `before`. Only the direction
+  // the length change allows is asked about.
+  for (let shift = 1; shift <= before.length - now.length; shift += 1) {
```
j-piasecki: **A mid-prefix delete of a diverging twin keeps its verdict, and the finish reports it against the twin that passed.**

A delete at index `p` with `0 < p < n` leaves `now[0..n-1] = before[0..p-1] ++ before[p+1..n]`. That is neither alignment this asks about: line 3499 accepts it whenever `before[p..n-1]` render alike, and the shift window here starts at `before[1]`, a different comparison. `dropMovedWarnings` states the opposite — "a delete or an insert at or before it… moves it or changes it, and dropping is the only answer that cannot convict an innocent step".

Throwaway probe through the real recorder on the existing fixture harness: step 2 diverges, step 3 is its byte-identical twin against a tree that holds the element, hand-delete step 2, record one more step.

```
message: Finished recording "probe1" flow (3 steps) — 1 step carries a cross-tree warning
         about converting a recorded wait; read `summary` before converting or replaying

summary:
 1. tool: await-ui-element {"condition":"visible","selector":{"text":"Ready marker"}}
 2. tool: await-ui-element {"condition":"visible","selector":{"text":"Continue"}}
    warning: recorded, but this condition does NOT hold against the tree the runner …
 3. tool: await-ui-element {"condition":"visible","selector":{"text":"Ready marker"}}
```

Step 2 is the twin that passed the probe, and `discarded` is 0, so `message` reports a clean carry rather than a shortfall. `drops the verdict a hand edit moved onto an identical twin step` (flow-record-cross-tree.test.ts:2147) deletes index 0, where the shift-1 window does match, so both verdicts drop there.

j-piasecki (commented):
Checked and cleared:

- The per-platform numbers the divergence stories rest on: chromium `describe` 5000 nodes (`describe/platforms/chromium.ts:53`) vs the flow tree's 12000 (`flow-chromium-tree.ts:94`); the android helper's 5000 vs `FLOW_MAX_NODES` 12000 (`flow-android-tree.ts:46`); `native-full-hierarchy`'s `maxDepth` default of 8 (`native-full-hierarchy.ts:48`) vs `queryFullHierarchyTree`'s 40 (`flow-ios-tree.ts:392`).
- `flow-add-step`'s new `longRunning: true` against the tools it proxies — `await-ui-element:441`, `run-sequence:155`, and `flow-run.ts:1152` (whose tool id is `flow-execute`, `flow-run.ts:1085`).
- "A wait inside `run-sequence` gets no recorder warning… Any `success: false` fails the sequence during replay": every warning is gated on `params.command === AWAIT_UI_ELEMENT_TOOL_ID`, and `run-sequence`'s break reaches the report through `nestedOrchestratorOutcome` (`flow-run.ts:2476`).
- "it adds about one second" in reliability-and-recovery.md: `waitForCondition` polls the full 1000ms grace plus `finalPoll` even when every read throws (`flow-actions.ts:1413`); the unreadable-tree tests land at 1006–1055 ms.
- `unmetUiWaitCause`'s note fallback: `timeoutNote` appends diagnostics after the base, so `startsWith(HIDDEN_UNREADABLE_NOTE)` matches, and the fetch-error arm returns the prefix unwrapped.
- `cappedReason` emits ≤ `MAX_PROBE_REASON_CHARS` and never longer than its input, across the fuzzed range.
- The four touched test files: 300 passed. Slowest test with no explicit timeout is 3013 ms against vitest's 5000 ms default (`vitest.config.ts` sets none). tool-server typecheck and `prettier --check` clean.

One comment, on the anchor check.

packages/tool-server/src/tools/flows/flow-utils.ts
hubgan: Fixed in 6a2b7cf27

j-piasecki (approved):
It looks like most of the changes in this PR are comments. Can they be reigned in at least to a some degree?

Please fix unit tests before merging.

hubgan:
> It looks like most of the changes in this PR are comments. Can they be reigned in at least to a some degree?
> 
> Please fix unit tests before merging.

I reduced the comments. The unit tests are main branch regression. @latekvo is working on the fix. I'll merge when that will be fixed.

latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Ran the full sweep against `dc304c45fc9f5fadb59ba7bb28914b56e50c3226`: claims vs code, nearest twin, non-happy paths, adversarial inputs, reachability both ways, what outlives the call, and the absence/mutation pass. I also re-ran the focused suites — `flow-record-cross-tree.test.ts` (74 passed), `flow-record-tap.test.ts` (12 passed), `flow-concurrent-recording.test.ts` (56 passed), plus a boundary-mocked cancellation/timing repro I wrote and deleted afterward. All findings from the sweep were either disproved by repro or already resolved in the current head. No HIGH, MEDIUM, or LOW issues remain. Thank you for contributing!

PR #721, opened 2026-08-04
fix(flow): raise the flow tree's depth cap and say why targeting failed

hubgan:
Rebased onto `main` (was stacked on #720), with a single hunk dropped — see
**Restack notes** at the bottom. Two commits: `a71c0f0` is exactly this PR's own
delta from the stacked version, `54fe761` is a review pass over it on the
post-rebase base. Squash on merge.

## Depth cap: 40 → 100

The flow selector tree capped raw `UIView` nesting at **40**. That does not clear the invisible wrapper layers React Native stacks on every screen — nested navigators (`RNSScreenStackView` → `RNSScreenView`, often two deep) plus a drawer/root wrapper routinely bury on-screen content **40–60 levels down** before the first tappable view.

In a deeply nested production app, plainly visible interactive elements sat at depths **41–62**. So `id:` and `text:` selectors silently failed to resolve, and only coordinate taps worked — which is the origin of coordinate-heavy flows on those apps.

The overflow is **silent**: nothing reports "truncated", selectors just stop matching. So the new cap carries generous headroom rather than hugging the deepest observed measurement.

**What it costs.** The tree itself is internal to selector resolution — consumed by `selectorToFrame`/`evaluateCondition`, never returned — so no tool result carries it. Text derived from it does reach the agent, though: `assertReason`'s `text` arm quotes the matched node's hoisted `subtreeText` verbatim into a failing `assert`/`await` reason, and nothing truncates it downstream. Descendants a depth-40 read dropped now hoist, so that quoted string grows with the cap. That is the trade — under the old cap those selectors did not resolve at all. Otherwise the cap only grows the `getFullHierarchy` payload over the native-devtools socket, which is already field-limited: ~11KB at depth 40, ~15KB at 48, and nothing at all past the tree's real depth.

## Targeting errors that could not be acted on

**`Provide bundleId explicitly`** — `resolveNativeTargetApp`'s own errors close with that advice. A flow selector step cannot follow it: the call hardcodes auto-targeting. Each failure now carries the remedy that actually exists:

[...]

PR #722, opened 2026-08-04
fix(flow): refuse a recorded tap that would replay somewhere else

hubgan:
> Stacked on #721.

## The core mismatch

Replay taps a selector's **centre**, not the point that was recorded. The recorder only checked that the resolved frame *contained* the point — which lets two silent misfires through.

## 1. Container-sized targets

A tap on blank space inside a drawer resolved to the drawer's whole scroll area — **0.72 of the screen**. It contains the point, so it recorded. Replay then tapped its centre, hit the "Chat" item, and reported **pass** while navigating somewhere the walkthrough never went.

A frame covering most of the screen is a container, not a control, and is now refused with that reason. Narrowing cannot help here: the problem is the element, not the selector.

**The threshold is a judgement between two real failures.** Too permissive records containers (above); too strict makes ordinary widgets unrecordable — a feed post is half the screen and tapping it is a perfectly normal QA step.

**Why not "does the centre resolve back to the same node"?** That test was tried and is wrong on a *flattened* tree: a control's own label is a sibling rect sitting on its centre, so a like button, a search field, a full-width row and every grid cell were refused — while replaying perfectly, because the touch is still inside the control. Node identity cannot tell a label from an independent control; size can tell a control from a container.

## 2. Ambiguity is answered by narrowing, not coordinates

When the derived selector matches the tapped element **and** something else and ranks the other first, the recorder now tries narrower forms — the node's own role, then its identifier — before giving up. The runner resolves either form, so answering ambiguity with coordinates was throwing away a perfectly good target.

**A `within` scope is deliberately not derived**, even though it would separate one feed row's button from another's. The flow tree is flattened and carries no z-order, so a container can only be found geometrically — and with a modal open, the foreground modal is a perfectly good geometric ancestor of a background element. A tap on a composer's text input recorded as a feed post "inside" the composer, then failed on any screen whose feed content differed. The scopes that survive are the ones an author writes knowingly at polish.

## 3. Positional ids are no longer derived

`profilePager-selector-2` encodes the element's index among its siblings, so it names a slot rather than a thing: it survives no re-order and silently addresses a different control once one is inserted before it.

It matters most where the recorder is least reliable. Observed on an edit-profile sheet: a tap on the display-name field derived the profile pager's "Media" tab — a view *behind* the modal. An oversized or ambiguous background match is caught by the two checks above; a positional id passes all of them and records silently.

## 4. Every kept-coordinate warning names the retarget

A coordinate tap replays fine today and breaks on the first layout change, so this warning is the author's only signal — and it is read at the one moment the screen is still there to retarget against.

The two failures need opposite advice: an element nothing can address needs a real target found in the tree; one that several things address equally needs **disambiguating**. Saying "no selector could be derived" for the second sent authors off to re-discover a selector they already had.

Raw-coordinate steps where capture never ran at all — `delayMs` set, `gesture-custom`, taps buried inside a `run-sequence` — now say why the rewrite was skipped and how to record it so that it happens.


packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -129,6 +130,23 @@ const UNSUPPORTED_PLATFORM = {
...
+ *
+ * `native-find-views` declares Apple capability only, so it is named for iOS
+ * alone; iOS `describe` is the AX tree — the RECORDER's side — so it is NOT
+ * listed here, where the point is to name the runner's reader.
+ */
+function treeReaderFor(udid: unknown): string {
+  const platform = platformOf(udid);
+  if (platform === "ios" || platform === "ios-remote") return "`native-find-views`";
+  if (platform === "chromium") return "`describe` (this platform's DOM walker)";
+  return "`describe`";
```
latekvo, line 147: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: On Android this fallback hands the author `\`describe\`` as the reader to find the real target with, but on Android `describe` returns the recorder's trimmed tree, not the projection the runner resolves selectors against at replay — this file's own Android clause in `runnerSideReadClause` states exactly that ("no read-only tool exposes the runner's full hierarchy on Android — `describe` returns the trimmed tree the recorder read"). The doc comment above claims this helper is "only ever called here for iOS / Chromium / Vega", but `coordinateRemedy` runs unconditionally for every kept-coordinate tap on every platform.

Reproduced against this head: recording a kept-coordinate tap with udid `emulator-5554` produces "Find the real target with `describe` and tap ITS centre". An Android author who follows it retargets against the one projection known to differ from what replay resolves, so the re-recorded tap can still fire somewhere else while reading as fixed.

latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Reviewed at 5834835. Ran the full sweep — claims vs code, nearest twin, non-happy paths, inputs, reachability both ways, what outlives the call, plus the absence pass (prose promises, symmetry across the delayMs opt-outs, sibling checks). Ran the four touched test suites locally (335 passing), traced the runner-side resolution chain end to end to confirm combined `{text, role}` selectors resolve identically at replay, and executed a reproduction against the head commit for each finding below before reporting it.

One non-inline note: `packages/skills/skills/argent-create-flow/references/live-authoring.md` describes the tap-recording outcomes as three, says a role-only `tap:` is "appended with no warning", and calls it "the recorder's silent fallback" — the recorder already warned on the role-only case before this PR, and this PR adds the container-refusal and ambiguity-narrowing outcomes, so that prose has drifted further out of sync with what recording actually reports.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -890,6 +1025,87 @@ export function directiveCommandHint(command: string): string | undefined {
...
+        step !== null &&
+        (step as { tool?: unknown }).tool === "gesture-tap"
+    )
+  ) {
+    return (
+      "run-sequence contains coordinate taps and was recorded as one opaque raw step; record taps " +
+      "individually so each can become a tap selector"
+    );
+  }
+  return undefined;
```
latekvo, line 1108: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: A `flow-execute` recorded with `delayMs` also skips its rewrite (`captureRunTarget` is gated on `delayMs === undefined`, like tap capture and the launch rewrite), lands in the raw-tool branch, and reaches this `return undefined` — so it is recorded as a raw non-portable `tool:` step with no warning at all, while the sibling opt-outs (`gesture-tap` + `delayMs`, `restart-app` + `delayMs`) both say why their rewrite was skipped and how to record so it happens.

Reproduced against this head: `flow-execute` with `delayMs: 500` writes `{kind: tool, name: flow-execute, args: {flow: …}, delayMs: 500}` and reports a bare "Step added" — the author gets no signal that the portable `run:` form was silently lost, which is the exact class of silence this warning table exists to close.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -890,6 +1025,87 @@ export function directiveCommandHint(command: string): string | undefined {
...
+    );
+  }
+  if (
+    command === "run-sequence" &&
+    Array.isArray(args.steps) &&
+    args.steps.some(
+      (step) =>
+        typeof step === "object" &&
+        step !== null &&
+        (step as { tool?: unknown }).tool === "gesture-tap"
```
latekvo, line 1100: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: This predicate only recognizes `gesture-tap`, but `run-sequence`'s own allowlist also admits `gesture-custom`, whose events are pure coordinates. A sequence built from `gesture-custom` steps therefore records as one opaque step with no warning, while the identical gesture recorded directly is told it was kept as raw coordinates and to re-record its taps individually — so a coordinate tap still hides from the recorder inside a run-sequence.

Reproduced against this head: a `run-sequence` wrapping a single `gesture-custom` Down/Up pair records with a bare "Step added" and neither the opaque-step nor the raw-coordinates warning.

packages/tool-server/src/utils/ui-tree-match.ts
```diff
@@ -1336,14 +1336,34 @@ const GENERIC_ROLES = new Set([
...
+ * It matters most where the recorder is least reliable. The flow tree is
+ * flattened and carries no z-order, so a tap inside a full-screen modal can
+ * resolve against a view BEHIND it; observed on Bluesky's edit-profile sheet,
+ * where a tap on the display-name field derived the profile pager's "Media"
+ * tab. An ambiguous or oversized background match is already caught and warned
+ * about, but a positional id on a background node passes every one of those
+ * checks and records silently. Refusing it turns that case back into the
+ * kept-coordinate warning the author is told to act on.
+ */
+const POSITIONAL_ID = /-selector-\d+$/i;
```
latekvo, line 1530: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: A deliberately chosen stable identifier that merely ends in `-selector-<n>` (say `bottomBar-selector-3`) is refused here just like a generated slot id, falling through to text/role or to null. When the id was the element's only anchor, the author then sees "tapped element has no stable text/id … If the element genuinely has no id or label, that is usually worth fixing in the app" — which is false in this case: the element has an id and the recorder discarded it by shape, with nothing in any warning saying so. The failure mode is a weaker-but-honest record rather than a misfire, so this is minor, but the refusal is invisible to exactly the author who did name things well.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -628,6 +646,92 @@ function roleOnlySelectorWarning(selector: Selector): string | undefined {
...
+ * button, a search field, a full-width row and every grid cell were refused —
+ * while replaying perfectly, because the touch is still inside the control.
+ * Node identity cannot tell a label from an independent control; size can tell
+ * a control from a container, which is the distinction that matters here.
+ */
+function replayReproducesTap(
+  frame: DescribeFrame,
+  point: { x: number; y: number }
+): "ok" | "container" | "retargets" {
+  if (isContainerSized(frame)) return "container";
```
latekvo, line 730: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Container size is checked before containment, so when the ranking elects an element that does not contain the tapped point and that element happens to be oversized, the verdict is `container` and the resulting warning asserts "the tap landed on <X>, which covers most of the screen" — false: the tap landed on the node under the point, and X is a different element ranked first somewhere else entirely. Coordinates are kept either way, so the cost is a wrong diagnosis that sends the author hunting for a container-sized control they never tapped instead of an ambiguity.

packages/tool-server/src/tools/flows/flow-add-step.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of 2eac0ef. Reproduced against this head with a `fetchFlowTree`-mocked recorder: a kept-coordinate `gesture-tap` on udid `emulator-5554` (Android) records the remedy `… Find the real target with `describe` and tap its centre …`, while the identical tap on an iOS udid names `` `native-find-views` ``. `coordinateRemedy` calls `treeReaderFor(udid)` unconditionally on every kept-coordinate tap, and its Android fall-through returns `` `describe` `` — the tree this file's own `runnerSideReadClause` Android clause describes as "the trimmed tree the recorder read, not the runner's". The `treeReaderFor` doc comment still asserts the helper "is only ever called here for iOS / Chromium / Vega".

packages/tool-server/src/tools/flows/flow-add-step.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of 2eac0ef. Reproduced against this head: recording a `flow-execute` of a sibling fragment with `delayMs: 500` writes `{ kind: tool, name: flow-execute, args: { name, project_root }, delayMs: 500 }` and reports the bare `Step added to "…" flow` — no warning. `runTarget` (the `run:` rewrite) is gated on `delayMs === undefined`, and `rawCoordinateWarning` has no `flow-execute` arm, so the portable `run:` form is dropped to a raw, non-portable `tool:` step silently — the one class of silence the sibling `delayMs` opt-outs (`gesture-tap`, `restart-app`) exist to close.

packages/tool-server/src/tools/flows/flow-add-step.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of 2eac0ef. Reproduced against this head: a `run-sequence` wrapping a single `gesture-custom` Down/Up pair records with a bare `Step added to "…" flow` and no warning. The `run-sequence` arm of `rawCoordinateWarning` matches only `step.tool === "gesture-tap"`, so a sequence built from `gesture-custom` (also admitted by run-sequence, and pure coordinates) records as one opaque step unflagged — while the same `gesture-custom` recorded directly is told it was kept as raw coordinates and to re-record its taps individually.

packages/tool-server/src/utils/ui-tree-match.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of 2eac0ef. `POSITIONAL_ID = /-selector-\d+$/i` is unchanged, so a deliberately-stable identifier that merely ends in `-selector-<n>` (e.g. `bottomBar-selector-3`) is refused by shape and falls through to text/role or to null. When that id was the node's only anchor, the author is then shown "tapped element has no stable text/id" — false for a node that does carry an id — with nothing in any warning saying the id was discarded because of its shape.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -892,6 +1027,101 @@ export function directiveCommandHint(command: string): string | undefined {
...
+    return (
+      "gesture-tap was kept as a raw coordinate tool step because flow-add-step delayMs prevents " +
+      "selector capture; remove delayMs, add a separate wait step before the tap if the pre-action " +
+      "delay is necessary, then record the tap again"
+    );
+  }
+  if (command === "restart-app" && delayMs !== undefined) {
+    return (
+      "restart-app was kept as a raw tool step because flow-add-step delayMs prevents the launch rewrite; " +
+      "remove delayMs so restart-app records as the leading launch, then record a post-launch " +
```
latekvo, line 1083: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: This remedy names `delayMs` as the reason `restart-app` was kept raw and tells the author to remove it so the step "records as the leading launch" — but `delayMs` is not the only thing that blocks the launch rewrite. `isLaunch` also requires the stripped args to be `bundleId` alone (`Object.keys(strippedArgs).length === 1`), so a `restart-app` carrying any extra arg — e.g. an Android `activity` (the case the comment at the `isLaunch` site explicitly calls out as staying raw) — is kept raw whether or not `delayMs` is set.

Reproduced against 2eac0ef: recording `restart-app { bundleId, activity }` with `delayMs: 500` emits this remedy; removing `delayMs` and re-recording `restart-app { bundleId, activity }` then records `{ kind: tool, name: restart-app, args: { bundleId, activity } }` and reports the bare `Step added to "…" flow` with no warning at all. So an author who follows the named fix lands on a step that is still raw — and now silently so, since no `rawCoordinateWarning` arm covers the extra-args reason.

packages/tool-server/test/utils/ui-tree-match.test.ts
```diff
@@ -1667,3 +1667,34 @@ describe("after / next (sibling) scoping", () => {
...
+
+  it("keeps no selector at all when the positional id was the only anchor", () => {
+    expect(deriveSelector(at({ identifier: "tab-selector-0" }))).toBeNull();
+  });
+
+  it("leaves ordinary ids alone, including ones merely containing digits", () => {
+    expect(deriveSelector(at({ identifier: "bottomBarProfileBtn" }))).toEqual({
+      identifier: "bottomBarProfileBtn",
+    });
+    expect(deriveSelector(at({ identifier: "selector-2-row" }))).toEqual({
```
latekvo, line 1695: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: This case does not exercise the trailing `$` it reads as if it guards. `selector-2-row` has no `-selector-` substring (no hyphen precedes the first `selector`), so it never enters the `POSITIONAL_ID` branch at all — it is kept for the same reason `post-2` is, not because of the anchor. As a result both load-bearing parts of `/-selector-\d+$/i` are unpinned: mutating it to drop the `$`, or to `\d*`, leaves the whole added suite green. The inputs that would actually discriminate are absent — an id like `myList-selector-2-label` (only the `$` keeps it) and one like `foo-selector-` (only the `\d+` keeps it).

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Re-reviewed at 2eac0ef. Ran the full sweep across the diff — claims vs code, nearest twin, non-happy paths, inputs, reachability both ways, what outlives the call, plus the absence pass (sibling coverage, prose promises, symmetry across the `delayMs`/raw opt-outs, and mutation of the new constants and regex). Rebuilt `@argent/registry` and ran the touched suites locally to reproduce each finding against this head before posting.

Two inline findings below are new since the last pass. The four earlier threads (the Android `describe` remedy, `flow-execute` + `delayMs`, `gesture-custom` inside `run-sequence`, and the positional-id refusal of a deliberately-stable id) I reproduced again against 2eac0ef — each is unchanged and still present, replied on its own thread rather than reopened here.

PR #723, opened 2026-08-04
fix(flow): wait for iOS instrumentation before capturing a recorded tap

hubgan:
> Stacked on #722.

## The problem

Replay has a launch readiness gate. **Recording had no counterpart.**

A live `restart-app` returns before the injected dylib dials back, so a tap recorded right after one read the tree before the app had connected, found nothing, and silently kept coordinates — with only the generic capture warning to show for it. That is the *first* tap of a walkthrough, and the one most likely to be repeated across takes.

## What changes

The recorder rides out that window before reading the tree, on the same budget replay uses: a cold start the replay gate would absorb, recording absorbs too.

- **With a leading `launch`**, it polls the exact bundle's connection bit — the same synchronous check replay makes.
- **For a fragment**, it polls until auto-targeting finds one connected, foreground-like app.

When the budget lapses it stops and lets the single tree read report whatever is really there, so the capture warning stays accurate rather than being replaced by a timeout.

## Three cases that had to stay cheap or correct

**An app started outside Argent can never connect during this recording.** One exhausted probe is remembered per device and per recording session — otherwise a 20-tap walkthrough pays the full budget twenty times. A successful tree read, or a successful `launch-app`/`restart-app`, clears the entry, so instrumentation that becomes available later still gets picked up.

**An Apple system app can never inject at all** (library validation), so it is not polled.

**The fragment path inspects app state over an RPC that can itself wedge.** It is raced against the remaining budget, so the gate stays a hard cap rather than becoming 8 seconds *between* potentially multi-second calls.

## Cancellation

A disconnect during the readiness poll now aborts the call. Previously it stopped the poll and then executed the tap anyway — moving the device after the caller had given up.

## Cost

When the app was never Argent-launched, this adds one budget's worth of latency before the (accurate) capture warning. That is the deliberate trade: it beats silently downgrading a post-launch tap to coordinates.


packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -710,12 +844,34 @@ function replayReproducesTap(
...
 ): Promise<{ selector?: Selector; warning?: string; ambiguous?: boolean; container?: boolean }> {
   try {
     const device = resolveDevice(udid);
+    // iOS's tree source connects asynchronously after launch — absorb the
+    // post-restart-app window replay's launch gate covers (see above). Apple
+    // system apps can never connect, and an exhausted probe is cached for this
+    // recording session so later taps do not each wait another full budget.
+    if (device.platform === "ios") {
+      const misses = readinessMissesFor(session);
+      const launchApp = recordedLaunchApp(session, device.platform);
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: The gate's readiness target is `recordedLaunchApp`, which walks `mostRecentLaunch` - a scan over steps recorded as kind `launch`, i.e. `restart-app`. A mid-flow `launch-app` is deliberately recorded as a raw tool step (not rewritten to a `launch`), so after "restart-app A ... launch-app B ... tap" this resolves to A even though the walkthrough switched to B. The gate then polls `isConnected(A)`, which is still true because launch-app never touched A, returns ready at once, and reads the tree before B's dylib has dialed back - silently downgrading the tap to a coordinate tap that survives only until the layout moves. Reproduced on the PR head: record a restart of A, a launch-app of B whose connection is still riding out, then a tap on B. The gate calls `isConnected` only with A (never B), and the recorded step is `{kind:"tap", x, y}` with "selector capture failed (B not connected yet); kept coordinates" - the exact post-restart-app failure this gate exists to prevent, still reachable through the launch-app sibling channel. The same mid-flow app-switch class is already handled in this PR for restart-app (the "gates on the most recent launch" case); the launch-app path is the sibling that does not get it.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -710,12 +844,34 @@ function replayReproducesTap(
...
+      if (!misses.has(device.id) && (!launchApp || isInjectableBundleId(launchApp))) {
+        const readiness = await awaitIosDevtoolsTarget(
+          registry,
+          device,
+          launchApp ?? undefined,
+          signal
+        );
+        if (readiness === "aborted") throw abortError();
```
latekvo, line 904: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: This new abort path reuses `abortError()`, whose message reads "flow-add-step aborted while re-probing the recorded wait against the runner's tree" - a phase unrelated to the readiness poll that actually aborted here (and the same stale text is carried by the post-capture abort at the execute call site). A caller that surfaces this AbortError's text on the new connection-wait path is told it was re-probing a recorded wait, which it was not.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: Two findings from a review pass over the new recorder readiness gate: one MEDIUM (reachable, with a hard reproduction on this PR's head) and one LOW.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -710,12 +844,34 @@ function replayReproducesTap(
...
   try {
     const device = resolveDevice(udid);
+    // iOS's tree source connects asynchronously after launch — absorb the
+    // post-restart-app window replay's launch gate covers (see above). Apple
+    // system apps can never connect, and an exhausted probe is cached for this
+    // recording session so later taps do not each wait another full budget.
+    if (device.platform === "ios") {
+      const misses = readinessMissesFor(session);
+      const launchApp = recordedLaunchApp(session, device.platform);
+      if (!misses.has(device.id) && (!launchApp || isInjectableBundleId(launchApp))) {
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: The gate is skipped outright when the most recent recorded launch is a non-injectable bundle: the `(!launchApp || isInjectableBundleId(launchApp))` guard means a recorded `restart-app com.apple.*` (e.g. Preferences) runs the tap's tree read with no readiness poll at all. That skip rests on the comment's premise that an Apple system app "can never connect" (:856) - but this repo's replay twin does not share it: `treeSourceGate`/`waitForNativeDevtools` (flow-run.ts:510-518) waits on every iOS bundle regardless of injectability and documents that on iOS 18.3 and 26.5 simulators Preferences/Maps/Contacts/Health/Photos/Safari all connect after the restart the launch step just ran. So a recorded restart-tap of an Apple system app still reads the tree before the app has connected and silently logs a coordinate tap - the exact post-restart downgrade this gate exists to ride out - while replay at least waits, and fails the launch loudly if the connection never comes up.

packages/tool-server/src/tools/flows/flow-add-step.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Still present as of 57090a31d880fcbbe62a06fb7a9d02b237709c45.

Reproduced on this head: record `restart-app A`, then a mid-flow `launch-app B` whose connection is still riding out, then a tap on B. The gate calls `isConnected` with `"com.example.a"` only — B is never queried — and the step is recorded as coordinate `tap: { x, y }` (`launch-app` stays a raw `tool:` step per the rewrite rule at flow-add-step.ts:2043-2044, so `recordedLaunchedApp` still resolves to A).

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -848,12 +972,34 @@ async function captureTapSelector(
...
+    // connection bit, so polling A would return ready at once and B's own
+    // connect window would never be ridden out — the exact downgrade this gate
+    // exists to prevent. Replay gates each launch step on its own bundle for the
+    // same reason.
     const launched = recordedLaunchedApp(session, device.platform);
+    // iOS's tree source connects asynchronously after launch — absorb the
+    // post-restart-app window replay's launch gate covers (see above). Apple
+    // system apps can never connect, and an exhausted probe is cached for this
+    // recording session so later taps do not each wait another full budget.
+    if (device.platform === "ios") {
```
latekvo, line 900: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: This gate keys on iOS only, but the replay twin it mirrors covers another platform with an attach window: `treeSourceGate` also runs `waitForVegaAutomation` after a Vega launch, because the automation toolkit attaches at app launch and serves no page source until it does. Recording has no counterpart for that window. A tap recorded right after a VVD `restart-app` reaches `queryVegaTree` while the toolkit has not attached; `describeVega` degrades to its empty tree, `nodeAtPoint` finds nothing under the point, and the step is kept as bare coordinates with "no element found under the tap" — the empty-tree relaunch hint is passed through for replay's blind-read guard but is dropped on this capture path, which reads only `{ tree, source }`. That is the same post-launch downgrade class this gate exists to ride out, on the sibling platform whose replay side does wait.

latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Ran the review moves over the readiness gate: claims vs code against the PR description, nearest-twin comparison with replay's `waitForNativeDevtools`/`treeSourceGate`, non-happy paths (wedged and rejecting app-state RPCs, unavailable service, abort mid-poll, miss-cache lifetime across session eviction, multi-device keying, restart-app/launch-app result shapes and device-id equality), reachability in both directions, and a mutation pass over the new branches and constants (the touched test suites catch every consequential mutation — all 137 tests pass on this head). Both prior threads' scenarios were replayed against 57090a31 with executable reproductions in a throwaway worktree: the mid-flow `launch-app` downgrade reproduces (replied on that thread); the Apple-system-app skip did not reproduce as described — the flow tree source statically refuses non-injectable bundles, so the outcome there is an explicit terminal explanation rather than a silent downgrade, and no amount of waiting changes it. One new finding inline.

packages/tool-server/src/tools/flows/flow-add-step.ts
latekvo: [\[Diplomat\](https://github.com/latekvo/Diplomat), X Preview F Free]: Still present as of 57090a31d880fcbbe62a06fb7a9d02b237709c45. Re-ran the scenario against this head: record restart-app A, then launch-app B whose instrumentation dials back only 300 ms later, then tap. The gate calls `isConnected` only with `com.example.a`, returns ready at once, and the step is recorded as a coordinate tap with "kept coordinates" while B's tree read fails.

packages/tool-server/src/tools/flows/flow-add-step.ts
latekvo: [\[Diplomat\](https://github.com/latekvo/Diplomat), X Preview F Free]: Still present as of 57090a31d880fcbbe62a06fb7a9d02b237709c45. The guard at flow-add-step.ts:994 still skips polling outright when the most recent recorded launch is a `com.apple.*` bundle, while the replay twin it cites deliberately keeps waiting on every iOS bundle (flow-run.ts:464-477, #453: Preferences connected:true on iOS 18.5).

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -115,6 +126,119 @@ function fallbackSourceWarning(source: DescribeSource, platform: string): string
...
+// successful tree read below or a successful app launch/restart clears the
+// entry, allowing recovery when instrumentation becomes available later.
+const captureReadinessMisses = new WeakMap<RecordingSession, Set<string>>();
+
+/**
+ * The cancellation a recorder step raises once its abort signal has fired.
+ *
+ * Thrown rather than returned: an abort during the readiness poll has to cancel
+ * the live action as well, not merely stop polling and dispatch the tap anyway.
+ * `name` is "AbortError" so the boundary reads it as a cancellation rather than
```
latekvo, line 158: [\[Diplomat\](https://github.com/latekvo/Diplomat), X Preview F Free]: The claim that the boundary "reads it as a cancellation rather than a tool failure" has no counterpart in the code: nothing between here and the caller special-cases `AbortError`, so the registry wraps the throw into `REGISTRY_TOOL_EXECUTION_FAILED` with the fallback signal `error_kind: "unknown"` (any `execute()` throw becomes a `ToolExecutionError`; only format-error.ts maps `AbortError`, and only for simulator-server calls). This same file states the opposite doctrine at CANCELLED_PROBE_WARNING (:625-627: "`AbortError` is special-cased nowhere ... every other cancellation in the server reports in band"). Consequence: a client disconnect during the up-to-15 s readiness poll - realistic for a longRunning tool - now reports a routine cancellation as an unclassified internal fault to the caller and in telemetry.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -115,6 +126,119 @@ function fallbackSourceWarning(source: DescribeSource, platform: string): string
...
+ * app. Stop when the budget lapses, then let the single tree read report
+ * whatever is really there. The budget mirrors replay's NATIVE_READY_TIMEOUT_MS:
+ * a cold start the replay gate would ride out, recording rides out too. When the
+ * app was never Argent-launched this adds one budget's worth of latency before
+ * the (accurate) capture warning; that beats silently downgrading a post-launch
+ * tap.
+ */
+type CaptureReadiness = "ready" | "unavailable" | "timed-out" | "aborted";
+
+// A third-party app started outside Argent can never connect during the active
```
latekvo, line 146: [\[Diplomat\](https://github.com/latekvo/Diplomat), X Preview F Free]: The premise is false: injection is `launchctl setenv DYLD_INSERT_LIBRARIES` on the simulator's launchd (ios-host.ts:235), which applies to every app process launched while it holds - however it is launched. A third-party app started from the home screen mid-recording picks up the dylib and dials back like any other, so it can connect during the active recording. The cache itself stays defensible (the successful-read clear below and the app-start invalidation recover), but this justification also contradicts the next sentence's own "allowing recovery when instrumentation becomes available later".

packages/tool-server/src/tools/flows/flow-add-step.ts
latekvo, line 900: [\[Diplomat\](https://github.com/latekvo/Diplomat), X Preview F Free]: The gate covers iOS only, but Vega has the same async-attach window and its downgrade is quieter. Replay gates a Vega launch on the toolkit serving a page source precisely because describeVega degrades to an empty tree + relaunch hint until attach (waitForVegaAutomation / treeSourceGate's vega arm), and a recorded `restart-app` is supported on Vega (restart-app/platforms/vega.ts returns `restarted: true`, so it records as kind "launch"). Here the tap's fetchFlowTree instead SUCCEEDS with that empty tree (queryVegaTree passes the empty tree + hint through unchanged), nodeAtPoint finds nothing, and the step is silently kept as coordinates. Reproduced on this head with a mocked empty vega read after restart-app: no readiness wait happens and the recorded step is `{kind:"tap", x, y}`.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -115,6 +126,119 @@ function fallbackSourceWarning(source: DescribeSource, platform: string): string
...
+    if (bundleId) {
+      try {
+        if (api.isConnected(bundleId)) return "ready";
+      } catch {
+        // Treat a transient connection read as not-ready and keep polling.
+      }
+    } else {
+      // Fragment auto-targeting must inspect app state, whose RPC can itself
+      // wedge. Race it against the remaining budget so the advertised
+      // 8-second gate stays a hard cap rather than 8 seconds between
```
latekvo, line 205: [\[Diplomat\](https://github.com/latekvo/Diplomat), X Preview F Free]: The budget raced against here is NATIVE_READY_TIMEOUT_MS = NATIVE_DEVTOOLS_CONNECT_BUDGET_MS = 15_000, and nothing in the repo advertises an 8-second gate - the only nearby "8s" is native-devtools.ts's note explaining why the budget had to exceed 8 s. The quoted figure does not exist for a reader to find.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -115,6 +126,119 @@ function fallbackSourceWarning(source: DescribeSource, platform: string): string
...
 }
 
+/**
+ * Recording has no counterpart to replay's launch readiness gate
+ * (`waitForNativeDevtools` in flow-run.ts): a live `restart-app` returns before
+ * the injected dylib dials back, so a tap recorded right after it would read
+ * the tree before the app has connected and silently keep coordinates. Ride
+ * out that window: when the recording has a launch, poll that (most recent)
+ * launch's exact bundle's synchronous connection bit (the same check replay
+ * uses); otherwise poll until auto-targeting finds one connected, foreground-like
```
latekvo, line 136: [\[Diplomat\](https://github.com/latekvo/Diplomat), X Preview F Free]: The wait re-probe is the other recorder-side read that fires right after a launch, and it gets no counterpart of this gate: record restart-app, then await-ui-element - the live wait evaluates the AX tree (available immediately) and can pass before the dylib connects, then probeAgainstRunnerTree reads fetchFlowTree on its 6-second PROBE_BUDGET_MS against this 15-second connect window. Every post-launch conversion check then lands on indeterminate UNKNOWN, or on the timeout arm's "the source is slow, not down" advice, which misreads a normal cold start. Never silent - each outcome is reported honestly - but systematically noisy across exactly the window this PR exists to absorb.

latekvo (commented):
[\[Diplomat\](https://github.com/latekvo/Diplomat), X Preview F Free]: Second review pass over the recorder readiness gate at head 57090a31d. Ran the full move set: claims vs code, nearest twin, non-happy paths, inputs, reachability both ways, and lifetime, plus the absence pass; re-verified the two earlier threads by reproducing them on this head (replies posted there). New findings below, each confirmed against source or reproduced in a scratch vitest run. Suite state on this head: flow-record-tap 40/40, flow-record-cross-tree 97/97, tsc --noEmit clean.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -115,6 +126,119 @@ function fallbackSourceWarning(source: DescribeSource, platform: string): string
...
+function invalidateReadinessMissAfterAppStart(
+  session: RecordingSession,
+  command: string,
+  args: Record<string, unknown>,
+  result: unknown
+): void {
+  const didStart =
+    typeof result === "object" &&
+    result !== null &&
+    ((command === "restart-app" && (result as { restarted?: unknown }).restarted === true) ||
```
latekvo, line 230: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: This allow-list clears the per-device readiness miss only for `restart-app` (`restarted:true`) and `launch-app` (`launched:true`), but `open-url` also starts an app: on iOS it cold-launches/foregrounds an injectable app via a custom scheme (returning `{ opened: true }`, no `launched`/`restarted`), whose injected dylib then dials back asynchronously — the exact connect window this gate exists to ride out. `open-url` is recorded as a raw `tool:` step, so `recordedLaunchedApp` never sees it either.

Effect: once a miss has been cached for the device (a prior fragment tap whose readiness poll timed out), an `open-url` that launches the app does not clear it, so the next tap finds `misses.has(device.id)` true, skips the readiness poll entirely (flow-add-step.ts:902), and `fetchFlowTree` runs during the connect window — throwing, so the tap is silently kept as coordinates. The identical sequence with `restart-app`/`launch-app` in that middle step captures a selector. This is a different code path from the launch-app-target thread: the trigger here is the command allow-list, not `recordedLaunchedApp`.

It also diverges from replay's foreground-changing set (`flow-run.ts` `FOREGROUND_CHANGING_TOOLS`, which includes `open-url`) and from this cache's own contract comment above ('a successful app launch/restart clears the entry, allowing recovery when instrumentation becomes available later'). Reproduced against 5ad14b9eb9 with a vitest unit test: tap1 (fragment, nothing connected) caches the miss and downgrades; `open-url` launches the app (dylib connects ~300ms later); tap2's connection-bit poll count is unchanged (gate skipped) and the step is kept as coordinates.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -115,6 +126,119 @@ function fallbackSourceWarning(source: DescribeSource, platform: string): string
...
+        if (api.isConnected(bundleId)) return "ready";
+      } catch {
+        // Treat a transient connection read as not-ready and keep polling.
+      }
+    } else {
+      // Fragment auto-targeting must inspect app state, whose RPC can itself
+      // wedge. Race it against the remaining budget so the advertised
+      // 8-second gate stays a hard cap rather than 8 seconds between
+      // potentially multi-second getAppState calls.
+      const inspected = await settleWithin(inspectConnectedNativeApps(api), remaining, signal);
```
latekvo, line 207: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: On the fragment (auto-target) path, `inspectConnectedNativeApps` fans out `Promise.all(bundleIds.map(getAppState))` — all-or-nothing. If any one connected app's `getAppState` rejects (`Application.getState` has a 5s RPC timeout in native-devtools.ts, and iOS suspends a backgrounded app so its `getState` stops answering), the whole inspect rejects, discarding the state of a responsive frontmost app. `settleWithin` maps that rejection to `{ type: "error" }`, which none of the arms here (aborted/timeout/value) match, so the loop falls through to the sleep and re-issues the same inspect every poll until the 15s deadline lapses — returning `timed-out` and caching a readiness miss even though a foreground-active app was connected the whole time.

Effect: the first fragment-path tap taken after any second app is connected (an app switch/deep link, or an app a prior session left connected) waits the full `NATIVE_READY_TIMEOUT_MS` and caches a spurious miss on a healthy device, which then suppresses the gate for the following tap. The sibling twin in this same package, `detectFrontmostBundleId` (native-devtools.ts), wraps each `getAppState` in a per-app try/catch precisely so one unresponsive app can't sink frontmost detection; `inspectConnectedNativeApps` (reached here and by auto-target) has no such guard, and the `settleWithin` race the comment above describes caps a single hang but not a repeated reject.

Reproduced against 5ad14b9eb9 with a vitest unit test (`listConnectedBundleIds` -> [good, wedged]; `getAppState(good)` resolves foreground-active; `getAppState(wedged)` rejects): the gate re-polls the full budget instead of concluding on the first poll where the good app is visible.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Review pass over the recorder readiness gate at head 5ad14b9eb9. Ran the move set — claims vs code, nearest twin (against replay's `treeSourceGate`/`waitForNativeDevtools` and the blueprint's own `detectFrontmostBundleId`), non-happy paths, inputs, reachability both ways, lifetime — plus a mutation pass over the new branches and the per-session miss cache; the touched suites are 137/137 on this head. Two new low-severity findings below, each reproduced with a scratch vitest unit test against this head; both sit on code paths distinct from the earlier threads, so they are filed rather than replied there.

PR #724, opened 2026-08-04
fix(flow): stop recording and scoring checks that cannot fail

hubgan:
> Stacked on #723.

Three kinds of step were being written into flows — and passing on every replay — without being able to prove anything. A silently-green check is worse than a flake: the test has stopped testing and nothing downstream will ever flag it.

## 1. A `hidden` wait whose selector never matched

That is not proof of dismissal. It is a check that **cannot fail**: a typo'd selector, a renamed id, and the wrong screen entirely all satisfy it.

**Why "never matched" is too narrow on its own.** The action that removes an element runs *before* the check, so the correct authoring order — prove `visible` → act → prove gone — always reaches the wait with nothing matching. Judging on the wait's own poll window would reject exactly the discipline we ask authors for.

So the **flow** is the wider evidence. `flow-selector-evidence` collects what a flow has positively established (acted on, or proved present), and a `hidden` check is falsifiable when an earlier step established its selector. A prior `hidden` proves absence, never presence, so it never counts. Only steps that **passed** count — a failed `visible` check proves nothing was there.

**The recorder and the runner answer differently, on purpose:**

| | verdict | why |
| --- | --- | --- |
| recorder | **refuses** to write it | the screen is still in front of the author; recording `visible` first costs one call |
| runner | **passes**, with a warning | the condition genuinely held, and three legitimate shapes reach it |

Those three shapes: a scoped `hidden` whose match sits outside the container, a fragment whose `executionPrerequisite` established the element, and a baseline absence check. Failing the run would report a regression that did not happen. What the two sides agree on is the **evidence rule**, not the verdict — so a hand-edited flow keeps its permanently-green gate and is told so on every run.

The scoped case is also detected directly: if the selector matches once its scope is dropped, the check is doing real work and is not flagged.

**The gate reads nested selectors too.** Wrapping the identical wait in a one-step `run-sequence` used to buy a permanently-green check that the recorder refuses to write directly.

## 2. An `await-ui-element` whose condition was not met

It returns `{ success: false }` rather than throwing — the same shape the runner and `run-sequence` already special-case. The recorder did not, so an unmet wait was recorded as a passing step, baking a gate that fails every replay.

## 3. `await-screen-idle`

It reports a screen that never settled as a soft `settled: false` instead of failing. A recorded one is therefore green on every replay no matter what the screen does. It is a live diagnostic, not a gate — the refusal names the two things that *are* gates: the element you actually need next, or `await: { idle: true }` at polish (added in a later PR in this stack), which hard-fails.


latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Ran the review moves across the stack: claims vs code (every docstring/tool-description/docs claim added here checked against the implementation), nearest-twin and symmetry across all four sites that implement the evidence rule, non-happy paths and hostile inputs, reachability both ways, lifetime/state, plus an absence pass and mutation checks on the new tests. Verified with executed evidence in a scratch checkout at head 72963348b9374e9eeb615f5a9f4d770866df254b: the touched suites pass 306/306, `tsc --noEmit` is clean, and every new gate was mutation-tested (reverting each production change makes its test fail), so the tests discriminate. Two findings below are backed by executed reproductions; the rest were confirmed by tracing the reachable paths. One note I could not anchor inline: packages/tool-server/test/flows/flow-record-cross-tree.test.ts:1889 still carries the comment "Delete the diverging step 1, the remedy `UNMET_WAIT_WARNING` offers." — both the constant and the delete-after-finish remedy were removed by this PR.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -2211,6 +2231,13 @@ async function execWhenStep(
   // Marker for the block, then the guarded steps inline — same fragment
   // attribution, one level deeper, failures hard-stop as anywhere else.
   pushReport(state, { ...marker, status: "pass", reason: `condition met (${label})` });
+  // The guard HELD, so its (non-hidden) selector was present — evidence for a
+  // later `hidden` check, exactly as an inline `assert: { visible }` would be.
+  if (step.condition.kind === "ui" && step.condition.condition !== "hidden") {
```
latekvo, line 2248: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: The `when:` path drops the vacuous-hidden verdict entirely. `waitForCondition` now returns `{ ok: true, warning }` for an unfalsifiable `hidden`, and the `await:`/`assert:` leaf arm carries that warning into the step report — but `execWhenStep` reads only `probe.aborted` / `.ok` / `.indeterminate`, so `probe.warning` never reaches any report. Reproduced end-to-end against the mocked tree: a hand-written flow `- when: { hidden: "zzz-never-rendered" }` runs green with zero warnings anywhere in the result, while the identical check spelled `- assert: { hidden: ... }` warns "cannot fail and proves nothing". A vacuously-true guard silently admits its block — one spelling away from exactly the permanently-green gate this PR refuses to let the recorder write.

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
```diff
@@ -0,0 +1,122 @@
...
+ * its guard held; treating a conditional proof as no proof would reject the
+ * correct authoring order for any flow that dismisses an interstitial.
+ */
+export function selectorEstablishedInSteps(steps: FlowStep[], selector: unknown): boolean {
+  const wanted = selectorIdentityTerms(selector);
+  if (wanted.length === 0) return false;
+  const seen = new Set<string>();
+  const walk = (list: FlowStep[]): void => {
+    for (const step of list) {
+      for (const term of establishedTerms(step)) seen.add(term);
```
latekvo, line 108: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: This walk credits steps inside a `when:` block regardless of whether the guard will ever hold, while the runner credits a block's steps only after the guard held (`execWhenStep` returns before executing the block when the probe fails). So for `- when: { visible: Ghost }` containing `await: { visible: Sheet }` where Ghost never renders, this function reports Sheet established (reproduced at unit level: it returns `true`), the recorder records a later `hidden Sheet` clean, and replay then attaches the "proves nothing" warning to that step on every run because the skipped block never established anything. Recorder and runner disagree on the evidence rule for this input — the agreement this file's header says they must keep.

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
```diff
@@ -0,0 +1,122 @@
...
+ * The wait tool itself cannot tell the difference, because it only sees its own
+ * poll window and the action that removes an element runs BEFORE the check —
+ * so the correct authoring order (prove visible → act → prove gone) always
+ * reaches it with nothing matching. The flow is the wider evidence.
+ *
+ * Shared by the recorder and the runner, which answer it differently on
+ * purpose. The recorder REFUSES to write such a check: the screen is still in
+ * front of the author, so the fix (record `visible` first) costs one call. The
+ * runner still scores it a PASS, carrying `warning` — by then the condition
+ * genuinely held, and three legitimate shapes reach it (a scoped `hidden`
```
latekvo, line 19: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Two of the three listed shapes do not hold. (1) A scoped `hidden` whose match sits outside the container cannot reach this warning from the directive path this outcome belongs to: `hiddenCheckIsFalsifiable` reprieves exactly that shape (scope anchors resolve visibly + visible unscoped match → falsifiable → clean pass, no warning). (2) An `executionPrerequisite` establishes nothing: it is acknowledge-gated free text that is echoed into the summary, and nothing executes it into `establishedSelectors` — the set's only writers are passing steps and entered non-hidden `when:` guards. The same wording is duplicated in the `DirectiveOutcome.warning` docstring in flow-actions.ts.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -1983,27 +1982,109 @@ If a step was recorded by mistake, remove it from the .yaml after \`flow-finish-
...
-      // them separately.
-      let waitWarning: { warning: string; kind: "conversion" | "wait" } | undefined;
+      // WHY it came back false picks the wording — see {@link unmetWaitRefusal}.
+      // A signal that is already down counts as cancelled whatever the result
+      // says, since the tool can return before the abort reaches its poll loop.
+      if (isUnmetUiWaitResult(params.command, toolResult)) {
+        const { stepCount, note } = await activeFlowState(session, true);
+        const waitNote = (toolResult as { note?: unknown }).note;
+        const detail = typeof waitNote === "string" ? `: ${waitNote}` : "";
+        const cause = ctx?.signal?.aborted === true ? "cancelled" : unmetUiWaitCause(toolResult);
```
latekvo, line 1734: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: When the caller's abort lands while this refusal is being assembled, a genuinely judged-false wait gets relabelled as cancelled. `invokeSubTool` forwards the signal into the sub-tool, so `await-ui-element` can conclude a real unmet verdict on trusted reads and the abort can arrive afterwards — and this branch awaits `activeFlowState` (a disk read + parse in host mode) before sampling the signal, widening the window past the tool's return. The author is then told the condition was "never settled ... UNKNOWN, not known-bad" and sent to re-run blindly, instead of being told the condition was judged false. flow-run's own abort handling explicitly rejects this inference pattern elsewhere ("Deliberately narrow, not 'the signal is set, so this step is a skip'"). Reproduced with a mock returning `{ success: false, cause: "unmet" }` with the signal already aborted: the message comes back as the cancelled/UNKNOWN wording.

packages/skills/skills/argent-create-flow/references/live-authoring.md
```diff
@@ -136,15 +136,15 @@ For every retained raw gesture, add an echo and a recorded result check.
...
+- `unreadable` — the wait ended with no trustworthy read, so the condition was never judged. Restore the tree source, then record the check again. Do not change the selector on this alone.
+- `cancelled` — the caller stopped the wait. Record the check again.
 
-A stale `hidden` whose selector matches nothing replays as a silent pass — the unfalsifiable gate that [Record absence in three steps](#record-absence-in-three-steps) exists to prevent. Never proceed as though the gate passed. See the `await-ui-element` section of `argent-device-interact` for the full live condition and selector reference.
+Nothing has to be deleted, because nothing was written. A stale `hidden` whose selector matches nothing would replay as a silent pass — the unfalsifiable gate that [Record absence in three steps](#record-absence-in-three-steps) exists to prevent, and which the recorder refuses to write for the same reason. Never proceed as though a refused gate passed. See the `await-ui-element` section of `argent-device-interact` for the full live condition and selector reference.
```
latekvo, line 146: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Two lines below this paragraph, the unchanged sentence "A wait inside `run-sequence` gets no recorder warning." is now false: this PR adds exactly such a recorder response — a sequence-wrapped vacuous `hidden` is refused outright with a named message plus a partial-mutation warning (pinned by the seq-vac test in flow-tools.test.ts). An agent authoring from this page will trust the stale sentence over behavior the rest of this section now has.

packages/argent-mcp/src/content.ts
```diff
@@ -211,13 +211,15 @@ export type FlowStepResult = {
...
   reason?: string;
   /**
-   * A step that passed in a way that weakens it as proof — raised today by
-   * `await: { idle: true }`, which never fails a run and says here what its
-   * green actually bought (see StepReport.warning in the tool-server's
-   * flow-run). Also carries the caveat older tool-servers put on a snapshot
-   * that adopted a missing baseline, which now fails the step instead. Live
-   * either way: dropping the field would silently delete the only thing the
-   * readiness check reports.
+   * A step that PASSED but whose pass is weaker than it looks. Two sources
```
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: "Two sources today" undercounts: `StepReport.warning` also carries the unsettled selector-less gesture warning raised during a tree-source outage (`GestureSettle.warning` in flow-actions.ts, spread onto directive outcomes and step reports in flow-run). The sibling docstring this text points at (StepReport.warning in flow-run.ts) lists all three producers; this rewrite re-asserts a count the same PR made false.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
```diff
@@ -204,7 +198,7 @@ export const flowFinishRecordingTool: ToolDefinition<
...
+A warning flow-add-step raised on a recorded \`await-ui-element\` is repeated in \`summary\` as a \`warning:\` line of its own, right below the step it judges, and \`message\` counts them by kind. A warning is repeated only while the step it judges is still identifiable by its number: hand-editing the .yaml during the recording moves the steps, so those warnings are DROPPED rather than pinned on whichever step inherited the number, and \`message\` says how many were dropped. A step that carries a cross-tree warning was re-probed against the runner's tree: read it before converting that wait to \`await:\`/\`assert:\`, which is what the verdict is about and what this moment is for. A wait that did not pass is never among them — \`flow-add-step\` refuses to record one at all, naming the cause on the call itself — so every warning here is a question about converting, not about replaying.
```
latekvo, line 200: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: "counts them by kind" is made stale by this same PR: `RecordedStepWarning` loses its `kind` discriminator (`conversion` | `wait`) and `warningHeadline` now counts the single conversion category, so there is no plural of kinds left to count by.

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
```diff
@@ -0,0 +1,122 @@
...
+        // `visible` check inside the block would carry.
+        if (step.condition.kind === "ui" && step.condition.condition !== "hidden") {
+          for (const term of selectorIdentityTerms(step.condition.selector)) seen.add(term);
+        }
+        walk(step.steps);
+      }
+    }
+  };
+  walk(steps);
+  return wanted.some((term) => seen.has(term));
```
latekvo, line 121: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Any-one-term overlap licenses the whole selector, but a check only fails when EVERY field matches. A flow that established `{ text: "Confirm" }` licenses a later `hidden { identifier: "confrimB", text: "Confirm" }` (typo'd id): `text:confirm` satisfies this `.some()` — mirrored in flow-run.ts:2458, flow-actions.ts:1236 and await-ui-element/index.ts:129 — so the recorder records it clean and the runner scores it clean, yet the full selector can never match anything on any screen because the identifier ANDs. Reproduced end-to-end through the real runner on a mocked tree that always shows Confirm: `assert: { visible: "Confirm" }` followed by `assert: { hidden: { identifier: "confrimB", text: "Confirm" } }` replays pass with NO warning, while the identical check without the overlapping text field warns. This is exactly the typo-class permanently-green gate this PR exists to block, admitted through the gate whenever any single term of the selector collides with an earlier step.

latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Ran the full sweep at head 9689cf1e: claims vs code (every docstring, tool description and skill-doc statement introduced here checked against the implementation), nearest-twin and symmetry across all five sites implementing the evidence rule (shared model, recorder refusals, directive-path gate, raw-tool gate, nested-sequence exoneration), non-happy paths and hostile inputs (degenerate selectors, scope interactions, abort/unmet/unreadable shapes, sequence nesting and index alignment), reachability both ways plus blast radius outside the diff (consumers of RecordedStepWarning/DirectiveOutcome/StepReport shapes in argent-mcp, argent-cli, ui, skills), what outlives the call (ExecState/session lifetime, reset points, cross-recording leak checks), and the absence pass (sibling mechanisms missing the discipline, stale prose repo-wide, mutation-read of the new tests). Verified in a scratch checkout at head 9689cf1e: the flow suite is green (1434 tests; three load-induced failures on a busy machine pass on a quiet re-run), `tsc --noEmit -p tsconfig.test.json` is clean, and four findings below are backed by executed end-to-end reproductions through the real runner on mocked trees rather than code reading alone.

One additional item I could not anchor inline because the file is unchanged by this PR — packages/tool-server/src/tools/flows/flow-start-recording.ts:93-97 still tells agents an unmet `await-ui-element` wait "is still recorded (it returns success:false rather than failing)", that flow-add-step warns only when you record the wait DIRECTLY, and that a wait nested in a recorded run-sequence "gets neither warning". All three are retired by this PR: unmet waits are refused outright and nested vacuous waits refuse the whole call. Agents authoring from that description will expect steps to exist that the recorder now declines to write.

Replies are left on the earlier review's threads where the finding still reproduces at this head; the live-authoring.md item from that sweep has since been rewritten and no longer applies.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -1752,26 +1763,109 @@ If a step was recorded by mistake, remove it from the .yaml after \`flow-finish-
...
+        const { stepCount, note } = await activeFlowState(session, true);
+        const settled = (toolResult as { settled?: unknown }).settled;
+        return {
+          message:
+            "`await-screen-idle` is a live diagnostic, not a gate — step NOT recorded. It " +
+            "reports a screen that never settled as `settled: false` instead of failing, so a " +
+            "recorded one passes on every replay no matter what the screen does" +
+            (settled === false ? " — and it just reported `settled: false`" : "") +
+            ". Record readiness as the element you actually need next (`await-ui-element`), or " +
+            "add `await: { idle: true }` during polish, which FAILS when the screen never " +
```
latekvo, line 1758: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: This tells the author `await: { idle: true }` "FAILS when the screen never settles" — the opposite of what the directive does. waitForIdle's own contract (flow-actions.ts:1661-1664) and the flow-execute tool description both say an unsettled screen spends the timeout and passes carrying a warning; the only non-passing outcome is an unreadable tree. Reproduced end-to-end on a mocked tree that never holds still: `await: { idle: true, timeout: 800, stableFor: 200 }` returns status pass, run ok:true, with a warning. An agent following this refusal believes they hardened the flow, but has re-created the same green-with-a-warning gate this PR refuses to record `await-screen-idle` for — recommended by name, under inverted semantics.

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
```diff
@@ -0,0 +1,122 @@
...
+        if (!Array.isArray(nested)) return [];
+        return nested.flatMap((s) => {
+          const inner = s as { tool?: unknown; args?: { condition?: unknown; selector?: unknown } };
+          if (inner.tool !== AWAIT_UI_ELEMENT_TOOL_ID) return [];
+          return inner.args?.condition === "hidden"
+            ? []
+            : selectorIdentityTerms(inner.args?.selector);
+        });
+      }
+      if (step.name !== AWAIT_UI_ELEMENT_TOOL_ID) return [];
```
latekvo, line 87: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Evidence crosses a `run-sequence` (lines 76-86) and crosses `run:` fragments (inline expansion shares the run's ExecState), but not `tool: flow-execute` — this returns nothing for it, and the sub-run seeds its own empty set (flow-run.ts:1265). Same composition, different verdict per spelling. Reproduced end-to-end with an identical parent (awaits Sheet visible, element then gone) and fragment (`assert: { hidden: "Sheet" }`): under `run:` the gate observes the parent's establishment and passes clean; under `tool: flow-execute` it observes an empty set and attaches the "cannot fail and proves nothing" warning to the very same check, permanently. The reverse direction mislabels too: a fragment that positively establishes a selector leaves no trace, so the parent's later `hidden` on it is refused at recording time / warned at replay under flow-execute but accepted under `run:`. Given the module header's stated invariant that recorder and runner agree on the evidence rule, today the rule itself forks on how a composed flow is spelled.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -2423,6 +2448,19 @@ async function execLeafStep(
...
           };
         }
+        // Covers a wait nested in a `run-sequence` as well as a direct one —
+        // the recorder refuses both, so scoring only the direct one would let a
+        // wrapped check replay as a clean pass forever.
+        let vacuousWarning: string | undefined;
+        for (const selector of vacuousHiddenSelectors(step.name, result, step.args)) {
+          const terms = selectorIdentityTerms(selector);
+          const falsifiable =
+            terms.length === 0 || terms.some((t) => state.establishedSelectors.has(t));
```
latekvo, line 2633: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: This replay-side gate covers `hidden` only, so a raw `tool: await-screen-idle` step — hand-written YAML or a recording made before this gate existed — keeps replaying as a silent pass forever. It is exactly the class the recorder now refuses ("passes on every replay no matter what the screen does"), but with no ⚠ analogue to the vacuous-hidden one telling the author so on every run, which is the treatment this PR gives hand-edited unfalsifiable `hidden` gates. Reproduced end-to-end: a flow whose step is `tool: await-screen-idle` with the tool answering `{ settled: false, waitedMs: 100, polls: 1 }` finishes ok:true, the step passing with no warning anywhere in the report. (The run-sequence route is closed — its allowlist rejects the tool name outright — so only the raw-tool spelling stays open.)

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * The identifier/text a selector names, lowercased. `role` is deliberately
+ * excluded — "some button existed earlier" is not evidence about THIS element.
+ */
+export function selectorIdentityTerms(selector: unknown): string[] {
+  if (selector === null || typeof selector !== "object") return [];
+  const s = selector as { identifier?: unknown; text?: unknown };
+  const terms: string[] = [];
+  if (typeof s.identifier === "string" && s.identifier !== "") {
+    terms.push(`id:${s.identifier.toLowerCase()}`);
```
latekvo, line 36: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: The identity term keeps the identifier verbatim (lowercased), but the matcher folds Android resource-ids — `'submit'` and `'com.example.app:id/submit'` resolve to the same element (selectorSchema's own description promises this). A flow that establishes presence via the qualified spelling therefore leaves no term a short-spelled `hidden { identifier: submit }` can find, yielding a false refusal at recording time and a false "cannot fail" warning at replay for the very element the flow did prove.

packages/tool-server/src/tools/flows/flow-run.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Still present as of 9689cf1e2de3f69b25749d90b17a8e01a6a27a66 — `execWhenStep` still reads only `probe.aborted`/`.ok`/`.indeterminate` (flow-run.ts:2039-2056), so a vacuously-held `when: { hidden: ... }` guard still executes its block green with no warning anywhere in the report.

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Still present as of 9689cf1e2de3f69b25749d90b17a8e01a6a27a66 — the walk at flow-selector-evidence.ts:106-117 still credits `when:` guards and inner steps unconditionally, and the unit test this PR adds (flow-selector-evidence.test.ts:72-81) pins exactly that unconditional credit.

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Still present as of 9689cf1e2de3f69b25749d90b17a8e01a6a27a66 — the three-shapes wording remains at flow-selector-evidence.ts:19-21 and flow-actions.ts:144-148; the scope reprieve (flow-actions.ts:1249-1257) still exonerates the scoped shape, and nothing anywhere executes `executionPrerequisite` into `establishedSelectors`.

packages/tool-server/src/tools/flows/flow-add-step.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Still present as of 9689cf1e2de3f69b25749d90b17a8e01a6a27a66 — flow-add-step.ts:1779 still overrides the result's own cause with "cancelled" whenever the signal is already down, after awaiting `activeFlowState` first.

packages/argent-mcp/src/content.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Still present as of 9689cf1e2de3f69b25749d90b17a8e01a6a27a66 — content.ts:196-199 still counts "two sources today" for `warning`, while `StepReport.warning`'s own docstring (flow-run.ts:194-205) names three.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Still present as of 9689cf1e2de3f69b25749d90b17a8e01a6a27a66 — flow-finish-recording.ts:194 still says `message` counts the warnings "by kind", while `RecordedStepWarning` now carries no `kind` and `warningHeadline` counts a single conversion category (flow-finish-recording.ts:107-118).

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
```diff
@@ -0,0 +1,122 @@
...
+          return inner.args?.condition === "hidden"
+            ? []
+            : selectorIdentityTerms(inner.args?.selector);
+        });
+      }
+      if (step.name !== AWAIT_UI_ELEMENT_TOOL_ID) return [];
+      const args = step.args as { condition?: unknown; selector?: unknown };
+      return args.condition === "hidden" ? [] : selectorIdentityTerms(args.selector);
+    }
+    default:
```
latekvo, line 91: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: A selector-anchored `swipe` proves its anchor element present, but this evidence rule credits it nothing. `establishedTerms` has cases for tap/long-press/type/scroll-to/await/assert/pinch/rotate/snapshot/tool, but no `swipe`, so a swipe step falls through to this `default: return []`. A swipe whose `from`/`to` is a selector resolves both ends through `waitForFrames` and hard-fails when an end is absent (flow-actions.ts:1333-1335), the very gate the pinch/rotate case just above (lines 58-63) credits, with the comment "a passing one proves the selector present." Because this is the single rule read by both the runner (`state.establishedSelectors`, flow-run.ts:2113) and the recorder (`selectorEstablishedInFlow`, flow-add-step.ts:1788), a flow that dismisses an element with `swipe: { from: { text: "Card" }, direction: up }` and then checks `assert: { hidden: "Card" }` is refused at recording time and carries the "cannot fail and proves nothing" warning on every replay, even though it genuinely proved Card present and then gone. Reproduced end-to-end on a mocked tree: the swipe-then-hidden flow warns while an identical tap-then-hidden flow does not (same tree, same hidden check); and a swipe on an absent selector hard-fails, confirming a passing swipe is real evidence.

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
```diff
@@ -0,0 +1,122 @@
...
+export function selectorIdentityTerms(selector: unknown): string[] {
+  if (selector === null || typeof selector !== "object") return [];
+  const s = selector as { identifier?: unknown; text?: unknown };
+  const terms: string[] = [];
+  if (typeof s.identifier === "string" && s.identifier !== "") {
+    terms.push(`id:${s.identifier.toLowerCase()}`);
+  }
+  // A regex text locator (`{ text: { matches } }`) names no fixed string, so it
+  // cannot serve as, or be satisfied by, positive evidence.
+  if (typeof s.text === "string" && s.text !== "") terms.push(`text:${s.text.toLowerCase()}`);
```
latekvo, line 40: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: The text term is built with a bare `.toLowerCase()`, but the tree matcher this rule has to agree with folds text far more aggressively: Unicode NFC normalization, NBSP and other space-likes to a plain space, whitespace-run collapse, and zero-width/invisible stripping (ui-tree-match.ts `foldText`). So the matcher treats two spellings of the same on-screen text as one element while this function keeps them as two different terms. A flow that proves an element present under one spelling and checks `hidden` under an equivalent one (the accented e in `Café` written precomposed, U+00E9, vs decomposed, e + U+0301; or an NBSP a label carries vs a plain space the author types) leaves the `hidden` uncredited, so the runner attaches "cannot fail and proves nothing" to a check that would in fact fail the instant the element stayed on screen, and the recorder refuses to write it. Reproduced end-to-end: `visible` under the precomposed spelling then `hidden` under the decomposed one (both rendering identically) replays pass with the false warning, while the same-spelling control does not, and the decomposed selector is confirmed to match the precomposed node. This is the text-field sibling of the identifier/resource-id divergence on line 36, a distinct trigger (reachable on iOS/macOS text, not only Android ids) that an identifier-only fix would not close.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Re-reviewed at efd7855a (the branch has merged its base a number of times since the last pass, touching several of these files). Ran the full sweep on the evidence model and the recorder/runner gates: claims vs code across the new docstrings and skill docs, nearest-twin and symmetry across the recorder refusals / runner warnings / the single shared evidence rule, non-happy paths and hostile selectors (degenerate, regex-text, scoped, unicode), reachability both ways, and an absence pass against the base features merged in since this PR was written (the swipe directive, the ui-tree-match text folding, the gesture-flag rename, the relaxed tapping rule). Two new findings are inline below, each reproduced end-to-end through the real runner on a mocked tree. The findings from the two earlier passes all still reproduce at this head, confirmed as replies on their existing threads rather than re-filed.

packages/tool-server/src/tools/flows/flow-run.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of efd7855a. `execWhenStep` (flow-run.ts:2213-2229) still reads only `probe.aborted`/`.indeterminate`/`.ok` and sets `met = probe.ok`; `probe.warning` reaches no report, so a vacuously-true `when: { hidden: ... }` still runs its block green with nothing to flag it. The base merges since 9689cf1e touched this file but did not change this path.

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of efd7855a. flow-selector-evidence.ts is byte-identical to 9689cf1e (no base merge touched it): the walk still recurses `walk(step.steps)` at line 116 and credits the guard's selector at 113-115 for every `when:` block, whether or not the guard can ever hold, and the added unit test (flow-selector-evidence.test.ts:72-81) still pins that unconditional credit.

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of efd7855a. The three-shapes wording is unchanged at flow-selector-evidence.ts:19-21 and has shifted (base merge) to flow-actions.ts:187-189. The scope reprieve still exonerates the scoped shape (`hiddenCheckIsFalsifiable`, flow-actions.ts:1569-1600, used at 1687), and nothing still writes `executionPrerequisite` into `establishedSelectors`: its only writers remain passing steps (via establishedTerms) and entered non-hidden `when:` guards.

packages/tool-server/src/tools/flows/flow-add-step.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of efd7855a. Now at flow-add-step.ts:1734: `const cause = ctx?.signal?.aborted === true ? "cancelled" : unmetUiWaitCause(toolResult)` still overrides the wait's own verdict with `cancelled` whenever the signal is down, after awaiting `activeFlowState` first (1731), which widens the window. flow-run.ts:2604-2609 still takes the opposite stance for the same shape ("The wait's own cause decides, not this run's signal").

packages/argent-mcp/src/content.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of efd7855a. content.ts:196-199 still says `warning` has "Two sources today", while its sibling docstring `StepReport.warning` (flow-run.ts:196-206) now names three producers: `await: { idle: true }`, a vacuous `hidden`, and a selector-less gesture left unsettled by a tree-source outage (`GestureSettle.warning`, flow-actions.ts:919).

packages/tool-server/src/tools/flows/flow-finish-recording.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of efd7855a. flow-finish-recording.ts:200 still says `message` "counts them by kind", but `RecordedStepWarning` (flow-utils.ts:236-252) still carries no `kind`, and `warningHeadline` (flow-finish-recording.ts:113-118) still counts a single conversion category, so there is no plural of kinds to count by.

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of efd7855a. flow-selector-evidence.ts is byte-identical to 9689cf1e: line 121 still licenses the whole selector on `wanted.some(term => seen.has(term))`, so a `hidden { identifier: <typo>, text: <established> }` whose full selector can never match is still admitted clean by both recorder and runner because the `text` term collides with an earlier step.

packages/tool-server/src/tools/flows/flow-add-step.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of efd7855a. Now at flow-add-step.ts:1758-1759: the refusal still recommends `await: { idle: true }` "which FAILS when the screen never settles", the opposite of `waitForIdle`'s own contract (flow-actions.ts:2005-2006: an unsettled screen spends the timeout and passes with a warning; only an unreadable tree hard-fails).

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of efd7855a. flow-selector-evidence.ts is byte-identical to 9689cf1e: line 87 still returns `[]` for any `tool` step that is neither `run-sequence` nor `await-ui-element`, so `tool: flow-execute` still contributes no evidence. One refinement on the `run:` side of this thread: the two sides diverge there too. The runner crosses a `run:` fragment (execRunStep shares `state.establishedSelectors`, flow-run.ts:2321 -> 2113), but the recorder's walk (flow-selector-evidence.ts:106-118) recurses only into `when:` blocks and has no `run` case, so a `run:`-established selector is invisible to the recording-time refusal at flow-add-step.ts:1788. A later `hidden` on it is refused at record time yet passes clean at replay, the reverse of what "accepted under run:" implies for the record side.

packages/tool-server/src/tools/flows/flow-run.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of efd7855a. The replay-side gate is now at flow-run.ts:2626-2638 and still reads `vacuousHiddenSelectors` only (the `hidden` class). A raw `tool: await-screen-idle` step answering `{ settled: false }` still finishes `ok:true`, passing with no warning anywhere in the report, the exact class the recorder now refuses, with no replay-side warning analogue.

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of efd7855a. flow-selector-evidence.ts is byte-identical to 9689cf1e: line 36 still stores the identifier verbatim (lowercased), so a flow that proves presence via a qualified `com.example.app:id/submit` still leaves no term a short-spelled `hidden { identifier: submit }` can find, though the matcher folds the two to the same element.

packages/tool-server/src/tools/flows/flow-selector-evidence.ts
```diff
@@ -0,0 +1,122 @@
...
+ * flow keeps its permanently-green gate and is told so on every run.
+ */
+
+/**
+ * The identifier/text a selector names, lowercased. `role` is deliberately
+ * excluded — "some button existed earlier" is not evidence about THIS element.
+ */
+export function selectorIdentityTerms(selector: unknown): string[] {
+  if (selector === null || typeof selector !== "object") return [];
+  const s = selector as { identifier?: unknown; text?: unknown };
```
latekvo, line 33: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Same evidence-model class as the swipe/text-field gaps, one axis over: this function reads only the selector's own `identifier`/`text` and never recurses into `within`/`after`/`next`, so a scope container is never credited. A scoped action like `tap: { text: "OK", within: { identifier: "dialog" } }` requires `dialog` to resolve for the match to succeed (the same premise `hiddenCheckIsFalsifiable` relies on at flow-actions.ts:1596), so it does prove `dialog` present, yet only `text:ok` is credited. A later `hidden: { identifier: "dialog" }` on the container is then falsely refused by the recorder / warned by the runner. Safe direction (a false "proves nothing", never a green vacuous gate), and scroll-to is correctly exempt: its `within` degrades to whole-screen scrolling rather than hard-failing, so a passing scroll-to genuinely does not prove `within` present.

PR #725, opened 2026-08-04
fix(flow): require a positive check to survive a second settled read

hubgan:
> Stacked on #724.

## The problem

A node that no longer belonged to the current screen was still present in the accessibility tree and **satisfied a readiness check** — producing a false-positive save completion.

A silently-wrong green is worse than a flake. The test has stopped testing, and nothing downstream will ever flag it.

## What changes

`exists`, `visible` and `text` must now hold across **two consecutive reads**.

`hidden` is excluded on purpose: it already carries the blind-read guard, and there is no such thing as a node that is stale *into* existence.

## Two bars, and why the second one is deliberately weaker

**Inside the step's budget — the strong bar:** the same elements, with the same frames, on a tree that did not change. A node the tree has not finished evicting cannot survive that: it is surrounded by churn, and one settled beat is enough for it to disappear.

Frames are part of the comparison because an element still sliding into place has not settled, and a `tap` resolved against a moving frame lands where the element no longer is.

**Once the budget is spent — the weak bar:** "it held on the previous read too."

Anything stricter would **regress** this check rather than tighten it. An element sliding into place reports a new frame on every read and would never produce two identical ones; a screen with something permanently in motion (a spinner, a looping animation) never produces two identical trees. Both of those used to pass and must keep passing. The weak form is still strictly stronger than the single read it replaces.

## Blind reads break the chain

A blank tree or a throw clears the weak claim as well as the fingerprints. Leaving `heldPrev` set let the post-deadline branch honour "it held on the previous read" across a read that never actually happened.

## Included here: indeterminate is scored `error`, not `fail`

Three of the tests in this PR depend on it, so it rides along. An indeterminate outcome is not a verdict about the app — the check could not run at all (unreadable tree, an environment failure). Scoring it `fail` makes CI read an environment problem as a regression, and makes a QA author reset a pass streak over it. `error` keeps the run non-ok while saying plainly that the app was never judged.


packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1148,20 +1180,55 @@ async function waitForCondition(
...
+        //
+        // Once the budget is spent the bar drops to "it held on the previous
+        // read too". Anything stricter would REGRESS this check: an element
+        // sliding into place reports a new frame on every read and would never
+        // produce two identical ones, and a screen with something permanently
+        // in motion never produces two identical trees — both used to pass and
+        // must keep passing. The weak form is still strictly stronger than the
+        // single read this replaced.
+        const strong = pendingMatch === fp && pendingTree === treeFp;
+        if (heldPrev && (strong || Date.now() >= deadline)) return { ok: true };
```
latekvo, line 1742: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: A transient blind or thrown read (an empty + degraded tree, or a fetch that throws) landing in the final read window clears `heldPrev` and the pending fingerprints here, and the `finalPoll` break below allows exactly one more read after the deadline — which cannot re-establish the two-consecutive-read chain. That turns a genuinely present, stable element on a screen whose tree is churning (so the strong two-identical-trees form never held pre-deadline) into a hard `fail` at the timeout, where the single-read logic this replaces passed it. Reproduced with three scripted reads of a stable element — a matching read, then a blind empty tree with a degraded hint, then a matching read again — on a churning tree: the head returns `ok:false` (`status: fail`) for the step, while the identical reads on the pre-PR base return `ok:true`. The failure message can also mislead here: with the match still held from the final read it reports the same 'none was visible (zero-area frame)' branch as a genuine visibility miss.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1148,20 +1180,55 @@ async function waitForCondition(
...
+            !everMatched &&
+            !hiddenCheckIsFalsifiable(env, step, lastTree)
+          ) {
+            return { ok: true, warning: vacuousHiddenReason(step.selector) };
+          }
+          return { ok: true };
         }
-        return { ok: true };
+        const fp = matchFingerprint(lastMatches);
+        const treeFp = treeFingerprint(data.tree);
```
latekvo, line 1727: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: The strong confirmation requires two byte-identical fingerprints of the *whole* tree, and `treeFingerprint` walks every descendant. Any node outside the matched set that changes between the two reads — a spinner icon, a clock, a 'updated just now' label — makes the strong form permanently unsatisfiable, so the 'settled' confirmation never fires and the step defers to the deadline weak form. The concrete impact is twofold: every positive `await`/`assert` on a screen with any persistent motion now waits the full `timeout` instead of passing on the first reading (the earlier behavior), and in that same common case the settle-confirmation this change is meant to add is effectively unreachable. The older scoped settle path in this file fingerprints the tree with an `include` predicate precisely so an unrelated animating node cannot defeat it.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1019,6 +1019,31 @@ async function runType(
...
+ * has not settled, and a `tap` resolved against a moving frame lands where the
+ * element no longer is.
+ */
+function matchFingerprint(matches: DescribeNode[]): string {
+  return matches
+    .map(
+      (n) =>
+        `${n.role}|${Math.round(n.frame.x * 1000)},${Math.round(n.frame.y * 1000)},` +
+        `${Math.round(n.frame.width * 1000)},${Math.round(n.frame.height * 1000)}` +
+        `|${n.label ?? ""}|${n.value ?? ""}|${n.identifier ?? ""}|${n.subtreeText ?? ""}`
```
latekvo, line 1582: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: `matchFingerprint` folds `subtreeText` into the match identity, but the `treeFingerprint` it is compared against (ui-tree-match.ts) does not. For a condition located on a container whose `subtreeText` embeds a churning descendant while the label/value the condition actually keys on stay stable, the match fingerprint differs on every read even though the matched content is settled — so the strong form is blocked on a tree that is otherwise identical, and the step is pushed to the deadline weak fallback (and the full-timeout latency that comes with it).

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1019,6 +1019,31 @@ async function runType(
   return { ok: true };
 }
 
+/**
+ * Conditions whose verdict is "the element is there" and which therefore have
+ * to survive a second read before they pass. `hidden` is excluded on purpose:
+ * it already carries the blind-read guard, and there is no such thing as a
+ * node that is stale into existence.
+ */
+const NEEDS_SECOND_READ: ReadonlySet<WaitCondition> = new Set(["exists", "visible", "text"]);
```
latekvo, line 1568: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: Nothing in the new tests pins the `text` entry in this set. `flow-stale-node.test.ts` only exercises the `visible` condition; removing `text` from `NEEDS_SECOND_READ` leaves the entire flow test suite (all 1259 flow tests) green. The claim this change carries — that a `text` positive must survive a second read — is therefore not asserted, so a future change that dropped `text` from the two-read requirement (or broke it through the subtreeText-churn path above) would go uncaught.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: Ran the full sweep on this change at `90844032`: claims vs code, nearest twin (the raw `await-ui-element` tool and the file's own scoped-settle path), non-happy paths, inputs/boundaries, reachability both ways, what outlives the loop, and mutation. The flow suite (1259 flow tests) is green on the PR head; I reproduced one false-failure regression head-vs-base (details in the inline finding on line 1209). Mutations over the new branches: the deadline weak bar is pinned by the tests, but the `text` entry of the two-read set is not.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Still present as of b27748704e5c8e09ccd585783364853ad013a522. Re-ran the scripted repro at head: three reads of a stable `Saved` element on a churning tree — a matching read (~335ms), a blind empty+degraded read completing ~900ms into a 700ms window (clearing `heldPrev` and both fingerprints), then a matching read at ~1249ms. The step returns `status: fail`, and because the final trusted read matched, the reason quotes the zero-area branch (`element(s) matched text="Saved" but none was visible (zero-area frame)`) even though the frame was full-size.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Still present as of b27748704e5c8e09ccd585783364853ad013a522. Measured at head: an `await visible` for an element present and unchanged from the very first read, with one spinner label churning elsewhere on the tree, returns `ok: true` only after the full 2000ms budget (6 polls). The strong form never fires on such screens, so every positive check pays the entire timeout where the single-read logic passed immediately.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Still present as of b27748704e5c8e09ccd585783364853ad013a522. Confirmed at head both mechanically and end-to-end: two trees differing only in a matched node's `subtreeText` produce identical `treeFingerprint` strings while the match fingerprints differ; and an `await exists` keyed on an identifier whose matched node carries churning hoisted text waits out its entire budget on an otherwise fully settled tree.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Still present as of b27748704e5c8e09ccd585783364853ad013a522. Mutation re-run at head: dropping `text` from this set leaves the full flow suite (1448 tests) green.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1484,20 +1516,55 @@ async function waitForCondition(
...
+        // enough for it to disappear.
+        //
+        // Once the budget is spent the bar drops to "it held on the previous
+        // read too". Anything stricter would REGRESS this check: an element
+        // sliding into place reports a new frame on every read and would never
+        // produce two identical ones, and a screen with something permanently
+        // in motion never produces two identical trees — both used to pass and
+        // must keep passing. The weak form is still strictly stronger than the
+        // single read this replaced.
+        const strong = pendingMatch === fp && pendingTree === treeFp;
```
latekvo, line 1741: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: The whole-tree half of this comparison is not pinned by the suite: mutating it to compare only `pendingMatch === fp` leaves all 1448 flow tests green — nothing asserts the extra latency or read count that the tree fingerprint produces (the spinner-churn test asserts only eventual `ok`, and the read-count assertions run on static trees where the match fingerprint alone already agrees). The tree-fingerprint requirement, and the multi-second waits it adds on screens with any persistent motion, can therefore be silently dropped or loosened without any test failing.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Still present as of b27748704e5c8e09ccd585783364853ad013a522

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Still present as of b27748704e5c8e09ccd585783364853ad013a522

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Still present as of b27748704e5c8e09ccd585783364853ad013a522

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Still present as of b27748704e5c8e09ccd585783364853ad013a522

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1357,6 +1357,31 @@ async function runType(
   return { ok: true };
 }
 
+/**
+ * Conditions whose verdict is "the element is there" and which therefore have
+ * to survive a second read before they pass. `hidden` is excluded on purpose:
+ * it already carries the blind-read guard, and there is no such thing as a
+ * node that is stale into existence.
+ */
+const NEEDS_SECOND_READ: ReadonlySet<WaitCondition> = new Set(["exists", "visible", "text"]);
```
latekvo, line 1568: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: A positive condition spelled as a `- tool: await-ui-element` step still passes on a single read: that tool's poll loop returns on the first trusted read (await-ui-element/index.ts:609) with no second-read gate, and this set only governs the directive path. Reproduced end-to-end with the real tool behind a faked AX edge: in the stale-eviction world (read 1 carries "Saved", every later read does not), the tool step reports status "pass" off exactly one read while the directive twin of the same wait fails after 12 reads. The stale-node false-positive class this change targets stays reachable from a flow file one spelling away.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -2501,19 +2501,17 @@ async function execLeafStep(
...
+        // could not run at all (an unreadable tree, no route reader, focus
+        // unconfirmed with nothing to read it from). It is `idle`'s only
+        // non-passing outcome — a screen that merely kept moving passes with a
+        // warning, and so does one that rendered nothing, so what is left there
+        // is a wait that could not run at all — and the selector conditions
+        // reach it the same way. Reporting it as `fail` makes CI read an
+        // environment problem as a regression and a QA author reset a pass
+        // streak over it. `error` keeps the run non-ok while saying plainly
+        // that the app was never judged.
+        if (!r.ok && r.indeterminate) return { ...base, status: "error", reason: r.reason };
```
latekvo, line 2513: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: After this widening, the same total tree-source outage scores differently depending on how the wait is spelled: the directive reaches indeterminate and maps to `error` here, while an unmet `- tool: await-ui-element` result is mapped to `fail` at flow-run.ts:2619 without consulting the cause classification (`unmetUiWaitCause`) that already separates an unreadable tree from a genuine miss — in the unreadable case the tool's own note says the empty tree is not evidence that nothing is on screen. Reproduced with both spellings against an always-throwing tree source: the `await` step ends `error` ("could not read the UI tree…"), the tool step ends `fail`. The reasoning this comment makes — an environment problem must not read as a regression — is contradicted by the tool-step case directly below it.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1484,20 +1516,55 @@ async function waitForCondition(
...
+        // in motion never produces two identical trees — both used to pass and
+        // must keep passing. The weak form is still strictly stronger than the
+        // single read this replaced.
+        const strong = pendingMatch === fp && pendingTree === treeFp;
+        if (heldPrev && (strong || Date.now() >= deadline)) return { ok: true };
+        heldPrev = true;
+        pendingMatch = fp;
+        pendingTree = treeFp;
+      } else {
+        heldPrev = false;
```
latekvo, line 1747: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Nothing pins these resets. Deleting the three lines leaves the entire flow suite green (verified by mutation across ~1461 tests), so a match that holds once and then vanishes would green at the deadline through the weak form even though "it held on the previous read too" is false — the false-green class this change exists to kill, with no test signal when it comes back. The catch-block clear and the `heldPrev` term of the gate above are similarly unpinned.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1357,6 +1357,31 @@ async function runType(
...
+ * it already carries the blind-read guard, and there is no such thing as a
+ * node that is stale into existence.
+ */
+const NEEDS_SECOND_READ: ReadonlySet<WaitCondition> = new Set(["exists", "visible", "text"]);
+
+/**
+ * Identity of the set of nodes a selector matched, for comparing one read
+ * against the next. Frames are included: an element still sliding into place
+ * has not settled, and a `tap` resolved against a moving frame lands where the
+ * element no longer is.
```
latekvo, line 1574: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: The mis-aimed-tap hazard this rationale names is left standing where taps actually resolve their frame: `settleTree` requires two identical consecutive tree fingerprints, but once its budget is spent it returns the previous tree once its minimum read count has landed, even when those reads differ. Measured on a stable matched button with a churning sibling: every selector gesture paid the full ~3000ms settle window and then dispatched anyway from that unsettled read.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -2501,19 +2501,17 @@ async function execLeafStep(
...
+        // An INDETERMINATE outcome is not a verdict about the app: the check
+        // could not run at all (an unreadable tree, no route reader, focus
+        // unconfirmed with nothing to read it from). It is `idle`'s only
+        // non-passing outcome — a screen that merely kept moving passes with a
+        // warning, and so does one that rendered nothing, so what is left there
+        // is a wait that could not run at all — and the selector conditions
+        // reach it the same way. Reporting it as `fail` makes CI read an
+        // environment problem as a regression and a QA author reset a pass
+        // streak over it. `error` keeps the run non-ok while saying plainly
+        // that the app was never judged.
```
latekvo, line 2512: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: This behavior change needs its docs update in the same PR per repo convention. packages/docs/docs/reference/flow-yaml.mdx still states the errored-and-stop outcome inside the `idle` subsection only ("One outcome stops the run: the runner cannot read the UI tree at all…"), the skills copy mirrors it, and nothing documents either the widened errored scope or the new two-consecutive-reads minimum for positive `await`/`assert` conditions.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Ran a fresh full sweep at `b27748704`: claims vs code, nearest twin (`treeFingerprint` in ui-tree-match.ts, the file's own scoped settle path, the raw `await-ui-element` tool, every `indeterminate` consumer), non-happy paths, inputs/boundaries, reachability both ways, lifetime, plus the absence pass with seven mutations over the new branches (each reverted after measuring; full flow suite, ~1461 tests per run). Everything below was reproduced at this head with scripted-tree vitest runs through `createRunFlowTool`: the transient-blind-read false fail, the full-timeout latency under persistent motion, subtreeText churn blocking the strong form, a single-read pass through a `- tool: await-ui-element` step, the error/fail outage split between directive and tool spellings, and the ~3s-per-gesture settle fallback. The first four inline threads from the earlier review were re-verified as still present and answered on their own threads.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1337,20 +1369,55 @@ async function waitForCondition(
...
+      // post-deadline branch honour that claim across a read that never
+      // happened. The blind-read branch above clears all three for the same
+      // reason, and a throw is strictly blinder than a blank tree.
       lastReadTrusted = false;
+      heldPrev = false;
+      pendingMatch = undefined;
+      pendingTree = undefined;
     }
     if (Date.now() >= deadline) {
       if (finalPoll) break;
```
latekvo, line 1767: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: A positive condition that first becomes true on the latched final-poll read is scored `fail`. That read satisfies the condition and sets `heldPrev`, but the `break` here exits before any further read could confirm, and the determinate tail then builds its fail verdict from `lastMatches` — which on that very read matched. Reproduced at `dcb098155`: reads absent ×3 then present, `await visible "Saved"` timeout 600ms → status `fail` with reason `element(s) matched text="Saved" but none was visible (zero-area frame)` even though the final trusted read carried the node with a nonzero frame (contrast case absent/present/present passes). Before this change the satisfying read returned `ok: true` on its own, so the effective detection window for `exists`/`visible`/`text` shrank by roughly one poll interval (~300ms default): an element rendering in the last poll slot of a tight `await` timeout — or late inside an `assert`'s 1000ms grace — now fails where it used to pass.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -2321,17 +2321,17 @@ async function execLeafStep(
...
+        if (!r.ok && r.indeterminate) return { ...base, status: "error", reason: r.reason };
```
latekvo, line 2513: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: With the widening to all step kinds, an APP-caused hang during an unmet `await`/`assert` is now reported as `error` — 'the app was never judged' — even though the app was judged and failed: it wedged, never reached the expected state, and took the tree source down with it. The dark-tail tier cannot tell that apart from an environment death. Both create-flow skills direct an `errored` step to repair-the-environment-and-rerun, so a genuine hang regression gets classified as an environment flake and rerun instead of ever being reported. Reproduced at `dcb098155`: `await visible` timeout 1000ms, first read trusted and showing no match, every subsequent read rejecting with `native devtools is unavailable` → `failed: 0, errored: 1`, status `error`, reason quoting the unreadable tail. Before this PR the same window scored `fail`.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Fresh sweep at `dcb098155` (the branch moved by rebase since `b27748704`): claims vs code, nearest twin (`treeFingerprint`, the raw `await-ui-element` tool, `settleTree`, the skills' triage copy), non-happy paths, inputs/boundaries, reachability both ways, lifetime, and the absence pass over the widened `indeterminate` rule and every new test. All previously reported threads were re-checked at head and still hold (replies on each thread). Two new findings, each proven with a scripted reproduction run against this head, posted inline below.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Still present as of dcb098155e9756459486fdfb6c38583d720d0517. Re-ran the scripted repro at head — reads \[match, throwing-after-500ms, match\], `await visible "Saved"` timeout 700ms → `ok: false`, status `fail`, 3 reads, ~827ms; the element matched on read 1 and read 3 yet the step failed.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Still present as of dcb098155e9756459486fdfb6c38583d720d0517. Re-measured at head: an element present and unchanged from the first read with one churning sibling label returns `ok: true` only after the full budget (2009ms, 8 polls) — the strong form never fires on such screens.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Still present as of dcb098155e9756459486fdfb6c38583d720d0517. `matchFingerprint` still folds in `subtreeText` while `treeFingerprint` (ui-tree-match.ts:1169) still does not, so the mismatch blocks the strong form on otherwise settled trees.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Still present as of dcb098155e9756459486fdfb6c38583d720d0517. No new test pins the `text` entry — flow-stale-node.test.ts still exercises only the `visible` condition.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Still present as of dcb098155e9756459486fdfb6c38583d720d0517. The spinner-churn test still asserts only eventual `ok`, and the read-count assertions still run on static trees where the match fingerprint alone already agrees — the tree-fingerprint half remains unpinned.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Still present as of dcb098155e9756459486fdfb6c38583d720d0517. `await-ui-element`'s poll loop (`onSample`) still returns `done: true` on the first trusted matching read, so a positive check spelled as a tool step still passes on a single read while the directive path requires two.

packages/tool-server/src/tools/flows/flow-run.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Still present as of dcb098155e9756459486fdfb6c38583d720d0517. The rule is now widened to all step kinds at flow-run.ts:2334, but the unmet `- tool: await-ui-element` branch (~flow-run.ts:2446) still maps to `fail` without consulting the indeterminate/unreadable classification — the same total outage still scores differently by spelling.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Still present as of dcb098155e9756459486fdfb6c38583d720d0517. No new test pins these resets — there is still no suite coverage for a match/blind-or-throw/match sequence asserting the chain stays broken.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Still present as of dcb098155e9756459486fdfb6c38583d720d0517. `settleTree` is unchanged: once its minimum read count has landed it still returns the previous tree after the budget even when those reads differ, so the mis-aimed-tap hazard the rationale names stands where taps resolve their frames.

packages/tool-server/src/tools/flows/flow-run.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Still present as of dcb098155e9756459486fdfb6c38583d720d0517. flow-yaml.mdx still documents the errored-and-stops-the-run outcome inside the `idle` subsection only, and neither it nor the skills copy documents the widened `errored` scope or the two-consecutive-reads minimum for positive conditions.

packages/tool-server/src/tools/flows/flow-run.ts
latekvo, line 2513: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: The error-vs-fail contract this line establishes is broken one nesting level down. A composed `run:` whose inner directive went indeterminate now produces an inner run with `failed: 0, errored: 1, ok: false` — and `flowExecuteOutcome` (flow-nested-outcome.ts:137-145) maps any `ok: false` to outer status `fail`, so the environment outage this change reclassifies as 'never judged' reads as a step FAILURE in the outer run's report and the recorder's refusal text, exactly the misclassification the widening exists to prevent. Confirmed at `dcb098155` by tracing the mapping: the reason string does quote `1 errored`, but the step status the skills' triage keys on is `fail`.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: One more consequence found on a second pass, same widening.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Two adjacent variants of this same mechanism found on a second pass: (1) if the matching read is followed only by reads that throw or go blind within the 600ms dark-tail tolerance, the loop exits through the tier-3 branch instead of the indeterminate one and issues a DETERMINATE fail quoting `lastMatches` from the very read that matched — a self-contradicting verdict where pass or indeterminate would be honest. (2) The same applies to a `when:` guard whose positive first becomes true on the latched read: instead of a scored fail it produces a silent block skip ("condition not met"), so a guarded section is skipped without any flag. Both pre-change passed.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1681,20 +1713,55 @@ async function waitForCondition(
...
+            return { ok: true, warning: vacuousHiddenReason(step.selector) };
+          }
+          return { ok: true };
         }
-        return { ok: true };
+        const fp = matchFingerprint(lastMatches);
+        const treeFp = treeFingerprint(data.tree);
+        // Two consecutive reads must agree. Inside the budget the bar is the
+        // strong one — the SAME elements, on a tree that did not change — so a
+        // node the tree has not finished evicting cannot be mistaken for the
```
latekvo, line 1730: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: The strong two-read form (`pendingMatch === fp && pendingTree === treeFp`) is satisfied by two byte-identical reads, so a stale node that lingers unchanged for one extra poll passes on the second read — the check returns ok on the stale match without ever reading the destination screen. The premise here ("surrounded by churn, and one settled beat is enough for it to disappear") holds only while a stale node is being actively evicted; the common "success state shown for a beat, then navigate" shape — which is the false-positive-save-completion this change targets — is static during that window and survives both reads. Reproduced with scripted reads [{Saved, Old screen}, {Saved, Old screen} (identical), {Compose}] and `await visible "Saved" timeout 3000`: the step returns status `pass` on read 2 (reads=2), never reaching {Compose}; the transient-eviction control [{Saved, Old screen}, {Compose}] correctly fails after the full budget. The dedicated test "refuses a match the tree had not finished evicting" only covers eviction by the very next read, so this residual is unpinned.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Re-ran the full sweep on this change at c20dfd4be0805b91c6729de56467e4f6c65b9e49 — claims vs code, nearest twin (the file's own scoped `settleTree` / `treeFingerprint(…, include)` path and the raw `await-ui-element` tool), non-happy paths, inputs, reachability both ways, lifetime, plus an absence/mutation pass across the new branches and constants — and reproduced the timing behaviors against the runner with a scripted `fetchFlowTree`. Every prior thread is re-confirmed at this head (replied inline), and one new correctness finding is left inline.

Two comments this change silently falsified fall outside the diff hunks, so they can't be anchored inline and are noted here:

- **`flow-actions.ts`, the `DirectiveOutcome.indeterminate` JSDoc (~lines 172-181):** still states "a plain `assert` reports it as an ordinary failure; `idle`, which has no condition to fall back on, is scored `error`". After the widening in flow-run.ts, an indeterminate `assert` (and `await`) is scored `error`, not an ordinary failure — the assert-vs-idle contrast on the canonical doc for the very field this change turns on is now inverted, and the PR's own test edits assert `assert:error`.
- **`flow-add-step.ts:560-562`, the `PROBE_BUDGET_MS` rationale:** still states "the clean case returns from the first read that satisfies the condition". The recorder's cross-tree re-probe runs `probeWhenCondition` → `waitForCondition`, so a clean positive re-probe (`visible`/`exists`/`text`) now costs two reads, not one; the 6000ms budget stays adequate only because both branches now cost two reads (the number is fine — the rationale is stale).

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of c20dfd4be0805b91c6729de56467e4f6c65b9e49. Re-ran the scripted repro at head: churning matches, a throw landing at ~702ms (the final window past a 700ms deadline), then the single post-deadline final poll matches with a full-size frame. The step returns status `fail`, and because that final trusted read matched, the reason quotes `element(s) matched text="Saved" but none was visible (zero-area frame)` even though the frame was full-size. The throw clears `heldPrev`+fingerprints and `finalPoll` allows exactly one more read, which cannot re-establish the two-read chain.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of c20dfd4be0805b91c6729de56467e4f6c65b9e49. Measured at head: an `await visible` for an element present and unchanged from the first read, with one spinner label churning elsewhere on the tree, returns `ok: true` only after the full 2000ms budget (8 reads). The no-churn baseline passes in ~303ms (2 reads). `treeFingerprint(data.tree)` walks every descendant with no `include` predicate, so any unrelated motion keeps the strong form from ever firing.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of c20dfd4be0805b91c6729de56467e4f6c65b9e49. Confirmed by code at head: `matchFingerprint` (flow-actions.ts:1576) folds `subtreeText` into the identity, while `treeFingerprint` (ui-tree-match.ts:1333-1350) still does not; `subtreeText` is stamped on the flow read path by the flatten step (flow-tree-flatten.ts:153). So a matched container whose `subtreeText` embeds a churning descendant while its label/value/frame stay stable produces a differing match fingerprint every read on an otherwise-identical tree, blocking the strong form and deferring to the deadline weak fallback.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of c20dfd4be0805b91c6729de56467e4f6c65b9e49. Mutation re-run at head: dropping `text` from `NEEDS_SECOND_READ` leaves the full flow suite green (59 files / 1655 tests). The PR's dedicated test (flow-stale-node.test.ts) exercises only the `visible` condition, so the two-read requirement for `text` is unasserted.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of c20dfd4be0805b91c6729de56467e4f6c65b9e49. Mutation re-run at head: reducing the strong compare to `pendingMatch === fp` (dropping the tree-fingerprint half) leaves all 1655 flow tests green; the complementary mutation to `pendingTree === treeFp` (dropping the match half) is also green. Both halves of the strong-form comparison are unpinned, so the extra latency and read count the tree fingerprint produces can be dropped or loosened with no test failing.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of c20dfd4be0805b91c6729de56467e4f6c65b9e49. Confirmed by code at head: `await-ui-element`'s poll loop returns `{ done: true, result: { success: true } }` on the first trusted matching read (await-ui-element/index.ts, `onSample`), with no second-read gate. `NEEDS_SECOND_READ` governs only the directive path, so a positive check spelled `- tool: await-ui-element` still passes off a single read while the `- await:` twin of the same wait requires two.

packages/tool-server/src/tools/flows/flow-run.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of c20dfd4be0805b91c6729de56467e4f6c65b9e49. Confirmed by code at head: an unmet `await-ui-element` result is mapped to `status: "fail"` unconditionally at flow-run.ts:2617-2624 (only the `cancelled` cause is special-cased, at 2614); `unmetUiWaitCause`'s unreadable classification is not consulted. So the same total tree-source outage scores `error` through the `- await:` directive (flow-run.ts:2513) but `fail` through the `- tool: await-ui-element` spelling.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of c20dfd4be0805b91c6729de56467e4f6c65b9e49. Mutation re-run at head: deleting the three else-branch resets (`heldPrev`/`pendingMatch`/`pendingTree`) leaves the full flow suite green (1655 tests). No test pins a match/reset/match sequence, so a match that holds once and then vanishes would still green at the deadline through the weak form even though "it held on the previous read too" is false.

packages/tool-server/src/tools/flows/flow-run.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of c20dfd4be0805b91c6729de56467e4f6c65b9e49. flow-yaml.mdx:259 still documents the errored-and-stops-the-run outcome only inside the `idle` subsection, and the skills copy mirrors it (argent-qa-flows/SKILL.md:103 and argent-create-flow/references/*.md scope the environment-errored example to `idle`/`run:`). Neither the widened `errored` scope (now every step kind, flow-run.ts:2513) nor the two-consecutive-reads minimum for positive `await`/`assert` conditions is documented, and the PR touches no packages/docs file.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of c20dfd4be0805b91c6729de56467e4f6c65b9e49. Re-ran the scripted repro at head: reads carry a non-matching screen three times, then the target from read 4; `await visible "Saved"` timeout 600ms. The target first appears on the post-deadline final poll (~602ms) with a full-size frame, yet the step is scored `fail` with reason `element(s) matched text="Saved" but none was visible (zero-area frame)`. The contrast case (target present from read 2) passes. The satisfying read sets `heldPrev`, but the `finalPoll` break exits before any read can confirm it, and the determinate tail builds the fail verdict from the very `lastMatches` that matched.

packages/tool-server/src/tools/flows/flow-run.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of c20dfd4be0805b91c6729de56467e4f6c65b9e49. Reproduced at head: `await visible` timeout 1000ms, a first trusted read with no match, then the tree source rejecting for the rest of the window -> `failed: 0, errored: 1`, status `error`, reason quoting the unreadable tail. The dark-tail tier cannot tell an app that wedged and took the tree source down from an environment death, and the skills route `errored` to repair-the-environment-and-rerun (argent-create-flow/references/reliability-and-recovery.md:113, argent-qa-flows/SKILL.md:103) with a carve-out only for `launch:` — so an app-hang regression during a selector wait is rerun rather than reported.

latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of c20dfd4be0805b91c6729de56467e4f6c65b9e49. Confirmed by code at head: `flowExecuteOutcome` (flow-nested-outcome.ts:136-144) maps any inner `result.ok === false` to outer `status: "fail"`, without consulting `errored` vs `failed`. So a nested `run:` whose inner directive went indeterminate (inner `errored: 1, failed: 0, ok: false`) surfaces in the outer run as a step `fail`, contradicting the error-vs-fail contract this line establishes one nesting level up.

PR #726, opened 2026-08-04
fix(flow): refuse to type into a field that never took focus

hubgan:
> Stacked on #725.

## The problem

`runType` called `waitForFocus` and **threw the result away**. The helper's own documentation admitted it "falls through to typing after the timeout rather than failing the step."

Keys are injected at the **HID level** — they go to whatever holds focus, not to the element the step names. So an unconfirmed focus meant typing into whatever the app last put the caret in.

The damage surfaces far from its cause. Captured live: the leading character was dropped, iOS autocorrect completed the truncated word, and the app saved **"Regent QA Merchant"** instead of "Argent QA Merchant". A green run, with wrong data in it.

Recording cannot catch this. The authoring skill requires confirming focus with `describe` before recording a `keyboard` step — but at replay there is no `describe` step and no agent, so `type:` typed into nothing.

## Three ways of not confirming focus, kept apart

They warrant opposite responses, so collapsing them would be wrong in both directions:

| verdict | meaning | response |
| --- | --- | --- |
| `unreported` | the source cannot surface focus at all | nothing to conclude — type as before |
| `unreadable` | every read of the window threw | environment failure → **indeterminate** |
| `unconfirmed` | reads succeeded, nothing focused overlapped the field | real evidence → retry once, then **fail** |

The retry exists because losing the focus race is far more common than the field being untappable. The failure text names what to look for — a sheet mid-animation, a toast, or an element that is not a text input at all.

## The overlap test was wrong in both directions

It was a bare **intersection**, which an element merely clipping a corner of the target satisfies — that is evidence of nothing.

Plain **nesting** is wrong the other way, and this one was live: a modal composer three quarters of the screen tall *contains* every node of the screen behind it, so it confirmed focus for any of them. `type:` then injected the keys into the composer while reporting a pass on a static label behind it. Silent, and wrong in the worst way — the text lands somewhere real.

The direction that carries evidence is the focused element **inside** the target (the documented case: the selector matches a testID container, the input inside it reports focus). A focused **ancestor** splits by whether it can receive text itself: an editable one — any composer or rich-text editor covering elements behind it, including an ARIA `role=textbox` on Chromium — is precisely the wrong element and never vouches, at any size. A non-editable container cannot take HID text itself, so its flag can only reflect a descendant; genuine wrappers come in every size (a shadow-DOM host, a tappable form row), so no size bound separates them from coverers without refusing fields that work.


packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -496,25 +534,29 @@ function collectFocused(node: DescribeNode, acc: DescribeNode[]): DescribeNode[]
...
+      if (!FOCUS_REPORTING_SOURCES.has(source)) return "unreported";
+      read = true;
       const target = flowSelectorToFrame(tree, into) ?? tappedFrame;
-      if (collectFocused(tree, []).some((n) => framesOverlap(n.frame, target))) return;
+      if (collectFocused(tree, []).some((n) => framesNest(n.frame, target))) return "confirmed";
     } catch {
       // transient describe failure — retry until the deadline
     }
-    if (Date.now() >= deadline) return;
+    if (Date.now() >= deadline) return read ? "unconfirmed" : "unreadable";
```
latekvo, line 751: [[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: The `unreadable` verdict this return feeds - and the `indeterminate: true` / "could not read the UI tree" outcome it drives in `runType` (1079-1087) - is the one focus verdict with no asserting test. I dead-coded this arm (always return `"unconfirmed"`, dropping `indeterminate: true`) and the full flow-type-focus suite still passed 6/6, so a future change that collapses the env-outage tier into a plain `unconfirmed` -> `fail` would ship unflagged. The counter-arm IS pinned (the "fails the step rather than typing" test asserts status `fail` for `unconfirmed`), so nothing distinguishes the two verdicts that are the entire point of the split: an all-reads-throw fetch during a `type` step (no mock in the suite produces one) would silently drop from `error` - which the runner scores for an indeterminate outcome and which says "env failure, app never judged" - to a hard `fail` that reads in CI as an app regression.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -988,6 +1030,22 @@ async function runRotate(
  * later steps (chained form fields that end in an explicit submit `tap` should
  * pass `submit: false`).
  */
+/**
+ * Tap a field's centre and wait for focus to land there, re-resolving the
+ * field's frame first (the keyboard sliding up can scroll it).
+ */
+async function tapForFocus(
```
latekvo, line 1359: [[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: When `tapForFocus` was extracted, runType's original docblock (lines ~1026-1032, "Resolve `into` -> tap to focus -> wait for focus to land -> type text ... unless `submit` is explicitly `false`...") was left sitting directly above this new function instead of being moved down to `runType`, which now has no doc comment at all. A reader scanning the file attaches the type-via-keyboard / submit story to `tapForFocus`, which does neither, while `runType` (line 1049) - the function that actually types and owns the submit semantics - reads as undocumented. (The `tapForFocus` docblock below it also describes re-resolving the field's frame, which this function never does - that re-resolution lives in the caller.)

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -997,18 +1055,37 @@ async function runType(
...
+  // every keyboard dispatch (the keyboard tool has no abort handling of its
+  // own), so a cancelled run can never type into, or submit, whatever the app
+  // has focused after the caller gave up.
+  if (focus === "aborted" || env.signal?.aborted) return ABORTED_OUTCOME;
+  if (focus === "unconfirmed") {
+    return {
+      ok: false,
+      reason:
+        `${describeSelector(step.into)} did not take keyboard focus within ` +
+        `${TYPE_FOCUS_TIMEOUT_MS}ms, after two taps — nothing was typed. Keys are injected at ` +
```
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: This "after two taps" message can be emitted after only ONE tap. On `unconfirmed`, the retry only re-taps if `retryFrame = await waitForFrame(step.into)` (line 1060) resolves truthy; if the field can no longer be re-found - it scrolled off-screen or was covered during the keyboard slide, the exact situation this message's own "something is covering the field" clause describes - the second `tapForFocus` is skipped, `focus` stays `unconfirmed`, and control reaches this line having tapped once while the text reports two.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -997,18 +1055,37 @@ async function runType(
...
+  if (focus === "unconfirmed") {
+    const retryFrame = await waitForFrame(env, step.into);
+    if (retryFrame === "aborted") return ABORTED_OUTCOME;
+    if (retryFrame) focus = await tapForFocus(env, step.into, retryFrame);
+  }
+  // waitForFocus reports abort separately from focus/timeout — re-check before
+  // every keyboard dispatch (the keyboard tool has no abort handling of its
+  // own), so a cancelled run can never type into, or submit, whatever the app
+  // has focused after the caller gave up.
+  if (focus === "aborted" || env.signal?.aborted) return ABORTED_OUTCOME;
```
latekvo, line 1643: [[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: The abort re-checks on the new focus-wait call graph - the retry's `waitForFrame` returning "aborted" (1060-1061), this one, and the pre-Enter check (1091) - have no exercising test. The sibling gestures get dedicated abort coverage (e.g. flow-pinch, flow-rotate), but `type`'s new retry/settle path is bare; a dropped re-check here would let a cancelled run type into, or submit, whatever the app has focused, reported as a pass rather than the uniform aborted skip.

latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: Ran the full sweep against the PR head (0ffc3ad): claims vs code, nearest twin, non-happy paths, inputs, reachability both ways, lifetime, and the absence pass (mutation). Built the head (tsc --build, exit 0) and ran flow-type-focus.test.ts (6/6 pass). The focus-gating itself is sound - typing is correctly blocked on confirmed focus, and the composer/voucher geometry bug is genuinely fixed. Three LOW items and one MEDIUM are posted inline. No HIGH findings.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: Ran the full look-don't-touch sweep against the PR head (0ffc3ad14): claims vs code, nearest twin, non-happy paths, inputs, reachability both ways, what-outlives-the-call, and the absence pass (mutation / test discrimination), across the two changed files. Ran flow-type-focus.test.ts at head - 6/6 green. The four findings already on this PR from the prior pass are all current-state at this head and are not re-filed. Two new findings, each proven by vitest repros driven through the real flow-run path: (1) the framesNest ancestor area-ratio heuristic misfires in both directions - a focused non-wrapper element at <=4x the target's area vouches and types into the wrong element, while a genuine >4x wrapper/host (a Chromium shadow-DOM input, or a chrome-heavy row) is refused; the boundary is exact at 4.0x vs 4.1x. (2) waitForFocus's unchanged docblock now asserts the opposite of the code (overlap vs framesNest containment; 'falls through to typing' vs a hard fail on unconfirmed).

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -436,6 +437,43 @@ function framesOverlap(a: DescribeFrame, b: DescribeFrame): boolean {
...
+ * documented case where the selector matches a testID container and the input
+ * inside it reports the focus. The reverse, a focused ancestor, is accepted
+ * only while it stays close to the target's own size, because a platform may
+ * legitimately mark a field's immediate wrapper focused rather than the field.
+ */
+function framesNest(focused: DescribeFrame, target: DescribeFrame): boolean {
+  if (frameWithin(focused, target)) return true;
+  return (
+    frameWithin(target, focused) &&
+    frameArea(focused) <= frameArea(target) * FOCUS_ANCESTOR_AREA_RATIO
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: `framesNest`'s ancestor arm is decided purely by an area ratio (`FOCUS_ANCESTOR_AREA_RATIO = 4`), which cannot distinguish the field's real wrapper from any other focused element that merely encloses the field - and it errs in both directions. False pass: a focused element that only CONTAINS the target is accepted whenever its area is at or below 4x the target's (reproduced with a focused composer covering a field behind it at 2.46x area: the step reported ok and the text was injected at HID level into the composer, which is whichever element actually holds focus; the boundary is exact - at 4.0x a foreign covered element vouches and types, at 4.1x it is refused). False refusal: a wrapper/host that genuinely holds focus but is larger than 4x the field (a Chromium shadow-DOM input whose focus is reported on its host, or any input inside a bigger tappable row) is rejected, so the step hard-fails a field that accepts keys. One size heuristic therefore silently injects into the wrong element on one side and refuses genuinely-typeable fields on the other.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -496,25 +534,29 @@ function collectFocused(node: DescribeNode, acc: DescribeNode[]): DescribeNode[]
  * through to typing after the timeout rather than failing the step, since "no
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: `waitForFocus`'s docblock was left unchanged by this change and now reads as the inverse of the code it documents: it says the poll waits for an element reporting `focused` that *overlaps* the typed-into element, and that 'an unconfirmed poll falls through to typing after the timeout rather than failing the step'. The function now returns a verdict that `runType` turns into a hard step failure on `unconfirmed` (after one retry), and the check is `framesNest` containment, not overlap. A maintainer reading the helper's contract is told it types anyway when focus is not seen and that it uses overlap - the opposite of what it now does.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Still present as of 53943df996decafb1d52832a0fc376a9ab000d11. Re-confirmed by an independent mutation pass: collapsing `"unreadable"` into `"unconfirmed"` (M4) and dropping `indeterminate: true` from the outcome (M9) both leave the full flow-type-focus suite green.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Still present as of 53943df996decafb1d52832a0fc376a9ab000d11.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Still present as of 53943df996decafb1d52832a0fc376a9ab000d11.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Still present as of 53943df996decafb1d52832a0fc376a9ab000d11. Re-confirmed by an independent mutation pass: deleting the abort re-check at flow-actions.ts:1401 outright leaves the entire flow-type-focus suite green.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Still present as of 53943df996decafb1d52832a0fc376a9ab000d11. Additional mutation evidence: bumping FOCUS_ANCESTOR_AREA_RATIO from 4 to 100 leaves the suite green — every ratio between the tested ~1.27x wrapper and the ~157x composer is unpinned, so the window this thread describes can silently widen to ~100x with nothing failing.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Still present as of 53943df996decafb1d52832a0fc376a9ab000d11.

latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Ran a fresh full sweep against head 53943df9: claims vs code, nearest twin, non-happy paths, inputs, reachability both ways, what outlives the call, plus the absence pass (stale prose / siblings / symmetry / dead code) and an 11-mutant discrimination pass over the new branches — baseline flow-type-focus suite 6/6 green. The focus-gating logic itself holds up; all previously reported findings are current-state and are confirmed on their existing threads rather than re-filed.

One new finding, posted here because the file is outside this PR's diff so it cannot carry an inline comment:

packages/skills/skills/argent-create-flow/references/live-authoring.md:126 — the Typing paragraph still tells flow authors that `type:` replay's focus wait "falls through to typing rather than failing — so retain the committed-value check." This PR makes replay do the opposite: an unconfirmed poll now retries once and then fails the step with nothing typed. Anyone following this guidance will model `type:` replay as silently typing into whatever holds focus and will treat lost characters as an expected hazard to be caught manually with a value check, when the runner now surfaces exactly that condition as a step failure.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Fixed in 848fc605. Added a test that scripts reads 1-2 to succeed and every read from the focus wait's first look on to throw, then asserts the step scores `error` (indeterminate) rather than `fail`, that no key was injected, and that no second tap was made. Mutation-checked your dead-coding scenario: collapsing `unreadable` into `unconfirmed` now fails this test (`expected error, received fail`), so the env-outage tier can no longer be collapsed unflagged.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Fixed in 848fc605. The misplaced runType docblock now sits on `runType` (extended with the focus-confirm/retry/indeterminate story), and `tapForFocus`'s own docblock no longer claims the re-resolution it never did — it states plainly that the caller resolves the frame.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Fixed in 848fc605. The tap count is now tracked (`taps`) and the message says `after N tap(s)`. Regression test added for exactly your shape: the field vanishes after the first tap (nothing to re-find), asserting the reason contains "after 1 tap", contains no "two taps", and only one `gesture-tap` fired.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Fixed in 848fc605. Three tests now cover the graph: cancel during the first focus wait (asserts skip + zero keys after one tap), cancel during the initial frame resolution (asserts skip with zero taps AND zero keys), and cancel during the keyboard-text call (asserts the submitting Enter never fires). Mutation-checked: deleting the pre-Enter re-check fails its dedicated test. Deleting the retry-`waitForFrame` or settle-abort sites individually does NOT change any outcome — the surviving layers (`env.signal?.aborted` before dispatch, `settleTree`'s loop-top check) still produce an aborted skip with nothing tapped or typed — which is the actual property the tests pin: no cancelled run can type or submit through any single-site regression.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Fixed in 848fc605. Reproduced both directions against the old head first: a focused EditText composer at 3x area covering a label behind it typed into the composer and reported a pass (your 2.46x repro, same mechanism), and a genuine focused wrapper at 8x area hard-failed a typeable field. The fix replaces the size heuristic as the decider with an editable-role split: an EDITABLE ancestor never vouches at any size — it is itself the input that would swallow the HID keys, which is the entire wrong-element bug class including the PR's motivating composer case; a NON-editable container always can — it cannot take HID text itself, its flag can only reflect a descendant, and real wrappers come in every size (shadow-DOM host, tappable form row), so any numeric cutoff refuses genuinely-typeable fields while a role test refuses precisely the elements that could misdirect text. Both repros are now permanent tests, plus a mutation check (`return true` for the ancestor arm) kills the editable-ancestor test. Residual risk documented in the docblock: a non-editable coverer marked focused alongside a truly-focused element elsewhere would vouch — but it can no longer receive the text itself, so the failure mode degrades from silent misdirection into a wrong-but-real focus read.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Fixed in 848fc605. The docblock now describes what the code does: nesting per `framesNest` rather than overlap, the five-verdict return, and that the caller retries once and hard-fails `unconfirmed` instead of the claimed fall-through-to-typing.

latekvo:
[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Full sweep complete against head 9ce2055c0 — all six reviewer threads resolved (fixed in 848fc605 / a6a5827a / 941ac3df / b7bbeb2e6 / 9ce2055c0, each replied on-thread and mutation-verified), plus five self-review passes that came back progressively cleaner:

- **848fc605** — replaced the area-ratio ancestor rule with an editable-role split (both failure directions reproduced first: false pass via a covering composer at 3x area, false refusal of a genuine focused wrapper at 8x); pinned the previously untested `unreadable`/indeterminate tier, three abort paths, and the one-tap message.
- **a6a5827a** — ARIA `role=textbox`/`searchbox` ancestors no longer vouch (reproduced through the real run path); refreshed the skill/docs prose that still described fall-through-to-typing, and aligned the PR body with the implemented rule.
- **941ac3df** — same fix for the text-entry `combobox`/`spinbutton` patterns; corrected the stale flow-ios-tree comment about old injected frameworks.
- **b7bbeb2e6 + 9ce2055c0** — the walker now surfaces `isContentEditable` as an `editable` flag (a bare contenteditable div carries no distinguishing role), and the call is receiver-guarded after verification caught it throwing Illegal invocation on inline-SVG pages; both pinned by walker-script regression tests.

Final state: flows suite 1489/1489 green, tsc/eslint/prettier/docusaurus build clean. Known residuals, all fail-closed or pre-existing and documented in-code: closed shadow hosts whose focused input lives inside them are the one editable shape still unmarked (docblock), structural-collapse promotion drops `editable`/`focused` for `<div contenteditable><p>…</p>` editors exactly as it already dropped `focused` before this branch (parity, follow-up candidate), and iOS keyboard-proxy fields now fail loudly instead of typing blind — which is this PR's point.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Ran a fresh full sweep against head 9ce2055c0: claims vs code, nearest twin, non-happy paths, inputs, reachability both ways, what outlives the call, plus the absence pass (stale prose / siblings / symmetry) and a second independent pass over the peripheral files. Baseline: tsc build clean after workspace deps built, flow-type-focus + describe-chromium-script suites 48/48 green. Two prior review rounds' threads were checked first — nothing re-filed; all six earlier threads are fixed at this head and were confirmed rather than re-opened.

Findings below as inline comments. One behavioral finding was proven by a reproduction driven through the real run path before reporting; the prose findings were verified against the runner's actual scoring paths. The previously documented residuals (closed shadow hosts, non-editable coverer degradation, structural-collapse flag drops) were not re-reported.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -540,6 +541,60 @@ function framesOverlap(a: DescribeFrame, b: DescribeFrame): boolean {
...
+ * canonical editable roles `textbox`/`searchbox`, plus the text-entry
+ * `combobox`/`spinbutton` patterns, which take DOM focus on the widget itself)
+ * or the raw tag name (`input`, `textarea`) otherwise. A bare contenteditable
+ * div carries none of those — its role is its raw tag — so the walker surfaces
+ * it as `editable: true` instead. The one shape this rule still cannot see is
+ * a closed shadow host whose focused input lives inside it, noted here
+ * because it is the editable form left unmarked.
+ */
+const EDITABLE_ROLES =
+  /^(textfield|axtextfield|input|textarea|textbox|searchbox|combobox|spinbutton)$/i;
```
latekvo, line 625: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Reproduced at 9ce2055c0 through the real run path (android-devtools hierarchy stub, same harness as this PR's own tests): a focused `android.widget.AutoCompleteTextView` covering a TextView behind it vouches for that target — the step reported **pass** while the `keyboard` call fired, so the HID keys landed in the autocomplete field and nothing flagged it. `AutoCompleteTextView` extends `EditText`, but `deriveUiAutomatorRole` maps it to `StaticText` (uiautomator-parser.ts:129-130: "autocompletetextview" contains "textview" and not "edittext"), so `isEditable()` is false here and the focused-ancestor leg of `framesNest` treats it as a non-editable wrapper. The identical geometry with class `EditText` fails correctly, so the class name alone decides between a clean fail and the exact silent misdirection this PR exists to prevent. Custom `EditText` subclasses with names falling through to the raw-name branch hit the same hole. It also makes the docblock's claim above that the closed shadow host is "the one shape this rule still cannot see" inaccurate on Android.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -540,6 +541,60 @@ function framesOverlap(a: DescribeFrame, b: DescribeFrame): boolean {
...
+ * DOM walker reports the ARIA `role` attribute when one is set (so the
+ * canonical editable roles `textbox`/`searchbox`, plus the text-entry
+ * `combobox`/`spinbutton` patterns, which take DOM focus on the widget itself)
+ * or the raw tag name (`input`, `textarea`) otherwise. A bare contenteditable
+ * div carries none of those — its role is its raw tag — so the walker surfaces
+ * it as `editable: true` instead. The one shape this rule still cannot see is
+ * a closed shadow host whose focused input lives inside it, noted here
+ * because it is the editable form left unmarked.
+ */
+const EDITABLE_ROLES =
```
latekvo, line 624: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Test-discrimination gap, verified by mutation at this head: deleting the `input`, `textarea`, `searchbox` and `spinbutton` alternations from this regex leaves the entire flow-type-focus suite green (16/16). These are exactly the ARIA text-entry shapes the PR's comments rely on this set to catch, so a future refactor could drop any of them and reintroduce the composer-swallows-the-keys false pass for those ancestors with nothing failing.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -586,35 +641,49 @@ function collectFocused(node: DescribeNode, acc: DescribeNode[]): DescribeNode[]
...
+      if (!FOCUS_REPORTING_SOURCES.has(source)) return "unreported";
+      read = true;
       const target = flowSelectorToFrame(tree, into) ?? tappedFrame;
-      if (collectFocused(tree, []).some((n) => framesOverlap(n.frame, target))) return;
+      if (collectFocused(tree, []).some((n) => framesNest(n, target))) return "confirmed";
     } catch {
       // transient describe failure — retry until the deadline
     }
-    if (Date.now() >= deadline) return;
+    if (Date.now() >= deadline) return read ? "unconfirmed" : "unreadable";
```
latekvo, line 751: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: One successful read followed by an outage for the rest of the window scores as a hard app-verdict fail, not an environment error: `read` flips true on the first success (and never resets), so a tree source that answers once and then disconnects (devtools dropping mid-run) produces `unconfirmed` → "did not take keyboard focus", which CI reads as an app regression, rather than the `indeterminate` outcome the all-reads-throw case correctly gets. `settleTree` prices exactly this hazard with its minimum-reads floor (its comment explains how one slow read can consume the whole window); this loop has no equivalent floor, so the env-vs-app boundary here depends on whether the single lucky read happened to land before the field's focus round-trip completed.

packages/docs/docs/reference/flow-yaml.mdx
```diff
@@ -143,23 +143,23 @@ On iOS and Android, an id absent from `describe` can still resolve in a flow. On
...
+| `tap`        | `<selector>`, `{ on: <selector>, times }` or `{ x, y }`            | Tap an element. `times: 2` double-taps. `x`, `y` are normalized coordinates                                                                                                                                                    |
+| `long-press` | `<selector>` or `{ on: <selector>, duration }`                     | Press and hold. `duration` in milliseconds                                                                                                                                                                                     |
+| `type`       | `{ into: <selector>, text, submit }`                               | Focus the element and type. Presses Enter unless `submit: false`. Fails without typing if the field never takes keyboard focus (one retry tap); scores an environment error if the UI tree can't be read during the focus wait |
```
latekvo, line 152: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: This guarantee is stated unconditionally, but it does not hold on Vega: Vega's flow tree source is `vega-automation`, which is not in `FOCUS_REPORTING_SOURCES` (flow-actions.ts:215-219), so `waitForFocus` returns "unreported" immediately and `runType` types with zero focus confirmation — no retry tap, no failure. Vega is a first-class flow platform elsewhere in this very table (the `launch` row), so as written the reference promises behavior the runner does not provide there. Note live-authoring.md scopes the same statement correctly to iOS/Android/Chromium.

packages/skills/skills/argent-create-flow/references/live-authoring.md
```diff
@@ -257,7 +257,7 @@ Run `flow-execute` on the complete YAML with the absolute project root. For a fr
...
+Manual rescue invalidates the pass. An `errored` step was never evaluated: an `idle` wait whose tree source could not be read, a step that threw, an unresolvable `run:` target, or a `launch:` that did not start the app — plus a `type:` whose field never took focus while the UI tree was unreadable, which scores an environment error rather than an app verdict. Read the reason — most name the environment, but a failed `launch:` is a verdict about the app. Unconfirmed focus IS in the failure class: the replay focus wait fails a `type:` step whose focus was never confirmed after one retry tap, with nothing typed, so a green `type:` means the field held keyboard focus.
```
latekvo, line 262: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: "a green `type:` means the field held keyboard focus" overstates what the code guarantees. A green `type:` also results when (a) the tree source cannot report focus at all — Vega types blind, see the flow-yaml directive table; (b) the focused element is an editable shape the role/editable check cannot see — reproduced at this head with a focused `android.widget.AutoCompleteTextView` ancestor covering the typed-at element behind it; and (c) the closed-shadow-host residual documented in the code itself. The Typing section above keeps the committed-value check as the proof the value is right; this sentence drops that hedge and presents focus as settled fact, so an author following it will trust a green step that the runner's own gate cannot fully vouch for.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of aa9e35fb. Re-confirmed by reading `deriveUiAutomatorRole` (uiautomator-parser.ts:129-130): `edittext`/`textinput` are tested before `textview`, and `autocompletetextview` contains `textview` but neither `edittext` nor `textinput`, so it maps to `StaticText` -> `isEditable` false -> the focused-ancestor arm of `framesNest` vouches and the keys land in the autocomplete. The surface is wider than the one named class: `MaterialAutoCompleteTextView` (the widget behind Material's exposed-dropdown `TextInputLayout`, common in real apps), `MultiAutoCompleteTextView`, and `AppCompatAutoCompleteTextView` all contain `textview`; `SearchView$SearchAutoComplete` contains neither token and falls through to the literal class-name branch, still outside `EDITABLE_ROLES`. Android never emits the `editable` flag, so `role` is the only signal and the hole is unmitigated on all of them.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of aa9e35fb. The suite exercises only `EditText` (-> `TextField`), `textbox`, and `combobox`, so deleting `input`, `textarea`, `searchbox` or `spinbutton` from the alternation leaves it green. `axtextfield` is unpinned the same way and is the iOS-critical one: `UITextField`/`UITextView`/`SearchField` -> `AXTextField` is iOS's only editable role and reaches `framesNest`'s ancestor arm through the `native-devtools` source, yet no test places an `AXTextField` in that arm (the one AXTextField fixture matches by identity, the `frameWithin(f, target)` branch, not the ancestor branch). So the entire iOS editable-ancestor path — the O1 mechanism on iOS — could be dropped from the regex with nothing failing.

packages/tool-server/src/tools/flows/flow-actions.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of aa9e35fb. `waitForFocus` sets `read = true` on the first successful in-set read and never resets it (flow-actions.ts:745), and the deadline arm returns `read ? "unconfirmed" : "unreadable"` (:751). So a source that answers once and then throws for the rest of the window returns `unconfirmed` -> the hard "did not take keyboard focus" app-fault fail, not the `unreadable`/indeterminate tier the all-reads-throw case gets. There is still no minimum-reads floor here, unlike `settleTree`'s `SETTLE_MIN_READS`, so the env-vs-app boundary turns entirely on whether the single lucky read happened to land before the focus round-trip completed.

packages/docs/docs/reference/flow-yaml.mdx
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Revisited this at aa9e35fb and it does not reproduce. `dispatchDirective` rejects a `type` step on Vega up front (flow-actions.ts:930-943: `platform === "vega" && step.kind === "type"` -> `ok: false` with the reason "type is a touch directive and Vega is remote-driven", pointing the author to `tool: keyboard` instead), so a Vega `type:` never reaches `runType`/`waitForFocus` and never hits the `unreported` path — it fails without typing, which is exactly what the row promises. `unreported` is in fact unreachable for every source that can reach `waitForFocus`: iOS `native-devtools`, Android `android-devtools` and Chromium `cdp-dom` are all in `FOCUS_REPORTING_SOURCES`, Vega is blocked at dispatch, and `ios-remote` throws at `waitForFrame` first. So the type row is accurate as written and the Vega blind-typing concern is not a live path.

packages/skills/skills/argent-create-flow/references/live-authoring.md
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: The sentence still overstates the guarantee at aa9e35fb, but the standing counterexample is (b), not (a): a Vega `type:` cannot produce a green step at all — it is rejected up front at `dispatchDirective` (flow-actions.ts:930-943), so it never types blind. The live way a green `type:` occurs without the target holding focus is the misclassified editable ancestor from the O1 thread — a focused `AutoCompleteTextView` (role `StaticText`, `isEditable` false) vouches for a covered target and the keys land in the autocomplete. So the hedge this sentence drops is warranted, just for the O1 reason rather than the Vega one.

packages/tool-server/src/tools/describe/platforms/chromium.ts
```diff
@@ -431,6 +435,22 @@ const buildDescribeDomScript = ({ maxDepth, maxNodes }: ChromiumWalkLimits) => `
...
+    // isContentEditable lives on HTMLElement.prototype without
+    // [LegacyLenientThis], so calling it with an SVGElement/MathMLElement
+    // receiver throws "Illegal invocation" and aborts the whole walk — and
+    // inline SVG is ubiquitous. SVG elements are never contenteditable, so
+    // gating on the receiver's type is semantically exact.
+    if (
+      (typeof HTMLElement === "undefined" || el instanceof HTMLElement) &&
+      getIsContentEditable.call(el)
+    ) {
+      node.editable = true;
```
latekvo, line 452: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: This new `editable` flag never reaches the focus rule for the one shape the docblock names it for — a bare `<div contenteditable>` — when that div has a single element child. The boxed structural collapse above (chromium.ts:413-424) returns `childResults[0]` whenever `depth > 0 && childResults.length === 1 && !clickable && !name && !text && !id && !scrollable && role === "div"`, and a plain `<div contenteditable><p>note</p></div>` matches it: no ARIA role -> role `"div"`, no id/testID, and its own direct text is empty (the text lives in the `<p>`). So the walk returns the `<p>` before this block and before the `focused` block below ever run. `activeElement` is the editing-host div, not the `<p>`, so `collectFocused` finds nothing focused, `waitForFocus` returns `unconfirmed`, and `runType` retries once then fails the step ("did not take keyboard focus") on a genuinely-focused, typeable editor. The collapse predates this PR, but before the new gate `type:` typed into that editor regardless; the gate turns the dropped `focused` into a hard failure. Editors that set `role="textbox"` or a testID escape the collapse — the bare single-paragraph case does not, which is exactly the shape the flag was added to support.

packages/tool-server/src/tools/describe/platforms/chromium.ts
```diff
@@ -72,6 +72,10 @@ const buildDescribeDomScript = ({ maxDepth, maxNodes }: ChromiumWalkLimits) => `
   const getAttr = Element.prototype.getAttribute;
   const hasAttr = Element.prototype.hasAttribute;
   const getBCR = Element.prototype.getBoundingClientRect;
+  // The test harness's mock DOM defines no HTMLElement constructor; degrade to
+  // a direct read (undefined on mocks) instead of throwing at script top.
+  const htmlProto = typeof HTMLElement === "undefined" ? {} : HTMLElement.prototype;
```
latekvo, line 77: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: This comment's premise is falsified by this same PR. The only harness that evals `DESCRIBE_DOM_SCRIPT` (describe-chromium-script.test.ts) now defines `HTMLElement` — `g.HTMLElement = MockHTMLElement`, plus the save/restore around `run` — and real Chromium always defines it. So `typeof HTMLElement === "undefined"` is never true here: `htmlProto` is always `HTMLElement.prototype`, the `{}` fallback is dead, and the `typeof HTMLElement === "undefined" ||` short-circuit in the editable guard below never fires either. The adjacent `Document` guard is the correct model — the harness never sets `g.Document`, so that `{}` branch really is exercised; this one is not, and the "mock DOM defines no HTMLElement constructor" rationale no longer holds.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -607,6 +608,60 @@ function framesOverlap(a: DescribeFrame, b: DescribeFrame): boolean {
...
+ * wrapper focused rather than the field, and those wrappers come in every size
+ * (a shadow-DOM host, a tappable form row), so no size bound separates them
+ * from coverers — any numeric cutoff refuses genuinely-typeable fields. What
+ * bounds the risk instead is that the ancestor cannot itself be the key-focus
+ * holder: for its flag to mean anything, some descendant took focus, and the
+ * tap that started this wait was aimed at the target's centre.
+ */
+function framesNest(focused: DescribeNode, target: DescribeFrame): boolean {
+  const f = focused.frame;
+  if (frameWithin(f, target)) return true;
```
latekvo, line 660: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: The descendant arm confirms focus for any focused node inside the target without the editability check the ancestor arm applies two lines down. A `type: { into: <container> }` whose tap lands on a focusable non-input inside it — a `<button>`, `<a href>` or `[tabindex]` element, which becomes `document.activeElement` with `focused: true`, a role outside `EDITABLE_ROLES`, and no `editable` flag — makes `frameWithin(f, target)` true, so `runType` reports `confirmed` and injects the keys, and the step returns `ok: true` though no text field ever held focus. That is the same "keys never reached a field, green anyway" class the ancestor arm was tightened against; the two arms disagree on whether a non-editable focused element counts as evidence.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Ran a fresh full sweep against head aa9e35fb: claims vs code, nearest twin, non-happy paths, inputs, reachability both ways, what outlives the call, plus the absence pass (siblings / symmetry / stale prose, and a discrimination trace over the new branches). Built the head and ran the flow-type-focus + describe-chromium-script suites (48/48 green) — prior rounds verified at 9ce2055c, before the base-branch merges, so this re-confirms them on the current head. The five open threads from the last round are re-confirmed or corrected on their own threads rather than re-filed (one — the Vega type-row concern — does not reproduce, corrected there). Three new low-severity items below.

PR #526, opened 2026-07-20
fix(flow): add a visual settle step

j-piasecki:
## Summary

Flow actions now wait for both the UI hierarchy and rendered pixels to stabilize before resolving coordinates or capturing snapshots. This covers transitions whose presentation-layer motion is invisible to the tree, such as iOS navigation and modal animations.

This surfaced in [react-native-gesture-handler#4318](https://github.com/software-mansion/react-native-gesture-handler/pull/4318), whose Argent E2E test often failed after dismissing a modal unless the flow included explicit `wait:` steps.

## What changed

- Added a bounded combined settle: stable tree → stable pixels → final tree revalidation.
- Made `scroll-to` visually settle once before scrolling, then use tree-only checks for subsequent momentum-free increments.
- Kept raw-coordinate flow actions tree/devtools-gated and clarified the related docs and authoring guidance.
- Added coverage for motion, capture/tree timeouts, aborts, unavailable sources, scroll behavior, and snapshot degradation.

## Validation

- Flow suite: 25 files, 408 tests passing.
- Relevant typechecks, ESLint, Prettier, and skill validation passing.

PR #728, opened 2026-08-04
feat(flow): add an idle condition that waits for the screen to stop moving

hubgan:
> Stacked on #574. Replaces the route-fingerprint approach that #727 and #729 carried; both are closed.
>
> Rebased off the #715-#726 chain so it can merge on #574 alone. That chain owns `DirectiveOutcome.warning` / `StepReport.warning` and the `indeterminate` scoring rule; the small amount of each that this condition needs is carried here, scoped to `idle`, and will conflict textually with #724/#725 for whichever lands second.

## One check a selector cannot express

```yaml
- await: { idle: true }
```

Has the screen stopped moving? Flows were substituting fixed `wait:` steps for it, which either wait too long on every run or too little on the run that mattered.

## Why it reads pixels as well as the tree

Each signal is blind to what the other sees.

The tree cannot see presentation-layer motion. An iOS push or modal dismissal commits its hierarchy up front and then animates a layer for a few hundred milliseconds; a cross-fade or a scrim moves no node at all. Measured on a real Android emulator: a page animating a large element continuously reads as a single static `WebView` node, and the existing `await-screen-idle` tool called it settled in **465 ms, five times out of five**. On a deterministic Chromium page whose DOM is byte-identical across the animation, that tool returned `settled: true` in **407 ms** while this condition failed 8/8.

Pixels alone would not do either: they cannot see a tree still churning behind an unchanged surface, and anything animated forever (a video, a shimmer) would make a pixel-only settle unsatisfiable on a screen the tree calls ready.

Unlike that tool, this cannot be silently stepped over. Readiness is not an acceptance criterion — the verdict belongs to the identity and outcome checks around it — so a screen that never settles still **passes**, carrying a `warning` that says what the green actually bought. That is what makes it safe to persist in a flow: it reports on every run without ever failing one over a video or a shimmer. The only thing that stops a run is a tree source that cannot be read.

## The rule the design follows

The absence of evidence is never evidence.

- A capture that did not arrive cannot complete the hold.
- Neither can a single agreeing pair. Every animation that reverses has a turning point, and two samples straddling it come back identical while the screen is moving. On a live 3 s cross-fade that passed a default-shaped step on roughly **one run in three**; settling now takes two consecutive still intervals, so `minStableMs: 0` still means three reads.
- A round is never started without the budget to observe it with. A round begun with nothing left neither captures nor reads, and both absences were otherwise recorded as facts about the device.
- No verdict comes from a latch. A screen that settles and then moves again, or goes blank, has not settled.

## `timeout:` is a real bound

No describe path takes an abort signal, so the tree read is raced against what is left of the budget. Without that, a wedged ViewInspector RPC ran 2.25 s past an 8000 ms budget, and a busy Android screen ran 5.4-6.0 s against a 3000 ms one. A read that runs out of step budget is the step ending, not the source failing, and only the latter is reported as an environment problem.

## Failure modes stay apart

They call for opposite responses. A screen that never stopped moving is a verdict about the app. A tree that could not be read is not, and it names the foreground check first, because a backgrounded app reads identically to an uninstrumented one and relaunching is the wrong repair. A run whose captures never produced a comparable pair still settles on the tree and says so in a warning, rather than passing off half a proof as the whole one.

## Capture routing and tolerance

Captures route exactly as the `screenshot` tool routes them, so tvOS and Vega, which have no simulator-server backend but are perfectly screenshottable through `xcrun` and the emulator console, are covered rather than written off.

The pixel comparison owns its tolerance rather than borrowing `screenshot-diff`'s. That one holds a baseline stored across sessions, machines and OS versions against a live capture and must absorb real drift; this one holds two captures one poll apart from a single session, where a static screen reads back byte-identical (measured: zero changed pixels across five consecutive pairs on an idle simulator). The margin is load-bearing, because uniform change is all-or-nothing: at the baseline tolerance any cross-fade slower than ~2 s counted zero changed pixels and read as settled mid-animation.

## Not a screen check

A dropped tap leaves the source screen perfectly idle. It belongs after the element check that names the destination, never instead of one. #730 documents the pairing.

## Verification

- 3307 tests pass; typecheck, `typecheck:tests`, lint and prettier clean. Eleven mutations applied to the source, all eleven caught by tests.
- Driven end to end against a real iOS simulator, a real Android emulator, and a Chromium app through a tool-server built from this branch: static screens pass with no warning (proving captures ran), animating screens pass carrying the warning that names the motion, and the pre-existing tool disagrees exactly where it should.


j-piasecki:
I think this fills the same role as https://github.com/software-mansion/argent/pull/526, no? In which case, it would be good to settle on one approach.

hubgan:
> I think this fills the same role as #526, no? In which case, it would be good to settle on one approach.

@j-piasecki Yeah there's an overlap. Both PR's try to fix the "has the screen stopped moving?" question. Especially for iOS devices. And the #526 is the better approach in my opinion. Additionally this and #727 answer the "Which screen is this?" question. I wonder if that's something worth keeping or if we should close both #727 and #728 PR's.

j-piasecki:
I'm actually not sure about #526 - on constantly animating screens, it effectively adds 5 seconds of delay to every step. An explicit, surgical step to await idle may make more sense.

Screen fingerprinting is also nice, but I don't like the fact that it only works for React Native apps in debug mode. E2E tests are usually run in the release variant.

j-piasecki (commented):
Adversarial pass over the whole branch at `f4c4ce6` — read every changed file, then probed `waitForIdle` with throwaway runner tests (deleted afterwards). `tsc --build`, eslint, prettier and the full tool-server suite (311 files / 3613 tests) are green.

Most of what I chased did not survive scrutiny, and I am deliberately not posting it. For the record, two that looked like findings and are not:

- **The Vega capture not taking the round's abort signal** (while the tvOS arm does) is correct as written: `runAdb` passes `timeoutMs: 20_000` with a `killSignal`, so the subprocess self-terminates, whereas `tvScreenshot`'s `execFileAsync` has no timeout at all and is genuinely unbounded without the signal. The asymmetry is the right call, not an oversight.
- **The `layoutViewport` fallback in `chromiumScrollOffset`** (device pixels used as a CSS-pixel clip origin) is unreachable on any Chrome this repo boots, since `cssVisualViewport` has been present for years.

One finding survived.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1117,6 +1151,409 @@ async function waitForCondition(
...
+        `the tree source never answered within the step's ${timeoutMs}ms — raise this step's ` +
+        `\`timeout:\` if it is merely slow (a tree read on a busy screen can take seconds), or ` +
+        `repair it if it has stopped answering altogether`,
+    };
+  }
+  // Reads worked, then stopped: a backgrounded app, a dropped instrumentation
+  // session. One early success does not license a verdict drawn from a window
+  // that went dark afterwards. (A read that merely ran out of budget is NOT
+  // this case — it is the step ending, and the evidence below still stands.)
+  if (lastRead === "error" && treeErrorMessage !== undefined) {
```
j-piasecki: **A single blip on the trailing poll turns every benign `idle` outcome into a run-stopping `error`.**

This branch fires on `lastRead === "error"` alone — one failed read, the last one — with no tolerance for how long the source was actually dark.

`waitForCondition` (line 1019), in this same file, deliberately does not do that. It keeps `lastTrustedReadAt` and only calls the window indeterminate once the dark tail exceeds `CONDITION_DARK_TAIL_TOLERANCE_MS` (line 242), precisely because "a fetch error on the trailing poll must not flip a clean skip into a hard error" (line 1100). `waitForIdle` has the *stricter* consequence — `error`, which sets `state.stopped` and skips every later step — and the *weaker* guard.

The exposure is exactly the screens this PR says are the common benign case. A screen that never settles always exits through the bottom of the loop, so the last read is always the one that decides. Measured against this branch:

| scenario (900ms timeout, `minStableMs: 300`) | outcome |
|---|---|
| never-settling tree, one `throw` on the read at t≥700ms | `status: "error"`, `ok: false`, later steps `skip` |
| **the identical blip 400ms earlier** | `status: "pass"` + "never held still" warning, run continues |
| blank tree (`children: []`) throughout + one blip on the trailing poll | `status: "error"` — instead of the "UI tree stayed empty" warning |

So a video, a shimmer, a carousel, Android live-updating text, or a canvas/video surface — all of which this PR is explicit about wanting to pass with a warning — turn a run red if one `describe` call blips on one particular poll. Whether the flow survives is decided by *when* the transient lands, which is timing luck.

Two things suggest this is not intended rather than a deliberate trade:

- `waitForIdle`'s own docstring says "Only an unreadable window is a hard stop". A window that answered on every poll but the last is not an unreadable window.
- The comment above this branch — "Reads worked, then stopped: a backgrounded app, a dropped instrumentation session" — describes a *sustained* failure, but the code fires on a single one. And `flow-idle-run.test.ts:691` ("restarts the hold after a failed read, and still settles") already pins the opposite behaviour for a blip one poll earlier, so the tolerated/fatal split is purely positional and nothing documents it.

The fix looks small: the loop does not currently keep the timestamp of the last read that *carried content* (`treeSince` is close, but it resets whenever the signature changes). With one more timestamp, this branch can gate on the dark tail the way tier 2 does, and let a single trailing error fall through to the determinate warning with the error appended — which is exactly what `waitForCondition` does with its `blipNote`.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1117,6 +1151,460 @@ async function waitForCondition(
...
+ * transient one poll earlier passed with a warning: whether a flow survived a
+ * screen this step is explicit about wanting to pass came down to where the
+ * blip landed.
+ *
+ * Two polls is what one blip costs: up to a poll of sleep since the last read
+ * that answered, plus a poll's worth of latency for the failing one. A longer
+ * tail means consecutive reads went dark, which is the window this step cannot
+ * describe.
+ */
+const IDLE_DARK_TAIL_TOLERANCE_MS = IDLE_POLL_MS * 2;
```
j-piasecki: **One blip costs a poll of sleep plus a whole round, and nothing bounds the round to a poll — so the pre-fix split returns whenever a round runs long.**

`lastAnsweredReadAt` advances only when a read resolves, and a round is `Promise.all([settleWithin(fetchFlowTree…), capturePixelsWithin…])` (line 1342). The gap it anchors is therefore `sleep + max(read, capture)`, and the tolerance leaves exactly `IDLE_POLL_MS` of it for the round. Neither half is held there: `capturePixelsWithin` grants the capture up to `PIXEL_CAPTURE_TIMEOUT_MS` (2000ms, `flow-pixels.ts:134`) and the tree read gets the whole remaining budget.

Measured in this repo's runner — never-settling tree, `await: { idle: true, timeout: 3000, minStableMs: 300 }`, exactly one `throw` from the tree source, `capturePixelsWithin` stubbed to take 300ms (bounded by its own budget, as the real one is). 6 rounds:

| blip lands on | step | run | later steps |
|---|---|---|---|
| the last round | `error` — "could not read the UI tree while waiting for the screen to settle…" | `ok: false` | `skip` |
| one round earlier | `pass` + "never held still" | `ok: true` | `pass` |

5/5 trials each. With the capture returning instantly, both positions pass — that is the case `flow-idle-run.test.ts:719` pins.

`SKILL.md` states the tolerated case without the condition: "a read that fails at the very end of the wait is named in the warning rather than stopping the run."

j-piasecki (commented):
Re-checked the trailing-blip finding against `1c843d80c`. Cleared:

- All three scenarios from the earlier comment now pass and carry the failed read in the warning; two consecutive failing reads still `error` and still skip later steps.
- `lastAnsweredReadAt === 0` cannot reach the dark-tail check — it and `readsSucceeded` are set in the same branch, and the `readsSucceeded === 0` guard returns first.
- The parenthetical on `blipNote` holds: the error branch clears `treeSettledAtLastRead` (line 1372), so the tree-only settle is genuinely unreachable with a failed final read.
- Full suite 317 files / 3711 tests pass; `tsc --build`, `typecheck:tests`, eslint and prettier clean.

One comment on the tolerance's derivation.

j-piasecki (approved):

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat)\]: Full end-to-end sweep of this branch at `d61b8395`, over three adversarial rounds.

**What I ran.** Built the branch in a clean worktree (`npm ci`, `tsc --build`, both green) and drove `await: { idle: true }` end to end through the real entry point — a branch-built tool-server, an Electron/Chromium device, and `flow-execute` on real flow files. All four documented outcomes reproduce: a static screen passes with no warning (so the captures genuinely ran), an animating screen passes carrying the motion warning, a blank page passes carrying the empty-tree warning, and killing the CDP-owning browser mid-step produces the `error`, `ok: false` and skipped later steps. Warnings propagate correctly through a nested `run:` sub-flow and render in both the CLI and the MCP content renderer. The flows suite is green in a clean worktree: 42 files / 1001 tests.

Everything below was then reproduced in this repo's own runner harness (`test/flows/flow-idle-run.test.ts`, stubbing only `fetchFlowTree` and `capturePixelsWithin`), each with a control run that isolates the single variable.

I read every existing comment and thread on this PR first. The two `waitForIdle` findings already raised — the single trailing blip, and the millisecond-based dark-tail tolerance — are fixed on this head and I am not restating them; the four inline comments below are separate mechanisms, two of which sit next to those guards rather than in them.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1117,6 +1151,467 @@ async function waitForCondition(
...
+  // Reads worked, then stopped: a backgrounded app, a dropped instrumentation
+  // session. One early success does not license a verdict drawn from a window
+  // that went dark afterwards. (A read that merely ran out of budget is NOT
+  // this case — it is the step ending, and the evidence below still stands.)
+  //
+  // Measured as a tail, not as a single read: the source failing on the last
+  // poll and the source having stopped answering are different windows, and
+  // only the second is unreadable. See IDLE_TOLERATED_DARK_READS.
+  if (
+    lastRead === "error" &&
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: **The dark-tail guard is keyed on `lastRead === "error"`, so a closing round that ends as a `timeout` throws the whole tail away — the accumulated `darkReads`, the error message, and the `blipNote` at line 1536 with it.**

A budget-starved closing round is routine rather than exotic, and the loop's own exit rule produces it: `if (left < MIN_ROUND_BUDGET_MS) break` lets a round through whenever `left >= 200`, then sleeps 200, so the next round begins with `left - 200`. Whenever `left` lands in `[200, 200 + read_latency)` that read is abandoned, `lastRead` flips from `"error"` to `"timeout"`, and `treeReadHung` stays false because the round had well under `HUNG_TREE_READ_MS` of budget. Nothing downstream looks at `darkReads` again.

Measured with this repo's own runner harness (`test/flows/flow-idle-run.test.ts`; only `fetchFlowTree` and `capturePixelsWithin` stubbed, real `settleWithin`, real `waitForIdle`). Tree source answers twice and then throws `native-devtools is not connected` on every later read, each read taking 150ms to fail; capture instant:

| step | outcome |
|---|---|
| `await: { idle: true, timeout: 1700, stableFor: 0 }` | `error`, `ok: false`, later steps `skip`, error named in the reason |
| **the same dead source at `timeout: 1500`** | `pass`, `ok: true`, later steps run, the read failure named nowhere in the report |

5/5 trials on each, with the control run in the same file. `darkReads` had reached 2 in the passing case and 3 in the erroring one; the only difference is which side of `[200, 350)` the closing `left` fell on. Sweeping `timeout` 1400→1800 in 50ms steps, 4 of the 9 land on the passing side.

The hit rate is `read_latency / (IDLE_POLL_MS + round_duration)`, so it scales with how long a failing read takes to come back. I measured ~10ms for the Chromium path (`ECONNREFUSED` on a dead CDP port), but the iOS and Android tree sources fail through an RPC or an adb round trip rather than a refused local socket.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1117,6 +1151,467 @@ async function waitForCondition(
...
+    treeErrorMessage !== undefined &&
+    darkReads > IDLE_TOLERATED_DARK_READS
+  ) {
+    return unreadable(treeErrorMessage);
+  }
+  // The same window going dark the other way: the source answered, then stopped
+  // answering with seconds of budget still in hand. A failing read has a
+  // dedicated error above, but a HANGING one used to fall through to the
+  // motion warning and tell the author a frozen screen was a carousel.
+  if (lastRead === "timeout" && treeReadHung) {
```
latekvo, line 1652: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: **`treeReadHung` cannot separate a wedged source from a merely slow one, so a healthy device hard-stops the run on a screen that is perfectly still.**

`treeReadHung` is set whenever a read is abandoned in a round that happened to have `HUNG_TREE_READ_MS` of budget. A source that simply takes longer than 2s per read produces exactly that, and the comment at line 1218 names that case as healthy — "a tree read on a busy screen genuinely takes seconds, and Android's `uiautomator dump` allows itself twenty". The PR description measures Android reads at 5.4-6.0s.

Measured with the runner harness, **static screen, every read answering with the same tree** — the most settled a screen can be — varying only the read latency:

| read latency | `timeout` | outcome |
|---|---|---|
| 2500ms | 7500 (default) | `error` "the UI tree source answered and then stopped", `ok: false`, later steps `skip` |
| 2500ms | 6000 | `pass` + warning |
| 3000ms | 6000 | `error`, run stopped |
| 3000ms | 7500 | `pass` + warning |
| 4000ms | 7500 | `error`, run stopped |

At the PR's own measured Android latency the errors dominate: over 5400/6000ms reads × `timeout` 8000/10000/12000/15000/20000, 6 of the 10 combinations hard-error a static screen and skip every later step.

Raising `timeout:` is the remedy SKILL.md prescribes ("that last one may simply be slow: raise the step's `timeout` before suspecting the app"), but it is not monotonic here — at 5400ms reads, 12000 passes, 15000 errors, 20000 passes. Whether a healthy run goes red is decided by where the deadline happens to fall relative to the read.

This is the step documented as one that never fails a run except for a source that cannot be read; in every row above the source answered every single read it was given time to finish.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1117,6 +1151,467 @@ async function waitForCondition(
...
+      stillIntervals = 0;
+    } else {
+      lastRead = "value";
+      readsSucceeded += 1;
+      darkReads = 0;
+      treeErrorMessage = undefined;
+      // It answered, so whatever wedged it has cleared.
+      treeReadHung = false;
+      const tree = read.value.tree;
+      if (tree.children.length === 0) {
```
latekvo, line 1480: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: **An empty tree carrying `hint` / `should_restart` is a degraded read, and this branch scores it as an observation about the app.**

`waitForCondition`, at line 1052 of this same file, treats `data.tree.children.length === 0 && Boolean(data.hint || data.should_restart || everMatched)` as blind and refuses to draw a verdict from it. This branch keys only on `children.length === 0`, so the same read clears `darkReads`, counts toward `readsSucceeded`, and ends in the `!sawContent` warning at line 1546.

`flow-vega-tree.ts:72` describes precisely this shape and states the intent it relies on: an unattached toolkit "surfaces as `describeVega`'s empty tree + relaunch hint, passed through unchanged so the runner's blind-read guard still sees the hint." There is no such guard on this path.

Driven through the real runner with `fetchFlowTree` returning `{ tree: { children: [] }, source: "vega-automation", hint: "the automation toolkit is not attached — relaunch the app so it can attach" }` and never throwing. Same read, two conditions, same file:

- `await: { visible: { id: … } }` → `fail`, `ok: false`, later steps `skip`, reason "could not evaluate the condition — every read of the UI tree was empty or degraded".
- `await: { idle: true }` → `pass`, `ok: true`, later steps run, warning "the UI tree stayed empty for 1200ms — the screen never rendered content, so there was nothing to settle. If the screen is meant to render accessible content, this is where it did not".

The second sentence is a verdict about the app drawn from a window that was never readable, and it points the author at their own screen rather than at the toolkit that never attached.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1117,6 +1151,467 @@ async function waitForCondition(
...
+        treeSettledAtLastRead =
+          treeStillIntervals >= MIN_STILL_INTERVALS && now - treeSince >= stableFor;
+
+        // A missing frame is the ABSENCE of visual evidence, never evidence of
+        // stillness. Letting it stand in for "the pixels held" is what turned a
+        // screen that never stopped moving into a pass.
+        let pixelsHeld = false;
+        let localizedThisInterval = false;
+        if (frame === undefined) {
+          captureFailed = true;
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: **`captureFailed` is a latch that is never cleared, so a single missed capture anywhere in the step routes a fully-verified pixel settle into the "could not be screenshotted" warning — and the reset two lines below makes the step burn its whole timeout getting there.**

Two things compound on a missed frame. `captureFailed = true` here is never reset (it is declared at 1332 and read at 1593 and nowhere else). And the `else` arm at 1458-1462 sets `bothSince = now; stillIntervals = 0`, so a missing frame does not merely fail to extend the combined hold, it destroys it — while the nearest twin in the same loop makes the opposite choice for the same class of event: an abandoned tree read "is the absence of an observation, not an observation … so the hold state is left as it was" (line ~1360).

Measured with the runner harness, **perfectly static screen** (constant tree signature, byte-identical 300x600 frames), the only variable being whether the capture backend misses:

| capture behaviour | elapsed | captures | result |
|---|---|---|---|
| never misses (control) | 1217ms | 7 ok / 0 missed | `pass`, no warning |
| misses every 4th round | **12002ms** (the full `timeout`) | **15 ok** / 5 missed | `pass` + "settled on the UI tree alone — this screen could not be screenshotted on enough polls to compare a pair of them, so animation that moves pixels without moving nodes (a push, a fade, a dismissing modal) was not waited out" |

Step: `await: { idle: true, stableFor: 1200, timeout: 12000 }`. Fifteen captures succeeded and every compared pair read still, so the sentence the author is handed is false about the run that produced it, and it points them at the capture backend as the thing that did not run.

The comment at 1587-1592 says this branch is unreachable "with captures arriving" — captures were arriving on three polls out of four. Only steps asking for a hold longer than the gap between misses are affected; the default 250ms hold spans about two polls and rides over a single miss.

packages/tool-server/src/tools/flows/flow-pixels.ts
```diff
@@ -0,0 +1,376 @@
...
+ * The fraction of the frame {@link comparePixels} must ignore for this device,
+ * because the system paints it and the app does not.
+ *
+ * Only iOS and Android put a status bar in the capture. A Chromium window's top
+ * band is page content, and Vega and tvOS render full-screen with no system
+ * chrome, so masking any of those would blind the check to real motion for
+ * nothing. tvOS shares iOS's platform tag and is only distinguishable by the
+ * runtime probe, which is memoized per UDID and already paid by the capture.
+ */
+export async function statusBarMaskFraction(device: ActionEnv["device"]): Promise<number> {
```
latekvo, line 183: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: **The platform enumeration here misses two platforms that reach it, in opposite directions.**

`device.platform === "android"` is true for Android TV / leanback as well as for phones, so 6% of every Android TV capture is dropped from both the changed count and the total — a target that renders full-screen with no system chrome, exactly the reason Vega and tvOS are excluded three lines down. `await-screen-idle`, the tool this check is positioned against, resolves it: `const androidIsTv = device.platform === "android" && (await isAndroidTv(device.id))` (`tools/await-screen-idle/index.ts:152`), using `isAndroidTv` from `utils/adb.ts:494`.

In the other direction, `device.platform !== "ios"` sends `ios-remote` down the no-mask arm, so a remote simulator's live status bar is compared as app content. `classifyDevice` (`utils/device-info.ts:49`) returns `"ios-remote"` for any `remote:`-prefixed udid, and flows treat it as a first-class iOS target — `flow-run.ts:1839` folds it explicitly, with the note that "'ios-remote' is an iOS simulator driven through sim-remote — for a platform guard it IS ios … without this fold no guard could ever match on a remote sim". `utils/status-bar.ts` covers only `"ios"` and `"android"`, so the run-level `pinStatusBar` does not cover it either.

The doc comment above enumerates which platforms need the mask and why, and reads as complete for both.

packages/tool-server/src/tools/flows/flow-pixels.ts
```diff
@@ -0,0 +1,376 @@
...
+//
+// The floor is not zero because a capture pair is not guaranteed byte-identical
+// on every backend; two captures of a static iOS screen changed 0 pixels of
+// 237k. Ten sits an order of magnitude under the smallest indicator measured
+// and clear of that noise. One smaller than ten pixels stays invisible, which
+// is the residual limit of comparing whole frames.
+//
+// MOTION_FRACTION above stays a fraction: a transition, a scroll or a
+// carousel moves a share of the screen, which is what scales with it.
+const LOCALIZED_MOTION_MIN_PIXELS = 10;
```
latekvo, line 143: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: **The floor here became a count while the ceiling above it stayed a fraction, so the `localized` band shrinks with the frame — and the recorded justification for the change does not hold arithmetically.**

The stated reason for a count is that an indicator's size in captured pixels does not scale with the frame. `MOTION_FRACTION` directly above did not change, so the band is `[10, 0.002 * total)`: its floor is frame-independent and its ceiling is not. Below `total = 5000` captured pixels the band is empty outright, and it is too narrow to hold a real indicator well before that.

Run against the built `comparePixels` from this branch, chromium mask 0, `CAPTURE_SCALE` 0.25, one fixed-size indicator:

| window (CSS, dpr 1) | frame | band | 22 changed px | 38 changed px |
|---|---|---|---|---|
| 400x300 | 100x75 | [10, 15.0) | `moving` | `moving` |
| 500x400 | 125x100 | [10, 25.0) | `localized` | `moving` |
| 700x520 | 175x130 | [10, 45.5) | `localized` | `localized` |
| 1200x800 | 300x200 | [10, 120.0) | `localized` | `localized` |

Same indicator, opposite verdict by window size alone. `moving` leaves `pixelsHeld` false, so the hold resets every round (1458-1462), the step runs to the deadline instead of returning when the screen is at rest, and the author gets "the screen never held still … a video, a looping animation, a carousel" for exactly the spinner this constant exists to report separately. That is the mirror of the large-window failure the comment says it fixed.

Separately, the numbers in that comment do not support it. The indicators listed are 50-66, 57 and 50 changed pixels, and the fraction-equivalent floor is then given as "~46 pixels, above every indicator measured here". 46 is below all three, so a 46-pixel floor would still have flagged every one of them. I could not reproduce 46 from the sizes quoted either: at the stated 0.005% (which is 10 / 198,112, the iPhone 16 Pro capture area) the 2400x2558 Chromium window this file measures elsewhere gives 19 pixels at `CAPTURE_SCALE`.

packages/tool-server/src/tools/flows/flow-pixels.ts
```diff
@@ -0,0 +1,376 @@
...
+  width: number;
+  height: number;
+  data: Buffer;
+}
+
+/** What {@link comparePixels} saw between two captures. */
+export type PixelChange = "still" | "localized" | "moving";
+
+// Hard downscale: motion detection only needs to see a region of the screen
+// change, and a quarter-scale frame decodes ~16x faster. Every route honours
```
latekvo, line 27: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: **"Every route honours it" holds for the output but not for the cost the sentence is justifying, and the Vega route is where that matters most.**

The stated reason for the downscale is that "a quarter-scale frame decodes ~16x faster". `captureVegaScreenshotPng` (`utils/vega-screen.ts:31-57`) reaches the quarter-scale frame the long way round: `PNG.sync.read` on the emulator's native-resolution PNG, a lanczos3 `resizeDecodedPng`, `PNG.sync.write` back to a temp file — and then `capturePng` here reads that file and decodes it a second time. So on Vega the full-resolution decode is paid on every poll anyway, plus a resample and a re-encode, all synchronous on the shared tool-server's event loop.

That is the same cost this file treats as a problem worth solving elsewhere: `captureChromiumPng`'s comment records "~79ms of every 200ms round" of blocking decode as the reason for taking the scale from the compositor instead of from `sharp`. Vega is the one remaining route that still pays it, and nothing in the comment flags it.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -1658,6 +1683,177 @@ function parseWaitFields(raw: unknown, kind: "await" | "assert" | "when"): WaitF
...
+  // default is what most steps run with, so leaving it out was the way to get
+  // an unsatisfiable step past the parser (`timeout: 100` was accepted while
+  // the identical `timeout: 100, stableFor: 250` was rejected).
+  const timeoutMs = step.timeout ?? IDLE_DEFAULT_TIMEOUT_MS;
+  const stableFor = step.stableFor ?? IDLE_DEFAULT_STABLE_FOR_MS;
+  const needed = stableFor + IDLE_SETTLE_OVERHEAD_MS;
+  if (timeoutMs < needed) {
+    badEntry(
+      entry,
+      `idle needs a timeout of at least ${needed}ms to hold still for ` +
```
latekvo, line 1841: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: **The cost model in this message (and in `IDLE_SETTLE_OVERHEAD_MS` and SKILL.md) is additive, but the runner measures the hold *across* the polls rather than after them, so satisfiable steps are rejected as impossible.**

`waitForIdle` sets `bothSince` on the first content round and settles when `stillIntervals >= MIN_STILL_INTERVALS && now - bothSince >= stableFor` — the hold clock starts at the first read and runs during the two polls; it is not appended to them. The true floor is therefore `max(2 polls, stableFor) + one closing round`, not `stableFor + 3 polls`.

`- await: { idle: true, timeout: 1000, stableFor: 800 }` is rejected here with "idle needs a timeout of at least 1400ms", and `- await: { idle: true, timeout: 700 }` with "at least 850ms". Running the same holds through the runner harness with the parser floor bypassed, both settle cleanly and well inside the timeout the parser calls impossible. The PR's own suite agrees: "holds a still screen for the whole requested hold before passing" uses `stableFor: 800` and completes in 805ms, and my own default-hold measurements land at 807ms for `stableFor: 800`.

The two agree exactly at `stableFor: 0` and diverge by up to 400ms above `stableFor: 400`. The over-demand itself is cheap, since `timeout:` is a ceiling rather than a duration — what carries is that the error text, `IDLE_SETTLE_OVERHEAD_MS`'s "What a settle costs before any hold is counted", and SKILL.md's "it has to leave room for the hold plus the 600ms a settle costs" all teach a flow author a cost model the runner does not implement.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1117,6 +1151,467 @@ async function waitForCondition(
...
+      previousFrame = undefined;
+      treeSince = 0;
+      treeStillIntervals = 0;
+      treeSettledAtLastRead = false;
+      bothSince = 0;
+      stillIntervals = 0;
+    } else {
+      lastRead = "value";
+      readsSucceeded += 1;
+      darkReads = 0;
```
latekvo, line 1473: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: **Nothing in the suite pins this reset.**

Deleting `darkReads = 0;` leaves the whole flows suite green — 42 files / 1001 tests pass in a clean worktree at `d61b8395`. With it gone, `darkReads` accumulates across the entire step instead of measuring a consecutive tail, so two tree-read blips that are *not* consecutive — the case the surrounding design treats as benign and rides out — sum past `IDLE_TOLERATED_DARK_READS` and reach the unreadable branch at 1507, turning a passing step into a run-stopping `error` that skips every later step.

That is the exact behaviour the two review rounds on this file were spent establishing, and the line that enforces it is unprotected.

packages/tool-server/test/flows/flow-pixels.test.ts
```diff
@@ -0,0 +1,746 @@
...
+        for (let i = 0; i < Math.round(IPHONE[0] * IPHONE[1] * 0.01); i++) {
+          const o = (39 * IPHONE[0] + i) * 4;
+          frames[1].data[o] = 0;
+          frames[1].data[o + 1] = 0;
+          frames[1].data[o + 2] = 0;
+        }
+        expect(comparePixels(...frames, MASK)).toBe("moving");
+      });
+
+      it("takes its fractions against the unmasked area, not the whole frame", () => {
```
latekvo, line 353: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: **This test does not discriminate the property its title claims.**

Replacing `const total = a.width * (a.height - maskedRows);` in `comparePixels` with `const total = a.width * a.height;` — taking the fractions against the whole frame, which is the bug the title and the comment name — leaves all 49 tests in this file green.

The sample chosen (466 changed pixels) clears the motion budget under both denominators: 466/186334 = 0.00250 and 466/198112 = 0.00235, both above `MOTION_FRACTION`. So the assertion holds whichever area the fraction is taken against, and the masked-area denominator is unpinned.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -1117,6 +1151,467 @@ async function waitForCondition(
...
+        `settled on the UI tree alone — this screen could not be screenshotted on enough polls ` +
+        `to compare a pair of them, so animation that moves pixels without moving nodes (a push, ` +
+        `a fade, a dismissing modal) was not waited out. Follow this with the element check the ` +
+        `next step actually needs.`,
+    };
+  }
+  return {
+    ok: true,
+    warning:
+      `the screen never held still for ${stableFor}ms within ${timeoutMs}ms, so this step went ` +
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: **This sentence can name a hold the run demonstrably achieved, because the requirement that was actually unmet is the interval count, not the time.**

The settle needs `stillIntervals >= MIN_STILL_INTERVALS && now - bothSince >= stableFor`. Only the second clause appears in the message, so a run that met the time and fell one interval short is reported as one that never held still at all — and with `stableFor: 0` the sentence reads "the screen never held still for 0ms", which nothing can fail.

Driven through the real runner: a screen whose tree and pixels both change for the first four reads and are then completely identical for the last two — one full still interval, with `stableFor: 0` satisfied the instant that pair agreed:

```
await: { idle: true, stableFor: 0, timeout: 1100 }
-> pass, warning: "the screen never held still for 0ms within 1100ms, so this step went ahead
   without waiting it out. Either something on it never stops (a video, a looping animation,
   a carousel, live-updating text) or the screen never finished loading."
```

The screen held still for 200ms against a 0ms requirement, and the advice sends the author looking for a video or a carousel when what was missing was a second agreeing interval — which `IDLE_MIN_STILL_INTERVALS` documents as deliberate and no `stableFor` can lower.

packages/tool-server/src/tools/flows/flow-actions.ts
hubgan: Resolved in ec6a99a9.

Gated on the dark tail alone. `treeErrorMessage` already survives an abandoned read and is cleared by a successful one, so it means "the last read that came back did so as a failure" — which is the tail the guard is about. Your 1500/1700 pair is pinned as a case: both now error, and the failure is named either way.

The same hole existed on the degraded-read path added for the blind-tree finding; closed in a3d5e1b0 with a flag rather than a message, since `should_restart` arrives without a hint.

packages/tool-server/src/tools/flows/flow-actions.ts
hubgan: Resolved in ce25fdc8.

You are right that the budget alone cannot make the split — every read here is abandoned at the deadline, so how much it had is a fact about the step arithmetic. It is now measured against the slowest read that came back: a source answering in 2.5s has proved a 2.1s read means nothing, one answering in 100ms has proved a 6.9s read is a wedge. The read is timed on itself rather than on the round, since a round is the slower of its two halves.

Pinned with a static screen on a 2300ms source in a 4600ms wait, which errored before and passes now.

packages/tool-server/src/tools/flows/flow-actions.ts
hubgan: Resolved in a3cbee21.

Took `waitForCondition`'s guard minus its `everMatched` term — that one protects a claim this check does not make, and an undegraded blank screen has to stay an observation here since it is what resets both holds. A degraded read now counts with the dark reads, leaves the hold untouched, and reports as indeterminate quoting the reader's own repair.

packages/tool-server/src/tools/flows/flow-actions.ts
hubgan: Resolved in 31f36116.

Both halves. The pixel verdict is now three-state, so only a compared pair that disagreed breaks the hold — a missing frame is left as the absence of an observation, matching what the loop already decides about an abandoned tree read. And the screenshot warning requires that no pair was ever compared, rather than that a capture failed once.

Your measurement reproduced at 6000ms/900ms with one capture in three missing: 5876ms and the tree-alone warning before, ~1100ms and a clean settle after. A round with no pair still measures no interval, so an all-missing run cannot settle by accident.

packages/tool-server/src/tools/flows/flow-pixels.ts
hubgan: Resolved in 21f453ff.

Both directions. Android now takes `isAndroidTv`, pairing each platform with the probe it has the way `await-screen-idle` does, and `ios-remote` masks like a local sim. It is not probed for tvOS — that probe reads the local simulator list, so it cannot see another machine's device.

packages/tool-server/src/tools/flows/flow-pixels.ts
hubgan: Resolved in 1feca418.

The ceiling has an absolute floor now, so a change smaller than any measured indicator is never the screen moving, and that floor is itself capped at a tenth of the frame so it cannot close off a verdict the way the fraction did.

Measured wider than your table before the fix: a 57-pixel indicator read `moving` on 400x300, 500x400 and 700x520 and `localized` only at 1200x800. Confirmed end to end against a real 400x300 Chrome window with a small spinner — "never held still ... a video, a looping animation, a carousel" 3/3 before, the spinner warning 3/3 after.

The arithmetic is corrected too: the fraction works out at 19 captured pixels on the window this file measures elsewhere, and every indicator quoted clears 19, so it had not yet lost any of them.

packages/tool-server/src/tools/flows/flow-pixels.ts
hubgan: Resolved in ec0c00a5.

Comment only — the fix belongs in `captureVegaScreenshotPng`, which the `screenshot` tool shares. It now records that three routes keep the large frame out of this process (compositor, simulator-server, `sips`) and that Vega decodes at native resolution, resamples and re-encodes before `capturePng` decodes again.

packages/tool-server/src/tools/flows/flow-utils.ts
hubgan: Resolved in 65ca2415.

Replaced with `idleMinimumTimeoutMs`: the longer of the hold and the settle span, plus the closing round's budget. Your two cases are pinned — `timeout: 1000, stableFor: 800` settles in ~820ms and the default hold in ~411ms, against demands of 1400 and 850. The message, the constant and SKILL.md all say the two costs overlap now, and the first of those steps was run end to end on a simulator.

packages/tool-server/src/tools/flows/flow-actions.ts
hubgan: Resolved in 90ffae0c.

Covered with the shape that separates the two readings: reads failing on 2 and 4 of 4, so the tail is one read long and the step total is two. It passes with the reset and stops the run without it.

packages/tool-server/test/flows/flow-pixels.test.ts
hubgan: Resolved in 41857fee.

The count comes from the gap between the two budgets now — 380 is above 0.2% of the 186334 visible pixels and below 0.2% of the whole 198112-pixel frame — with both bounds asserted beside it so it cannot drift back onto a value that clears both.

packages/tool-server/src/tools/flows/flow-actions.ts
hubgan: Resolved in 759aa264.

Split into two endings. A wait that ran out mid-hold now says how long the screen had been still and which term was missing — a second agreeing interval, or the rest of the hold — and points at `timeout:`. The other keeps the motion advice and names the interval requirement, so the hold it quotes is one a screen can fail.

Your `stableFor: 0` case is pinned, and the new sentence turned up on a real simulator run: "the screen was still for the last 221ms of the 1000ms wait, over 1 200ms interval".

packages/tool-server/src/tools/flows/flow-actions.ts
hubgan: Already resolved on this branch, in 1c843d80 and d61b8395 — this thread predates both, which is why GitHub marks it outdated.

The tolerance is counted in rounds (`IDLE_TOLERATED_DARK_READS`) rather than milliseconds, for the reason your follow-up gave. All three rows of your table are pinned in `flow-idle-run.test.ts` at :742, :780, :811 and :841.

PR #729, opened 2026-08-04
feat(flow): record the screen identity gate from a fingerprint read

hubgan:
> Stacked on #728.

## What changes

A `screen-fingerprint` call made through `flow-add-step` is rewritten into the `await: { screen: <route> }` gate for the screen it just read.

A fingerprint **read** becomes a fingerprint **check** — recorded as the portable directive rather than a raw tool call, so the runner *compares* the route instead of merely re-reading it and discarding the answer. It is the same shape as the existing `gesture-tap` → `tap:` and `restart-app` → `launch:` rewrites.

## Three cases the rewrite has to get right

**A fingerprint that could not name the screen is refused.** There is no identity to gate on. Recording it raw would persist a `tool:` step whose output nothing checks; the failure belongs where the evidence still is. The refusal distinguishes "this app can never have a reader" (gate on an element instead) from "no route at this instant" (let it settle and call again).

**A fragment has no `launch` step**, so at replay the gate would have no app to read the route from and would fail as an environment error — discovered long after the recording looked fine. The fingerprint call had to name the app to run at all, so that id is carried onto the step, with a warning that it pins the fragment to one platform and how to avoid that (compose it from an e2e flow whose `launch` declares the app).

**A route equal to the flow's last gated route warns, but still records.**

This one is a judgement call worth stating. Such a gate *may* prove nothing: an app whose sign-in form is presented inside its landing route reports one fingerprint for both screens, so a gate recorded after "tap Sign in" passes whether or not the tap landed.

But it may equally be a **return trip** — Home → Search → Home — where the gate is perfectly falsifiable. Telling the two apart needs the route as it was *immediately before the last action*, and the recorder does not have it: the last route it gated on says nothing about where the app went in between, since intermediate screens are routinely gated on an element instead — exactly what the skills prescribe for a route-less screen.

So it warns. A refusal needs certainty, and without it blocking would make the ordinary "go there, come back, verify" shape unrecordable — while the QA contract calls a navigation with no identity gate a blocking defect. The recorder would be making its own contract unsatisfiable.

## `delayMs` is dropped loudly

A `screen` gate has no pre-step delay because it **polls**. Every other rewrite refuses to convert while `delayMs` is set, precisely so the delay survives; this one converts anyway — and dropping an author's parameter without a word is what must not happen. The warning names `timeout:` as the field that expresses the same intent.

### DIFF
--- a/packages/skills/rules/argent.md
+++ b/packages/skills/rules/argent.md
@@ -84,6 +84,7 @@ Decision order:
   If the user started Metro separately, ask whether to call `stop-metro` (specify the port if not 8081).
 - If tools provided by mcp-server are not sufficient and action can be done using `xcrun`, `adb`, or other commands, use the command. Examples: changing device options, performing a device action such as lock, shake, etc.
 - When waiting for an action, do not call `screenshot` repeatedly without a proper wait mechanism. Use the `await-ui-element` tool to block until the UI settles (e.g. wait for an element to become `visible`/`hidden`, or to contain expected `text`) instead of polling.
+- To confirm which screen a React Native app is on, use `screen-fingerprint` — it reads the app's focused navigation route. It answers only "which screen": it does not see a modal or system alert on top, and the route commits before the transition finishes animating.
   </general_rules>
 
 <react_native_detection>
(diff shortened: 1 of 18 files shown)


hubgan:
Closing with #727: this recorded the identity gate from a route fingerprint, and there is no fingerprint tool any more.

The recorder work here has no equivalent in the new shape, because the replacement identity gate is an ordinary `await-ui-element` on a destination-only element - already recordable, with no new command to teach `flow-add-step`. The readiness half, `await: { idle: true }`, is deliberately not recordable: it is a directive with no tool behind it, so it is added during polish alongside `snapshot:`.

#730 documents the pairing and now stacks directly on #728.

PR #734, opened 2026-08-04
feat(secrets): resolve {{secret:…}} from .env files, not just the environment

filip131311:
## Why

A credential could only reach a typing tool through an `ARGENT_SECRET_<NAME>` environment variable. That is the right CI contract — and stays the highest-priority source — but interactively it is poor ergonomics: an editor-hosted tool-server inherits the environment its editor was launched with, so changing a credential means restarting the editor, and there is nowhere to write one down that survives a reboot. "Is loading these from a `.env` file on the roadmap?" is the recurring ask.

## What

`{{secret:NAME}}` now resolves through an ordered chain, first match wins, re-read on every resolution — so editing a file applies to the next keystroke with **no server restart** (the same runtime-read design `ios.additionalDeviceSets` already uses).

| # | Source | Which keys it exposes |
| - | ------ | --------------------- |
| 1 | `ARGENT_SECRET_<NAME>` in the environment | prefixed variables only — unchanged, still wins |
| 2 | `<project>/.argent/secrets.env` | every key (gitignore it) |
| 3 | `<project>/.env.local`, then `<project>/.env` | **only `ARGENT_SECRET_`-prefixed keys** |
| 4 | `~/.argent/secrets.env` | every key, any project |

Covers `keyboard`, `paste`, `run-sequence` keyboard steps, and flow `type:` steps — all four already funnel through the same resolution point, so nothing else changed.

Also adds `argent secrets [list] [--json]` so a setup is verifiable without an agent: names only, never values, shadowed names marked, and a warning when a linked remote tool-server means resolution happens on a different machine.

## The one design decision worth reviewing

**Row 3 is why this isn't just "read the `.env`".** An agent composes the tool call, so anything reachable through `{{secret:…}}` is exfiltratable by a prompt-injected one — type it into a field, then read the field back. "Which values may argent type" has to stay an allowlist the user wrote deliberately.

So: a *dedicated* argent secrets file exists for no other purpose and therefore **is** the allowlist; a file the app *shares* keeps the prefix as the allowlist, exactly as the process environment does. Your `STRIPE_SECRET_KEY` sitting in the same `.env` stays unreachable — verified live, not just asserted. Adding `ARGENT_SECRET_` to one line is the deliberate act that exposes it.

Two smaller calls, both reversible:

- The chain lives in `@argent/configuration-core`, not the tool-server, so the CLI reads the same code; the tool-server keeps only the placeholder engine.
- The project scope is discovered by the **cwd walk-up every other project-scoped configuration uses**. A per-request project root (header → `ToolContext`) was considered and rejected as new protocol surface across two client paths; reusing flows' module-global `setActiveProjectRoot` was rejected as a cross-project leak on a reused server. The cost is that a tool-server whose cwd sits outside the project skips project files — mitigated by the failure now listing every path it consulted, and by `~/.argent/secrets.env` needing no project at all.

## Verification

Unit + integration: tool-server 3122, configuration-core 122 (19 new), argent-cli 370 (6 new). Lint and prettier clean. All four publish bundles build.

Live, driving this branch's tool-server on :3001 against iPhone 17 Pro / iOS 26.4 and the Electron testbed:

- global `~/.argent/secrets.env` → typed correctly; the response echoed only the placeholder
- prefixed key in the project `.env` → typed; **unprefixed `STRIPE_SECRET_KEY` in that same file → rejected**
- `.argent/secrets.env` beat `.env`; the environment variable beat both
- edited a secrets file mid-session → applied on the very next call, same server pid
- a chromium flow with `type: { text: "{{secret:APP_PASSWORD}}" }` → 3/3 pass, asserting on the greeting the app rendered, so the *correct value* provably landed
- an undefined name inside a flow surfaces the full source list in the step report

Sample of the failure output (no value ever appears):

```
Unknown secret "LOGIN_PASSWORD" — no source on the machine running the tool-server defines
LOGIN_PASSWORD. Available secrets: APP_USER.
Sources consulted, first match wins:
  1. environment (ARGENT_SECRET_*) — 0 secrets
  2. /path/to/project/.argent/secrets.env — found, 1 secret
  3. /path/to/project/.env.local — not found
  4. /path/to/project/.env — found, 1 secret
  5. /Users/me/.argent/secrets.env — not found
To make it available, ask the user to add `LOGIN_PASSWORD=…` to …
```

## Out of scope

No general flow variable/parameter system — a non-secret external value (a test account's email, a staging tenant id) can use the same placeholder, but argent treats every such value as a secret: redacted from errors, never echoed, so it is not for something a report should show. No config key pointing at an arbitrary secrets file. Remote (`argent link`) still resolves on the remote host, which the CLI now says out loud.


filip131311:
## Cost of resolving a placeholder — measured on two hosts

Prompted by a fair question on review: this puts synchronous file reads on the typing hot path, so how expensive is it? Pessimistic layout throughout — all four sources present, a 200-key `.env`, tool-server cwd **12 levels deep** inside the project so `findProjectRoot` does the full walk-up before anything is read.

**Steady state** (2000 iterations, warm cache — what a running server sees, since it re-reads the same four paths every call):

| Scenario | MacBook (node 22) | Raspberry Pi 5 / NVMe (node 18) |
| --- | --- | --- |
| No placeholder — every ordinary keyboard call | 0.0002ms | — (never builds the chain) |
| Typical: 1 small file, 3 absent | 0.020ms | 0.072ms |
| All 4 files present, small | 0.036ms | 0.116ms |
| **Pessimistic: 200-key `.env` + deep cwd** | **0.187ms** (p99 0.32) | **0.710ms** (p99 1.06) |
| Adversarial: 1.03 MB / 20k-key `.env` | 12.5ms | 53.6ms |

**One-off costs**, paid once per tool-server lifetime, not per call — Pi: first resolution in a fresh process 4.98ms warm, **8.28ms with the page cache genuinely dropped** (`drop_caches` between runs), so ~3.3ms of real disk on cold NVMe. Mac equivalent 1.17ms warm.

**End-to-end it is invisible.** A 9-character `keyboard` call on the iOS simulator, 8 runs each: plain text 928.8ms, `{{secret:…}}` 930.0ms, plain text again 928.8ms. At 50ms/keystroke the resolution is ~0.02% of the call.

**Where the time goes** (pessimistic case, Pi): `dotenv.parse` of the 200-key `.env` 323µs, `findProjectRoot` walk-up 162µs, 4× `readFileSync` 55µs. So the cost is **parsing, not I/O** — a warm small-file read is ~5–16µs, roughly a syscall. That is also why the adversarial case is bad: it is parse-bound, and since the read is synchronous it blocks the event loop for that long. I left it uncapped deliberately — a size limit that silently stopped honoring a large `.env` is a worse failure than 54ms — but say the word and I will add a cap that *reports* the skip.

### The Pi found a real bug (fixed in 1eb60e4)

On the Pi the *typical* layout measured **slower** (226µs) than the strictly larger all-files-present one (100µs), reproducibly across fresh processes. Cause: absent sources were detected by letting `readFileSync` reject, and constructing that Error — stack capture included — costs 40.6µs against 3.0µs for a non-throwing `stat`. Since the common setup misses three of four sources on every call, the empty case was the expensive one.

Now probed with `statSync({ throwIfNoEntry: false })` first; the read stays wrapped, because `stat` cannot rule out a directory, a permission error, or a delete between the two calls. Typical layout: **226µs → 59µs** on the Pi, 25.5µs → 19.7µs on macOS, and the ordering is sane again. Behavior is unchanged and the existing absent-file / directory cases still cover it.

PR #736, opened 2026-08-04
test(telemetry): pin the wire format and what a dead collector costs

latekvo:
Two test files that look at what #570's exporter actually *does*, plus the configuration holes they exposed.

The suite on #570 proves the exporter is **configured** right - the url, the headers, the timeouts, all read off the constructor. Nothing looked at the request that comes out of it, or at what happens when the collector does not answer. Both gaps are invisible from inside Argent by design: a failed export never reaches the caller, and an OTLP exporter counts any 2xx as delivered.

| | |
|---|---|
| **`otel-wire.test.ts`** | Captures the real exporter's request off a loopback socket: uncompressed OTLP/JSON on `POST /v1/logs`, where each field the ingestion schema reads actually sits, the batch-size cap, and connection reuse. |
| **`otel-unreachable.test.ts`** | Drives the same exporter at a silent socket, a refused port, a 503, a 401 and a 204, and asserts each drain finishes inside its deadline and what the collector really received. |
| **`src/otel.ts`** | Two holes the tests exposed: `compression` and the six OTLP certificate variables are now settled in code instead of left to the environment. |

**No docs update needed** - no user-facing capability, CLI surface, config key, flow file or MCP tool changed.

<details>
<summary><b>The two source fixes</b></summary>

**`compression` was the one OTLP knob still left to the environment.** With `OTEL_EXPORTER_OTLP_COMPRESSION=gzip` (or its `_LOGS_` variant) set - and a machine already running OpenTelemetry has it set for its own collector - argent gzips its batches. Nothing downstream notices: a gzipped 2xx is a delivered batch, while the request this suite and the deploy-time smoke test are written against quietly stops being the request argent sends. `createExporter` now passes `compression` explicitly, which beats both variables. Unlike the header channel, where the SDK *merges* whatever the code does not set and the variables have to be cleared outright, one explicit value settles this.

**The certificate variables could hang the command outright.** The SDK resolves `OTEL_EXPORTER_OTLP_{,LOGS_}{CERTIFICATE,CLIENT_CERTIFICATE,CLIENT_KEY}` with a synchronous `fs.readFileSync` while the exporter is constructed, and only then discards the https agent it built from them in favour of the explicit `httpAgentOptions`. The file is read for nothing - and a path that never answers hangs the command that emitted the event, uninterruptibly, on `getClient()`'s path. Measured with a fifo that has no writer: `createExporter` never returned (killed at 12s); with the variables cleared, 1.0 ms. They join the header pair in the list cleared across the constructor and restored on the way out.

Some source prose was also wrong about the SDK and is corrected: it does **not** reject a null attribute - it serializes one as an empty OTLP value, `{"key":"cloud_agent","value":{}}` with `droppedAttributesCount: 0`, stored as and unrecoverable from a property that really was empty. That is the actual reason `toAttributes` drops the key.

</details>

<details>
<summary><b>What the tests pin, and what they deliberately do not</b></summary>

The wire test **mirrors** `OtelClient.emit` rather than calling it: `getClient()` resolves its endpoint from the hard-coded `OTLP_LOGS_ENDPOINT`, and that being unredirectable is the anti-exfiltration property the client is meant to have, so there is deliberately no seam to aim it at a test server. `otel-endpoint.test.ts` covers the other half against the real emit path - `SERVICE_NAME`, the logger name, event-name-as-body, the dropping of null/undefined properties - so a rename is caught there rather than in the mirror.

What enforces the drain budgets is the exporter's `timeoutMillis`, *not* the processor's `exportTimeoutMillis`, which `forceFlush()` awaits straight past. Measured against a silent socket: `exportTimeout` 200 / `timeoutMillis` 5000 drains in 5012 ms, the reverse in 202 ms.

503 is the one collector state where the batch is sent twice, and the request count is the only thing that tells a retried failure from a rejected one - the SDK reports the class, not the code. The retry itself is only reachable while the next backoff fits in what is left of `timeoutMillis`, so the jitter is pinned to its low draw and the assertion turns on the round trip alone.

`httpAgentOptions.timeout` is the one option with no behavioural coverage: it covers a socket stuck in *connect*, which needs an address whose packets are dropped and is not portably reproducible here. It stays pinned at the constructor and the docstring says so. Its `keepAlive` half **is** covered, on the two batches of a single drain.

Real deadlines cost real time: `otel-unreachable.test.ts` waits ~5.1 s, taking the package suite from ~0.9 s to ~5.3 s. Inherent to driving real deadlines against real sockets.

</details>

<details>
<summary><b>Verification</b></summary>

- `npm test -w @argent/telemetry` - **336 passed**, 19 files
- `typecheck:tests`, `eslint --max-warnings 0`, `prettier --check`, `knip`, `tsc --build` all clean; CI green
- **Mutation-tested**, whole-suite counts rather than one test each:

  | mutation in `src/otel.ts` | tests red |
  |---|---|
  | `compression` pin dropped | 2 |
  | `compression` pinned to `"gzip"` | 5 |
  | any one of the 8 cleared env-var names dropped | 1 each |
  | env-clearing loop deleted | 2 |
  | the `finally` that restores it deleted | 1 |
  | `timeoutMillis` dropped | 5 |
  | `httpAgentOptions` dropped, or just its `timeout` | 1 |
  | `keepAlive` flipped to `false` | 2 |
  | `installDiagLogger`'s `!isDebugEnabled()` guard removed | 1 |
  | `maxExportBatchSize` 20 → 19 or 25; `scheduledDelayMillis` changed | 2 |
  | `SERVICE_NAME` renamed | 2 |
  | `LOGGER_NAME` changed | 1 |
  | record body no longer the event name | 4 |
  | `severityText` INFO → WARN | 1 |
  | null/undefined property skip removed | 2 |
  | `url` given a suffix the config did not set | 9 |

- **Absence assertions carry positive controls.** The certificate test reads its answer out of a `Failed to read ` prefix that lives in a transitive dependency, and the 204 case asserts an empty diag channel; both would pass just as well against a dead observer. Each now proves its own observer live first.
- **Flake-tested.** The 503 case's exact two-request assertion used to turn on the jitter draw: standing a 500 ms collector delay in for a loaded runner made it fail 20/20 (13 on the request count, 7 on `Timeout` masking the error class). With the jitter pinned, 20/20 pass.

</details>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved telemetry export reliability by preventing unintended compression and certificate-related environment settings from affecting exporter configuration.
  - Ensured telemetry attributes with null or undefined values are omitted rather than serialized as misleading empty values.
  - Improved handling of unreachable or unresponsive telemetry collectors with bounded failures and appropriate diagnostic reporting.
  - Preserved connection reuse and batching behavior for more efficient OTLP delivery.

- **Tests**
  - Added comprehensive coverage for OTLP formatting, authorization, batching, compression, certificates, connection handling, and failure scenarios.


packages/telemetry/test/otel-wire.test.ts
```diff
@@ -0,0 +1,229 @@
...
+}
+
+let capture: Capture;
+
+beforeEach(async () => {
+  capture = await startCapture();
+});
+
+afterEach(async () => {
+  await new Promise<void>((resolve) => capture.server.close(() => resolve()));
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `beforeEach` can fail before assigning `capture` - `listenLoopback` rejects on a bind failure, which is the whole point of it rejecting - and this line then dereferences it, so vitest reports a second failure under the real one.

Reproduced by exhausting the ephemeral port range with six parallel binders and running this file:

```
Error: listen EADDRNOTAVAIL: address not available 127.0.0.1     [1/6]
TypeError: Cannot read properties of undefined (reading 'server') [2/6]
  ❯ test/otel-wire.test.ts:137:48
```

The real cause is reported first, so this costs a confusing second line rather than a wrong diagnosis. `packages/telemetry/test/otel-endpoint-live.test.ts:75` has the same shape. Left alone because narrowing `capture` to `Capture | undefined` puts a `!` on every use of it in the file, which reads worse than the extra failure line.

For scale, the same repro against the pre-`listenLoopback` code hung for 61.21s and surfaced the bind error as an uncaught exception attributed to whichever test was running; it now fails in 204ms.

packages/telemetry/test/otel-unreachable.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    // real ingest, which is why the check has to live at deploy time, against
+    // the endpoint, rather than here. The empty error channel IS that missing
+    // signal: it is what a delivered batch and a discarded one both look like.
+    const errors = captureExportErrors();
+    const collector = await startResponding(204);
+
+    await expect(exportAndDrain(collector.url)).resolves.toBeLessThan(FAILURE_BUDGET_MS);
+
+    expect(collector.requests).toEqual(["/v1/logs"]);
+    expect(errors).toEqual([]);
```
latekvo, line 318: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This assertion reads the same whether the diagnostics channel is live or dead - a `captureExportErrors` whose `error` callback never fires would satisfy it too.

Confirmed by mutating that callback to a no-op and running the file: 2 failed | 3 passed. The 401 case and the refused case catch it; this one stays green. So the control is the siblings, not the test.

Left as is: a positive control inside this test means driving a second, failing export through the same channel, which duplicates the 401 case one test above for a property three tests in the file already exercise.

packages/telemetry/test/otel-unreachable.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    severityText: "INFO",
+    body: "tool:invoke",
+    attributes: { "distinct_id": "a".repeat(64), "event.name": "tool:invoke" },
+  });
+  const started = performance.now();
+  await provider.shutdown();
+  return performance.now() - started;
+}
+
+describe("a collector that cannot take the batch", () => {
```
latekvo, line 194: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This file is most of what `@argent/telemetry` costs to test, and none of it parallelises.

Measured by moving the file aside and back, six runs each: package suite 1.08 / 1.13 / 1.41 s without it, 5.45 / 5.55 / 5.57 s with. The file alone is 5.15-5.84 s against 0.45-0.49 s for `otel-wire.test.ts`. It is real waiting - roughly 1.5 s silent, 1.5 s teardown, 0.8-1.2 s refused backoff, 0.8-1.2 s retryable backoff - and the six `it`s share one `describe` in one file, so vitest runs them serially in a single worker whatever the core count.

CI is worse rather than better: `maxWorkers` resolves to `max(availableParallelism() - 1, 1)`, so on a 2-core `ubuntu-latest` runner it is 1 worker and this file's ~5 s adds to the sum instead of hiding under the max. Inherent to driving real deadlines against real sockets, so it is a cost to know about rather than a thing to fix, but it is not visible from the diff.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Three things I found while going over this and decided not to change, recorded so they are on the record rather than lost. None of them is a defect in what the tests assert; each is a limit on what a failure here will tell the next person.

packages/telemetry/test/otel-wire.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in bcf138e8caf275a972cf509e4d40e47594ae2847. Both capture suites now register the close on a `closers` stack at the moment `listenLoopback` returns, matching the idiom this PR introduces in `otel-unreachable.test.ts:47-55`, so `afterEach` never dereferences a variable a failed `beforeEach` left unassigned and `Capture` no longer needs to carry `server`.

Repro, forcing the second `startCapture()` in the live suite to reject and grepping the run: before the change the output carried both `Error: SIMULATED BIND FAILURE` and `TypeError: Cannot read properties of undefined (reading 'server')`; after it, only the real error. `process._getActiveHandles()` reported `listening=0` in both runs, so nothing was leaking - the defect was the masked message, as described.

321/321 still green.

latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Correcting my previous reply: `listening=0` held only for the case I probed, where the bind itself rejects and no server exists yet. When setup fails *after* a server has bound, the old shape did leak it.

Repro - throw at the end of the second `startCapture()`, i.e. once `listenLoopback` has already returned, with an `afterAll` counting handles where `listening === true`:

- at `924f9e2c1`: `LEAKPROBE stillListening=1`, plus `TypeError: Cannot read properties of undefined (reading 'server')`
- at `bcf138e8c`: `LEAKPROBE stillListening=0`, and only the real error

The old teardown could only close what the setup had finished assigning, so a server bound by a call that never returned had nothing holding a reference to it and stayed listening for the rest of the process. Registering the closer at bind time rather than at assignment time is what closes that gap - so this was a leak as well as a masked message, not just the latter.

packages/telemetry/test/otel-endpoint-live.test.ts
```diff
@@ -137,16 +142,75 @@ describe("what reaches the collector, against the real OTLP exporter", () => {
...
+    process.env.OTEL_EXPORTER_OTLP_CLIENT_KEY = "/nonexistent/argent-probe-key.pem";
+    process.env.OTEL_EXPORTER_OTLP_LOGS_CLIENT_KEY = "/nonexistent/argent-probe-logs-key.pem";
+
+    try {
+      createExporter({ endpoint: code.url, token: "real-ingest-token", isUsable: true });
+    } finally {
+      diag.disable();
+    }
+
+    expect(warnings).toEqual([]);
```
hubgan: This captures the entire diag stream, not just the certificate warnings, so any unrelated SDK warning emitted during the constructor fails a test about certificates. `snapshotEnv` saves the OTLP variables without clearing them, and `createExporter` deliberately leaves `OTEL_EXPORTER_OTLP_COMPRESSION` and `_TIMEOUT` in place - both of which the SDK warns about unconditionally when the value is invalid.

Reproduced from `packages/telemetry`:

```
OTEL_EXPORTER_OTLP_COMPRESSION=zstd npx vitest run test/otel-endpoint-live.test.ts

  x reads no file an OTLP certificate variable names
    expected [ Array(1) ] to deeply equal []
    + "Configuration: OTEL_EXPORTER_OTLP_COMPRESSION is invalid, expected 'none' or 'gzip' (actual: 'zstd')"
```

`OTEL_EXPORTER_OTLP_TIMEOUT=-5` fails the same test the same way. A realistic valid OTel developer environment (endpoint, protocol, headers, `COMPRESSION=gzip`, service name) passes, so the trigger is an invalid value rather than merely a present one.

packages/telemetry/test/otel-endpoint-live.test.ts
```diff
@@ -137,16 +142,75 @@ describe("what reaches the collector, against the real OTLP exporter", () => {
...
+    process.env.OTEL_EXPORTER_OTLP_CERTIFICATE = "/nonexistent/argent-probe-ca.pem";
+    process.env.OTEL_EXPORTER_OTLP_LOGS_CERTIFICATE = "/nonexistent/argent-probe-logs-ca.pem";
+    process.env.OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE = "/nonexistent/argent-probe-cert.pem";
+    process.env.OTEL_EXPORTER_OTLP_LOGS_CLIENT_CERTIFICATE =
+      "/nonexistent/argent-probe-logs-cert.pem";
+    process.env.OTEL_EXPORTER_OTLP_CLIENT_KEY = "/nonexistent/argent-probe-key.pem";
+    process.env.OTEL_EXPORTER_OTLP_LOGS_CLIENT_KEY = "/nonexistent/argent-probe-logs-key.pem";
+
+    try {
+      createExporter({ endpoint: code.url, token: "real-ingest-token", isUsable: true });
```
hubgan, line 195: The harm this test exists to prevent is the synchronous read hanging the command that emitted the event, but what it asserts is the absence of a warning. That does catch a regression today - dropping the six names from the cleared list makes the SDK's "Failed to read root certificate file" fire - but it pins "the SDK did not warn", not "no file was read". If a future SDK stops warning on a failed read, this keeps passing while the hang comes back.

The measurement quoted in the comment above ("the constructor never returned") is the property that actually matters, and it is the one not encoded anywhere in the suite.

packages/telemetry/test/otel-endpoint.test.ts
```diff
@@ -139,6 +158,76 @@ describe("otel endpoint invariance", () => {
...
+    // agent, which is nearly all of them.
+    const client = getClient();
+    client!.emit({
+      distinctId: "d".repeat(64),
+      event: "tool:invoke",
+      properties: { cloud_agent: null, os_version: undefined, tool: "screenshot" },
+    });
+
+    const record = otelMock.providers[0]!.emit.mock.calls[0]![0] as EmittedRecord;
+    expect(record.attributes).toEqual({
```
hubgan, line 238: `toEqual` ignores a key whose value is `undefined`, so this assertion cannot see `os_version` surviving into the record. Only the `null` half of the guard in `toAttributes` is covered here, despite the test passing `os_version: undefined` and the comment describing both.

The two are not interchangeable downstream: an `undefined` attribute value serializes exactly like a `null` one. Captured off a real exporter, both come out as `{"key":"os_version","value":{}}` with `droppedAttributesCount: 0`.

`index.test.ts` already covers this case and covers it strictly better, asserting `expect("missing" in attributes).toBe(false)`. A change that drops only `null` and keeps `undefined` turns that test red and leaves this one green.

packages/telemetry/test/otel-endpoint.test.ts
```diff
@@ -139,6 +158,76 @@ describe("otel endpoint invariance", () => {
...
+  });
+
+  it("identifies itself as the service and scope the collector's schema keys on", () => {
+    // These two strings are how a stored record is attributed: the resource's
+    // service.name lands in ServiceName and the instrumentation scope in
+    // ScopeName, and the ingestion side filters on both. They are load-bearing
+    // identifiers rather than labels, so renaming either silently orphans every
+    // record written afterwards from every query written before.
+    getClient();
+    expect(otelMock.providers[0]!.config.resource).toEqual({
```
hubgan, line 184: This resource assertion is already made in `index.test.ts` ("identifies the record by service.name resource and event.name attribute"), and a `SERVICE_NAME` rename turns both red.

The `loggerNames` assertion on the next line is the half that does new work - it is the only place in the package that catches a `LOGGER_NAME` rename.

packages/telemetry/test/otel-wire.test.ts
```diff
@@ -0,0 +1,232 @@
...
+async function exportRecords(
+  endpoint: string,
+  events: Array<{ event: string; attributes: Record<string, string | number | boolean> }>
+): Promise<void> {
+  const provider = new LoggerProvider({
+    resource: resourceFromAttributes({ "service.name": "argent" }),
+    processors: [
+      new BatchLogRecordProcessor({
+        exporter: createExporter({ endpoint, token: "wire-probe-token", isUsable: true }),
+        maxExportBatchSize: 20,
```
hubgan, line 117: These three numbers are copies of `MAX_EXPORT_BATCH_SIZE`, `SCHEDULED_DELAY_MS` and `EXPORT_TIMEOUT_MS`, which `src/otel.ts` does not export; `otel-unreachable.test.ts` and `otel-endpoint-live.test.ts` carry the same copies.

The consequence lands on the batching test below. Change the shipped batch size and the only test that goes red is `otel-endpoint.test.ts > uses the queued batching config`, which reads the real value through the mock. Update that one literal to match - which is what changing the constant would involve - and the suite is green again while this file still builds its own provider at 20 and still asserts `<= 20`. The test whose stated purpose is keeping a request under the body caps in front of ClickHouse would then be asserting about a number production no longer uses.

packages/telemetry/test/otel-wire.test.ts
```diff
@@ -0,0 +1,232 @@
...
+    // exporter treats a gzipped 2xx as delivered just the same, and the request
+    // this suite and the deploy-time smoke test are written against silently
+    // stops being the request argent sends.
+    expect(request.headers["content-encoding"]).toBeUndefined();
+
+    expect(request.headers.authorization).toBe("Bearer wire-probe-token");
+    expect(() => JSON.parse(request.raw.toString("utf8")) as OtlpPayload).not.toThrow();
+  }, 15_000);
+
+  it("puts every column the analytics table reads where the collector looks for it", async () => {
```
hubgan, line 176: Every value this test checks was written by the same test 60 lines up in `exportRecords` - the resource, the scope name, the body and the attributes are all its own literals rather than argent's. As a pin on the SDK's OTLP/JSON shape that is genuinely useful, and it is the only such pin in the repo.

But renaming `SERVICE_NAME`, renaming `LOGGER_NAME`, or changing `OtelClient.emit`'s `body: record.event` all leave it green; each of those is caught only in `otel-endpoint.test.ts`.

The file header is explicit about the mirror-rather-than-call tradeoff. The test name and the `// -> ClickHouse ServiceName` / `// -> ScopeName` markers still read as though argent's identifiers were guarded here, which is where a future reader would look for them.

packages/telemetry/test/otel-wire.test.ts
```diff
@@ -0,0 +1,232 @@
...
+    expect(request.path).toBe("/v1/logs");
+
+    // JSON, not protobuf. @opentelemetry/otlp-transformer ships both
+    // serializers, but the exporter package this one depends on hardcodes the
+    // JSON one and sends this content-type alongside it; protobuf-over-HTTP for
+    // logs is a separate package. So this pins a dependency choice rather than a
+    // default anything could flip.
+    expect(request.headers["content-type"]).toBe("application/json");
+
+    // No compression - which `createExporter` has to pin, because the SDK
```
hubgan: This file never sets `OTEL_EXPORTER_OTLP_COMPRESSION`, so the risk the comment describes is not exercised here - deleting `compression: NO_COMPRESSION` from `src/otel.ts` leaves all three tests in this file green.

The guard actually lives in `otel-endpoint.test.ts` (the constructor value) and `otel-endpoint-live.test.ts` (the env-var precedence). What this assertion does pin is that the shipped value is not `gzip`, which is worth having on the wire; the comment just credits it with more than it does.

packages/telemetry/test/otel-unreachable.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    // where it receives the same batch twice. The status never reaches the error
+    // channel - the SDK reports the class, not the code - so a retried failure
+    // and a rejected one are told apart by the request count, not the message.
+    const errors = captureExportErrors();
+    const collector = await startResponding(503);
+
+    const elapsed = await exportAndDrain(collector.url);
+
+    expect(elapsed).toBeLessThan(FAILURE_BUDGET_MS);
+    expect(collector.requests).toEqual(["/v1/logs", "/v1/logs"]);
```
hubgan, line 280: This assertion sits inside a timing band of roughly 300 ms.

The retrying transport sets `deadline = now + 1500` before the first attempt, then skips the retry entirely when the computed backoff exceeds the remaining budget. Backoff is `1000 * (1 + jitter)` with jitter in +/-0.2, so 800-1200 ms - which means the retry only happens while the first loopback round trip finishes inside `1500 - backoff`, i.e. under 300 ms in the worst case. Above roughly 700 ms it cannot happen at all and this becomes `['/v1/logs']`. Three requests is not reachable, since the second backoff is 1200-1800 ms against at most 700 ms of remaining budget, so the only failure direction is one-instead-of-two.

I could not make it fail naturally: 14 runs of this file at load average 110-165 on a 12-core box, 10 shuffled full-suite runs, and a run on Node 20 were all clean, matching the sweep in the description. Recording the margin anyway, since CI is a 2-core runner forking 19 files.

packages/telemetry/test/otel-unreachable.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    expect(errors.join("\n")).toContain("retryable");
```
hubgan: Two things about this line.

It passes whether or not a retry happened. `otlp-export-delegate` maps any final retryable status to `OTLPExporterError("Export failed with retryable status")`, so a single 503 with no re-send produces the same string. The comment above already says the request count is what tells the two apart, so this reads as belt-and-braces rather than as a second signal.

It is also the assertion that notices if the export runs long. The processor rejects with `Error: Timeout` at `exportTimeoutMillis` (1500, set at line 176), and once that promise settles the exporter's real error is dropped - so an export taking over 1500 ms leaves `errors` holding only `Timeout`. The 503 path spends 800-1200 ms of that in backoff before the second attempt, leaving around 300 ms of slack. The `ECONNREFUSED` assertion above shares the mechanism.

hubgan (commented):
Reviewed the source change and the four OTLP suites against the real SDK (`@opentelemetry/*` 0.221.0, api 1.9.1) rather than from the diff alone.

**Nothing here is blocking.** I found no correctness problem in the shipped code, and every claim the new prose makes about the SDK checks out. The comments below are all test-quality and documentation points.

What I verified independently:

- Under a hostile environment (`OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`, and `OTEL_EXPORTER_OTLP_{ENDPOINT,LOGS_ENDPOINT,HEADERS,COMPRESSION,LOGS_COMPRESSION,TIMEOUT,PROTOCOL,CERTIFICATE}` all set hostile) the request still leaves as uncompressed JSON `POST /v1/logs` to the code-supplied URL, carrying only argent's bearer token, with `service.name=argent` and scope `@argent/telemetry` - and every variable is restored synchronously by the time `createExporter` returns.
- The certificate hang is real and the fix lands. With `OTEL_EXPORTER_OTLP_CERTIFICATE` pointed at a writer-less fifo, the pre-fix constructor never returned (killed at 12s); `createExporter` on this branch returns in 1 ms.
- The rewritten `shutdown()` comment is right about which deadline binds: `exportTimeoutMillis` 200 / `timeoutMillis` 5000 drains in 5482 ms, the reverse in 202 ms, and the shipped 1500/1500 pair in 1503 ms.
- The corrected null-attribute prose is right, and the comment it replaced was wrong: a null attribute serializes as `{"key":"cloud_agent","value":{}}` with `droppedAttributesCount: 0` and no diag warning.
- The env surface really is closed. The only OTLP variables this SDK acts on are the six certificate/key names and headers (all cleared), plus compression, timeout and endpoint (all settled by an explicit option). `OTEL_SERVICE_NAME` / `OTEL_RESOURCE_ATTRIBUTES` never reach the wire, because `LoggerProvider` only falls back to `defaultResource()` when no resource is passed.
- 321/321 pass. 10 shuffled full-suite runs, 14 runs of `otel-unreachable` at load average 110-165, and a run on Node 20 (what CI uses) were all clean. `typecheck:tests`, `eslint --max-warnings 0` and `prettier --check` are clean, and `knip` reports nothing from this diff.

One process point with no natural line to attach to: the repo CLAUDE.md asks that a PR say so in its description when no docs update is needed. This one changes which OTLP environment variables argent honours, and the description does not mention docs either way. No page is actually made stale - `packages/docs` never mentions OTEL or OTLP anywhere, and the Telemetry table in `reference/configuration.mdx` lists only `DO_NOT_TRACK`.

packages/telemetry/test/otel-endpoint-live.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in de227e661. The capture now filters to the SDK's `Failed to read ...` cert-read warnings before asserting, so an invalid `OTEL_EXPORTER_OTLP_COMPRESSION` or `_TIMEOUT` in the ambient environment - both of which `createExporter` leaves in place by design - no longer fails this cert test.

Verified against your repro: `OTEL_EXPORTER_OTLP_COMPRESSION=zstd` and `OTEL_EXPORTER_OTLP_TIMEOUT=-5` each failed this test before and both pass now, while the mutation that drops `OTEL_EXPORTER_OTLP_CERTIFICATE` from the cleared list still fails cleanly with `Failed to read root certificate file`.

packages/telemetry/test/otel-endpoint-live.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Comment corrected in de227e661 - it no longer claims the test measures the hang; it now says the absence of a `Failed to read ...` warning is the observable that the clear held.

The assertion stays warning-based on purpose. A direct "does not hang" test would point a cleared var at a writer-less fifo and bound `createExporter`, but the SDK reads certs with a synchronous `fs.readFileSync`, so a regression would wedge the vitest worker until the harness timeout - a strictly worse failure mode than the clean, fast assertion here (`fs.readFileSync` blocks the whole event loop, so there is no in-process way to bound it). The regression this actually needs to catch is the cleared list drifting, and the warning catches that immediately - verified by mutation: dropping a name yields `Failed to read root certificate file`. The remaining case you name, a future SDK that stops warning on a failed read, changes the read path itself and so lands at the SDK-upgrade review, where the version bump is the signal.

packages/telemetry/test/otel-endpoint.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in de227e661. Added `expect("cloud_agent" in record.attributes).toBe(false)` and the same for `os_version`, so both halves of the guard are under test rather than only the null one, and left the `toEqual` in place for the positive shape.

Verified by mutation: changing `toAttributes` to `if (value === null) continue` - dropping only null, keeping undefined - now fails at the `os_version` assertion, where before it passed the `toEqual` untouched. As you noted, `index.test.ts` already covers this strictly; this keeps the local assertion honest about what it claims to test.

packages/telemetry/test/otel-endpoint.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Kept as is. The duplication is intentional across layers: this asserts the resource `OtelClient` hands the SDK with the SDK mocked (the constructor contract in isolation), while `index.test.ts` asserts it through the full `track()` -> `getClient()` path. The `loggerNames` line right below is unique to this file, as you point out. A `SERVICE_NAME` rename going red at both layers is the intended belt-and-suspenders rather than noise to trim - the two tests fail for different reasons (wiring vs. end-to-end), so neither makes the other safe to delete.

packages/telemetry/test/otel-wire.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Left as is. `otel-endpoint.test.ts > uses the queued batching config` is the deliberate canonical pin of the shipped values - asserted as literals against the real, mocked processor config - and it going red on a batch-size change is the intended forcing function.

Exporting the constants would either make that literal pin self-referential, or, if only the mirror suites imported them, entangle their fixtures with production: this test's `Array.from({ length: 25 })` encodes "more than one batch of 20", a premise that silently breaks if the batch size rises past 25. The mirror suites rebuild the provider wiring by hand on purpose (the file headers state the mirror-not-call rationale), so `<= 20` here is a body-cap ceiling whose "one batch per request" purpose holds whatever the exact number. The drift you describe is real but rare and loud - the endpoint test is red the moment the constant moves - so I would rather not couple fixture sizes to a production constant to chase it.

packages/telemetry/test/otel-wire.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Addressed in de227e661. Added a scope note before the assertions: the asserted values are this suite's own literals, so what the `// -> ClickHouse X` markers pin is where each field lands in the OTLP/JSON envelope, not that argent still names them this - and a `SERVICE_NAME`/`LOGGER_NAME` rename, or moving the event off `body`, is caught in `otel-endpoint.test.ts`. That is exactly where a future reader would otherwise go looking for those identifiers, so the note now sends them there. The mirror-vs-call tradeoff itself stays as the file header already frames it.

packages/telemetry/test/otel-wire.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in de227e661. The comment no longer credits this assertion with guarding the `OTEL_EXPORTER_OTLP_COMPRESSION` precedence - this file sets no such env var, and you are right that deleting `compression: NO_COMPRESSION` from `src/otel.ts` leaves all three tests here green. It now says the assertion pins the shipped uncompressed value on the wire, and points at `otel-endpoint{,-live}.test.ts` for the env-precedence guard.

packages/telemetry/test/otel-unreachable.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Kept as is. The two-request assertion is the test's whole purpose - "re-sends once on a retryable status" - so loosening it to `<= 2` would let a no-retry run pass and hollow the test out. The ~300 ms band you measured is inherent to driving a real 1.5 s deadline against a real socket (the file header calls this cost out), and the sweep - your 14 runs at load 110-165, 10 shuffled full-suite, Node 20, plus a rerun here - came back clean. Recording the margin is the right outcome; there is no tighter assertion that still asserts the retry actually happened, and three requests is unreachable as you showed, so the only failure direction is a real regression to one.

packages/telemetry/test/otel-unreachable.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Kept as is. The request-count line above is the primary retry signal; this `retryable` assertion corroborates that the failure reached the diag channel at all rather than being dropped silently - the two together say "retried AND surfaced", which is more than either alone.

The Timeout-masking you flag is real but is the same ~300 ms slack as the sibling: the 503 path spends 800-1200 ms of the 1500 ms `exportTimeoutMillis` in backoff, and an export that overran the budget would leave `errors` holding only `Timeout` and fail this line - it cannot pass it wrongly. Unreproduced across the same sweep, so I am leaving it as a documented margin rather than widening the assertion to `retryable|Timeout`, which would mask exactly that overrun.

packages/telemetry/src/otel.ts
```diff
@@ -105,44 +124,81 @@ function toAttributes(record: EmitRecord): LogAttributes {
...
       //
       // keepAlive is restated because supplying httpAgentOptions at all
-      // replaces the agent the SDK would otherwise build, whose default is
-      // keepAlive: true; without it the long-lived tool-server pays a fresh
-      // TCP+TLS handshake for every batch.
+      // replaces the agent the SDK would otherwise build, and its default is
+      // keepAlive: true. Without it the long-lived tool-server pays a fresh
+      // TCP+TLS handshake for every 10s batch — measured as one socket per
+      // request against a loopback collector, versus one socket shared.
       httpAgentOptions: { timeout: EXPORT_TIMEOUT_MS, keepAlive: true },
```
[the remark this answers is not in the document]
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Good catch on the mechanism - confirmed. A raw `http.Agent({ keepAlive: true, timeout: 1500 })` opens two connections across a 2500ms gap (idle socket reaped at 1500ms) versus one with no timeout, so at the 10s `SCHEDULED_DELAY_MS` cadence the pooled socket is always gone before the next batch and keepAlive buys no steady-state reuse.

I fixed the comment rather than the code, because the code here is the intended one. That `timeout` is load-bearing: it is the connection-establishment timeout that bounds a blackholed collector - a socket stuck in `connect` that `req.setTimeout()` (armed only post-connect) cannot reach - which is the whole reason this PR added `httpAgentOptions`. Raising or dropping it to keep idle sockets warm would reintroduce the ~75s connect hang the PR exists to fix. keepAlive stays because supplying `httpAgentOptions` replaces the SDK's default agent (itself keepAlive: true), so `keepAlive: false` would open and close a socket per request - strictly worse - and it still reuses within a burst that flushes inside the timeout (e.g. a multi-batch force-flush). The one handshake per 10s against the collector is negligible for a long-lived process, and not worth weakening the anti-hang guarantee.

Corrected the overclaiming comment (and its twin in `otel-endpoint.test.ts`) in 3f38a2d70.

packages/telemetry/test/otel-unreachable.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Reversing my own "kept as is" — you were right, and the reason I gave for closing it (unreproducible across the sweep) no longer holds. Fixed in af2b06a6b.

Reproduced two ways. Naturally, in the isolated form `-t "re-sends once"` at load average 450-540: **1 failure in 40 runs**, on exactly this line, with a 610 ms first round trip.

```
× re-sends once on a retryable status and still stops inside the budget 610ms
AssertionError: expected [ '/v1/logs' ] to deeply equal [ '/v1/logs', '/v1/logs' ]
```

And deterministically, which is what I should have built instead of running the sweep harder: a 500 ms collector response delay stands in for a loaded runner's first round trip, and the old test then fails **20 of 20** — 13 on this line, 7 on the `Timeout`-masking mode in the thread below. Your derivation was exact, `backoff ~ U(800,1200)` against `1500 - t1`, which at t1 = 500 ms leaves the retry a coin flip on the jitter draw.

The fix removes the coin flip rather than widening the band or loosening the assertion. `getJitter()` is the only `Math.random()` consumer on the whole export path — `otlp-exporter-base`, `exporter-logs-otlp-http` and `sdk-logs` contain exactly one between them — so pinning it to the low draw perturbs nothing else. The retry then turns on the round trip alone and fits until 700 ms rather than 300 ms, and the second backoff, 1200 ms against at most 700 ms remaining, still cannot fire, so the exact two-request assertion stays exact. Same 500 ms delay against the new test: **20 of 20 pass**.

packages/telemetry/test/otel-unreachable.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Also reversing my own reply here. Both halves of your comment were right and I acted on neither. Assertion deleted in af2b06a6b.

On redundancy you were simply correct: `otlp-export-delegate` maps any final retryable status to the same string, so a single 503 with no re-send produces it too, and the request-count line above is the only thing that tells the two apart.

On the `Timeout` masking, reproduced. With the backoff pinned to its worst legal draw, the isolated test failed 1 in 20 at load ~400:

```
AssertionError: expected '{"stack":"Error: Timeout\n    at Time…' to contain 'retryable'
```

and deterministically, under a 500 ms collector delay, 7 of 20 failures land on this line rather than on the count. `_exportWithTimeout` arms `setTimeout(() => reject(new Error('Timeout')), 1500)` and routes that to `globalErrorHandler`, so once `t1 + backoff + t2` passes 1500 ms the exporter's real result never reaches `diag` at all.

Which makes it both the redundant assertion and the one with the tighter timing band, so deleting it is the entire fix rather than a concession. The diag channel keeps its coverage from the siblings: stubbing `captureExportErrors`' `error` callback to a no-op still takes three tests red at `ea9863484` (refused, 401, and now the 204 case).

packages/telemetry/test/otel-unreachable.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in ea9863484, reversing my own "left as is" above. The reason I gave — that a positive control means driving a second failing export through the channel, duplicating the 401 case — was only true of the expensive version of it. A direct write through the same channel costs nothing and duplicates nothing:

```ts
diag.error("capture probe");
expect(errors).toEqual(["capture probe"]);
errors.length = 0;
```

Verified with the same mutation that recorded the finding. Stubbing `captureExportErrors`' `error` callback to a no-op used to leave this test among the survivors (3 red: refused, 503, 401). At `ea9863484` it takes this test red too — the red set is now refused, 401 and this one, and it stayed at three only because the 503 case's `retryable` assertion was deleted in the same pass as redundant.

packages/telemetry/test/otel-unreachable.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Still accurate at ea9863484, re-measured. Recording that and resolving, since it is a cost to know about rather than a thing to fix.

Package suite without this file 0.85 / 0.90 / 0.91 s, with it 4.94 / 5.27 / 5.27 s, the file alone 5.13 / 5.15 / 5.17 s — the same shape as the original measurement, and the six `it`s still share one `describe` in one file, so it is still serial in a single worker whatever the core count. Pinning the 503 case's backoff to the low draw shaved a little off the retry wait without changing that.

PR #738, opened 2026-08-05
fix(flow): keep snapshot steps working where a full-res frame cannot be streamed

latekvo:
## Reproduced on a Pixel_9a emulator (API 37, 1080x2424)

Every `snapshot` step errors, so visual regression in flows is unavailable on this device:

```
$ argent flow run snap-test --device emulator-5554 --update-baselines
Flow "snap-test"
  › snapshot baseline seeding test
  ✗  1 snapshot "pokedex-home" — [Tool:screenshot] Screenshot failed: wrong data size, expected 7853760 got 17627328.

FAIL on emulator-5554 — 0 passed, 0 failed, 1 errored, 0 skipped
```

`--update-baselines` fails the same way, so there is no way to seed a baseline either — the
documented CI workflow (seed, review, commit `__baselines__/`) cannot start on this device.

Scale sweep isolates it to the unscaled request, and nothing else:

```
scale=0.3 ok   scale=0.5 ok   scale=0.75 ok   scale=0.9 ok   scale=0.99 ok   scale=1.0 FAIL
```

The device is not the limitation — `screen-recording` streams the same screen at native
1080x2424 h264 without complaint.

## What the two numbers are

Both come from the same frame. simulator-server's unscaled path hands the PNG encoder the frame
buffer whole, and the Android emulator sizes that buffer for a square of the screen's longer side
so a rotated frame still fits it (`max_dimension` in `android_emulator.rs`), then packs a `w × h`
frame at its head. So for 1080x2424 the encoder is handed `2424² · 3` = 17627328 bytes, wants
`1080 · 2424 · 3` = 7853760, and reports both. Every scale below 1 takes the resizer instead,
which reads only as far as the frame — the rest of the buffer is slack, which is why 0.3 through
0.99 all serve a correct image.

Nothing about the failure is a resolution limit, so nothing about it justifies a reduced capture.

## The fix

Retry at `REFUSED_CAPTURE_RETRY_SCALE`, one part in a million below 1. That is the whole margin
the condition needs: it clears simulator-server's own `1.0 - scale > f32::EPSILON` test for
"below 1", and it stays inside half a pixel of the screen's dimensions for any screen under
500_000 px a side, so `round(dimension × scale)` is the dimension itself and Lanczos3 between
equal sizes is an identity. The frame that comes back is byte-for-byte the frame the unscaled
request would have returned.

That matters beyond fidelity. The baseline key is capture geometry
(`pokedex-home__android-1080x2424`), so a reduced retry gives an affected host its own private
device class: it reports `no baseline for "<name>" on this device class` with the other host's
baseline sitting in the same directory, and following that advice commits a second permanent
file that neither host reconciles. Coming back at the screen's own resolution keeps one AVD one
device class.

**The retry names its own scale rather than inheriting the screenshot tool's default.** That
default is `ARGENT_SCREENSHOT_SCALE` when set — a knob for how much detail the *agent* sees.
Inheriting it would key every committed baseline on the value of the host that seeded it.

**Only the buffer-length refusal is retried** (`isPixelBufferSizeMismatch`, matched on the
server's message rather than on a failure code: the same wording arrives under
`SIMULATOR_SCREENSHOT_FAILED` in band and `SIMULATOR_HTTP_ERROR_RESPONSE` on a non-200, so no
static code identifies it). The scale is the only lever the retry has, so nothing else is worth
re-requesting for. Not even a cold frame stream, which *would* answer a second call —
`httpScreenshot` opens a first-frame window per call — because doubling how long a snapshot waits
for a frame is a separate decision from recovering a refused one.

**The refusal is learned once per device, not once per snapshot step.** The doomed unscaled
attempt is a real `screenshot` failure: registry `toolFailed`, a `tool:fail` telemetry event, a
line on stderr. Left unremembered, a flow with N snapshots reports N screenshot failures on an
affected device — on a run that passes, in exactly the metric someone would use to find this class
of problem. `ExecState` carries the device ids proven to refuse (the run-scoped shape it already
uses for `snapshotApps`, keyed by device because the verdict is the device's), and later steps go
straight to the retry.

**A cancelled capture is a `skip`, not an `error`.** An abort rejects a capture as an ordinary
network failure, which is not a size mismatch — so both attempts sit inside one `try` whose catch
decides on the signal rather than on the error, and a cancelled snapshot reports what every other
cancelled step reports.

**`screenshot-diff` retries at the same scale**, and gains the same pre-retry abort guard, so a
cancelled diff no longer dispatches a second capture on an already-aborted signal. It has the
identical fallback for the identical refusal and was landing on the env-overridable default (0.25
today), so a diff on an affected device compared quarter-size captures — the resolution a diff
exists to inspect. Its catch stays unconditional; see "Not fixed here".

## Checks

- **End to end on a real emulator.** A 1080x2424 AVD driven by a simulator-server built to
  reproduce the refusal: the unscaled request returns the reported
  `wrong data size, expected 7853760 got 17627328`, and `scale: 0.999999` returns 1080x2424
  bytes identical to what an unaffected server serves for the same screen (0 differing bytes of
  7853760; against an independent `adb exec-out screencap` reference, mean abs diff 0.58). Then
  the same flow through `flow-execute` on that device: `baseline written
  (pixel-home__android-1080x2424.png)`, and a second run `diff 0.00% ≤ 0.5%`.
- **The shipped binary is now held to the resize the scale rests on, in CI.** The unit suite's
  server stand-in *models* `round(dimension × scale)`; the binary is not in-tree (a rolling tag),
  so a build that truncated instead would flip every Android snapshot key with the suite still
  green. `scripts/e2e/drive-device.sh` captures unscaled and at the constant's own value — read
  out of `simulator-client.ts`, not restated — and fails on a dimension difference. It runs on the
  iOS simulator job, the only `e2e-device-smoke` cell on a simulator-server backend, which already
  triggers on `simulator-client.ts`. Green on this push: `unscaled=1206x2622  scale
  0.999999=1206x2622`. Locally on an iPhone simulator the two PNGs are byte-identical (0 differing
  bytes of 11360148 raw); against a stub server that truncates the check fails naming both
  dimensions, at 0.3 it fails at 324x727, and with the constant renamed it hard-fails rather than
  silently skipping.
- Full `packages/tool-server` suite: 360 files, 4504 pass / 1 skip. prettier, eslint
  (`--max-warnings 0`), `tsc --build`, `typecheck:tests`, `npm run knip` clean, and
  `flow-visual.test.ts` passes under `--sequence.shuffle` at seeds 1, 7 and 42.
- Each decision is pinned by a mutation over the three suites the change touches (56 tests). Retry
  scale to 0.3 fails 6; to 0.999 — still below 1, still ~full res, but half a pixel too far —
  fails the same 6, so the tests pin the dimension-preserving property and not merely "some
  scale". Dropping the mismatch guard fails 2; dropping the memo write fails 1; never consulting
  the memo fails 1; dropping the outer aborted-to-skip catch fails 1; dropping the pre-retry abort
  guard fails 1; dropping `screenshot-diff`'s abort guard fails 1; dropping the regex's `/i` fails
  1; dropping its `instanceof Error` fails 1; reverting `screenshot-diff`'s retry to the inherited
  default fails 1.

## Docs

No docs update needed. `reference/flow-yaml.mdx` ("Argent keys a baseline by platform and capture
geometry") and `features/visual-regression.mdx` ("when the resolution differs, Argent scales one
screenshot down") both stay accurate: the retry no longer changes capture geometry, so the
fallback introduces no resolution difference and no second device class for a page to explain.

## Not fixed here

Every agent-facing description of a comparison capture still promises full resolution, or tells the
agent to ask for `scale: 1.0` — the one call these devices refuse. That is one family spanning two
tools' parameter descriptions, the #689 resize hint and three skills, so it wants a single pass
rather than a qualifier fixed in one place and left in the others: **#838**.

`captureLiveInput` keeps its blanket catch: it re-requests after any capture failure, not only the
refusal. This changes what it retries *with* and stops it retrying a run already cancelled, but
narrowing it further changes what a cold frame stream or a dead backend reports there, which is
**#806**.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved screenshot capture reliability on Android emulators that reject full-resolution frames.
  - Automatically retries refused captures at a compatible scale while preserving image output.
  - Remembers emulator-specific capture behavior to avoid repeated failed attempts.
  - Handles cancelled runs cleanly without unnecessary retry attempts.
  - Updated screenshot comparisons to use the same recovery behavior.

- **Tests**
  - Added coverage for retry, cancellation, failure, per-device behavior, and successful capture scenarios.
  - Added end-to-end validation for consistent screenshot dimensions during recovery.


packages/tool-server/src/tools/flows/flow-visual.ts
```diff
@@ -200,10 +200,28 @@ export async function runSnapshot(
...
+  // instead of being compared against a full-res one. If the device is
+  // genuinely unreachable the retry fails too, and its error is what surfaces.
+  let shot: { image: ArtifactHandle };
+  try {
+    shot = (await invokeOnDevice(env, "screenshot", {
+      scale: 1.0,
+      includeImageInContext: false,
+    })) as { image: ArtifactHandle };
+  } catch {
+    shot = (await invokeOnDevice(env, "screenshot", {
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: On the fallback path the baseline key silently becomes a function of the ARGENT_SCREENSHOT_SCALE env var.

The retry omits `scale`, so the capture size is whatever `getScreenshotScale()` resolves: `httpScreenshot` does `const resolvedScale = scale ?? getScreenshotScale()` (simulator-client.ts:228) and `getScreenshotScale` returns `parseFloat(process.env.ARGENT_SCREENSHOT_SCALE)` when set, else 0.3 (simulator-client.ts:199-206). Because `snapshotKey` is built from the dimensions that came back (`const { w, h } = await pngDimensions(shot.image.hostPath)` / `${opts.name}__${env.device.platform}-${w}x${h}` at flow-visual.ts:231-236), a baseline seeded on a fallback device is only reusable by a tool-server running with the same value of that env var — an unrelated knob documented as controlling how much detail the *agent* sees (screenshot/index.ts:36). Before this change the snapshot capture was pinned at scale 1.0 and the key was independent of it. Two consequences: (1) changing or differing on ARGENT_SCREENSHOT_SCALE re-keys every committed baseline and the step reports `no baseline for "<name>" on this device class ...

**Trigger.** On an Android emulator where full-res capture fails (the PR's Pixel_9a), run `argent flow run <f> --update-baselines`: the fallback captures at the default 0.3 and writes `home__android-324x727.png`, which the user reviews and commits. Run the same flow against the same emulator from a tool-server started with ARGENT_SCREENSHOT_SCALE=0.5: the capture is 540x1212, the key becomes `home__android-540x1212`, and the step fails with "no baseline for \"home\" on this device class — expected .../home__android-540x1212.png, nothing was compared" despite an identical device class.

**Checked against.** flow-visual.ts:220-224 `} catch {\n    shot = (await invokeOnDevice(env, "screenshot", {\n      includeImageInContext: false,\n    })) as { image: ArtifactHandle };\n  }` — no explicit scale. simulator-client.ts:199-206 `export function getScreenshotScale(): number { const v = process.env.ARGENT_SCREENSHOT_SCALE; if (v) { const n = parseFloat(v); if (!Number.isNaN(n) && n > 0 && n <= 1) return n; } return DEFAULT_SCREENSHOT_SCALE; }`; simulator-client.ts:228,231 `const resolvedScale = scale ?? getScreenshotScale();` / `if (resolvedScale !== 1.0) body.scale = resolvedScale;`.

latekvo (commented):

packages/tool-server/src/tools/flows/flow-visual.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 48430f62e6afe9630e80e290db6bd0ac3e506cb4.

Reproduced end to end before fixing, with a vitest harness that drives the real `invokeOnDevice` → real `screenshot` tool → real `httpScreenshot` against a fake simulator-server standing in for the emulator (HTTP 200 + `wrong data size, expected 7853760 got 17627328` whenever the body carries no `scale`, i.e. the full-res request; otherwise a PNG at `round(1080*scale) x round(2424*scale)`). Seeding under `ARGENT_SCREENSHOT_SCALE=0.3` then re-running under `0.5` gave exactly the predicted split: `expected 'home__android-540x1212' to be 'home__android-324x727'`, and the re-run failed with `no baseline for "home" on this device class` on an identical device.

The retry now names its own scale (`FALLBACK_CAPTURE_SCALE = 0.3` in flow-visual.ts) instead of leaving it to `getScreenshotScale()`, so the key is a function of the device alone. 0.3 rather than some other value: that is where the retry already landed whenever the env var is unset, so nothing the PR verified on the device changes — only the env sensitivity goes away.

Both halves are pinned by `test/flows/flow-snapshot-capture-scale.test.ts`, which asserts the wire-level scales the fake server was asked for (`[undefined, 0.3]` under `ARGENT_SCREENSHOT_SCALE=0.5`) and that a baseline seeded at one env value still compares at another. Reverting only the `scale: FALLBACK_CAPTURE_SCALE` line fails both.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Swarm review pass came back with no HIGH and no MEDIUM findings. Four parallel agents covered claims-vs-code, nearest-twin + absence, what-outlives-the-call, and mutation-verification of the added tests (all pins fire: guard removal fails 2 tests, scale change fails 3, suffix removal fails 1; full production revert fails 5 of 8 new tests). The resolved thread's fix is verified at HEAD: `requested === [undefined, 0.3]` under ARGENT_SCREENSHOT_SCALE=0.5 through the real invokeOnDevice → screenshot tool → httpScreenshot path.

LOW findings posted for the record:

1. **PR body carried a stale numeric justification** — "0.3 is where the retry lands anyway with the var unset" was true when the branch was cut but #878 has since moved the tool default to 0.25, so the pinned constant now also fixes fallback dimensions against drift. Body corrected directly rather than commented.

2. **captureLiveInput's blanket catch** diverges from this PR's targeted `isFramebufferSizeMismatch` guard one module away — a dead server pays two connection-refused round-trips in screenshot-diff. Already tracked as #806; no action here.

3. **The agent-facing `screenshot` tool honors an explicit scale 1.0 with no fallback**, so affected devices still cannot take a full-res capture through that path. Arguably correct (silently downscaling an explicit request would under-deliver without saying so) and covered by the prose sweep in #838.

4. **No baseline pruning exists**, and this PR means every affected host now commits its own reduced-scale key set alongside full-res ones permanently. By design (keys are device-class identity), but a docs note on stale geometry keys would fit in the visual-regression page.

5. **Integration-test environment sensitivity**: on a host running a real `emulator-5554`, the Android rotation probe could enter the request and perturb the call-count assertions in flow-snapshot-capture-scale.test.ts. Fails soft today (adb probes resolve to null here); noting only.

packages/tool-server/src/tools/flows/flow-visual.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Re-verified at current HEAD (38d34e71, post-main merge): the flow suites run green — 40/40 across flow-visual.test.ts and flow-snapshot-capture-scale.test.ts, including the wire-level pin `requested === [undefined, 0.3]` under ARGENT_SCREENSHOT_SCALE=0.5 and the seed-at-one-env-value/compare-at-another case. The env-var sensitivity this thread flagged is not reachable in the merged code.

packages/tool-server/src/tools/flows/flow-visual.ts
```diff
@@ -294,7 +334,12 @@ export async function runSnapshot(
       });
       return {
         status: "pass",
-        reason: exists ? `baseline updated (${key})` : `baseline written (${key})`,
+        // Said where the baseline is adopted — what the gate checks is what the
+        // reviewer is deciding to commit — rather than on every later
+        // comparison, which on an affected device is every run.
+        reason:
+          (exists ? `baseline updated (${key})` : `baseline written (${key})`) +
+          (reducedScale ? ` at reduced scale — this device cannot stream a full-res frame` : ""),
```
hubgan: `reducedScale` is true on every path out of the fallback but is read only here. The branch an affected device reaches first is the other one: a host that can stream full-res commits `home__android-1080x2424.png`, the affected host falls back to 0.3, keys `home__android-324x727`, and reports `no baseline for "home" on this device class` with the full-res baseline sitting in the same directory. The device class is identical — only the capture scale differs — so the message names a cause that is not the one that happened, and following its own remedy commits a second permanent baseline. Nothing reconciles the pair afterwards, so the two hosts gate on disjoint files and neither exercises the other's.

packages/tool-server/src/tools/flows/flow-visual.ts
```diff
@@ -197,13 +209,41 @@ export async function runSnapshot(
...
+  let reducedScale = false;
+  try {
+    shot = (await invokeOnDevice(env, "screenshot", {
+      scale: 1.0,
+      includeImageInContext: false,
+    })) as { image: ArtifactHandle };
+  } catch (err) {
+    if (!isFramebufferSizeMismatch(err)) throw err;
+    shot = (await invokeOnDevice(env, "screenshot", {
+      scale: FALLBACK_CAPTURE_SCALE,
```
hubgan: At 0.3 a 1080x2424 screen becomes 324x727, and `cropPngFile`'s independent edge rounding then collapses thin elements. Running that arithmetic over 2000 vertical positions: a 1-physical-pixel element degenerates in 70% of positions (0.1% at full resolution), a 1dp divider at 2.625 dpr in 21%, at 3 dpr in 10%. Those land on `cropOn matched ... but its on-screen region is empty at this resolution`, which reads as a statement about the device — the element is not empty at the device's actual resolution. That return sits above the `updateBaselines` branch, making it the one snapshot outcome `--update-baselines` cannot get past, and it is the only crop path that never carries the reduced-scale note.

packages/tool-server/src/tools/flows/flow-visual.ts
```diff
@@ -197,13 +209,41 @@ export async function runSnapshot(
...
+  let shot: { image: ArtifactHandle };
+  let reducedScale = false;
+  try {
+    shot = (await invokeOnDevice(env, "screenshot", {
+      scale: 1.0,
+      includeImageInContext: false,
+    })) as { image: ArtifactHandle };
+  } catch (err) {
+    if (!isFramebufferSizeMismatch(err)) throw err;
+    shot = (await invokeOnDevice(env, "screenshot", {
```
hubgan: `diffPngFiles` renders the inline context diff at 0.3 of the compared image, so on this path the two scales compound. The image a client renders so an agent can see what differed comes back 97x218 for a 1080x2424 device, with the 4px changed-region outlines down to roughly one pixel.

packages/tool-server/src/tools/flows/flow-visual.ts
```diff
@@ -197,13 +209,41 @@ export async function runSnapshot(
...
+  // Only that mismatch is retried. Any other capture failure — an unreachable
+  // server, a frame stream that has not warmed up — answers a smaller request
+  // the same way, so a retry buys nothing and costs the baseline's identity: a
+  // transient that happened to clear would key the step off a resolution the
+  // device does not otherwise produce, and the miss would surface as the device
+  // class having no baseline. Those propagate instead.
+  let shot: { image: ArtifactHandle };
+  let reducedScale = false;
+  try {
+    shot = (await invokeOnDevice(env, "screenshot", {
```
hubgan: On an affected device this attempt fails on every snapshot step, and it fails as a real registry event before the retry recovers it: `toolFailed` is emitted in `registry.invokeTool`, the telemetry listener turns that into a `tool:fail` for `screenshot` carrying `SIMULATOR_SCREENSHOT_FAILED`, the tool-server event log writes it at error level, and the registry logger prints it to stderr. A snapshot step that passes therefore reports one screenshot failure per run in exactly the metric someone would use to find this class of problem.

packages/tool-server/src/tools/flows/flow-visual.ts
```diff
@@ -197,13 +209,41 @@ export async function runSnapshot(
...
+      scale: 1.0,
+      includeImageInContext: false,
+    })) as { image: ArtifactHandle };
+  } catch (err) {
+    if (!isFramebufferSizeMismatch(err)) throw err;
```
hubgan: No `env.signal?.aborted` re-check between the failed attempt and the retry, unlike the two settle guards earlier in this function. A run cancelled in that window dispatches a second device capture that then aborts mid-fetch and throws, and the runner maps a throwing snapshot to `status: "error"` while every other cancelled step reports `skip`.

hubgan (commented):
Reviewed the diff and the paths it touches, on iOS and Android.

I could not reproduce the framebuffer mismatch itself on this host — no API 37 system image is available, and a freshly created Pixel 9 AVD at the exact 1080x2424 plus five other local emulators all stream a full-resolution frame without complaint, including across forced rotations. So anything that depends on observing an affected device is left out here rather than asserted. What I did verify end to end, driving the real `flow-execute` against a device, is the unaffected path: seeding and then comparing a baseline works on Android (`settings-home__android-1080x2424`, `diff 0.00%`) and on iOS (`prefs-home__ios-1170x2532`, `diff 0.21%`), with a single full-resolution capture and no reduced-scale note.

No blocking findings. Everything below is either message and convention quality on the fallback path, or test and docs work.

packages/tool-server/src/tools/flows/flow-visual.ts
```diff
@@ -14,10 +14,22 @@ import {
...
+/**
+ * Scale a snapshot retries at when the device cannot stream a full-res frame.
+ *
+ * Stated here rather than left to the screenshot tool's default, which
+ * `ARGENT_SCREENSHOT_SCALE` overrides: the baseline key carries the dimensions
+ * the capture came back at, so an unscaled retry keys every committed baseline
+ * on that env var, and a host that set it differently is told it has no
+ * baseline for a device class it does have one for.
+ */
+const FALLBACK_CAPTURE_SCALE = 0.3;
```
hubgan: The comment above argues why the scale is pinned here rather than inherited, but not why it is 0.3. The scale sweep in the description has 0.5, 0.75, 0.9 and 0.99 all working on the affected device, so this is the loosest of the known-good values for a gate whose whole job is pixel comparison — it discards about 91% of the pixels that get compared. As written a reader cannot tell 0.3 from the screenshot-tool default it happened to match when the branch was cut.

packages/tool-server/src/tools/flows/flow-visual.ts
```diff
@@ -197,13 +209,41 @@ export async function runSnapshot(
...
-    scale: 1.0,
-    includeImageInContext: false,
-  })) as { image: ArtifactHandle };
+  //
+  // Full-res is preferred: it is the strictest comparison available. But some
+  // Android emulator configurations cannot stream a full-res frame — the
+  // simulator-server rejects it with a "wrong data size" framebuffer mismatch —
+  // so demanding one makes `snapshot` unusable on those devices, including
+  // under --update-baselines, where there is nothing to compare yet. Retry at a
+  // reduced scale, which captures reliably; `screenshot-diff` falls back for
```
hubgan: `screenshot-diff` does fall back for the same limitation, but not in the same way, and the paragraph five lines below states the opposite rule as this file's. `captureLiveInput` is a bare `catch {}` that retries every capture failure, and it retries with no scale at all, so it lands on the env-overridable default (0.25 today) rather than a pinned one. Following this pointer, a reader infers a parallel implementation that is not there.

packages/tool-server/src/tools/flows/flow-visual.ts
```diff
@@ -294,7 +334,12 @@ export async function runSnapshot(
       });
       return {
         status: "pass",
-        reason: exists ? `baseline updated (${key})` : `baseline written (${key})`,
+        // Said where the baseline is adopted — what the gate checks is what the
+        // reviewer is deciding to commit — rather than on every later
+        // comparison, which on an affected device is every run.
+        reason:
+          (exists ? `baseline updated (${key})` : `baseline written (${key})`) +
```
hubgan: This introduces a user-facing string no page describes, and the fallback splits a single AVD into two baseline device classes. `packages/docs/docs/reference/flow-yaml.mdx:237` ("Argent keys a baseline by platform and capture geometry") and `packages/docs/docs/features/visual-regression.mdx:41` do not mention either. `features/visual-regression.mdx:56` also now reads wrong for this path — it says a resolution difference is resolved by scaling one screenshot down, which holds for `screenshot-diff` but not for a flow snapshot, where the key gates before any normalization can happen. The repo CLAUDE.md asks for the docs change in the same pull request, or a note in the description that none is needed.

packages/tool-server/src/utils/simulator-client.ts
```diff
@@ -34,6 +34,25 @@ const NO_IMAGE_ERROR = /no image to export/i;
 export const FIRST_FRAME_WAIT_MS = 6_000;
 const FIRST_FRAME_POLL_MS = 250;
 
+// Some Android emulator configurations answer a full-resolution request with a
+// framebuffer size mismatch: the frame they stream is not the size the server
+// computed for the device, so it refuses to export it. The same device serves
+// every scale below full, which is what makes this the one capture failure a
+// caller can answer by asking for less.
+const FRAMEBUFFER_SIZE_MISMATCH = /wrong data size/i;
```
hubgan: `wrong data size` is not framebuffer-specific. `strings` on the shipped `packages/argent/bin/darwin/simulator-server` puts both variants of it (`wrong data size, expected ... got ...` and `wrong data size, got ... bytes too many`) inside the `png-0.17.16` crate's strings, next to `Invalid combination of bit-depth`, `Chunk is too short` and the `png-0.17.16/src/encoder.rs` / `decoder/stream.rs` paths. It is the png crate reporting a pixel buffer of unexpected length, raised on encode or decode, so a torn or mid-resize frame produces the same text as a transient — the case the comment block in flow-visual.ts says it is deliberately not retrying. There is also no platform gate, while the comment scopes the condition to Android emulator configurations.

packages/tool-server/src/utils/simulator-client.ts
```diff
@@ -34,6 +34,25 @@ const NO_IMAGE_ERROR = /no image to export/i;
...
+
+/**
+ * Whether a screenshot failure is the emulator refusing to export a full-res
+ * frame — the only one worth retrying at a lower scale.
+ *
+ * Matched on the server's message, not on the failure code: the code the
+ * mismatch carries (`SIMULATOR_SCREENSHOT_FAILED`) is shared with every other
+ * in-band capture error, the first-frame timeout included.
+ */
+export function isFramebufferSizeMismatch(err: unknown): boolean {
```
hubgan: The condition is classified by regexing an already-thrown error's message, while the site that mints that error is 280 lines below in this same file and its comment there already spells out `wrong data size` by hand. `FailureSignal.error_code` is documented as `Static code; never derive it from an Error message`, and the established shape for classify-or-rethrow is `getFailureSignal(err)?.error_code !== FAILURE_CODES.X` — `flow-run.ts:1325` is the same pattern in the same directory. This is also the only single-use export in this module (`NO_IMAGE_ERROR`, twelve lines above, stays private and is tested inline where the error is minted). The message form additionally couples the fix to wording emitted by a prebuilt binary, and no test here would catch a wording change: all four call-site tests hand-feed the literal.

packages/tool-server/test/flows/flow-visual.test.ts
```diff
@@ -697,3 +697,92 @@ describe("runSnapshot cropOn", () => {
...
+    // buys nothing — and a transient that happened to clear on the retry would
+    // key the step off a resolution the device does not otherwise produce,
+    // reporting a device class whose baseline is right there as having none.
+    const scaled = path.join(tmpDir, "scaled.png");
+    await writeFakePng(scaled, 324, 727);
+    vi.mocked(invokeOnDevice).mockClear();
+    vi.mocked(invokeOnDevice).mockRejectedValueOnce(
+      new Error("Screenshot failed: no image to export.")
+    );
+    vi.mocked(invokeOnDevice).mockResolvedValueOnce({ image: { hostPath: scaled } });
```
hubgan: This queues a resolved value the test never consumes — the assertion is that only one call happens — so the residue outlives the test. The block's `beforeEach` resets the mock, which contains it only while this describe stays last in the file and this test is not last in the block. `npx vitest run test/flows/flow-visual.test.ts --sequence.shuffle --sequence.seed=1` fails: `runSnapshot cropOn > crops exactly the frame's pixel rect from the capture` dies with `ENOENT ... /flow-visual-ZVtNTV/scaled.png` in `pngDimensions`, the leaked handle pointing into a tmpdir the file-level afterEach already removed.

packages/tool-server/test/flows/flow-snapshot-capture-scale.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    factory: async () => ({
+      api: { apiUrl: API_URL, streamUrl: "", pressKey: () => {} },
+      dispose: async () => {},
+      events: new TypedEventEmitter<ServiceEvents>(),
+    }),
+  });
+  registry.registerTool(createScreenshotTool(registry));
+
+  env = {
+    device: { platform: "android", id: "emulator-5554", kind: "emulator" },
```
hubgan, line 101: Driving the real `screenshot` tool with an Android device id also drives the real rotation probe: `captureScreenshotUpright` calls `readAndroidSurfaceRotation("emulator-5554")`, which tries `adbServerShell` first — a TCP connection to the local adb server, not the adb binary. On a machine with an emulator attached at that serial the suite reads that device on every capture. With such a device in landscape the run takes the wrong-aspect re-capture path and `requested` becomes `[undefined, 0.3, 0.3]`; I reproduced the failure at line 133 by pointing a copy of this file at a landscape emulator on this host. The file's header comment lists only the settle and the simulator-server as stood in for.

packages/tool-server/test/flows/flow-snapshot-capture-scale.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    expect(seeded.snapshotKey).toBe("home__android-1080x2424");
+
+    // ...and a host that cannot seeds its own rather than passing a downscaled
+    // capture off against it. Same message any new device class gets.
+    fullResFailure = { status: 200, error: "wrong data size, expected 7853760 got 17627328" };
+    const constrained = await runSnapshot(env, opts());
+
+    expect(constrained.snapshotKey).toBe("home__android-324x727");
+    expect(constrained.status).toBe("fail");
+    expect(constrained.reason).toContain('no baseline for "home" on this device class');
```
hubgan: `toContain` here and at line 122 leaves the deliberate omission of the reduced-scale note from the non-adoption paths unpinned: appending a marker to both the missing-baseline reason and the comparison reason when `reducedScale` is true leaves the whole tool-server suite green. Three more behaviours the changed code reaches are uncovered — a fallback combined with `cropOn` (the reduced-scale crop key, and the degenerate-region failure whose message literally says "at this resolution"), the retry itself hitting the same framebuffer mismatch a second time, and `isFramebufferSizeMismatch` directly, which has no unit test anywhere: dropping the `/i` flag and the `instanceof Error` guard together also leaves the suite green. Against that, four assertions here are duplicated by the new block in flow-visual.test.ts (full-res-then-0.3, the fallback key, the non-mismatch propagation, the full-res-host key) at strictly lower fidelity.

hubgan (approved):

latekvo:
This is a very bad approach.
We're explicitly given the upper bound for the snapshot, so arbitrarily rescaling it to 0.3 is strictly wrong, it neither guarantees success nor fulfills the agent's intention for highest possible res.

latekvo:
Fixed in 8eee3874f — you were right, and the mechanism turns out to make the fix free.

I traced the two numbers. Both describe the same frame: simulator-server's unscaled path hands
the PNG encoder the frame buffer whole, and the Android emulator sizes that buffer for a square
of the screen's longer side so a rotated frame still fits it (`max_dimension` in
`android_emulator.rs`), packing the `w × h` frame at its head. For 1080x2424 the encoder is
handed `2424² · 3` = 17627328 bytes and wants `1080 · 2424 · 3` = 7853760. Every scale below 1
takes the resizer instead, which reads only as far as the frame — the rest is slack. That is why
0.3 through 0.99 all worked.

So the bound is not a resolution at all — it is "anything below 1.0". The retry now goes to
`1 - 1e-6`: enough to clear the server's own `1.0 - scale > f32::EPSILON` test, and inside half a
pixel of the screen's dimensions for any screen under 500_000 px a side, so
`round(dimension × scale)` is the dimension itself and Lanczos3 between equal sizes is an
identity. Full resolution, not an approximation of it.

Verified on a 1080x2424 AVD against a simulator-server built to reproduce the refusal: the
unscaled request returns `wrong data size, expected 7853760 got 17627328`, and `scale: 0.999999`
returns 1080x2424 bytes identical to what an unaffected server serves for the same screen — 0
differing bytes of 7853760. Then the same flow through `flow-execute` on that device:
`baseline written (pixel-home__android-1080x2424.png)`, second run `diff 0.00% ≤ 0.5%`. On the
previous commit, same device and server: `baseline written
(pixel-home__android-324x727.png) at reduced scale`.

Two things fall out. The baseline key is capture geometry, so at full resolution an affected host
writes and gates on the same file every other host does, instead of splitting one AVD into two
device classes that never reconcile — which also removes the reduced-scale note, since there is
nothing reduced to disclose. And `screenshot-diff` had the same fallback landing on the
env-overridable default (0.25), so it now retries at the same scale; a diff on an affected device
was comparing quarter-size captures.

hubgan (commented):
Reviewed the retry end to end against a real device rather than by reading. Nothing here blocks the merge: the mechanism holds up, and everything below is comment accuracy, one residual gap in the abort handling, and test coverage.

What I verified first-hand, so the notes are grounded: with a proxy in front of a real 1080x2400 emulator answering the unscaled request exactly as the affected AVD does (HTTP 200, `wrong data size`), `snapshot` errors without the retry and writes `home__android-1080x2400.png` with it, and a second run diffs 0.00%. The same holds in landscape (`home__android-2400x1080`) and on iOS (`home__ios-1206x2622`). Against the shipped simulator-server the near-1 scale is not merely close: `scale: 0.999999` returns a PNG byte-identical to the unscaled capture on both platforms (same file md5), and a boundary sweep puts the server's rounding at `round`, not truncation (1080x2400 at 0.9996 gives 1080x2399, floor would give 1079). One thing I could not close from outside: whether the server's below-1 gate engages at exactly 1e-6, since identity dimensions and identity bytes look the same either way - that rests on the reading of the server source quoted in the constant's doc comment.

packages/tool-server/src/tools/flows/flow-visual.ts
```diff
@@ -197,13 +201,39 @@ export async function runSnapshot(
...
+  // Some Android emulators refuse the unscaled capture outright, which alone
+  // would leave `snapshot` unusable there — including under --update-baselines,
+  // where there is nothing to compare yet. The retry asks the same device for
+  // the same pixels down the one code path that does not trip on it, so such a
+  // device gates at its own resolution, on the baseline every other host writes.
+  // `screenshot-diff` retries at the same scale.
+  //
+  // Only that refusal is retried: the retry re-requests the frame at the
+  // resolution that just failed, so anything else — an unreachable server, a
+  // frame stream that has not warmed up — fails identically a second time.
```
hubgan: The cold-stream half of this justification does not hold. `httpScreenshot` computes its first-frame deadline inside the call (`const deadline = Date.now() + FIRST_FRAME_WAIT_MS`), so a second request gets a fresh six-second window and a stream that needed longer than one window would come back on it. Not retrying that case is defensible on the narrower ground already stated one line up - a scale only answers the size mismatch - but as written the comment picks the one example where re-requesting can change the outcome. The same wording is repeated in the test comment at test/flows/flow-visual.test.ts:742.

packages/tool-server/src/tools/flows/flow-visual.ts
```diff
@@ -197,13 +201,39 @@ export async function runSnapshot(
...
+    shot = (await invokeOnDevice(env, "screenshot", {
+      scale: 1.0,
+      includeImageInContext: false,
+    })) as { image: ArtifactHandle };
+  } catch (err) {
+    if (!isPixelBufferSizeMismatch(err)) throw err;
+    // Same guard the settles above use: without it a run cancelled in this
+    // window dispatches a capture that aborts mid-fetch, and a throwing
+    // snapshot is an `error` step where every other cancelled step is a `skip`.
+    if (env.signal?.aborted) {
```
hubgan, line 257: This covers only the window between the refusal arriving and the FailureError being constructed. An abort during the first capture rejects the fetch with an abort error, which is not a size mismatch, so line 225 rethrows and flow-run.ts:2308 maps it to `error`; an abort during the retry throws for the same reason. The asymmetry the comment describes - a snapshot reporting `error` where every other cancelled step reports `skip` - therefore still holds across the two much wider windows on either side of this check.

packages/tool-server/src/tools/screenshot-diff/index.ts
```diff
@@ -332,11 +332,10 @@ async function captureLiveInput(params: {
   signal?: AbortSignal;
   captureScreenshot: CaptureScreenshot;
 }): Promise<string> {
-  // Full-res gives the best diff fidelity, but some Android emulators reject a
-  // full-res frame ("wrong data size" framebuffer mismatch), which broke the whole
-  // baselinePath + captureCurrent flow there. The server's default scale captures
-  // reliably, and diffPngFiles' same-aspect normalization keeps a scaled capture
-  // comparable to a baseline saved at any scale.
+  // Some Android emulators refuse the unscaled capture outright, taking the
```
hubgan, line 382: The rewritten comment reads as the narrow, refusal-only retry described in flow-visual, but the catch below is unconditional: it fires for a dead backend, a cold frame stream and a cancelled run alike. It also drops the recorded reason for the change this PR actually makes on this path - that the retry now names its own scale instead of inheriting ARGENT_SCREENSHOT_SCALE. That reasoning now survives only in a test comment at test/screenshot-diff-tool.test.ts:173.

packages/tool-server/src/tools/screenshot-diff/index.ts
```diff
@@ -354,7 +353,7 @@ async function captureLiveInput(params: {
       params.device,
       params.rotation,
       params.signal,
-      undefined,
+      REFUSED_CAPTURE_RETRY_SCALE,
```
hubgan, line 415: flow-visual gained an aborted re-check between its two attempts in this commit; this site did not, so a cancelled diff still dispatches a second capture on an already-aborted signal. The two files now pin opposite contracts in tests added in the same commit: test/screenshot-diff-tool.test.ts asserts two attempts for a non-refusal failure, test/flows/flow-visual.test.ts asserts one.

packages/tool-server/src/utils/simulator-client.ts
```diff
@@ -34,6 +34,38 @@ const NO_IMAGE_ERROR = /no image to export/i;
...
+// two numbers. Every scale below 1 goes through the resizer instead, which
+// reads only as far as the frame.
+const PIXEL_BUFFER_SIZE_MISMATCH = /wrong data size/i;
+
+/**
+ * Whether a screenshot failure is the encoder rejecting the frame buffer's
+ * length — the one capture failure a caller can answer by naming a scale.
+ *
+ * Matched on the server's message, not on the failure code: the code it arrives
+ * with (`SIMULATOR_SCREENSHOT_FAILED`) is shared with every other in-band
```
hubgan: `SIMULATOR_SCREENSHOT_FAILED` is not the only code this message can arrive with. The `!res.ok` branch at line 313 formats the same server `error` field into `Screenshot failed: <msg>.` and mints it as `SIMULATOR_HTTP_ERROR_RESPONSE`, and the missing-fields throw at line 355 is also in-band under a third code. Matching on the message is what makes the non-200 variant match too, so the choice is right - the stated reason is just narrower than the real one.

packages/tool-server/test/screenshot-diff-tool.test.ts
```diff
@@ -169,10 +170,12 @@ describe("screenshotDiffTool", () => {
...
     );
 
-    // Full-res attempted first, then a default-scale retry without an explicit scale.
+    // Unscaled attempted first, then a retry at a scale of the tool's own —
+    // leaving it undefined would inherit ARGENT_SCREENSHOT_SCALE and diff a
+    // quarter-size capture, which is the resolution a diff exists to inspect.
     expect(captureScreenshot).toHaveBeenCalledTimes(2);
     expect(captureScreenshot.mock.calls[0]![3]).toBe(1.0);
-    expect(captureScreenshot.mock.calls[1]![3]).toBeUndefined();
+    expect(captureScreenshot.mock.calls[1]![3]).toBe(REFUSED_CAPTURE_RETRY_SCALE);
```
hubgan: This assertion imports the same constant it checks, so it holds for whatever value the constant takes - setting REFUSED_CAPTURE_RETRY_SCALE to 0.5 or to 0.999 leaves this test green. The comment three lines above says it guards against diffing a quarter-size capture, and it would not catch 0.25 either. The literal `1 - 1e-6` in the two flow tests is what pins the value today; the assertion the PR touched pins nothing.

packages/tool-server/test/flows/flow-visual.test.ts
```diff
@@ -697,3 +697,88 @@ describe("runSnapshot cropOn", () => {
...
+
+    expect(r.status).toBe("pass");
+    expect(vi.mocked(invokeOnDevice).mock.calls).toEqual([
+      [env, "screenshot", { scale: 1.0, includeImageInContext: false }],
+      [env, "screenshot", { scale: 1 - 1e-6, includeImageInContext: false }],
+    ]);
+    // The retry comes back at the device's own resolution, so the baseline it
+    // writes is the file every other host writes, under the key they key it by
+    // — and the reason says nothing a caveat-free capture would not say.
+    expect(r.snapshotKey).toBe("home__ios-390x844");
```
hubgan: The mocked invokeOnDevice returns the same `h.shotPath` (390x844) for both attempts, so this holds whatever scale the retry asks for - it would still pass if the retry captured at 0.1. The property the comment above describes is pinned by the call-argument assertion at line 720, not by this line.

packages/tool-server/test/flows/flow-snapshot-capture-scale.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    const body = JSON.parse(init.body) as { scale?: number };
+    requested.push(body.scale);
+    if (body.scale === undefined && unscaledFailure !== null) {
+      return new Response(JSON.stringify({ error: unscaledFailure.error }), {
+        status: unscaledFailure.status,
+      });
+    }
+    const file = path.join(shotDir, `shot-${requested.length}.png`);
+    const scale = body.scale ?? 1.0;
+    await writePng(file, Math.round(SCREEN_W * scale), Math.round(SCREEN_H * scale));
```
hubgan, line 84: The fake server sizes its reply with `Math.round(dimension * scale)`, which is the exact simulator-server behaviour the retry scale is chosen to exploit, so the suite asserts against the model of the server rather than the server. Nothing in the repo pins that contract: the binary is not in-tree and scripts/download-simulator-server.sh defaults to the rolling `radon-main` tag. A build that truncated instead of rounding, or that widened its below-1 gate past 1e-6, would flip every Android snapshot key or send the retry straight back into the refusing path, and this suite would stay green. I confirmed the current binary rounds (1080x2400 at 0.9996 returns 1080x2399) and returns a byte-identical frame at 0.999999, so this is about keeping that true, not about it being wrong today.

packages/tool-server/src/tools/flows/flow-visual.ts
latekvo: Fixed in fea3ec79b.

You are right: `httpScreenshot` computes `deadline` inside the call, so a second request opens a fresh `FIRST_FRAME_WAIT_MS` window and a stream that needed longer than one would come back on it. The comment no longer offers that as an example of a failure that repeats. It states the reason that does hold - the scale is the only lever the retry has - and records the cold-stream case for what it actually is: re-requesting would wait again, and doubling how long a snapshot waits for a frame is not a decision for a scale fallback to make on the side.

The test comment carried the same wording and was rewritten with it; the test is renamed to "does not retry a capture failure a scale cannot answer".

packages/tool-server/src/tools/flows/flow-visual.ts
latekvo: Fixed in fea3ec79b, by closing the windows rather than by rewording.

Both attempts now sit inside one `try`, and the catch decides on the signal instead of on the error: any throw on an aborted run returns `skip`, whether it came from the first capture, from the retry, or from the refusal itself. The pre-retry guard stays, so a cancelled run does not dispatch a second capture at all.

Pinned by mutation over the new tests: removing the outer catch fails 1, removing the pre-retry guard fails 1.

packages/tool-server/src/tools/screenshot-diff/index.ts
latekvo: Fixed in fea3ec79b.

The comment now says plainly that this catch is unconditional, unlike the flow snapshot's - it re-requests after any capture failure, not only the refusal - and that what that costs a cold frame stream or a dead backend is #806. The scale rationale is back on this path too: an omitted scale resolves to `ARGENT_SCREENSHOT_SCALE` (0.25 unset), a knob for how much detail the *agent* sees, and diffing quarter-size captures gives up the resolution a diff exists to inspect.

packages/tool-server/src/tools/screenshot-diff/index.ts
latekvo: Fixed in fea3ec79b: `catch (err)` with `if (params.signal?.aborted) throw err;` before the retry, so a cancelled diff no longer dispatches a second capture on an already-aborted signal. New test "does not re-request a capture the run was cancelled during" asserts one attempt; removing the guard fails it.

The two files still differ on *what* they retry - unconditional here, refusal-only there. That half is the deliberate one (#806), and the comment now states it instead of leaving a reader to infer a parallel implementation.

packages/tool-server/src/utils/simulator-client.ts
latekvo: Fixed in fea3ec79b. The doc comment now gives the real reason rather than the narrow one: no failure code identifies the message, because `httpScreenshot` formats the server's `error` field under `SIMULATOR_SCREENSHOT_FAILED` in band and under `SIMULATOR_HTTP_ERROR_RESPONSE` on a non-200, and both of those also carry capture failures no scale can answer.

packages/tool-server/test/screenshot-diff-tool.test.ts
latekvo: Fixed in fea3ec79b. The assertion spells out `1 - 1e-6`, like the two flow tests, and the import is dropped.

Checked by mutation: setting `REFUSED_CAPTURE_RETRY_SCALE` to 0.999 now fails this test - it passed before - and 0.3 fails it too.

packages/tool-server/test/flows/flow-visual.test.ts
latekvo: Fixed in fea3ec79b. The `snapshotKey` assertion is gone: `expect(r.reason).toBe("baseline written (home__ios-390x844.png)")` already carries the key, and neither line pins the scale - the call-argument assertion above does.

The comment no longer claims otherwise. It says what these lines do pin (the recovered capture is adopted with no caveat, indistinguishable from one the device served first time) and points at `flow-snapshot-capture-scale.test.ts`, whose server stand-in actually sizes its reply by the scale asked for.

packages/tool-server/test/flows/flow-snapshot-capture-scale.test.ts
latekvo: Fixed in 41a8ed83a.

`scripts/e2e/drive-device.sh` now captures twice on a booted device - unscaled, then at the constant's own value - and fails if the dimensions differ. It reads the scale out of `simulator-client.ts` rather than restating it, so the check follows the constant. It runs on the iOS simulator job, the only one there on a simulator-server backend (a scale below 1 on Chromium goes through the optional `sharp` dependency), and `e2e-device-smoke` already triggers on `simulator-client.ts` - so the file that owns the constant is the file that runs it.

Verified four ways: on an iPhone simulator through a real tool-server, unscaled and 0.999999 both return 1320x2868 and the two PNGs are byte-identical (0 differing bytes of 11360148 raw); against a stub server that truncates instead of rounding the check fails naming both dimensions; with the constant mutated to 0.3 it fails at 324x727; with the constant renamed it hard-fails rather than silently skipping. The unit suite's header now records that its `round(dimension x scale)` server is a model of the binary, and points here.

One half of your finding this does not close: the server's own below-1 gate staying at `f32::EPSILON`. On a device that serves the unscaled frame the gated and ungated paths are byte-identical, so nothing distinguishes them without an affected emulator.

packages/tool-server/src/tools/flows/flow-visual.ts
latekvo: Fixed in 8eee3874f. The fallback no longer captures at a reduced scale, so there is no second device class and no `reducedScale` flag: an affected host and an unaffected one both key `home__android-1080x2424` and gate on the same committed file. This branch is back to naming only the cause it can name.

packages/tool-server/src/tools/flows/flow-visual.ts
latekvo: Fixed in 8eee3874f. The retry comes back at the screen's own dimensions, so `cropPngFile` rounds against full resolution as it does on every other host, and "empty at this resolution" is reachable only for a region that really is empty at the device's resolution.

Re-confirmed on the shipped binary while adding the E2E pin: a 1320x2868 capture at `1 - 1e-6` is byte-identical to the unscaled one.

packages/tool-server/src/tools/flows/flow-visual.ts
latekvo: Fixed in 8eee3874f. The compared image is full-resolution on this path now, so `diffPngFiles` renders its context diff at 0.3 of the full capture - exactly what it renders on an unaffected host. The two scales no longer compound.

packages/tool-server/src/tools/flows/flow-visual.ts
latekvo: Fixed in fea3ec79b.

The failed invocation is real and the retry cannot un-happen it, but the refusal is a property of the device, so it is now discovered once per run instead of once per snapshot step: `ExecState` carries the device ids proven to refuse, in the same run-scoped shape it already uses for `snapshotApps`, and later steps go straight to the retry. A flow with N snapshots on an affected device reports one `tool:fail` rather than N, and saves N-1 doomed round trips (each one an unscaled encode of a 17 MB buffer).

Two tests pin it: that a second snapshot skips the unscaled attempt, and that a device the run moved onto re-proves the refusal for itself. Dropping the memo write fails one; never consulting it fails the other.

packages/tool-server/src/tools/flows/flow-visual.ts
latekvo: Fixed in 8eee3874f, and widened in fea3ec79b after your follow-up on this code: the guard alone only covered the window between the refusal arriving and the retry dispatching, so both captures now sit inside one `try` whose catch decides on the signal rather than on the error. An abort during either capture is a `skip`.

packages/tool-server/src/tools/flows/flow-visual.ts
latekvo: Fixed in 8eee3874f. The scale is no longer a reduction needing a justification: `REFUSED_CAPTURE_RETRY_SCALE` is `1 - 1e-6`, chosen for the only two properties it needs - below 1 by more than `f32::EPSILON`, and inside half a pixel of the screen's dimensions so `round(dimension x scale)` is the dimension itself - and the constant's doc says exactly that. Nothing is discarded: the frame is byte-identical to the unscaled one on the shipped binary.

packages/tool-server/src/tools/flows/flow-visual.ts
latekvo: Fixed in 8eee3874f and finished in fea3ec79b. `screenshot-diff` now does share the scale - the same exported constant, not the env-overridable default - and its comment states the part that is still not parallel: the catch there is unconditional, and narrowing it is #806.

packages/tool-server/src/tools/flows/flow-visual.ts
latekvo: Fixed in 8eee3874f. The user-facing string went with the reduced-scale capture, and the fallback no longer splits one AVD into two baseline device classes - so `flow-yaml.mdx:237` ("platform and capture geometry") and `visual-regression.mdx:41`/`:56` all stay accurate as written: the retry introduces no resolution difference for :56 to be wrong about. The PR description carries the "no docs update needed" note CLAUDE.md asks for, under ## Docs.

packages/tool-server/src/utils/simulator-client.ts
latekvo: Fixed in fea3ec79b. The comment now says the wording belongs to the `png` crate rather than to any one emulator, that a buffer wrong-length for another reason - a torn or half-resized frame - reads as this too, and that re-requesting is a fair answer to those as well. That is why the match stays as wide as the crate's message and carries no platform gate, which the previous wording implied without saying.

packages/tool-server/src/utils/simulator-client.ts
latekvo: Addressed in 8eee3874f, with the stated reason corrected in fea3ec79b.

The message match stays, for the reason your own follow-up on `simulator-client.ts:50` gives: the same wording arrives under `SIMULATOR_SCREENSHOT_FAILED` in band and `SIMULATOR_HTTP_ERROR_RESPONSE` on a non-200, so no static code identifies it and `getFailureSignal(err)?.error_code !== FAILURE_CODES.X` cannot express the condition. The doc comment now says that instead of the narrower claim.

It is also no longer a single-use export - `flow-visual` and `screenshot-diff` both take the constant, `flow-visual` takes the predicate - and it has direct coverage: dropping the `/i` flag fails a test, dropping the `instanceof Error` guard fails a test.

packages/tool-server/test/flows/flow-visual.test.ts
latekvo: Fixed in 8eee3874f, which rewrote this block: the unconsumed `mockResolvedValueOnce` is gone (every queued one-shot is now consumed by the test that queues it) and the block's `beforeEach` resets the mock.

`npx vitest run test/flows/flow-visual.test.ts --sequence.shuffle --sequence.seed=N` passes at N = 1, 7 and 42 - 39/39 each.

packages/tool-server/test/flows/flow-snapshot-capture-scale.test.ts
latekvo: Fixed in 8eee3874f. The file mocks `readAndroidSurfaceRotation`, so the suite opens no adb connection and reads no emulator this host happens to have at `emulator-5554`, and the header comment lists the rotation probe among what is stood in for.

packages/tool-server/test/flows/flow-snapshot-capture-scale.test.ts
latekvo: Addressed across 8eee3874f and fea3ec79b.

The reduced-scale note the `toContain`s left unpinned no longer exists - the retry adopts an ordinary baseline, so there is no caveat to omit, and the reason is asserted with `toBe` in both files.

`isPixelBufferSizeMismatch` has direct unit coverage now: dropping the `/i` flag fails a test, dropping the `instanceof Error` guard fails a test.

On the duplication: `flow-visual.test.ts` is down to what only a mock can express - call counts, error propagation, cancellation, and the once-per-device memo - and the key and scale assertions live in the file whose server stand-in sizes its reply by the scale asked for. The one assertion there that held for any scale is dropped (see the thread on line 727).

latekvo:
This sounds like a bug in sim-server, NOT something we should be fixing in Argent

latekvo:
Confirmed, this fix was made due to misconfigured scope. Will be reduced before handing off to review again.

latekvo:
Superseded by the root-cause fix in software-mansion/radon#210.

The failure was simulator-server handing the PNG encoder the whole Android capture buffer (a `max(w,h)²·3` rotation-headroom square) for a `w×h` frame; the encoder's exact-length check rejected the surplus as `wrong data size`. radon#210 slices the unscaled encode arm to the frame - matching the scaled arm, which already read only `w×h` - so no device refuses an unscaled capture and the client-side retry-scale workaround here is unnecessary.

argent picks the fix up through the rolling `radon-main` simulator-server asset with no code change. Closing in favour of the upstream fix.

PR #730, opened 2026-08-04
docs(skills): split argent-create-flow into references and add argent-qa-flows

hubgan:
> Now rebased directly onto `main`. #728 and #574 have landed, so this PR carries only its own 13 documentation commits; it still does not sit on the #715-#726 chain.
>
> Guidance describing behaviour owned by that chain is therefore **not** carried here: the recorder's refusal of a vacuous `hidden` and of `await-screen-idle` (#724), its cross-tree re-probe warning (#717), the per-step `recorded` line (#715), and `type:` failing on unconfirmed focus (#726). Each of those paragraphs states what this base actually does instead; they are worth restoring when that chain lands.

## `argent-create-flow`: ~5,000 words to ~900 plus three references

Every word of that skill loaded on every invocation, whether the task was replaying a two-step fragment or authoring a full regression test. The core skill now routes to three references read on demand:

- **`live-authoring.md`** - recording a walkthrough
- **`flow-yaml.md`** - the file format and selector vocabulary
- **`reliability-and-recovery.md`** - what to do when a step or a replay goes wrong

Nothing is dropped in the split - it moves behind a router.

## New skill: `argent-qa-flows`

Turning a **test case, ticket, or acceptance criteria** into a repeatable regression test is a different job from recording a path worth replaying, and it was being done by a skill that never asks for acceptance criteria. It orchestrates `argent-create-flow` as its engine, records the first walkthrough live, requires every requested screen and state to be proved with stable evidence, and completes only after the unchanged flow passes **twice consecutively**.

## Proving a navigation

Spelled out here for the first time, as two element-level checks rather than a route read:

```yaml
- await: { visible: { id: settings-screen } } # identity: WHICH screen
- await: { idle: true } # readiness: it stopped moving
```

Neither implies the other - a dropped tap leaves the source screen perfectly idle, and the destination's elements enter the tree while the transition still animates over them. Both belong after every screen change, including the one `launch:` performs.

## Routing

The three skills were being confused by name alone, so the rules now draw the line:

| you want | skill |
| --- | --- |
| a one-off interactive check, nothing saved | `argent-test-ui-flow` |
| a saved, replayable path | `argent-create-flow` |
| a saved test with acceptance criteria and two-pass proof | `argent-qa-flows` |

"Record a flow" also stops being ambiguous with **screen recording**, which is video.

## Tests

[...]

Issue #645, opened 2026-07-31
argent run prints a raw Zod JSON blob when a required flag is missing


filip131311:
## Symptom

Omitting a required flag on `argent run <tool>` surfaces the raw Zod validation error as JSON, rather
than a CLI-shaped message naming the missing flag and pointing at `argent run <tool> --help`.

## Why it matters

`argent run <tool> --help` produces good per-flag help, so the information needed for a decent message is
already there — the error path just doesn't use it.

## Suggested fix

Map validation failures to `missing required flag --<name>` (plus the `--help` hint), and keep the raw
issue list behind `--json` for programmatic callers.

## Related

#586 (boolean flags silently ignoring a space-separated value) is the same family of CLI argument
handling.

---
*Found during a full-surface QA sweep of **0.18.1** (`swmansion-argent-0.18.1.tgz`; the global install
is byte-identical to the tarball). Matrix: iPhone 17 Pro + iPhone 17 / iOS 26.5, Pixel_9 / Android 16
(API 36), Apple TV 4K / tvOS 26.5, Electron 42 (Chrome 148), Linux aarch64 (Lima) incl. the remote-link
leg, and the deviceless CLI. Host: macOS Darwin 25.5.0, Node v22.23.1.*

PR #739, opened 2026-08-05
fix(flow): give every flow in a directory run a verdict on stdout

latekvo:
## Every flow run says how it ended, on stdout

`argent flow run` wrote some outcomes only to stderr. A CI job that archives stdout — the
common case — got a log with a hole in it.

**A directory run**, three flows, one with a `run:` the parser rejects:

```
$ argent flow run <dir> > out.txt 2> err.txt ; echo $?
1

── out.txt ──                      ← the log that gets archived
[1/3] a-ok.yaml
  PASS — 0 passed, 0 failed, 0 errored, 0 skipped (no test steps)
[2/3] b-broken.yaml
[3/3] c-ok.yaml
  PASS — 0 passed, 0 failed, 0 errored, 0 skipped (no test steps)

FAIL — 3 flows: 2 passed, 1 failed, 0 skipped
```

`[2/3]` has a header and no outcome. The tally says one flow failed and stdout names none;
the reader's inference that `b-broken` was the one is available only because it is the sole
entry missing a line.

**A single-flow run** had the same gap, and worse — with live step lines the last thing the
log holds is a `✓`, so a run that died reads as one that passed and got truncated:

```
$ argent flow run checkout.yaml > out.txt 2> err.txt ; echo $?
1

── out.txt ──
Flow "checkout"
  ✓  1 tap
  ✓  2 assert                      ← ends here
```

## The fix

An indented verdict on stdout for every branch that had none, on both runners:

```
  ✗ not run (invalid flow)          ← rejected by validation; a batch continues
  ✗ not run (no device resolved)    ← nothing matched the run, or several did and none was singled out
  ✗ not run (rejected)              ← rejected for a reason neither of the above names
  ✗ did not finish (run error)      ← in a batch, this flow is the one that stopped it
  ✗ did not finish (no run report)
```

stderr is untouched, so anything reading that stream sees exactly what it did before, and the
detail is not duplicated onto the merged stream. Both runners print the verdict directly under
the header and the stderr detail after it, so a `2>&1` log of a batch reads like a `2>&1` log of
a single run. No verdict reaches `--json` — every one is behind `!args.json` — so neither runner
ever mixes prose into that stream. What each emits is unchanged: a batch still prints its one
aggregate document on every path, and a single run still prints the report alone, which means
nothing at all when the run produced no report.

### What a batch's `--json` carries

The aggregate is the mode a CI job consumes for a directory, so the classification the verdict is
chosen from travels with it: a failed entry carries `error_code` and `error_kind` beside `error`,
under the names `--json-stream` already publishes. Without them an invalid flow file and an
unresolvable device are the same `{"status": "fail", "error": "<prose>"}` entry with the same key
set, told apart only by a message the server assembles from device lists and remedies.

### Picking the wording from the failure code, not the kind

`error_kind: "validation"` marks any rejection the server scopes to one call; it says nothing
about the flow file. Device resolution rejects that way too (`FLOW_DEVICE_RESOLUTION`), so
keying the wording off the kind made a batch run with nothing booted report every one of a
directory of good YAML files as `not run (invalid flow)` — sending the reader to the wrong
file. `rejectionVerdict` switches on `error_code` instead: only a code whose subject is the
file (`FLOW_FILE_INVALID`, `FLOW_ENTRY_UNRECOGNIZED`, `FLOW_E2E_HAS_PREREQUISITE`) licenses
"invalid flow"; anything unrecognized stays neutral and leaves the reason to the stderr line
beside it.

`FLOW_NAME_INVALID` is deliberately absent. `argent flow run` only ever sends `flow_path`, and
both of its paths screen the stem against `SAFE_FLOW_NAME` first — which *is*
`FLOW_NAME_PATTERN`, the pattern `assertSafeFlowName` applies on the server — so the code
cannot come back from a flow run. `FLOW_PROJECT_ROOT_INVALID` is out for the same reason:
`project_root` is `process.cwd()`, always absolute and never `..`-bearing.

Which rejections let the batch continue is unchanged — still `errorKind === "validation"`.

### Also on the single-flow path

`!report || !("steps" in report)` took a primitive wire value straight into `in`, which throws
on one — `TypeError: Cannot use 'in' operator to search for 'steps' in a-string`, surfacing
through the CLI's top-level catch as a raw stack trace. The batch path guards this deliberately
("must classify as no report, not as an infra throw"); this one now does too.

### Docs

The help text said a directory run prints "only failing steps plus a final summary" and named
an invalid file as the only rejection it survives. It now describes what it prints and what it
continues past.

`packages/docs/docs/reference/flow-yaml.mdx` still described the batch as printing "only the
failing steps and a summary" and named an invalid file as the only rejection it survives. It now
describes what a directory run prints, tables the six outcome lines with their meanings, states
what the `--json` aggregate holds down to `error_code`, notes that a single-flow `--json` failure
writes nothing at all to stdout, and adds `--json-stream` to the option table — the mode that
carries the failure as a record.

Two claims that survived into the first pass of that page were wrong and are corrected here: a
flow that fails its *steps* also leaves the batch running (only a transport failure, a rejection
of another kind, or a reply that is not a report stops it), and the aggregate entry for such a
flow carries a report and no `error`. The `flow run --help` blurb carried the first of those two
as well.

`argent-create-flow` carried the same stale sentence when this branch opened, and an earlier
commit here fixed it too. #730 has since split that skill into `references/` and dropped the
standalone-runner paragraph from the tree, so there is nothing left to correct there and the
merge resolution takes the hunk out.

## Checks

- E2E through the built CLI (`packages/argent/dist/cli.js`) for every transcript above, plus
  the `--json` shapes piped through `python3 -m json.tool` to prove one document.
- `packages/argent-cli`: 512 passed across 24 files (`flow.test.ts` 148).
- Mutation matrix over the new behaviour, 24 mutations applied and reverted individually:
  deleting each of the five verdicts, dropping either `!args.json` guard, weakening
  `isFlowReport` in six ways (each conjunct dropped in turn, `every`→`some`, and reverts to
  both earlier readings), deleting each `rejectionVerdict` arm, swapping two wordings,
  dropping either `liveSteps === 0` header guard, dropping `stopped = true`, restricting the
  batch verdict to the first flow, and emitting a verdict on the success path. Every one
  reddens a test; no gaps.
- Reverting only the source hunks fails exactly the tests added here; no pre-existing test
  changes state.
- E2E against a real tool-server for the review follow-ups: a directory of one unparseable
  YAML plus one flow needing a device, run with the verdict and the stderr detail merged
  (`2>&1`) to show the order, and again under `--json` — `FLOW_FILE_INVALID` / `validation`
  arrive in the aggregate entry and stdout parses as one document. A `--platform vega` run with
  no VVD gives the real `not run (no device resolved)` line.
- Mutations for the follow-ups, each applied and reverted: dropping the failure signal from the
  aggregate, relaxing the batch guard to `!args.json || rejectedThisFlowOnly`, dropping
  `&& !args.jsonStream` from the single-flow guards, restoring the old detail-before-verdict
  order, and deleting the batch verdict. Every one reddens a test.
- prettier, eslint (`--max-warnings 0`), `tsc --build`, `tsc -p tsconfig.test.json`, knip and
  `npx docusaurus build` in `packages/docs` all clean.

Two findings from reviewing this branch are out of scope and filed instead: #1025 (`--json`
prints the usage text on stdout for a flag error, where `--json-stream` keeps stdout
machine-readable) and #1026 (a directory run's artifact export sits outside the loop's `try`, so
an export throw loses the batch summary, the skip entries and the aggregate).

Directory runs arrived in #576, so the batch half is a defect in a new feature rather than a
regression; the single-flow half predates it.

## Summary by CodeRabbit

* **New Features**
  * Added clear outcome lines for flow runs, including validation rejections, execution errors, missing reports, and skipped flows.
  * Batch runs continue after flow-specific validation rejections and stop when other errors occur.
  * Added machine-readable failure classifications to JSON output, including batch aggregates.
  * Added `--json-stream` for one JSON record per line during single-flow runs.
  * JSON modes now emit only machine-readable output, with aggregate details for directory runs.

* **Documentation**
  * Expanded flow command documentation with batch behavior, outcome statuses, and JSON output details.


packages/argent-cli/src/flow.ts
```diff
@@ -1133,13 +1133,25 @@ async function runFlowDirectory(
...
         err instanceof ToolInvocationError && err.errorKind === "validation";
+      // A verdict on stdout for every entry, next to the `[i/n]` header stdout
+      // already carries. The detail above goes to stderr, so without this line
+      // a redirected stdout log shows this flow's header followed by the next
+      // flow's — an entry that reads as if it never ran, while the final tally
+      // still counts it failed and names nothing.
+      if (!args.json) {
+        console.log(
+          `  ${STATUS_GLYPH.error} ` +
+            (rejectedThisFlowOnly ? "not run (invalid flow)" : "did not finish (run error)")
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: A device-resolution failure now prints "not run (invalid flow)" under every flow in the batch.

The new stdout verdict picks its wording from `rejectedThisFlowOnly`, i.e. `err instanceof ToolInvocationError && err.errorKind === "validation"` (flow.ts:1134-1135). But `"validation"` is not only the flow-file signal. `deviceResolutionError` raises it for a problem that has nothing to do with the YAML: packages/tool-server/src/tools/flows/flow-device.ts:78-89 returns `new FailureError(`${message} Available devices: ${list}.`, { error_code: FAILURE_CODES.FLOW_DEVICE_RESOLUTION, ... error_kind: "validation" })` for both "No booted device found. Pass a device id or platform explicitly." and "N booted devices matched — pass --device or --platform to disambiguate.". That throw comes from `resolveRunDevice` (flow-run.ts:789), which sits outside any try in `execute`, so it reaches the HTTP layer, where every tool-error response spreads `...errorSignalFields(err)` (http.ts:952/962/966) and therefore carries `error_kind`.

**Trigger.** With no simulator/emulator booted, run `argent flow run .argent/flows` over three ordinary device-touching flows. Each flow's `resolveFlowDevice` throws FLOW_DEVICE_RESOLUTION with error_kind "validation", so stdout reads `[1/3] a-login.yaml` / `  ✗ not run (invalid flow)` … for all three, while stderr carries "No booted device found. Pass a device id or platform explicitly. Available devices: none." Same for the opposite case: two booted devices and no --device/--platform.

**Checked against.** packages/argent-cli/src/flow.ts:1134-1145 — `const rejectedThisFlowOnly = err instanceof ToolInvocationError && err.errorKind === "validation";` … `console.log(`  ${STATUS_GLYPH.error} ` + (rejectedThisFlowOnly ? "not run (invalid flow)" : "did not finish (run error)"));`  ||  packages/tool-server/src/tools/flows/flow-device.ts:78-89 — `function deviceResolutionError(message: string, all: RawDevice[]): FailureError { … return new FailureError(`${message} Available devices: ${list}.`, { error_code: FAILURE_CODES.FLOW_DEVICE_RESOLUTION, failure_stage: "flow_device_resolution", failure_area: "tool_server", error_kind: "validation" }); }`  ||  packages/tool-server/src/tools/flows/flow-run.ts:789 …

latekvo (commented):

packages/argent-cli/src/flow.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Fixed in d5fa1676.

Confirmed, and reproduced end-to-end through the built CLI (`packages/argent/dist/cli.js`) before touching anything — three ordinary device-touching flows, nothing booted:

```
$ argent flow run .argent/flows > out.txt 2> err.txt ; echo $?
1
-- out.txt --
[1/3] a-login.yaml
  ✗ not run (invalid flow)
[2/3] b-checkout.yaml
  ✗ not run (invalid flow)
[3/3] c-logout.yaml
  ✗ not run (invalid flow)

FAIL — 3 flows: 0 passed, 3 failed, 0 skipped
-- err.txt --
[Tool:flow-execute] No booted device found. Pass a device id or platform explicitly. Available devices: ...
```

Three valid files, all blamed. The whitelist is wider than device resolution too: `HTTP_ZOD_VALIDATION_FAILED` (http.ts:721-724) and the capability gate also reach the CLI as `error_kind: "validation"`, so `errorKind` alone can never license a claim about the YAML.

The verdict now comes from `error_code`, and defaults to saying nothing about the cause:

- `FLOW_FILE_INVALID` / `FLOW_ENTRY_UNRECOGNIZED` → `not run (invalid flow)`
- `FLOW_DEVICE_RESOLUTION` → `not run (no device resolved)` (covers both directions — none booted, and several booted with no `--device`/`--platform`)
- anything else → `not run (rejected)`, leaving the reason to the stderr line beside it

`rejectedThisFlowOnly` is unchanged, so which rejections stop the batch is exactly as before.

Same repro after, with `--platform ios` to force the same `FLOW_DEVICE_RESOLUTION` throw deterministically (another agent had an emulator up on the box):

```
[1/3] a-login.yaml
  ✗ not run (no device resolved)
[2/3] b-checkout.yaml
  ✗ not run (no device resolved)
[3/3] c-logout.yaml
  ✗ not run (no device resolved)
```

and the PR's original repro (`run:` with an absolute path) still reads `✗ not run (invalid flow)` on stdout — that one is `FLOW_ENTRY_UNRECOGNIZED`, which is why it is in the whitelist rather than only `FLOW_FILE_INVALID`.

Pinned by a table-driven test over the four codes. Both mutations were checked: reverting the mapping to the old hardcoded string fails the `FLOW_DEVICE_RESOLUTION` and `HTTP_ZOD_VALIDATION_FAILED` rows, and dropping `FLOW_ENTRY_UNRECOGNIZED` from the whitelist fails that row. `flow.test.ts` 104 passed / 1 skipped; prettier, eslint `--max-warnings 0`, `tsc --build` clean.

latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Correction to my reply above, and a follow-up in 9460428e.

I claimed `HTTP_ZOD_VALIDATION_FAILED` also reaches the CLI as `error_kind: "validation"`. It does not. `http.ts:718-728` passes that signal to `emitHttpFailure` (telemetry) but answers with `res.status(400).json({ error: parseResult.error.message })` — no `error_code`, no `error_kind` on the wire. Same for the capability gate at `http.ts:753-762`. Only signals raised from inside tool execution reach the client, via `errorSignalFields`. The fix is unaffected — the switch keys on the code that does arrive — but the whitelist argument stands on `FLOW_DEVICE_RESOLUTION` alone, not on that second example.

Re-walking the codes that genuinely do arrive turned up one more mis-bucketed: `FLOW_E2E_HAS_PREREQUISITE`, thrown by `validateFlow` off `parseFlow` (`flow-utils.ts:2095-2105`, called at `flow-run.ts:677`). Reproduced in a batch:

```
[2/3] b-prereq.yaml
  ✗ not run (rejected)     <- a pure YAML-authoring error
```

It is as much "invalid flow" as a bad `run:` path, so it joins that arm. Same repro after: `✗ not run (invalid flow)`.

The same walk found the gap in the sibling path. `argent flow run <file>` — the single-flow runner — had exactly the defect this PR fixes for the batch: both failure branches (`flow.ts:1531-1534`, `:1536-1539`) wrote only to stderr. Verified through the built CLI, `--platform ios` with no iOS device:

```
--- STDOUT (0 bytes) ---
--- STDERR ---
[Tool:flow-execute] No booted ios device found. ...
```

Worse than the batch case it mirrors: with live step lines the last thing a redirected log holds is a `✓`, so a run that died reads as one that passed and got truncated. Now:

```
Flow "a-login"
  ✗ not run (no device resolved)
```

While there, `!report || !("steps" in report)` on that path took a primitive `data` straight into `in`, which throws — `TypeError: Cannot use 'in' operator to search for 'steps' in a-string`, out through `cli.ts:189` as a raw stack trace. The batch path guards this deliberately (`flow.ts:1127-1129`, "must classify as no report, not as an infra throw"); the single path did not. It now does.

Checks: `flow.test.ts` 113 passed / 1 skipped, `packages/argent-cli` 470 / 3 skipped. Mutation matrix over the new behaviour — deleting either single-flow verdict, dropping either `!args.json` guard, dropping the `typeof` guard, dropping the `FLOW_E2E_HAS_PREREQUISITE` arm, restricting the batch verdict to the first flow, and emitting a verdict on the success path — each one reddens a test and each was checked individually. prettier, eslint `--max-warnings 0`, `tsc --build` clean.

latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Re-verified on the current head (440c3567), independently of the reply above.

Built `packages/argent-cli/dist/` from this branch and drove the real `flow()` entry point in its own process against a real tool-server (`packages/tool-server/dist/index.js start`, port 45739), so the stdout/stderr split is the real one and the `error_code` travels the real HTTP path. `--platform vega` with nothing Vega running forces `FLOW_DEVICE_RESOLUTION` without touching any device on this box.

```
$ argent flow run ./batch --platform vega > out.txt 2> err.txt ; echo $?
1
-- out.txt --
[1/3] a-login.yaml
  ✗ not run (no device resolved)
[2/3] b-checkout.yaml
  ✗ not run (no device resolved)
[3/3] c-logout.yaml
  ✗ not run (no device resolved)

FAIL — 3 flows: 0 passed, 3 failed, 0 skipped
-- err.txt --
[Tool:flow-execute] No booted vega device found. Pass a device id or platform explicitly. Available devices: ...
```

The repro discriminates: reverting only `rejectionVerdict(toolErr?.errorCode)` to the original hardcoded `"not run (invalid flow)"`, rebuilding, and re-running the same command reproduces the reported defect verbatim (all three good files blamed), and reddens 2 of the 125 tests in `flow.test.ts`. Restored and re-run: green, and stdout back to `not run (no device resolved)`.

Same harness on the sibling path, which the follow-up commit covered:

```
$ argent flow run ./batch/a-login.yaml --platform vega
Flow "a-login"
  ✗ not run (no device resolved)
```

and `--json` on both runners: the batch emits exactly one document that `python3 -m json.tool` parses, the single-flow rejection emits 0 bytes on stdout — no prose leaks into either.

One correction to the second reply, for the record: it listed `FLOW_NAME_INVALID` among the codes licensing "invalid flow". That arm was deleted in f59a2103 (the CLI only ever sends `flow_path`, having already checked the stem itself, so the code cannot come back from a flow run) and the shipped whitelist is `FLOW_FILE_INVALID` / `FLOW_ENTRY_UNRECOGNIZED` / `FLOW_E2E_HAS_PREREQUISITE`. The PR body still carries the stale list; fixing it there.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Two review passes over this PR, plus the reviewer finding already on it. Recording what changed and what I decided not to change.

**Method.** Pass 1 ran seven agents, one per review move (claims-vs-code, nearest twin, non-happy paths, inputs, reachability both ways, what outlives the call) plus the absence pass. Pass 2 ran four, aimed at the fix pass 1 produced. Alongside both I drove the built CLI in its own process against a real tool-server, so the stdout/stderr split and the `error_code` round-trip are the real ones, and ran a mutation matrix against the suite. Nothing was reproduced on a device — `--platform vega` with nothing Vega running forces `FLOW_DEVICE_RESOLUTION` without touching one.

**Existing thread** (device resolution blamed on the flow file) — re-verified independently on head rather than taken on trust: the batch prints `not run (no device resolved)`, and reverting only `rejectionVerdict(toolErr?.errorCode)` reproduces the reported defect verbatim and reddens 2 tests. Resolved.

**Fixed here**

- `isFlowReport` admitted `{ steps: [null] }`. `Array.isArray` passed it, then `resolveArtifactDisplayPaths` — which runs *past* the try on both runners — threw on `s.artifacts`. In a batch that costs the whole ledger: no verdict for the flow, no summary, no tally, and the flows after it never run. That is the failure this PR exists to close, one level in. The guard now checks elements as well (b3b823842).
- The guard's nullish-data half was load-bearing and unpinned: deleting it left all 125 tests green while a `null` wire value crashed with an empty stdout. Pinned on both runners.
- The element check's `typeof step === "object"` half was likewise unpinned, and it is *not* redundant with `!!step`. A truthy primitive element throws nowhere — every field reads `undefined`, so the run prints a step line of them, tallies `PASS — undefined passed`, and exits **0**. A run that produced no report, reported as a success. Pinned on both runners (6853598d1).
- `rejectionVerdict`'s comment claimed the ambiguous device-resolution throw needs "no `--device`/`--platform`". It does not: `--platform ios` with two booted iOS simulators still lands there, because `scoped` is filtered by platform *before* the count check.
- A test named for a signal the server "leaves unclassified" was fed `errorKind: "subprocess"` — classified, just not as validation. Renamed to match the accurate wording its own sibling already used, and given a row for the genuinely absent kind a pre-signal server sends.
- PR body: it listed `FLOW_NAME_INVALID` among the licensed codes (that arm was deleted in f59a2103, and the code is genuinely unreachable — `SAFE_FLOW_NAME` *is* `FLOW_NAME_PATTERN`, and both CLI paths screen the stem first); claimed a SKILL.md change that #730 has since removed from the tree; carried the stale device-resolution parenthetical; and said `--json` "emits exactly one document on every path", which the single-flow rejection path contradicts by emitting zero. All corrected.

**Mutation matrix** — 24 mutations, applied and reverted individually: each of the five verdicts deleted, either `!args.json` guard dropped, `isFlowReport` weakened six ways (each conjunct in turn, `every`→`some`, and reverts to both earlier readings), each `rejectionVerdict` arm deleted, two wordings swapped, either `liveSteps === 0` header guard dropped, `stopped = true` dropped, the batch verdict restricted to the first flow, and a verdict emitted on the success path. Every one reddens a test; no gaps.

**Considered and not changed**

- `FLOW_E2E_HAS_PREREQUISITE` → `not run (invalid flow)`, though one of its two throw sites (`flow-run.ts`, the leading-`run:`-chain guard) can also be resolved with `--device`. For ios/android/vega `chromiumPinnable` is false, so the only remedies are YAML edits and the wording is exactly right; only the unpinned-chromium case has a third, invocation-level way out, and even there the two remedies the message leads with are YAML edits in the two files it names. Narrowing the arm would mislabel the far more common parse-time case to fix a rare one.
- A device-resolution rejection lets the batch continue, so a run with nothing booted burns one round-trip per flow. Pre-existing and unchanged here — `rejectedThisFlowOnly` is what it always was — and worth deciding on separately.
- `--json` on a rejected single-flow run prints nothing, where a batch prints its aggregate. Pre-existing, and deliberate: the single-run `--json` *is* the report, so prose would break `| jq` and a synthesised report would lie.

**Filed separately** — #823: `argent flow run --platform iOS` prints a raw Zod dump and aborts the batch. The CLI never checks the value, and `http.ts`'s pre-invoke 400 answers `{ error }` alone, dropping the `error_kind: "validation"` it just derived — so no client can classify it. Same class as #645, which #648 fixed for `argent run` only.

Green: prettier, eslint `--max-warnings 0`, `tsc --build`, `typecheck:tests`, 496 tests across 24 files, and all six CI checks on 6853598d1.

packages/argent-cli/src/flow.ts
```diff
@@ -1057,23 +1096,34 @@ async function runFlowDirectory(
...
+      // A verdict on stdout for every entry, next to the `[i/n]` header stdout
+      // already carries. The detail above goes to stderr, so without this line
+      // a redirected stdout log shows this flow's header followed by the next
+      // flow's — an entry that reads as if it never ran, while the final tally
+      // still counts it failed and names nothing.
+      if (!args.json) {
+        console.log(
+          `  ${STATUS_GLYPH.error} ` +
+            (rejectedThisFlowOnly
+              ? rejectionVerdict(toolErr?.errorCode)
```
hubgan, line 1086: A directory run's `--json` aggregate cannot express the distinction this line adds for humans. `BatchFlowResult` carries only `path`, `status`, `report` and `error`, so an invalid flow file and an unresolvable device both come back as `{"status": "fail", "error": "<server prose>"}` with identical key sets and the same exit 1 - I ran both against a directory and the only discriminator is the free-text message, which is assembled from device lists and remedies rather than being a contract. `toolErr` on the line above holds `errorCode` and `errorKind`, and they are dropped immediately after `results.push`. The single-flow `--json-stream` mode already puts those two fields on the wire as `error_code`/`error_kind`, so the directory aggregate - the mode a CI job would actually consume for a batch - is the one place the new classification stops at the human renderer.

packages/argent-cli/src/flow.ts
```diff
@@ -995,6 +996,43 @@ async function exportAndResolveArtifacts(
...
+  }
+}
+
+/**
+ * Whether a wire value is a report the renderers can walk. They run past the
+ * try on both runners, so a value this admits and they then throw on takes
+ * down the whole ledger rather than one line of it — hence the elements are
+ * checked too. A non-array `steps` throws on `for (… of report.steps)` and a
+ * nullish element on the first field read off it; a primitive element throws
+ * nowhere, which is worse — every field reads `undefined`, so the run renders
```
hubgan: The last clause of this paragraph does not hold. PASS and exit 0 come from the report's own top-level `ok`, which this predicate never inspects - driving the shipped renderers with `{steps: [5]}` and no `ok` prints `FAIL - undefined passed, ...` and exits 1, the opposite of what is written, while a well-formed report with `ok: true` prints PASS and exits 0 regardless of the elements. The step elements are causally irrelevant to the verdict.

"renders a step line of them" is also true of one runner only. `renderFailedSteps`, which is what the batch uses, skips a step whose status is neither `fail` nor `error` and carries no warning, so a batch prints no step line at all for that shape - it prints just the summary. The paragraph opens by saying the reasoning applies to both runners.

The guard itself is right and the hazard is real; it is the stated mechanism that is wrong.

packages/argent-cli/src/flow.ts
```diff
@@ -1038,7 +1077,7 @@ async function runFlowDirectory(
 
   const outputBase = args.output ? path.resolve(args.output) : undefined;
   const results: BatchFlowResult[] = [];
-  // A validation rejection is specific to one flow file, so the batch keeps
+  // A validation rejection is scoped to the one call, so the batch keeps
   // going. Anything else — transport death, or a failure the server didn't
   // classify — could make every remaining flow burn a device run against the
```
hubgan: "a failure the server didn't classify" names only one kind. `FAILURE_KINDS` has ten values, `unknown` is the unclassified bucket, and the code below stops the batch on any kind other than `validation` - so eight fully classified kinds stop it too. Reproduced: a flow whose leading `launch:` is a chromium target whose Electron spawn fails comes back as `CHROMIUM_ELECTRON_SPAWN_FAILED` with `error_kind: "subprocess"`, and the batch stops with the remaining flows skipped.

The `runFlowDirectory` docstring above says the same thing ("transport throw, unclassified failure, non-report result"), and both lines were rewritten in this PR. The test fixture added here states the rule correctly - `errorKind: "subprocess"` under a comment reading "a kind other than validation" - so the source comments and the test comments now disagree.

packages/argent-cli/src/flow.ts
```diff
@@ -106,9 +106,10 @@ chromium (a lone \`{ chromium: ... }\` target, or --platform chromium); a
 multi-platform launch auto-detects a device instead. Pass --device to attach to
 a running instance.
 
-A directory run prints only failing steps plus a final flow summary;
---recursive walks subdirectories too (dot-directories and node_modules are
-skipped). An invalid flow file fails alone and the batch continues; an infra
+A directory run prints each flow's failing steps and its outcome, then a final
+flow summary; --recursive walks subdirectories too (dot-directories and
+node_modules are skipped). A flow the server rejects up front — an invalid file,
+or a device it cannot resolve — fails alone and the batch continues; an infra
```
hubgan: `packages/docs/docs/reference/flow-yaml.mdx:269` still carries the sentence this hunk replaces: "A directory run prints only the failing steps and a summary. An invalid file fails alone and the batch continues." Both halves are now stale - a directory run also prints a per-flow outcome, and a device it cannot resolve fails alone too, which is what the second commit on this branch is about.

That line is the only place in `packages/docs` that describes batch behaviour (nothing in `features/` or `reference/cli.mdx` covers the directory form), and `git diff --stat origin/main...HEAD -- packages/docs packages/skills` is empty. The repo CLAUDE.md asks for the matching `docs/reference/` page in the same PR, or a line in the description saying no docs update is needed. The five new verdict strings are documented nowhere, and `--json-stream` is missing from that page's option table - which matters here because it is the only mode carrying `error_code`.

packages/argent-cli/src/flow.ts
```diff
@@ -1424,10 +1474,29 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
...
+    // The same stdout verdict a directory run gives every entry. Live step
+    // lines make the gap worse here: the last thing a redirected log holds is
+    // a passing step, so a run that died reads as one that passed and got cut
+    // off. The header is the flow's name, printed by the first step event or
+    // here when the run failed before any of them. Streaming mode owns stdout,
+    // so no prose verdict there either.
+    if (!args.json && !args.jsonStream) {
+      if (liveSteps === 0) console.log(`Flow "${flowName}"`);
+      console.log(
+        `  ${STATUS_GLYPH.error} ` +
```
hubgan, line 1459: The two runners put the same event onto a merged stream in opposite orders. Here the verdict is printed before `fail()` writes the detail to stderr; the batch writes the detail first and the verdict after. Captured on the same rejection:

```
$ argent flow run ./flows/b-broken.yaml 2>&1
Flow "b-broken"
  x not run (invalid flow)
[Tool:flow-execute] Unrecognized flow entry ...

$ argent flow run ./flows 2>&1
[2/3] b-broken.yaml
[Tool:flow-execute] Unrecognized flow entry ...
  x not run (invalid flow)
```

Cosmetic, and neither stream alone is affected, but the two ledgers are otherwise deliberately the same format.

packages/argent-cli/test/flow.test.ts
```diff
@@ -1717,6 +1885,133 @@ describe("argent flow run <dir>", () => {
...
+    // as though it never ran, while the tally still counts a failure and names
+    // no flow.
+    const headers = lines.filter((l) => /^\[\d+\/\d+] /.test(l));
+    expect(headers).toHaveLength(2);
+    for (const header of headers) {
+      expect(lines[lines.indexOf(header) + 1]).toMatch(/^ {2}\S/);
+    }
+    // A ledger nobody can read backwards: a ✗ appears only for the flow that
+    // earned one, so completeness above cannot be met by verdicting everything.
+    expect(lines.filter((l) => l.startsWith("  ✗ "))).toHaveLength(1);
```
hubgan, line 1955: The comment reads this as "a verdict marks the runs that earned one, so completeness cannot be met by verdicting everything", but a failed step line carries the same `  x ` prefix - `renderStepLine` emits `  x  2 run sub.yaml ...`. On a real batch of one step-failing flow plus one rejected flow, `grep -c '^  x '` returns 2. The assertion holds here only because the other fixture flow passes cleanly, so it is not the negative control the comment describes. Same applies to the copy at line 1956.

packages/argent-cli/test/flow.test.ts
```diff
@@ -1717,6 +1885,133 @@ describe("argent flow run <dir>", () => {
...
+    expect(lines).toContain("  ✗ not run (invalid flow)");
+    // stdout on its own is a complete ledger: every `[i/n]` header is followed
+    // by an indented outcome. Without one, a redirected stdout log shows this
+    // flow's header immediately followed by the next flow's — the entry reads
+    // as though it never ran, while the tally still counts a failure and names
+    // no flow.
+    const headers = lines.filter((l) => /^\[\d+\/\d+] /.test(l));
+    expect(headers).toHaveLength(2);
+    for (const header of headers) {
+      expect(lines[lines.indexOf(header) + 1]).toMatch(/^ {2}\S/);
```
hubgan: Same prefix collision from the other side: a failed step line satisfies `/^ {2}\S/` as well, so this checks that something indented follows each header, not that the flow's outcome does. On a real batch the line immediately after `[1/2]` is the failed step and the outcome is two lines further down, yet the assertion passes. Worth saying in the comment what the predicate can and cannot tell apart, since the verdict lines are not distinguishable from step lines by prefix.

packages/argent-cli/test/flow.test.ts
```diff
@@ -1717,6 +1885,133 @@ describe("argent flow run <dir>", () => {
...
+  });
+
+  // Every verdict is stdout prose, so --json must emit the aggregate alone —
+  // one `| jq` away from a parse error otherwise.
+  it("keeps --json a single document when a flow returns no report", async () => {
+    toolsClientMock.callTool.mockResolvedValueOnce({
+      data: { flow: "a-login", notice: "prerequisite" },
+    });
+
+    await expect(flow(["run", flowsDir, "--json"], opts)).rejects.toThrow("process.exit:1");
```
hubgan, line 2089: The batch `--json` guard is exercised on the no-report branch here, and on the rejection branch only by the pre-existing test that rejects with a plain `Error("boom")` - which takes the infra arm, not the rejection arm. No test runs a directory with `--json` and a `validation`-kind rejection, so relaxing the guard to `if (!args.json || rejectedThisFlowOnly)` leaves the whole suite green while prose leaks into the aggregate.

packages/argent-cli/test/flow.test.ts
```diff
@@ -1515,6 +1515,174 @@ describe("argent flow run", () => {
...
+    expect(logs.join("\n")).not.toContain("✗");
+  });
+
+  // A single run's --json is the report itself, so a run that produced none
+  // has nothing to print: stdout stays empty rather than gaining the prose
+  // verdict, which would be the one thing `| jq` cannot survive.
+  it.each([
+    ["produces no report", { data: { flow: "checkout", notice: "prereq" } }, "process.exit:2"],
+    ["is rejected", new ToolInvocationError("nope", { errorKind: "validation" }), "process.exit:1"],
+  ])("prints no prose on stdout under --json when the run %s", async (_case, outcome, exitCode) => {
```
hubgan, line 1678: This matrix pins the `--json` half of the single-flow guard, but both new verdict blocks are gated on `!args.json && !args.jsonStream` and nothing added here exercises the `--json-stream` half. Dropping that conjunct is still caught, but only by the pre-existing NDJSON tests that `JSON.parse` every stdout line - the guard this PR introduces has no test of its own.

hubgan (commented):
Reviewed the two changed files and the paths they reach, and drove the built CLI end to end for every transcript in the description (directory and single-flow, invalid flow, ambiguous device resolution, `--json`, `--json-stream`, merged streams) across iOS, Android, Chromium and Vega. The behaviour claimed here holds on all of them, `--json` stays a single document on every failure path, and the argent-cli suite is green (506 tests; flow.test.ts 142, run 130 times without a flake).

Nothing below blocks the merge - no correctness defect survived verification. What is left is one output-parity gap, three comment/doc inaccuracies, and a few coverage notes.

One thing to note outside the diff: the branch is behind main, which has since refactored `parseRunArgs` onto `parseCommandArgs` and un-exported `resolveArtifactDisplayPaths` under the knip ceiling drop in #770, so this will need another merge before its knip job matches main's.

hubgan (approved):

packages/argent-cli/src/flow.ts
latekvo: Fixed in 5d3ab0e44bf5a2f7e983b1a049fbf036bc52cbd3 - `BatchFlowResult` carries `error_code` and `error_kind` beside `error`, filled from a `failureSignal(err)` helper that `writeJsonStreamError` now uses too, so the two modes share one spelling rather than gaining a parallel shape.

Verified against a real tool-server started from the worktree (a directory of one unparseable YAML plus one flow that needs a device, driven through `flow()`):

```
$ ... run <dir> --platform ios --device NO-SUCH-DEVICE --json
{'path': 'a-bad.yaml', 'status': 'fail', 'error': '[Tool:flow-execute] Invalid flow file: ...', 'error_code': 'FLOW_FILE_INVALID', 'error_kind': 'validation'}
```

stdout still parses as one document. Two new rows cover it - `FLOW_DEVICE_RESOLUTION`/`validation` and `CHROMIUM_ELECTRON_SPAWN_FAILED`/`subprocess` - and both fail on the un-changed source with the two fields absent.

packages/argent-cli/src/flow.ts
latekvo: Fixed in 5d3ab0e44bf5a2f7e983b1a049fbf036bc52cbd3 - confirmed both halves by driving the shipped renderers before rewriting.

`renderSummary({flow, steps: [5]})` prints `FAIL - undefined passed, undefined failed, ...` and the run exits 1; adding `ok: true` to the same value prints `PASS ...` and exits 0, so the elements are causally irrelevant. `renderFailedSteps` returns `[]` for both shapes, while `renderReport` prints `  undefined  1 undefined`.

The docstring now says the verdict is the report's own `ok`, that `ok: true` exits 0 over steps no renderer can read, and names which renderer prints a line of `undefined` fields and which prints none. The guard itself is unchanged.

packages/argent-cli/src/flow.ts
latekvo: Fixed in 5d3ab0e44bf5a2f7e983b1a049fbf036bc52cbd3 - `FAILURE_KINDS` has ten values and the guard is `errorKind !== "validation"`, so the eight other classified kinds stop the batch alongside `unknown`, as does a transport throw carrying no kind at all.

Both comments state that now. The inline one reads "Any other kind - a boot failure, a timeout - stops it, as does a transport throw carrying no kind at all"; the `runFlowDirectory` docstring reads "a transport throw, a failure of any other kind, a result that is not a report". The `flow run --help` text carried the same claim ("an infra error stops the batch") and now reads "any other failure". Source and test comments agree.

packages/argent-cli/src/flow.ts
latekvo: Fixed in 5d3ab0e44bf5a2f7e983b1a049fbf036bc52cbd3 - `packages/docs/docs/reference/flow-yaml.mdx` is updated in this PR.

The stale sentence is replaced by what a directory run prints, a table of the six outcome lines with their meanings, what the `--json` aggregate holds (including `error_code`/`error_kind`), and a `--json-stream` row in the option table. `npx docusaurus build` in `packages/docs` and `npm run format` from the root are both clean. The PR description's Docs section is rewritten to match.

packages/argent-cli/src/flow.ts
latekvo: Fixed in 5d3ab0e44bf5a2f7e983b1a049fbf036bc52cbd3 - the batch prints the verdict first and `console.error(message)` after it, which is the single-flow order. Captured on a real merged stream (worktree tool-server, `2>&1`):

```
[1/2] a-bad.yaml
  ✗ not run (invalid flow)
[Tool:flow-execute] Invalid flow file: Missing closing "quote at line 2, column 27:
```

A new test captures both console spies into one ordered array and asserts `[["out", "[1/2] a-login.yaml"], ["out", "  ✗ not run (invalid flow)"], ["err", "flow file is not valid YAML"]]`. It fails on the previous order.

packages/argent-cli/test/flow.test.ts
latekvo: Fixed in 5d3ab0e44bf5a2f7e983b1a049fbf036bc52cbd3 - reproduced first: a batch of one step-failing flow plus one rejected flow prints `  ✗  1 assert - never visible` and `  ✗ not run (invalid flow)`, so the count is 2, exactly as you describe.

The comment says what the assertion is: the count discriminates only because the other fixture flow passes cleanly, since `renderStepLine` gives a failed step the same `  ✗ ` prefix that no count tells apart from a verdict. The copy under `renders a %s rejection as %j` carries the same note.

packages/argent-cli/test/flow.test.ts
latekvo: Fixed in 5d3ab0e44bf5a2f7e983b1a049fbf036bc52cbd3 - reproduced: with a step-failing first flow the line after `[1/2]` is `  ✗  1 assert - never visible`, and `/^ {2}\S/` still passes.

Rather than document the ambiguity I removed it. The assertion matches the outcome lines exactly - `toBe("  ✗ not run (invalid flow)")` under `[1/2]` and `/^ {2}PASS /` under `[2/2]` - neither of which a step line satisfies, and the comment says why it is matched exactly. Deleting the verdict `console.log` reddens it.

packages/argent-cli/test/flow.test.ts
latekvo: Fixed in 5d3ab0e44bf5a2f7e983b1a049fbf036bc52cbd3 - added a directory `--json` case over a rejection: an `it.each` with a `validation` row (`FLOW_DEVICE_RESOLUTION`) and a non-validation row, asserting `logs` holds exactly one entry and that `flows[0]` equals the full object including `error_code`/`error_kind`.

Checked with the mutation you named: `if (!args.json || rejectedThisFlowOnly)` reddens the validation row (stdout gains the prose line, so `logs` has two entries). On the un-changed source it stays green, as you said.

packages/argent-cli/test/flow.test.ts
latekvo: Fixed in 5d3ab0e44bf5a2f7e983b1a049fbf036bc52cbd3 - added `keeps stdout NDJSON-only under --json-stream when the run %s`, covering both branches the guard sits on (rejected, and no run report). Each asserts every stdout line parses and equals the exact `error` record, including `error_code`/`error_kind` on the rejection.

Dropping `&& !args.jsonStream` from the two guards reddens both new cases, alongside the pre-existing NDJSON tests.

packages/argent-cli/src/flow.ts
```diff
@@ -995,20 +1000,65 @@ async function exportAndResolveArtifacts(
...
+ * down the whole ledger rather than one line of it — hence the elements are
+ * checked too. A non-array `steps` throws on `for (… of report.steps)` and a
+ * nullish element on the first field read off it; a primitive element throws
+ * nowhere, which is worse — the verdict is the report's own `ok`, so `ok: true`
+ * exits 0 over steps no renderer can read: renderReport prints a line of
+ * `undefined` fields for each, and the batch's renderFailedSteps prints none.
+ */
+function isFlowReport(data: unknown): data is FlowReport {
+  const steps = (data as FlowReport | undefined)?.steps;
+  return Array.isArray(steps) && steps.every((step) => !!step && typeof step === "object");
```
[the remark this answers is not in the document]
latekvo: Checked and not acting on this one: the rendering is exactly as described, but no reply `flow-execute` can send reaches it.

Confirmed the behaviour first, driving the single-flow runner with `{ data: { ok: true, steps: [] } }` as the tool reply:

```
exit: process.exit:0
stdout: Flow "undefined"

         PASS - undefined passed, undefined failed, undefined errored, undefined skipped
```

So the description of what a bare value renders is right. What it needs is a producer, and there is none:

- `isFlowReport` only ever sees the `data` of a `flow-execute` reply (`packages/argent-cli/src/flow.ts:1111` for the batch, `:1483` for a single flow are its only paths), and `flow run` refuses any other server: `requireLocalToolServer` (`:933`) fails the run when `ARGENT_TOOLS_URL` or a link file is configured, so the peer is always the auto-started local one from the same bundle.
- That tool has two returns. `summarize(...)` (`packages/tool-server/src/tools/flows/flow-run.ts:1598-1613`) is one object literal that always carries `flow`, `device`, `ok`, all four counters and `steps` - and `summarize` takes `flowName: string` / `deviceId: string` as parameters, so the two strings are pinned by the signature. The other is `FlowPrerequisiteNotice` (`:1165-1170`), which has no `steps`; the guard already rejects it and `test/flow.test.ts` pins that as `did not finish (no run report)`.

The narrower contract is deliberate and the docstring above the guard states it: it admits values the renderers can walk, because both runners render outside the `try`, so a value that gets in and then throws takes down the whole ledger instead of one line of it. A non-array `steps` throws on the `for` loop and a nullish element on the first field read, which is why those two are rejected; `{ ok: true, steps: [] }` walks, so tightening would only re-label an input nothing emits.

On the twin: `isValidationIssue` (`packages/argent-cli/src/run-validation.ts:176`) does check every field, and its trigger is reachable - it runs on `JSON.parse(err.message)` (`:71-76`), so any tool whose error message happens to parse to a JSON array lands there. There is no comparable ambiguity about `data` on this path.

packages/docs/docs/reference/flow-yaml.mdx
```diff
@@ -292,7 +293,28 @@ argent flow run ~/shared-flows/checkout.yaml --device <UDID> --update-baselines
...
+| `not run (rejected)`             | The tool-server rejected the flow for another reason            |
+| `did not finish (run error)`     | The run failed before it produced a report                      |
+| `did not finish (no run report)` | The tool-server answered without a run report                   |
+| `not run (batch stopped)`        | An earlier failure ended the directory run. Directory runs only |
+
+`--json` and `--json-stream` print no header, no outcome line and no summary.
+
+A directory run with `--json` writes one document. The document holds `ok`, `total`, `passed`, `failed`, `skipped` and a `flows` array. The array holds one entry for each flow. Each entry holds the `path` and the `status`. An entry holds `report` when the flow produced one. An entry holds `error` when Argent called the tool-server for that flow and got no report. Such an entry also holds `error_code` and `error_kind` when the tool-server sends them. An entry with the status `skip` holds neither, because Argent made no call for that flow. `--json-stream` puts the same two fields on its `error` record.
```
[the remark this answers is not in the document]
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in a821baaaa. Confirmed against `runFlowDirectory`: `error` is pushed on the catch arm (flow.ts:1091 — a rejection *or* a run error, carrying `error_code`/`error_kind` from `failureSignal` when the err has them) and on the no-report arm (:1099, `error` only); an entry that produced a report — passing or failing its steps (:1111) — carries `report` and no `error`, and a `skip` entry (:1060) carries neither. So you are right that the old "got no report" undersold the contract, and it collided with the `did not finish (no run report)` outcome name two paragraphs up.

I did not take the suggested "when Argent called the tool-server and the flow failed" verbatim: a flow that fails its *steps* also "failed" yet carries a report rather than `error`, so that phrasing would overstate it. The line now enumerates the three no-report cases (a rejected run, a run error, a reply that is not a report) and pins `report` to a produced report; `error_code`/`error_kind` stay conditional on a server signal and the `skip` distinction is preserved. `npx docusaurus build` and `prettier --check` clean.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Follow-up pass.

**Reviewer findings.** The one open thread was CodeRabbit's on `flow-yaml.mdx:315`: the `--json` aggregate said an entry holds `error` when Argent "got no report", which undersells the contract and collides with the `did not finish (no run report)` outcome name two paragraphs up. Fixed in a821baaaa — the line now enumerates the three no-report cases (a rejected run, a run error, a reply that is not a report), pins `report` to a produced report (passing or failing its steps), and keeps `error_code`/`error_kind` conditional on a server signal and `skip` holding neither. I did not take the suggested "when … the flow failed" wording: a flow that fails its *steps* also failed yet carries a report, not `error`. `docusaurus build` and `prettier --check` clean; all eight CI checks green on a821baaaa. Every other thread was already resolved.

**Fresh review.** Seven agents, one per review move — claims-vs-code, nearest-twin, non-happy-paths, inputs, reachability both ways, what-outlives-the-call, and the absence pass (including the mutation) — over the diff against the merge base, reading the real source and the server throw sites, with a full `flow.test.ts` run (149 pass). Any H/M candidate would have gone to a three-lens adversarial verify; none was raised. No HIGH, MEDIUM or LOW surfaced. Independently re-derived: all four `rejectionVerdict` arms are reachable and carry `error_kind: "validation"` from their throw sites (`FLOW_FILE_INVALID`, `FLOW_ENTRY_UNRECOGNIZED`, `FLOW_E2E_HAS_PREREQUISITE`, `FLOW_DEVICE_RESOLUTION`); `isFlowReport` is strictly safer than the old `"steps" in` guard, admits an empty `steps: []`, and rejects the primitive-element shape that would otherwise exit 0 over steps no renderer can read (pinned on both runners); `error_code`/`error_kind` travel both the buffered and the streamed HTTP paths; the batch↔single-flow divergences (`--json-stream` gating, exit codes) are each justified.

One pre-existing item, out of scope and already filed as #1026: the batch-path `exportAndResolveArtifacts` (flow.ts:1105) runs past the loop's `try`, so an export throw escapes `runFlowDirectory` with no summary, no skip entries and no aggregate — byte-identical to the pre-PR baseline and untouched by this diff.

Nothing here blocks the merge.

packages/docs/docs/reference/flow-yaml.mdx
```diff
@@ -292,7 +293,28 @@ argent flow run ~/shared-flows/checkout.yaml --device <UDID> --update-baselines
...
+| `not run (rejected)`             | The tool-server rejected the flow for another reason            |
+| `did not finish (run error)`     | The run failed before it produced a report                      |
+| `did not finish (no run report)` | The tool-server answered without a run report                   |
+| `not run (batch stopped)`        | An earlier failure ended the directory run. Directory runs only |
+
+`--json` and `--json-stream` print no header, no outcome line and no summary.
+
+A directory run with `--json` writes one document. The document holds `ok`, `total`, `passed`, `failed`, `skipped` and a `flows` array. The array holds one entry for each flow. Each entry holds the `path` and the `status`. An entry holds `report` when the flow produced one, whether it passed or failed its steps. An entry holds `error` instead when the tool call produced no report — a rejected run, a run error, or a reply that is not a report. Such an entry also holds `error_code` and `error_kind` when the tool-server sends them. An entry with the status `skip` holds neither, because Argent made no call for that flow. `--json-stream` puts the same two fields on its `error` record.
```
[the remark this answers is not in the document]
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in 90e062c5 — split into two sentences, each under the one-idea-per-sentence limit, keeping the `error`-vs-`report` "instead" contrast that pairs with the `report` line above it. `docusaurus build` and `prettier --check` clean.

PR #743, opened 2026-08-06
refactor(tool-server): use @swmansion/argent-cloud-sdk for MoQ + input protocol

jwajgelt:
## Summary

argent and the other client of the same device backend each hand-rolled the same MoQ client and `DataChannelCommand` protobuf encoder, with no shared source — argent carried a vendored copy of the `.proto` as an inline string that could silently drift from the server schema. Both now use `@swmansion/argent-cloud-sdk`, where the encoder is checked byte-for-byte against the canonical `.proto` in tests, so drift fails the build instead of shipping.

- **`utils/moq-client.ts`** (195 → ~50 lines) is now a thin adapter over the SDK's `MoqDeviceSession`. It keeps the `MoqClient` interface exactly (`sendControl`, `screenshot` → `Buffer`, `close`), so `blueprints/simulator-server.ts` and `createMoqTransport` are untouched. argent still resolves its endpoint through `sim-remote moq-info` (preserving daemon-based session sharing) and still publishes on its own `"argent"` broadcast path.
- **`utils/datachannel-proto.ts`** (201 lines → re-exports) so its ~10 import sites keep working unchanged.
- **Dependencies:** drops `protobufjs` and `@moq/lite`, both now unused. `@moq/net` is kept and still dedupes to a single copy.

No behavioural change intended.

## Verified end-to-end against live infrastructure

Ran the migrated client against a real remote simulator (iPhone 17 Pro, iOS 26.5) on a live staging machine, driving this branch's `moq-client.ts` through the published SDK **0.1.2**:

| Check | Result |
|---|---|
| MoQ session open (WebTransport, cert-pinned) | **PASS** |
| Screenshot | **PASS** — valid PNG 1206×2622 |
| Sequential screenshots | **PASS** |
| Concurrent screenshots | **PASS** |
| Input round-trip (tap an app icon) | **PASS** — tapping Settings opened Settings, confirmed visually in the returned screenshot |
| Screenshot after concurrent burst | **PASS** |
| Session teardown | **PASS** |

Plus the existing suite: full workspace `tsc --build`, **3372 tests passing** (1 skipped), and `npm ls @moq/net` confirming a single deduped copy.

### A bug this surfaced, now fixed

Concurrent `screenshot()` calls hung under SDK **0.1.0**. The server answers one screenshot per session at a time; 0.1.0 had dropped the serialisation the old hand-rolled client had, so a second in-flight request was never answered and its promise never settled — a regression against pre-SDK behaviour that only showed up under live traffic, with no unit test catching it.

Fixed in SDK 0.1.1 (queue requests, plus a 30s request timeout so a dropped response errors instead of hanging). This PR depends on `^0.1.2` — a republish of 0.1.1 through the release workflow carrying npm build provenance, whose shipped `dist` is byte-identical (verified by SHA256), so the live results above hold for it unchanged.

## Notes for reviewers

- The SDK subscribes to the screenshot track lazily, so a video-only consumer doesn't open a subscription it never reads. argent requests screenshots, so this path is exercised.
- Screenshot responses are matched by the request id the server echoes, so a late or duplicate frame can't be handed to the wrong waiter.


packages/tool-server/src/utils/moq-client.ts
```diff
@@ -86,109 +33,20 @@ export interface MoqClient {
...
-  let screenshotSeq = 0;
-
-  const api: MoqClient = {
-    async sendControl(payload: Uint8Array): Promise<void> {
-      const track = await getControlTrack();
-      track.writeFrame(payload);
+  return {
+    sendControl: (payload) => session.sendControl(payload),
+    async screenshot(opts) {
+      return Buffer.from(await session.screenshot({ scale: opts?.scale }));
```
latekvo, line 46: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Routing through `session.screenshot()` brings in the SDK's `ScreenshotChannel`, which adds a 30s request timeout (`DEFAULT_REQUEST_TIMEOUT_MS`; `MoqDeviceSession` constructs the channel with two args, so it can't be overridden from here). When that timeout fires the pending entry is dropped and the queue advances. A late answer is then matched against the id the server echoes, but `dispatch()` falls back to the oldest waiter whenever the frame carries no string `id`:

```js
const id = typeof payload.id === "string" ? payload.id : undefined;
const index = id === undefined ? 0 : this.pending.findIndex((p) => p.id === id);
```

So against a server that does not echo the id, the slow answer to request #1 is handed to request #2 as its result. `createMoqTransport.screenshot` writes those bytes to disk and returns `{ url, path }` with nothing marking them stale, so the agent acts on a picture of the device from more than 30s earlier.

What I ran: argent's compiled `openMoqClientFromInfo` + `createMoqTransport` over a real `@moq/net` session (`createMockTransportPair` + `Connection.connect`/`accept`), against a fake simulator-server that answers request #1 at 33s, with the real 30 000 ms timeout — plus the pre-PR client transcribed verbatim from `6d907cc8` as a control against the same server:

```
new, server does NOT echo id:  #1 = MoQ screenshot timed out after 30000ms   #2 = IMG-for-req-1
new, server DOES echo id:      #1 = MoQ screenshot timed out after 30000ms   #2 = times out on its own
pre-PR client, no echo:        #1 = IMG-for-req-1                           #2 = IMG-for-req-2
```

The whole thing therefore rests on the deployed simulator-server echoing `ScreenshotCommand.id`. The only statement about that in this repo is the comment this PR deletes a few lines up — `// (server has no per-request id)`, added in 20f5a128 and unchanged since — and the SDK keeps an explicit fallback for "a server too old to echo it". Nothing in the diff or the test suite establishes which holds for the servers `sim-remote moq-info` resolves to. (Separately and in argent's favour: on that same 33s-late server the pre-PR client hangs forever, so the timeout itself is an improvement — it is only the oldest-waiter fallback behind it that is new.)

packages/tool-server/src/utils/datachannel-proto.ts
```diff
@@ -1,200 +1,25 @@
...
- * Kept inline so we don't need a build step to ship the .proto file alongside
- * the compiled tool-server. If you bump the simulator-server schema, regenerate
- * this string from the canonical .proto.
+ * Encoders for the `datachannel.DataChannelCommand` messages simulator-server
+ * reads off the MoQ control track.
+ *
+ * These now live in `@swmansion/argent-cloud-sdk`, alongside the canonical
+ * `.proto` and a test that checks the encoder against it, so the clients of this
+ * protocol can no longer drift apart. Re-exported here so the existing import
+ * sites keep working.
```
latekvo, line 8: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The inline `PROTO_SOURCE` copy is gone, but it was not the only vendored schema in this package — `packages/tool-server/src/proto/datachannel.proto` was added alongside it in 20f5a128 and is untouched by this PR. It has already drifted: diffing it against `node_modules/@swmansion/argent-cloud-sdk/proto/datachannel.proto` shows it missing `TouchPointer`, `TouchStateCommand` and the `touch_state = 7` arm of the `DataChannelCommand` oneof.

Nothing imports it (`git grep -e 'src/proto' -e 'datachannel\.proto'` returns no hits at this head), so there is no runtime effect. The impact is that the file a reader is most likely to open when looking for argent's copy of this schema now describes a protocol the server has since extended, while the comment here states that the copies can no longer drift apart — which holds for the SDK's encoder and its canonical `.proto`, but not for the one still sitting in this package.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat)\]: Reviewed end-to-end rather than by reading. What I ran against `189c4790`, all green unless noted below: workspace `tsc --build`; `npm test --workspaces` (305 files / 3372 passed, 1 skipped in tool-server, matching the PR body, plus 87 / 282 / 31 elsewhere); `typecheck:tests`, `typecheck:scripts`, `test:scripts`, prettier and eslint on the changed files; the Lockfile CI gate (`npm install --package-lock-only` leaves `package-lock.json` unchanged); `npm ls` confirming a single deduped `@moq/net@0.1.9` and no `protobufjs` / `@moq/lite` left in the tree; and `packages/argent/scripts/bundle-tools.cjs`, which inlines the SDK and `@moq/net` into `dist/tool-server.cjs` with no unresolved `require`, keeping only the existing `@fails-components/webtransport` dynamic import external.

Wire parity old-vs-new: 1627 vectors across all six encoders (every enum value including the proto3 zero cases, doubles at 0/-0/NaN/±Infinity/1e308/subnormals, float32 rounding, unicode and empty screenshot ids, `secondX`/`secondY` present/absent), old side driven by the merge-base's own `PROTO_SOURCE` through `protobufjs@7.6.4` (the version the base lockfile pins). Identical bytes everywhere except `encodeKey` with a code outside int32 range or a negative non-integer/NaN — none of which any caller can produce, since codes only ever come from `charToKeyPress` (4–56), `NAMED_KEYS` (40–82), `SHIFT_KEYCODE` and the hardcoded CMD/V, so I am not raising it.

Behaviour: drove the published-shape esbuild bundle through a real `@moq/net` session (`createMockTransportPair` + `Connection.connect`/`accept`) against a fake simulator-server, with only `globalThis.WebTransport` faked — `connectMoq`, cert pinning, the handshake, `MoqDeviceSession` and `ScreenshotChannel` all real. Token appended to the URL, fingerprint decoded and pinned as one sha-256 hash, one transport attempt (no WebSocket race), screenshot round-trip byte-exact, touch/button/key/rotate decoding correctly server-side against the canonical `.proto`, and screenshot-after-close rejecting rather than hanging. Screenshot serialisation confirmed on the wire (5 concurrent calls, max 1 request in flight at the server, each caller getting its own image). Control-frame drops under an unpaced burst are identical before and after (6 of 7 lost in both, all 7 delivered in both once paced), so that is not a regression here. Two things below.

packages/tool-server/src/utils/datachannel-proto.ts
jwajgelt: [79655b6](https://github.com/software-mansion/argent/commit/79655b6f6ceb59ba221d8436a9b8a4462bafed95)

packages/tool-server/src/utils/moq-client.ts
jwajgelt: The deployed `simulator-server` instances now send the ids.
It should be fine as is, we don't expect the screenshot to ever take this long -- the moq stream would disconnect before the timeout triggered on network failure, and the server-side screenshot operation itself is fast.

jwajgelt: Updated the SDK to reflect that the `id` is always present.

filip131311 (approved):

latekvo (approved):

PR #582, opened 2026-07-28
feat(tool-server): add human-readable log message

sjchmiela:
## Summary

Adds tool-owned, human-readable interaction messages for every tool definition.

Each tool defines its own started, completed, and failed wording (as implemented in https://github.com/software-mansion/argent/pull/367. Messages can use validated parameters, successful results, and safe failure codes. They are to avoid sensitive input such as typed text, cookie values, JavaScript expressions, and URL credentials.

Bridges the gap between current Argent

Screenshot%202026-07-28%20at%2015 48 37

and current agent-device

Zrzut ekranu 2026-07-28 o 20 14 42

## Test

Selected fields from the generated JSONL:

```jsonl
{"level":30,"type":"tool.invoked","toolId":"list-devices","toolInvocationId":"decb864f-51c1-41ee-91fc-15bc2494e268","msg":"Listing devices","time":"2026-07-29T10:37:19.896Z"}
{"level":30,"type":"tool.completed","toolId":"list-devices","toolInvocationId":"decb864f-51c1-41ee-91fc-15bc2494e268","durationMs":96.77333299999918,"msg":"Listed 22 devices and 0 AVDs","time":"2026-07-29T10:37:19.993Z"}
{"level":30,"type":"tool.invoked","toolId":"screenshot","toolInvocationId":"8504abb8-83ed-4490-8210-3355d6c638c9","msg":"Capturing screenshot","time":"2026-07-29T10:37:24.503Z"}
{"level":50,"type":"tool.failed","toolId":"screenshot","toolInvocationId":"8504abb8-83ed-4490-8210-3355d6c638c9","failureSignal":{"error_code":"SIMULATOR_SERVER_READY_EXITED","failure_stage":"simulator_server_spawn_ready","failure_area":"tool_server","error_kind":"subprocess","failure_command":"simulator_server","failure_exit_code":1},"durationMs":750.4639999999999,"msg":"Failed to capture screenshot: SIMULATOR_SERVER_READY_EXITED","time":"2026-07-29T10:37:25.254Z"}
{"level":30,"type":"tool.invoked","toolId":"screenshot","toolInvocationId":"63b938fc-3d86-4eb5-9972-a008474ffb01","msg":"Capturing screenshot","time":"2026-07-29T10:38:01.107Z"}
[...]

PR #744, opened 2026-08-06
refactor(flow): let each step kind own how it reads

latekvo:
Stacked on #574 (merged; retargeted at `main`).

## Summary

`summarizeSteps` (flow-finish-recording.ts) was a 50-line `switch` over `FlowStep["kind"]`. `stepTarget` (flow-run.ts) was a second one, over the same union, with its own selector and condition spellings. Neither could be found from the other, and `summarizeSteps`' `case "tool": default:` arm meant a step kind added without wording rendered as a `tool:` step instead of failing.

Both surfaces now read from one definition per kind — the same shape as a tool owning its own log wording in `ToolDefinition.interaction` (#582):

```ts
const FLOW_STEP_DEFINITIONS: {
  [K in FlowStep["kind"]]: FlowStepDefinition<Extract<FlowStep, { kind: K }>>;
} = {
  "scroll-to": {
    summary: (step) => `${yamlSelectorLabel(step.target)} (${step.direction})`,
    target: (step) => { ... },
  },
  ...
};
```

Keying the record by `FlowStep["kind"]` is what replaces the default arm: a new kind is a compile error until it says how it reads on **both** surfaces. That also retires the `renderToolArgs` interpolation guard, which existed only because the default arm could hand it a step with no `args` field.

`flow-run.ts` sheds a net 113 lines and five imports; `flow-finish-recording.ts` sheds a net 180 and is now just the tool. The two shared label templates (`conditionLabel` parameterized by selector spelling, and a new `whenLabel` for the platform-vs-UI guard arm) mean the recorder's `when:` tail and the runner's no longer duplicate the platform arm.

## Test

The extraction is rendering-neutral: every summary line and report target the switches produced moved over byte-for-byte (with four later corrections, below).

- Proved with a throwaway equivalence test that inlined both pre-refactor switches verbatim and compared them against the new definitions over 1264 generated steps: all 15 kinds, 10 selector shapes (loose, regex, scoped, quote-bearing), every wait condition × text-match mode × expectation, every scroll direction, and the cyclic-`args` path. Identical on all of them. Mutating two renderings (`equals` comparator, default scroll direction) failed it, so it discriminates.
- Replaced by `packages/tool-server/test/flows/flow-step-definitions.test.ts`, kept: assertions pinning the exact summary line and report target of every kind, in a table keyed by `FlowStep["kind"]` so a new kind fails `typecheck:tests` here too (verified by deleting the `wait` entry: `TS1360`). Deleting a kind from the definitions record fails the build the same way (`TS2741`).
- `npm run build`, `eslint --max-warnings 0`, `prettier --check .`, `npm run typecheck:tests --workspaces`: clean.
- `packages/tool-server`: 329 files / 3906 tests passed. `packages/argent-cli`: 459 tests passed.

## Review follow-up (9cf51322)

Four corners where the carried-over rendering misdescribed what replays, fixed with pinning tests:

- `delayLabel`: a hand-edited `delayMs` above `setTimeout`'s 2³¹−1 ceiling sleeps a clamped 1ms tick but read `(after 3000000000ms)` — the summary now describes no delay there.
- `scroll-to`: the `within` container is parsed, serialized, and changes what replays, yet neither surface rendered it; both do now.
- `snapshot`: the summary now carries `cropOn` and `maxMismatch` (baseline key and tolerance). Only `cropOn` was on the report target; `maxMismatch` was rendered on neither surface (the run report surfaces the tolerance through the step's `reason`, not its target).
- `type`: the summary JSON-quotes the text, the module's own convention for literals, so embedded quotes/newlines stay one unambiguous line.

## Review follow-up (7c804907)

A second pass (7 lenses + a 32-mutation pass that also re-derived the neutrality claim against main over a 136-step corpus, and confirmed the mapped-type compile gate by adding a kind) found no HIGH or MEDIUM. One rendering defect this PR introduced, plus the boundaries the case table left open:

- `scroll-to`'s report target spelled the scroll container ` within (…)` — exactly how `selectorLabel` spells a selector's own relational scope. `scroll-to: { target: Footer, within: { id: list } }` (scroll that container) and `scroll-to: { target: { text: Footer, within: { id: list } } }` (match only inside it) both rendered `"Footer" within (id=list)`; carrying both put the direction between two identical-looking scopes. It now reads ` in scroll container (…)`, the noun flow-actions already uses in its own "scroll container … is not visible" failure. `main` did not collide here only because it dropped `within` entirely.
- New pinning cases, each verified to fail without the code it pins: `setTimeout`'s ceiling from both sides (`2**31-1` keeps its label, `2**31` loses it) and its 1ms floor; `maxMismatch: 0`, the one value separating `!== undefined` from a truthiness test; `cropOn` + `maxMismatch` together, which fixes their suffix order; and a multi-field selector, which fixes the report target's field order.
- `Number.isFinite(ms) &&` in `delayLabel` became redundant the moment correction #1 added the ceiling — NaN and ±Infinity both fail the range test. Removed (proved a no-op by mutation).
- Prose corrected where it had stopped matching the code: the point-gesture comment argued a criterion the file does not follow (`snapshot` and `scroll-to` are equally not recorder-built and this PR renders their fields anyway) and listed two of the four still-unrendered replay-affecting fields; `flow-tools.test.ts` still claimed to be the only coverage of the long-press rendering; and the new case table justified a no-op `as number` cast with a reachability claim that is wrong in both halves (`flow_add_step`'s `delayMs` is `z.number().int().min(0)` with no ceiling, so the recorder reaches these values too).

Follow-ups filed rather than folded in, both pre-existing and unchanged here: #808 (an `echo` message or `tool` name containing a newline breaks the one-line-per-step summary contract, the same hazard correction #4 fixed for `type`) and #809 (`type.submit`, `await.timeout`, `idle.timeout`, `idle.stableFor` change what replays and render on neither surface).

## Review follow-up (4560f412)

Round from @hubgan's review (11 inline) plus CodeRabbit's one. Nine acted on, two verified and rejected — the rejection reason is now in the code, so it stays answered rather than being re-raised.

**Rendering**

- `type`'s report target carried the typed text with **no bound**. Unlike the summary, that target is emitted on every replay — stdout, `--json`, the MCP blocks handed to the agent, and the appended `~/.argent/mcp-calls.log`, which has no cap or rotation — so a literal written into a flow file shipped its whole length to each. Now capped at 200 chars with the same `…(+N chars)` elision `parseFlow` uses on the entry it echoes (`MAX_ENTRY_RENDER_CHARS`). The **summary keeps the text whole**: it is also `stepAnchor`, so eliding there would collide two steps differing only past the cap and drop the recording's verdicts on both. Elides before quoting, so a cut cannot halve an escape.
- `selectorLabel` and `describeSelector` interpolated `textMatches`, `identifier` and `role` raw — a newline in one splits a step across two report lines, an escape byte drives the terminal. That is the hazard `text`'s quoting already names, so it now covers its neighbours: new `escapeInline` (flow-utils) spells out control characters **only**. Not JSON escaping: doubling a backslash would print a regex source nobody could copy back into the .yaml. `describeSelector`'s literal arm moved from a raw `"${v}"` to `JSON.stringify` — byte-identical for ordinary values, unambiguous for a quote in the content.

**Types**

- `FlowStepDefinition`'s two renderers were method shorthand, and a method's parameters are checked **bivariantly even under `strict`**: a definition annotated narrower than its slot (one requiring `times` on `tap`) was accepted and then read a field the step need not carry, with a green typecheck. Property types now — the same assignment is `TS2322`.

**Tests** — each verified against the mutation it exists to catch

- Every case in the table goes through `serializeFlow` → `parseFlow` and is compared on its anchor. That round-trip stability is what `scroll-to.within`, `snapshot.cropOn` and `snapshot.maxMismatch` newly depend on, and nothing asserted it; deleting `yamlSelectorLabel`'s key sort now reddens the suite. The `when` case with `steps: []` became a two-step block — `parseFlow` refuses an empty guarded block, so no count below 1 is renderable, and `when #1` already pins the singular.
- The case table's value type is a non-empty tuple of cases **narrowed to the key's own kind**. `"wait": []` and a `wait` slot holding an `echo` case both typechecked clean before; both are `TS2322` now.
- The `when:` pluralization coupling only ran a two-step block, so hardcoding either surface's tail to `(N steps)` left everything green. Both counts run now, and against the whole skip reason rather than its suffix.
- New pins for the escaping above, on both surfaces, each confirmed to fail without the line it pins.

**Prose corrected against the code**

- `stepAnchor`: there are **two** anchor comparisons, not three, and only the per-verdict one is unconditionally raw-vs-parsed — host mode re-parses both sides of the whole-file check, which is raw-vs-parsed only in client persist mode.
- `delayMs` bounds: `.int()` admits nothing strictly between 0 and 1, so the 1 ms floor is hand-edit-only; the ceiling is recorder-reachable, as stated.
- `0` is not a parse-only `delayMs`: `min(0)` accepts it and `serializeFlow` writes it back out.

**Rejected, verified**

- Escaping `snapshot.name` (CodeRabbit) would be dead code: `parseFlow` gates it on `FLOW_NAME_PATTERN` — letters, digits, `_`, `-` — because it becomes a baseline filename, and that parse is the **only** construction site of a snapshot step. A comment there now says so.
- "`textMatches`' `/…/` delimiters are unpinned" no longer holds at this head: `flow-progress.test.ts`'s swipe-`from` target asserts `to (0.1, 0.5) from /^Card \d+$/`, and dropping the delimiters fails it. Measured by mutation.

**Verification**: 371 files / 4950 tests in `packages/tool-server`; `argent-cli` 527, `argent-mcp` 87, `registry` 120. `npm run build`, `typecheck:tests --workspaces`, `eslint --max-warnings 0`, `prettier --check .`, `npm run knip`: clean. 13 mutations aimed at the new code, all killed.

**Docs**: no update needed. `packages/docs/` documents flow YAML, the CLI and the tools; it does not specify how a step summary or a `StepReport.target` is spelled, so nothing there goes stale. That covers the two earlier rendering changes from `3e2244b3` as well — report-target text literals becoming JSON-quoted, and the `type` target carrying the typed text (now bounded, above).

## Review follow-up (9f8a1e29)

The two threads left open after approval.

- `typedTextLabel` compared and cut on `text.length`, which counts UTF-16 units. 199 ASCII chars plus one astral char is 200 characters but 201 units, so the target quoted a lone high surrogate and announced `(+1 chars)` for text that fits under the cap whole. It counts and slices code points; the boundary is pinned on both sides of the cap and each assertion is red without the fix. Two sibling truncations are deliberately left, both pre-existing and outside the step-summary code: `badEntry` slices an already JSON-stringified entry, and `cappedReason`'s cap bounds the EMITTED length, where counting code points would let the emitted string overshoot its own bound.
- The `textMatches` delimiter thread, re-measured at this head: dropping the `/…/` fails `flow-progress.test.ts:231` and `flow-step-definitions.test.ts:451` - 2 failed | 1571 passed of 1573. Restoring them returns the suite to green.

**Verification**: `packages/tool-server` 371 files / 4950 tests; `eslint --max-warnings 0`, `prettier --check .`, `tsc --noEmit` on both `tsconfig.json` and `tsconfig.test.json`: clean.

## Summary by CodeRabbit

* **Improvements**
  * Flow step summaries and run reports now use consistent formatting across supported step types.
  * Type steps display their target identifier and entered text more clearly.
  * Selector, gesture, condition, delay, and snapshot details are rendered consistently, including safer handling of special characters.
  * Tool steps with cyclic arguments are handled safely in summaries.

* **Tests**
  * Added comprehensive coverage for step rendering, flow numbering, conditions, edge cases, and report output.


latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat)\]: Review pass over the refactor (8 parallel moves: claims-vs-code, nearest twin, non-happy paths, inputs, reachability both ways, what-outlives-the-call, absence pass, plus an independent old-vs-new equivalence rebuild). No HIGH/MEDIUM findings — the byte-for-byte claim held on all 1141 reachable comparisons, and the suite caught all 6 mutations thrown at it. Four LOWs found and fixed in 9cf51322:

- **delayLabel out-of-range**: a hand-edited `delayMs: 3000000000` rendered `(after 3000000000ms)` while `setTimeout` clamps >2³¹−1 to a 1ms tick. The label now describes no delay above the ceiling, matching the runner.
- **scroll-to `within`**: parsed (`within?` admitted) and serialized, container-scoped scroll changes what replays, yet neither surface rendered it. Summary now reads `… (down) within {"id":"list"}`; target `… within (id=list)`.
- **snapshot summary**: dropped `cropOn`/`maxMismatch`, both of which change what replays (baseline key gains `-crop-<hash>`; tolerance). Summary now carries them, as the report target already did for `cropOn`.
- **type summary quoting**: `"${step.text}"` broke the one-line-per-step contract on embedded quotes/newlines, against the module’s own JSON-quoting convention. Now `JSON.stringify(step.text)`.

Dropped without verification per severity policy: comment-only nitpicks and pre-existing renderings this PR preserves verbatim. One pre-existing issue spotted out of scope for a follow-up: `flow-visual.ts` `cropIdentity` hashes only flat selector fields, so two `cropOn` selectors differing only by scope collide on the baseline key.

latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Two review rounds over this branch. **No HIGH and no MEDIUM in either.** Round 1 was 7 parallel lenses (claims-vs-code, nearest twin, non-happy paths, inputs, reachability both ways, what-outlives-the-call, absence) plus a 32-mutation pass; round 2 was 4 lenses re-run against the fixes. Everything found was LOW; all of it is fixed here rather than left as comments.

**What held up.** The mutation pass re-derived the rendering-neutrality claim independently — main's two switches inlined verbatim and diffed against the new definitions over a 136-step corpus (every kind, 10 selector shapes, `delayMs` of `0.5 / 1 / 2**31-1 / 2**31 / 3e9 / Infinity / NaN / "2000" / -5`, text with quotes and newlines, `when` on both arms). Exactly 12 differences, every one inside a declared correction. The compile gate is real: adding `| { kind: "shake"; … }` to `FlowStep` fails `typecheck:tests` at the definitions record. 25 of 32 mutations were killed; of the 7 survivors, 4 were provably equivalent or unreachable (the `times`/`held` swap is guarded by mutually exclusive kind discriminants, so one clause is always `""`).

**Fixed in 7c804907** — one rendering defect this PR introduced, and the boundaries the case table left open:

- `scroll-to`'s report target spelled the step's scroll container ` within (…)`, which is exactly how `selectorLabel` spells a selector's *own* relational scope. `scroll-to: { target: Footer, within: { id: list } }` (scroll that container) and `scroll-to: { target: { text: Footer, within: { id: list } } }` (match only inside it) both rendered `"Footer" within (id=list)`; a step carrying both rendered `"Delete" within (id=cards) (up) within (id=settings)`, direction wedged between two identical-looking scopes. Reproduced through the real parser, then fixed to ` in scroll container (…)` — the noun `flow-actions` already uses in its own `scroll container … is not visible` failure. `main` avoided the collision only by dropping `within` entirely, so this arrived with correction #2.
- Pinning cases for the boundaries, each verified to fail without the code it pins: `setTimeout`'s ceiling from both sides (`2**31-1` keeps its label, `2**31` loses it) and its 1ms floor; `maxMismatch: 0`, the one value separating `!== undefined` from a truthiness test; `cropOn` + `maxMismatch` together, fixing their suffix order; a multi-field selector, fixing the report target's field order.
- `Number.isFinite(ms) &&` in `delayLabel` went dead the moment correction #1 added the ceiling — NaN and ±Infinity both fail the range test. Removed, proved a no-op by mutation.
- Prose that had stopped matching the code: the point-gesture comment argued a criterion the file does not follow (`snapshot` and `scroll-to` are equally not recorder-built, and this PR renders their fields anyway) and named two of the four still-unrendered replay-affecting fields; `flow-tools.test.ts` still claimed to be the only coverage of the long-press rendering; and the new case table justified a no-op `as number` cast with a reachability claim wrong in both halves — `flow-add-step`'s `delayMs` is `z.number().int().min(0)` with no ceiling, so the recorder reaches these values too.

**Fixed in 30594244** — round 2 caught the replacement comment being wrong in its turn (every test in that block calls `summarizeStep` directly; nothing there goes through the recording tools), and one overstatement carried in from `main`: `selectorLabel` and `describeSelector` share the *scope shape*, not field order. Confirmed by running both — a selector with a regex matcher renders `/Sa.*/ id=save-btn role=button` as a target and `id="save-btn" role="button" text=/Sa.*/` as a reason, because one order is fixed and the other follows parsed key order.

**Filed rather than folded in**, both pre-existing and unchanged by this PR: #808 (a newline in an `echo` message or `tool` name splits one summary line in two — the hazard correction #4 fixed for `type`) and #809 (`type.submit`, `await.timeout`, `idle.timeout`, `idle.stableFor` change what replays and render on neither surface; the field-by-field table across all 16 kinds confirms it is exactly those four).

Verification on 30594244: `npm run build`, `eslint --max-warnings 0`, `prettier --check .`, `tsc -p tsconfig.test.json`, and `packages/tool-server` 47 files / 1199 tests — all clean, and CI green.

latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Third full review pass (7 parallel moves: inputs, what-outlives-the-call, absence matrix, claims-vs-code, nearest-twin, non-happy-paths, reachability-both-ways). **No HIGH and no MEDIUM.** The exhaustiveness gate, both-surface pinning, session-cleanup ordering, cyclic-args guard equivalence, anchor stability, and consumer compatibility of the new `in scroll container (...)` suffix all verified against source; `typecheck:tests` and the touched suites pass in-worktree. Four LOW findings remain, each confirmed by direct read:

1. **Report-target selector labels interpolate unquoted** (`flow-step-definitions.ts:77-78`): a selector like `{ text: 'Say "hi"' }` parses fine but renders `visible "Say "hi""` — the closing delimiter is indistinguishable from content; an embedded newline breaks the one-line report target. Pre-existing, moved verbatim from old `flow-run.ts`; `describeSelector` (reasons) does quote its values, so the two surfaces are inconsistent here. Display-only.
2. **`type` report target omits the typed text** (`flow-step-definitions.ts:270`): two steps differing only by `text` into the same field render identical targets, while sibling await/assert targets carry their expectation. Pre-existing, but unlike every other one-surface omission it carries no justifying comment, and the new test enshrines it.
3. **Universal-selector branch unasserted** (`flow-step-definitions.ts:76`): the `*` push has no test asserting its rendered string on any surface — deleting it leaves the suite green while an `{ any: true, next: ... }` target silently loses the universal marker.
4. **Pluralization twin pair still two-homed** (`flow-run.ts:1902` vs `flow-step-definitions.ts:262`): byte-equality is enforced only by a comment; no test compares them, so a future edit to one desyncs the `when:` summary from the skip reason.

Also fixed directly: the PR body's line-count figures were stale (claimed −100/−177; actual −113/−180) — description corrected.

latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: All four findings from the third pass are now fixed and re-verified in 3e2244b3 (rebased onto 65f26502 after main moved):

1. **Report-target selector labels interpolate unquoted** — `selectorLabel` now JSON-quotes the `text` literal, matching the module's own literal convention (`yamlTextConditionLabel`). Reproduced before the fix (`{ text: 'Say "hi"' }` rendered `"Say "hi""`, closing delimiter indistinguishable from content); after the fix it renders `"Say \"hi\""`. Simple-text pins are byte-unchanged.
2. **`type` report target omits the typed text** — now `into id=email ← "a@b.c"`, the way await/assert targets carry their expectation. Two steps differing only by text no longer collapse to identical targets (was reproduced: both rendered `into id=email`). Pins updated, including a case that fails if the text drops off again.
3. **Universal-selector branch unasserted** — new case pins `* within (id=nav)` on both surfaces; deleting the `*` push now reddens the suite. (A scopeless `{ any: true }` case was dropped during verification: `selectorToYaml` rejects it, so the scoped shape is the reachable one.)
4. **Pluralization twin pair two-homed** — new coupling test runs a skipped `when:` block through the runner and asserts the skip reason ends with the exact `(N steps)` tail `summarizeStep` spells for the same block, so an edit desyncing either homed spelling now fails a test instead of drifting silently.

Verification on 3e2244b3: full packages/tool-server suite 363 files / 4637 tests green, `typecheck:tests` clean, eslint and prettier clean on touched files.

latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Fourth full review pass over 3e2244b3 (two parallel lenses: claims-vs-code / inputs / reachability / anchor-stability, and an absence matrix / prose-audit / test-table hand-evaluation / consumer-parsing sweep). **No HIGH and no MEDIUM — clean pass, stopping here.** Every behavioral claim in `flow-step-definitions.ts` re-verified against its cited site; all 40+ case-table expectations hand-evaluated correct; no consumer parses the rendered strings; the render-on-neither-surface set is still exactly the four declared fields.

Three LOWs surfaced, noted for a follow-up rather than folded in (each display-only, pre-existing in shape):
- a `matches:` pattern containing `/` or a newline renders ambiguously inside the `/…/` delimiters on both surfaces;
- an empty `echo` message renders as a bare `N. echo: ` line;
- the "as-written path" comment on `run:`'s summary is slightly loose — `run: login` summarizes as `login.yaml`, which is the documented one-way migration, not the literal spelling.

CI green on 3e2244b3.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
```diff
@@ -0,0 +1,376 @@
...
+    target: (step) => whenLabel(step.condition, (ui) => conditionLabel(ui, selectorLabel)),
+  },
+  "tap": POINT_GESTURE_STEP,
+  "long-press": POINT_GESTURE_STEP,
+  "type": {
+    summary: (step) => `${yamlSelectorLabel(step.into)} ← ${JSON.stringify(step.text)}`,
+    // The typed text rides along, the way an await/assert target carries its
+    // expectation: two steps into the same field with different text would
+    // otherwise render identical targets.
+    target: (step) => `into ${selectorLabel(step.into)} ← ${JSON.stringify(step.text)}`,
```
hubgan: A `type` step's report target now carries the typed text itself, with no length bound. `StepReport.target` is printed to stdout by `argent flow run`, emitted in `--json`/`--json-stream`, rendered into the MCP text block handed to the agent, and appended to `~/.argent/mcp-calls.log`, which has no gate, redaction, cap or rotation. Before this change the typed text reached those surfaces only through `flow_finish_recording`'s summary at authoring time; it now appears on every replay, including CI runs on another machine.

Measured on a real Chromium run with a 50,000-character `type` text: the target grows from 25 to 50,030 characters, the CLI report from 252 B to 50,259 B, the MCP text blocks from 266 B to 50,273 B, and the appended log line from 495 B to 50,503 B, each carrying the string verbatim.

A `{{secret:...}}` placeholder is unaffected: resolution happens inside the `keyboard` tool's `execute`, downstream of the step object this renders. Confirmed on a passing step whose field received the resolved value while the target showed only the placeholder. The exposure is for a credential written literally into the flow YAML.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
```diff
@@ -0,0 +1,376 @@
...
+ * How one flow step kind reads to a human — the recording summary and the run
+ * report's target — owned by the kind itself, the way each tool owns its log
+ * wording in `ToolDefinition.interaction`.
+ */
+interface FlowStepDefinition<S extends FlowStep> {
+  /**
+   * The step's summary line minus the `<n>. <key>: ` prefix
+   * {@link summarizeStep} adds, in the flow file's own spellings.
+   */
+  summary(step: S): string;
```
hubgan: `summary` and `target` are declared in method shorthand, so TypeScript checks their parameter types bivariantly even under `strict`. The exhaustiveness half of the guarantee holds - a kind added to `FlowStep` without an entry here is TS2741 - but the per-kind step type is not enforced: a definition annotated with a step type narrower than its slot (one that requires `times` on `tap`, say) is accepted here and then reads a field the step need not carry, with a fully green typecheck. Declaring these two members as property types rather than methods rejects the same assignment with TS2322, which is how I confirmed the bivariance is what admits it.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
```diff
@@ -0,0 +1,376 @@
...
+function selectorLabel(sel: FlowSelector): string {
+  const parts: string[] = [];
+  // The universal selector prints as CSS spells it, so a scope-only target
+  // never renders as an empty label.
+  if (sel.any) parts.push("*");
+  // Text literals use JSON quoting, the module's own convention for literals
+  // (see `yamlTextConditionLabel`): a raw `"…"` delimiter is indistinguishable
+  // from content when the text itself carries quotes, and an embedded newline
+  // breaks the one-line-per-target report shape.
+  if (sel.text !== undefined) parts.push(JSON.stringify(sel.text));
```
hubgan, line 86: Reverting this to the pre-PR `"${sel.text}"` form leaves all 1358 tests in `test/flows/` green. Every report-target case in the new table uses plain text, so nothing pins the escaping this line exists for. The `type` cases do exercise quoting, but on `step.text`, which does not pass through `selectorLabel`. A selector such as `tap: { text: "Delete \"all\"\nnow" }` parses today and is exactly the input this change is for.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
```diff
@@ -0,0 +1,376 @@
...
+  if (sel.textMatches !== undefined) parts.push(`/${sel.textMatches}/`);
```
hubgan: Dropping the slashes here also leaves all 1358 tests in `test/flows/` green - nothing in the suite renders a `textMatches` selector into a report target or a summary; `flow-selector-matches.test.ts` only round-trips it through parse and serialize. Without the delimiters a regex target reads the same as a literal-text one, which is the distinction the comment just below reasons about.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
```diff
@@ -0,0 +1,376 @@
...
+  if (sel.identifier) parts.push(`id=${sel.identifier}`);
+  if (sel.role) parts.push(`role=${sel.role}`);
```
hubgan: The comment three lines up gives "an embedded newline breaks the one-line-per-target report shape" as the reason for JSON-quoting `text`, but `textMatches`, `identifier` and `role` are still interpolated raw on these three lines, so the same newline in any of them splits one step across two console lines. The new ` in scroll container (...)` clause routes a further selector through these branches, onto every `scroll-to` report line rather than only into flow-actions' failure reason. Verified on a real run: a `within` selector containing a newline renders one step as two lines through `renderReport`, and one containing an ANSI escape repaints the terminal. The class is pre-existing and this PR reduces it for `text`; what is inconsistent is that the stated rationale does not extend to its own neighbours.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
```diff
@@ -0,0 +1,376 @@
...
+    // beyond the screen itself: a target would render as "idle screen idle".
+    target: () => undefined,
+  },
+  "wait": {
+    summary: (step) => `${step.ms}ms`,
+    target: () => undefined,
+  },
+  "scroll-to": {
+    summary: (step) =>
+      `${yamlSelectorLabel(step.target)} (${step.direction})` +
```
hubgan, line 379: `within` joins `cropOn` and `maxMismatch` in the summary, and the summary is also `stepAnchor`, which is compared across a serialize-then-parse round trip. Nothing in the new test file exercises that round trip - it contains no `serializeFlow`, `parseFlow` or `stepAnchor` - so the round-trip stability these three new fields now depend on rests on the doc comment rather than an assertion.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
```diff
@@ -0,0 +1,376 @@
...
+}
+
+/**
+ * WHICH step this is, told apart from where it sits.
+ *
+ * The same renderer as the summary, on a fixed number, so the identity does not
+ * move with the position.
+ *
+ * The anchor rests on {@link summarizeStep} being STABLE across a
+ * serialize-then-parse round trip: two of the three comparisons put a raw
```
hubgan: There are two `stepAnchor` comparisons in the repo, not three - both in `flow-finish-recording.ts`, at lines 125 and 131. Only the per-warning one at 131 is unconditionally raw-vs-parsed; the whole-file check at 125 compares parsed against parsed in host mode, where `session.flow` is itself `parseFlow` output, and is raw-vs-parsed only in client persist mode.

hubgan (commented):
Reviewed the diff and the code it transitively affects, over two full passes. No blocking findings - everything inline is non-blocking.

Verified and clear along the way: rendering is equivalent to `main` over 7,175 generated steps beyond the seven intended deltas; `describeWhenCondition` is byte-identical over 141,241 conditions; `stepAnchor` round-trips cleanly for the three newly-rendered fields; `delayLabel` matches what `setTimeout` actually does at every boundary; no import cycles; eslint, prettier, knip and tsc are green; no `packages/docs/` page is made stale; and a `{{secret:...}}` placeholder is never resolved into a report. Exercised end to end with real flow runs on iOS, Android and Chromium against a tool-server built from this branch.

One process note with no line to hang it on: `CLAUDE.md` asks for a docs update in the same PR or an explicit statement that none is needed, and the description has neither. It also does not cover the two rendering changes from `3e2244b3a` - report-target literals becoming JSON-quoted, and the `type` target carrying the typed text.

packages/tool-server/test/flows/flow-step-definitions.test.ts
```diff
@@ -0,0 +1,382 @@
...
+    {
+      // The ceiling itself: setTimeout accepts 2³¹−1, so this is a real
+      // ~24.8-day sleep the runner performs and the label must keep.
+      step: { kind: "tool", name: "screenshot", args: {}, delayMs: 2 ** 31 - 1 },
+      summary: "1. tool: screenshot {} (after 2147483647ms)",
+      target: undefined,
+    },
+    {
+      // One past it sleeps a clamped tick instead, so the summary describes no
+      // delay. Neither bound is hand-edit-only: `flow-add-step`'s `delayMs` is
```
hubgan: `flow-add-step`'s `delayMs` is `z.number().int().min(0)`. The `.int()` means the recorder cannot emit any value strictly between 0 and 1, so the 1 ms floor the next case covers is reachable only by hand-editing the YAML. The ceiling is recorder-reachable as described; "neither bound" is not.

packages/tool-server/test/flows/flow-step-definitions.test.ts
```diff
@@ -0,0 +1,382 @@
...
+  target: string | undefined;
+  step: FlowStep;
+}
+
+/**
+ * One case per step kind, keyed by kind so a kind added to {@link FlowStep}
+ * without wording here fails to compile — the same gate the definitions
+ * themselves carry, asserted against the strings both surfaces actually print.
+ */
+const CASES = {
```
hubgan, line 33: The compile gate proves a key exists, not that the kind is exercised. `"vibrate": []` satisfies `readonly Case[]`, so a new kind can clear both this table and the definitions record with zero assertions. `Case.step` is also typed `FlowStep` rather than narrowed to its key: replacing the whole `"wait"` entry with an `echo` case passes `typecheck:tests` and all 85 assertions, leaving `wait` uncovered.

packages/tool-server/test/flows/flow-tools.test.ts
```diff
@@ -2799,8 +2797,9 @@ describe("the flow-add-step schema the CLI tests hand-copy", () => {
 // summarizeStep is the single spelling shared by the recorder's per-step
 // `recorded` line and flow-finish-recording's `summary`. `times` (tap),
 // `duration` (long-press) and `delayMs` (tool) change what replays, so a
-// summary that drops them misdescribes the file. long-press steps have no
-// live recorder path, so this is the only coverage of that rendering.
+// summary that drops them misdescribes the file. flow-step-definitions.test.ts
+// pins one case per kind; these add the `delayMs` shapes only a real YAML parse
+// yields — `soon`, `.nan`, `"2000"`, `0`, `-5`.
```
hubgan: `0` is not parse-only: `z.number().int().min(0)` accepts it, `serializeFlow` writes it back out, and it is an ordinary in-memory literal. The other four in the list are genuinely parse- or hand-edit-only.

packages/tool-server/test/flows/flow-when.test.ts
```diff
@@ -1288,4 +1289,32 @@ describe("when: run cancellation", () => {
...
+    };
+    await writeFlow("plural", { executionPrerequisite: "", steps: [block] });
+    currentTree = () => screen([]);
+
+    const result = await run("plural");
+
+    const summaryTail = summarizeStep(block, 1).match(/\([^()]*\)$/)?.[0];
+    expect(summaryTail).toBe("(2 steps)");
+    // Plain suffix compare: the tail's parentheses are regex metacharacters.
+    expect(result.steps[0].reason?.endsWith(`block skipped ${summaryTail}`)).toBe(true);
```
hubgan: The stated purpose is to couple flow-run's skip-reason pluralization to the summary's, but the test only exercises a two-step block, where both spellings trivially produce "steps". Hardcoding `flow-run.ts`'s tail to `(${n} steps)`, dropping the `n === 1` branch, leaves the whole of `test/flows/` green - so the singular direction the coupling is meant to guard is unguarded. The final assertion is also a suffix check rather than an exact comparison.

hubgan (approved):

packages/tool-server/src/tools/flows/flow-step-definitions.ts
latekvo: Fixed in 4560f41265bdb0cb5b0fc0d98a65d7fe60c293d2. Confirmed the reach before acting: `renderStepLine` prints `${s.kind} ${s.tool ?? s.target}` verbatim, `stepLabel` in argent-mcp/src/content.ts does the same into the text block, and the log at `ARGENT_MCP_LOG ?? ~/.argent/mcp-calls.log` has no cap or rotation — so this rode every replay, unbounded.

The target now stops at 200 chars with the `…(+N chars)` elision `parseFlow` already uses on the entry it echoes (`MAX_ENTRY_RENDER_CHARS`), which carries the same rationale about `StepReport.reason`. It elides before quoting, so a cut cannot halve an escape sequence.

The summary keeps the text whole, deliberately: it is also `stepAnchor`, so an elision there would collide two steps differing only past the cap and silently drop the recording's verdicts on both. Both halves are pinned, and capping the summary too is one of the mutations the new test kills.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
latekvo: Fixed in 4560f41265bdb0cb5b0fc0d98a65d7fe60c293d2. Reproduced the bivariance standalone first — a definition annotated `(step: Tap & { times: number })` assigned into the `tap` slot compiles under method shorthand and is `TS2322` under property types, with the shared `POINT_GESTURE_STEP`/`UI_CONDITION_STEP` assignments (wider param, contravariant) and `definitionOf`'s cast unaffected either way.

Both members are property types now. Re-confirmed in the real file: annotating `wait`'s `summary` narrower than its slot is `TS2322` at the record, where it typechecked clean before.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
latekvo: Fixed in 4560f41265bdb0cb5b0fc0d98a65d7fe60c293d2. Confirmed by mutation before acting — reverting the line to `"${sel.text}"` left all 55 files / 1570 tests in `test/flows/` green.

Pinned now on both surfaces, with a case whose text carries the two things the line exists for at once (`Delete "all"\nnow`); the revert reddens it. The same test covers the neighbours — see the thread on line 86.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
latekvo: This one no longer holds at this head, and I do not think it is a difference of commit: dropping the delimiters reddens `flow-progress.test.ts`, whose swipe `from` target asserts `to (0.1, 0.5) from /^Card \\d+$/` — a `textMatches` selector through `selectorLabel`. Measured the same way as the others: apply the mutation, run `test/flows`, read the failure.

(Your run reports 1358 tests where mine reports 1570, so the trees differ somewhere.) Nothing changed for this, but the escaping fix on line 86 does add a `textMatches` case that pins the delimiters alongside the escaping.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
latekvo: Fixed in 4560f41265bdb0cb5b0fc0d98a65d7fe60c293d2 — the rationale now covers its neighbours instead of being narrowed to fit them.

New `escapeInline` in flow-utils spells out control characters, and `textMatches`, `identifier` and `role` all go through it in `selectorLabel`; `describeSelector` gets the same treatment, since its output lands on the same report line through `reason`, and its literal arm moved from a raw `"${v}"` to `JSON.stringify` (byte-identical for ordinary values).

**Control characters only, not JSON escaping** — that was the one real trap here. JSON-escaping the regex arm doubles a pattern's own backslashes: `flow-progress.test.ts` and `flow-selector-matches.test.ts` both went red rendering `/^Card \\\\d+$/`, a regex nobody could copy back into the .yaml. The hazard you named is the control characters, so that is exactly the class escaped; a backslash is content.

Pinned on both surfaces, including the case that separates the two rules: a `textMatches` of `^Row\n\\d+$` renders `/^Row\\n\\d+$/` — newline spelled out, backslash intact. Dropping `escapeInline` from any of the four call sites, or shrinking its character class, now fails a test.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
latekvo: Fixed in 4560f41265bdb0cb5b0fc0d98a65d7fe60c293d2. Correct that nothing exercised it — the file had no `serializeFlow`, `parseFlow` or `stepAnchor` at all.

It does now, over the whole case table rather than the three fields: each case is serialized, parsed back and compared on its anchor. Deleting `yamlSelectorLabel`'s key sort — the round-trip property the anchor doc reasons about — fails it, so it is not vacuous.

One case had to change to get there: the `when` case with `steps: []` cannot round-trip, because `parseFlow` rejects an empty guarded block (`when needs a non-empty steps list to guard`). It is a two-step block now, which pins the plural arm; `when #1` already pinned the singular.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
latekvo: Fixed in 4560f41265bdb0cb5b0fc0d98a65d7fe60c293d2. Traced both sites: in host mode `appendStepToFlow` assigns `session.flow = parseFlow(flowFile)` (flow-utils.ts:3801) and the finish parses its own read, so line 125 is parsed-vs-parsed there and raw-vs-parsed only in client persist mode, where `session.flow.steps.push(step)` keeps the in-memory object. Line 131 compares against `verdict.step`, taken at flow-add-step.ts:1318 off the raw step, so that one always is.

The comment now says two comparisons, names which one is unconditional, and gives host mode as the reason the other is not.

packages/tool-server/test/flows/flow-step-definitions.test.ts
latekvo: Fixed in 4560f41265bdb0cb5b0fc0d98a65d7fe60c293d2. `.int()` does rule out the floor: the comment now says the ceiling is recorder-reachable and the 1 ms floor is hand-edit-only, on the case each belongs to.

packages/tool-server/test/flows/flow-step-definitions.test.ts
latekvo: Fixed in 4560f41265bdb0cb5b0fc0d98a65d7fe60c293d2. Confirmed both halves against `typecheck:tests` first — `"wait": []` and a `wait` slot holding `{ kind: "echo", … }` each compiled clean, exactly as you describe.

The table's value type is now `readonly [Case<Extract<FlowStep, { kind: K }>>, ...Case<…>[]]`: a non-empty tuple of cases narrowed to the key's own kind. Both of those are `TS2322` now — the empty list as "Source has 0 element(s) but target requires 1", the wrong-kind case as `Type '"echo"' is not assignable to type '"wait"'`.

packages/tool-server/test/flows/flow-tools.test.ts
latekvo: Fixed in 4560f41265bdb0cb5b0fc0d98a65d7fe60c293d2. You are right — `z.number().int().min(0)` takes `0`, and `serializeFlow` writes it back out (`if (step.delayMs !== undefined)`), so it is neither parse-only nor hand-edit-only.

The comment now lists the four that are, says why for each (`soon`/`.nan`/`"2000"` are not numbers, `-5` is below `min(0)`), and keeps `0` with its own reason: it is the falsy number the runner's truthiness gate turns on.

packages/tool-server/test/flows/flow-when.test.ts
latekvo: Fixed in 4560f41265bdb0cb5b0fc0d98a65d7fe60c293d2. Confirmed by mutation that neither direction was pinned: hardcoding `flow-run.ts`'s tail to `(${n} steps)` — and, separately, the `when` summary's — each left all 55 files in `test/flows/` green.

The test now runs a one-step and a two-step block, and compares the whole reason rather than its suffix. Both hardcodes redden it.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
```diff
@@ -0,0 +1,426 @@
...
+    },
+  },
+  "snapshot": {
+    summary: (step) =>
+      step.name +
+      (step.cropOn ? ` cropOn ${yamlSelectorLabel(step.cropOn)}` : "") +
+      (step.maxMismatch !== undefined ? ` maxMismatch ${step.maxMismatch}` : ""),
+    target: (step) =>
+      step.cropOn ? `"${step.name}" cropOn ${selectorLabel(step.cropOn)}` : `"${step.name}"`,
+  },
```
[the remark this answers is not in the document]
latekvo: Verified and not applied — it would be dead code.

`snapshot.name` becomes a baseline filename, so `parseFlow` gates it on `FLOW_NAME_PATTERN` (`^[A-Za-z0-9_-]+$`) at flow-utils.ts:3083, rejecting the entry outright. And that parse is the *only* construction site of a `{ kind: "snapshot" }` step in the repo (flow-utils.ts:3089) — both renderers run over parsed steps, so no quote, backslash or newline can reach either one.

`JSON.stringify` there would also change the summary's spelling from `snapshot: home` to `snapshot: "home"` — that summary is the flow file's own spelling and is also `stepAnchor` — while leaving the report target byte-identical, since the manual quotes already produce `"home"`.

A comment at the definition now records the constraint, so the next reader can see why the raw interpolation is safe here: 4560f41265bdb0cb5b0fc0d98a65d7fe60c293d2.

The neighbouring class you are pointing at is real, just not on this field — a newline in an `echo` message or a `tool` name does break the one-line summary, and that is tracked as #808. A control-character escape for selector fields landed in the same commit.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
latekvo: Re-ran the mutation at `9f8a1e2972a8b64f77994f1e38c47bcee96a6a17` and it is caught. Applied exactly the drop - `parts.push(escapeInline(sel.textMatches))` in place of the `/…/` interpolation - then `npx vitest run test/flows` in `packages/tool-server`:

```
Test Files  2 failed | 53 passed (55)
     Tests  2 failed | 1571 passed (1573)
```

- `flow-progress.test.ts:231` - `labels swipe reports with their travel and optional starting target`: `"to (0.1, 0.5) from /^Card \d+$/"` came back as `"to (0.1, 0.5) from ^Card \d+$"`.
- `flow-step-definitions.test.ts:451` - `escapes every selector field value into a one-line report target`: expected `"/^Row\n\d+$/"`, received `"^Row\n\d+$"`.

Restoring the delimiters puts the suite back at 55 files / 1573 tests green, which is also the baseline I measured before mutating. Your 1358 is a different tree - the second of those two tests came in with the escaping fix on line 86, so a checkout from before it only has the `flow-progress` one.

packages/tool-server/src/tools/flows/flow-step-definitions.ts
```diff
@@ -0,0 +1,464 @@
...
+ * recording's verdicts on both.
+ */
+const MAX_TARGET_TEXT_CHARS = 200;
+
+function typedTextLabel(text: string): string {
+  if (text.length <= MAX_TARGET_TEXT_CHARS) return JSON.stringify(text);
+  // Elide before quoting, not after: a cut through the rendered form can halve
+  // an escape sequence.
+  const elided = text.length - MAX_TARGET_TEXT_CHARS;
+  return `${JSON.stringify(text.slice(0, MAX_TARGET_TEXT_CHARS))}…(+${elided} chars)`;
```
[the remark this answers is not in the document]
latekvo: Fixed in `9f8a1e2972a8b64f77994f1e38c47bcee96a6a17`.

Reproduced first: `"x".repeat(199) + "😀"` is 200 code points but 201 UTF-16 units, so the old cut rendered `…xxx\ud83d"…(+1 chars)` - a lone high surrogate for text that fits under the cap whole. `typedTextLabel` now counts and slices `Array.from(text)`.

The boundary case is pinned in `flow-step-definitions.test.ts` on both sides of the cap: the 200-code-point string is quoted entire, and one astral char past it keeps the emoji intact and reports `(+1 chars)`. Reverting only the source change reddens each assertion on its own (`\ud83d"…(+1 chars)` and `\ud83d"…(+3 chars)` respectively). `test/flows` is 55 files / 1573 tests green, the whole `packages/tool-server` suite 371 files / 4950 tests.

Two sibling truncations are deliberately left: `badEntry` (`flow-utils.ts`) slices an already JSON-stringified entry, and `cappedReason` (`flow-add-step.ts`) budgets a head and tail against a cap that bounds the EMITTED length, where switching to code points would let the emitted string exceed its own bound. Both pre-date this branch and neither is step-summary code.

PR #748, opened 2026-08-07
refactor(flow): generalize when's block machinery so a second block directive can't miss a site

j-piasecki:
`when` is the only directive with a sibling `steps:` list, and four runner sites asked `kind === "when"` directly to decide whether a step has authored children to expand as skip lines. A second block directive would have had to remember all four, and a forgotten one silently drops a whole block from the report — exactly the invariant `reportBlockSkipped` exists to hold.

This introduces `blockSteps(step)` as the single runtime predicate and `BLOCK_DIRECTIVE_KEYS` as its parse-time counterpart, and extracts the parse-side block plumbing (`assertBlockDepth`, `parseBlockSteps`) that a second directive must share — notably the depth cap, renamed `MAX_BLOCK_DEPTH`, which has to be ONE counter across directives or an alternating nesting chain evades it and reopens the cyclic-YAML-alias hole it was added to close.

No behavior change, with one deliberate exception: the depth-cap message now reads `block directives nest deeper than 20 levels` where it said `when blocks nest deeper than 20 levels` — the cap stopped being `when`'s the moment it became shared, and a message naming only `when` would misdirect an author who hit it through a different directive. Every other error message, their ordering, and the report shapes are untouched.

559 flow tests pass unchanged.

---

**Stack**

1. **#748 — you are here** · `refactor(flow)`: generalize when's block machinery → `main`
2. #749 · `feat(flow)`: add `repeat:` blocks → #748

### DIFF
--- a/packages/argent-cli/src/flow.ts
+++ b/packages/argent-cli/src/flow.ts
@@ -40,7 +40,7 @@ export interface StepReport {
   /** Human-readable step target (selector / snapshot name), set by the runner. */
   target?: string;
   /**
-   * Nesting depth: absent/0 at top level, +1 inside each block directive
+   * Nesting depth: absent/0 at top level, +1 inside each nesting step
    * (`when:` guarded steps, `run:` fragment steps). Renderers indent by it; a
    * pre-depth tool-server sends none and the report renders flat, as before.
    */
--- a/packages/tool-server/src/tools/flows/flow-device.ts
+++ b/packages/tool-server/src/tools/flows/flow-device.ts
@@ -205,7 +205,7 @@ export function stripDeviceKeys(args: Record<string, unknown>): Record<string, u
  *
  * - `when` needs a device whatever its body contains, because the guard itself
  *   reads one — the device's platform, or its view tree. That classifies the
- *   header alone; a block's body is answered by {@link flowRequiresDevice}'s
+ *   header alone; a block's body is left to {@link flowRequiresDevice}'s
  *   walk, not here.
  * - `idle` needs one despite carrying no selector: it reads the device twice
  *   over, the UI tree and a screenshot of it.
@@ -244,10 +244,11 @@ export function stepRequiresDevice(registry: Registry, step: FlowStep): boolean 
 }
 
 /**
- * Whether any step in a flow acts on a device, block children included — each
- * block header's own classification OR, via {@link blockSteps}, the steps it
- * actually CONTAINS. Both halves matter: a `when: { platform: ios }` guard
- * reads the device even over a device-free body.
+ * Whether any step in a flow acts on a device - each block header's own
+ * classification OR, via {@link blockSteps}, the steps it actually CONTAINS.
+ * Today the header half answers every case: `when`, the only block kind,
+ * classifies device-requiring in {@link stepRequiresDevice}, so the child walk
+ * is a no-op until a block kind classifies `false`.
  *
  * Header classification alone is not enough: a block directive whose header
  * reads nothing off the device would naturally be classified `false` in
--- a/packages/tool-server/src/tools/flows/flow-device.ts
+++ b/packages/tool-server/src/tools/flows/flow-device.ts
@@ -2,7 +2,7 @@ import type { DeviceInfo, Registry, ToolContext } from "@argent/registry";
 import { FAILURE_CODES, FailureError } from "@argent/registry";
 import { resolveDevice } from "../../utils/device-info";
 import { invokeSubTool } from "../../utils/sub-invoke";
-import type { FlowStep, WhenPlatform } from "./flow-utils";
+import { blockSteps, type FlowStep, type WhenPlatform } from "./flow-utils";
 
 /**
  * Device resolution + binding for the flow runner. Flows store no device id
@@ -204,7 +204,9 @@ export function stripDeviceKeys(args: Record<string, unknown>): Record<string, u
  * Three of the classifications are worth stating outright:
  *
  * - `when` needs a device whatever its body contains, because the guard itself
- *   reads one — the device's platform, or its view tree.
+ *   reads one — the device's platform, or its view tree. That classifies the
+ *   header alone; a block's body is answered by {@link flowRequiresDevice}'s
+ *   walk, not here.
  * - `idle` needs one despite carrying no selector: it reads the device twice
  *   over, the UI tree and a screenshot of it.
  * - `run` needs one without the fragment being read here. The flow it names is
@@ -241,9 +243,24 @@ export function stepRequiresDevice(registry: Registry, step: FlowStep): boolean 
   }
 }
 
-/** Whether any step in a flow acts on a device. */
+/**
+ * Whether any step in a flow acts on a device, block children included — each
+ * block header's own classification OR, via {@link blockSteps}, the steps it
+ * actually CONTAINS. Both halves matter: a `when: { platform: ios }` guard
+ * reads the device even over a device-free body.
+ *
+ * Header classification alone is not enough: a block directive whose header
+ * reads nothing off the device would naturally be classified `false` in
+ * {@link stepRequiresDevice}, and a flow that is only that block would then
+ * resolve device-free and hard-stop on the first device step in its body. The
+ * exhaustiveness check there forces a new kind to be classified, but not to be
+ * classified `true`; the walk is what makes either answer safe.
+ */
 export function flowRequiresDevice(registry: Registry, steps: FlowStep[]): boolean {
-  return steps.some((step) => stepRequiresDevice(registry, step));
+  return steps.some(
+    (step) =>
+      stepRequiresDevice(registry, step) || flowRequiresDevice(registry, blockSteps(step) ?? [])
+  );
 }
 
 /**
--- a/packages/tool-server/src/tools/flows/flow-device.ts
+++ b/packages/tool-server/src/tools/flows/flow-device.ts
@@ -255,7 +255,8 @@ export function stepRequiresDevice(registry: Registry, step: FlowStep): boolean 
  * {@link stepRequiresDevice}, and a flow that is only that block would then
  * resolve device-free and hard-stop on the first device step in its body. The
  * exhaustiveness check there forces a new kind to be classified, but not to be
- * classified `true`; the walk is what makes either answer safe.
+ * classified `true`; the walk - here and in {@link flowScopesDevice}, the
+ * question the runner asks next - is what makes either answer safe.
  */
 export function flowRequiresDevice(registry: Registry, steps: FlowStep[]): boolean {
   return steps.some(
@@ -273,10 +274,21 @@ export function flowRequiresDevice(registry: Registry, steps: FlowStep[]): boole
  * (keeping the cross-agent protection the scope exists for), or, where no
  * single device is resolvable, run the step's unscoped meaning rather than
  * failing a flow whose whole purpose is to clear the machine.
+ *
+ * Walks a block's body via {@link blockSteps}. The walk answers nothing in a
+ * run today: `when`, the only block kind, classifies device-requiring, so a
+ * flow holding a block is answered by {@link flowRequiresDevice} and never
+ * reaches this question at all. The pair needs both walks - under
+ * a block kind that reads nothing off the device, a `devices` scope in its body
+ * would be invisible here, the run would resolve no device, and the teardown
+ * would execute with its unscoped meaning, the machine-wide sweep the scope
+ * exists to prevent.
  */
 export function flowScopesDevice(registry: Registry, steps: FlowStep[]): boolean {
   return steps.some(
-    (step) => step.kind === "tool" && declaresAny(registry, step.name, DEVICE_BIND_LIST_KEYS)
+    (step) =>
+      (step.kind === "tool" && declaresAny(registry, step.name, DEVICE_BIND_LIST_KEYS)) ||
+      flowScopesDevice(registry, blockSteps(step) ?? [])
   );
 }
 
--- a/packages/tool-server/src/tools/flows/flow-run.ts
+++ b/packages/tool-server/src/tools/flows/flow-run.ts
@@ -1804,11 +1804,11 @@ function describeWhenCondition(cond: WhenCondition): string {
 }
 
 /**
- * Report every step of a `when:` block that will not run as skipped — so a
- * run where the block was skipped (unmet guard, errored guard, hard stop, or
- * cancellation) produces the same report shape (one line per authored step,
- * at the same depth) as a run where it entered, and reports stay comparable
- * run-to-run. Nested block directives expand (their literal steps are known);
+ * Report every step of a block directive that will not run as skipped — so a
+ * run where the block was skipped (a `when:` guard unmet or errored, a hard
+ * stop, a cancellation) produces the same report shape (one line per authored
+ * step, at the same depth) as a run where it entered, and reports stay
+ * comparable run-to-run. Nested blocks expand (their literal steps are known);
  * a `run:` composition stays one line, matching how post-hard-stop skips report
  * a fragment that was never loaded. `scope` is the scope the steps would have
  * executed in — already the block's child scope, not the marker's.
--- a/packages/tool-server/src/tools/flows/flow-run.ts
+++ b/packages/tool-server/src/tools/flows/flow-run.ts
@@ -1710,7 +1710,7 @@ function scopeFlowDir(scope: StepScope): string {
   return path.dirname(scope.runStack[scope.runStack.length - 1]!.canonical);
 }
 
-/** The scope a block directive's children execute in — one level deeper. */
+/** The scope a nesting step's children execute in — one level deeper. */
 function childScope(
   scope: StepScope,
   overrides: Partial<Omit<StepScope, "depth">> = {}
@@ -1719,8 +1719,8 @@ function childScope(
 }
 
 /**
- * The depth stamp for a report — omitted at top level, so a flow with no block
- * directives produces a report byte-identical to the pre-depth shape.
+ * The depth stamp for a report — omitted at top level, so a flow with no
+ * nesting steps produces a report byte-identical to the pre-depth shape.
  */
 function depthOf(scope: StepScope): Pick<StepReport, "depth"> {
   return scope.depth ? { depth: scope.depth } : {};
--- a/packages/tool-server/src/tools/flows/flow-run.ts
+++ b/packages/tool-server/src/tools/flows/flow-run.ts
@@ -23,6 +23,7 @@ import {
   appIdForPlatform,
   assertSafeFlowName,
   assertValidProjectRoot,
+  blockSteps,
   chromiumLaunchSpec,
   classifyOnDiskSpelling,
   describeSelector,
@@ -1733,9 +1734,10 @@ async function execSteps(state: ExecState, steps: FlowStep[], scope: StepScope):
         // line rather than vanishing — matching reportBlockSkipped.
         ...(step.kind === "echo" ? { message: step.message } : {}),
       });
-      // A when block's literal steps are known — expand them so the report
+      // A block directive's literal steps are known — expand them so the report
       // keeps one line per authored step no matter where the stop landed.
-      if (step.kind === "when") reportBlockSkipped(state, step.steps, childScope(scope));
+      const inner = blockSteps(step);
+      if (inner) reportBlockSkipped(state, inner, childScope(scope));
       continue;
     }
     // The flow was resolved as needing no device, yet a step that acts on one
@@ -1752,7 +1754,8 @@ async function execSteps(state: ExecState, steps: FlowStep[], scope: StepScope):
         ...depthOf(scope),
         reason: `step needs a device but the flow was resolved as device-free — pass an explicit device`,
       });
-      if (step.kind === "when") reportBlockSkipped(state, step.steps, childScope(scope));
+      const inner = blockSteps(step);
+      if (inner) reportBlockSkipped(state, inner, childScope(scope));
       continue;
     }
     if (state.signal?.aborted) {
@@ -1767,9 +1770,8 @@ async function execSteps(state: ExecState, steps: FlowStep[], scope: StepScope):
         ...depthOf(scope),
         ...(step.kind === "echo" ? { message: step.message } : {}),
       });
-      if (step.kind === "when") {
-        reportBlockSkipped(state, step.steps, childScope(scope), "run aborted");
-      }
+      const inner = blockSteps(step);
+      if (inner) reportBlockSkipped(state, inner, childScope(scope), "run aborted");
       continue;
     }
 
@@ -1799,8 +1801,8 @@ function describeWhenCondition(cond: WhenCondition): string {
  * run where the block was skipped (unmet guard, errored guard, hard stop, or
  * cancellation) produces the same report shape (one line per authored step,
  * at the same depth) as a run where it entered, and reports stay comparable
- * run-to-run. Nested when blocks expand (their literal steps are known); a
- * `run:` composition stays one line, matching how post-hard-stop skips report
+ * run-to-run. Nested block directives expand (their literal steps are known);
+ * a `run:` composition stays one line, matching how post-hard-stop skips report
  * a fragment that was never loaded. `scope` is the scope the steps would have
  * executed in — already the block's child scope, not the marker's.
  */
@@ -1821,7 +1823,8 @@ function reportBlockSkipped(
       ...depthOf(scope),
       ...(step.kind === "echo" ? { message: step.message } : {}),
     });
-    if (step.kind === "when") reportBlockSkipped(state, step.steps, childScope(scope), reason);
+    const inner = blockSteps(step);
+    if (inner) reportBlockSkipped(state, inner, childScope(scope), reason);
   }
 }
 
--- a/packages/tool-server/src/tools/flows/flow-run.ts
+++ b/packages/tool-server/src/tools/flows/flow-run.ts
@@ -29,8 +29,10 @@ import {
   describeSelector,
   describeTextExpectation,
   getFlowPath,
+  isBlockStep,
   parseFlow,
   runTargetName,
+  type BlockStep,
   type FlowFile,
   type FlowSelector,
   type FlowStep,
@@ -1791,8 +1793,8 @@ async function execSteps(state: ExecState, steps: FlowStep[], scope: StepScope):
       await execRunStep(state, step, scope);
       continue;
     }
-    if (step.kind === "when") {
-      await execWhenStep(state, step, scope);
+    if (isBlockStep(step)) {
+      await execBlockStep(state, step, scope);
       continue;
     }
 
@@ -1837,6 +1839,26 @@ function reportBlockSkipped(
     });
     const inner = blockSteps(step);
     if (inner) reportBlockSkipped(state, inner, childScope(scope), reason);
+  }
+}
+
+/**
+ * Dispatch a block directive to its executor. The `never` default arm is the
+ * run-time site a kind registered in BLOCK_DIRECTIVE_KEYS cannot miss: an
+ * unhandled registered kind fails tsc here instead of falling through to
+ * execLeafStep as an "unsupported step kind" error that skips the block's
+ * children without a report. Binds `step.kind` rather than `step` - while the
+ * registry has one entry BlockStep is not a union, so only the discriminant
+ * narrows to `never`.
+ */
+async function execBlockStep(state: ExecState, step: BlockStep, scope: StepScope): Promise<void> {
+  switch (step.kind) {
+    case "when":
+      return execWhenStep(state, step, scope);
+    default: {
+      const unhandled: never = step.kind;
+      void unhandled;
+    }
   }
 }
 
--- a/packages/tool-server/src/tools/flows/flow-run.ts
+++ b/packages/tool-server/src/tools/flows/flow-run.ts
@@ -1750,6 +1750,11 @@ async function execSteps(state: ExecState, steps: FlowStep[], scope: StepScope):
     // The flow was resolved as needing no device, yet a step that acts on one
     // reached execution — the two decisions disagree. Report it as this step's
     // error and stop, rather than letting it fail obscurely further in.
+    // They cannot disagree today, nor for a future block directive whichever way
+    // stepRequiresDevice classifies it — flowRequiresDevice recurses through
+    // blockSteps, so no nesting hides a step: under `true` a device is resolved
+    // and `!state.device` is false; under `false` this guard's own
+    // stepRequiresDevice conjunct fails. The expansion below stays regardless.
     if (!state.device && stepRequiresDevice(state.registry, step)) {
       state.stopped = true;
       pushReport(state, {
--- a/packages/tool-server/src/tools/flows/flow-run.ts
+++ b/packages/tool-server/src/tools/flows/flow-run.ts
@@ -874,23 +874,30 @@ function displayFlowName(params: { name?: string; flow_path?: string }): string 
   return params.name || stem || params.flow_path || "(unspecified)";
 }
 
-/** Yield every parsed step, recursing into `when:` blocks (the parser's only nesting). */
+/**
+ * Yield every parsed step, recursing into a block directive's children through
+ * {@link blockSteps} rather than testing one kind: this is the sole feeder of
+ * {@link assertUploadSelfContained}, so a block absent from the recursion would
+ * carry an uploaded flow's nested `run:`/`snapshot` past the preflight and let
+ * an unrunnable flow report green.
+ */
 function* walkSteps(steps: FlowStep[]): Generator<FlowStep> {
   for (const step of steps) {
     yield step;
-    if (step.kind === "when") yield* walkSteps(step.steps);
+    const inner = blockSteps(step);
+    if (inner) yield* walkSteps(inner);
   }
 }
 
 /**
  * Reject an uploaded root flow that is not self-contained — one with a `run:`
- * or `snapshot` step (even inside a `when:` block) — before anything executes:
- * a mid-run or guard-gated error could execute half the flow first, or first
- * surface in CI. Both step kinds anchor at the flow file's real directory,
- * which an uploaded flow does not have: a run: step's referenced files stayed
- * on the client, and snapshot baselines live beside the flow's file — against
- * a per-call temp materialization a plain snapshot can only fail (no baseline)
- * and updateBaselines writes PNGs no later run can find.
+ * or `snapshot` step (at any depth inside a block directive) — before anything
+ * executes: a mid-run or guard-gated error could execute half the flow first,
+ * or first surface in CI. Both step kinds anchor at the flow file's real
+ * directory, which an uploaded flow does not have: a run: step's referenced
+ * files stayed on the client, and snapshot baselines live beside the flow's
+ * file — against a per-call temp materialization a plain snapshot can only fail
+ * (no baseline) and updateBaselines writes PNGs no later run can find.
  */
 function assertUploadSelfContained(flow: FlowFile): void {
   for (const step of walkSteps(flow.steps)) {
--- a/packages/tool-server/src/tools/flows/flow-run.ts
+++ b/packages/tool-server/src/tools/flows/flow-run.ts
@@ -240,8 +240,8 @@ export interface StepReport {
   /** Snapshot-step artifacts (baseline/current/diff) as materializable handles. */
   artifacts?: SnapshotArtifacts;
   /**
-   * Nesting depth for display: omitted at top level, +1 inside each block
-   * directive's expanded steps (a `when:` block's guarded steps, a `run:`
+   * Nesting depth for display: omitted at top level, +1 inside each nesting
+   * step's expanded steps (a `when:` block's guarded steps, a `run:`
    * fragment's steps). Renderers indent by it without knowing which directives
    * nest — the report is a flat list with no block-end marker, so depth cannot
    * be reconstructed downstream.
--- a/packages/tool-server/src/tools/flows/flow-run.ts
+++ b/packages/tool-server/src/tools/flows/flow-run.ts
@@ -1626,8 +1626,22 @@ function stepTarget(step: FlowStep): string | undefined {
       // The as-written path, so a report line shows exactly what the flow
       // references (`run ../shared/login.yaml`), not just the attribution stem.
       return step.flow;
-    default:
+    case "echo":
+    case "tool":
+      // Each carries its subject in a report field of its own (`message`,
+      // `tool`) that renderers print in the target's place.
       return undefined;
+    case "launch":
+      // A launch's app id may be per-platform (`appIdForPlatform`), and a step
+      // alone does not know the run device.
+      return undefined;
+    case "wait":
+      return undefined;
+    default: {
+      const unclassified: never = step;
+      void unclassified;
+      return undefined;
+    }
   }
 }
 
@@ -1845,11 +1859,12 @@ function reportBlockSkipped(
 /**
  * Dispatch a block directive to its executor. The `never` default arm is the
  * run-time site a kind registered in BLOCK_DIRECTIVE_KEYS cannot miss: an
- * unhandled registered kind fails tsc here instead of falling through to
- * execLeafStep as an "unsupported step kind" error that skips the block's
- * children without a report. Binds `step.kind` rather than `step` - while the
- * registry has one entry BlockStep is not a union, so only the discriminant
- * narrows to `never`.
+ * unhandled registered kind fails tsc here instead of returning silently and
+ * leaving the block out of the report entirely, not even its own marker.
+ * Preventing execLeafStep's "unsupported step kind" error is the isBlockStep
+ * gate's doing, not this arm's - a registered kind never reaches the leaf
+ * switch. Binds `step.kind` rather than `step` - while the registry has one
+ * entry BlockStep is not a union, so only the discriminant narrows to `never`.
  */
 async function execBlockStep(state: ExecState, step: BlockStep, scope: StepScope): Promise<void> {
   switch (step.kind) {
--- a/packages/tool-server/src/tools/flows/flow-utils.ts
+++ b/packages/tool-server/src/tools/flows/flow-utils.ts
@@ -729,7 +729,7 @@ export type FlowFile = {
 /**
  * The literal child steps of a block directive, or undefined for a leaf step.
  *
- * The single predicate for "this step has authored children", read at six
+ * The single predicate for "this step has authored children", read at seven
  * sites. Four expand a block that will NOT execute into skip lines, so a report
  * keeps one line per authored step no matter where the run ended: `execSteps`'
  * three gates — a hard stop, a device-free flow, a cancellation — plus
@@ -737,12 +737,12 @@ export type FlowFile = {
  * reaches those same skip lines by another route: `execWhenStep` has the block
  * in hand and passes `step.steps` straight to `reportBlockSkipped`.) The fifth
  * is the upload preflight's walk, where a block it cannot see hides a nested
- * `run:`/`snapshot` from validation. The sixth, `flowRequiresDevice`
- * (flow-device.ts), reads children not to skip them but to resolve the flow's
- * device requirement from a block's body — the guard against a later block
- * directive whose header reads nothing off the device, and a no-op while
- * `when`, whose header already classifies device-requiring, is the only block
- * kind.
+ * `run:`/`snapshot` from validation. The sixth and seventh,
+ * `flowRequiresDevice` and `flowScopesDevice` (flow-device.ts), read children
+ * not to skip them but to resolve the flow's device decisions from a block's
+ * body — the guard against a later block directive whose header reads nothing
+ * off the device, and dead in a run while `when`, whose header already
+ * classifies device-requiring, is the only block kind.
  *
  * Five sites asked `kind === "when"` directly before this existed — the five in
  * the runner; `flowRequiresDevice`'s walk arrived after the predicate and never
--- a/packages/tool-server/src/tools/flows/flow-utils.ts
+++ b/packages/tool-server/src/tools/flows/flow-utils.ts
@@ -744,13 +744,13 @@ export type FlowFile = {
  * off the device, and dead in a run while `when`, whose header already
  * classifies device-requiring, is the only block kind.
  *
- * Five sites asked `kind === "when"` directly before this existed — the five in
- * the runner; `flowRequiresDevice`'s walk arrived after the predicate and never
- * had to. A second block directive would have had to remember all five,
- * and a forgotten one drops a whole block from the report — or from the
- * preflight — silently. Now it is one case here — and it is the same case the
- * PARSER exempts from the single-key sibling check, because the kinds come from
- * {@link BLOCK_DIRECTIVE_KEYS} rather than being restated.
+ * Six sites asked `kind === "when"` directly before this existed: five for the
+ * children, one for the dispatch (now {@link isBlockStep}/`execBlockStep`);
+ * `flowRequiresDevice`'s walk arrived after the predicate and never had to. A
+ * second block directive would have had to remember all six; a forgotten one
+ * drops a whole block from the report or preflight silently. Now both are one
+ * case here, the same case the PARSER exempts from the single-key sibling
+ * check: the kinds come from {@link BLOCK_DIRECTIVE_KEYS}, not restated.
  */
 export function blockSteps(step: FlowStep): FlowStep[] | undefined {
   return isBlockStep(step) ? (step.steps satisfies FlowStep[]) : undefined;
--- a/packages/tool-server/src/tools/flows/flow-utils.ts
+++ b/packages/tool-server/src/tools/flows/flow-utils.ts
@@ -2064,10 +2064,11 @@ const _everyChildBearingKindIsRegistered: [UnregisteredBlockKind] extends [never
   : UnregisteredBlockKind = true;
 
 /**
- * Is this directive key a block directive? The parser and runner's only read of
- * {@link BLOCK_DIRECTIVE_KEYS}; the widening is the lookup itself — the const
- * tuple's own `includes` accepts only keys already known to be block kinds,
- * which is the question being asked.
+ * Is this directive key a block directive? The parser and runner's only
+ * CLASSIFYING read of {@link BLOCK_DIRECTIVE_KEYS} ({@link assertBlockDepth}
+ * reads it to NAME the keys in its message); the widening is the lookup
+ * itself — the const tuple's own `includes` accepts only keys already known to
+ * be block kinds, which is the question being asked.
  */
 function isBlockDirectiveKey(key: string): key is BlockDirectiveKind {
   return (BLOCK_DIRECTIVE_KEYS as readonly string[]).includes(key);
--- a/packages/tool-server/src/tools/flows/flow-utils.ts
+++ b/packages/tool-server/src/tools/flows/flow-utils.ts
@@ -1450,7 +1450,7 @@ const SELECTOR_KEYS: readonly string[] = [
 /**
  * Total number of scopes one selector may carry, counted across its whole
  * relation TREE rather than down a single branch — the selector analog of
- * MAX_WHEN_DEPTH. A size bound, not a depth bound, because each level can open
+ * MAX_BLOCK_DEPTH. A size bound, not a depth bound, because each level can open
  * three branches: capping depth alone still admits 3^depth scopes, and the
  * runner's loose-alternative expansion is exponential in the number of
  * bare-string scopes (`selectorAlternatives`), so a few hundred bytes of YAML
--- a/packages/tool-server/src/tools/flows/flow-utils.ts
+++ b/packages/tool-server/src/tools/flows/flow-utils.ts
@@ -1998,10 +1998,13 @@ const STEP_DIRECTIVE_KEYS: readonly string[] = [
  * The directive keys that carry a sibling `steps:` list — the single registry
  * of what a block directive is: {@link blockSteps} reads THIS list rather than
  * restating the kinds, so parse time and run time cannot answer differently.
- * Two constraints keep an entry honest: `satisfies` rejects a key that is not a
- * real step kind, and blockSteps' `Extract` rejects a kind whose step type
- * carries no `steps` (a directive listed here without children is a compile
- * error, not a silent runtime `undefined`).
+ * Three constraints keep an entry honest. Two judge the keys already listed:
+ * `satisfies` rejects a key that is not a real step kind, and blockSteps'
+ * `Extract` rejects a kind whose step type carries no `steps` (a directive
+ * listed here without children is a compile error, not a silent runtime
+ * `undefined`). Neither can force a key IN, which is what
+ * {@link _everyChildBearingKindIsRegistered} does — without it a child-bearing
+ * kind added later is simply absent here, and absent is not an error.
  *
  * At parse time these keys are exempt from the single-key sibling check,
  * because their own parser validates their siblings with pointed messages.
@@ -2010,6 +2013,24 @@ export const BLOCK_DIRECTIVE_KEYS = ["when"] as const satisfies readonly FlowSte
 
 /** The step kinds {@link BLOCK_DIRECTIVE_KEYS} names. */
 type BlockDirectiveKind = (typeof BLOCK_DIRECTIVE_KEYS)[number];
+
+/** The child-bearing step kinds {@link BLOCK_DIRECTIVE_KEYS} fails to list. */
+type UnregisteredBlockKind = Exclude<
+  Extract<FlowStep, { steps: FlowStep[] }>["kind"],
+  BlockDirectiveKind
+>;
+
+/**
+ * Forces a child-bearing kind INTO the registry — the direction the two
+ * constraints on {@link BLOCK_DIRECTIVE_KEYS} cannot cover, since both only
+ * judge kinds already listed. An unlisted one parses like any other step and
+ * then reads as `undefined` from {@link blockSteps}, so every reader listed
+ * there silently misses its children. Spelled as a conditional rather than a
+ * bare `never` so the compile error names the missing kind.
+ */
+const _everyChildBearingKindIsRegistered: [UnregisteredBlockKind] extends [never]
+  ? true
+  : UnregisteredBlockKind = true;
 
 /**
  * Is this directive key a block directive? The parser and runner's only read of
--- a/packages/tool-server/src/tools/flows/flow-utils.ts
+++ b/packages/tool-server/src/tools/flows/flow-utils.ts
@@ -737,10 +737,21 @@ export type FlowFile = {
  * nested `run:`/`snapshot` from validation. Five sites asked `kind === "when"`
  * directly before this existed; a second block directive would have had to
  * remember all five, and a forgotten one drops a whole block from the report —
- * or from the preflight — silently. Now it is one case here.
+ * or from the preflight — silently. Now it is one case here — and it is the
+ * same case the PARSER exempts from the single-key sibling check, because the
+ * kinds come from {@link BLOCK_DIRECTIVE_KEYS} rather than being restated.
  */
 export function blockSteps(step: FlowStep): FlowStep[] | undefined {
-  return step.kind === "when" ? step.steps : undefined;
+  return isBlockStep(step) ? step.steps : undefined;
+}
+
+/**
+ * Narrow a step to the kinds {@link BLOCK_DIRECTIVE_KEYS} lists. `Extract` is
+ * what makes the list load-bearing: {@link blockSteps}' `.steps` typechecks
+ * only while EVERY listed kind's step type carries children.
+ */
+function isBlockStep(step: FlowStep): step is Extract<FlowStep, { kind: BlockDirectiveKind }> {
+  return isBlockDirectiveKey(step.kind);
 }
 
 /**
@@ -1984,12 +1995,31 @@ const STEP_DIRECTIVE_KEYS: readonly string[] = [
 ];
 
 /**
- * The directive keys that carry a sibling `steps:` list. Parse-time counterpart
- * to {@link blockSteps} (which answers the same question about an already
- * parsed step): these are exempt from the single-key sibling check, because
- * their own parser validates their siblings with pointed messages.
+ * The directive keys that carry a sibling `steps:` list — the single registry
+ * of what a block directive is: {@link blockSteps} reads THIS list rather than
+ * restating the kinds, so parse time and run time cannot answer differently.
+ * Two constraints keep an entry honest: `satisfies` rejects a key that is not a
+ * real step kind, and blockSteps' `Extract` rejects a kind whose step type
+ * carries no `steps` (a directive listed here without children is a compile
+ * error, not a silent runtime `undefined`).
+ *
+ * At parse time these keys are exempt from the single-key sibling check,
+ * because their own parser validates their siblings with pointed messages.
  */
-const BLOCK_DIRECTIVE_KEYS: readonly string[] = ["when"];
+export const BLOCK_DIRECTIVE_KEYS = ["when"] as const satisfies readonly FlowStep["kind"][];
+
+/** The step kinds {@link BLOCK_DIRECTIVE_KEYS} names. */
+type BlockDirectiveKind = (typeof BLOCK_DIRECTIVE_KEYS)[number];
+
+/**
+ * Is this directive key a block directive? The parser and runner's only read of
+ * {@link BLOCK_DIRECTIVE_KEYS}; the widening is the lookup itself — the const
+ * tuple's own `includes` accepts only keys already known to be block kinds,
+ * which is the question being asked.
+ */
+function isBlockDirectiveKey(key: string): key is BlockDirectiveKind {
+  return (BLOCK_DIRECTIVE_KEYS as readonly string[]).includes(key);
+}
 
 /**
  * Parse `times` on a tap body: an integer tap count dispatched as ONE
@@ -2517,7 +2547,7 @@ function fromYamlStep(raw: YamlStep, blockDepth = 0): FlowStep {
   // it), but its own parser validates them with pointed messages, so the
   // generic check stays out of its way.
   const kind = kinds[0]!;
-  if (!BLOCK_DIRECTIVE_KEYS.includes(kind)) {
+  if (!isBlockDirectiveKey(kind)) {
     const siblings = kind === "tool" ? ["tool", "args", "delayMs"] : [kind];
     const extras = Object.keys(entry).filter((k) => !siblings.includes(k));
     if (extras.length > 0) {
--- a/packages/tool-server/src/tools/flows/flow-utils.ts
+++ b/packages/tool-server/src/tools/flows/flow-utils.ts
@@ -2310,7 +2310,9 @@ const MAX_BLOCK_DEPTH = 20;
  * Guard a block directive's recursion depth. Called FIRST in a block's parse —
  * before its own key/shape checks — so a cyclic alias reports as a depth error
  * rather than surfacing whatever the cycle happens to make the shallower checks
- * say.
+ * say. That early call buys the error PRECEDENCE only, not the cap itself:
+ * {@link parseBlockSteps} asserts again before it recurses, so forgetting the
+ * early call costs a directive the precedence and nothing more.
  */
 function assertBlockDepth(raw: unknown, depth: number): void {
   if (depth >= MAX_BLOCK_DEPTH) {
@@ -2327,12 +2329,18 @@ function assertBlockDepth(raw: unknown, depth: number): void {
  * chain. `emptyDetail` is the directive's own message for an absent or empty
  * list — the one part worth phrasing per directive, since it is where an author
  * lands when they wrote the guard but not the body.
+ *
+ * Asserts the depth cap here too, on the one path every block directive must go
+ * through to recurse, so no directive can opt out of the cap by forgetting the
+ * early {@link assertBlockDepth} call. `depth` is unchanged between the two
+ * calls, so for a directive that made the early one this is a no-op.
  */
 function parseBlockSteps(
   raw: Record<string, unknown>,
   depth: number,
   emptyDetail: string
 ): FlowStep[] {
+  assertBlockDepth(raw, depth);
   if (!Array.isArray(raw.steps) || raw.steps.length === 0) badEntry(raw, emptyDetail);
   return (raw.steps as unknown[]).map((s) => {
     if (s !== null && typeof s === "object") return fromYamlStep(s as YamlStep, depth + 1);
--- a/packages/tool-server/src/tools/flows/flow-utils.ts
+++ b/packages/tool-server/src/tools/flows/flow-utils.ts
@@ -2385,9 +2385,10 @@ const MAX_BLOCK_DEPTH = 20;
  */
 function assertBlockDepth(raw: unknown, depth: number): void {
   if (depth >= MAX_BLOCK_DEPTH) {
+    const directives = BLOCK_DIRECTIVE_KEYS.map((key) => `\`${key}:\``).join("/");
     badEntry(
       raw,
-      `block directives nest deeper than ${MAX_BLOCK_DEPTH} levels — check for a cyclic YAML alias (\`steps: &s … steps: *s\`)`
+      `${directives} blocks nest deeper than ${MAX_BLOCK_DEPTH} levels — check for a cyclic YAML alias (\`steps: &s … steps: *s\`)`
     );
   }
 }
--- a/packages/tool-server/src/tools/flows/flow-utils.ts
+++ b/packages/tool-server/src/tools/flows/flow-utils.ts
@@ -729,17 +729,28 @@ export type FlowFile = {
 /**
  * The literal child steps of a block directive, or undefined for a leaf step.
  *
- * The single predicate for "this step has authored children". The runner reads
- * it wherever a block does NOT execute — a hard stop, a device-free flow, a
- * cancellation, an unmet guard — to expand the block's steps as skip lines, so
- * a report keeps one line per authored step no matter where the run ended, and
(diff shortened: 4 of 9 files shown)


packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -395,6 +395,21 @@ export type FlowFile = {
...
+ *
+ * The single predicate for "this step has authored children". The runner reads
+ * it wherever a block does NOT execute — a hard stop, a device-free flow, a
+ * cancellation, an unmet guard — to expand the block's steps as skip lines, so
+ * a report keeps one line per authored step no matter where the run ended. Four
+ * sites asked `kind === "when"` directly before this existed; a second block
+ * directive would have had to remember all four, and a forgotten one drops a
+ * whole block from the report silently. Now it is one case here.
+ */
+export function blockSteps(step: FlowStep): FlowStep[] | undefined {
```
hubgan, line 755: The JSDoc above calls this "the single predicate" for whether a step has authored children, and says four sites asked `kind === "when"` directly. There is a fifth that was not converted: `walkSteps` at flow-run.ts:837 still tests `kind === "when"` and yields `step.steps`. Unlike `toYamlStep` and `stepRequiresDevice`, it is a bare `if` rather than an exhaustive switch, so tsc will not flag it when a second block directive lands - I confirmed that by adding a hypothetical block kind to the union and typechecking each of these shapes in isolation.

That site is not incidental. It is the sole feeder of `assertUploadSelfContained`, and flow-run.ts:2009 cites that preflight as the reason there is deliberately no path fence between `execRunStep` and the read ("The one route that carries untrusted content, an uploaded flow, never arrives here at all"). A directive registered in `blockSteps` and `BLOCK_DIRECTIVE_KEYS` but absent from `walkSteps` lets an uploaded flow carrying a nested `run:` or `snapshot` past the preflight and report green, which is the outcome flow-composition.test.ts:1432 and :1733 pin for `when`. On the stacked `repeat:` branch `walkSteps` is untouched, and `run:` inside `repeat:` is supported and has its own test.

Related: `walkSteps`'s docstring at flow-run.ts:833 still calls `when:` blocks "the parser's only nesting". That reads as settled fact right where the missed case lives, and it now contradicts the `MAX_BLOCK_DEPTH` JSDoc's "A block directive is the only kind of step whose parse recurses into child steps".

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -1426,6 +1441,14 @@ const STEP_DIRECTIVE_KEYS: readonly string[] = [
   "snapshot",
 ];
 
+/**
+ * The directive keys that carry a sibling `steps:` list. Parse-time counterpart
+ * to {@link blockSteps} (which answers the same question about an already
+ * parsed step): these are exempt from the single-key sibling check, because
+ * their own parser validates their siblings with pointed messages.
+ */
+const BLOCK_DIRECTIVE_KEYS: readonly string[] = ["when"];
```
hubgan: Nothing links this parse-time list to `blockSteps`'s runtime answer. They are two hand-maintained registries with no type or test forcing agreement, and the suite has no assertion that every kind listed here yields a defined array from `blockSteps` while no other kind does.

The four skip-site tests have the same shape problem: each hardcodes a `kind: "when"` literal instead of iterating a shared block-directive table, so a second directive would still be silently missing from all four report expansions - the exact failure mode this change is meant to remove.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -1719,13 +1742,50 @@ function parseWhenCondition(raw: unknown): WhenCondition {
...
+ * Nesting cap for block directives — the parse-side analog of flow-run's
+ * MAX_RUN_DEPTH. A block directive is the only kind of step whose parse
+ * recurses into child steps, and the yaml library happily materializes a cyclic
+ * alias (`steps: &s … steps: *s`) as a cyclic object; without a cap that cycle
  * escapes parseFlow as a raw RangeError instead of a structured parse error.
+ *
+ * ONE counter shared by every block directive, deliberately: a per-directive
+ * counter would let an alternating chain evade all of them and reopen the hole.
+ */
+const MAX_BLOCK_DEPTH = 20;
```
hubgan, line 2390: The rename leaves two references to the old name dangling: flow-utils.ts:1087, where the `MAX_SELECTOR_SCOPES` JSDoc calls itself "the selector analog of MAX_WHEN_DEPTH", and flow-composition.test.ts:1466, "The parser nests when: to MAX_WHEN_DEPTH (20)". Both name a symbol that no longer exists, so neither cross-reference survives a grep.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -1719,13 +1742,50 @@ function parseWhenCondition(raw: unknown): WhenCondition {
...
+ * Guard a block directive's recursion depth. Called FIRST in a block's parse —
+ * before its own key/shape checks — so a cyclic alias reports as a depth error
+ * rather than surfacing whatever the cycle happens to make the shallower checks
+ * say.
+ */
+function assertBlockDepth(raw: unknown, depth: number): void {
+  if (depth >= MAX_BLOCK_DEPTH) {
+    badEntry(
+      raw,
+      `block directives nest deeper than ${MAX_BLOCK_DEPTH} levels — check for a cyclic YAML alias (\`steps: &s … steps: *s\`)`
```
hubgan: This is the PR's one deliberate behavior change, and no test pins it. The only assertion that reaches this path, flow-when.test.ts:183, matches `/nest deeper than 20 levels/i` - precisely the substring that did not change - so the reworded half is unverified and free to regress to anything. flow-utils.test.ts:1202 looks like a second guard, but I ran its fixture: it throws the `MAX_SELECTOR_SCOPES` message, so its `nest deeper than` alternative never fires.

The cap's value is unpinned too. I confirmed against the real parser that 20 nested levels parse and 21 reject, but nothing asserts either side of that boundary, so the constant can drift in either direction unnoticed.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -1719,13 +1742,50 @@ function parseWhenCondition(raw: unknown): WhenCondition {
...
+
+/**
+ * Parse a block directive's sibling `steps:` list: non-empty, every entry an
+ * object, each parsed one level deeper so the shared depth cap sees the whole
+ * chain. `emptyDetail` is the directive's own message for an absent or empty
+ * list — the one part worth phrasing per directive, since it is where an author
+ * lands when they wrote the guard but not the body.
  */
-const MAX_WHEN_DEPTH = 20;
+function parseBlockSteps(
```
hubgan, line 2428: This takes `depth` and threads it into the recursion but never asserts it. The cap lives in `assertBlockDepth`, which each directive's parser has to remember to call first, and nothing in the types or the tests ties the two together. A second directive that calls this and omits the assert gets no cap at all: a cyclic alias then recurses through parseX -> parseBlockSteps -> fromYamlStep -> parseX unbounded and escapes `parseFlow` as a raw RangeError instead of the structured error flow-when.test.ts:178 pins. That is the same hole the cap was added to close, and the reason the commit message gives for the counter having to be shared.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -1719,13 +1742,50 @@ function parseWhenCondition(raw: unknown): WhenCondition {
...
+  raw: Record<string, unknown>,
+  depth: number,
+  emptyDetail: string
+): FlowStep[] {
+  if (!Array.isArray(raw.steps) || raw.steps.length === 0) badEntry(raw, emptyDetail);
+  return (raw.steps as unknown[]).map((s) => {
+    if (s !== null && typeof s === "object") return fromYamlStep(s as YamlStep, depth + 1);
+    return badEntry(s, "step must be an object");
```
hubgan, line 2437: This message is asserted nowhere in the suite, neither here nor at the top-level entry parse in flow-utils.ts:2216, so the two copies are free to diverge and either could be dropped unnoticed. No test feeds a non-object entry into a block's `steps:` list.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -1673,7 +1675,8 @@ async function execSteps(state: ExecState, steps: FlowStep[], scope: StepScope):
         ...depthOf(scope),
         reason: `step needs a device but the flow was resolved as device-free — pass an explicit device`,
       });
-      if (step.kind === "when") reportBlockSkipped(state, step.steps, childScope(scope));
+      const inner = blockSteps(step);
+      if (inner) reportBlockSkipped(state, inner, childScope(scope));
```
hubgan, line 1786: Two observations about this branch.

For a `when` step it cannot currently fire. `state.device` is null only when `flowRequiresDevice` returned false at flow-run.ts:1151; that maps `stepRequiresDevice` over the root flow's top-level steps only, and both `when` and `run` return true there. So no block step can reach a device-free run, and `blockSteps(step)` on this path always returns undefined - one of the four generalized sites is dead. The branch's reason string is also asserted in no test, so the whole path is uncovered.

Why it holds is worth flagging for the next directive: `flowRequiresDevice` (flow-device.ts:193) never recurses into block children, so the invariant rests entirely on `stepRequiresDevice` hardcoding `case "when": return true`. The exhaustiveness check there forces a new kind to be classified, but not to be classified `true`, and a block whose header reads nothing off the device would naturally be classified false. A flow like `[{repeat: 3, steps: [{tap: ...}]}]` would then resolve device-free and hard-stop on its own inner `tap`.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -1720,8 +1722,8 @@ function describeWhenCondition(cond: WhenCondition): string {
  * run where the block was skipped (unmet guard, errored guard, hard stop, or
  * cancellation) produces the same report shape (one line per authored step,
  * at the same depth) as a run where it entered, and reports stay comparable
- * run-to-run. Nested when blocks expand (their literal steps are known); a
- * `run:` composition stays one line, matching how post-hard-stop skips report
+ * run-to-run. Nested block directives expand (their literal steps are known);
```
hubgan: This line was generalized to "Nested block directives expand", but the opening sentence of the same docstring, two lines up, still reads "Report every step of a `when:` block that will not run as skipped". The function now receives an already-extracted array and recurses through `blockSteps`, and none of its callers test `kind` any more, so that opener is the only thing still scoping it to `when`.

hubgan (commented):
Checked this for behavioral equivalence rather than for new behavior, and it holds. I bundled the pre-PR and post-PR `parseFlow` separately and ran an identical 61-case YAML battery through both (multi-check collisions, the depth boundary, `steps: [[]]` / `[null]` / `[42]` / non-array, cyclic aliases, prototype-key siblings), comparing full parse output on success and the full failure payload on error: six lines differ across all 61 cases, all of them the documented depth-cap rewording, with error code, stage, area and kind byte-identical. The four converted runtime sites are equivalent by construction, since `blockSteps(step)` is truthy exactly when `kind === "when"` and an empty array is still truthy. 779 flow tests pass, prettier is clean, and tsc reports nothing on either changed file.

The notes inline are about the completeness of the generalization and its test and doc coverage, not about the conversion itself.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Confirmed and fixed in 772c5da — `walkSteps` now recurses through `blockSteps`. You were right on both counts: it's the one shape tsc can't catch, and the preflight is what makes it matter rather than incidental.

Its docstring no longer claims `when:` is the parser's only nesting, and `assertUploadSelfContained`'s "even inside a `when:` block" is now "at any depth inside a block directive". `blockSteps`' own JSDoc says five sites rather than four and names the preflight as the fifth reader — the count was wrong the moment this site was missed, so bumping the number alone would have left the reader unable to find it.

No behaviour change today: `blockSteps` returns `step.steps` exactly when `kind === "when"`, so the recursion set is unchanged.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: First half fixed in 2aea303, structurally rather than by test. `BLOCK_DIRECTIVE_KEYS` is now `["when"] as const satisfies readonly FlowStep["kind"][]`, and `blockSteps` narrows through it with a `step is Extract<FlowStep, { kind: BlockDirectiveKind }>` predicate instead of restating the kind. That turns both directions into compile errors rather than silent disagreement: adding `"tap"` gives `TS2339: Property 'steps' does not exist…` on `blockSteps`, and a typo like `"wen"` gives `TS2820 … Did you mean '"when"'?`. Both verified by mutation. The agreement test iterates the list and fails loudly (`no flow fixture for block directive \`X\``) when a key has no fixture, so a second directive can't join without being covered.

On the second half I'd push back. The four report-expansion tests can't be driven off a shared table the way the agreement test can, because each directive needs its own header YAML — `when: { visible: X }` and `repeat: 3` don't come from the key, so the table would have to carry a hand-written fixture per directive either way. With one entry that generates nothing a literal doesn't, and building the machinery now means building it against a header syntax that doesn't exist yet. The part that does generalize for free is the agreement test, which picks up `repeat:` the moment it's registered; the per-directive report fixtures land naturally in #749, where the second header actually exists.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Both fixed in 60cb461 — `grep MAX_WHEN_DEPTH` is now empty across the repo.

Also caught the citation sitting right beside one of them: `flow-composition.test.ts` described the one-level-peek regression as `yield* step.steps` in place of `yield* walkSteps(step.steps)`, which stopped matching `walkSteps`' body after 772c5da.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Both pinned in f8c5b0a.

The cyclic-alias test now asserts the whole message as a literal — wording, cap number and the alias hint — plus the structured signal (`error_kind: "validation"`, `FLOW_ENTRY_UNRECOGNIZED`), which a raw `RangeError` doesn't carry. A new test generates 20 and 21 levels of nesting programmatically: the 20 side walks the parsed tree and asserts depth 20 with the authored leaf, so a silently-truncating parser can't pass it; the 21 side asserts it still lands as the structured depth error rather than escaping.

Mutation-checked one at a time: `block directives` → `when blocks` fails only these; `MAX_BLOCK_DEPTH = 19` fails the at-the-cap side; `= 21` fails the one-deeper side. The literal 20 in the test is deliberate — reading the constant would move with it and pin nothing.

Your read of `flow-utils.test.ts` was right, and it's worth recording why: that fixture is a cyclic `within:` alias, so it throws the `MAX_SELECTOR_SCOPES` message, whose own text also contains "cyclic YAML alias" — the alternation is satisfied by the wrong half and the `nest deeper than` branch never fires. Left alone as the selector-scope test it actually is.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Fixed in c9997ab: `parseBlockSteps` now asserts the cap itself, on the one path every block directive has to go through to recurse.

The early call in `parseWhenStep` stays rather than moving, because moving it would have run the assert after the `else`/single-key/condition checks and changed which error a cyclic alias reports — the ordering `assertBlockDepth`'s own JSDoc documents. So the split is now explicit: the early call buys error precedence, `parseBlockSteps` buys the cap. `depth` is the same value at both sites, so for a directive that made the early call the second check is a no-op.

Verified by deleting the early call and confirming the cyclic-alias fixture still throws the structured depth error; deleting both brings back `Maximum call stack size exceeded`, so the experiment isn't vacuous.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Both copies pinned in 8af0689, and independently — mutating one site's text fails only that site's tests, which is the property worth having here, since the whole risk you named is the two drifting apart. Each asserts the full `Unrecognized flow entry (step must be an object)` plus `FLOW_ENTRY_UNRECOGNIZED`, so a raw `TypeError` wouldn't satisfy them.

Cases are the two the guard actually separates (`null` and a non-object primitive), nothing beyond that. The two older top-level entry tests turned out to be strictly subsumed — same fixture, weaker assertion — so they fold into the new `it.each` rather than sitting beside it.

packages/tool-server/src/tools/flows/flow-run.ts
j-piasecki: Second observation acted on in 1596b89: `flowRequiresDevice` now recurses through `blockSteps`, so a flow's device requirement follows the steps it actually contains rather than resting on `stepRequiresDevice` hardcoding `case "when": return true`. Strictly a no-op today — `when` classifies true, so `.some()` short-circuits before the recursion is ever entered — which also means no test can distinguish before from after, so I added none rather than one that passes either way. Both JSDoc blocks now say where the invariant lives, and the `when` bullet notes that it classifies the header only.

On the first observation I'll leave the branch as it is. The analysis is right: no block step can reach a device-free run today, so `blockSteps` there always returns undefined. But that branch is a can't-happen consistency check for the two decisions disagreeing, not a live path — there is no input, correct or otherwise, that reaches it, so there's nothing to assert the reason string against, and a test would have to fabricate an inconsistent `ExecState` to exist at all. Generalizing it wasn't a wasted conversion either: it stops being unreachable the moment a directive is classified `false`, which is exactly the case you're flagging. And after 1596b89 that classification no longer has to be right for the flow to resolve correctly.

packages/tool-server/src/tools/flows/flow-run.ts
j-piasecki: Fixed in 0b08465. The opener now reads "Report every step of a block directive that will not run as skipped", and the guard cases in the parenthetical are attributed to `when:` specifically rather than framing the whole function around it.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -2270,13 +2325,58 @@ function parseWhenCondition(raw: unknown): WhenCondition {
...
  * escapes parseFlow as a raw RangeError instead of a structured parse error.
+ *
+ * ONE counter shared by every block directive, deliberately: a per-directive
+ * counter would let an alternating chain evade all of them and reopen the hole.
+ */
+const MAX_BLOCK_DEPTH = 20;
+
+/**
+ * Guard a block directive's recursion depth. Called FIRST in a block's parse —
+ * before its own key/shape checks — so a cyclic alias reports as a depth error
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This rationale does not hold for the case it names. For a cyclic alias the same object repeats at every level, so a shape defect in it fires at depth 0 — long before 20 — and a well-formed cycle reaches the cap either way; the early call cannot be what decides it.

Deleting `assertBlockDepth(raw, depth)` from `parseWhenStep` and running the repo's own cyclic-alias fixture from `flow-when.test.ts:183` gives a message byte-identical to unmutated:

```
block directives nest deeper than 20 levels — check for a cyclic YAML alias (`steps: &s … steps: *s`)
```

Deleting *both* asserts is what changes it, to a raw `RangeError: Maximum call stack size exceeded`.

The inputs the early call actually changes are non-cyclic — a 20-deep `when` nest whose innermost entry carries a second defect. Unmutated, all three report the depth error; with the early call deleted they report:

| innermost also has | without the early call |
|---|---|
| `else:` | `when has no else — paths may only reconverge…` |
| a third sibling `foo: 1` | `a when step takes exactly { when: <condition>, steps: [...] }` |
| `when: { nope: X }` | `when needs exactly one condition key (exists, visible, hidden, text, platform)` |

Nothing pins that precedence: deleting the early call leaves the flow suite at 43 files / 1032 tests passing, so the one behaviour this JSDoc names as the line's entire purpose can regress unnoticed. The neighbouring pins do discriminate for contrast — `MAX_BLOCK_DEPTH` 20→19, 20→21, and rewording the message each fail 2 tests.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -726,6 +726,34 @@ export type FlowFile = {
...
 
+/**
+ * The literal child steps of a block directive, or undefined for a leaf step.
+ *
+ * The single predicate for "this step has authored children". The runner reads
+ * it wherever a block does NOT execute — a hard stop, a device-free flow, a
+ * cancellation, an unmet guard — to expand the block's steps as skip lines, so
+ * a report keeps one line per authored step no matter where the run ended, and
+ * once more in the upload preflight's walk, where a block it cannot see hides a
+ * nested `run:`/`snapshot` from validation. Five sites asked `kind === "when"`
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This paragraph reads as a closed enumeration of where the predicate is read, and two parts of it no longer match the code at 603835a8.

`flowRequiresDevice` (`flow-device.ts:262`) is a sixth reader, added by a later commit on this branch (`96748ae0`), but the list closes at five with "once more in the upload preflight's walk". Someone auditing block coverage from this docstring counts five and misses the one this PR itself introduced.

"an unmet guard" is also not a site that reads `blockSteps` — `execWhenStep` passes `step.steps` directly at `flow-run.ts:1856`, `:1866` and `:1879`. The fifth *direct* reader is `reportBlockSkipped`'s own nested-block recursion at `flow-run.ts:1810`, which the list never names.

The "Five sites asked `kind === "when"` directly" count itself checks out: I enumerated all six `step.kind === "when"` tests in the merge-base `flow-run.ts` (857, 1715, 1732, 1747, 1757, 1801) and confirmed 1757 is the `execWhenStep` dispatch, a different question, leaving exactly five.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -1966,6 +1994,33 @@ const STEP_DIRECTIVE_KEYS: readonly string[] = [
...
+ * restating the kinds, so parse time and run time cannot answer differently.
+ * Two constraints keep an entry honest: `satisfies` rejects a key that is not a
+ * real step kind, and blockSteps' `Extract` rejects a kind whose step type
+ * carries no `steps` (a directive listed here without children is a compile
+ * error, not a silent runtime `undefined`).
+ *
+ * At parse time these keys are exempt from the single-key sibling check,
+ * because their own parser validates their siblings with pointed messages.
+ */
+export const BLOCK_DIRECTIVE_KEYS = ["when"] as const satisfies readonly FlowStep["kind"][];
```
latekvo, line 2027: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Both constraints named here validate only entries that are *already* in the list, so neither closes the direction that "parse time and run time cannot answer differently" depends on: nothing forces a new children-carrying step kind **into** the registry. `satisfies` checks that a listed key is a real kind; `Extract` checks that a listed kind carries `steps`. There is no counterpart to the `never` binding that `stepRequiresDevice` uses to force a new kind to be classified at all.

Built in a scratch worktree at this SHA: a `repeat` directive whose children live inside its value (`repeat: { times: 2, steps: [...] }`), parsed through `parseBlockSteps`, but absent from `BLOCK_DIRECTIVE_KEYS`. `tsc --build` is clean and `parseFlow` is clean, yet:

```
parsed     = [{"kind":"repeat","times":2,"steps":[{"kind":"run",...},{"kind":"snapshot",...}]}]
blockSteps = undefined
```

So `walkSteps` never descends, and an uploaded flow's nested `run:`/`snapshot` reaches execution — verbatim the outcome `flow-run.ts:855-859` states cannot happen.

The sibling-`steps:` spelling is genuinely safe: removing `repeat` from the list there produces a loud parse error from the single-key sibling check (``a `repeat` step has unknown key `steps` — step options go inside the `repeat:` value, not beside it``). It is only the children-inside-the-value spelling that passes silently, and `parseBlockSteps` — documented at `:2363` as the one path every block directive must go through to recurse — asserts no membership.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat)\]: Reviewed at `603835a8`. What I ran, so the coverage is auditable:

- **Differential of the real parser, merge-base vs head.** 190 YAML inputs through `parseFlow` in one process, comparing thrown/not, exact message, `error_kind`/`error_code`/`failure_stage`, the parsed object as key-sorted JSON, and the `serializeFlow` round-trip. **27 differ, all 27 the intended depth-cap wording; 0 other differences.** Error *ordering* survives a 36-cell boundary matrix (depth 18/19/20/21 × `else`, empty `steps`, third sibling, bad condition, non-object child, `optional`, two directive keys, leaf sibling, clean) — the empty-`steps` check still loses to `parseWhenCondition` exactly as at base. Covered 4 cycle shapes, cycles in non-block positions, alias bombs, unicode-lookalike sibling keys, `__proto__`/`constructor`, and all 15 non-block directives × `steps`/`else`/unknown sibling.
- **End-to-end through `POST /tools/flow-execute`**, two tool-servers built from merge-base and head, 22 flows each, reports byte-compared: guard met/unmet at 3 levels, hard stop before a block, failure inside a block, a real mid-stream cancellation (socket destroy → `controller.abort()`) with the block both adjacent and non-adjacent to the abort point, a guard errored by SIGKILLing the Electron mid-probe, `run:`-fragment boundaries, and the upload path with `run:`/`snapshot:` nested 0-3 `when` levels deep. Report shapes are untouched; only the 3 parse-error flows differ, in the intended wording.
- **Mutation**, 27 isolated full-suite runs. 22 go red. `blockSteps`→`undefined` fails 10 tests across 3 files; the one-level peek (`yield* inner`) fails exactly the test whose comment claims it; removing both depth asserts restores the raw `RangeError`, so the cap does what it documents. Per-site: hard-stop 2 red, abort 1 red, `reportBlockSkipped` recursion 4 red, `walkSteps` 3 red.
- **Type-level claims executed, not reasoned about.** `BLOCK_DIRECTIVE_KEYS = ["when","wen"]` → `TS2820 … Did you mean '"when"'?` on the list; `["when","tap"]` → `TS2339: Property 'steps' does not exist` on `blockSteps`' `.steps`. Both land where the JSDoc says, verified against a control error.
- Also swept the nearest twins (`flowScopesDevice`, `toYamlStep`, `summarizeSteps`, `stepTarget`, `execSteps`' dispatch, the recorder and CLI step consumers), recursion bounds and report growth under alias bombs, and temp-file/process lifetime on the upload throw paths.

The refactor holds up: the behaviour-preservation claim is true under real inputs, and the five converted sites are well pinned. Three notes below, all on the new prose and the one unpinned invariant.

I also chased a possible gap in `flowScopesDevice` (the twin of the `flowRequiresDevice` walk this PR adds) and am **not** reporting it — nesting a scoping step is strictly safer than leaving it at top level here, since the top-level path already falls open to the machine-wide sweep whenever resolution is ambiguous, and `#749` classifies `repeat` `true`, so the premise never materialises.

packages/tool-server/src/tools/flows/flow-run.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Still present as of `603835a8`, and the reason given for leaving it no longer holds.

The reply says the branch "stops being unreachable the moment a directive is classified `false`". Under that classification the branch's own guard `stepRequiresDevice(state.registry, step)` is false for that step, so it still cannot fire; and under a `true` classification `flowRequiresDevice` is true, a device is resolved, and `!state.device` is false. Both classifications are dead, so `blockSteps` here is unreachable for every present *and* future block directive — and the recursion added to `flowRequiresDevice` in `96748ae0` is what closes the last route, by making the device-free walk exhaustive over block children.

Measured two ways. Instrumenting the branch and running 22 flows through `POST /tools/flow-execute` (device-free, plus a Chromium/Electron device for the guard cases) gives **0 hits** here, against 5 for the hard-stop site, 1 for the abort site and 52 for `reportBlockSkipped`'s recursion. And deleting `const inner = blockSteps(step); if (inner) reportBlockSkipped(...)` from this branch alone leaves the suite at 43 files / 1032 tests passing, while the same deletion at the hard-stop and abort sites fails 2 and 1 tests respectively.

packages/tool-server/src/tools/flows/flow-utils.ts
hubgan, line 2027: Registering a key here reaches the report expansions, the upload preflight and the parse-time sibling exemption, but not the step dispatch in `execSteps` (`if (step.kind === "when")`), which is the site that actually runs the block.

I added a second block kind (`{ kind: "repeat"; times: number; steps: FlowStep[] }`) to `FlowStep` and to this list, then ran `tsc --noEmit`. The compiler flags `flow-device.ts:239`, `flow-finish-recording.ts:221` and `flow-utils.ts:1321-1324`, and reports nothing at all in `flow-run.ts`. A directive registered here but absent from that dispatch therefore compiles clean and falls through to `execLeafStep`'s `default` arm at run time, reporting a single `unsupported step kind` error; because `execWhenStep` is what calls `reportBlockSkipped` on the not-taken path, its children then appear in neither the executed report nor a skip expansion.

That is the one block-related site a second directive can still miss, and it is the one whose failure costs the whole block.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -1966,6 +1994,33 @@ const STEP_DIRECTIVE_KEYS: readonly string[] = [
   "snapshot",
 ];
 
+/**
+ * The directive keys that carry a sibling `steps:` list — the single registry
+ * of what a block directive is: {@link blockSteps} reads THIS list rather than
+ * restating the kinds, so parse time and run time cannot answer differently.
+ * Two constraints keep an entry honest: `satisfies` rejects a key that is not a
```
hubgan: This makes "block directive" a precise term meaning "carries a sibling `steps:` list", and the `Extract` guard provably excludes `run:` from it (listing `run` here fails with `Property 'steps' does not exist on type '{ kind: "run"; flow: string; }'`).

Two comments outside this package already use the term with the opposite meaning: `packages/argent-cli/src/flow.ts:43-44` and `packages/argent-mcp/src/content.ts:232-233` both describe report depth as "+1 inside each block directive (`when:` guarded steps, `run:` fragment steps)". After this PR the codebase has one name for two different sets.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -2453,12 +2542,12 @@ function fromYamlStep(raw: YamlStep, whenDepth = 0): FlowStep {
...
-  // step also carries siblings (`steps`, and the rejected `else`), but
-  // parseWhenStep validates them itself with pointed messages, so the generic
-  // check stays out of its way.
+  // value, so a sibling key is a mis-nested or misspelled option. A block
+  // directive also carries siblings (`steps`, and whatever it rejects beside
+  // it), but its own parser validates them with pointed messages, so the
+  // generic check stays out of its way.
   const kind = kinds[0]!;
-  if (kind !== "when") {
+  if (!isBlockDirectiveKey(kind)) {
```
hubgan, line 2611: Registration in `BLOCK_DIRECTIVE_KEYS` now doubles as an opt-out from the generic unknown-sibling-key rejection. The comment above justifies the exemption with "its own parser validates them with pointed messages", but nothing holds a new entry to that promise: a directive added to the list whose parser does not check its own siblings loses sibling validation entirely and silently accepts a misspelled option, where every non-block directive would reject it.

The two guards on the list (`satisfies` and `Extract`) both check the kind's shape; neither checks that a validating parser exists.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -2270,13 +2325,58 @@ function parseWhenCondition(raw: unknown): WhenCondition {
...
+ * say. That early call buys the error PRECEDENCE only, not the cap itself:
+ * {@link parseBlockSteps} asserts again before it recurses, so forgetting the
+ * early call costs a directive the precedence and nothing more.
  */
-const MAX_WHEN_DEPTH = 20;
+function assertBlockDepth(raw: unknown, depth: number): void {
+  if (depth >= MAX_BLOCK_DEPTH) {
+    badEntry(
+      raw,
+      `block directives nest deeper than ${MAX_BLOCK_DEPTH} levels — check for a cyclic YAML alias (\`steps: &s … steps: *s\`)`
```
hubgan: This string is the only user-visible behavior change in the PR. I ran 43 malformed and well-formed `when:` inputs through the pre-PR and post-PR parsers side by side and diffed every outcome (AST, message, failure code, serialized form): six differ, and all six are this message.

The new wording is less actionable than the text it replaces. `when` is the only block directive today, so an author who hits the cap wrote `when:` and nothing else; the message now drops that word. And "block directive" appears nowhere author-facing — `argent-create-flow/SKILL.md:235` documents `when:` blocks at length and never uses the term, and no skill doc mentions a nesting limit at all — so the reader is handed a concept they cannot look up in place of the one they typed.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -2285,12 +2385,7 @@ const MAX_WHEN_DEPTH = 20;
  * back on the known path), so paths may only reconverge, never diverge.
  */
 function parseWhenStep(raw: Record<string, unknown>, depth: number): FlowStep {
-  if (depth >= MAX_WHEN_DEPTH) {
-    badEntry(
-      raw,
-      `when blocks nest deeper than ${MAX_WHEN_DEPTH} levels — check for a cyclic YAML alias (\`steps: &s … steps: *s\`)`
-    );
-  }
+  assertBlockDepth(raw, depth);
```
hubgan, line 2448: This early call is what decides which error an author sees when a block is both over the cap and independently malformed, and nothing in the suite pins it.

I removed this line in a worktree build: all 314 tests in `flow-when` + `flow-utils` + `flow-composition` still pass. The new cap tests nest a well-formed `when`, and for that shape `parseBlockSteps`' assert produces the identical message, so they cannot see the difference.

The line is load-bearing, though. With it removed, a 21-level nest whose innermost block also carries `else:`, an extra sibling key, `platform: bogus`, or an unknown condition key reports that shallower error instead of the depth error — all four inputs I tried changed their reported message. So the precedence property the doc comment describes is real and currently undefended.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -726,6 +726,34 @@ export type FlowFile = {
...
+ * a report keeps one line per authored step no matter where the run ended, and
+ * once more in the upload preflight's walk, where a block it cannot see hides a
+ * nested `run:`/`snapshot` from validation. Five sites asked `kind === "when"`
+ * directly before this existed; a second block directive would have had to
+ * remember all five, and a forgotten one drops a whole block from the report —
+ * or from the preflight — silently. Now it is one case here — and it is the
+ * same case the PARSER exempts from the single-key sibling check, because the
+ * kinds come from {@link BLOCK_DIRECTIVE_KEYS} rather than being restated.
+ */
+export function blockSteps(step: FlowStep): FlowStep[] | undefined {
```
hubgan, line 755: This reads as a census of the helper's readers — five runner sites plus the upload preflight walk — but it omits the reader this same PR adds, `flowRequiresDevice` in `flow-device.ts`.

Since the comment's purpose is to tell the next person adding a block directive where the helper is consulted, leaving out device resolution is the omission most likely to cost them: it is the one reader outside `flow-run.ts`.

packages/tool-server/src/tools/flows/flow-device.ts
```diff
@@ -241,9 +243,24 @@ export function stepRequiresDevice(registry: Registry, step: FlowStep): boolean
...
+ * {@link stepRequiresDevice}, and a flow that is only that block would then
+ * resolve device-free and hard-stop on the first device step in its body. The
+ * exhaustiveness check there forces a new kind to be classified, but not to be
+ * classified `true`; the walk is what makes either answer safe.
+ */
 export function flowRequiresDevice(registry: Registry, steps: FlowStep[]): boolean {
-  return steps.some((step) => stepRequiresDevice(registry, step));
+  return steps.some(
+    (step) =>
+      stepRequiresDevice(registry, step) || flowRequiresDevice(registry, blockSteps(step) ?? [])
```
hubgan, line 264: The right-hand side of this `||` cannot execute today. `blockSteps` returns a list only for `when`, and `stepRequiresDevice` returns `true` unconditionally for `when` two functions up, so the left operand short-circuits in exactly the cases where the recursion would have children to walk; every other kind reaches it as `flowRequiresDevice(registry, [])`.

I diffed this against the previous `steps.some(stepRequiresDevice)` over a corpus of flows — device-free leaves, device leaves, a `when` over a device-free body, three-deep nested `when`, device-free and device-bearing `tool` steps — and the answer is identical in every case.

Two consequences worth deciding on. The doc above states the new behavior in the present tense ("block children included", "Both halves matter", "a block's body is answered by `flowRequiresDevice`'s walk"), which reads as though `when` bodies are now walked; they are not, and will not be until a block header can be classified `false`. And nothing tests it: no test in the repo calls `flowRequiresDevice` at all (`flow-deviceless.test.ts` covers only `stepRequiresDevice`), so the whole recursion can be deleted with 1032/1032 flow tests still green.

hubgan (commented):
Reviewed the block-directive generalization. Behavior-preservation checked by diffing the pre-PR and post-PR builds directly: 43 parser inputs (identical except the six carrying the renamed depth-cap message), and full flow runs on a booted iOS simulator, Android emulator and Vega device (reports byte-identical on all three). Notes are inline.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -1729,7 +1738,8 @@ async function execSteps(state: ExecState, steps: FlowStep[], scope: StepScope):
         ...depthOf(scope),
         reason: `step needs a device but the flow was resolved as device-free — pass an explicit device`,
       });
-      if (step.kind === "when") reportBlockSkipped(state, step.steps, childScope(scope));
+      const inner = blockSteps(step);
```
hubgan, line 1785: This expansion cannot run. `state.device` is null only via the two `return { device: null, ... }` paths, both gated on `!flowRequiresDevice(registry, flow.steps)`; that predicate being false means no top-level step is `when` or `run`, since both are classified as requiring a device. So by the time this branch is reachable, the step in hand is an `echo`, a `wait` or a device-free `tool` — none of which has children.

Deleting these two lines leaves all 1032 flow tests passing. It predates the PR, but the PR rewrote it, so it is a reasonable moment to decide whether it stays.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Confirmed and fixed in 52a7e846. You were right that the two constraints only judge keys already in the list, and that `parseBlockSteps` asserts no membership - the direction that "parse time and run time cannot answer differently" actually rests on was never closed.

Reproduced your scratch before fixing it: adding a child-bearing kind to `FlowStep` without registering it gives 8 tsc errors, and not one of them mentions `BLOCK_DIRECTIVE_KEYS` or `blockSteps`. They all come from unrelated exhaustiveness (`stepRequiresDevice`'s `never` binding, the `tool`-kind property accesses in `toYamlStep` and flow-finish-recording), so a directive author satisfies every one of them and still ships a registry gap.

The missing direction is now a compile-time binding over the kinds whose step type carries `steps`:

```
error TS2322: Type 'true' is not assignable to type '"repeat"'.
```

Spelled as a conditional rather than a bare `never` deliberately - `not assignable to 'never'` names nothing, and the whole value here is that the error tells you which kind you forgot. Verified in both directions: your unregistered value-carried `repeat` now fails tsc where it was silent, registering it goes clean, and the two existing constraints still fire on `["when", "tap"]` and `["when", "wen"]`.

One gap I did not close, since it is narrower than what you found: "child-bearing" means a *required* `steps: FlowStep[]`, so a kind spelled `steps?: FlowStep[]` still slips past. Three of the four near-miss spellings are self-limiting (naming the field anything but `steps`, nesting it, or making it `readonly` also makes the kind impossible to register, so the author hits a wall immediately); the optional one is genuinely asymmetric, but it contradicts `parseBlockSteps`' own invariant that it returns a non-empty `FlowStep[]`, and no step type in the union spells a required field optional. Closing it needs a `"steps" extends keyof S` distributive conditional, which costs more legibility than the case is worth today.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Both corrections confirmed and fixed in ea839bdf. I re-enumerated rather than taking the counts on faith, and got exactly your six: `flowRequiresDevice` (flow-device.ts:262), `walkSteps` (flow-run.ts:864), the three `execSteps` gates (:1723, :1741, :1757), and `reportBlockSkipped`'s own nested-block recursion (:1810). And `execWhenStep` does pass `step.steps` straight through at :1856/:1866/:1879, so "an unmet guard" was never a reader - it displaced the one that is.

The paragraph now groups the six by what they read children *for*: four expand a block that will not run, one feeds the upload preflight, and the device walk resolves a requirement rather than skipping anything. The unmet guard keeps a parenthetical, because it does still produce skip lines and dropping it entirely would read as a behaviour claim I did not mean.

Your history check holds up too - I verified the five against the merge base independently (857, 1715, 1732, 1747, 1801 yield children; 1757 is the dispatch). The sentence stays, but now says *which* five, so it cannot be read as naming today's six.

One thing your comment prompted that I got wrong on the first pass: I initially wrote the device walk as fixing an answer top-level steps get wrong today. It does not - `when` classifies device-requiring, so `.some()` short-circuits before the recursion is ever entered, exactly as 8fe589d2's message says. The text now states the hypothetical it guards against and marks it a no-op while `when` is the only block kind, matching `flowRequiresDevice`'s own docstring instead of inventing a third phrasing of it.

`flowRequiresDevice` is named the way this file names its other cross-file symbols rather than with `{@link}`: importing it here would be circular, so the link would not resolve.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: You are right on both halves, and I reproduced your table before touching anything. Fixed in 269d0f42.

Deleting the early `assertBlockDepth` from `parseWhenStep` leaves the cyclic-alias fixture byte-identical, and the three at-the-cap non-cyclic inputs flip exactly as you listed - `else:` to "when has no else", a third sibling to "a when step takes exactly {...}", `when: { nope: X }` to "needs exactly one condition key". The flow suite stayed fully green under that mutation, so the one behaviour the line exists for was regressible in silence.

The JSDoc now names the case the call actually decides - an entry that has REACHED the cap and also carries a second defect - and says outright that a cyclic alias is not it, with your reason: the same object repeats at every level, so a shape defect fires at depth 0 and a well-formed cycle reaches the cap through `parseBlockSteps`' assert either way. The "buys PRECEDENCE only, not the cap" half was already true and stays.

The pin is an `it.each` over all three spellings, asserting both that the depth message is reported and that the shallower check's message is *not* - the second assertion is the one that actually captures the difference, since "contains the depth error" alone would survive a reordering. Sensitivity checked the way you checked the neighbours: deleting the early call fails exactly these three and nothing else, and the cyclic-alias and 20/21 tests still pass, which is what "precedence only, not the cap" means operationally.

`parseBlockSteps`' and `MAX_BLOCK_DEPTH`'s docstrings I left alone - the same evidence supports both (the double-deletion RangeError is exactly what they name), so only the early-call rationale was wrong.

Not covered: the other at-cap spellings that also flip (a `timeout` sibling, `when: 5`, a bad platform value, `when: { idle: true }`). Any one of the three chosen already discriminates the mutation, and the three span both `parseWhenStep`'s own checks and `parseWhenCondition`, so the extra cases would add runtime without adding signal.

packages/tool-server/src/tools/flows/flow-run.ts
j-piasecki: You are right and my reason was wrong. Corrected in 95b2cddc.

I re-derived it rather than just accepting the measurement: `state.device` is null only through flow-run.ts:1186-1187, gated on `!flowRequiresDevice`, and since 8fe589d2 that walk recurses through `blockSteps`, so no nesting hides a step from it. `stepRequiresDevice` returns true for `run` as well as `when`, so a block inside a fragment does not escape either - a root flow containing `run:` is never resolved device-free. That leaves your dichotomy intact in both directions: under `true` a device is resolved and `!state.device` is false; under `false` the guard's own `stepRequiresDevice` conjunct fails. "It stops being unreachable the moment a directive is classified `false`" was backwards - that classification is the one that most clearly cannot reach it.

What I have not done is delete the expansion, and I want to be explicit that this is a disagreement about the remedy rather than about the facts. The enclosing branch is a can't-happen consistency check that nobody is proposing to remove; if it ever does fire, a block whose children went unexpanded produces a report that violates the one-line-per-authored-step invariant in precisely the case the check exists to report. Deleting the two lines optimises a coverage number and costs the report shape at the only moment it would matter, so the asymmetry with the hard-stop and abort sites would be a defect rather than a cleanup.

So the commit is comment-only: it replaces the wrong reason with the two-case argument, names `flowRequiresDevice`'s recursion as what makes it hold, and says why the expansion stays anyway. The next reader gets the unreachability without re-deriving it, and specifically does not get my mistaken belief that a device-free classification revives it.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Closed from both directions. 405c9e187344 forces the dispatch: `execSteps` now narrows through `isBlockStep` into `execBlockStep`, whose never-bound default arm makes a registered-but-undispatched kind a tsc error naming it - the scratch `repeat` mutation now fails in flow-run.ts where it previously compiled silent. Behavior-preserving today: with `when` the only entry, `isBlockStep(step)` is exactly `step.kind === "when"`. And 52a7e846f6f2 adds the missing reverse binding - a child-bearing kind absent from the registry is also a compile error naming the kind - so a second directive can miss neither the registry nor the one site whose failure costs the whole block.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Fixed in 8e9e8d08aa3a: the run-inclusive depth set is no longer called "block directive" anywhere. The five sites that used the words for it - the two depth docstrings you found in argent-cli and argent-mcp, `childScope` and `depthOf` in the runner, and the composition test annotating run-only depths - now say "nesting step", so the codebase has one name per set. Renaming the depth set rather than the registry was settled by the sharper version of your observation: under the precise meaning, a flow with `run:` but no `when:` has no block directives yet stamps depth, so those docstrings were not merely ambiguous but false.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Held in 0efa1d61bc3f, at the test layer - the compile-time guards judge shapes, as you note, and "a validating parser with pointed messages exists" is only checkable by running one. The pin is a fixtures table typed as a full `Record` over the registry's key type, so registering a directive without a sibling-rejection fixture is a typecheck error, and the test drives each fixture's bogus sibling through `parseFlow` and asserts it dies with the directive's own pointed message rather than being silently accepted. Sensitivity checked by bypassing `parseWhenStep`'s exact-keys check: exactly the new test fails, on the silently-accepted sibling.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Fixed in 0a839647c0bd. The prefix now derives from BLOCK_DIRECTIVE_KEYS, so the message reads `` `when:` blocks nest deeper than 20 levels `` today - the word the author typed is back - and it names every registered directive automatically once more register. Naming `when` statically was rejected for the original reason: it would misdirect an author who hits the cap through a different directive once one exists. The test pins stay full literals, so a new registration updates the expectation consciously rather than the message drifting unpinned.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Pinned in 269d0f4290aa, in all three spellings the parser checks after the early call - `else:`, a third sibling key, and an unknown condition key, each sitting at the cap - asserting both that the depth error is reported and that the shallower check's message is not. Sensitivity checked by deleting the early call: exactly these three fail and the two neighbouring depth tests still pass, which is "precedence only, not the cap" made observable. The JSDoc now names this as the case the call decides, and says explicitly that a cyclic alias is not it - as your experiment showed, the cycle reports identically without the early call, since a shape defect in a cycle fires at depth 0 and a well-formed cycle reaches the cap through `parseBlockSteps`' assert anyway.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Corrected in ea839bdfe582. The census is now six, grouped by what they read children FOR - four expand a block that will not run, one feeds the upload preflight, and `flowRequiresDevice`'s walk resolves a requirement rather than skipping anything - with the new reader named the way this file names its other cross-file symbols, since importing it for a {@link} would be circular. Writing it up surfaced a second defect: "an unmet guard" was never a reader at all - `execWhenStep` has the block in hand and passes `step.steps` straight to `reportBlockSkipped` - and the entry it displaced, `reportBlockSkipped`'s own recursion into a nested block, was never named. The "five sites asked `kind === "when"` directly" count was still correct; the text now says which five, so it cannot be read as naming today's six.

packages/tool-server/src/tools/flows/flow-device.ts
j-piasecki: Both consequences taken, in 68a5541b4f6b. The doc's first paragraph now says the no-op part out loud - the header half short-circuits in exactly the cases where the recursion would have children, so the walk answers nothing until a block header can classify `false` - and `stepRequiresDevice`'s cross-reference no longer implies the walk answers anything yet. The walk itself stays: it is the declared guard for that future classification, and the unreachability argument on flow-run's device-free branch (95b2cddc1386) leans on it. On the test gap: seven tests now pin `flowRequiresDevice`'s answers across leaf, tool, when-headed and nested shapes - explicit that they cannot discriminate the recursion itself, since no FlowStep value reaches it while `when` is the only block kind.

packages/tool-server/src/tools/flows/flow-run.ts
j-piasecki: Kept, with the reason on record corrected in 95b2cddc1386 - the old one ("stops being unreachable once a block directive is classified device-free") was wrong in both directions. Under that classification the guard's own `stepRequiresDevice` conjunct is false, so the branch still cannot fire; under the opposite one `flowRequiresDevice` resolves a device and `!state.device` is false - and since its walk recurses through `blockSteps`, no nesting hides a step, with a block inside a `run:` fragment covered because `run:` itself classifies device-requiring. So the expansion is dead for every present and future block directive, not just today's, exactly as you measured. It stays because the enclosing branch is a can't-happen check between two decisions that cannot disagree, and if they ever did, a block whose children went unexpanded would break the one-line-per-authored-step invariant in precisely the case the check exists to report.

packages/tool-server/src/tools/flows/flow-device.ts
```diff
@@ -241,9 +243,25 @@ export function stepRequiresDevice(registry: Registry, step: FlowStep): boolean
...
```
hubgan, line 264: This function now resolves the requirement from a block's children, but `flowScopesDevice` twelve lines below still asks its question only of top-level steps. The two are consulted as a pair in flow-run.ts - `flowScopesDevice` is reached only when `flowRequiresDevice` answered no - so today the asymmetry is masked: `when` classifies device-requiring, so a flow containing a block never reaches the second question.

It stops being masked under exactly the scenario the docblock above invokes: a block directive whose header reads nothing off the device. There, a `stop-all-simulator-servers` carrying a `devices:` scope inside the block is invisible to `flowScopesDevice`, the run resolves no device, and the teardown executes with its unscoped meaning - the machine-wide sweep that the scope exists to prevent. That also narrows the claim two lines above: the walk makes either classification safe for this function, not for the pair.

Separately, nothing in the suite exercises the new recursion - deleting it leaves all 1066 flow tests green, since `stepRequiresDevice` returns true for `when` before the right-hand side of the `||` is ever evaluated.

hubgan (commented):
Reviewed the diff and the code it transitively affects. **No blocking findings** - everything below is non-blocking.

The central claim holds up under direct measurement:

- 314 valid and malformed flow YAML inputs parsed through the merge-base and PR trees: after normalizing the one declared depth-cap message change, the two outputs are byte-identical (parse result, serialize round-trip, error message, failure signal). A mutant registry was used as a negative control to confirm the harness detects real changes at this site.
- The merge-base test files run unmodified against the PR source: 1049 passed, 0 failed.
- Both builds driven end to end through `argent flow run` against a real Chromium target - guard met, guard unmet, nested block with the outer guard unmet, hard stop followed by a block, 20 and 21 levels of nesting, cyclic alias, and an uploaded flow with a nested `run:` - byte-identical output on every one except the depth-cap string.
- Full flow suite on this branch: 1066 passed. tsc, prettier, eslint and knip all clean; knip's ceiling moves by 0.

The comments are mostly about claims the new documentation makes that the shipped code doesn't quite support, one place where the generalization stopped one function short, and a couple of gaps in what the type-level guards actually catch.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -1966,6 +2005,57 @@ const STEP_DIRECTIVE_KEYS: readonly string[] = [
...
+/**
+ * Forces a child-bearing kind INTO the registry — the direction the two
+ * constraints on {@link BLOCK_DIRECTIVE_KEYS} cannot cover, since both only
+ * judge kinds already listed. An unlisted one parses like any other step and
+ * then reads as `undefined` from {@link blockSteps}, so every reader listed
+ * there silently misses its children. Spelled as a conditional rather than a
+ * bare `never` so the compile error names the missing kind.
+ */
+const _everyChildBearingKindIsRegistered: [UnregisteredBlockKind] extends [never]
+  ? true
```
hubgan, line 2063: Checked with the repo's own tsc against a child-bearing kind added to the union and left unregistered: this catches `steps: FlowStep[]` and names the missing kind, but stays silent for `steps?: FlowStep[]` and for `steps: readonly FlowStep[]`, because `Extract<FlowStep, { steps: FlowStep[] }>` matches only a required, mutable property of that exact name.

The optional spelling is the one with no backstop on either side - `blockSteps`' `Extract` doesn't catch it either, since `step.steps` is then `FlowStep[] | undefined`, which is already `blockSteps`' declared return type. A kind spelled that way compiles clean while reading as a leaf at all six call sites, which is the silent-drop outcome this line is described as preventing.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -1821,7 +1837,28 @@ function reportBlockSkipped(
...
+    const inner = blockSteps(step);
+    if (inner) reportBlockSkipped(state, inner, childScope(scope), reason);
+  }
+}
+
+/**
+ * Dispatch a block directive to its executor. The `never` default arm is the
+ * run-time site a kind registered in BLOCK_DIRECTIVE_KEYS cannot miss: an
+ * unhandled registered kind fails tsc here instead of falling through to
+ * execLeafStep as an "unsupported step kind" error that skips the block's
```
hubgan: The alternative described here isn't the one that would occur. `execSteps` dispatches on `isBlockStep` before it reaches `execLeafStep`, so a registered-but-unhandled kind returns from this default arm and produces no report line at all - not even the block's own marker. Reaching `execLeafStep`'s "unsupported step kind" error is what the `isBlockStep` gate prevents; this arm prevents the block vanishing entirely. The benefit is attributed to the wrong construct.

Worth noting alongside it: `stepTarget` (flow-run.ts:1629) ends in a bare `default: return undefined` with no such binding. Adding an unregistered child-bearing kind does fail tsc in `stepRequiresDevice`, `toYamlStep` and `summarizeStep`, but not there - a registered block kind would render `target: undefined` on its marker and every skip line with nothing objecting.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -1966,6 +2005,57 @@ const STEP_DIRECTIVE_KEYS: readonly string[] = [
...
+  : UnregisteredBlockKind = true;
+
+/**
+ * Is this directive key a block directive? The parser and runner's only read of
```
hubgan: This claims to be the parser and runner's only read of `BLOCK_DIRECTIVE_KEYS`, but `assertBlockDepth` reads it directly to build the depth-cap message. That second read arrived two commits after this comment was written and the comment wasn't revisited.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -726,6 +726,45 @@ export type FlowFile = {
...
+ * in hand and passes `step.steps` straight to `reportBlockSkipped`.) The fifth
+ * is the upload preflight's walk, where a block it cannot see hides a nested
+ * `run:`/`snapshot` from validation. The sixth, `flowRequiresDevice`
+ * (flow-device.ts), reads children not to skip them but to resolve the flow's
+ * device requirement from a block's body — the guard against a later block
+ * directive whose header reads nothing off the device, and a no-op while
+ * `when`, whose header already classifies device-requiring, is the only block
+ * kind.
+ *
+ * Five sites asked `kind === "when"` directly before this existed — the five in
```
hubgan: The count is off by one. At the merge base, six sites in flow-run.ts asked `kind === "when"` directly: the five that became `blockSteps` readers, plus the dispatch that became `isBlockStep` / `execBlockStep`. In a docblock this precise about counts, "the five in the runner" reads as an exhaustive enumeration, and the follow-on about what a second directive would have had to remember undercounts for the same reason - it would also have had to remember the dispatch.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -2270,13 +2360,63 @@ function parseWhenCondition(raw: unknown): WhenCondition {
...
+ * this call.) That early call buys the error PRECEDENCE only, not the cap
+ * itself: {@link parseBlockSteps} asserts again before it recurses, so
+ * forgetting the early call costs a directive the precedence and nothing more.
+ */
+function assertBlockDepth(raw: unknown, depth: number): void {
+  if (depth >= MAX_BLOCK_DEPTH) {
+    const directives = BLOCK_DIRECTIVE_KEYS.map((key) => `\`${key}:\``).join("/");
+    badEntry(
+      raw,
+      `${directives} blocks nest deeper than ${MAX_BLOCK_DEPTH} levels — check for a cyclic YAML alias (\`steps: &s … steps: *s\`)`
```
hubgan, line 2409: The shipped message is `` `when:` blocks nest deeper than 20 levels `` - with one registered directive the joined list names only `when`.

The PR description says the message now reads `block directives nest deeper than 20 levels`, and argues that a message naming only `when` would misdirect an author who reached the cap through a different directive. The code does the opposite of what the description claims for it; the actual delta from the base message is `when blocks` to `` `when:` blocks ``. Confirmed end to end - the base and PR builds of the CLI differ on exactly this string and nothing else across seven flow scenarios.

The description's "559 flow tests pass unchanged" is also off: the suite is 1066 on this branch against 1049 at the merge base, so the PR adds 17.

packages/argent-cli/src/flow.ts
```diff
@@ -40,7 +40,7 @@ export interface StepReport {
   /** Human-readable step target (selector / snapshot name), set by the runner. */
   target?: string;
   /**
-   * Nesting depth: absent/0 at top level, +1 inside each block directive
+   * Nesting depth: absent/0 at top level, +1 inside each nesting step
```
hubgan, line 43: The final commit re-worded this field's docblock here and in the MCP copy, but the tool-server's own declaration at packages/tool-server/src/tools/flows/flow-run.ts:242 still reads "+1 inside each block directive's expanded steps (a `when:` block's guarded steps, a `run:` fragment's steps)". That's the source of truth these two client copies mirror, and under this PR's definition `run:` is not a block directive - so the three copies now describe the same field differently.

packages/tool-server/test/flows/flow-utils.test.ts
```diff
@@ -898,6 +926,80 @@ describe("parseFlow", () => {
   });
 });
 
+// ── block directives ─────────────────────────────────────────────────
+
+// A flow per key in BLOCK_DIRECTIVE_KEYS, with the children its block authors.
+// Deliberately not derived from the key: a fixture that agreed with the parser
+// by construction would test nothing.
+const BLOCK_DIRECTIVE_FLOWS: Partial<
+  Record<(typeof BLOCK_DIRECTIVE_KEYS)[number], { yaml: string; children: FlowFile["steps"] }>
```
hubgan: This table is `Partial<Record<...>>` while its neighbour below is a full `Record<...>`. Only the full one makes the "registering a directive without a fixture is a compile error" property hold - compiling with a second key registered raises TS2741 for the `Record` table and nothing for this one. Here the gap falls back to the runtime `toBeDefined()` at line 978, and the two `fixture!` assertions below exist only because of the `Partial`. The comment's "A flow per key in BLOCK_DIRECTIVE_KEYS" is aspirational rather than enforced.

packages/tool-server/test/flows/flow-when.test.ts
```diff
@@ -179,11 +179,94 @@ describe("when: parse/serialize", () => {
...
+      thrown = err;
+    }
+    // The whole message is pinned, not only the "nest deeper than" half: the
+    // prefix names the registered block directives (a literal on purpose - it
+    // must be updated consciously when a new directive registers, not drift
+    // with the source), and the alias hint is what points the author at the
+    // actual mistake.
+    expect(thrown).toBeInstanceOf(Error);
+    expect((thrown as Error).message).toContain(
+      "`when:` blocks nest deeper than 20 levels — check for a cyclic YAML alias (`steps: &s … steps: *s`)"
```
hubgan: The `` `when:` `` half of this literal is not a constant - it's derived from `BLOCK_DIRECTIVE_KEYS` at runtime, and it's restated by hand at three places in this file (here, 232, 263). Registering a second directive turns the prefix into `` `when:`/`repeat:` `` and all three have to be found and updated, which is the restatement the registry was introduced to eliminate. The literal 20 is a different case - the comment's argument for pinning that one holds.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -2270,13 +2360,63 @@ function parseWhenCondition(raw: unknown): WhenCondition {
...
+ * through to recurse, so no directive can opt out of the cap by forgetting the
+ * early {@link assertBlockDepth} call. `depth` is unchanged between the two
+ * calls, so for a directive that made the early one this is a no-op.
+ */
+function parseBlockSteps(
+  raw: Record<string, unknown>,
+  depth: number,
+  emptyDetail: string
+): FlowStep[] {
+  assertBlockDepth(raw, depth);
```
hubgan, line 2433: Nothing in the suite pins this second assertion: removing it leaves all 1066 flow tests green, because `parseWhenStep`'s earlier call always fires first at the same depth. The docblock is candid that it's a no-op for a directive that made the early call, so this is expected rather than wrong - but the property it exists to hold (a directive that skips the early call still gets the cap) is currently unverified.

hubgan (approved):

packages/tool-server/src/tools/flows/flow-device.ts
j-piasecki: Confirmed and fixed in ccebfc0a8204 — `flowScopesDevice` now recurses through `blockSteps` too, so the pair is symmetric.

You were right that the asymmetry is masked rather than absent, and right about which scenario unmasks it: under a block header classified device-free, a `devices` scope in the body was invisible here, the run resolved no device, and the teardown ran its unscoped meaning — the machine-wide sweep the scope exists to prevent. Rather than narrowing the claim two lines above I made it true: the walk now covers both questions, which is what "either classification is safe" was supposed to mean.

On the coverage half, one correction to my own earlier framing. `flowRequiresDevice`'s walk genuinely cannot be discriminated by any `FlowStep` value — the header short-circuits first. This one can: a `when` header scopes nothing itself, so `flowScopesDevice` over a `when` wrapping a `stop-all-simulator-servers` answers `false` without the recursion and `true` with it. The seven new tests in `flow-deviceless.test.ts` do bite it — deleting the recursion fails "answers true for a scope inside a when, nested three deep" and nothing else. What stays unreachable is the *run*, not the function: `flowRequiresDevice` answers yes for any flow holding a block, so this question is never asked with one present. The docblock says that rather than borrowing the other function's "no-op" wording, since the two are unreachable for different reasons.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: You are right, and this reverses what I argued last round. Fixed in ed09e3df6659.

My earlier reason for leaving the optional spelling — that it "contradicts `parseBlockSteps`' own invariant that it returns a non-empty `FlowStep[]`" — was a runtime invariant answering a compile-time question. It cannot make a guard fire. And your second point is the one that settles it: there was no backstop on the other side either, since `FlowStep[] | undefined` is already `blockSteps`' declared return type, so the kind reads as a leaf at every call site with nothing objecting anywhere.

The derivation is now `"steps" extends keyof S` distributed over the union, so it reaches any spelling, and `blockSteps`' `.steps` read carries a `satisfies FlowStep[]` that rejects an optional or non-`FlowStep[]` one on the registered side. Measured with the repo's tsc, unregistered `repeat` in each spelling:

| `steps` spelling | before | after |
|---|---|---|
| `steps: FlowStep[]` | `TS2322 … not assignable to type '"repeat"'` | same |
| `steps?: FlowStep[]` | silent | `TS2322 … not assignable to type '"repeat"'` |
| `steps: readonly FlowStep[]` | silent | `TS2322 … not assignable to type '"repeat"'` |

`steps: string[]` is caught too. The two existing constraints still fire unchanged — `["when", "tap"]` gives `TS2339: Property 'steps' does not exist`, `["when", "wen"]` gives `TS2820 … Did you mean '"when"'?`.

Writing it up surfaced that the registry's own docblock had the credit wrong in the other direction: it said `Extract` rejects a kind with no *required* `steps`, when `Extract` only catches a missing one and the `satisfies` catches the rest. Both are named accurately now.

packages/tool-server/src/tools/flows/flow-run.ts
j-piasecki: Both halves confirmed and fixed in f91c4d3d0c42.

The attribution was wrong exactly as you describe: `execSteps` gates on `isBlockStep` before the `execLeafStep` call, so a registered kind never reaches the leaf switch and the "unsupported step kind" outcome is that gate's doing. Traced what the arm actually prevents — `execWhenStep` is what pushes the `when` marker, so without a case for it the default arm returns having pushed nothing and the block leaves no report line at all, not even its own marker. The docblock now says that and credits the gate separately.

`stepTarget` I took as part of this rather than a note, since it is the same failure mode one function over. It now enumerates `echo`, `tool`, `launch` and `wait` explicitly and binds `never`, matching `stepRequiresDevice`. Behavior-identical — all four already returned `undefined` through the bare default. Verified the binding bites: an unregistered child-bearing kind now errors in `stepTarget` naming it, where before the only objectors were `stepRequiresDevice`, `toYamlStep` and `summarizeStep`, exactly your list.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Correct, fixed in 87e0586523ac. `assertBlockDepth` reads the list directly to build the depth-cap prefix, and that read arrived in aaccc2d43d2c without this docblock being revisited.

The claim now scopes to the CLASSIFYING read, which is the property actually worth stating — it is what keeps parse time and run time from answering the block question differently — and names `assertBlockDepth` inline so the other read is findable without a grep. Re-enumerated the reads while making the change: the definition, the `BlockDirectiveKind` type derivation, `isBlockDirectiveKey`, and `assertBlockDepth`. No other docblock in the file makes an "only read" claim.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Confirmed and fixed in 44351446cb0a. Re-enumerated against the merge base rather than taking the count on faith, and got your six, all in flow-run.ts: 880 `walkSteps`, 1738/1755/1770/1824 the `reportBlockSkipped` calls, and 1780 the dispatch into `execWhenStep`.

"the five in the runner" was the misleading half — the dispatch is in the runner too, so the locator read as a census of the file when it was a census of one of the two questions being asked. The paragraph now names both roles, five for the children and one for the dispatch, and the follow-on says all six, since a second directive would have had to remember the dispatch as well. It is `isBlockStep`/`execBlockStep` now, which 405c9e187344 made a compile error to miss.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Both wrong in the description, not in the code.

The message: the shipped string is `` `when:` blocks nest deeper than 20 levels ``, derived from `BLOCK_DIRECTIVE_KEYS`, and you are right that the delta from base is `when blocks` to `` `when:` blocks ``. That is 0a839647c0bd, which landed in response to your earlier note that "block directive" appears in no author-facing doc. The description still carried the pre-fix wording *and* the argument for it, so it reads as claiming the opposite of what ships. The argument itself survives in a narrower form: the message is not hardcoded to `when`, it names whatever is registered, so an author who reaches the cap through a different directive still is not misdirected — it just happens to name only `when` today because that is the whole registry.

The count: also wrong, and stale rather than merely imprecise. The flow suite is 1072 on this branch now (the nine commits in this round add six `flowScopesDevice` tests to the 1066 you measured). I have not re-measured 1049 at the merge base myself, so I am quoting your number for the base rather than asserting it.

Third one you did not flag, same staleness: the opening paragraph says "four runner sites", which is the count 44351446cb0a just corrected to six in the docblock.

All three are description-only, so no commit — I will rewrite the body rather than leave it describing a different change.

packages/argent-cli/src/flow.ts
j-piasecki: Confirmed and fixed in 0cb38b116054. `StepReport.depth`'s own declaration was the site 8e9e8d08aa3a missed, and it is the one the other two mirror, so the source of truth was the copy still describing the wrong set. It now says "nesting step" like the clients.

Audited the term rather than just fixing the hit you named: 20 other uses of "block directive" remain across the tool-server, and every one is the precise sense — anchored to `BLOCK_DIRECTIVE_KEYS` or `blockSteps`, none putting `run:` in the set. The reverse direction is clean too: "nesting step" now appears only where the set really is run-inclusive (`childScope`, `depthOf`, the three depth docblocks), and `childScope` is called on both the `when:` and `run:` paths, so the term matches its call sites.

packages/tool-server/test/flows/flow-utils.test.ts
j-piasecki: Correct, fixed in 3efe384f6dbb. The `Partial` is gone, so both tables are full `Record`s over the registry's key type and the comment states what the type enforces rather than what it hoped for. The `toBeDefined()` fallback and both `fixture!` assertions went with it, and the test destructures its fixture the way its neighbour already did.

Verified from both directions with a scratch second registration, as you did: before, tsc raised `TS2741: Property 'unless' is missing` for the sibling table only and nothing for this one; after, both tables raise it.

packages/tool-server/test/flows/flow-when.test.ts
j-piasecki: Fixed in 6b23264e9533, and you drew the line in the right place: the objection is to the restatement, not to the literal.

The message is now one `DEPTH_CAP_MESSAGE` constant in this file, used by all three sites. Still hand-written rather than derived from `BLOCK_DIRECTIVE_KEYS` — deriving it would move the expectation with the source and pin nothing, which is the argument the old comment made and it still holds. It just holds in one place now instead of three. The literal 20 stays a literal for the same reason, as you say.

The three sites did not assert the same amount of the message, and that survives: the cyclic-alias test still pins the whole string including the alias hint (the constant plus the hint composed in a template), the other two still pin the prefix half alone.

Mutation-checked: rewording the message fails all five message pins, `MAX_BLOCK_DEPTH` 19 and 21 each fail the side they should, and deleting `parseWhenStep`'s early assert still fails exactly the three precedence cases with both depth tests passing.

packages/tool-server/src/tools/flows/flow-utils.ts
j-piasecki: Reproduced — deleting it leaves all 1072 flow tests green. But I am not adding a test, and I want to be explicit that this is a claim about reachability rather than an appeal to "expected".

The property is unobservable, not merely unobserved. `parseBlockSteps` has exactly one caller, `parseWhenStep`, which calls `assertBlockDepth` on the same `raw` at the same `depth` as its first statement; `assertBlockDepth` throws purely as a function of `depth`; `badEntry` is unconditional and nothing on the parse path catches it. So if the inner assert could throw for some input, the outer one already threw on that same input. There is no YAML for which this is the assert that fires. `parseBlockSteps` is module-private too, so the only ways to build the scenario are exporting a private parser or adding a fake kind to the union, and either tests the fabrication rather than the guard.

What was wrong is that nothing said so. The docblock was candid that the second call is a no-op for a directive that made the early one, which reads as "untested" and sends the next reader looking for the missing test. 48059173484a adds the quantifier: while every registered directive makes the early call, no input can make this assert be the one that fires, so nothing pins it until one skips it.

Worth saying plainly rather than implying a near-term pin: #749 does not provide one. I checked, and `parseRepeatStep` opens with `assertBlockDepth(raw, depth)` too, so a second registered directive leaves this exactly as unpinnable as it is today. It becomes observable only when some directive deliberately trades the error precedence away, and nothing plans to. If that never happens, the honest reading is that this assert is a structural guarantee with no test rather than one waiting for coverage — which is a fair thing to hold against it, and the docblock now says enough for that call to be made deliberately.

PR #768, opened 2026-08-11
feat(flow): get a launch past the expo dev-client server chooser

hubgan:
## Problem

An expo dev build shows its `DEVELOPMENT SERVERS` chooser whenever it cold-starts and cannot silently reconnect to the server it opened last. Every `launch:` cold-starts by construction (terminate + relaunch), so any run against a dev build can land there.

It reads as intermittent and hits Android hardest, because the remembered address rotates between the emulator's `10.0.2.2` host alias, `localhost` (only reachable through `adb reverse`), and a LAN address that goes stale on its own. iOS keeps a stable `localhost:8081`.

The run does not fail cleanly. Every selector resolves against the chooser instead of the app, so the first directive fails on a screen that looks nothing like the flow, and recovering costs an agent several turns of describing and tapping.

<details>
<summary>The chooser, reproduced on an emulator (Bluesky dev build)</summary>

```
StaticText "DEVELOPMENT SERVERS"
View "http://10.0.2.2:8082 / Chevron" [clickable]
View "http://10.0.2.2:8081 / Chevron" [clickable]
View "Plus / New development server" [clickable]
StaticText "RECENTLY OPENED"
View "Bluesky / http://192.168.92.72:8081 / Chevron" [clickable]   <- dead address, live port
```

Two bundlers live at once, and a history entry carrying the run's own port on a host that stopped answering.
</details>

## Change

`launch:` handles it. After the tree-source gate it waits for the screen to become either the chooser or the app, opens the row for the run's own bundler, waits for the bundle, and reports what it did:

```
launch  pass  app opened behind the expo dev-client launcher — dismissed it via http://10.0.2.2:8081
```

Which bundler that is comes from `flow-execute`'s new **`metroPort`** (default 8081). The caller passes it, because only the caller knows which Metro it started.

Nothing is guessed:

- Only rows in the live `DEVELOPMENT SERVERS` section are eligible, never a `RECENTLY OPENED` entry.
- The match is on the whole `http://<host>:<port>` origin with a trailing-digit guard, so port `808` cannot open the row for `8081`.
- A chooser with no row for the port fails the launch naming the port it wanted, rather than running the flow against another project's bundle.

Two costs are deliberately kept off everything else:

- The wait for the chooser to draw runs **only while the screen has drawn no text** — the one state it can still emerge from.
- The whole path is gated on a `dumpsys package` probe for the `exp+` scheme, so an app that is not an expo dev build pays nothing. Measured: `com.android.settings` launches in 5s with no wait and no warning.

iOS is not probed (that needs the installed bundle's Info.plist rather than one `dumpsys` read), so the chooser is not auto-dismissed there. Given iOS's stable `localhost`, that is the rare case — but it is a gap, not a non-issue.

## Verification

Run end-to-end through `flow-execute` against a real emulator, not only unit tests:

```
ok: True
  echo    pass
  launch  pass    app opened behind the expo dev-client launcher — dismissed it via http://10.0.2.2:8081
  idle    pass
```

Corroborated by logcat, which shows the chooser drew at +5.0s and then handed off to the app:

```
Displayed xyz.blueskyweb.app/expo.modules.devlauncher.launcher.DevLauncherActivity for user 0: +5s043ms
Displayed xyz.blueskyweb.app/.MainActivity for user 0: +492ms
```

Row selection checked against the live chooser — by port, not position. That rendering listed 8081 first, the reverse of the unit fixture, so both orderings are covered on real data:

```
port 8081 -> "http://10.0.2.2:8081" @y=0.253
port 8082 -> "http://10.0.2.2:8082" @y=0.327
port 9999 -> no row (reports instead of guessing)
```

Also run: 12 new unit tests over the real captured tree; all 44 flow test files (1055 tests); `npm run build`; `npm run lint`; `npx prettier --check .`; `typecheck:tests` + `typecheck:scripts`; repo-hygiene greps; `npm run knip` against an unbuilt tree per CONTRIBUTING step 5 (exit 0, no new exports flagged).

SpiderShield tool-description score is unchanged from `main` at **9.09 / 10** (threshold 9.0); `flow-execute` sits at 6.5 on both sides, so the `has_scenario_trigger` gap there is pre-existing and left alone.

## Notes for review

- `APPEAR_TIMEOUT_MS` is 12s. It applies only to expo dev builds whose screen has drawn nothing, and exits the moment the chooser or the app appears. If dev builds routinely sit blank longer than that, this is the number to raise.
- Detection needs two marks on screen (the section heading **and** the "New development server" affordance), so an app screen that merely says "development servers" is never tapped at.
- Headings and row text are matched on each node's *own* text, not the hoisted `subtreeText`, or the scroll container's repeated text would drag the live/history boundary above the rows it must sit below.


latekvo (commented):
Comprehensive review of the expo dev-client chooser recovery. The implementation is thorough — own-text matching, history-boundary filtering, geometric address-box exclusion, and abort handling are all well-considered. Verified: 35 new unit tests pass (flow-dev-launcher), 94 composition tests pass, 101 CLI tests pass, `npm run build` clean. Findings below are edge-case observations, none blocking.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
```diff
@@ -0,0 +1,395 @@
...
+      ok: false,
+      reason:
+        `the expo dev-client launcher is showing and the tap that opens ${target.url} failed: ` +
+        `${errMsg(err)}`,
+    };
+  }
+
+  const deadline = Date.now() + EXIT_TIMEOUT_MS;
+  for (;;) {
+    if (signal?.aborted) return { handled: true, ok: true, url: target.url };
```
latekvo, line 523: When the exit-wait poll loop is cancelled mid-bundle (signal aborts during `sleepOrAbort` at line 391), the function returns `{ handled: true; ok: true; url: target.url }`. The tap already landed and the chooser was already dismissed, so reporting success is defensible — but the caller (`runLaunch` line 477) re-checks `signal?.aborted` and returns `ABORTED_OUTCOME` instead of the warning. So the step is reported as a skip rather than a pass-with-warning, even though the app was successfully launched behind a dismissed chooser.

The recovery completed (the app IS up, the chooser IS gone), but the step is reported as "run aborted" rather than carrying the dismissal warning. If a run is later resumed, the next pass won't know the chooser was dismissed. Not blocking — the abort takes priority — but the warning is lost on abort-after-success.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
```diff
@@ -0,0 +1,395 @@
...
+ * already has what the run needs, and left the other one unhandled: the launch
+ * passed, and every later step read the chooser. That second face is the common
+ * one (it is what the client shows whenever it has not found a running
+ * packager), and it is exactly where the "no reachable server on port N"
+ * message has to come from. The build's own header subtitle is listed too: it is
+ * on both faces, so it still identifies the launcher if either affordance is
+ * reworded.
+ */
+const SECTION_HEADING = "DEVELOPMENT SERVERS";
+const LAUNCHER_MARKS = ["Development Build", "New development server", "Fetch development servers"];
```
latekvo, line 59: `"Development Build"` is listed as a launcher mark, but it is the app's own branding text, not a structural property of the chooser. If a real app screen (not the chooser) happened to contain both "DEVELOPMENT SERVERS" and "Development Build" as text, `detectDevLauncher` would false-positive and the launch would try to dismiss a chooser that doesn't exist — tapping at whatever `pickDevServerRow` finds above the history boundary (or returning `handled: false` if no row matches, which is the safer fallback).

The risk is low because the heading itself is unusual wording for an app screen, but "Development Build" is the weakest of the three marks since it's the most likely to appear in an app's settings or about screen. The other two marks ("New development server", "Fetch development servers") are structural affordances of the launcher, not text an app would carry.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
```diff
@@ -0,0 +1,395 @@
...
+  port: number,
+  historyY: number
+): { node: DescribeNode; url: string } | null {
+  const live = candidateRows(flatten(root), historyY);
+  for (const host of candidateHosts(device)) {
+    const url = `http://${host}:${port}`;
+    // A trailing-digit guard, so port 808 cannot open the row for 8081. Plain
+    // substring matching is otherwise right: the row label wraps the URL in
+    // decorations (a chevron, the project name) that vary by client version.
+    const origin = new RegExp(`${url.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}(?![0-9])`, "i");
```
latekvo: The regex uses `(?![0-9])` as a trailing-digit guard so port 808 cannot match the row for 8081. One edge case: if a row label ever contains a URL with a path segment starting with a digit (e.g. `http://10.0.2.2:8081/2`), the negative lookahead would reject the match because `2` follows the origin. The real expo-dev-client rows seem to use the bare origin (`http://host:port`), so this is unlikely in practice, but the guard is narrower than "port boundary" — it rejects any digit after the origin, not just digits that extend the port number.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
```diff
@@ -0,0 +1,395 @@
...
+  );
+}
+
+/**
+ * How long a blank, still-starting screen is given to become something. Only a
+ * launch that has drawn nothing at all waits, and it stops the moment the
+ * screen draws — so this bounds the pathological case (an app whose first
+ * screen is genuinely wordless) rather than describing the normal one.
+ */
+const APPEAR_TIMEOUT_MS = 12_000;
```
latekvo, line 385: 12 seconds is a long time for a blank screen to wait, and the PR body calls it out as the number to raise. One concern: this wait runs on EVERY launch of an expo dev build, not just cold starts. A flow with multiple `launch:` steps (e.g. a mid-flow relaunch) pays this wait on each one if the app's first screen is genuinely wordless — say, an app that starts on a video player or a camera preview. The `hasDrawnContent` check (2+ text nodes) would read the screen as blank even though the app IS up.

The `hasDrawnContent` threshold of 2 text nodes is the right call for distinguishing a splash from the chooser, but an app whose first screen has exactly one visible text element would also wait the full 12s. The test at line 297-306 covers a branded splash with one label, but not a real app screen that happens to have only one text element alongside non-text content.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
```diff
@@ -0,0 +1,395 @@
...
+
+  let root: DescribeNode | null;
+  let launcher: { historyY: number } | null;
+  const appearBy = Date.now() + APPEAR_TIMEOUT_MS;
+  for (;;) {
+    root = await readTree();
+    launcher = root ? detectDevLauncher(root) : null;
+    // Settled on something: the chooser to dismiss, or a drawn screen that
+    // isn't one. Either way there is nothing left to wait for.
+    if (launcher || (root && hasDrawnContent(root))) break;
```
latekvo, line 445: The appear-wait loop breaks on timeout (`Date.now() >= appearBy`) with `root` possibly a splash or null. The subsequent `if (!root || !launcher) return { handled: false }` returns false, and `runLaunch` falls through to `{ ok: true }` — a pass with no warning. A dev build that takes longer than 12s to draw anything at all (splash, chooser, or app) is reported as a successful launch with no indication the appear-wait timed out.

This is probably the right behavior (the tree source gate already verified readiness), but it means a genuinely stuck dev build (blank screen, no chooser, no app) is indistinguishable from a successful direct launch. The next step's selector failure would be the first sign something is wrong, rather than a launch warning.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -112,6 +113,14 @@ const zodSchema = z
...
       ),
+    metroPort: z.coerce
+      .number()
+      .int()
+      .positive()
+      .optional()
+      .describe(
+        `ANDROID ONLY. Port of the Metro bundler this run's app must load from (default ${DEFAULT_METRO_PORT}). It applies only when a \`launch\` step lands on the expo-dev-client "DEVELOPMENT SERVERS" chooser, which is probed for on Android alone (an iOS dev build reaches Metro at a stable localhost, so it rarely shows the chooser; there this parameter does nothing). The runner then opens the listed server on this port and waits for the chooser to go away. Set this to the port of the Metro YOU started when it is not the default, or when more than one bundler is up: the runner never guesses which bundler the flow wants, so a chooser listing no live server on this port fails the launch step outright rather than opening another one.`
```
latekvo: The `metroPort` zod description is a single ~450-character string. The SpiderShield tool-description score mentioned in the PR body (6.5 for `flow-execute`) suggests this description is already long. The description now includes Android-only caveats, the chooser explanation, and the default port inline — all useful, but worth confirming the published tool description is still readable for an LLM consumer that has to parse it as one field.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -2204,6 +2252,21 @@ async function execLeafStep(
...
+      // same app — so the bundler this run targets is the one it must target
+      // too. A `run:` fragment inherits `metroPort` by construction (it shares
+      // one ExecState), and the two composition forms must not disagree on which
+      // Metro a dev-client launch opens. Only a non-default port is passed on:
+      // the default is what the inner run would resolve anyway, so injecting it
+      // would only add noise to every nested step's reported args. An explicit
+      // value on the step still wins — there the author named a bundler for it.
+      if (
+        step.name === FLOW_EXECUTE_TOOL_ID &&
+        args.metroPort === undefined &&
```
latekvo, line 2313: The nested `flow-execute` metroPort injection only fires when `state.metroPort !== DEFAULT_METRO_PORT`. This means a nested flow-execute that explicitly sets `metroPort: 8081` (the default) in its recorded args would have that value preserved — which is correct, since `args.metroPort === undefined` is the guard. But if a recording explicitly recorded `metroPort: 8081`, and the outer run also uses 8081, the inner step's args would carry `metroPort: 8081` while a `run:` fragment would not — the two composition forms would report different args for the same logical step. Not a correctness issue (the inner run resolves 8081 either way), but the reported args would differ.

latekvo (commented):
Comprehensive review of the expo dev-client chooser recovery. The implementation is thorough — own-text matching, history-boundary filtering, geometric address-box exclusion, and abort handling are all well-considered. Verified: 35 new unit tests pass (flow-dev-launcher), 94 composition tests pass, 101 CLI tests pass, `npm run build` clean. Findings below are edge-case observations, none blocking.

latekvo:
Review above completed by GLM-5.2 on xhigh thinking. 
Environment setup mirrored into Hermes agent from my daily driver harness.
Runtime = 46 minutes. For reference Claude Opus 5.0 average = ~45-55 min per review.
I believe the agent displayed satisfactory level of scrutiny.

<h3>Personal opinion:</h3>

When i first ran review with GLM-5.2 on late-stage PR it did AWFULLY, really acting like a 2023 model. It turns out that an overwhelming majority of perceived models intelligence is steered by learned memories and user-created system prompts. After migrating both of these, GLM-5.2 turned from useless to Opus level, which makes me think i was too harsh in the past towards models like gemini, GPT and deepseek, perhaps given the same config as Claude they perform on a similar level. It really is fascinating, I wonder how much further this can be pushed and how much of the potential remains unused.

latekvo (commented):
Comprehensive review of the expo dev-client chooser recovery. The implementation is thorough — own-text matching, history-boundary filtering, geometric address-box exclusion, and abort handling are all well-considered. Verified: 35 new unit tests pass (flow-dev-launcher), 94 composition tests pass, 101 CLI tests pass, `npm run build` clean. Findings below are edge-case observations, none blocking.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
hubgan: The abort takes priority deliberately, and the value on that arm is never observable by anyone. `runLaunch` re-checks the signal on the statement immediately after the call — `flow-run.ts:477` invokes `dismissDevLauncher`, `:478` is `if (signal?.aborted) return ABORTED_OUTCOME;`, with no `await` between them and `env.signal` being the same `AbortSignal` object the module destructures. A signal never un-aborts, so every abort arm of `dismissDevLauncher` (including both `{ handled: true, ok: true }` ones) resolves to `ABORTED_OUTCOME`. That is the intent recorded in the comment at `flow-dev-launcher.ts:414-417`, and `flow-dev-launcher.ts` has exactly one caller in `src`.

The resume half does not apply: the runner has no partial-run resume. `resumeFrom|--resume|startAtStep|fromStep|skipTo` matches nothing under `packages/*/src` — replay always starts at step 1, so no later pass can consume a per-step warning from an aborted one.

The test at `test/flows/flow-dev-launcher.test.ts:842` pins the outcome, and its comment states the same reasoning. Left as is.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
hubgan: Fixed in 20b1ae5615bb5945aa59340fe39afd9b60e79145 — the comment was what was wrong, not the mark.

"Development Build" is not the app's branding text. `expo-dev-launcher` renders that literal itself, in `AppHeader.kt`: `NewText("Development Build")` sits as a sibling of `NewText(appName)`, so the app's branding is the line above and the subtitle is the launcher's own fixed text. It lives in `android/src/debug/` — the debug-variant source set the `dumpsys` dev-build probe already gates on — and it is unchanged between `expo-dev-launcher` 6.0.21 and 57.0.11. `HomeScreen.kt` renders `AppHeader` above the `runningPackagers.isNotEmpty()` branch that swaps the two faces, which is why the mark is on both.

It is also load-bearing rather than only a hedge against a rewording: `FetchDevelopmentServersButton.kt` swaps its label to `"Searching for development servers..."` while a fetch is in flight, so on the no-servers face during a refresh the subtitle is the only mark left.

The wording that produced the misreading was "the build's own header subtitle", which reads as the app's. The comment now names the launcher as the author and records both facts above.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
hubgan: The lookahead rejects only a digit that immediately follows the port digits, which is the port boundary and nothing wider — a path segment is not rejected, because the character after the port is `/`, not the digit. Checked directly against the pattern:

```
"http://10.0.2.2:8081/2"              -> 8081: "http://10.0.2.2:8081"   | 808: null
"http://localhost:8081/index.bundle"  -> 8081: "http://localhost:8081"
"http://localhost:80812"              -> 8081: null
"http://192.168.1.5:18081"            -> 8081: null
```

So the example in the finding still matches. Both boundary directions are already pinned by tests — `test/flows/flow-dev-launcher.test.ts:457` (a short port must not open a longer one) and `:485` (a longer port must not answer for a shorter one, the case a free host pattern could break). No change.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
hubgan: Two things narrow this further than the finding assumes.

Every `launch:` is terminate + relaunch, so there is no warm launch to contrast with a cold one — each one genuinely re-enters the state the wait exists for, and on an Android dev build the chooser is what it meets (measured at +5.0s in the PR body). The loop ends at the first read that is the chooser or that has drawn.

The threshold is also weaker than "2 text elements". On Android `labelOf` takes `content-desc` first and falls back to `text`, so any node carrying a content description counts — a labelled icon clears the bar. On the three captured dumps in the tests the counts are 15, 15 and 16 against a threshold of 2. To pay the full 12s a screen has to expose at most one labelled node for 12 continuous seconds, on an Android expo dev build that also skipped its own chooser.

The exit wait is unaffected either way: it polls on `detectDevLauncher`, not on `hasDrawnContent`, so a camera or video first screen costs nothing there.

Lowering the timeout re-opens the miss it exists for, and a miss costs the whole run rather than 12s — every later selector then resolves against the chooser. Left as is, and the PR notes it as the number to raise if dev builds routinely sit blank longer.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
hubgan: Considered, and left as is: the warning would not catch the case it looks like it catches, and it has a false-positive set that a passing-step warning cannot afford.

A chooser that draws but is not recognized (marks reworded) carries ~15 labelled nodes, so `hasDrawnContent` is true and the loop exits at the FIRST read — the deadline is never reached and no warning would fire. What does reach the timeout branch is: a chooser slower than 12s (a true positive), every tree read throwing for 12s, a splash outlasting 12s and then drawing the app, and a dev build that skipped the chooser onto a one-label screen. The last two are false positives, and `argent-qa-flows` requires an agent to resolve every passing-step warning before completion — so a false one turns a healthy run into an investigation.

The read-outage sub-case is already covered downstream rather than silently: `flow-tree.ts:31-33` states that transient failures are absorbed by the callers' retry loops and a persistent outage fails the step with the helper's reason, so the next directive errors naming the tree source rather than the app.

packages/tool-server/src/tools/flows/flow-run.ts
hubgan: Fixed in 833d4286752b367333aedb9861073aba40e379b2 — 731 characters down to 461.

Two of the four sentences were already said in full elsewhere: the tool description says the launch "opens the server listed on `metroPort`, waits for the chooser to go away", and the iOS rationale in the parenthetical is in the reliability reference and is what "ANDROID ONLY" has already told the caller two sentences earlier. What is left is what decides the value — the port, the default, where it applies, when to pass one, and the launch error a wrong port produces. Verified against the published schema: the tool-server serves the new text on `flow-execute`'s `metroPort` property, and a run passing `metroPort` still errors with the port it wanted.

On the score itself: parameter descriptions are not in the scanned payload at all. `scripts/extract-tools.mjs` emits `{ name, description }` only — no `inputSchema` — and the workflow's custom scoring excludes `has_param_docs` besides. `flow-execute`'s 6.5 was unaffected by this field either way.

packages/tool-server/src/tools/flows/flow-run.ts
hubgan: The two forms are not comparable on args in the first place, so there is nothing here that the guard makes diverge.

A `run:` fragment emits a `kind: "run"` marker report with no `args` field at all and then inlines the fragment's own steps; the raw `tool: flow-execute` form emits a single report carrying `tool`, `result` and `args`. That difference holds in every case, with or without `metroPort`. And in the scenario described the arg is written by the flow file itself — the guard is what keeps the injection from adding one — while behaviour is identical either way, since the inner run resolves 8081 from its own default.

Nothing consumes `StepReport.args` beyond `packages/argent-mcp/src/content.ts:334`, which passes it to `toMcpContent` purely for the `includeImageInContext === false` image-suppression check. Baselines key off `baselineKey`, no CLI renderer reads args, and no test asserts parity between composition forms. No change.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
```diff
@@ -0,0 +1,501 @@
...
+    if (Date.now() >= appearBy) break;
+    if (!(await sleepOrAbort(APPEAR_POLL_MS, signal))) return { handled: false };
+  }
+  if (!root || !launcher) return { handled: false };
+
+  const target = pickDevServerRow(root, port, launcher.historyY);
```
latekvo: [Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: The run's Metro row is picked in a single pass the moment the appear-wait breaks on the launcher, with no retry before a missing row becomes a launch error. The chooser has a face it shows while it is still discovering packagers - the no-servers layout made of the heading, the launcher's header subtitle and the address box, all full of visible text - and the appear-loop breaks as soon as `detectDevLauncher` returns non-null on that face, because it has drawn content. `pickDevServerRow` then runs once against that face, finds no live row for the port, and the launch errors `lists no live server on port N` even when Metro is actually up and a live row for it appears on the very next read. Verified against boundary-mocked fixtures: the first tree read returns the discovering face, the second returns the populated chooser, and `dismissDevLauncher` returns `{ handled: true, ok: false }` rather than ever opening the row. A launch whose first chooser-read lands inside that population window fails spuriously - on exactly the cold starts this module exists to recover - and the message blames the bundler for a state that resolves on its own a read later.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -2204,12 +2288,38 @@ async function execLeafStep(
...
       }
       try {
         const result = await invokeSubTool(registry, ctx, step.name, args);
+        // `reinstall-app` takes an arbitrary `appPath`, so a run can put a
+        // release build over a dev one between two `launch` steps — the one
+        // thing that moves what `devBuilds` remembers, and in that direction a
+        // remembered `false` would skip the chooser recovery on a build that
+        // shows the chooser. Dropped whole rather than by key: the step names a
+        // PATH, which says nothing about the app id the probe is keyed by.
+        if (step.name === REINSTALL_APP_TOOL_ID) state.devBuilds.clear();
```
latekvo: [Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: The dev-build probe cache is evicted only when a `reinstall-app` step runs as a DIRECT tool step of this run. The same reinstall nested inside a `tool: flow-execute` step executes against the inner run's own fresh `devBuilds` map, so the outer run's remembered answer is left untouched. A launch that probed a release build (remembered 'no'), followed by a nested `flow-execute` whose flow reinstalls a dev build of the same app, then another launch of that app, reuses the stale 'no' and skips the chooser recovery on a build that does show the chooser - the recovery the comment above says the run handles is switched off, and later selectors resolve against the chooser. Reproduced with a boundary-mocked run `[launch, tool:flow-execute (inner: reinstall-app), launch]`: the probe runs once, not twice, and the second launch carries no recovery warning, whereas the direct-step reinstall case re-probes. `run:` fragments are unaffected because they share this ExecState; the nested-`flow-execute` composition is the one form that diverges.

latekvo (commented):
[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: Ran the review moves (claims vs code, nearest twin, non-happy paths, inputs, reachability, lifetime) plus the absence pass (sibling, prose, symmetry, mutation) across the new chooser-recovery module, the launch-step/report changes, the nested-composition paths, the recorder-decline mechanism, the CLI plumbing and the skills docs. Sectioned full-suite runs green: flow-dev-launcher (54), flow-composition (98), flow-record-tap (25) at the PR head. Two LOW findings below, each confirmed with a concrete boundary-mocked repro (no Android device on this box, so the trees are fixtures captured from a real chooser).

packages/skills/skills/argent-create-flow/references/live-authoring.md
```diff
@@ -233,10 +234,12 @@ Resolve every hit and confirm:
...
+Manual rescue invalidates the pass. An `errored` step was never evaluated: an `idle` wait whose tree source could not be read, a step that threw, an unresolvable `run:` target, or a `launch:` that did not start the app. Read the reason — most name the environment, but a failed `launch:` is a verdict about the app. One class of launch reason is the exception: every reason that names the expo dev-client launcher names the bundler, not the app. The chooser listed no live server on the run's `metroPort` (`--metro-port` from the CLI); the tap that opens that server failed; the chooser was still showing 60s later, so the bundler at that address never served this app; or the screen could not be read for 60s afterwards. Repair the bundler or the port and rerun. Unconfirmed focus is not in this class at all: the replay focus poll has no failure return, so a `type:` step whose focus was never confirmed is scored a **pass**, and only the value check after typing catches it.
```
latekvo: [Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: The claim that every launch-error reason naming the expo dev-client launcher is about the bundler does not hold for two of the four reasons enumerated here. `flow-dev-launcher.ts` emits a tap-failure reason ("the expo dev-client launcher is showing and the tap that opens <url> failed: <err>") that names the gesture, and a tree-read reason ("the screen could not be read for 60s afterwards (...) so whether the chooser left is unknown") that the module's own comment says is produced exactly when the reads, not the bundler, are what is down. The blanket "Repair the bundler or the port and rerun" guidance thus points an agent at the wrong subsystem for the unreadable-screen case: a launch whose chooser never left because the tree source was unreachable would be re-run against Metro to no effect. The same overclaim appears in argent-qa-flows/SKILL.md ("every reason that names the expo dev-client launcher... Start that Metro, or correct the port") and in reliability-and-recovery.md ("unless its reason names the expo dev-client chooser, which names the bundler").

latekvo (commented):
[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731]: Follow-up pass over the claims-vs-code move for the skill docs; the device suites at PR head stay green. One finding below.

packages/skills/skills/argent-create-flow/references/live-authoring.md
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: The over-claim this thread names in live-authoring.md:241 also survives at head e8d5b961 in two sibling locations: reliability-and-recovery.md:123 ("unless its reason names the expo dev-client chooser, which names the bundler") and argent-qa-flows/SKILL.md:103 ("every reason that names the expo dev-client launcher" enumerated as naming the environment). In both, two of the four enumerated reasons do not name the bundler: the tap-failure reason (flow-dev-launcher.ts:460 names the gesture and its error) and the unreadable-screen reason (flow-dev-launcher.ts:490 names the tree source, not the bundler).

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -2204,12 +2288,38 @@ async function execLeafStep(
...
+      // Android only, which is all the parameter is for: the nested run binds
+      // this same device, and off Android its launch never reads the port, so
+      // forwarding it would only put an inert argument in the step's args.
+      if (
+        step.name === FLOW_EXECUTE_TOOL_ID &&
+        device?.platform === "android" &&
+        args.metroPort === undefined &&
+        state.metroPort !== DEFAULT_METRO_PORT
+      ) {
+        args.metroPort = state.metroPort;
```
latekvo, line 2316: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: The comment on this block says a nested `flow-execute` is "this same run one level down — same device, same app", and forwards `metroPort` to it on that basis — but the sibling per-run state the launch recovery actually consumes, `devBuilds`, is not forwarded with it. Every `flow-execute` invocation initializes a fresh `new Map()` for it (flow-run.ts:1145), the nested ones included, so an inner run whose flow launches the same dev build the outer run already probed re-runs the `dumpsys package` probe for the same (device, app) key the outer answer already covers, once per nesting level, each with the full 10s probe budget when the device is slow. Correctness is unaffected — the answer is re-derived identically — which makes this the exact duplication the `devBuilds` cache exists to prevent, on precisely the composition path this block declares equivalent to the outer run.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: Full review pass at head e8d5b961. Read the complete diff (16 files, +2735/−84), all 17 existing threads, and confirmed every cited "Fixed in" hash is an ancestor of the head. Dispatched five review agents across the moves — claims vs code, nearest twin + inputs, non-happy paths + reachability + abort handling, the absence pass (sibling / prose / symmetry / mutation) + lifetime, and composition / recorder-decline — working in a throwaway worktree at the head commit. Re-ran the suites at head: flow-dev-launcher 54/54, flow-composition 98/98, flow-record-tap 25/25; a scratch vitest repro additionally confirmed that the no-history chooser face (historyY fallback 1) reports "no live server" and picks a live row when one is present. One new finding below, plus a thread extension on the bundler-taxonomy docs claim. The three open threads (single-pass row pick at flow-dev-launcher.ts:432, nested-reinstall cache eviction at flow-run.ts:2322, bundler-taxonomy docs at live-authoring.md:241) are all confirmed present at head.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
latekvo: [Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Fixed in d19e3a540e0e83123030d25f2d4492331f81ea6f.

The single pick no longer decides against the discovering face. When the pick misses AND the screen lists no live row at all, the launch now treats the empty list as discovery still running and retries — a tree read every 500ms, up to 12s — ending on any of: a row for `port` (opens it), any listed row (discovery has finished, so a wrong-port miss fails immediately with the same message, spending no second window), the chooser leaving entirely (`handled: false` — the client connected itself, nothing here dismisses anything), or the deadline. A tree read that throws mid-window is transient and only spends it, matching the appear-wait's treatment.

Reproduced at the prior head with scripted reads [no-servers face, populated chooser, app] and port 8091: `{ handled: true, ok: false }` with "lists no live server on port 8091", never tapping — the exact spurious failure described. At the fix head the same script returns `{ handled: true, ok: true, url: "http://192.168.0.94:8091" }` with one tap. Two companions pin the edges: a populated list missing the port still fails after a single read (no second window on a finished discovery), and a chooser that leaves while discovering leaves the launch alone.

packages/tool-server/src/tools/flows/flow-run.ts
latekvo: [Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Fixed in d19e3a540e0e83123030d25f2d4492331f81ea6f.

`state.devBuilds.clear()` now also fires after every nested orchestrator step — `tool: flow-execute` and `run-sequence` — not only a direct `reinstall-app`. The nested run keeps its own map, and that map dies with it, so eviction in the OUTER run is the only thing that can keep its remembered answer honest; the composition step is the boundary at which an install can have happened unseen. Eviction is unconditional per composition step rather than install-aware — nothing crosses the registry boundary that says an install happened — so the cost of not knowing is one extra probe on a later launch, which is the cheap direction to be wrong in.

Reproduced at the prior head with run `[launch, tool: flow-execute (inner: reinstall-app), launch]`: one probe for two launches, exactly as reported — the second launch reused the stale answer. At the fix head, two probes, and both launches recover. The direct-step case keeps its existing coverage; a `run:` fragment remains unaffected as before (it shares this ExecState).

packages/skills/skills/argent-create-flow/references/live-authoring.md
latekvo: [Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Fixed in d19e3a540e0e83123030d25f2d4492331f81ea6f, including both sibling locations from the follow-up.

The blanket claim is gone. All three sites now split the recovery's four reasons the way the code actually emits them:

- **live-authoring.md** (manual-rescue paragraph): two bundler verdicts — no live server on `metroPort`, or the chooser still showing 60s later because the bundler never served this app — repaired by fixing Metro or the port; and two reasons that name other subsystems — a tap that failed (`the tap that opens <url> failed: …`) points at the device and is usually cleared by a rerun alone, while a screen unreadable for 60s afterwards says the tree source went down mid-bundle, so check the device connection, not Metro.
- **reliability-and-recovery.md** (environment-error row): the exception is narrowed to "names the expo dev-client chooser's bundler (no live server on `metroPort`, or still showing after 60s)"; the tap-failed and unreadable-screen reasons are environment like any other.
- **argent-qa-flows/SKILL.md** step 4: same split — the environment class covers the two bundler verdicts ("Start that Metro, or correct the port"), with one line each for where the other two point.

Verified against the reason strings emitted at flow-dev-launcher.ts (tap-failure names the gesture and its error; the unreadable-screen reason names the reads, per its own comment): no remaining prose claims every launcher-naming reason names the bundler. `rg -n "every reason that names" packages/skills` comes back empty.

packages/tool-server/src/tools/flows/flow-run.ts
latekvo: [Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Dismissed as a cost question rather than a defect — with the comment corrected so it no longer claims state parity.

Correctness is unaffected, as the finding itself notes: the inner probe re-derives an identical answer from the same device and app. The worst case is one extra `dumpsys package` read per nesting level where an inner flow launches a dev build — normally milliseconds, bounded by the 10s budget on a pathologically slow device. Sharing the map would mean threading mutable run state through the registry's validated invocation boundary: `invokeSubTool` dispatches as plain args into a zod-parsed `invokeTool`, so the map can only cross as either a public field on `flow-execute`'s schema (visible in every tool listing, subject to record/replay parity, and dead weight for every non-flow caller) or new plumbing in the registry package's invoke options. Both cost more than the duplicate read they save.

What was actionable in the thread is taken: the block's comment now states explicitly that `metroPort` is the only forwarded state and why (`d19e3a54`). The stale-answer hazard that makes cache hygiene matter at all is separately closed by the eviction added for the sibling thread on nested-reinstall eviction.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Full independent pass at head e8d5b961 over the complete diff (16 files, +2735/−84), reading every existing thread first so nothing below duplicates them. Dispatched agents across claims-vs-code (every PR-body claim checked against the implementation, including the probe target against upstream expo-dev-launcher manifests), inputs (port boundaries both directions, zod coercion edges, frame coordinate space, regex construction, arg mutation), the recorder-decline mechanism end-to-end against its nearest twins, the composition matrix (when:/run:/nested flow-execute × metroPort/devBuilds), non-happy paths (all 13 terminal states of the dismissal mapped to their step verdicts), reachability both ways, lifetime (devBuilds writers/clearers, abort points, leaked waits), plus the absence pass (sibling warning consumers in CLI/MCP/batch/--json, docs symmetry across all five skill files) and a mutation pass over the new branches and constants. Re-ran at head: flow-dev-launcher 54/54, flow-record-tap + flow-composition green, CLI flow suites green; scratch vitest repros were built and executed before anything below was accepted. Four low-severity findings, each with concrete evidence.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
```diff
@@ -0,0 +1,501 @@
...
+ * seconds after the launch step's read (see {@link APPEAR_TIMEOUT_MS}), so
+ * catching it means waiting; making an ordinary app wait out that window on
+ * every launch would be a tax it can never benefit from.
+ *
+ * A dev build is identified by the launcher itself: `expo-dev-launcher` declares
+ * its activities in a DEBUG-variant manifest, so the installed package mentions
+ * them only when the build can actually show a chooser. On Android that is one
+ * `dumpsys package` read.
+ *
+ * Deliberately NOT the `exp+<slug>` scheme the chooser's own rows open: the
```
latekvo, line 303: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: The pull-request description says the whole path is "gated on a `dumpsys package` probe for the `exp+` scheme", but the implementation deliberately probes the launcher's own debug-manifest marks instead (`expo.modules.devlauncher` / `Scheme: "expo-dev-launcher"`, line 352) precisely because the `exp+<slug>` scheme is written into the main manifest by the expo-dev-client config plugin and survives into release builds — verified against upstream expo-dev-launcher's `android/src/{main,debug}/AndroidManifest.xml`. The code is right; the description mischaracterizes it. Anyone consuming the description rather than this comment could conclude release builds pay the chooser wait, or "simplify" back to a scheme probe and reintroduce exactly the release-build tax this module documents having removed.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
```diff
@@ -0,0 +1,501 @@
...
+ * almost always lands on the splash and concludes, wrongly, that there is no
+ * chooser. What separates "the app is up, nothing to do here" from "still
+ * starting, the chooser may yet appear" is text: a splash is a logo on a blank
+ * ground, and both the chooser and a real first screen are full of words.
+ *
+ * Gating the wait on this keeps the cost off an app that has already rendered:
+ * it pays one read, and only a screen that has drawn nothing waits — the one
+ * state the chooser can still emerge from.
+ */
+export function hasDrawnContent(root: DescribeNode): boolean {
```
latekvo, line 371: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: The ≥ 2-visible-text-nodes threshold means a cold-start splash that renders two or more labels of its own (app name plus a version or status string suffices) reads as "the app has drawn", so the appear-wait exits before the chooser draws, `dismissDevLauncher` returns `{ handled: false }`, and the chooser appears seconds later onto a screen every subsequent step resolves selectors against — the exact failure this recovery exists to prevent — while the launch reports a plain pass with no hint the launcher was met. The comment's premise ("a splash is a logo on a blank ground") does not hold for apps whose splash carries more than one label; reproduced with a crafted tree where a two-label splash closes the window (`hasDrawnContent` returns true) while `detectDevLauncher` stays null.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -86,50 +88,73 @@ async function captureTapSelector(
...
-      return {
-        warning: `selector ${describeSelector(selector)} resolves to a different element on this screen; kept coordinates (brittle)`,
-      };
-    }
-    return { selector, warning: fallbackSourceWarning(source, device.platform) };
+    // Off the same read, and only where a replay would undo it: the launch's
+    // chooser recovery is Android-only, so on every other platform the chooser
+    // is still on screen at replay and this tap is the step that gets past it.
+    const devServerRow =
+      device.platform === "android" ? (devServerRowAt(tree, point) ?? undefined) : undefined;
```
latekvo, line 99: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: This gate keys on the RECORDING device's platform, but the decline it feeds only asks whether the previous step is a `launch:` — it cannot know the replay platform. For a cross-platform launch map recorded once on an Android emulator, the chooser-row tap is silently declined here; replayed on iOS that run has neither the tap nor any recovery (the recovery is Android-only), so the flow meets the iOS chooser with nothing to dismiss it. reliability-and-recovery.md:91 states the keep-rule purely in replay terms ("It keeps the tap … on every other platform, because there the tap is what gets past the chooser"), which does not hold whenever recording platform differs from replay platform.

packages/skills/skills/argent-create-flow/references/reliability-and-recovery.md
```diff
@@ -78,6 +79,19 @@ While the required source is down, selector failures and raw-point capture are v
...
+- When Metro is not on port 8081, set `metroPort`.
+- When more than one bundler is available, set `metroPort`.
+- Do not write a `when:` block to tap the chooser on Android. The launch dismisses the chooser before your step can run, so the step meets a different screen. The order of the rows also changes between launches. Where no launch does the recovery — iOS, or a `tool: restart-app` step — a `when:` block is the workaround.
+- Do not keep a recorded tap on a chooser row either. The recorder leaves that tap out of the file and says so, but only where the launch performs it at replay: on Android, and only when the step before it is the recorded `launch:`. It keeps the tap after a `tool: restart-app` step, after a `run:` fragment, and on every other platform, because there the tap is what gets past the chooser. Delete such a step by hand only in a flow recorded before this, or where a launch does the recovery for it.
+- If the chooser has no live row for the port, the launch fails. Start Metro on that port, or correct `metroPort`. This failure is about the bundler, not about the app.
```
latekvo, line 92: \[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: This section is the anchor every other document links to as the canonical reference for launcher failures (flow-yaml.md, live-authoring.md, both SKILL.md files), yet it describes only one of the four failure reasons the launcher emits — the missing live row. The tap-failure reason ("the tap that opens <url> failed"), the still-showing-after-60s reason, and the unreadable-for-60s reason exist only in live-authoring.md's manual-rescue paragraph and argent-qa-flows item 4. An author hitting any of those three and following the link lands on a section whose only failure bullet is about a different cause.

latekvo:
[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Review passes over the branch at head 7e550b42, following every reviewer thread being resolved.

**Fixes landed** (d19e3a54, b30783c1, 7e550b42), each reproduced before and after:

1. **Row pick retried during discovery** (was: single pick against the chooser's discovering face could fail a launch Metro was about to satisfy). The window now runs to its 12s deadline regardless of what the list already shows — mDNS answers arrive incrementally, so an early exit on the first foreign row would reintroduce the false verdict in a narrower race. Verified by mutation testing: reverting to early-exit fails the suite exactly at the new read-count assertion.
2. **Probe cache evicted after nested compositions and in a `finally`** — a reinstall inside `tool: flow-execute` (or a midway-failing direct reinstall) no longer leaves a stale answer; eviction now answers "the ground may have moved", not "the step passed".
3. **Docs taxonomy scoped** in live-authoring.md, reliability-and-recovery.md and argent-qa-flows/SKILL.md: two launcher reasons are bundler verdicts (repair Metro/port), two name other subsystems (failed tap → device/rerun; unreadable screen → tree source).

**Remaining LOW findings, disclosed not fixed** (each judged not worth its cure):

- Deadline expiry after reads that ALL throw reports "lists no live server" from evidence up to 12s old (diagnosis quality only; the exit-wait handles the symmetric case more carefully).
- Host-blind port match can open another developer's bundle on a shared LAN where coworkers run Metro on the same port.
- `historyY ?? 1` lets a scrolled-past history heading make remembered rows live during manual recording (`devServerRowAt`); unreachable on cold-start launches.
- A transient overlay that hides both chooser marks reads as "chooser left" (`handled: false`) for one frame.
- IPv6-literal and https row URLs can never match the origin pattern; current clients write resolved IPv4 over http.

Suites at head: flow-dev-launcher 60/60, all 45 flow files 1124+ green modulo two known load-flaky timing suites (flow-concurrent-recording, flow-idle-run) which fail identically at base e8d5b961 under this machine's load; build, lint, prettier clean.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
```diff
@@ -0,0 +1,545 @@
...
+  // launch would tap the middle of the list — an arbitrary point — instead of
+  // reporting that no live row exists. Nothing is lost by ignoring the hoist:
+  // it is additive, so the leaf that renders the URL still carries it itself.
+  // Tightest match still wins, since a row card whose own label spells the URL
+  // also wraps the leaf that renders it.
+  let best: { node: DescribeNode; url: string } | null = null;
+  for (const node of candidateRows(flatten(root), historyY)) {
+    const found = origin.exec(nodeText(node));
+    if (!found) continue;
+    if (best === null || area(node) < area(best.node)) best = { node, url: found[0] };
```
latekvo, line 206: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Row identity is port-only on whatever host the client wrote, so two live servers on the same port cannot be told apart — and Metro's default port makes that the ordinary shape, not an edge: any second machine advertising its own project's bundler on 8081 lands in the live section next to this run's row, and the pick between them comes down to card area alone.

Reproduced against this module's own harness: a tree listing a live `http://192.168.1.50:8081` row above a live `http://10.0.2.2:8081` row sends the tap to whichever card renders smaller, and `dismissDevLauncher` returns `{ handled: true, ok: true, url: "http://192.168.1.50:8081" }` once the chooser leaves. The launch then reports success and every later step resolves against the other project's bundle; the only trace is the warning URL, which nothing validates.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
```diff
@@ -0,0 +1,545 @@
...
+      if (Date.now() >= rowBy) break;
+      if (!(await sleepOrAbort(ROW_POLL_MS, signal))) return { handled: false };
+    }
+  }
+  if (!target) {
+    return {
+      handled: true,
+      ok: false,
+      reason:
+        `the expo dev-client launcher is showing and lists no live server on port ${port}. ` +
```
latekvo, line 482: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: When the initial pick misses and every tree read inside the retry window fails, this verdict still asserts present-tense screen state the window never observed: "the expo dev-client launcher **is showing** and lists no live server on port N. Start Metro on that port…"

Reproduced by scripting `fetchFlowTree` to throw after the first read: after 12s of unreadable screen the launch errors blaming the bundler, though the last confirmed sighting of the chooser predates the whole window — it may have left on its own, or the row may have appeared while nothing could be read. An agent following this reason restarts Metro instead of looking at the device or tree source. The exit-wait below already models exactly this distinction with its own "could not be read" reason (line 534), so the asymmetry sits inside one file.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -546,7 +577,48 @@ If a step was recorded by mistake, edit the .yaml to remove it — against a rem
...
+      // the fragment is one nothing at replay dismisses. A raw `restart-app`
+      // carrying an activity gets no recovery at all. After either, this tap is
+      // the step that gets past the chooser, and is kept.
+      // Decided by the append itself, against the flow it is about to extend:
+      // this asks what the PREVIOUS step is, and only there is that read of the
+      // file the append's own (see appendStepToFlow). Off the in-memory copy it
+      // would answer off a flow a hand edit — which this tool's own description
+      // invites — may already have outdated.
+      const declineDevServerRow =
+        captured?.devServerRow === undefined
```
latekvo, line 599: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: The decline fires only when the immediately previous step is a `launch`, so any append interleaved between the launch and the chooser tap defeats it silently. `flow-add-echo` exists precisely to insert steps mid-recording, and hand edits do the same in host mode.

Reproduced against `appendStepToFlow`: with steps `[launch]` the chooser-row tap is declined (`appended: false`); with steps `[launch, echo]` the identical tap is kept. At replay the launch recovery opens the row and the kept step then fires into the app the recovery just opened — the double-tap this mechanism exists to prevent. A selector derived from the chooser tree usually fails loudly there; when coordinates were kept instead it is an unconditionally green blind tap.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
```diff
@@ -0,0 +1,545 @@
...
+ *
+ * The host matches anything but a separator, so a row for 18081 cannot answer
+ * for 8081, and the trailing-digit guard stops port 808 opening the row for
+ * 8081. Plain substring matching is otherwise right: the row label wraps the URL
+ * in decorations (a chevron, the project name) that vary by client version.
+ * `"any"` asks only "is this a server row at all", which is what the recorder
+ * needs — it is identifying the row under a tap, not choosing one.
+ */
+function originPattern(port: number | "any"): RegExp {
+  return new RegExp(`http://[^\\s/:]+:${port === "any" ? "\\d+" : port}(?![0-9])`, "i");
```
latekvo, line 163: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: The host class `[^\s/:]+` cannot cross the colons of an IPv6 literal, so a row spelled `http://[::1]:8081` (or a link-local address with a zone) matches neither the port pattern here nor the `"any"` pattern the recorder uses — verified directly against both regexes. Impact: if a client ever writes a bracketed host into a live row, the launch spends its full retry window and then fails claiming no live server exists while the row is on screen.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -112,6 +117,15 @@ const zodSchema = z
       .describe(
         "Restrict auto-detection to this platform when several devices are booted. `chromium` does more than filter: with no `device` it SELECTS the self-boot branch for an e2e flow - the runner boots an Electron instance from the `launch` step's chromium value and tears it down after the run (a single-key `launch: { chromium: … }` map selects it on its own, without this parameter). When it selects that branch it never falls back to device auto-detection (a fragment, or an e2e launch map with no `chromium` key, still does), and the launch value must be a real Electron app path on the tool-server host: a bare-string `launch:` - what the recorder writes - holds an installed-app bundle id, so passing `chromium` for one fails the whole run with `Electron boot: path does not exist`. Edit the launch to `{ chromium: <app path> }` first."
       ),
+    metroPort: z.coerce
```
latekvo, line 120: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: zod v4's number coercion admits junk past what this schema intends: `metroPort: true` coerces to 1, and a single-element array (`[8081]`, `["8081"]`) coerces to 8081 — verified against the installed zod. A caller passing a boolean gets a launch hunting for rows on port 1 and failing with "lists no live server on port 1", rather than a schema error naming the bad argument.

latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Full independent pass at head 7e550b426ceb. Read the complete diff and all existing threads first; dispatched agents across claims-vs-code, inputs/boundaries, non-happy paths + abort interleavings, reachability both ways, lifetime/eviction, composition matrix, recorder-decline, docs absence-pass, and a convergence sweep over geometry, runLaunch integration, the new tests and the CLI/content plumbing. Re-ran the suites at head (flow-dev-launcher 54, flow-composition 98, flow-record-tap 25 — all green) and built scripted-tree reproductions for every finding below; each one was confirmed against actual behavior before posting. Findings below are new; three open threads from the earlier pass were reconfirmed at this head and answered on their own threads.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Still present as of 7e550b426ceb. Reconfirmed with a scripted-read repro at head: a splash rendering two visible text nodes of its own ("MyApp" / "Loading…") makes `hasDrawnContent` true on the first read, so the appear-wait exits immediately and `dismissDevLauncher` returns `{ handled: false }` — the chooser that draws seconds later is never dismissed and the run proceeds against it.

packages/tool-server/src/tools/flows/flow-dev-launcher.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Still present as of 7e550b426ceb. The pull-request description still says the path is "gated on a `dumpsys package` probe for the `exp+` scheme", while the code probes the launcher's own debug-manifest marks (`expo.modules.devlauncher` / `Scheme: "expo-dev-launcher"`, line 352) precisely because the scheme is not debug-variant-only.

packages/tool-server/src/tools/flows/flow-add-step.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Still present as of 7e550b426ceb. The Android gate at capture time (flow-add-step.ts:99) plus a decline that only asks whether the previous step is a `launch` are unchanged; an Android-recorded flow replayed on iOS still has neither the recorded chooser tap nor any recovery to open the server.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Full independent pass at head 7e550b426ceb. Read the complete diff (16 files, +2979/-84) and every existing review thread first, so nothing here duplicates them. Dispatched four agents across the review moves and the absence pass, each briefed with the full set of already-posted threads: (1) the tree-math and detection internals of flow-dev-launcher.ts, (2) the dismissDevLauncher / isExpoDevBuild control flow and its runLaunch + execLeafStep integration, (3) the recorder-decline path in flow-add-step.ts + appendStepToFlow, and (4) the metroPort / devBuilds plumbing across flow-run.ts, the CLI, content.ts and the five skills docs.

Re-derived the geometry the module rests on rather than trusting the comments: getDescribeTapPoint returns the frame centre, frameContains is inclusive on all four edges, nodeText reads a node's own label+value (never the hoisted subtreeText), and the address-box exclusion holds in both directions - a URL on a child leaf is excluded because the leaf's centre falls in the box, and a URL on the box's own label is excluded because a box contains its own centre. The `>= historyY` boundary drops a row only at or below the RECENTLY OPENED heading, which a live row never reaches. Re-walked all three DevLauncherOutcome states through to their StepReport status/warning, every abort exit (no aborted sleep falls through to a tap), and the isExpoDevBuild cache (error/timeout/abort answer false uncached; only a real read is remembered). The worst-case launch budget (probe 10s + appear 12s + row 12s + exit 60s) does not blow the MCP client cap because flow-execute is longRunning.

No new HIGH or MEDIUM. No new distinct, actionable LOW either:

- The one candidate that looked new - the published CLI/tool reference under packages/docs not carrying `--metro-port` / `metroPort` - is a false positive on this branch: packages/docs does not exist here (it was added to main later, in #814), so there is no reference page for this PR to update. The branch's own docs, the five packages/skills files, are updated.
- The transient-detection-miss variants a pass surfaced (the row-retry loop's `!nextFace` early exit at flow-dev-launcher.ts:466, and the exit-wait declaring success on a single `!detectDevLauncher` read) fold into the same "detectDevLauncher returns null therefore handled:false, silent pass" tradeoff already threaded and defended on the appear-loop findings; both require a defective/partial tree read of a chooser whose heading and marks are static, and the module treats such reads as transient by design. Not opening a fourth thread on the same acknowledged decision.

The disclosed LOW limitations already on their own threads - host-blind port match, the two-label-splash appear-close, the interleaved-append decline gap, IPv6 / https origins, zod coercion, the abort-after-success warning - remain present at this head and are left where they are.

Issue #775, opened 2026-08-11
native-devtools sockets: multi-server contention (attach-to-every-sim + latest-bind-wins + unlink-on-stop)


filip131311:
## Summary

When more than one argent tool-server runs on the same host (a second dev
instance, a stale/orphaned instance, or another agent session's server), the
native-devtools injection sockets become a three-way hazard:

1. **Attach-to-every-sim**: each tool-server's simulator watcher initializes a
   NativeDevtools service — and binds `/tmp/argent-nd-<UDID>.sock` — for
   **every** booted simulator, even devices that server will never drive. A
   session doing Android-only work through a dev tool-server knocked out an
   unrelated iOS session's devtools on a different simulator.
2. **Latest-bind-wins**: the blueprint unlinks any existing socket file before
   binding (`fs.unlinkSync(socketPath)` in `native-devtools.ts`, "remove stale
   socket file from a crashed previous run"). A *live* server's socket is
   indistinguishable from a crashed one, so every server start silently steals
   all future app connections; the losing server keeps its established
   connections but every relaunched app connects to the thief.
3. **Unlink-on-stop kills survivors**: when the last-binding server stops, its
   cleanup removes the socket file — the surviving servers' listeners are gone
   from the filesystem, so no app can connect to *anyone* until something
   rebinds.

## Observed failure signatures (all reproduced live, 2026-08-11)

- flow `launch` step: `could not connect to native devtools for <bundleId>…` —
  the app connected to another server's socket; the flow's server waits out its
  connect window in vain. Reproduced on demand: the failure windows correlate
  1:1 with another server instance's start timestamps (`22:07`, `22:10:57`,
  `22:15:51`, `22:16:55` across two independent sessions).
- `ViewInspector RPC timed out: <method>` mid-flow — when the steal happens
  between a flow's app relaunch and its first RPC.
- With three sessions (two scoped experiments + one unrelated Android A/B loop
  restarting its dev server per run), no session could keep a stable devtools
  connection on any simulator.

Also worth noting operationally: there is no in-band way to identify or
negotiate with the process holding the socket — `lsof` + process-tree forensics
was the only way to find the owner.

## Sketch of remedies (not prescriptive)

- Scope: only claim devtools sockets for devices a server is actually asked to
  drive (lazy claim on first tool call for that UDID), or honor an explicit
  device allowlist. (An `ARGENT_DEVTOOLS_ONLY_UDID` env filter prototype made
[...]

Issue #774, opened 2026-08-11
iOS simulator: keyboard shift state wedges mid-session — '@' types as '2', '+' as '='


filip131311:
## Summary

On the iOS simulator, `keyboard` types shifted characters via a shift-modifier HID keystroke. Part-way through a session the simulator's shift state wedges, and from then on **every shifted character lands unshifted** for the rest of that session:

- `@` types as `2`
- `+` types as `=`

So `filip+test@example.com` is entered as `filip=test2example.com`. Nothing errors — the tool reports success and the auto-screenshot shows the wrong text in the field.

## Environment

- argent 0.20.0
- iPhone 17 Pro simulator, iOS 26.4 (also seen on the same host with iOS 26.x generally)
- macOS host (Darwin 25.5.0, Apple silicon)
- App under test: Expensify NewDot (React Native), built from source; the affected field is the sign-in "Phone or email" input

## Steps to reproduce

1. Boot an iPhone simulator and launch any app with a plain text input.
2. Use `keyboard` repeatedly across a session to type strings containing `@` or `+` (e.g. drive a sign-in flow several times in a row, each time typing a unique email).
3. Early runs type the address correctly. After some number of runs, the shifted characters stop being shifted — `@` becomes `2`, `+` becomes `=` — and stay broken for the remainder of the session.

Rebooting the simulator clears it, but it recurs after further use.

## Why this looks like an argent bug, not a simulator quirk

In the same session, on the same simulator, typing the same strings:

- **maestro** (XCUITest `typeText`) — unaffected, types `@` correctly
- **agent-device** — unaffected, types `@` correctly
- **argent `keyboard`** — wedged

That isolates it to argent's sim-server keyboard path rather than to the simulator's HID stack in general.

## Hypothesis (untested)

The wedge direction is informative. If shift were stuck *down* we would see everything over-shifted; instead nothing is shifted, so the shift key-down is either not landing or being released before the character keypress registers — i.e. a modifier/keypress ordering race in the HID report sequence rather than a stuck modifier.

Worth checking whether the per-key cadence is involved: `keyboard`'s default `delayMs` is 50, and there is a related known issue where remote sim-server typing needs `delayMs >= 150` to be reliable. A first experiment would be to type `@` in a loop at varying `delayMs` until it wedges, and see whether a higher delay prevents it.

## Impact

[...]

PR #778, opened 2026-08-11
fix(flows): ride out iOS main-thread stalls instead of failing the step

filip131311:
## Problem

iOS flows against a heavy app (Expensify NewDot, release sim build) intermittently error with

```
✗  4 tap "Continue" — ViewInspector RPC timed out: Application.getState
```

The only known mitigation was enabling **Reduce Motion** on the simulator — which shouldn't be a functional requirement, and (as the investigation showed) wasn't actually the mechanism.

## Root cause

The injected ViewInspector handlers hop onto the app's **main thread** (`dispatch_sync` in `DevtoolsConnection.m`). During a heavy first-launch cold start (first Hermes parse, asset decode, on a loaded host) the main thread can be pinned past the flat **5s** RPC timeout, and the flow layer turns one timed-out read into a step error:

- every iOS tree read first runs target auto-resolution → an `Application.getState` probe per connected app, per read — the cheap probe is the first to hit the stall;
- `settleTree`'s 3s window < the 5s RPC timeout, so an action step's **first** stalled read exhausts the window and throws;
- a default `await` (7.5s) gets ~1 stalled attempt;
- the flow launch readiness gate (8s) sits below the same stall ceiling, because the dylib's connect is also main-thread-gated.

Steady-state RPCs are fast (instrumented: `getState` p50 0ms / max 402ms, `getFullHierarchy` p50 9–12ms — even with animations on and a 14-core CPU load). Only the stall window is the problem; Reduce Motion merely shortened it below the cliff (and correlated with cache warming from repeated relaunches).

Error-code note: these failures are `NATIVE_DEVTOOLS_RPC_TIMEOUT` on an **established** connection — distinct from the multi-server socket-contention family (#775), which surfaces as `could not connect to native devtools` at launch.

## Fix (three coordinated changes, tool-server only)

1. **`getFullHierarchy` RPC timeout 5s → 15s** — mirrors the Android devtools client's 5s default / 15s `getHierarchy` tiers. Every other method (incl. `getState` and the interactive point queries) keeps 5s, so one-shot agent tools stay snappy on a wedged app.
2. **Flows remember the launched app** (`launchedNativeApp` on the run state, threaded to iOS tree reads; invalidated by `launch-app`/`restart-app`/`reinstall-app`/`open-url`/`button` `tool:` steps; shared with nested `run:` flows). It is a **timeout arbiter only**: auto-resolution still runs on every read, and any *answered* resolution — including the deliberate "single app but backgrounded" guard — wins. The hint decides the target only when the resolution probe itself dies of an RPC timeout and the hinted app's devtools connection is still up.
3. **Launch readiness gate 8s → 15s** — same stall class, same budget.

Accepted trade-offs (from adversarial design review): a persistent outage now takes longer to report (a dead-but-connected app fails an `await` after up to ~2 slow reads instead of three 5s ones), and a stall that clears right at a settle deadline can act on a single post-stall read (previously a hard error). Both were judged worth the removal of the false-failure class; the review also killed an earlier variant of this fix that would have bypassed the backgrounded-app guard.

## Verification

- **Fault injection**: a DYLD-inserted dylib pins the app's main thread on demand (12s). Pre-fix, deterministically reproduces both historical signatures — mid-flow stall → `tap error: ViewInspector RPC timed out: Application.getState`; launch-window stall → `launch error: could not connect`. Post-fix: **5/5 green** across both stall placements ×2 + no-stall control.
- **Android regression**: the same sign-in flow on the Android emulator through the patched server — green (the hint is iOS-only by construction; Android/Chromium tree paths unchanged).
- **Unit tests**: new `flow-ios-tree-launch-hint.test.ts` (6 cases: timeout fallback, no-hint rethrow, disconnected-hint rethrow, backgrounded-guard preserved, non-timeout rethrow, answered-resolution-beats-hint); existing `flow-ios-tree-no-windows` and `native-target-app` suites green. Full tool-server suite: 3875 passed; the 7 failures in `bind-failure/startup/lens-telemetry` tests fail identically on pristine `origin/main` sources in this environment (pre-existing, unrelated).

Related: #775 (multi-server devtools socket contention, filed from the same investigation), #774 (sim keyboard shift wedge, unrelated mechanism ruled out during triage).

### DIFF
--- a/packages/tool-server/src/blueprints/native-devtools.ts
+++ b/packages/tool-server/src/blueprints/native-devtools.ts
@@ -478,6 +478,16 @@ export const nativeDevtoolsBlueprint: ServiceBlueprint<NativeDevtoolsApi, Device
         );
       }
       const id = nextRpcId++;
+      // The injected handlers hop onto the app's MAIN thread; a heavy cold
+      // start (first Hermes parse, asset decode) can pin it for several
+      // seconds, and the flow runner re-reads the hierarchy through exactly
+      // that window. Give the flows' workhorse read time to ride the stall out
+      // instead of failing the step — mirroring the Android devtools client's
+      // 5s default / 15s getHierarchy tiers. Everything else (including the
+      // per-read Application.getState probe and the interactive point queries)
+      // keeps the 5s ceiling so one-shot agent tools stay snappy on a wedged
+      // app.
+      const timeoutMs = method === "ViewHierarchy.getFullHierarchy" ? 15_000 : 5_000;
       return new Promise((resolve, reject) => {
         pendingRpc.set(id, { resolve, reject });
         conn.socket.write(
@@ -498,7 +508,7 @@ export const nativeDevtoolsBlueprint: ServiceBlueprint<NativeDevtoolsApi, Device
               })
             );
           }
-        }, 5000);
+        }, timeoutMs);
       });
     }
 
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -1417,7 +1417,11 @@ async function waitForIdle(
     // backend), so serializing them would double the round without buying
     // anything.
     const [read, frame] = await Promise.all([
-      settleWithin(fetchFlowTree(env.registry, env.device, env.launchedNativeApp), roundBudget, env.signal).then((r) => {
+      settleWithin(
+        fetchFlowTree(env.registry, env.device, env.launchedNativeApp),
+        roundBudget,
+        env.signal
+      ).then((r) => {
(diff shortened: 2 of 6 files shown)


filip131311:
Corroboration from a second environment (different session, iPhone 17 Pro / iOS 26.5, argent 0.20.0, Expensify full-login variant with a mock backend): Reduce Motion was load-bearing there too — the flow runner failed with the same ViewInspector RPC timeout signature without it, consistent with the cold-start-stall mechanism (that variant's first launch is heavier: from-scratch start plus full sign-in). Also for #775's impact ledger: the devtools socket steal cost that session a discarded measurement round and ~40 minutes across three incidents, two caused by a session working on a different platform entirely.

PR #780, opened 2026-08-12
fix(flow): floor every settle at two read attempts

j-piasecki:
`settleTree` bounded its polling by a 3s window but never bounded the number of reads it took, and a tree read can fail by TIMING OUT. Every tree source's RPC timeout outlasts that window, the shortest being iOS's 5s, so the window fit exactly one such read: "every read in the window failed", the condition the throw reports as a tree-source outage, collapsed into "the first read was slow", erroring a tap/type/scroll step on one transient blip with no retry at all.

The floor makes the deadline stop deciding how much evidence an outage verdict is drawn from. `await`/`assert` have never had this flaw, since `waitForCondition`'s final poll already guarantees a second attempt, so this is the same guarantee given to the directives that resolve through a settle.

A run whose reads answer inside the window pays nothing: two reads are what a settle already costs there. A run whose reads outlast it pays a second read, and gets a settle it never actually got before, since past the deadline the old code returned its single tree having compared it against nothing. A wedged source costs two full reads rather than one, and a read is several RPCs deep with each hop on its own timeout, so that is tens of seconds on a step that used to fail in one read. The alternative is the step failing on evidence the window was too short to gather.

### DIFF
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -220,6 +220,20 @@ const FOCUS_REPORTING_SOURCES: ReadonlySet<DescribeSource> = new Set([
 const SETTLE_POLL_MS = 250;
 const SETTLE_TIMEOUT_MS = 3000;
 
+// Read attempts every settle makes before it may conclude anything, enforced
+// even once the window has closed. A read can fail by TIMING OUT, and every
+// tree source's RPC timeout outlasts the 3s window (the shortest is iOS's 5s),
+// so the window fits only one such read: without a floor, "every read in the
+// window failed", which the throw below reports as a tree-source outage, would
+// collapse into "the first read was slow", erroring a step on one transient
+// blip with no retry at all. The second read is not bounded by the window
+// either, and a read is several RPCs deep, each on its own timeout, so a wedged
+// source costs two full reads and tens of seconds, and a source that answers
+// slowly pays a second read where it used to end on one. `await`/`assert` have
+// always had this floor in `waitForCondition`'s final poll; this is the same
+// guarantee for the directives that resolve through a settle.
+const SETTLE_MIN_READS = 2;
+
 // `scroll-to`: a bounded number of momentum-free increments. Each travels half
 // the clip window along the scroll axis (half the screen when no `within`
 // container is named) — < 1 viewport, so consecutive viewports overlap and a
@@ -428,25 +442,26 @@ function readFlowTree(env: ActionEnv): Promise<DescribeTreeData> {
  * from landing mid-deceleration (where a scroll view swallows it) or acting on a
  * frame that has already moved.
  *
- * Throws when EVERY read in the window failed: that is a tree-source outage
+ * Throws when EVERY read attempt failed: that is a tree-source outage
  * (e.g. native devtools disconnected mid-run — `fetchFlowTree` refuses to
  * degrade to a trimmed tree), not a mid-animation blip, and swallowing it would
  * convert the outage into a misleading "element not found" downstream. The
  * throw lands in the step's structured report via `execLeafStep`'s catch.
  *
+ * "Every attempt" is at least {@link SETTLE_MIN_READS} of them: the deadline
+ * bounds the polling, never the number of reads taken, so a read slow enough to
+ * outlast the window on its own is retried rather than treated as the whole
+ * evidence.
+ *
  * `skipProvenOutage` rethrows {@link ActionEnv.treeOutage} instead of buying
  * that window again. Not a shorter budget: a threshold low enough to spare a
  * source serving no tree at all would also abandon the mid-navigation blip the
- * retry above exists for. But the memo is a prediction: one window mints it,
- * nothing re-tests it, and on iOS a window can be one stalled read. The
- * deadline below is only tested after a read RETURNS, and the read a window
- * issues is `ViewHierarchy.getFullHierarchy` on the 15s RPC tier #778 gave it,
- * behind a serial 5s `Application.getState` probe, so a 3000ms window can spend
- * ~20s on the single sample it then mints the verdict from. Being wrong costs a
- * settle, not a step: {@link settleForGesture} swallows the throw - and says
- * so, warning every gesture it spares, so a run whose prediction was wrong
- * reports which greens went out unsettled. Any directive read that comes back
- * retires it, as does a relaunch.
+ * retry above exists for. But the memo is a prediction: one window mints it and
+ * nothing re-tests it. Being wrong costs a settle, not a step:
+ * {@link settleForGesture} swallows the throw - and says so, warning every
+ * gesture it spares, so a run whose prediction was wrong reports which greens
+ * went out unsettled. Any directive read that comes back retires it, as does a
+ * relaunch.
  *
  * What it costs: an outage that lasted a full window, in a run that then never
  * reads the tree again and never relaunches, leaves later gestures unsettled
@@ -463,6 +478,7 @@ export async function settleTree(
   let prevFp: string | undefined;
   let prevTree: DescribeNode | undefined;
   let lastError: Error | undefined;
+  let reads = 0;
   for (;;) {
     if (env.signal?.aborted) return undefined;
     // Inside the loop so the abort above still wins, but only ever true on the
@@ -481,6 +497,7 @@ export async function settleTree(
       // transient describe failure mid-navigation — retry until the deadline
       lastError = read.cause;
     }
+    reads += 1;
     // The abort can land while the read above is in flight (e.g. the HTTP
     // client disconnecting mid-flow trips the run's AbortController). Without
     // this re-check the two-identical-reads return below — or the deadline's
@@ -495,6 +512,10 @@ export async function settleTree(
       prevTree = tree;
     }
     if (Date.now() >= deadline) {
+      // Owed another attempt: retry back-to-back rather than sleeping out a
+      // poll interval the window no longer has (`waitForCondition`'s final
+      // poll does the same). Bounded — `reads` rises on every pass.
+      if (reads < SETTLE_MIN_READS) continue;
       if (prevTree === undefined && lastError !== undefined) {
         if (env.treeOutage) env.treeOutage.proven = { deviceId: env.device.id, error: lastError };
         throw lastError;
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -477,12 +477,12 @@ function readFlowTree(env: ActionEnv): Promise<DescribeTreeData> {
  * reports which greens went out unsettled. Any directive read that comes back
  * retires it, as does a relaunch.
  *
- * What it costs: an outage that lasted a full window, in a run that then never
- * reads the tree again and never relaunches, leaves later gestures unsettled
- * even once it clears. That is a flow of nothing but coordinate gestures giving
- * up a best-effort settle it never had before this directive existed, against
- * every one of those gestures otherwise paying the window for a settle that is
- * not coming.
+ * What it costs: an outage that failed every read attempt, in a run that then
+ * never reads the tree again and never relaunches, leaves later gestures
+ * unsettled even once it clears. That is a flow of nothing but coordinate
+ * gestures giving up a best-effort settle it never had before this directive
+ * existed, against every one of those gestures otherwise paying a whole settle,
+ * window and floor reads alike, for a tree that is not coming.
  */
 export async function settleTree(
   env: ActionEnv,
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -454,7 +454,11 @@ function readFlowTree(env: ActionEnv): Promise<DescribeTreeData> {
  * "Every attempt" is at least {@link SETTLE_MIN_READS} of them: the deadline
  * bounds the polling, never the number of reads taken, so a read slow enough to
  * outlast the window on its own is retried rather than treated as the whole
- * evidence.
+ * evidence. The price lands on the best-effort return: if that retry then
+ * fails, `prevTree` is the first read's tree handed back a read-duration later
+ * (bounded by the tree source's own RPC ceiling), older than what the bare
+ * deadline would have returned. Paid because a single successful read has been
+ * compared against nothing and is no settle at all.
  *
  * `skipProvenOutage` rethrows {@link ActionEnv.treeOutage} instead of buying
  * that whole settle again, window and floor reads alike. Not a shorter budget:
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -464,8 +464,13 @@ function readFlowTree(env: ActionEnv): Promise<DescribeTreeData> {
  * evidence. The price lands on the best-effort return: if that retry then
  * fails, `prevTree` is the first read's tree handed back a read-duration later
  * (bounded by the tree source's own RPC ceiling), older than what the bare
- * deadline would have returned. Paid because a single successful read has been
- * compared against nothing and is no settle at all.
+ * deadline would have returned. Paid for what the retry buys, which differs by
+ * branch. After a failed first read it buys evidence: the outage throw rests on
+ * two failed attempts rather than on one slow read, and a retry that succeeds
+ * there still returns a lone tree matched against nothing. After a successful
+ * first read it buys a fresher sample, since the retry's tree comes back
+ * whether or not the fingerprints match, and with it the settle's only chance
+ * to converge.
  *
  * `skipProvenOutage` rethrows {@link ActionEnv.treeOutage} instead of buying
  * that whole settle again, window and floor reads alike. Not a shorter budget:
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -1135,8 +1135,8 @@ async function runPinch(
  * the tree several times per step, so the extra fetch is noise, and the
  * resolution path every other directive shares stays untouched.
  *
- * The one read {@link ActionEnv.treeOutage} must not spare, unlike the settle
- * window {@link settleForGesture} skips: this answer is DISPATCHED, not waited
+ * The one read {@link ActionEnv.treeOutage} must not spare, unlike the whole
+ * settle {@link settleForGesture} skips: this answer is DISPATCHED, not waited
  * on. A stale verdict would degrade a centre rotate on a phone from the
  * edge-safe vertical placement the real aspect picks (Down points at
  * y = 0.278 / 0.722) to the aspect-1 fallback, which puts both fingers 0.48 off
@@ -1147,9 +1147,11 @@ async function runPinch(
  *
  * A true verdict costs one failed read instead, which the caller degrades past
  * exactly as it degrades past a skip - but not for free: this read carries no
- * budget and no signal, so on iOS it is the 15s hierarchy tier, longer than the
- * 3s window the memo saved. That is the price of never dispatching geometry off
- * a prediction, and it is what every rotate paid before the memo existed.
+ * budget and no signal, so on iOS it is the 15s hierarchy tier. Less than the
+ * whole settle the memo saved - window and floor reads alike, two reads on that
+ * same tier when the source fails by timing out - but still the price of never
+ * dispatching geometry off a prediction, and what every rotate paid before the
+ * memo existed.
  */
 async function fetchScreenAspect(env: ActionEnv): Promise<number | undefined> {
   try {
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -470,7 +470,10 @@ function readFlowTree(env: ActionEnv): Promise<DescribeTreeData> {
  * there still returns a lone tree matched against nothing. After a successful
  * first read it buys a fresher sample, since the retry's tree comes back
  * whether or not the fingerprints match, and with it the settle's only chance
- * to converge.
+ * to converge. The best-effort return also reaches `snapshot: { cropOn }`
+ * through {@link waitForFrame}, where the whole retry widens the gap between
+ * the read the crop rectangle comes from and the capture, so the crop can be
+ * measured against a layout the pixels no longer show.
  *
  * `skipProvenOutage` rethrows {@link ActionEnv.treeOutage} instead of buying
  * that whole settle again, window and floor reads alike. Not a shorter budget:
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -231,9 +231,13 @@ const SETTLE_TIMEOUT_MS = 3000;
 // blip with no retry at all. The second read is not bounded by the window
 // either, so a wedged source costs two full reads and tens of seconds, and a
 // source that answers slowly pays a second read where it used to end on one.
-// `scroll-to` multiplies that: it settles once per round, bounded by
-// `MAX_SCROLL_ITERATIONS` rather than a wall clock, so the added read lands on
-// every round and a slow source turns one step's tens of seconds into minutes.
+// Slower than the window is not the exceptional case: `HUNG_TREE_READ_MS`
+// reasons from 5400ms Android reads, so ending a window on a single read is
+// the ordinary outcome there and the second read an ordinary cost, not a
+// degraded one. `scroll-to` multiplies that: it settles once per round,
+// bounded by `MAX_SCROLL_ITERATIONS` rather than a wall clock, so the added
+// read lands on every round and a slow source turns one step's tens of seconds
+// into minutes.
 // `await`/`assert` have always had this floor in `waitForCondition`'s final
 // poll; this is the same guarantee for the directives that resolve through a
 // settle.
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -229,9 +229,12 @@ const SETTLE_TIMEOUT_MS = 3000;
 // blip with no retry at all. The second read is not bounded by the window
 // either, and a read is several RPCs deep, each on its own timeout, so a wedged
 // source costs two full reads and tens of seconds, and a source that answers
-// slowly pays a second read where it used to end on one. `await`/`assert` have
-// always had this floor in `waitForCondition`'s final poll; this is the same
-// guarantee for the directives that resolve through a settle.
+// slowly pays a second read where it used to end on one. `scroll-to` multiplies
+// that: it settles once per round, bounded by `MAX_SCROLL_ITERATIONS` rather
+// than a wall clock, so the added read lands on every round and a slow source
+// turns one step's tens of seconds into minutes. `await`/`assert` have always
+// had this floor in `waitForCondition`'s final poll; this is the same guarantee
+// for the directives that resolve through a settle.
 const SETTLE_MIN_READS = 2;
 
 // `scroll-to`: a bounded number of momentum-free increments. Each travels half
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -76,7 +76,7 @@ export interface ActionEnv {
   launchedNativeApp?: string;
   /**
    * Run-scoped memo of a tree source that answered nothing: set by a
-   * {@link settleTree} whose whole window failed, cleared by the next
+   * {@link settleTree} that failed every read attempt, cleared by the next
    * DIRECTIVE read that comes back. One holder per run (built in flow-run's
    * ExecState and shared by every `deviceEnv`), so the evidence one step
    * gathered is the evidence the next one acts on.
@@ -89,17 +89,17 @@ export interface ActionEnv {
    * ordered against the step that is running: `idle` hands its read to
    * `settleWithin` and stops waiting at the round budget, so that read can land
    * during a later step and retire a verdict minted after it was issued. Safe
-   * direction - that only ever costs a later gesture a window it would have
+   * direction - that only ever costs a later gesture a settle it would have
    * skipped - but it is a clear by arrival, not by sequence.
    *
-   * Only {@link settleForGesture} READS it, and only to skip a window it has
+   * Only {@link settleForGesture} READS it, and only to skip a settle it has
    * already been shown is unaffordable - never a read whose answer is
    * dispatched rather than waited on, which is why {@link fetchScreenAspect}
    * does not consult it. The skip is not silent: the gesture warns its step
    * report that it dispatched unsettled, so the memo never makes an outage
    * cheaper to miss than it was to prove. It is written by the one place that
-   * can prove a source dead (a {@link settleTree} whose whole window failed),
-   * whichever directive asked for that window. `waitForFrame` and
+   * can prove a source dead (a {@link settleTree} that failed every read
+   * attempt), whichever directive asked for that settle. `waitForFrame` and
    * `scrollToVisible` error the step on it, so a verdict of theirs is never
    * spent; `runSnapshot` is the one producer whose step then passes reporting
    * nothing, since it captures pixels regardless and has no warning to carry.
@@ -120,7 +120,7 @@ export interface ActionEnv {
    * move is behind `runLaunch`, which clears the verdict first - but it keeps
    * the property from resting on where that clear sits in another file. The
    * clear is deliberately NOT keyed: it only ever costs a later gesture a
-   * window it would otherwise have skipped, and that is the side to err on.
+   * settle it would otherwise have skipped, and that is the side to err on.
    */
   treeOutage?: { proven?: { deviceId: string; error: Error } };
 }
@@ -457,14 +457,14 @@ function readFlowTree(env: ActionEnv): Promise<DescribeTreeData> {
  * evidence.
  *
  * `skipProvenOutage` rethrows {@link ActionEnv.treeOutage} instead of buying
- * that window again. Not a shorter budget: a threshold low enough to spare a
- * source serving no tree at all would also abandon the mid-navigation blip the
- * retry above exists for. But the memo is a prediction: one window mints it and
- * nothing re-tests it. Being wrong costs a settle, not a step:
- * {@link settleForGesture} swallows the throw - and says so, warning every
- * gesture it spares, so a run whose prediction was wrong reports which greens
- * went out unsettled. Any directive read that comes back retires it, as does a
- * relaunch.
+ * that whole settle again, window and floor reads alike. Not a shorter budget:
+ * a threshold low enough to spare a source serving no tree at all would also
+ * abandon the mid-navigation blip the retry above exists for. But the memo is a
+ * prediction: one settle mints it and nothing re-tests it. Being wrong costs a
+ * settle, not a step: {@link settleForGesture} swallows the throw - and says
+ * so, warning every gesture it spares, so a run whose prediction was wrong
+ * reports which greens went out unsettled. Any directive read that comes back
+ * retires it, as does a relaunch.
  *
  * What it costs: an outage that lasted a full window, in a run that then never
  * reads the tree again and never relaunches, leaves later gestures unsettled
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -232,7 +232,13 @@ const FOCUS_REPORTING_SOURCES: ReadonlySet<DescribeSource> = new Set([
 // Settle detection: re-read the tree until two consecutive reads match, so a tap
 // never lands mid-fling and a resolved frame can't go stale before we act.
 const SETTLE_POLL_MS = 250;
-const SETTLE_TIMEOUT_MS = 3000;
+/**
+ * How long that re-reading gets. Exported for flow-settle-min-reads.test.ts,
+ * which sizes its slow reads past this window: a hand-copied number there would
+ * survive this one being raised above it, and the read counts it prices would
+ * quietly stop measuring the floor below.
+ */
+export const SETTLE_TIMEOUT_MS = 3000;
 
 // Read attempts every settle makes before it may conclude anything, enforced
 // even once the window has closed. A read can fail by TIMING OUT, and every
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -222,19 +222,21 @@ const SETTLE_TIMEOUT_MS = 3000;
 
 // Read attempts every settle makes before it may conclude anything, enforced
 // even once the window has closed. A read can fail by TIMING OUT, and every
-// tree source's RPC timeout outlasts the 3s window (the shortest is iOS's 5s),
-// so the window fits only one such read: without a floor, "every read in the
-// window failed", which the throw below reports as a tree-source outage, would
+// tree source's RPC timeout outlasts the 3s window: 5s is the shortest tier any
+// of them allows, and a whole read chains several of those (an iOS read spends
+// up to 5s resolving the target app before a 15s `getFullHierarchy`), so the
+// window fits only one such read. Without a floor, "every read in the window
+// failed", which the throw below reports as a tree-source outage, would
 // collapse into "the first read was slow", erroring a step on one transient
 // blip with no retry at all. The second read is not bounded by the window
-// either, and a read is several RPCs deep, each on its own timeout, so a wedged
-// source costs two full reads and tens of seconds, and a source that answers
-// slowly pays a second read where it used to end on one. `scroll-to` multiplies
-// that: it settles once per round, bounded by `MAX_SCROLL_ITERATIONS` rather
-// than a wall clock, so the added read lands on every round and a slow source
-// turns one step's tens of seconds into minutes. `await`/`assert` have always
-// had this floor in `waitForCondition`'s final poll; this is the same guarantee
-// for the directives that resolve through a settle.
+// either, so a wedged source costs two full reads and tens of seconds, and a
+// source that answers slowly pays a second read where it used to end on one.
+// `scroll-to` multiplies that: it settles once per round, bounded by
+// `MAX_SCROLL_ITERATIONS` rather than a wall clock, so the added read lands on
+// every round and a slow source turns one step's tens of seconds into minutes.
+// `await`/`assert` have always had this floor in `waitForCondition`'s final
+// poll; this is the same guarantee for the directives that resolve through a
+// settle.
 const SETTLE_MIN_READS = 2;
 
 // `scroll-to`: a bounded number of momentum-free increments. Each travels half
--- a/packages/tool-server/src/tools/flows/flow-actions.ts
+++ b/packages/tool-server/src/tools/flows/flow-actions.ts
@@ -101,32 +101,31 @@ export interface ActionEnv {
    * can prove a source dead (a {@link settleTree} that failed every read
    * attempt), whichever directive asked for that settle. `waitForFrame` and
    * `scrollToVisible` error the step on it, so a verdict of theirs is never
-   * spent; `runSnapshot` is the one producer whose step then passes reporting
-   * nothing, since it captures pixels regardless and has no warning to carry.
-   * Deliberate, but it means a snapshot's outage is heard about on the gesture
-   * that spends the verdict, or not at all if no gesture follows. It is cleared
-   * by {@link readFlowTree}, which every directive's reads go through, since a
-   * read that came back is evidence of health whichever directive asked for it.
-   * A relaunch clears it too, whether spelled `launch` or as one of flow-run's
-   * `FOREGROUND_CHANGING_TOOLS`: it is the repair the commonest of these errors
-   * asks for, and no read need follow it before a gesture does. So does a
-   * nested orchestrator step, which can do either out of this holder's sight.
-   * Absent for a caller that builds an `ActionEnv` by hand, which simply leaves
-   * every settle on its own budget.
+   * spent; `runSnapshot` captures pixels regardless and `VisualOutcome` has
+   * no `warning` field, so its step passes in silence: the outage surfaces on
+   * the gesture that spends the verdict, or not at all if none follows.
    *
-   * `runSnapshot` is the third caller {@link fetchFlowTree} names as swallowing
-   * such an outage, and it deliberately does NOT read the memo: a capture taken
-   * unsettled is one taken mid-animation, and it lands as a pixel mismatch on a
-   * step whose report has no `warning` to carry the caveat, so being wrong
-   * there would cost a step rather than a settle. Snapshots pay the full price
-   * for that: on a source that fails by timing out, every snapshot step re-buys
-   * the whole settle, window and floor reads alike, two reads where the window
-   * alone ended on one. That price buys something back. The settle reads
-   * through {@link readFlowTree}, so a snapshot re-tests the verdict rather
-   * than trusting it, and in a flow of nothing but snapshots and coordinate
-   * gestures - where the gesture settle skips itself while one stands - it is,
-   * with a rotate's {@link fetchScreenAspect}, one of the two reads left that
-   * can retire a wrong one.
+   * That same silence keeps `runSnapshot` deliberately off the reading side,
+   * though not on {@link fetchScreenAspect}'s grounds, since a snapshot's
+   * settle IS waited on. A capture taken on a stale verdict lands as a
+   * mid-animation pixel mismatch that reads as a visual regression, with
+   * nothing to say the screen was never settled: being wrong there costs a
+   * step, where the gesture risks only a settle. The price is that on a source
+   * failing by timeout, every snapshot step re-buys the whole settle, window
+   * and floor reads alike, two reads where the window alone ended on one. In
+   * exchange the settle reads through {@link readFlowTree}, so a snapshot
+   * re-tests the verdict instead of trusting it, and with a rotate's
+   * {@link fetchScreenAspect} it is one of the two reads left that can retire a
+   * wrong one in a flow of nothing but snapshots and coordinate gestures.
+   *
+   * It is cleared by {@link readFlowTree}, which every directive's reads go
+   * through, since a read that came back is evidence of health whichever
+   * directive asked for it. A relaunch clears it too, whether spelled `launch`
+   * or as one of flow-run's `FOREGROUND_CHANGING_TOOLS`: it is the repair the
+   * commonest of these errors asks for, and no read need follow it before a
+   * gesture does. So does a nested orchestrator step, which can do either out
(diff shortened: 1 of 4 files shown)


packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -195,6 +195,20 @@ const FOCUS_REPORTING_SOURCES: ReadonlySet<DescribeSource> = new Set([
...
 
+// Read attempts every settle makes before it may conclude anything, enforced
+// even once the window has closed. A read can fail by TIMING OUT, and every
+// tree source's RPC timeout outlasts the 3s window (the shortest is iOS's 5s),
+// so the window fits only one such read: without a floor, "every read in the
+// window failed", which the throw below reports as a tree-source outage, would
+// collapse into "the first read was slow", erroring a step on one transient
+// blip with no retry at all. The second read is not bounded by the window
+// either, and a read is several RPCs deep, each on its own timeout, so a wedged
+// source costs two full reads and tens of seconds, and a source that answers
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The per-settle cost is stated here, but one directive pays it up to `MAX_SCROLL_ITERATIONS = 25` times: `scrollToVisible` settles once per round and is bounded by iteration count, not by wall clock. Measured on a live Chromium/CDP device driving `flow-execute` with tree reads pinned at 4s (over the window), one `scroll-to` step: 51.7s / 12 tree reads on `3c0e87fb6` vs 95.8s / 23 reads at `c7462150` — +44s, +85%, both arms still passing. A synthetic run at 3.1s reads gives 77.6s → 155.2s for the same single step. Nothing hard-fails (there is no per-step or per-run budget, and `flow-execute` is `longRunning` so the MCP layer sets no fetch timeout), and reads that answer inside the window cost nothing — measured +8ms across a 6-tap flow, and identical read counts at 1.5s reads. The aggregate on the one looping consumer is what the sentence does not name.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -195,6 +195,20 @@ const FOCUS_REPORTING_SOURCES: ReadonlySet<DescribeSource> = new Set([
...
+// slowly pays a second read where it used to end on one. `await`/`assert` have
+// always had this floor in `waitForCondition`'s final poll; this is the same
+// guarantee for the directives that resolve through a settle.
+const SETTLE_MIN_READS = 2;
```
latekvo, line 264: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: Once a read outlasts the window, every consumer receives a tree sampled one read later than before, and `scrollToVisible` compares the settled tree's fingerprint ACROSS rounds to decide the container has hit its end. Base samples reads 1,2,3…; this samples 2,4,6…, so in-region motion whose period sits near the two-read cycle now aliases to equal fingerprints where it previously alternated.

Reproduced A/B at 3.2s reads against `3c0e87fb6`, with a list that swallows one scroll increment and a two-state blink inside the scrolled region:

- base: `swipes=2`, `["scroll-to:pass","tap:pass"]`
- head: `swipes=1`, `["scroll-to:fail","tap:skip"]`, `reason="reached the end of the scroll without finding text=\"Order #1234\""`

Same total read count on both sides — the difference is purely which reads the loop samples. A second variant of the same setup stops a round early and taps a still-clipped row at `y=0.975`, i.e. past the entry-edge rule at `flow-actions.ts:675-677`. The effect is not one-directional (a third variant has base burn all 25 swipes and fail where head passes), and a control with no in-region motion is identical on both sides, which is what isolates the trigger.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -388,12 +402,17 @@ function readFlowTree(env: ActionEnv): Promise<DescribeTreeData> {
  * from landing mid-deceleration (where a scroll view swallows it) or acting on a
  * frame that has already moved.
  *
- * Throws when EVERY read in the window failed: that is a tree-source outage
+ * Throws when EVERY read attempt failed: that is a tree-source outage
```
latekvo, line 474: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: This sentence was amended, but three twins stating the same condition were not — and after this change a whole window of failure is no longer sufficient for either the throw or the memo, since a settle that burns its window on one slow failed read takes another read first and concludes nothing if that one comes back:

- `flow-actions.ts:79` — "set by a {@link settleTree} whose whole window failed"
- `flow-actions.ts:86` — "the one place that can prove a source dead (a {@link settleTree} whose whole window failed)"
- `packages/tool-server/test/flows/flow-visual.test.ts:336` — "settleTree throws when every read in its window failed"

Line 417 in this same JSDoc ("instead of spending a window to re-derive it") names only part of what `skipProvenOutage` now saves, since the window can be followed by a second read the window does not bound.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -444,6 +465,10 @@ export async function settleTree(
       prevTree = tree;
     }
     if (Date.now() >= deadline) {
+      // Owed another attempt: retry back-to-back rather than sleeping out a
+      // poll interval the window no longer has (`waitForCondition`'s final
+      // poll does the same). Bounded — `reads` rises on every pass.
+      if (reads < SETTLE_MIN_READS) continue;
```
latekvo, line 559: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: On the path where the slow first read SUCCEEDED and the retry then throws, `prevTree` is still what comes back — but a full read-duration later. Reproduced against `3c0e87fb6` with a 3.2s first read and a throwing second: base returns the tree at 3202ms with an age of ~0ms; head returns the same tree at 6405ms with an age of 3205ms. With iOS's 15s `ViewHierarchy.getFullHierarchy` ceiling (`blueprints/native-devtools.ts:490`) the retry can hand the following gesture a frame up to 15s old. That cuts against the purpose stated three lines above the throw — "acting on a frame that has already moved" — and the JSDoc records the extra attempt without recording that the tree it returns can be older than the one it replaces.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -444,6 +465,10 @@ export async function settleTree(
...
       if (prevTree === undefined && lastError !== undefined) {
         if (env.treeOutage) env.treeOutage.proven = { deviceId: env.device.id, error: lastError };
         throw lastError;
```
latekvo, line 562: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: `lastError` is now the SECOND read's error where it used to be the first's, so a source that degrades between the two reads reports the less actionable of them — both in the step's `reason` and in the memo other steps rethrow. Reproduced with the strings this path really produces: read 1 slow and failing with `<bundleId> was launched before argent's instrumentation loaded — relaunch it (launch-app, or a flow \`launch\` step) so the full view hierarchy is readable` (`flow-ios-tree.ts:318`, reached only after `resolveNativeTargetApp`'s 5s `Application.getState` probe, i.e. a read that already outlasts the window), read 2 stalling harder. Base reports the relaunch instruction; head reports `native devtools RPC timed out after 5000ms`. The repair the first read named is the one dropped.

packages/tool-server/test/flows/flow-settle-min-reads.test.ts
```diff
@@ -0,0 +1,149 @@
...
+beforeEach(async () => {
+  tmpDir = await fs.mkdtemp(path.join(os.tmpdir(), "flow-settle-reads-"));
+  reads = 0;
+  onRead = async () => tree();
+});
+afterEach(async () => {
+  await fs.rm(tmpDir, { recursive: true, force: true });
+});
+
+describe("settleTree takes at least two read attempts", () => {
```
latekvo, line 109: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: All three slow-read cases in this block script the first read to FAIL slowly; none covers a first read that SUCCEEDS slowly. That is the second behaviour the PR description advertises — "gets a settle it never actually got before, since past the deadline the old code returned its single tree having compared it against nothing" — and nothing pins it. Probed by mutation in a scratch worktree at `c7462150`: making the retry conditional on no read having succeeded leaves all 4 tests in this file green and the whole flows suite green (46 files / 1070 tests). The behaviour is real, just unpinned — the same scroll-to flow on a live Chromium/CDP device with 4s reads takes 12 tree reads on `3c0e87fb6` and 23 at `c7462150`. For contrast, the eight other mutations I ran against the change (floor of 1, floor of 3, deleted `continue`, `continue`→`break`, floor moved below the throw, increment moved inside the `try`, window shrunk to 10ms, floor hoisted out of the deadline branch) were all caught.

packages/tool-server/test/flows/flow-settle-min-reads.test.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    const result = await run("tap-dead");
+
+    expect(result.ok).toBe(false);
+    expect(result.steps[0].reason).toMatch(/native devtools is unavailable/);
+    // Reads that fail fast still fill the window, so the floor changes nothing.
+    expect(reads).toBeGreaterThan(2);
+  }, 20_000);
+
+  it("does not spend a second read when the first one settles the screen", async () => {
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat)\]: The name says a second read is not spent, but the body asserts `expect(reads).toBe(2)` — one is — and the comment directly above that assertion says the same ("Two reads are what a settle costs a healthy run"). The condition in the name is also unreachable: a settle returns only when two consecutive fingerprints match (`flow-actions.ts:461-463`), so a first read can never settle the screen. Ran the file at `c7462150`: 4/4 green with `reads === 2`. What the test actually guards is that no THIRD read appears — it caught a floor applied outside the deadline branch when I mutated for it — but the name is what shows in suite output and in a `-t` filter, and it reads as a spec for a one-read fast path that does not exist.

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: Correct, and the sentence now names it (`8e27a285`). `scrollToVisible` really is the only consumer whose cost scales with round count rather than wall clock: its loop is `for (let i = 0; i < MAX_SCROLL_ITERATIONS; i++)` with the settle as the first statement of each round and no deadline check anywhere in the body, where `waitForFrame` bounds itself on `DEFAULT_ACTION_TIMEOUT_MS` and so overshoots by at most one settle.

The comment now reads: "`scroll-to` multiplies that: it settles once per round, bounded by `MAX_SCROLL_ITERATIONS` rather than a wall clock, so the added read lands on every round and a slow source turns one step's tens of seconds into minutes."

Tens of seconds to minutes is the honest shape of your 51.7s to 95.8s, and it is the number an author actually feels.

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: Correct on all four, fixed in `01c3e5fd`. The twins now read "a {@link settleTree} that failed every read attempt", reusing the vocabulary the throw paragraph already chose rather than inventing a third phrasing, and the same swap landed on the `flow-visual.test.ts` comment. Two more sites in that JSDoc carried the same shorthand for what the memo saves and went with them: "only to skip a settle it has already been shown is unaffordable" and "it only ever costs a later gesture a settle it would otherwise have skipped".

On your fourth point, `skipProvenOutage` now says it rethrows "instead of buying that whole settle again, window and floor reads alike".

Two "window" phrases were deliberately left, since the added read does not falsify either: "a source that failed a whole window and has answered nothing since" is excluded by its own second clause once the floor read comes back, and "an outage that lasted a full window, in a run that then never reads the tree again" is excluded the same way, because the floor read is a read of the tree.

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: Right, and it was unrecorded. Traced it the same way you did: a slow first read sets `prevTree`, the deadline has passed but `reads` is 1, so `continue` takes read 2; if that one throws, `prevTree === undefined` is false so the throw is skipped and the first read's tree comes back, one full read-duration later than the bare deadline would have returned it.

`5ff247ab` records it on the paragraph that describes the extra attempt: "The price lands on the best-effort return: if that retry then fails, `prevTree` is the first read's tree handed back a read-duration later (bounded by the tree source's own RPC ceiling), older than what the bare deadline would have returned. Paid because a single successful read has been compared against nothing and is no settle at all."

Left the ceiling non-numeric on purpose rather than hardcoding your 15s, since `fetchFlowTree` fans out to four platform sources and the iOS number lives in `native-devtools.ts`. The retry itself stays: a lone read is not a settle, and the second sentence is there so the staleness reads as the price of the comparison rather than an oversight.

packages/tool-server/test/flows/flow-settle-min-reads.test.ts
j-piasecki: Confirmed and covered in `096c4dcc`. Your mutation is exactly the one that survived, so the new test is built to kill it.

`settles against the retry when the only read so far succeeded slowly` scripts read 1 to sleep 3200ms and then succeed, and read 2 to return the same screen with the button moved from `y=0.4` to `y=0.7`. It asserts both the read count and the dispatched point, so it pins the consequence and not just the arithmetic: the gesture acts on the frame the second read saw.

Sensitivity, run rather than assumed. With `if (reads < SETTLE_MIN_READS && prevTree === undefined) continue;` applied the file goes `1 failed | 4 passed`, only the new test failing on `expected 1 to be 2`. Relaxing the read-count assertion while still mutated makes the coordinate assertion fail on its own, `y: 0.45` received against `0.75` expected, which is the mutant tapping the pre-deadline frame. Source then restored and verified byte-for-byte by hash.

`0.5 / 0.75` was read off an actual dispatch rather than derived from the tap-point formula.

packages/tool-server/test/flows/flow-settle-min-reads.test.ts
j-piasecki: Both halves correct, renamed in `be28d5c1`. The early return is `if (prevFp !== undefined && fp === prevFp) return tree;` and `prevFp` is undefined on the first pass, so the old name really did specify a one-read fast path the code cannot enter.

It now reads `adds no third read when the first two reads matched inside the window`, which is what the body pins and what your out-of-the-deadline-branch mutation was caught by. The "inside the window" clause also keeps it distinct from the new slow-success test, which lands on `reads === 2` too but by way of the floor.

The comment above the assertion lost the half that moved into the name and is now just "Two reads are what a settle has always cost a healthy run", which is the reason two is the right expectation: it is the pre-floor baseline, so the floor is free here.

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: The mechanism is real and traces cleanly: `scrollToVisible` decides end-of-scroll by comparing one round's in-region fingerprint against the previous round's, and once a read outlasts the window each round returns read N+1's tree where it used to return read N's. The compared samples move. Not going to change the code for it, though, and here is why.

That check is a two-sample equality test, so it has an aliasing set by construction: the motion periods that happen to land on equal samples. The base cadence has one too. It is not immune, it aliases a different set, and a change that re-phases sampling moves the set rather than creating it. `flow-actions.ts:735` already records the residual, in the comment explaining why text stays in the fingerprint for in-scope nodes: "an animating node INSIDE the scrolled content remains a known limitation." That comment predates this PR and its base, sitting on `main` at `fe59768`.

Telling "the container stopped" from "the container is moving with a period near the sampling cadence" needs more than the two samples the heuristic has. Any gate I could write instead keys on a proxy: more samples or longer waits are a cost every scroll pays for a case that only bites when a node inside the scrolled region animates, and none of it would carry a test that means anything. Your own data makes the point better than I can: a third variant has base burn all 25 swipes and fail where head passes. An effect that flips outcomes in both directions inside a documented residual is not something to build a guard against.

What I do concede is the part you caught in the twin thread: the cost sentence in the code named the per-settle price and stopped there, saying nothing about the one looping consumer. That is fixed in `8e27a285`.

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: The observation holds and the scenario is reachable. `flow-ios-tree.ts:318` is reached only after `resolveNativeTargetApp`'s 5s probe, so that relaunch instruction genuinely does arrive on a read that has already outlasted the window, and the diff does not touch the `lastError` assignment, so the second read's error wins. Leaving it as is, for two reasons.

Last-wins is not something this diff introduced. Checked the base body directly: `3c0e87fb6` has the identical `lastError = err instanceof Error ? ... : ...` with no first-wins guard, so a fast-failing source already fills the window with several reads and reports the last one's error while discarding every earlier one. That path is pinned today by `leaves a fast-failing outage on its existing budget`, which asserts `reads > 2`. What this PR changes is how many reads happen, which is its declared purpose. The error-selection rule is untouched.

And first-wins is not strictly better, so there is no signal to pick on. Neither direction dominates. In your scenario read 1 carries the actionable repair. In the mirror scenario, a source that only gets far enough to identify the real problem on the retry, read 2 does. Ranking two error strings by actionability needs information the code does not have, and switching to first-wins would regress the common multi-read window path by reporting a 3s-old first blip in place of the most current observation.

The loss you describe is real, though, and I would take a fix at the other end: if the relaunch instruction specifically has to survive a degrading source, the place to make that true is where the error is constructed, so the hint is sticky or classified and any later error can carry it. That beats inverting a uniform selection rule inside `settleTree` to rescue one error string.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -447,6 +475,10 @@ export async function settleTree(
       prevTree = tree;
     }
     if (Date.now() >= deadline) {
+      // Owed another attempt: retry back-to-back rather than sleeping out a
+      // poll interval the window no longer has (`waitForCondition`'s final
+      // poll does the same). Bounded — `reads` rises on every pass.
+      if (reads < SETTLE_MIN_READS) continue;
```
hubgan, line 559: The extra read is taken after the window has already closed, so it is bounded only by the tree source's own RPC ceiling rather than by anything the settle controls. On the callers that resolve through a single settle that is a clean doubling, and `scroll-to` multiplies it: `scrollToVisible` settles once per round and is capped only by `MAX_SCROLL_ITERATIONS`, with no wall clock above it — `execLeafStep` wraps the directive in a try/catch only, `flow-run` has no per-step or per-run deadline, and `http.ts` sets no request timeout.

Measured on a real device with 3.5s tree reads: a 25-round `scroll-to` went from 25 reads / 100.0s to 50 reads / 186.9s, both runs ending on the same `not found after 25 scroll attempts`. At the per-read ceilings the sources actually allow (15s for `ViewHierarchy.getFullHierarchy` and Android's `getHierarchy`, 10s per Chromium evaluate) that is the difference between minutes and tens of minutes for one step.

The comment at 207-213 does name this cost, but it frames a source answering slower than the window as the exception. On Android it is a case the codebase already sizes other flow logic against — `flow-actions.ts:1386` reasons from "the 5400ms Android reads this step is sized against", which is well past the 3s window, so `reads === 1` at the deadline is an ordinary outcome there rather than a degraded one.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -391,19 +408,28 @@ function readFlowTree(env: ActionEnv): Promise<DescribeTreeData> {
...
- * RPC timeout). Being wrong costs a settle, not a step: {@link settleForGesture}
- * swallows the throw. Any read that comes back retires it, as does a relaunch.
+ * that whole settle again, window and floor reads alike. Not a shorter budget:
+ * a threshold low enough to spare a source serving no tree at all would also
+ * abandon the mid-navigation blip the retry above exists for. But the memo is a
+ * prediction: one settle mints it and nothing re-tests it. Being wrong costs a
+ * settle, not a step: {@link settleForGesture} swallows the throw. Any read that
+ * comes back retires it, as does a relaunch.
  *
  * What it costs: an outage that lasted a full window, in a run that then never
```
hubgan: This paragraph was left in the framing the rest of the block just retired. It still says the memo is minted by "an outage that lasted a full window" and that the saving is those gestures "otherwise paying the window", but after this change the memo is minted only once the window has closed *and* the read floor is satisfied, and what a skipped gesture now avoids is two full reads, not a 3s window.

As it stands it contradicts line 427 in the same comment, which this PR rewrote to "that whole settle again, window and floor reads alike". The equivalent wording was updated in `flow-visual.test.ts`, so this reads like the one spot the sweep missed. Understating the saving as a window also inverts the cost/benefit this paragraph exists to weigh.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -198,6 +198,23 @@ const FOCUS_REPORTING_SOURCES: ReadonlySet<DescribeSource> = new Set([
...
+// window failed", which the throw below reports as a tree-source outage, would
+// collapse into "the first read was slow", erroring a step on one transient
+// blip with no retry at all. The second read is not bounded by the window
+// either, and a read is several RPCs deep, each on its own timeout, so a wedged
+// source costs two full reads and tens of seconds, and a source that answers
+// slowly pays a second read where it used to end on one. `scroll-to` multiplies
+// that: it settles once per round, bounded by `MAX_SCROLL_ITERATIONS` rather
+// than a wall clock, so the added read lands on every round and a slow source
+// turns one step's tens of seconds into minutes. `await`/`assert` have always
+// had this floor in `waitForCondition`'s final poll; this is the same guarantee
```
hubgan: "this is the same guarantee" claims more than the two mechanisms share. `waitForCondition` takes a read at or after its deadline unconditionally — its `finalPoll` flag fires however many reads already happened inside the window. The floor here is conditional on `reads < SETTLE_MIN_READS`, so a settle that already made a dozen reads takes no post-deadline read at all.

They coincide only in the slow-first-read case, which is the case this comment is about, so the reasoning holds — it is the equivalence claim that doesn't.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -198,6 +198,23 @@ const FOCUS_REPORTING_SOURCES: ReadonlySet<DescribeSource> = new Set([
 const SETTLE_POLL_MS = 250;
 const SETTLE_TIMEOUT_MS = 3000;
 
+// Read attempts every settle makes before it may conclude anything, enforced
+// even once the window has closed. A read can fail by TIMING OUT, and every
+// tree source's RPC timeout outlasts the 3s window (the shortest is iOS's 5s),
```
hubgan: "the shortest is iOS's 5s" reads as an iOS property, but 5s is the default RPC tier on Android too (`android-devtools-client.ts:25`), and Vega's `forward --remove` sits at 5s as well. The argument the sentence makes is sound — the shortest RPC timeout on any flow tree path is still greater than the 3s window — but the parenthetical is the number a reader anchors on when sizing what the floor costs, and it is neither iOS-specific nor the ceiling of a whole read, which chains several of these (5s + 15s on iOS).

packages/tool-server/test/flows/flow-settle-min-reads.test.ts
```diff
@@ -0,0 +1,180 @@
...
+      return tree();
+    };
+    await writeTap("tap-slow-blip");
+
+    const result = await run("tap-slow-blip");
+
+    // One slow blip is not an outage: the retry read the screen, so the tap
+    // resolved and went out instead of erroring the step.
+    expect(result.ok).toBe(true);
+    expect(reads).toBeGreaterThanOrEqual(2);
```
hubgan: The real count here is exactly 2 — I instrumented it — and the sibling tests below use `toBe(2)`. As written this would still pass if the floor degenerated into an open-ended retry on the slow-fail path, which is the one thing this test is positioned to catch.

The test also never asserts the dispatched coordinate, unlike its sibling at 136, so which tree the settle actually acted on stays unpinned: read 2 lands with `prevFp` still undefined, so what comes back is a lone read compared against nothing — precisely what the comment at 131-132 calls "no settle at all".

packages/tool-server/test/flows/flow-settle-min-reads.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    // the window being spent does not excuse skipping the retry.
+    expect(reads).toBe(2);
+    // Centre of the MOVED frame (0.45 would be the first read's): the tap lands
+    // where the second read saw the button, not on the stale pre-deadline one.
+    expect(result.calls).toEqual([
+      { tool: "gesture-tap", args: { udid: DEVICE, x: 0.5, y: 0.75 } },
+    ]);
+  }, 20_000);
+
+  it("still calls a sustained outage an outage, on the read after the slow one", async () => {
```
hubgan, line 182: The branch the JSDoc explicitly prices isn't covered: first read succeeds slowly, the floor read then fails. That is the one case where the floor changes what the caller acts on rather than just when — `prevTree` is the first read's tree, handed back a full read later, and no outage memo is minted. The four cases here cover slow-fail + success, slow-success + success, slow-fail + fail, and fast-fail, but not slow-success + fail.

packages/tool-server/test/flows/flow-settle-min-reads.test.ts
```diff
@@ -0,0 +1,180 @@
...
+beforeEach(async () => {
+  tmpDir = await fs.mkdtemp(path.join(os.tmpdir(), "flow-settle-reads-"));
+  reads = 0;
+  onRead = async () => tree();
+});
+afterEach(async () => {
+  await fs.rm(tmpDir, { recursive: true, force: true });
+});
+
+describe("settleTree takes at least two read attempts", () => {
```
hubgan, line 109: Nothing here or in `flow-scroll.test.ts` pins the multiplied cost the constant's own comment calls out for `scroll-to` — those tests all use fast reads, where the floor never fires. Against a source whose reads outlast the window I measured a 25-round `scroll-to` going from 25 reads to 50, so the behaviour the comment predicts is real and currently unguarded; a read-count assertion on a scroll round would catch a future change that made it worse.

hubgan (commented):
Reviewed the read floor end-to-end. I ran the branch and the base build side by side against a real Chromium device whose tree reads were made to take 3.5s (a page-side hook on `document.documentElement`, which is where the walker starts every read — argent itself is unmocked), and counted the reads each settle actually issued.

The fix does what it says: a settle that used to end on one uncompared read now takes two.

| step | base | this branch |
| --- | --- | --- |
| coordinate `tap` | 1 read / 3.87s | 2 reads / 7.87s |
| `tap` by selector, resolves | 1 read / 3.93s | 2 reads / 7.42s |
| `tap` by selector, never resolves | 3 reads / 11.53s | 4 reads / 14.75s |
| `scroll-to`, 25 rounds | 25 reads / 100.02s | 50 reads / 186.93s |

`npx vitest run test/flows` is green on the branch (46 files, 1074 tests), and prettier/eslint are clean on the changed files.

Comments below are mostly about the comment block and test coverage; one is about where the doubled cost lands.

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: Correct, and the parenthetical was doing exactly the anchoring you describe. 5s is the floor of the whole set rather than an iOS property: `android-devtools-client.ts:25` carries the same default, Vega's `forward --remove` is 5s, and Chromium's CDP default is 10s, so the argument the sentence makes survives unchanged.

`0f658473` rewrites it to: "every tree source's RPC timeout outlasts the 3s window: 5s is the shortest tier any of them allows, and a whole read chains several of those (an iOS read spends up to 5s resolving the target app before a 15s `getFullHierarchy`), so the window fits only one such read."

That picks up your second half too. `queryFullHierarchyTree` goes through `resolveNativeTargetApp`'s `Application.getState` on the 5s tier before the 15s hierarchy call, so the ceiling on one iOS read is 20s, not 5. The line further down that said "a read is several RPCs deep, each on its own timeout" was making the same point and went with the rewrite rather than stating it twice.

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: The cost half the comment already names, from `8e27a285`: `scroll-to` settling once per round under `MAX_SCROLL_ITERATIONS` with no wall clock above it, tens of seconds turning into minutes. Your 25 reads / 100.0s to 50 reads / 186.9s is the same shape as the numbers that sentence was written from.

The framing half it did not, and that is the better point. `flow-actions.ts:1391` reasons from "the 5400ms Android reads this step is sized against", and a 5400ms read ends the window on one attempt as a matter of course, so `reads === 1` at the deadline is the normal case there rather than a symptom.

`7f711664` adds: "Slower than the window is not the exceptional case: `HUNG_TREE_READ_MS` reasons from 5400ms Android reads, so ending a window on a single read is the ordinary outcome there and the second read an ordinary cost, not a degraded one."

The multiplied cost is now also pinned by a test rather than only described - see the reply on the `flow-settle-min-reads.test.ts` thread.

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: Right, and the code says so plainly. `waitForCondition`'s `finalPoll` never consults a read counter, so it fires at the deadline however many reads preceded it, while the floor here is gated on `reads < SETTLE_MIN_READS` and a settle that already made a dozen reads takes nothing extra.

`18a63d20` drops the equivalence: "`await`/`assert` never had this hole: `waitForCondition` takes a poll at its deadline unconditionally, however many reads preceded it. The floor here is narrower, firing only when the window closed on fewer than two reads, but it covers the same case: a first read that outlasts the window on its own."

That keeps the reasoning the sentence was there for and stops claiming the two mechanisms are one.

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: Correct on both phrases, and this is the one spot the `01c3e5fd` sweep missed.

I did defend two "window" phrases in the earlier twin thread as not falsified by the floor, but that answered a different objection. Yours is that the paragraph contradicts a line the same PR rewrote, and that understating the saving as a window inverts the cost/benefit the paragraph exists to weigh. Both hold.

`50a6ca47`: the minting condition becomes "an outage that failed every read attempt", the vocabulary already used at `flow-actions.ts:79` and `:101` and in `flow-visual.test.ts:336`; the saving becomes "otherwise paying a whole settle, window and floor reads alike, for a tree that is not coming", which reuses line 464's phrasing rather than inventing a third one.

packages/tool-server/test/flows/flow-settle-min-reads.test.ts
j-piasecki: Both halves correct, fixed in `9912e00b`.

Exactly 2 is what the path gives: read 1 fails past the window, `reads < SETTLE_MIN_READS` forces the retry, read 2 sets `prevTree` with `prevFp` still undefined, and the deadline branch returns it. The assertion is now `toBe(2)`, matching its siblings.

For the coordinate, read 2 returns the button at `MOVED_BUTTON_FRAME`, so the expected point is one only the retry's tree can produce, and the check is the sibling's full-call shape: `{ tool: "gesture-tap", args: { udid: DEVICE, x: 0.5, y: 0.75 } }`.

Sensitivity run rather than assumed. With `SETTLE_MIN_READS = 3` this test fails on `expected 3 to be 2`. I also ran the pre-change body against the same mutant to check your specific claim: it passes. The third read matches the second's fingerprint and returns early, so reads is 3, `>= 2` holds, and the coordinate is unchanged at `0.75`. The tightening to `toBe(2)` is the only thing that catches it.

packages/tool-server/test/flows/flow-settle-min-reads.test.ts
j-piasecki: Confirmed, and it was the one branch the JSDoc prices with nothing under it. `b265115f` adds `settles against the slow first read when the retry then fails`: read 1 sleeps past the window and succeeds, read 2 throws, and it asserts `reads === 2` plus the dispatched point `y: 0.45`, the centre of the first read's frame. It sits directly after the slow-success + success case, so the two pin which tree wins in each direction, `0.75` there and `0.45` here.

No separate memo assertion, and deliberately so: the memo write lives inside `if (prevTree === undefined && lastError !== undefined)`, the same branch that throws, so a dispatched tap already proves it was not entered. A comment records that, so nobody adds a redundant probe step later.

Sensitivity: mutating that guard's `prevTree` to `tree`, so a failed retry discards the good tree instead of handing it back, gives `1 failed | 5 passed` with only the new test failing.

packages/tool-server/test/flows/flow-settle-min-reads.test.ts
j-piasecki: Fair, and writing it corrected an assumption I had been carrying. `2b4c252e` adds `spends two reads a round against a source slower than the window`: a two-round `scroll-to` with every read at 3200ms, asserting one swipe and `reads === 4`.

The obvious fixture, a static screen with slow reads, turned out to be a dead guard. Two matching fingerprints return above the deadline branch in `settleTree`, so a static screen costs 2 reads per round whatever `SETTLE_MIN_READS` is set to. I ran it at 3 and it still passed at 4. The fixture therefore drifts the button between the two reads of each round, which forces every settle out through the deadline branch, and the count then tracks the constant in both directions: `SETTLE_MIN_READS = 3` gives `expected 6 to be 4`, deleting the floor gives `expected 2 to be 4`. The second of those is your 25-to-50 measurement at two rounds instead of 25.

Two rounds rather than more because it is the fewest that shows a per-round rather than a per-settle cost, and the wall clock is the reason to stop there: the test is 12.8s of scripted sleeps and takes the file from about 16s to about 29s. `SLOW_READ_MS` has to stay above the 3s window, so that is the floor for observing this without faking the read-versus-window relationship the test is measuring.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -89,16 +89,16 @@ export interface ActionEnv {
...
    * `settleWithin` and stops waiting at the round budget, so that read can land
    * during a later step and retire a verdict minted after it was issued. Safe
-   * direction - that only ever costs a later gesture a window it would have
+   * direction - that only ever costs a later gesture a settle it would have
    * skipped - but it is a clear by arrival, not by sequence.
    *
    * Only {@link settleForGesture} and {@link fetchScreenAspect} READ it, each
-   * to skip work it has been shown cannot pay off: a settle window, and a read
-   * whose failure the caller degrades past anyway. The skip is not silent: the
+   * to skip work it has been shown cannot pay off: a settle, and a read whose
```
hubgan: This clause was rewritten here from "a settle window" to "a settle", and `settleTree`'s docblock got the same treatment in two places. `fetchScreenAspect`'s docblock further down was left behind: it still describes the same skip as `settleForGesture` skipping "its window", and still prices the aspect read against "the 3s window just saved". On a source that fails by timing out that skip now spares two reads rather than one, so the paragraph names a quantity the code no longer saves. `settleForGesture`'s own docblock is unaffected - the passages there are about sources that fail fast, where the window still binds.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -421,32 +445,37 @@ function readFlowTree(env: ActionEnv): Promise<DescribeTreeData> {
...
  * throw lands in the step's structured report via `execLeafStep`'s catch.
  *
+ * "Every attempt" is at least {@link SETTLE_MIN_READS} of them: the deadline
+ * bounds the polling, never the number of reads taken, so a read slow enough to
+ * outlast the window on its own is retried rather than treated as the whole
+ * evidence. The price lands on the best-effort return: if that retry then
+ * fails, `prevTree` is the first read's tree handed back a read-duration later
+ * (bounded by the tree source's own RPC ceiling), older than what the bare
+ * deadline would have returned. Paid because a single successful read has been
+ * compared against nothing and is no settle at all.
```
hubgan: The floor is on read attempts, not on compared reads, so this justification does not hold on either branch it covers. When the slow first read fails and the retry succeeds, the settle returns exactly one successful read compared against nothing - the state this sentence calls "no settle at all". When the slow first read succeeds, the retry's comparison never changes the outcome: a matching fingerprint returns the retry's tree at the convergence check, and a mismatch returns the retry's tree from the deadline branch. What the second read buys on that path is a fresher sample, not the comparison.

packages/tool-server/src/tools/flows/flow-actions.ts
```diff
@@ -421,32 +445,37 @@ function readFlowTree(env: ActionEnv): Promise<DescribeTreeData> {
...
  * degrade to a trimmed tree), not a mid-animation blip, and swallowing it would
  * convert the outage into a misleading "element not found" downstream. The
  * throw lands in the step's structured report via `execLeafStep`'s catch.
  *
+ * "Every attempt" is at least {@link SETTLE_MIN_READS} of them: the deadline
+ * bounds the polling, never the number of reads taken, so a read slow enough to
+ * outlast the window on its own is retried rather than treated as the whole
+ * evidence. The price lands on the best-effort return: if that retry then
+ * fails, `prevTree` is the first read's tree handed back a read-duration later
+ * (bounded by the tree source's own RPC ceiling), older than what the bare
```
hubgan, line 485: Worth naming where that later return lands. `snapshot: { cropOn }` resolves its crop rectangle through `waitForFrame` and then captures pixels, and the comment at that call site states that the pixels captured are the state the frame was resolved from. The tree handed back is unchanged - still the first read's - but the gap between resolving the rectangle and taking the screenshot grows by the whole retry, up to the source's RPC ceiling. The consequence there differs in kind from a mis-aimed gesture: a crop measured against a layout the capture no longer shows. Nothing exercises the costly version of it either - the case added here has the retry fail instantly, and the cropOn tests mock `waitForFrame` out entirely.

packages/tool-server/test/flows/flow-visual.test.ts
```diff
@@ -333,7 +333,7 @@ describe("runSnapshot cross-app key collision", () => {
 
 describe("runSnapshot settle", () => {
   it("proceeds to capture when the tree source is down", async () => {
-    // settleTree throws when every read in its window failed (native devtools
+    // settleTree throws when every read attempt failed (native devtools
```
hubgan, line 336: `runSnapshot` is the one settle caller that swallows a tree-source outage and still does not pass `skipProvenOutage`, so on a run that has already proven the source dead every snapshot step re-buys the whole settle and then discards the throw. `flow-tree.ts` names three callers that swallow such a failure under a single criterion, but only two of them read the memo. The exclusion predates this change; what changes is the price - for a source that fails by timing out it goes from one read per snapshot step to two, so roughly 5s becomes 10s at the shortest RPC tier and 20s becomes 40s in the iOS worst case. Fast-failing sources are unaffected either way.

packages/tool-server/test/flows/flow-settle-min-reads.test.ts
```diff
@@ -0,0 +1,242 @@
...
+    return onRead(reads);
+  }),
+}));
+
+import { createRunFlowTool, type FlowRunResult } from "../../src/tools/flows/flow-run";
+import { serializeFlow } from "../../src/tools/flows/flow-utils";
+
+const DEVICE = "00000000-0000-0000-0000-0000000000ab"; // iOS UDID shape
+// Comfortably past the 3000ms settle window, the way a tree RPC's own timeout is.
+const SLOW_READ_MS = 3200;
```
hubgan: This 3200 is a hand-copied companion to the unexported `SETTLE_TIMEOUT_MS` of 3000, tied to it only by the comment above. The scroll-to case's `reads === 4` rests on that ordering rather than on the floor: with the window raised to 4000 and the floor removed entirely, that case still passes at 4 reads - the first read then lands inside the window and the second arrives by the ordinary poll, same number, different code path (I ran it). The four tap cases do fail under that mutation, so the file would still go red, but the one assertion that prices the per-round cost would have quietly stopped covering it.

packages/tool-server/test/flows/flow-settle-min-reads.test.ts
```diff
@@ -0,0 +1,242 @@
...
+  tmpDir = await fs.mkdtemp(path.join(os.tmpdir(), "flow-settle-reads-"));
+  reads = 0;
+  swiped = false;
+  onRead = async () => tree();
+});
+afterEach(async () => {
+  await fs.rm(tmpDir, { recursive: true, force: true });
+});
+
+describe("settleTree takes at least two read attempts", () => {
```
hubgan, line 109: Nothing here, and nothing in the sibling abort suites, cancels a run while the retry itself is in flight. `flow-gesture-settle.test.ts` covers a cancellation during the first read by hanging it, and its other abort cases run at zero read delay, so the deadline never passes and this branch is never reached. The retry introduces no new branch - the same two signal checks bracket it - but the `continue` does skip the poll sleep, and that interleaving is currently only reasoned about in the comments.

hubgan (commented):
Reviewed the read floor itself, all four `settleTree` call sites, and the timing, cancellation and outage-memo paths around them; also built the branch and ran flows through it against a real Chromium device and an iOS simulator, alongside the parent branch for comparison.

No blocking findings. The loop terminates, cancellation still bounds the wait, the outage verdict can only be minted on more evidence than before rather than less, and a healthy run measured the same as the parent branch end to end. Everything below is documentation and test coverage.

hubgan (approved):

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: Correct, and both phrases are gone in `17f18339` - though the rebase changed which fix was the right one. `fetchScreenAspect` no longer reads the memo at all on the new base, so there is no skip left to misprice; the docblock had already been rewritten around that, and what my commit changes is the two clauses that still carried the old quantity.

"unlike the settle window `settleForGesture` skips" becomes "unlike the whole settle `settleForGesture` skips", and the pricing clause inverts, which is the part your comment predicted: it used to read "the 15s hierarchy tier, longer than the 3s window the memo saved", and now reads "the 15s hierarchy tier. Less than the whole settle the memo saved - window and floor reads alike, two reads on that same tier when the source fails by timing out - but still the price of never dispatching geometry off a prediction". Once the floor lands, the saving is no longer the smaller number, so the sentence was not just understating a quantity, it was comparing in the wrong direction.

Your exclusion held on review: `settleForGesture`'s own docblock is untouched, since the passages there are about sources that fail fast, where the window still binds.

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: Right on both branches, and traced rather than taken on the word of the sentence. Fixed in `5d83db9d`.

After a failed first read the retry lands with `prevFp` still undefined, so the convergence check cannot fire and what comes back is exactly the lone successful read the sentence called no settle at all. After a successful first read the retry's tree comes back either way - matching fingerprints return it at the convergence check, a mismatch returns it from the deadline branch - so the comparison never steers the return. The floor is on attempts, as you said, and the old justification claimed a comparison it does not deliver.

The paragraph now states what each branch actually buys: "After a failed first read it buys evidence: the outage throw rests on two failed attempts rather than on one slow read, and a retry that succeeds there still returns a lone tree matched against nothing. After a successful first read it buys a fresher sample, since the retry's tree comes back whether or not the fingerprints match, and with it the settle's only chance to converge."

The lone-uncompared-read case is conceded in the text rather than dropped - it is the truth on that branch, and naming it is better than hiding it behind the old claim. The staleness sentence above it was accurate and is untouched.

packages/tool-server/src/tools/flows/flow-actions.ts
j-piasecki: Named in `e9525564`: "The best-effort return also reaches `snapshot: { cropOn }` through {@link waitForFrame}, where the whole retry widens the gap between the read the crop rectangle comes from and the capture, so the crop can be measured against a layout the pixels no longer show."

One wording note against your phrasing: the code order is read, retry, resolve rectangle, capture, so the staleness is in the tree the rectangle came from rather than in a pause between resolving it and capturing. Same consequence, and the sentence is written that way round.

On the coverage half I am not adding a test, and the reason is not that the code is obviously correct. cropOn resolves through `waitForFrame(env, opts.cropOn)`, the same call `tap` makes into the same body - there is no second frame-selection path to pin. The costly branch is already pinned through the real settle by `settles against the slow first read when the retry then fails`, which asserts the gesture acted on the first read's frame (`y: 0.45`, against `0.75` for the retry's) and `reads === 2`; that is precisely which tree a crop rectangle would be measured from. What separates "retry fails instantly" from "retry fails slowly" is only elapsed time between the first read and the capture: same tree, same rectangle, same read count. A test asserting that gap would be asserting the duration of its own `sleep` in the `fetchFlowTree` mock, and I could not name a mutation it would catch that the min-reads suite misses.

You are right that nothing drives cropOn through the real settle. What such a test would exercise is the `waitForFrame` return already pinned, composed with a resolve-then-capture ordering that `flow-visual.test.ts` already pins in three places (frame throws, frame misses, frame aborts - each asserts no screenshot is taken). Happy to be pushed on this if you can name the mutation it would kill.

packages/tool-server/test/flows/flow-visual.test.ts
j-piasecki: Every fact here checks out, and the exclusion is staying. Recorded instead in `c5a7b6d9`, on the `ActionEnv.treeOutage` docblock where the readers are enumerated.

The reason it is not an oversight: the memo is a prediction nothing re-tests, and what makes it affordable for `settleForGesture` is that being wrong costs a settle rather than a step - the gesture still dispatches and the step report carries `unsettledGestureWarning`, so the run is told what the green bought. A snapshot has no such escape. `VisualOutcome` has no `warning` field and flow-run's `case "snapshot"` maps only `status`/`reason`/`snapshotKey`/`artifacts`, so a capture taken on a stale verdict lands as a pixel mismatch that reads as a visual regression, with nothing attached to say the screen was never settled. Being wrong there costs a step.

There is a second reason I did not see until writing it up, and it is the stronger one. `runSnapshot`'s settle reads through `readFlowTree`, so a snapshot re-tests the verdict rather than trusting it. Passing `skipProvenOutage` would remove one of the few reads that can still retire a wrong memo in exactly the flows a wrong memo hurts most.

Your price is in the text, in reads rather than seconds, since flow-actions.ts deliberately keeps platform RPC numbers out of its prose: "on a source that fails by timing out, every snapshot step re-buys the whole settle, window and floor reads alike, two reads where the window alone ended on one."

One thing your comment is owed: since the rebase, the paragraph above this one already says only `settleForGesture` reads the memo and already names `runSnapshot` as the producer whose step passes reporting nothing with no warning to carry. So the addition now overlaps it - "the third caller ... does NOT read the memo" is stale framing against a paragraph that has narrowed the readers to one, and "no warning to carry" is said twice about fifteen lines apart. Worth a follow-up pass to merge the two rather than leaving both.

packages/tool-server/test/flows/flow-settle-min-reads.test.ts
j-piasecki: Confirmed by running your mutation before changing anything, and fixed in `924c276e`.

On the pristine file, window at 4000 with the floor deleted: the four tap cases fail (`expected 3 to be 2` twice, `expected 5 to be 2` twice) and the scroll case passes at 4 reads, exactly as you describe. The assertion that prices the per-round cost had stopped covering it while the file still went red for other reasons.

`SETTLE_TIMEOUT_MS` is now exported and `SLOW_READ_MS = SETTLE_TIMEOUT_MS + 200`, so the ordering cannot drift. Under the same mutation the scroll case now fails on `expected 2 to be 4`, which is the pre-floor cost of one read per round - what the assertion exists to catch.

The export carries its reason at the declaration rather than only in the test: "Exported for flow-settle-min-reads.test.ts, which sizes its slow reads past this window: a hand-copied number there would survive this one being raised above it, and the read counts it prices would quietly stop measuring the floor below."

Nothing derives from `SETTLE_MIN_READS` - only the sleep is tied to a constant, and every read count stays a literal, so the existing sensitivity survives: `SETTLE_MIN_READS = 3` still gives `expected 6 to be 4`. knip does not flag the new export, since test files are entry points there.

packages/tool-server/test/flows/flow-settle-min-reads.test.ts
j-piasecki: Right, and covered in `e9b125d6` by `ends a run cancelled while the floor's retry is in flight`. Read 1 sleeps past the window and succeeds, so the deadline closes on a lone read and the retry goes out back-to-back; read 2 never answers, and the abort fires once it is provably in flight (the fixture resolves a deferred from inside the read, after the counter increments, so the handshake cannot race ahead of it).

It asserts the skip and its reason, no gesture dispatched, `reads === 2` so it is the floor's read that was cancelled and not the first, and that the abort ends the run in about 13ms rather than waiting out a read that never answers.

Sensitivity, run rather than assumed:

- floor deleted: fails on `expected [ 'tap:pass' ] to deeply equal [ 'tap:skip' ]` - read 2 never happens, so the settle returns the slow first read's tree at the deadline and the tap goes out.
- second signal check deleted (the one after `reads += 1`): same failure, and against the whole `test/flows` suite it is the only test of 1113 that catches it. `flow-gesture-settle.test.ts` does not, for the reason you gave: its abort cases run at zero read delay, so the deadline has not passed and `sleepOrAbort` catches the already-aborted signal at the poll instead.

One thing your comment names that is still uncovered, and I am reporting it rather than claiming otherwise: replacing the floor's `continue` with a fall-through to the poll sleep passes the entire flows suite, 1113 of 1113. The back-to-back interleaving remains argued only in the comments. The observable difference is 250ms of wall clock, so a test for it would be a timing assertion on a sleep, but I would rather say so than leave the impression it is pinned.

packages/tool-server/test/flows/flow-visual.test.ts
j-piasecki: Follow-up on the overlap I flagged above: closed in `b62e7415`, a net deletion.

The `runSnapshot` story is now one argument in three paragraphs cut at topic seams, rather than two paragraphs each carrying half of it. The first ends on `runSnapshot` as a producer whose own step passes over the verdict it minted - "captures pixels regardless and `VisualOutcome` has no `warning` field, so its step passes in silence". The second opens "That same silence keeps `runSnapshot` deliberately off the reading side", so the no-warning fact is stated once and leaned on twice. The third is the clearing material, word-identical to before and only reflowed by the split.

The "third caller" opener is gone. Writing the merge also surfaced something the original addition had left implicit and worth stating: the exclusion is not on `fetchScreenAspect`'s grounds. That criterion is "never a read whose answer is dispatched rather than waited on", and a snapshot's settle IS waited on, so as written it would have admitted `runSnapshot`. The paragraph now says so and gives the actual ground - the missing warning channel, so being wrong costs a step where the gesture risks only a settle.

One detail I nearly lost and put back: "with a rotate's `{@link fetchScreenAspect}` it is one of the two reads left that can retire a wrong one in a flow of nothing but snapshots and coordinate gestures". A first pass softened that to "one of the few reads left", which drops the only place `fetchScreenAspect` is credited as a retirer rather than a non-consulter. Checked that it really is two: the snapshot's settle on both the plain and `cropOn` paths, and `fetchScreenAspect`; `settleForGesture` skips itself while a verdict stands, and the focus wait, the await/assert poll and the idle poll all need selector steps such a flow has none of.

`test/flows` is green at 46 files / 1118 tests, prettier and `tsc --noEmit` clean.

PR #783, opened 2026-08-13
fix(network): report UTF-8 response byte size

### LINKED ISSUE #662: view-network-logs reports response size in characters but renders it as bytes
Found while working on #633.

`packages/tool-server/src/utils/debugger/scripts/network-interceptor.ts:86` records

```js
entry.encodedDataLength = body.length;
```

where `body` is the result of `response.text()` — a **character** count, not a byte count.

`packages/tool-server/src/tools/network/network-logs.ts` then renders that value through `formatBytes`, so the log prints e.g. `559 B`.

For any non-ASCII payload the two differ: a UTF-8 JSON response of 1000 characters containing CJK text or emoji is ~3000 bytes, and the tool reports `1000 B`. The name `encodedDataLength` is borrowed from CDP, where it genuinely means bytes on the wire (that is what the Chromium branch of the same tool reports), so the RN and Chromium halves of one tool disagree about what the number means.

Not urgent — the ordering and rough magnitude are still useful — but it is a number presented with more precision than it has, and the fix is small: measure the encoded byte length, or rename the field and label the column as characters on the RN path.

mjq2020:
## Summary

- measure React Native response bodies from a cloned Blob instead of a decoded JavaScript string
- report one consistent metric: decoded response entity bytes
- avoid Content-Length because HEAD/304 and compressed responses give it different semantics
- keep response text capture independent, and leave the size unknown if Blob conversion is unavailable

## Root cause

body.length counts UTF-16 code units. Re-encoding the decoded text as UTF-8 fixes ordinary UTF-8 text, but it overcounts replacement characters from non-UTF-8 and binary responses. Content-Length is also not a safe fallback: it can describe an unsent GET body for HEAD/304 or the compressed wire representation while the response body is decompressed.

Response.blob().size measures the bytes in the decoded entity supplied by the React Native response. The same measurement therefore applies to text, binary, empty, cached, and compressed responses.

## Tests

- network interceptor regression: 23/23 passed
- covers Unicode JSON, HEAD, 304, non-UTF-8 replacement characters, compression, binary bodies, and unavailable Blob conversion
- npm run build
- npm run typecheck:tests -w @argent/tool-server
- targeted ESLint and Prettier checks on both changed files

Fixes #662

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reported network response sizes to reflect the actual decoded response data.
  - Corrected size calculations for binary content, compressed responses, and text containing non-ASCII characters.
  - Added reliable handling for `HEAD` and `304` responses, which now report zero size.
  - Prevented failed size calculations from incorrectly reporting a zero-byte response.

### DIFF
--- a/packages/tool-server/src/utils/debugger/scripts/network-interceptor.ts
+++ b/packages/tool-server/src/utils/debugger/scripts/network-interceptor.ts
@@ -20,6 +20,32 @@ export const NETWORK_INTERCEPTOR_SCRIPT = `(function() {
   var MAX_ENTRIES = 2000;
   function genId() { return 'rn-net-' + (nextReqId++); }
   function ts() { return Date.now() / 1000; }
+  function utf8ByteLength(value) {
+    if (typeof TextEncoder !== 'undefined') {
+      return new TextEncoder().encode(value).length;
+    }
+
+    var length = 0;
+    for (var i = 0; i < value.length; i++) {
+      var code = value.charCodeAt(i);
+      if (code <= 0x7f) {
+        length += 1;
+      } else if (code <= 0x7ff) {
+        length += 2;
+      } else if (code >= 0xd800 && code <= 0xdbff && i + 1 < value.length) {
+        var next = value.charCodeAt(i + 1);
+        if (next >= 0xdc00 && next <= 0xdfff) {
+          length += 4;
+          i += 1;
+        } else {
+          length += 3;
+        }
+      } else {
+        length += 3;
+      }
+    }
+    return length;
+  }
 
   function getOrCreate(reqId) {
     if (byId[reqId]) return byId[reqId];
@@ -83,7 +109,7 @@ export const NETWORK_INTERCEPTOR_SCRIPT = `(function() {
         var cloned = response.clone();
         cloned.text().then(function(body) {
           entry.state = 'finished';
-          entry.encodedDataLength = body.length;
+          entry.encodedDataLength = utf8ByteLength(body);
           entry.durationMs = Math.round((ts() - t) * 1000);
           entry.responseBody = body;
         }).catch(function() {
--- a/packages/tool-server/src/utils/debugger/scripts/network-interceptor.ts
+++ b/packages/tool-server/src/utils/debugger/scripts/network-interceptor.ts
@@ -20,56 +20,6 @@ export const NETWORK_INTERCEPTOR_SCRIPT = `(function() {
   var MAX_ENTRIES = 2000;
   function genId() { return 'rn-net-' + (nextReqId++); }
   function ts() { return Date.now() / 1000; }
-  function utf8ByteLength(value) {
-    if (typeof TextEncoder !== 'undefined') {
-      return new TextEncoder().encode(value).length;
-    }
-
-    var length = 0;
-    for (var i = 0; i < value.length; i++) {
-      var code = value.charCodeAt(i);
-      if (code <= 0x7f) {
-        length += 1;
-      } else if (code <= 0x7ff) {
-        length += 2;
-      } else if (code >= 0xd800 && code <= 0xdbff && i + 1 < value.length) {
-        var next = value.charCodeAt(i + 1);
-        if (next >= 0xdc00 && next <= 0xdfff) {
-          length += 4;
-          i += 1;
-        } else {
-          length += 3;
-        }
-      } else {
-        length += 3;
-      }
-    }
-    return length;
-  }
-
-  function responseByteLength(body, mimeType, headers) {
-    var keys = Object.keys(headers);
-    for (var i = 0; i < keys.length; i++) {
-      if (keys[i].toLowerCase() !== 'content-length') continue;
-
-      var value = String(headers[keys[i]]).trim();
-      if (/^\\d+$/.test(value)) return Number(value);
-    }
-
-    if (typeof body !== 'string') return 0;
-
-    var textual = mimeType.indexOf('text/') === 0
-      || mimeType === 'application/json'
-      || mimeType.slice(-5) === '+json'
-      || mimeType === 'application/xml'
-      || mimeType.slice(-4) === '+xml'
-      || mimeType === 'application/javascript'
-      || mimeType === 'application/x-javascript'
-      || mimeType === 'application/graphql'
-      || mimeType === 'application/x-www-form-urlencoded';
-
-    return textual ? utf8ByteLength(body) : body.length;
-  }
 
   function getOrCreate(reqId) {
     if (byId[reqId]) return byId[reqId];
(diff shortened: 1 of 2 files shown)


packages/tool-server/src/utils/debugger/scripts/network-interceptor.ts
```diff
@@ -83,7 +109,7 @@ export const NETWORK_INTERCEPTOR_SCRIPT = `(function() {
         var cloned = response.clone();
         cloned.text().then(function(body) {
           entry.state = 'finished';
-          entry.encodedDataLength = body.length;
+          entry.encodedDataLength = utf8ByteLength(body);
```
hubgan: `text()` gives text, not the bytes of the response.

If the body is not UTF-8 text, React Native on Android replaces each byte that is not UTF-8 with the character U+FFFD. One such byte becomes one character. Thus `body.length` was the same as the number of bytes. A UTF-8 count is larger, because U+FFFD is three bytes.

Measurements on a Pixel 7 emulator (API 34, RN 0.81.5):

- 1024-byte `application/octet-stream`: main gives 1024 B, this branch gives 2048 B
- 16-byte PNG: main gives 16 B, this branch gives 18 B

The 1024-byte body becomes 1024 characters. 512 of them are U+FFFD. The count is 512 + (512 x 3) = 2048. Most bytes in an image are not UTF-8. Thus the count for such a file is near three times the correct value.

On iOS, `text()` gives `undefined` for the same responses. Hermes changes `undefined` to the text "undefined". The count is then 9 B for each of these responses.

mjq2020: Good catch — the UTF-8 conversion is only valid for textual responses. I updated the interceptor to prefer a valid Content-Length header, use UTF-8 byte length only for textual MIME types, preserve body.length for binary/unknown responses, and return 0 when iOS exposes no body and there is no length header. Added regressions for Android replacement characters, Content-Length, and an unavailable binary body; the focused suite now passes 21/21, along with the repository build, test typecheck, and targeted ESLint.

packages/tool-server/src/utils/debugger/scripts/network-interceptor.ts
```diff
@@ -20,6 +20,56 @@ export const NETWORK_INTERCEPTOR_SCRIPT = `(function() {
...
+    return length;
+  }
+
+  function responseByteLength(body, mimeType, headers) {
+    var keys = Object.keys(headers);
+    for (var i = 0; i < keys.length; i++) {
+      if (keys[i].toLowerCase() !== 'content-length') continue;
+
+      var value = String(headers[keys[i]]).trim();
+      if (/^\\d+$/.test(value)) return Number(value);
```
hubgan: This line returns the `Content-Length` value before the code looks at the body.

Two responses carry a `Content-Length` for a body that the server does not send. A `HEAD` response gives the length that a `GET` would return. A `304` response gives the length of the cached entity. The client receives zero bytes in both cases.

Before this change, `body.length` gave `0` for these responses. That value was correct.

I measured a `HEAD` request on an iPhone 17 Pro (iOS 26.4) and on an Android emulator (API 30). The server sent `Content-Length: 524288000` and no body. Both platforms gave the same result.

| | reported |
|---|---|
| before this change | `0` |
| with this change | `524288000` |

`view-network-logs` prints the row as `Fetch 500 MB`. The request transferred no data.

The `304` case is the same on iOS. The app must set `If-None-Match` in the request. JS then receives the `304` with an empty body. Before this change the value was `0`. With this change it is the size of the cached entity.

packages/tool-server/src/utils/debugger/scripts/network-interceptor.ts
```diff
@@ -20,6 +20,56 @@ export const NETWORK_INTERCEPTOR_SCRIPT = `(function() {
...
+    }
+
+    if (typeof body !== 'string') return 0;
+
+    var textual = mimeType.indexOf('text/') === 0
```
hubgan: This test matches the `text/` prefix. It ignores the `charset` parameter of the media type.

React Native decodes the body as UTF-8. If the charset is not UTF-8, each byte outside ASCII becomes U+FFFD. `utf8ByteLength` then encodes each U+FFFD as three bytes.

Before this change, `body.length` counted one UTF-16 unit for each U+FFFD. For a single-byte charset, that count is equal to the number of bytes on the wire.

I measured this on an Android emulator (API 30). The server sent `Café naïve résumé` as ISO-8859-1. The response was 17 bytes and had no `Content-Length`.

| | reported |
|---|---|
| before this change | `17` (equal to the bytes on the wire) |
| with this change | `25` |

The count of 25 is 13 ASCII characters, plus 4 replacement characters at 3 bytes each.

The description of this pull request says that the code keeps the decoded string length for binary responses, so that it does not encode replacement characters again. That protection covers only the binary branch. A `text/*` response with a charset that is not UTF-8 takes this branch.

iOS does not show this result. `clone().text()` returns `undefined` for the same response, so both versions report `0`.

packages/tool-server/src/utils/debugger/scripts/network-interceptor.ts
```diff
@@ -20,6 +20,56 @@ export const NETWORK_INTERCEPTOR_SCRIPT = `(function() {
...
+      || mimeType === 'application/json'
+      || mimeType.slice(-5) === '+json'
+      || mimeType === 'application/xml'
+      || mimeType.slice(-4) === '+xml'
+      || mimeType === 'application/javascript'
+      || mimeType === 'application/x-javascript'
+      || mimeType === 'application/graphql'
+      || mimeType === 'application/x-www-form-urlencoded';
+
+    return textual ? utf8ByteLength(body) : body.length;
```
hubgan: This line returns the size of the decoded body. Line 56 returns the value of the `Content-Length` header. These are two different measurements.

On iOS, NSURLSession gives React Native a decompressed body. It keeps the original `Content-Length` and `Content-Encoding` headers. A compressed response with the header therefore reports the compressed size. A compressed response without the header reports the decompressed size.

Before this change, every row used one measurement. A reader could compare the rows with each other.

I measured two public endpoints on an iPhone 17 Pro (iOS 26.4):

| endpoint | `content-encoding` | `content-length` | body | reported |
|---|---|---|---|---|
| `api.github.com/repos/facebook/react-native` | gzip | `1510` | 6308 chars | `1510` (compressed) |
| `registry.npmjs.org/react-native/latest` | gzip | absent | 4729 chars | `4729` (decompressed) |

The two rows appear together in one log. The output does not show which measurement each row uses.

The platforms also disagree. On Android, OkHttp removes `Content-Length` and `Content-Encoding` when it decompresses the body. One gzipped response reported `107` on iOS and `40025` on Android. OkHttp removes the headers only when the app does not set `Accept-Encoding` or `Range`.

packages/tool-server/src/utils/debugger/scripts/network-interceptor.ts
mjq2020: Agreed. I removed Content-Length from the calculation entirely. The interceptor now measures the cloned response Blob, so HEAD and 304 responses report the actual empty entity as 0 even when the header describes a GET or cached body. Both cases now have regressions.

packages/tool-server/src/utils/debugger/scripts/network-interceptor.ts
mjq2020: Agreed. The MIME/charset branch and UTF-8 re-encoding are gone. Blob.size is read before text decoding, so the ISO-8859-1 example remains 17 bytes even if text() exposes four U+FFFD characters. The new regression explicitly checks 17 rather than the 25-byte re-encoding.

packages/tool-server/src/utils/debugger/scripts/network-interceptor.ts
mjq2020: Agreed. The implementation now ignores Content-Length for every response and consistently reports the decoded entity Blob size. A compressed response therefore uses the same decompressed-body measurement whether or not Content-Length is present, and across both platforms. The regression uses a compressed length of 1510 and a decoded entity size of 6308 and expects 6308.

hubgan (approved):

hubgan:
Thanks for contributing

PR #787, opened 2026-08-13
fix(installer): never create Cursor's permissions.json (it turns off Run Everything)

filip131311:
## The bug

Reported by Magda: with Cursor's run mode set to **Run Everything (unsandboxed)**, running `argent init` silently reverts it to **Allowlist**.

Her diagnosis was exactly right. `argent init` creates `~/.cursor/permissions.json` as `{ "mcpAllowlist": ["argent:*"] }` when the user has no such file, and Cursor derives its Run Mode from that file (verified in the Cursor 3.15.6 bundle):

```js
shouldPermissionsFileConstrainUnrestrictedMode(){
  …
  if (t === "unrestricted") return false;
  if (t === "allowlist" || t === "manual") return true;
  return mcpAllowlist.length > 0 || terminalAllowlist.length > 0;   // ← our write
}
```

The UI then reads `Run Mode (Enforced by ~/.cursor/permissions.json)`. It doesn't matter *what* we write — the file merely being non-empty is enough.

**Second effect, same write:** once the file has an `mcpAllowlist`, it *supersedes* the in-IDE MCP allowlist — `getEffectiveMcpAllowlist` returns the file's list and ignores `composerState.mcpAllowedTools`, and `canAddToAllowlistFromIde("mcp")` goes `false` (editor read-only). A user who had "Always allow"-ed other MCP servers silently stops getting them.

Not to be confused with the *older* clobbering bug (argent rewriting the file and dropping the user's rules) — that one was fixed by the JSONC append path and is covered by existing tests.

## The fix

`addAllowlist` becomes **append-only, never create**:

| `~/.cursor/permissions.json` | before | after |
|---|---|---|
| absent | created → **run mode demoted** | left alone |
| exists, no `mcpAllowlist` | key added → **in-IDE allowlist taken over** | left alone |
| exists with `mcpAllowlist` | `argent:*` appended | `argent:*` appended (unchanged) |

Skipping costs those users nothing: under Run Everything the allowlist is never consulted, so the rule we were adding was already a no-op. Where the user already keeps a file-based allowlist, both effects are their own doing and `argent:*` changes no state — so we still append there.

Existing files are never deleted, argent-only shape included. A file an earlier argent created is indistinguishable from one the user keeps on purpose, and if they like that allowlist it should stay. Removing it is a manual step (or `argent uninstall`, whose `removeAllowlist` already prunes the entry and drops the file when nothing else is left).

`init` now reports the skip instead of claiming it added the rule:

```
- Cursor (skipped - creating ~/.cursor/permissions.json would turn off Cursor's "Run Everything" mode)
```

## Rejected alternative

Writing `"approvalMode": "unrestricted"` alongside the allowlist does silence the constraint — but Cursor treats a permissions-file `approvalMode` as authoritative, not permissive:

```js
getModeFullAutoRun(e){ switch(…){ case "unrestricted": return true;   // forces it ON
```

That would push users who deliberately chose "Ask every time" into Run Everything. Disqualified.

## Scope

IDE only. The `cursor-agent` CLI is unaffected — its permissions-file provider reports `approvalMode: "unrestricted"` and only a provider reporting `"allowlist"` downgrades the merge.

Already-demoted machines are not repaired by this PR: those users delete `~/.cursor/permissions.json` (or drop the `argent:*` rule) once, and Run Everything returns — Cursor evaluates the constraint on read and never writes the demotion back, so nothing needs reconfiguring.

## Test plan

- 4 new cases in `mcp-configs.test.ts`: no file created (+ note), file without `mcpAllowlist` untouched, existing argent-only file left in place, user-owned file with `argent:*` left byte-intact
- Existing Cursor allowlist tests (comment/foreign-rule preservation, remove round-trip, argent-only detection) unchanged and green
- `packages/argent-installer` suite: 558 passed; `tsc --build`, `eslint --max-warnings 0`, both `knip` gates (215/215) clean

---

## Adopted from #742, generalized

Two ideas from #742 land here in reworked form (credit @softwarebyze):

- **`--no-allowlist`** on `init` and `update` skips the editor auto-approve allowlist step entirely — a scriptable opt-out for CI images and provisioning, where the prompt never runs. Documented in both `--help` outputs (the flags-sync test keeps them honest); `allowlist_decision` telemetry gains `decided_by: flag | default | prompt` so an automated skip is distinguishable from a user declining.

- **Update's allowlist refresh never changes the approval state the user has.** `addAllowlist` takes `refresh: true` on update's re-run; every adapter now declines (with a note) to re-impose its opted-in value when it is absent or was changed since init:
  - Cursor: `argent:*` missing from an existing list is not re-appended
  - Zed: a `tool_permissions.default` the user reset from `"allow"` stays (the guard honors Zed's project-over-global settings merge)
  - Codex: opt-in derives from entry values (≥1 `"approve"`), a restricted entry keeps its mode, an all-deny or removed table stays; the one refresh write left anywhere is backfilling tool ids a new version added, for opted-in tables
  - Claude Code: the permissions rule is not re-added (and `settings.json` not created) on refresh
  - Windsurf/Kiro: a narrowed `alwaysAllow`/`autoApprove` stays narrowed
  - Gemini/opencode: an unset/falsified `trust`/tools toggle stays
  
  Unlike #742's `createIfMissing:false` (which silently made update read-only), the option says what it means, and each skip prints a note (`- Cursor: skipped - argent:* is not on this allowlist; opt in via argent init`) so a missing auto-approve entry after an update is explainable from the output. `argent init` remains the explicit (re-)opt-in path.

## Known limitations (follow-up material)

- No consent bit is persisted anywhere, so refresh infers intent from file shape — it cannot tell "removed on purpose" from "never added" and errs toward not writing.
- The agent-triggered auto-update (`update-argent` tool) spawns `argent update --yes` detached with `stdio: "ignore"`, so the skip notes are not visible on that path and `--no-allowlist` cannot be passed there. The refresh guards themselves do apply, which removes the harmful overrides; durable observability there needs a log file or telemetry from update.

## Test plan (additions)

- 571 installer tests (13 new: per-adapter refresh guards incl. Zed scope-merge and Codex all-deny, `--no-allowlist` parsing), 295 telemetry, 80 argent (flags-sync green)
- Live from a packed tgz in a sandboxed `HOME`: `--help` shows the flag; `init -y --no-allowlist` writes nothing and says so; removed `argent:*` / reverted Zed default / deleted opencode entry all survive `argent update` with notes printed; fully opted-in update prints zero notes and keeps state; `update --no-allowlist` prints its trace line

### DIFF
--- a/packages/argent-installer/src/init-allowlist.ts
+++ b/packages/argent-installer/src/init-allowlist.ts
@@ -54,8 +54,12 @@ export async function configureAllowlist(args: {
       continue;
     }
     try {
-      adapter.addAllowlist!(effectiveRoot, scope);
-      lines.push(`${pc.green("+")} ${adapter.name}`);
+      const note = adapter.addAllowlist!(effectiveRoot, scope);
+      lines.push(
+        note
+          ? `${pc.yellow("-")} ${adapter.name} ${pc.dim(`(${note})`)}`
+          : `${pc.green("+")} ${adapter.name}`
+      );
     } catch (err) {
       lines.push(`${pc.red("x")} ${adapter.name}: ${pc.dim(String(err))}`);
     }
--- a/packages/argent-installer/src/init-allowlist.ts
+++ b/packages/argent-installer/src/init-allowlist.ts
@@ -14,12 +14,19 @@ export async function configureAllowlist(args: {
   effectiveRoot: string;
   scope: "local" | "global";
   nonInteractive: boolean;
+  /** --no-allowlist: skip the whole step without prompting */
+  noAllowlist: boolean;
 }): Promise<AllowlistResult> {
-  const { adapters, effectiveRoot, scope, nonInteractive } = args;
+  const { adapters, effectiveRoot, scope, nonInteractive, noAllowlist } = args;
   const adaptersWithAllowlist = adapters.filter((a) => a.addAllowlist);
   const adaptersWithoutAllowlist = adapters.filter((a) => !a.addAllowlist);
 
   let enabled = false;
+
+  if (noAllowlist) {
+    p.log.info(pc.dim("Skipping editor auto-approve allowlists (--no-allowlist)."));
+    return { enabled, lines: [] };
+  }
 
   if (adaptersWithAllowlist.length > 0) {
     p.log.info(
--- a/packages/argent-installer/src/init-allowlist.ts
+++ b/packages/argent-installer/src/init-allowlist.ts
@@ -5,6 +5,8 @@ import { InitCancelled } from "./init-args.js";
 
 interface AllowlistResult {
   enabled: boolean;
+  /** How `enabled` was reached: --no-allowlist, the -y default, or an answer. */
+  decidedBy: "flag" | "default" | "prompt";
   lines: string[];
 }
 
@@ -22,10 +24,11 @@ export async function configureAllowlist(args: {
   const adaptersWithoutAllowlist = adapters.filter((a) => !a.addAllowlist);
 
   let enabled = false;
+  let decidedBy: AllowlistResult["decidedBy"] = "default";
 
   if (noAllowlist) {
     p.log.info(pc.dim("Skipping editor auto-approve allowlists (--no-allowlist)."));
-    return { enabled, lines: [] };
+    return { enabled, decidedBy: "flag", lines: [] };
   }
 
   if (adaptersWithAllowlist.length > 0) {
@@ -47,10 +50,11 @@ export async function configureAllowlist(args: {
 
       if (p.isCancel(allowlistChoice)) throw new InitCancelled("allowlist");
       enabled = allowlistChoice as boolean;
+      decidedBy = "prompt";
     }
   }
 
-  if (!enabled) return { enabled, lines: [] };
+  if (!enabled) return { enabled, decidedBy, lines: [] };
 
   const lines: string[] = [];
 
@@ -78,5 +82,5 @@ export async function configureAllowlist(args: {
     );
   }
 
-  return { enabled, lines };
+  return { enabled, decidedBy, lines };
 }
--- a/packages/argent-installer/src/init.ts
+++ b/packages/argent-installer/src/init.ts
@@ -246,7 +246,10 @@ export async function init(args: string[]): Promise<void> {
       nonInteractive: parsed.nonInteractive,
       noAllowlist: parsed.noAllowlist,
     });
-    track("installation:allowlist_decision", { is_enabled: allowlist.enabled });
+    track("installation:allowlist_decision", {
+      is_enabled: allowlist.enabled,
+      decided_by: allowlist.decidedBy,
+    });
     if (allowlist.enabled && allowlist.lines.length > 0) {
       p.note(allowlist.lines.join("\n"), "Tool Auto-Approval");
     }
--- a/packages/argent-installer/src/init.ts
+++ b/packages/argent-installer/src/init.ts
@@ -244,6 +244,7 @@ export async function init(args: string[]): Promise<void> {
       effectiveRoot,
       scope: normalizedScope,
       nonInteractive: parsed.nonInteractive,
+      noAllowlist: parsed.noAllowlist,
     });
     track("installation:allowlist_decision", { is_enabled: allowlist.enabled });
     if (allowlist.enabled && allowlist.lines.length > 0) {
--- a/packages/argent-installer/src/mcp-configs.ts
+++ b/packages/argent-installer/src/mcp-configs.ts
@@ -42,6 +42,28 @@ function getAvailableToolIds(): string[] {
 
 // The exact opted-in shape Windsurf's alwaysAllow / Kiro's autoApprove get:
 // anything else there is a value the user narrowed since.
+// `update` rewrites the argent MCP entry on every run (to repair drift), but
+// only command/args/env are argent's to author. Editors keep state INSIDE the
+// entry too — Gemini's `trust`, Kiro's `autoApprove`, Windsurf's `alwaysAllow`,
+// Codex's `tools` table, a user-set `timeout`/`disabled` — and a wholesale
+// replace drops all of it. Dropping the opt-in keys used to be masked by the
+// unconditional allowlist re-add that followed; update's refresh guard has
+// retired that re-add, so the rewrite itself must carry them over.
+function mergeServerEntry(existing: unknown, entry: McpServerEntry): Record<string, unknown> {
+  const kept: Record<string, unknown> = {};
+  if (existing && typeof existing === "object" && !Array.isArray(existing)) {
+    for (const [key, value] of Object.entries(existing as Record<string, unknown>)) {
+      if (key !== "command" && key !== "args" && key !== "env") kept[key] = value;
+    }
+  }
+  return {
+    command: entry.command,
+    args: entry.args,
+    ...(hasEnv(entry) ? { env: entry.env } : {}),
+    ...kept,
+  };
+}
+
 function isWildcardOnly(value: unknown): boolean {
   return Array.isArray(value) && value.length === 1 && value[0] === "*";
 }
@@ -1093,11 +1115,10 @@ const windsurfAdapter: McpConfigAdapter = {
 
   // JSONC-safe writes (see the Cursor adapter).
   write(configPath: string, entry: McpServerEntry): void {
-    editJsoncFile(configPath, ["mcpServers", MCP_SERVER_KEY], {
-      command: entry.command,
-      args: entry.args,
-      ...(hasEnv(entry) ? { env: entry.env } : {}),
-    });
+    const existing = (readJsonc(configPath).mcpServers as Record<string, unknown> | undefined)?.[
+      MCP_SERVER_KEY
+    ];
+    editJsoncFile(configPath, ["mcpServers", MCP_SERVER_KEY], mergeServerEntry(existing, entry));
   },
 
   remove(configPath: string): boolean {
@@ -1280,11 +1301,10 @@ const geminiAdapter: McpConfigAdapter = {
 
   // JSONC-safe writes (see the Cursor adapter).
   write(configPath: string, entry: McpServerEntry): void {
-    editJsoncFile(configPath, ["mcpServers", MCP_SERVER_KEY], {
-      command: entry.command,
-      args: entry.args,
-      ...(hasEnv(entry) ? { env: entry.env } : {}),
-    });
+    const existing = (readJsonc(configPath).mcpServers as Record<string, unknown> | undefined)?.[
+      MCP_SERVER_KEY
+    ];
+    editJsoncFile(configPath, ["mcpServers", MCP_SERVER_KEY], mergeServerEntry(existing, entry));
   },
 
   remove(configPath: string): boolean {
@@ -1371,11 +1391,7 @@ const codexAdapter: McpConfigAdapter = {
   write(configPath: string, entry: McpServerEntry): void {
     const config = readToml(configPath);
     const servers = (config.mcp_servers ?? {}) as Record<string, unknown>;
-    servers[MCP_SERVER_KEY] = {
-      command: entry.command,
-      args: entry.args,
-      ...(hasEnv(entry) ? { env: entry.env } : {}),
-    };
+    servers[MCP_SERVER_KEY] = mergeServerEntry(servers[MCP_SERVER_KEY], entry);
     config.mcp_servers = servers;
     writeToml(configPath, config);
   },
@@ -1660,11 +1676,10 @@ const kiroAdapter: McpConfigAdapter = {
   // Read and written as JSONC through readJsonc / editJsoncFile so comments and
   // foreign servers survive (see the Cursor adapter).
   write(configPath: string, entry: McpServerEntry): void {
-    editJsoncFile(configPath, ["mcpServers", MCP_SERVER_KEY], {
-      command: entry.command,
-      args: entry.args,
-      ...(hasEnv(entry) ? { env: entry.env } : {}),
-    });
+    const existing = (readJsonc(configPath).mcpServers as Record<string, unknown> | undefined)?.[
+      MCP_SERVER_KEY
+    ];
+    editJsoncFile(configPath, ["mcpServers", MCP_SERVER_KEY], mergeServerEntry(existing, entry));
   },
 
   remove(configPath: string): boolean {
--- a/packages/argent-installer/src/mcp-configs.ts
+++ b/packages/argent-installer/src/mcp-configs.ts
@@ -89,10 +89,13 @@ export interface McpConfigAdapter {
   // say so instead of claiming a write.
   //
   // `refresh: true` marks update's periodic re-run, as opposed to init's
-  // explicit opt-in. An adapter whose rule has side effects beyond argent may
-  // treat a rule missing from an existing allowlist as a deliberate removal
-  // and decline to re-add it on refresh. Only Cursor does today; the other
-  // adapters re-add on both paths.
+  // explicit opt-in. An adapter whose write can override a choice the user
+  // made since init — Cursor's file allowlist, Zed's global tool_permissions
+  // default, Codex's per-tool approval modes — declines to re-impose it on
+  // refresh; adapters whose entries are argent-scoped and inert keep re-adding
+  // on both paths. Absence-based checks cannot tell "removed on purpose" from
+  // "never added" (no consent is persisted anywhere), so refresh errs toward
+  // not writing; `argent init` is the explicit (re-)opt-in path.
   addAllowlist?(
     root: string,
     scope: "local" | "global",
@@ -729,11 +732,13 @@ const cursorAdapter: McpConfigAdapter = {
       return `skipped - an mcpAllowlist here would turn off Cursor's "Run Everything" mode and override its in-app allowlist`;
     }
     if (list.includes(CURSOR_ALLOWLIST_PATTERN)) return;
-    // On update's refresh, argent:* missing from a list the user maintains is
-    // a deliberate removal — re-adding it would override that choice on every
-    // update. Re-opting in is `argent init`'s allowlist step.
+    // On update's refresh, argent:* missing from a list the user maintains
+    // usually means they removed it — re-adding would override that choice on
+    // every update. (It can also mean they never opted in; absence cannot tell
+    // the two apart, so refresh errs toward not writing.) Opting in is
+    // `argent init`'s allowlist step.
     if (options?.refresh) {
-      return `skipped - argent:* was removed from this allowlist; re-add it via argent init`;
+      return `skipped - argent:* is not on this allowlist; opt in via argent init`;
     }
     editJsoncFile(permPath, ["mcpAllowlist"], [...list, CURSOR_ALLOWLIST_PATTERN]);
   },
@@ -1170,13 +1175,30 @@ const zedAdapter: McpConfigAdapter = {
     return this.getArgentEntry(configPath) !== null;
   },
 
-  // Zed has no server-level wildcard for MCP tools — each tool would need its
-  // own entry — so the global default is set to "allow" instead.
-  addAllowlist(root: string, scope: "local" | "global"): void {
+  // Zed doesn't support server-level wildcards for MCP tools — each tool
+  // would need its own entry.  Setting the global default to "allow" is the
+  // documented opt-in; built-in security rules still protect against
+  // destructive operations.
+  //
+  // That default governs EVERY agent tool in Zed, not just argent's, so
+  // update's refresh never re-imposes it: any other value is a choice the
+  // user made since init (or an opt-in that never happened), and overriding
+  // it on each update would be the Cursor run-mode bug in another skin.
+  addAllowlist(root, scope, options): string | void {
     const settingsPath =
       scope === "global"
         ? path.join(homedir(), ".config", "zed", "settings.json")
         : path.join(root, ".zed", "settings.json");
+    if (options?.refresh) {
+      const config = fs.existsSync(settingsPath) ? readJsonc(settingsPath) : {};
+      const perms = (config.agent as Record<string, unknown>)?.tool_permissions as
+        | Record<string, unknown>
+        | undefined;
+      if (perms?.default !== "allow") {
+        return `skipped - Zed's tool_permissions default is not "allow"; opt in via argent init`;
+      }
+      return;
+    }
     editJsoncFile(settingsPath, ["agent", "tool_permissions", "default"], "allow");
   },
 
@@ -1330,7 +1352,11 @@ const codexAdapter: McpConfigAdapter = {
     return this.getArgentEntry(configPath) !== null;
   },
 
-  addAllowlist(root, scope): void {
+  // On update's refresh the per-tool table is maintained, not re-imposed: ids
+  // a new version added are filled in, but an entry the user restricted keeps
+  // their approval_mode, and a table they emptied or removed stays gone —
+  // rewriting either would override a choice made since init.
+  addAllowlist(root, scope, options): string | void {
     const configPath = scope === "global" ? this.globalPath() : this.projectPath(root);
 
     if (!configPath) {
@@ -1339,6 +1365,22 @@ const codexAdapter: McpConfigAdapter = {
 
     const tools = getAvailableToolIds();
     const config = readToml(configPath) as CodexConfig;
+
+    if (options?.refresh) {
+      const toolsConfig = config.mcp_servers?.argent?.tools;
+      if (toolsConfig === undefined || Object.keys(toolsConfig).length === 0) {
+        return `skipped - no argent tools allowlist in config.toml; opt in via argent init`;
+      }
+      let added = false;
+      for (const tool of tools) {
+        if (!(tool in toolsConfig)) {
+          toolsConfig[tool] = { approval_mode: "approve" };
+          added = true;
+        }
+      }
+      if (added) writeToml(configPath, config);
+      return;
+    }
 
     config.mcp_servers ??= {};
     config.mcp_servers.argent ??= {};
--- a/packages/argent-installer/src/mcp-configs.ts
+++ b/packages/argent-installer/src/mcp-configs.ts
@@ -85,7 +85,9 @@ export interface McpConfigAdapter {
   // Report argent state outside the projectPath/globalPath pair that can keep
   // the entry written at `writtenScope` from taking effect.
   findShadowingConfigs?(root: string, writtenScope: "local" | "global"): ShadowingConfigFinding[];
-  addAllowlist?(root: string, scope: "local" | "global"): void;
+  // Returns a short note when the rule was deliberately NOT added (or an
+  // earlier one was undone) so init can say so instead of claiming a write.
+  addAllowlist?(root: string, scope: "local" | "global"): string | void;
   removeAllowlist?(root: string, scope: "local" | "global"): void;
 }
 
@@ -672,12 +674,50 @@ const cursorAdapter: McpConfigAdapter = {
     return this.getArgentEntry(configPath) !== null;
   },
 
-  // The allowlist lives in a separate ~/.cursor/permissions.json that may carry
-  // the user's own rules, so edit that one key instead of rewriting the file.
-  addAllowlist(): void {
+  // Cursor's allowlist lives in a separate ~/.cursor/permissions.json, which —
+  // like every Cursor config — is JSONC (comments, trailing commas) and may
+  // carry the user's own unrelated rules. Route through readJsonc / editJsoncFile
+  // so a hand-commented or multi-rule permissions.json survives: the old readJson
+  // path ran it through `catch { return {} }` and rewrote the whole file with
+  // only { mcpAllowlist }, dropping every foreign rule and comment. Newly matters
+  // because `hasArgentEntry` now reads mcp.json with readJsonc, so `update`
+  // detects a commented config as configured and calls this.
+  //
+  // APPEND-ONLY, NEVER CREATE. Cursor derives its Run Mode from this file:
+  // `shouldPermissionsFileConstrainUnrestrictedMode()` disables "Run Everything"
+  // whenever a non-empty mcpAllowlist/terminalAllowlist is present and no
+  // approvalMode overrides it, which the UI reports as
+  // `Run Mode (Enforced by ~/.cursor/permissions.json)`. The same non-emptiness
+  // makes the file SUPERSEDE the in-IDE MCP allowlist and freeze its editor
+  // (`getEffectiveMcpAllowlist` ignores composerState.mcpAllowedTools,
+  // `canAddToAllowlistFromIde("mcp")` goes false), so the user's own
+  // "Always allow" entries stop being consulted.
+  //
+  // Creating the file therefore costs a run mode the user chose — and buys
+  // nothing, because under "Run Everything" every tool auto-runs already and
+  // the allowlist is never read. So argent only appends where the user already
+  // keeps a file-based MCP allowlist: there both effects are their own doing
+  // and `argent:*` changes no state. Existing files are never deleted either:
+  // even an argent-only one may be an allowlist the user keeps deliberately.
+  //
+  // Setting `approvalMode: "unrestricted"` to lift the constraint is NOT an
+  // option — Cursor treats a permissions-file approvalMode as authoritative
+  // (`getModeFullAutoRun` returns true for it), so it would force Run
+  // Everything ON for users who chose "Ask every time".
+  addAllowlist(): string | void {
     const permPath = path.join(homedir(), ".cursor", "permissions.json");
+    if (!fs.existsSync(permPath)) {
+      return `skipped - creating ~/.cursor/permissions.json would turn off Cursor's "Run Everything" mode`;
+    }
+    // An existing file is left as the user has it, argent-only shape included:
+    // we cannot tell a file an earlier argent created from one the user keeps
+    // deliberately, and deleting the latter would take away an allowlist they
+    // want. Removing it stays a manual step (or `argent uninstall`).
     const config = readJsonc(permPath);
-    const list = Array.isArray(config.mcpAllowlist) ? (config.mcpAllowlist as string[]) : [];
+    const list = Array.isArray(config.mcpAllowlist) ? (config.mcpAllowlist as string[]) : null;
+    if (list === null || list.length === 0) {
+      return `skipped - an mcpAllowlist here would override Cursor's in-app allowlist`;
+    }
     if (list.includes(CURSOR_ALLOWLIST_PATTERN)) return;
     editJsoncFile(permPath, ["mcpAllowlist"], [...list, CURSOR_ALLOWLIST_PATTERN]);
   },
--- a/packages/argent-installer/src/mcp-configs.ts
+++ b/packages/argent-installer/src/mcp-configs.ts
@@ -87,7 +87,17 @@ export interface McpConfigAdapter {
   findShadowingConfigs?(root: string, writtenScope: "local" | "global"): ShadowingConfigFinding[];
   // Returns a short note when the rule was deliberately NOT added so init can
   // say so instead of claiming a write.
-  addAllowlist?(root: string, scope: "local" | "global"): string | void;
+  //
+  // `refresh: true` marks update's periodic re-run, as opposed to init's
+  // explicit opt-in. An adapter whose rule has side effects beyond argent may
+  // treat a rule missing from an existing allowlist as a deliberate removal
+  // and decline to re-add it on refresh. Only Cursor does today; the other
+  // adapters re-add on both paths.
+  addAllowlist?(
+    root: string,
+    scope: "local" | "global",
+    options?: { refresh?: boolean }
+  ): string | void;
   removeAllowlist?(root: string, scope: "local" | "global"): void;
 }
 
@@ -704,7 +714,7 @@ const cursorAdapter: McpConfigAdapter = {
   // option — Cursor treats a permissions-file approvalMode as authoritative
   // (`getModeFullAutoRun` returns true for it), so it would force Run
   // Everything ON for users who chose "Ask every time".
-  addAllowlist(): string | void {
+  addAllowlist(_root, _scope, options): string | void {
     const permPath = path.join(homedir(), ".cursor", "permissions.json");
     if (!fs.existsSync(permPath)) {
       return `skipped - creating ~/.cursor/permissions.json would turn off Cursor's "Run Everything" mode`;
@@ -719,6 +729,12 @@ const cursorAdapter: McpConfigAdapter = {
       return `skipped - an mcpAllowlist here would turn off Cursor's "Run Everything" mode and override its in-app allowlist`;
     }
     if (list.includes(CURSOR_ALLOWLIST_PATTERN)) return;
+    // On update's refresh, argent:* missing from a list the user maintains is
+    // a deliberate removal — re-adding it would override that choice on every
+    // update. Re-opting in is `argent init`'s allowlist step.
+    if (options?.refresh) {
+      return `skipped - argent:* was removed from this allowlist; re-add it via argent init`;
+    }
     editJsoncFile(permPath, ["mcpAllowlist"], [...list, CURSOR_ALLOWLIST_PATTERN]);
   },
 
--- a/packages/argent-installer/src/mcp-configs.ts
+++ b/packages/argent-installer/src/mcp-configs.ts
@@ -85,8 +85,8 @@ export interface McpConfigAdapter {
   // Report argent state outside the projectPath/globalPath pair that can keep
   // the entry written at `writtenScope` from taking effect.
   findShadowingConfigs?(root: string, writtenScope: "local" | "global"): ShadowingConfigFinding[];
-  // Returns a short note when the rule was deliberately NOT added (or an
-  // earlier one was undone) so init can say so instead of claiming a write.
+  // Returns a short note when the rule was deliberately NOT added so init can
+  // say so instead of claiming a write.
   addAllowlist?(root: string, scope: "local" | "global"): string | void;
   removeAllowlist?(root: string, scope: "local" | "global"): void;
 }
@@ -716,7 +716,7 @@ const cursorAdapter: McpConfigAdapter = {
     const config = readJsonc(permPath);
     const list = Array.isArray(config.mcpAllowlist) ? (config.mcpAllowlist as string[]) : null;
     if (list === null || list.length === 0) {
-      return `skipped - an mcpAllowlist here would override Cursor's in-app allowlist`;
+      return `skipped - an mcpAllowlist here would turn off Cursor's "Run Everything" mode and override its in-app allowlist`;
     }
     if (list.includes(CURSOR_ALLOWLIST_PATTERN)) return;
     editJsoncFile(permPath, ["mcpAllowlist"], [...list, CURSOR_ALLOWLIST_PATTERN]);
--- a/packages/argent-installer/src/mcp-configs.ts
+++ b/packages/argent-installer/src/mcp-configs.ts
@@ -40,6 +40,13 @@ function getAvailableToolIds(): string[] {
   return tools.map((t) => t.id);
 }
 
+// The exact opted-in shape Windsurf's alwaysAllow / Kiro's autoApprove get:
+// anything else there is a value the user narrowed since.
+function isWildcardOnly(value: unknown): boolean {
+  return Array.isArray(value) && value.length === 1 && value[0] === "*";
+}
+
+// ── Types ─────────────────────────────────────────────────────────────────────
 // MARK: Types
 
 export interface McpServerEntry {
@@ -89,13 +96,17 @@ export interface McpConfigAdapter {
   // say so instead of claiming a write.
   //
   // `refresh: true` marks update's periodic re-run, as opposed to init's
-  // explicit opt-in. An adapter whose write can override a choice the user
-  // made since init — Cursor's file allowlist, Zed's global tool_permissions
-  // default, Codex's per-tool approval modes — declines to re-impose it on
-  // refresh; adapters whose entries are argent-scoped and inert keep re-adding
-  // on both paths. Absence-based checks cannot tell "removed on purpose" from
-  // "never added" (no consent is persisted anywhere), so refresh errs toward
-  // not writing; `argent init` is the explicit (re-)opt-in path.
+  // explicit opt-in. Refresh never changes the effective approval state the
+  // user has: when an adapter's opted-in value is absent or was changed since
+  // init, it skips with a note — re-imposing the value would override that
+  // choice (the Cursor run-mode bug in another skin). The one write refresh
+  // still performs is Codex's backfill of tool ids a new version added, for
+  // users whose table shows they are opted in. Absence cannot tell "removed
+  // on purpose" from "never added" (no consent is persisted anywhere), so
+  // refresh errs toward not writing; `argent init` is the explicit
+  // (re-)opt-in path. The option is advisory — an adapter that ignores it
+  // re-adds on both paths — so a new adapter must decide its refresh
+  // behavior explicitly.
   addAllowlist?(
     root: string,
     scope: "local" | "global",
@@ -949,7 +960,15 @@ const claudeAdapter: McpConfigAdapter = {
     return findings;
   },
 
-  addAllowlist(root: string, scope: "local" | "global"): void {
+  // On refresh there is nothing to maintain — the rule is a single
+  // argent-scoped entry — so a missing rule is only reported: re-adding it
+  // (and creating settings.json for a user who declined at init) would
+  // override a removal.
+  addAllowlist(root, scope, options): string | void {
+    if (options?.refresh) {
+      if (hasClaudePermission(root, scope)) return;
+      return `skipped - the argent permission rule is not in settings.json; opt in via argent init`;
+    }
     addClaudePermission(root, scope);
   },
 
@@ -1101,13 +1120,23 @@ const windsurfAdapter: McpConfigAdapter = {
     return this.getArgentEntry(configPath) !== null;
   },
 
-  // Targets just the argent entry's alwaysAllow key, so comments and foreign
-  // servers survive.
-  addAllowlist(): void {
+  // JSONC-safe allowlist edits (see the Cursor adapter): the argent entry lives
+  // in this same mcp_config.json, and `init` runs write() (comment-preserving)
+  // before addAllowlist(). The old readJson path choked on any user comment and
+  // silently skipped the toggle; editJsoncFile targets just the argent entry's
+  // alwaysAllow key so comments and foreign servers survive.
+  addAllowlist(_root, _scope, options): string | void {
     const configPath = path.join(homedir(), ".codeium", "windsurf", "mcp_config.json");
     const config = readJsonc(configPath);
     const servers = config.mcpServers as Record<string, unknown> | undefined;
     if (!servers?.[MCP_SERVER_KEY]) return;
+    // A user-narrowed or removed alwaysAllow is their choice — rewriting it
+    // to ["*"] on refresh would override it.
+    if (options?.refresh) {
+      const entry = servers[MCP_SERVER_KEY] as Record<string, unknown>;
+      if (isWildcardOnly(entry.alwaysAllow)) return;
+      return `skipped - argent's alwaysAllow is not ["*"]; opt in via argent init`;
+    }
     editJsoncFile(configPath, ["mcpServers", MCP_SERVER_KEY, "alwaysAllow"], ["*"]);
   },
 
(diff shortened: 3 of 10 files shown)


packages/argent-installer/src/init-allowlist.ts
```diff
@@ -56,8 +56,12 @@ export async function configureAllowlist(args: {
       continue;
     }
     try {
-      adapter.addAllowlist!(effectiveRoot, scope);
-      lines.push(`${pc.green("+")} ${adapter.name}`);
+      const note = adapter.addAllowlist!(effectiveRoot, scope);
```
latekvo, line 68: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: `enabled` is set before this loop from `adaptersWithAllowlist.length > 0` plus the confirm/nonInteractive decision, so when Cursor is the only allowlist-capable adapter and its addAllowlist returns the new skip note (permissions.json absent, or a missing/empty mcpAllowlist), this line's note renders but no write happens while `enabled` stays true. Replayed end-to-end: with a fresh HOME, only the Cursor adapter, and nonInteractive=true, configureAllowlist returns `enabled: true`, prints the `- Cursor (skipped ...)` line, and creates nothing. That false `enabled` is what init.ts:253 records via track('installation:allowlist_decision', { is_enabled: true }), and the nonInteractive path (init-allowlist.ts:33-34) sets it true unconditionally with no human to catch the mismatch - so the analytics log the allowlist as enabled for an outcome that did not happen and, under Run Everything, never could.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -93,7 +93,9 @@ export interface McpConfigAdapter {
   // pair that the client resolves ahead of (or gates) the entry written at
   // `writtenScope`. Only clients with hidden scopes implement this.
   findShadowingConfigs?(root: string, writtenScope: "local" | "global"): ShadowingConfigFinding[];
-  addAllowlist?(root: string, scope: "local" | "global"): void;
+  // Returns a short note when the rule was deliberately NOT added (or an
+  // earlier one was undone) so init can say so instead of claiming a write.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: This interface doc promises the note also fires when 'an earlier one [was] undone'. None of the eight addAllowlist implementations (Cursor, Windsurf, Zed, Gemini, Claude, Codex, OpenCode, Kiro) ever removes or undoes a rule - all are strictly additive, and undoing is exclusively removeAllowlist's job. The Cursor implementation is the only one that returns a note, and it returns it only for choosing not to add. A future implementer extending this interface against the documented contract would be told a return shape no current adapter produces.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -723,10 +725,42 @@ const cursorAdapter: McpConfigAdapter = {
...
+    }
+    // An existing file is left as the user has it, argent-only shape included:
+    // we cannot tell a file an earlier argent created from one the user keeps
+    // deliberately, and deleting the latter would take away an allowlist they
+    // want. Removing it stays a manual step (or `argent uninstall`).
     const config = readJsonc(permPath);
-    const list = Array.isArray(config.mcpAllowlist) ? (config.mcpAllowlist as string[]) : [];
+    const list = Array.isArray(config.mcpAllowlist) ? (config.mcpAllowlist as string[]) : null;
+    if (list === null || list.length === 0) {
+      return `skipped - an mcpAllowlist here would override Cursor's in-app allowlist`;
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: This branch fires for `list === null || list.length === 0`, which includes a permissions.json that exists but is empty ({}). An empty file does not currently constrain Cursor's run mode, so adding an mcpAllowlist there would also bring on the primary harm the file-absent branch (line 753) names - shouldPermissionsFileConstrainUnrestrictedMode demoting Run Everything - yet this message mentions only the in-app-allowlist override. A user with an empty file and Run Everything set sees a note that omits the main consequence.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -93,7 +93,9 @@ export interface McpConfigAdapter {
   // pair that the client resolves ahead of (or gates) the entry written at
   // `writtenScope`. Only clients with hidden scopes implement this.
   findShadowingConfigs?(root: string, writtenScope: "local" | "global"): ShadowingConfigFinding[];
-  addAllowlist?(root: string, scope: "local" | "global"): void;
+  // Returns a short note when the rule was deliberately NOT added (or an
+  // earlier one was undone) so init can say so instead of claiming a write.
+  addAllowlist?(root: string, scope: "local" | "global"): string | void;
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: The new `string | void` return is documented as 'so init can say so', but update.ts:750 calls addAllowlist and discards the value. On an `argent update` where Cursor's permissions.json is absent or empty - a machine that a pre-change build would have created the file for - the never-create skip is completely silent; a user refreshing allowlists (the stated intent at update.ts:744) gets no signal that the Cursor rule was not (re)added, even though the SKIP is now the default outcome for exactly this population.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Deepseek V4 Flash 0731\]: Ran the full sweep across the changed lines and their call sites: claims vs code, nearest twin (add vs remove vs sibling adapters), non-happy paths, inputs, reachability both ways, lifetime, plus the absence pass (sibling / prose / symmetry / mutation). Replayed configureAllowlist end-to-end through a sandboxed HOME with only the Cursor adapter; replayed the package suite (558 tests pass, eslint --max-warnings 0 clean; tsc --build is blocked by type-drift from a cross-branch symlinked node_modules, not this PR). Returned one medium and three low findings below.

packages/argent-installer/src/mcp-configs.ts
filip131311: Fixed in 88063b9e7 — the note now names both effects: `skipped - an mcpAllowlist here would turn off Cursor's "Run Everything" mode and override its in-app allowlist`. One nuance: when the file already carries a non-empty `terminalAllowlist`, the run mode is already constrained and only the override half is new — but a single combined message is accurate for the common case and never understates the harm.

packages/argent-installer/src/mcp-configs.ts
filip131311: Fixed in 88063b9e7 — dropped the "(or an earlier one was undone)" clause. Correct: nothing in any adapter undoes rules from addAllowlist; that's removeAllowlist's job, and the doc shouldn't promise a return shape no implementation produces.

packages/argent-installer/src/init-allowlist.ts
filip131311: Accurate chain, but not new semantics from this PR: the event is named allowlist_**decision** and records the user's choice (or the nonInteractive default), not per-adapter write outcomes. `enabled: true` alongside a skipped adapter was already possible before this change via the `(no config for this scope)` branch a few lines up — decision ≠ outcome has always been the contract here. If we want outcome-level analytics, that's a separate per-adapter event, out of scope for this fix. Not changing.

packages/argent-installer/src/mcp-configs.ts
filip131311: True, but consistent with update's existing reporting: the allowlist-refresh loop in update.ts has never surfaced per-adapter results — successes print nothing and errors are swallowed with `catch { /* non-fatal */ }`. The skip is exactly as silent as a success always was there. Surfacing allowlist outcomes on `argent update` would be a reasonable enhancement to that loop as a whole (all eight adapters, success and error included), but it's a separate change to update's reporting pattern, not a gap this PR introduces. Not changing here.

filip131311:
@latekvo i need actuall review

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -672,13 +720,59 @@ const cursorAdapter: McpConfigAdapter = {
...
+  // even an argent-only one may be an allowlist the user keeps deliberately.
+  //
+  // Setting `approvalMode: "unrestricted"` to lift the constraint is NOT an
+  // option — Cursor treats a permissions-file approvalMode as authoritative
+  // (`getModeFullAutoRun` returns true for it), so it would force Run
+  // Everything ON for users who chose "Ask every time".
+  addAllowlist(_root, _scope, options): string | void {
     const permPath = path.join(homedir(), ".cursor", "permissions.json");
+    if (!fs.existsSync(permPath)) {
+      return `skipped - creating ~/.cursor/permissions.json would turn off Cursor's "Run Everything" mode`;
```
latekvo, line 756: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The skip fires for every user who has no `~/.cursor/permissions.json`, but the reasoning for it - that under Run Everything the allowlist is never consulted, so the rule was already a no-op - only holds for users who actually have Run Everything on.

From the shipped Cursor bundle (3.7.36), the mode is picked as `fullAutoRun && !shouldPermissionsFileConstrainUnrestrictedMode() ? "unrestricted" : autoRun ? "allowlist" : "ask-every-time"`, and in the `"allowlist"` arm the permissions handed to the agent include `getEffectiveMcpAllowlist().map(t => "Mcp(" + t + ")")`. That resolver (`ogw`, non-admin branch) returns the permissions-file list whenever it is non-empty and the in-IDE list otherwise. So for a user with Auto-run on and Run Everything off, this file was the only thing that made argent's tools auto-run - nothing writes `argent:*` into the in-IDE list.

Same input through both builds, sandboxed `HOME`, `~/.cursor/mcp.json` present:

```
main    addAllowlist returned: undefined
        ~/.cursor now contains: [ 'mcp.json', 'permissions.json' ]
PR 787  addAllowlist returned: "skipped - creating ~/.cursor/permissions.json would
                                turn off Cursor's \"Run Everything\" mode"
        ~/.cursor now contains: [ 'mcp.json' ]
```

That user answered yes to `Add Argent tools to editor auto-approve lists? - recommended`, ends up with no auto-approval on Cursor, and the reason printed names a run mode they do not have.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -1130,11 +1274,7 @@ const zedAdapter: McpConfigAdapter = {
         ? path.join(homedir(), ".config", "zed", "settings.json")
         : path.join(root, ".zed", "settings.json");
     if (!fs.existsSync(settingsPath)) return;
-    const config = readJsonc(settingsPath);
-    const perms = (config.agent as Record<string, unknown>)?.tool_permissions as
-      | Record<string, unknown>
-      | undefined;
-    if (!perms || perms.default !== "allow") return;
+    if (zedToolPermissionsDefault(settingsPath) !== "allow") return;
```
latekvo, line 1277: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The refresh guard twenty lines above declines to write this key on the grounds that it "governs EVERY agent tool in Zed, not just argent's" and that there is no record of who set it. `removeAllowlist` keys on the same value with no such record, and `uninstall.ts:461-472` runs it across `ALL_ADAPTERS` with no check that argent was ever configured for that adapter or scope.

Reproduced against this head's build, sandboxed `HOME`, on a settings file with no `context_servers.argent` anywhere in it:

```
argent configured in this Zed settings file? false
--- before ---
{
  // I want my agent tools to run without confirmation
  "theme": "One Dark",
  "agent": { "tool_permissions": { "default": "allow" } }
}
--- after `argent uninstall` ---
{
  // I want my agent tools to run without confirmation
  "theme": "One Dark",
  "agent": { "tool_permissions": { "default": "confirm" } }
}
```

`argent uninstall` turns confirmation back on for every agent tool that Zed user has, in a run where argent never set the value.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -888,7 +982,15 @@ const claudeAdapter: McpConfigAdapter = {
...
 
-  addAllowlist(root: string, scope: "local" | "global"): void {
+  // On refresh there is nothing to maintain — the rule is a single
+  // argent-scoped entry — so a missing rule is only reported: re-adding it
+  // (and creating settings.json for a user who declined at init) would
+  // override a removal.
+  addAllowlist(root, scope, options): string | void {
+    if (options?.refresh) {
+      if (hasClaudePermission(root, scope)) return;
+      return `skipped - the argent permission rule is not in settings.json; opt in via argent init`;
```
latekvo, line 992: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `hasClaudePermission` reads one file per scope and never the other. Claude Code merges `~/.claude/settings.json` into every project and permission lists merge rather than override - that merge is what makes `addClaudePermission(root, "global")` effective at all. The Zed guard in this same diff consults the global file before concluding the user is not opted in; this one does not, so a local-scope refresh reports a skip for a user who is opted in globally.

Reproduced on this head's build, sandboxed `HOME`, opted in globally, project carrying a committed `.mcp.json`:

```
~/.claude/settings.json = { "permissions": { "allow": [ "mcp__argent" ] } }
<proj>/.claude/settings.json exists? false
scopes update will refresh for Claude Code: [ 'project' ]
addAllowlist(proj, "global", {refresh:true}) -> undefined
addAllowlist(proj, "local",  {refresh:true}) -> "skipped - the argent permission rule is not
                                                in settings.json; opt in via argent init"
Zed  addAllowlist(proj, "local", {refresh:true}) -> undefined  (no local .zed/settings.json)
```

Every `argent update` in that project tells a user who is auto-approved that the rule is missing, and sends them back to `argent init`.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -1274,28 +1418,46 @@ const codexAdapter: McpConfigAdapter = {
...
 
-    for (const tool of tools) {
-      toolsConfig[tool] = {
-        approval_mode: "approve",
-      };
+    let changed = false;
+    for (const tool of getAvailableToolIds()) {
+      // Refresh only backfills ids the table doesn't know; init's explicit
+      // opt-in (re)writes them all.
+      if (refresh && tool in toolsConfig) continue;
```
latekvo, line 1456: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: An id absent from the table is read as one a new version added, but it is equally the shape left by a user who deleted an entry to stop auto-approving that tool. The interface doc this implements states that refresh "never changes the effective approval state the user has" and "errs toward not writing", and the block comment above enumerates the protected cases - restricted entry, emptied table, removed table, all-deny - without this one. The write also returns no note, so nothing in update's output accounts for it.

Reproduced on this head's build, sandboxed `HOME`:

```
tool ids after init opt-in: 4 -> screenshot, tap, type-text, describe
user deleted the "tap" entry; present now? false
refresh note: undefined
"tap" back, as approve? true
```

The sibling adapters treat exactly this gesture as opt-out: Windsurf and Kiro require an exact `["*"]` and skip on any narrowing.

packages/argent-installer/src/update.ts
```diff
@@ -721,16 +722,32 @@ export async function update(args: string[]): Promise<void> {
...
+        p.log.info(pc.dim("Skipped editor auto-approve allowlist refresh (--no-allowlist)."));
+      } else {
+        // Set-dedup: an adapter configured in both scopes (Cursor's allowlist
+        // is machine-global regardless) would otherwise note twice.
+        const allowlistNotes = new Set<string>();
+        for (const [scope, adapters] of adaptersByScope) {
+          for (const adapter of adapters) {
+            if (!adapter.addAllowlist) continue;
+            try {
+              const note = adapter.addAllowlist(projectRoot, scope, { refresh: true });
```
latekvo, line 741: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This is the only call site that makes update's refresh semantics take effect, and nothing in the suite discriminates it. Baseline in a clean checkout of this head is 576/576; dropping the option here leaves it at 576/576:

```
=== BASELINE ===   Tests  576 passed (576)
741:  const note = adapter.addAllowlist(projectRoot, scope);   // { refresh: true } removed
=== MUTATED  ===   Tests  576 passed (576)
```

All nine new refresh tests call `addAllowlist(..., { refresh: true })` directly, so none of them can observe the production caller dropping it. With the option gone, `argent update` goes back to re-adding `argent:*` to a Cursor allowlist the user cleaned, re-setting Zed's global default, re-imposing the Windsurf/Kiro wildcards, Gemini's `trust`, opencode's entry and Claude's rule - the behaviour this PR exists to stop.

`--no-allowlist` is in the same position on both commands:

```
src/update.ts:731:        if (false) {     // the --no-allowlist branch
src/init-allowlist.ts:29:  if (false) {    // the --no-allowlist branch
=== MUTATED  ===   Tests  576 passed (576)
```

`installer-flags-sync.test.ts` pins the spelling against the help text by scraping the parser source as text, and `init-args.test.ts` pins that the boolean is parsed, but nothing exercises what either flag does.

packages/argent-installer/src/init-allowlist.ts
```diff
@@ -54,8 +65,12 @@ export async function configureAllowlist(args: {
...
+      lines.push(
+        note
+          ? `${pc.yellow("-")} ${adapter.name} ${pc.dim(`(${note})`)}`
+          : `${pc.green("+")} ${adapter.name}`
```
latekvo, line 72: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The falsy branch renders the green `+ <adapter>` line, which reads as "added". Three adapters reach it having written nothing: Windsurf (`mcp-configs.ts:1153`), Gemini (`:1344`) and Kiro (`:1715`) each early-return on `if (!servers?.[MCP_SERVER_KEY]) return;`, ahead of their refresh guards. That exit is reachable at init whenever the entry write failed, since `init-mcp-write.ts` returns the selected adapters rather than the successfully written ones.

Driving the real `write()` then `configureAllowlist` on this head's build, with a hand-mangled non-object JSON root:

```
Gemini write THREW: Can not add index to parent of type array
Kiro   write THREW: Can not add index to parent of type null
note-line: "+ Gemini"
note-line: "+ Kiro"
Gemini file afterwards: "[]\n"
Kiro   file afterwards: "null\n"
```

The contract added at `mcp-configs.ts:115` says the return value exists so init "can say so instead of claiming a write" - Zed, Claude Code, opencode and Codex propagate the error and Cursor returns a note, leaving these three as the only exits where the rule is not added and nothing comes back.

packages/argent-installer/src/update.ts
```diff
@@ -721,16 +722,32 @@ export async function update(args: string[]): Promise<void> {
...
+              if (note) allowlistNotes.add(`- ${adapter.name}: ${note}`);
+            } catch {
+              // non-fatal
+            }
```
latekvo, line 745: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Two gaps against the stated purpose of this block ("so a missing auto-approve entry after an update is explainable from the output"):

The dedup key drops the `scope` that is in hand one line above, and the notes themselves name a bare filename (`settings.json`, `config.toml`) that six of the eight adapters resolve differently per scope. With Claude Code configured in both scopes, opted in locally but not globally, opted in globally but not locally, and opted out in both all print the same single line - and the `Set` then collapses the third case, where two files need attention, into one. The justification given ("Cursor's allowlist is machine-global regardless") holds for Cursor and Windsurf; Claude Code, Codex, Gemini, opencode, Kiro and Zed all read a different file per scope and share the same dedup.

The `catch` is the third way an entry goes missing and it leaves no line at all, while success prints nothing either - so a swallowed failure is byte-identical to a clean run. Codex is the reachable case, being the only adapter that still writes on refresh: with a read-only `~/.codex/config.toml` and ids to backfill, the update printed nothing and the `EACCES` was swallowed. init's sibling loop reports the same failure as `x <adapter>: <err>`.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -1040,13 +1141,23 @@ const windsurfAdapter: McpConfigAdapter = {
...
   },
 
-  // Targets just the argent entry's alwaysAllow key, so comments and foreign
-  // servers survive.
-  addAllowlist(): void {
+  // JSONC-safe allowlist edits (see the Cursor adapter): the argent entry lives
+  // in this same mcp_config.json, and `init` runs write() (comment-preserving)
+  // before addAllowlist(). The old readJson path choked on any user comment and
+  // silently skipped the toggle; editJsoncFile targets just the argent entry's
+  // alwaysAllow key so comments and foreign servers survive.
```
latekvo, line 1148: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This block, and four others in this file (`:723-730` Cursor, `:1239-1242` Zed, `:1330-1334` Gemini, `:1705-1709` Kiro), restore the verbose pre-audit comments that `b23211442` ("docs: audit every comment in src against the code it describes", #931) removed on `main`, which is an ancestor of this branch. 27 lines that #931 deleted are present again at this head:

```
$ git show b23211442 -- .../mcp-configs.ts | grep "^-  //" | sed 's/^-//' > /tmp/r931.txt
$ while read -r l; do grep -qxF "$l" .../mcp-configs.ts && echo "$l"; done < /tmp/r931.txt
  ... TOTAL RE-ADDED: 27
```

They carry the transition narration that audit existed to remove - "the old readJson path ran it through `catch { return {} }`", "Newly matters because `hasArgentEntry` now reads mcp.json with readJsonc", "The old readJson path choked on any user comment" (three times). Read with the diff hidden, they describe a code path that has not existed for several releases. Four of the five blocks sit on adapters whose `addAllowlist` changed only by gaining an `options` parameter. The Zed one also preserves the pre-audit double space in `own entry.  Setting`.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -1066,6 +1177,17 @@ const windsurfAdapter: McpConfigAdapter = {
 
 // MARK: Zed
 
+// The one place Zed's agent.tool_permissions.default is read, so the opt-in
+// schema can't drift between addAllowlist's refresh guard and removeAllowlist.
```
latekvo, line 1181: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: "The one place ... is read" is not accurate, and the drift the helper is introduced to prevent is exactly what the third reader will do. `zedSettingsLooksArgentOnly` at `:534-546` also reads `agent.tool_permissions` and its `default`, with its own inline copy of the accepted shape:

```
$ grep -rn "tool_permissions" packages/argent-installer/src
520, 538, 539, 1180, 1185, 1264, 1268, 1278
```

So the helper unifies `addAllowlist`'s guard with `removeAllowlist`, and the reader it leaves out is the one that decides whether a Zed settings file counts as argent-only for `detect()`.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -40,6 +40,35 @@ function getAvailableToolIds(): string[] {
   return tools.map((t) => t.id);
 }
 
+// The exact opted-in shape Windsurf's alwaysAllow / Kiro's autoApprove get:
+// anything else there is a value the user narrowed since.
```
latekvo, line 44: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: These two lines documented `isWildcardOnly` when `2d25c90f1` added them; `7739795b0` then inserted `mergeServerEntry` between the comment and its function. As it now stands, the sentence about "the exact opted-in shape Windsurf's alwaysAllow / Kiro's autoApprove get" heads `mergeServerEntry`, which keeps every key other than `command`/`args`/`env` regardless of shape and has nothing to do with wildcards, while `isWildcardOnly` at `:67` is left undocumented. `git log -L 43,72:packages/argent-installer/src/mcp-configs.ts` shows the comment arriving with `isWildcardOnly` in one commit and being separated from it in the next.

The same hunk also adds `// -- Types --` at `:71` directly above the pre-existing `// MARK: Types`, where every other section in this file uses the bare `// MARK:` form.

packages/argent-installer/test/mcp-configs.test.ts
```diff
@@ -2282,12 +2468,101 @@ describe("installer preserves foreign MCP config", () => {
     const cursor = ALL_ADAPTERS.find((a) => a.name === "Cursor")!;
     homedirOverride = fs.mkdtempSync(path.join(os.tmpdir(), "argent-fc-home-"));
     const permPath = path.join(homedirOverride, ".cursor", "permissions.json");
-    cursor.addAllowlist!(tmpDir, "global"); // fresh create: { mcpAllowlist: ["argent:*"] }
+    fs.mkdirSync(path.dirname(permPath), { recursive: true });
+    // As argent <= 0.20.x created it before addAllowlist stopped creating files.
```
latekvo, line 2472: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The version range is not right: every released version creates the file, current one included. `git show v0.20.0/v0.21.0/v0.22.0/v0.22.1:packages/argent-installer/src/mcp-configs.ts` all contain the same unguarded `editJsoncFile(permPath, ["mcpAllowlist"], [...list, CURSOR_ALLOWLIST_PATTERN])`, and `packages/argent/package.json` is at `0.22.1`. A reader sizing the affected population - the description notes that already-demoted machines are not repaired - would exclude two whole minor lines that are in fact affected.

The trailing clause is also transition narration in a permanent comment: read later, "before addAllowlist stopped creating files" describes a change rather than the state the fixture is in.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Ran the full sweep over the five commits at `7739795b0`: claims vs code, nearest twin (add vs remove, adapter vs adapter, init vs update), non-happy paths, inputs, reachability both ways, lifetime, plus the absence pass - a sibling that has it, prose that promises it, spawn/reap symmetry, and mutation across the new branches, constants and call sites.

Baseline in a clean worktree of the head: 576 installer, 80 argent and 314 telemetry tests pass; `tsc --build`, eslint and both knip gates clean.

I checked the diagnosis before reviewing the fix, against the Cursor bundle on this machine (3.7.36, `out/vs/workbench/workbench.desktop.main.js`). Both effects the description names are real: `shouldPermissionsFileConstrainUnrestrictedMode()` returns true purely on `mcpAllowlist.length > 0 || terminalAllowlist.length > 0` when no `approvalMode` is set, and `canAddToAllowlistFromIde("mcp")` returns `this._permissionsFileAllowlists.mcp.length === 0`. This machine had the argent-created `permissions.json` in the wild.

Every item below was reproduced against the head's own build under a sandboxed `HOME` before being written up. Mutation results are from a clean checkout of the head, one edit at a time, reverted after each. Findings inline.

packages/argent-installer/src/init-allowlist.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Still present as of `7739795b0`, and worth a second look because this change moves it from an edge case to the default one.

The point that `allowlist_decision` records a decision rather than an outcome is fair for the telemetry event. But the same value drives a user-facing string: `init.ts:362` renders `Auto-approve enabled` in the Summary box off `allowlist.enabled`. On the now-standard Cursor path the two appear in the same terminal:

```
Tool Auto-Approval
  - Cursor (skipped - creating ~/.cursor/permissions.json would turn off Cursor's "Run Everything" mode)
...
Summary
  Auto-approve enabled
```

Before this PR the `(no config for this scope)` branch you cite made that pairing rare. After it, a machine with no pre-existing `~/.cursor/permissions.json` - which is every machine argent has not already written one on - takes the skip every time, so "enabled" alongside a skipped adapter becomes the common output rather than the exception.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -40,6 +40,35 @@ function getAvailableToolIds(): string[] {
...
+    for (const [key, value] of Object.entries(existing as Record<string, unknown>)) {
+      if (key !== "command" && key !== "args" && key !== "env") kept[key] = value;
+    }
+  }
+  return {
+    command: entry.command,
+    args: entry.args,
+    ...(hasEnv(entry) ? { env: entry.env } : {}),
+    ...kept,
+  };
```
latekvo, line 64: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `kept` carries every key that is not `command`/`args`/`env`, and this return then adds `command` and `args` unconditionally. When the existing `[mcp_servers.argent]` is a streamable-HTTP entry, the result holds `command` and `url` in the same table, which Codex rejects at config-load time - not just the argent server, the whole `config.toml`.

The path is reachable on `argent update` precisely because of the repair branch: `normalizeServerEntry` returns the unreadable sentinel `{command:"",args:[]}` for such an entry, so `update.ts:701`'s `isUnreadableSentinel` check deliberately bypasses the "customized entry untouched" guard and calls `write()`. `argent init` reaches the same `write()` with no classification guard at all.

Reproduced with the real `codex-cli 0.142.5`, entry created by Codex itself:

```
$ codex mcp add argent --url https://mcp.example.com/sse
$ codex mcp list
argent  https://mcp.example.com/sse  -  enabled  Unsupported

# this head's write(), via update.ts's own classification
RESULT getArgentEntry: {"command":"","args":[]}
RESULT isUnreadableSentinel: true
RESULT -> update ran write(); reports '+ Codex'

$ cat config.toml
[mcp_servers.argent]
command = "argent"
args = [ "mcp" ]
url = "https://mcp.example.com/sse"

$ codex mcp list
Error: failed to load configuration
Caused by:
    0: config.toml:1:1: url is not supported for stdio
    1: url is not supported for stdio
       in `mcp_servers.argent`
```

Same input on `main`'s build:

```
$ cat config.toml
[mcp_servers.argent]
command = "argent"
args = [ "mcp" ]
$ codex mcp list
argent  argent  mcp  -  -  enabled  Unsupported
```

argent prints `+ Codex` in both cases. Every config-reading `codex` command fails afterwards until the file is hand-edited. Windsurf, Gemini and Kiro route their `write()` through the same helper and would form the same hybrid from a `url`/`type`-shaped entry; I did not establish whether those clients also hard-fail on it.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -40,6 +40,35 @@ function getAvailableToolIds(): string[] {
...
+// entry too — Gemini's `trust`, Kiro's `autoApprove`, Windsurf's `alwaysAllow`,
+// Codex's `tools` table, a user-set `timeout`/`disabled` — and a wholesale
+// replace drops all of it. Dropping the opt-in keys used to be masked by the
+// unconditional allowlist re-add that followed; update's refresh guard has
+// retired that re-add, so the rewrite itself must carry them over.
+function mergeServerEntry(existing: unknown, entry: McpServerEntry): Record<string, unknown> {
+  const kept: Record<string, unknown> = {};
+  if (existing && typeof existing === "object" && !Array.isArray(existing)) {
+    for (const [key, value] of Object.entries(existing as Record<string, unknown>)) {
+      if (key !== "command" && key !== "args" && key !== "env") kept[key] = value;
```
latekvo, line 56: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The exclusion set makes `enabled`/`disabled` a preserved key, which changes what update's rewrite means for a server the user switched off. The comment above names `a user-set timeout/disabled` as state to carry over, but `update.ts:696-699` describes this same `write()` as the step that "REPAIRS state the normalized view can't see (an opencode `enabled: false` ...)", and the opencode adapter still hardcodes `enabled: true` at `:1602`. The two adapters now treat the identical key in opposite directions, and the comment in `update.ts` describing the repair is true only for opencode.

The user-visible half is on `init`, which has no `isArgentManagedEntry` guard: it writes the entry, prints `+ Codex`, and finishes with "Argent is ready!" for a server that stays off. Real `codex-cli 0.142.5`:

```
$ codex mcp list                      # user has argent switched off
argent  argent  mcp  -  -  disabled  Unsupported
# this head's write() -> reports '+ Codex'
$ codex mcp list
argent  argent  mcp  -  -  disabled  Unsupported     # still off

# main's build, same input -> reports '+ Codex'
$ codex mcp list
argent  argent  mcp  -  -  enabled   Unsupported
```

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -672,13 +720,59 @@ const cursorAdapter: McpConfigAdapter = {
...
+      return `skipped - creating ~/.cursor/permissions.json would turn off Cursor's "Run Everything" mode`;
+    }
+    // An existing file is left as the user has it, argent-only shape included:
+    // we cannot tell a file an earlier argent created from one the user keeps
+    // deliberately, and deleting the latter would take away an allowlist they
+    // want. Removing it stays a manual step (or `argent uninstall`).
     const config = readJsonc(permPath);
-    const list = Array.isArray(config.mcpAllowlist) ? (config.mcpAllowlist as string[]) : [];
+    const list = Array.isArray(config.mcpAllowlist) ? (config.mcpAllowlist as string[]) : null;
+    if (list === null || list.length === 0) {
```
latekvo, line 764: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Half of this guard is pinned by nothing. The `list === null` arm is covered by the two new Cursor tests (missing file, and a file holding only `terminalAllowlist`); no test uses an `mcpAllowlist` that exists and is empty, so the `list.length === 0` half can be deleted with the suite fully green:

```
764:    if (list === null) {
=== MUTATED === Tests  576 passed (576)      (baseline: 576 passed)
```

What that half is holding back, built both ways and run against `{"mcpAllowlist": []}` on init's path:

```
head      note: "skipped - an mcpAllowlist here would turn off Cursor's \"Run Everything\"
                 mode and override its in-app allowlist"
          file: {"mcpAllowlist": []}          unchanged: true
mutated   note: undefined
          file: {"mcpAllowlist": ["argent:*"]}  unchanged: false
```

By the premise this PR rests on, that write is what switches Run Everything off - so the headline claim has two halves and only the no-file one is protected.

packages/argent-installer/test/mcp-configs.test.ts
```diff
@@ -679,6 +717,43 @@ describe("Zed adapter", () => {
...
+  // Zed's tool_permissions default governs every agent tool, not just
+  // argent's — refresh must not flip a value the user changed back to "allow".
+  it('addAllowlist on refresh does not re-impose default:"allow" the user reverted', () => {
+    const configPath = path.join(tmpDir, ".zed", "settings.json");
+    fs.mkdirSync(path.dirname(configPath), { recursive: true });
+    const original = `{\n  "agent": { "tool_permissions": { "default": "confirm" } }\n}\n`;
+    fs.writeFileSync(configPath, original);
+    const note = adapter.addAllowlist!(tmpDir, "local", { refresh: true });
+    expect(fs.readFileSync(configPath, "utf8")).toBe(original);
+    expect(note).toContain("not");
```
latekvo, line 730: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `toContain("not")` is three letters, and every other refresh guard's test asserts the full `opt in via argent init` phrase. Any replacement note containing "nothing", "not opted in", or similar satisfies it. The byte comparison two lines up does not compensate: re-writing `"allow"` over `"allow"` is byte-identical, so deleting the guard's `return;` and letting refresh fall through to `editJsoncFile(..., "allow")` also survives this test in isolation.

Two other assertions in the new set do not discriminate the value that moved:

- `:1080` asserts `toContain("[mcp_servers.argent.tools.tool-b]")` - the table header only - while its sibling on the next line asserts tool-c's full `approval_mode = "deny"`. The stated point of the backfill is that a new version's ids arrive *approved*; changing the refresh backfill to write `"deny"` keeps the suite green.
- `:484` wraps its check in `if (fs.existsSync(settingsPath))`. I ran the test's own sequence: `removeClaudePermission` prunes the document to `{}` and deletes the file, so `exists` is `false` there and the block never executes. The step the test is named for asserts nothing.

Three of the six new Cursor cases also pass verbatim against the merge-base source (`d4a712c40`) - `leaves an existing argent-only permissions.json in place`, `keeps a user-owned permissions.json intact`, and `on refresh is a byte-level no-op when argent:* is present` - so they guard the pre-existing de-dup rather than this change.

packages/telemetry/src/sanitize.ts
```diff
@@ -164,6 +164,7 @@ export const ALLOWED: ValidatorMap = {
   },
   "installation:allowlist_decision": {
     is_enabled: bool,
+    decided_by: oneOf(["flag", "default", "prompt"] as const),
```
latekvo, line 167: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Nothing in the repo pins this value set. `grep -rn "allowlist_decision\|decided_by\|decidedBy"` across the tree returns no match in any test file, and the `ValidatorMap` mapped type only proves that *a* validator exists for the prop, not that its values match the TS union. Narrowing it so two of the three values are silently dropped at the export boundary leaves everything green:

```
decided_by: oneOf(["flag"] as const)
$ npx tsc --build              exit 0
$ packages/telemetry           Tests 314 passed (314)
$ packages/argent-installer    Tests 576 passed (576)
```

The producing side is unpinned too: no test imports `configureAllowlist`, so `decidedBy: "flag"` and `decidedBy = "prompt"` can each be changed to `"default"`, or `decided_by` dropped from the `track()` call at `init.ts:251`, with the installer suite green. Since the field's only consumer is a remote collector, nothing local would ever surface the loss.

packages/argent-installer/src/update.ts
```diff
@@ -721,16 +722,32 @@ export async function update(args: string[]): Promise<void> {
...
-            adapter.addAllowlist(projectRoot, scope);
-          } catch {
-            // non-fatal
+      // Refresh allowlists only for scopes that already had argent configured —
+      // matches the editor list above. `refresh: true` lets an adapter treat a
+      // rule missing from an existing allowlist as a deliberate removal
+      // instead of re-adding it; --no-allowlist skips the refresh entirely.
+      // Both skips leave a trace so a missing auto-approve entry after an
+      // update is explainable from the output.
+      if (noAllowlist) {
```
latekvo, line 731: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `update` takes its own allowlist branch here and emits no telemetry, while `installation:allowlist_decision` has exactly one emitter, at `init.ts:249`. So the `decided_by` dimension this PR adds to measure allowlist opt-out - whose docstring defines `"flag"` as `--no-allowlist`, a flag the PR gives to both commands - can never record an `argent update --no-allowlist` run.

```
$ grep -rn "installation:allowlist_decision" --include='*.ts' packages/*/src
packages/argent-installer/src/init.ts:249
$ grep -rn "installation:stale_config_cleanup" --include='*.ts' packages/*/src
packages/argent-installer/src/init.ts:203
packages/argent-installer/src/update.ts:798
```

The nearest twin is emitted from both commands and documented as such in `events.ts` ("Post-write sweep in init/update"); `InstallationAllowlistDecisionProps` carries no init-only scoping note, and `update` already emits five other `installation:*` events. An analyst sizing opt-out from `decided_by` would undercount one whole command surface with nothing in the data indicating it is missing.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Second sweep over the same head, on the angles the first pass covered least: the telemetry and CLI surface, the discrimination of the new tests themselves, cross-command behaviour, and a fresh-eyes hunk-by-hunk re-read with no assigned lane.

The first item below is a regression this change introduces, reproduced against the real `codex-cli 0.142.5` on this machine and confirmed absent on `main`. The rest are coverage and reporting gaps, each with the mutation or run that establishes it.

Re-verified as clean in this pass, so they need no attention: `smol-toml` ordering and round-tripping of a merged entry (scalars hoist above sub-tables; datetimes, inline tables, arrays-of-tables all survive); `mergeServerEntry`'s key ordering and its `hasEnv` interaction (no key in the spread can collide with the three it sets); the eight refresh guards each having both an opted-in and an opted-out arm; `decided_by` reaching the OTLP wire for all three values and being stripped rather than dropping the event when absent or bogus; `--no-allowlist` parsing, forwarding and help rendering on both commands, including `init --no-allowlist help`; and the absence of any other flag-enumerating surface in the repo that this PR missed.

packages/argent-installer/src/init.ts
```diff
@@ -244,8 +244,12 @@ export async function init(args: string[]): Promise<void> {
...
       nonInteractive: parsed.nonInteractive,
+      noAllowlist: parsed.noAllowlist,
+    });
+    track("installation:allowlist_decision", {
+      is_enabled: allowlist.enabled,
+      decided_by: allowlist.decidedBy,
     });
-    track("installation:allowlist_decision", { is_enabled: allowlist.enabled });
     if (allowlist.enabled && allowlist.lines.length > 0) {
       p.note(allowlist.lines.join("\n"), "Tool Auto-Approval");
```
latekvo, line 254: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The note printed here and the Summary block at `init.ts:362` disagree about the same run. `printSummary` is handed `allowlistEnabled: allowlist.enabled` (`:287`), and `enabled` is the answer to the prompt (or the `-y` default), never what the step achieved - the per-adapter outcomes only ever reach `lines`.

Before this PR that was accurate for Cursor, because answering yes always wrote the rule. Now the default Cursor install - the one with no `~/.cursor/permissions.json` - takes the new skip exit, and a single run prints both halves:

```
A0 ~/.cursor before: [ 'mcp.json' ]
A1 configureAllowlist -> {"enabled":true,"decidedBy":"default"}
A2 Tool Auto-Approval note: ["- Cursor (skipped - creating ~/.cursor/permissions.json would turn off Cursor's \"Run Everything\" mode)"]
A3 Summary line: Auto-approve enabled
A4 ~/.cursor after: [ 'mcp.json' ]
```

Nothing was written, and the last thing on screen says auto-approve is enabled. Cursor is also the only adapter that can produce this at `init`: every other `addAllowlist` note in this file is behind `options?.refresh`, so it is unreachable from `init`, which passes no options.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -40,6 +40,35 @@ function getAvailableToolIds(): string[] {
...
+  const kept: Record<string, unknown> = {};
+  if (existing && typeof existing === "object" && !Array.isArray(existing)) {
+    for (const [key, value] of Object.entries(existing as Record<string, unknown>)) {
+      if (key !== "command" && key !== "args" && key !== "env") kept[key] = value;
+    }
+  }
+  return {
+    command: entry.command,
+    args: entry.args,
+    ...(hasEnv(entry) ? { env: entry.env } : {}),
```
latekvo, line 62: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `env` is handled here and excluded from `kept` at `:56`, but the reader accepts two spellings - `normalizeServerEntry` at `:249` takes `raw.env` **or** `raw.environment`, as the comment at `:243` notes for opencode. Only one of the two is excluded, so an `environment` key survives the merge and is then read back as `env`, which is the one thing that makes `hasCustomizingEnv` true and `isArgentManagedEntry` false.

The result is that argent's own write produces an entry argent's own classifier reads as a user customization. Same input through this head's `write()`, then `update.ts`'s own classification:

```
B Gemini    keysAfterWrite=["command","args","environment"] getArgentEntry={"command":"argent","args":["mcp"],"env":{"FOO":"1"}} isArgentManagedEntry=false -> update SKIPS: left a customized entry untouched
B Windsurf  keysAfterWrite=["command","args","environment"] getArgentEntry={"command":"argent","args":["mcp"],"env":{"FOO":"1"}} isArgentManagedEntry=false -> update SKIPS: left a customized entry untouched
B Kiro      keysAfterWrite=["command","args","environment"] getArgentEntry={"command":"argent","args":["mcp"],"env":{"FOO":"1"}} isArgentManagedEntry=false -> update SKIPS: left a customized entry untouched
```

From there every `argent update` takes the `! left a customized entry untouched` exit at `update.ts:701-705` and stops repairing `command`/`args` - which is what a local install depends on, since the args carry a `node_modules` path that moves between versions. On `main` the wholesale replace dropped the key, so the entry stayed managed and kept being repaired.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Third pass over `7739795b0`. What I ran this round: real `gemini` 0.46.0 against configs produced by this head's `write()` and by `main`'s under a sandboxed `HOME`; `configureAllowlist` driven directly on this head's build for a Cursor-only install; and the `write()` -> `getArgentEntry` -> `isArgentManagedEntry` chain replayed the way `update.ts:687-712` classifies. Two new threads below, plus a follow-up on the open `mergeServerEntry` thread now that I have the client behaviour it was missing.

packages/argent-installer/src/mcp-configs.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Following up on the last paragraph - I have now run the Gemini half. Real `gemini` 0.46.0 under a sandboxed `HOME` with `security.folderTrust.enabled: false`, same starting entry (`{httpUrl, trust}`) through each build's `write()`:

```
=== PR head write() output (hybrid) ===
  file: {"command": "argent", "args": ["mcp"], "httpUrl": "https://argent.example.com/mcp", "trust": true}
  gemini mcp list:  argent: https://argent.example.com/mcp (http) - Disconnected
=== main write() output (wholesale replace) ===
  file: {"command": "argent", "args": ["mcp"]}
  gemini mcp list:  argent: argent mcp (stdio) - Connected
```

`command`/`args` are never read - the transport is chosen from `httpUrl`/`url` ahead of `command`, so the merged entry dials the URL and `argent mcp` is never spawned. No error, no warning; the only signal is `(http)` in `gemini mcp list`, and argent prints `+ Gemini` either way. Windsurf and Kiro I did not run, so the other two remain unestablished.

One more consequence of the same `kept` set on Gemini: its per-server schema is closed, so any foreign key the merge carries over draws a config error on every invocation, and the merge re-preserves it on each subsequent write:

```
Invalid configuration in .../.gemini/settings.json:
Error in: mcpServers.argent
    Unrecognized key(s) in object: 'alwaysAllow', 'autoApprove'
Please fix the configuration.
...
argent: argent mcp (stdio) - Connected
```

Same with a single `disabled` key. The server still loads, so that half is a permanent banner rather than a break - `main`'s output produced no error at all.

filip131311:
Closing in favor of https://github.com/software-mansion/argent/pull/986, which replaces the per-adapter refresh heuristics with a recorded decision: a new `allowlist.enabled` config key that `argent init` writes and `argent update` honors, plus the never-create/key-presence guard for Cursor's permissions.json. The review findings on this branch that are independent of the refresh machinery (Zed uninstall resetting a default argent never set, the `hasClaudePermission` scope merge, mergeServerEntry vs streamable-HTTP entries) don't apply to the replacement since that code isn't carried over.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -672,13 +720,59 @@ const cursorAdapter: McpConfigAdapter = {
...
+  // option — Cursor treats a permissions-file approvalMode as authoritative
+  // (`getModeFullAutoRun` returns true for it), so it would force Run
+  // Everything ON for users who chose "Ask every time".
+  addAllowlist(_root, _scope, options): string | void {
     const permPath = path.join(homedir(), ".cursor", "permissions.json");
+    if (!fs.existsSync(permPath)) {
+      return `skipped - creating ~/.cursor/permissions.json would turn off Cursor's "Run Everything" mode`;
+    }
+    // An existing file is left as the user has it, argent-only shape included:
+    // we cannot tell a file an earlier argent created from one the user keeps
```
latekvo, line 759: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This rationale is contradicted twice inside this PR, and the add/remove pair it produces no longer round-trips.

`cursorDirLooksArgentOnly` (`:419-448`) does tell those two apart, and its own comment at `:433` names the exact document - `// Written by this adapter's addAllowlist: { mcpAllowlist: ["argent:*"] }`. `detect()` already trusts that discriminator for a larger decision, whether Cursor counts as installed at all:

```
permissions.json absent            detect()=false  -> argent-only (its own leftovers)
exactly what older argent wrote    detect()=false  -> argent-only (its own leftovers)
user rule alongside argent:*       detect()=true   -> user-owned
user rule only                     detect()=true   -> user-owned
```

The PR's own new test asserts the same attribution in the opposite direction - `test/mcp-configs.test.ts:2467`, *"Cursor removeAllowlist deletes the file when argent was the only rule"*, commented *"As argent <= 0.20.x created it before addAllowlist stopped creating files"*. `removeAllowlist` is allowed to attribute that document to argent and delete it; `addAllowlist` is not allowed to attribute it and act on it.

Composing the two is a one-way door for a user who accepted the recommended prompt on a released version and keeps no other Cursor rules:

```
1 state left by an older argent init      : "{\n  \"mcpAllowlist\": [\n    \"argent:*\"\n  ]\n}"
2 after argent uninstall (removeAllowlist): <FILE ABSENT>
3 argent init addAllowlist returns        : "skipped - creating ~/.cursor/permissions.json would turn off Cursor's \"Run Everything\" mode"
4 file after re-opt-in via argent init    : <FILE ABSENT>
5 argent update refresh returns           : "skipped - creating ~/.cursor/permissions.json would turn off Cursor's \"Run Everything\" mode"
6 file after update                       : <FILE ABSENT>
```

`argent uninstall` destroys argent-authored state that `argent init` - the route every other adapter's note names, "opt in via argent init" - cannot restore, and every later `argent update` repeats the same dead end. Cursor's note is also the one that does not name a remedy.

The same claim is what the description leans on for "Already-demoted machines are not repaired by this PR". Those are the machines that hit the reported bug: every user who ran a released argent against Cursor with no prior `permissions.json` is still in Allowlist mode after this ships, and argent has the discriminator to find them.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -1114,13 +1236,35 @@ const zedAdapter: McpConfigAdapter = {
...
     const settingsPath =
       scope === "global"
         ? path.join(homedir(), ".config", "zed", "settings.json")
         : path.join(root, ".zed", "settings.json");
+    if (options?.refresh) {
+      // Zed merges project settings over global, so a project file with no
+      // agent block inherits the global default — check both before deciding
+      // the user is not opted in.
+      let current = zedToolPermissionsDefault(settingsPath);
+      if (current === undefined && scope === "local") {
```
latekvo, line 1258: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The cell this fallback exists for is the one no fixture stages. The two new Zed refresh tests cover a project file with an explicit `"confirm"` and no global file (`test:723`), and a project `{}` over a global `"allow"` (`test:745`) - neither has a project value AND a global value, so nothing separates `current === undefined` from `current !== "allow"` here. Widening the condition that way keeps the suite at 576 passed:

```
1258:      if (current !== "allow" && scope === "local") {
=== MUTATED === Tests  576 passed (576)      (baseline: 576 passed)
```

What that reads as opted in is a project `.zed/settings.json` that deliberately sets `default: "confirm"` while the global file still says `"allow"` - the merge outcome is `"confirm"`, and it is the one shape where the project file is overriding rather than inheriting. Given what this default governs (every agent tool in Zed, per the comment at `:1244`), that is the case worth pinning.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fourth pass over `7739795b0`, run as four lanes: mutation testing of the new production code, `--no-allowlist` end to end, add/remove lifecycle symmetry, and hostile inputs. Most of what came back was already on the PR - the `{ refresh: true }` and `--no-allowlist` mutation survivors are the thread on `update.ts:741`, and `decided_by` being emitted for a run that was never prompted is already on `events.ts:68`. Two survive as new, below. Also checked and clean: `addAllowlist(..., { refresh: true })` on a fully opted-in state leaves the file byte-identical for all seven JSON adapters, so the "the one write refresh still performs is Codex's backfill" claim at `:118-125` holds; the installer suite is 576 passed at this head.

packages/argent-installer/src/mcp-configs.ts
```diff
@@ -40,6 +40,35 @@ function getAvailableToolIds(): string[] {
...
+      if (key !== "command" && key !== "args" && key !== "env") kept[key] = value;
+    }
+  }
+  return {
+    command: entry.command,
+    args: entry.args,
+    ...(hasEnv(entry) ? { env: entry.env } : {}),
+    ...kept,
+  };
+}
```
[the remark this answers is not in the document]
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This half does not reproduce, so it is probably not worth chasing. `readJsonc` goes through jsonc-parser, which assigns with `obj[key] = value`; for `__proto__` that sets the object's prototype instead of creating an own property, so `Object.entries(existing)` never yields it and `kept[key] = value` never runs for it. Real input through this head's built `write()`:

```
input JSON        : {"mcpServers":{"argent":{"__proto__":{"polluted":1},"constructor":{"x":1},"trust":true}}}
own keys of entry : ["constructor","trust"]
Object.entries    : [["constructor",{"x":1}],["trust",true]]
after write()     : {"mcpServers":{"argent":{ "command": "argent", "args": [ "mcp" ], "constructor": { "x": 1 }, "trust": true }}}
({}).polluted     : undefined
Object.prototype.polluted: undefined
```

`constructor` is copied, but onto a fresh object literal and re-serialized as an ordinary JSON key - no `Object.prototype` mutation and nothing downstream reads it. The arbitrary-key copying does have consequences, just different ones: they are on the two threads at `:56` and `:64`.

PR #803, opened 2026-08-17
feat(shake): wobble the host window behind a `microinteractions` flag

kacperkapusciak:
`shake` only touches the guest, so nothing on screen explains why an app popped its dev menu. This wobbles the host window carrying that device while the gesture runs.

- Opt-in `microinteractions` flag, off by default (`argent enable microinteractions`). Read inside the handlers, not as `ToolDefinition.featureFlag`, which would hide the whole tool.
- Can never fail a shake: every path warns once to stderr and resolves. Skipped off macOS, on headless boots, and for `sim-remote`.
- Windows are title-matched on the iOS display name, so a host with several booted simulators wobbles the right one.
- `prepareHostWindowShake` resolves flag/platform/device-set/script once per call; `begin()` per gesture, `settle()` once after the loop. Not in the `finally`, so a real failure throws immediately rather than waiting on cosmetics.
- `ARGENT_SIMULATOR_NO_WINDOW` / `ARGENT_EMULATOR_NO_WINDOW` parsing extracted to `utils/no-window-env`, shared with `boot-device`.

### iOS simulators only

The wobble originally covered Android emulators too and has been removed there. Picking the emulator's window meant matching a title among several `qemu-system-*` host processes, using needles built from the console port plus an extra `adb emu avd name` round trip per shake. That lookup was the fragile half of the feature: a build that drops the port from the title, or a serial with no console port, falls back to either the wrong window or no animation at all. An Android shake also already produces visible in-guest feedback, so the amplification bought less there.

`shake` itself is unchanged on Android: the same accelerometer burst, the same resting-vector capture and restore, the same per-device lock. Only the cosmetic host-window animation is gone. `ARGENT_EMULATOR_NO_WINDOW` handling stays in `utils/no-window-env`, where `boot-device` still uses it.

Verified live on an iPhone 17 Pro Max: window returns to its exact origin and the wobble fits inside one gesture. The Android sensor path was verified live on a Pixel_9 before the removal and is unchanged by it; it has not been re-run since. Full tool-server suite green. The AppleScript motion itself is only verified by eye; unit tests stub `osascript`.

### DIFF
--- a/packages/configuration-core/src/flags.ts
+++ b/packages/configuration-core/src/flags.ts
@@ -70,7 +70,7 @@ export const FLAG_REGISTRY: readonly FlagDefinition[] = [
   {
     name: "microinteractions",
     description:
-      "Amplify a device gesture with a matching animation of the host window — the Simulator / emulator window physically shakes while `shake` runs. Purely cosmetic, macOS only, and never affects whether the gesture itself succeeds. Off by default.",
+      "Amplify device actions with matching animations of the host window, so what happens on the guest is also visible on the desktop. Purely cosmetic, macOS only, and never affects whether the underlying action succeeds. Off by default.",
   },
   {
     name: "video-watermark",
--- a/packages/configuration-core/src/flags.ts
+++ b/packages/configuration-core/src/flags.ts
@@ -68,6 +68,11 @@ export const FLAG_REGISTRY: readonly FlagDefinition[] = [
     description: "Write structured tool-server lifecycle events to a JSONL file.",
   },
   {
+    name: "microinteractions",
+    description:
+      "Amplify a device gesture with a matching animation of the host window — the Simulator / emulator window physically shakes while `shake` runs. Purely cosmetic, macOS only, and never affects whether the gesture itself succeeds. Off by default.",
+  },
+  {
     name: "video-watermark",
     description:
       "Overlay the argent corner watermark on recorded screen videos. On by default; turn it off with `argent disable video-watermark`.",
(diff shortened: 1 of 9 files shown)


packages/configuration-core/src/flags.ts
```diff
@@ -67,6 +67,11 @@ export const FLAG_REGISTRY: readonly FlagDefinition[] = [
     name: "tool-server-event-log",
     description: "Write structured tool-server lifecycle events to a JSONL file.",
   },
+  {
+    name: "microinteractions",
+    description:
+      "Amplify a device gesture with a matching animation of the host window — the Simulator / emulator window physically shakes while `shake` runs. Purely cosmetic, macOS only, and never affects whether the gesture itself succeeds. Off by default.",
```
filip131311: the description is "shake focused" but this flag should be used in the future for other microinteractions pleas refrase

packages/skills/skills/argent-device-interact/SKILL.md
```diff
@@ -78,6 +78,8 @@ Common schemes: `messages://`, `settings://`, `maps://?q=<query>`, `tel://<numbe
 | Wait for UI       | `await-ui-element`  | Block until an element is visible/hidden/exists/contains text    |
 | Wait for idle     | `await-screen-idle` | Block until a non-empty screen tree stops changing               |
 
+`shake` drives the guest only, so nothing about it is visible on the host. On macOS, `argent enable microinteractions` makes the Simulator / emulator window itself wobble for the length of each gesture, which is useful when demoing or recording. It is cosmetic: the shake reports the same result whether or not the animation runs (it needs Accessibility permission for the process running the tool-server, and is skipped on a headless boot).
```
filip131311: remove this commone

filip131311 (approved):
i have some reservations about the skills and description of the flag but the rest seems to be resonable guarded by the flag so just go for it if it works

packages/configuration-core/src/flags.ts
kacperkapusciak: Rephrased in 030c0734 so the description covers host-window microinteractions generally instead of just `shake`.

packages/skills/skills/argent-device-interact/SKILL.md
kacperkapusciak: Removed in 030c0734.

PR #821, opened 2026-08-17
fix(keyboard): read the field back, instead of trusting a select-all that exits 0

latekvo:
Stacked on #581.

`input keycombination 113 29` exits 0 whether or not the select-all actually took, and on a React Native `TextInput` it repeatedly does not — so the DEL that follows removes one character instead of the value, the `text` is appended to the residue, and the tool returns `cleared: true`.

**The fix:** after the DEL, read the field back. A residue goes to `clearByDeleting`, the selection-free run the legacy path already uses. Only a *positive* reading redirects — unreadable leaves the fast path alone rather than spending a blind run on a clear that most likely worked.

Measured on a Pixel 6 / API 34 against the Expo dev-launcher URL field: **85/85 clean, against 15/85 corrupt on the same path unfixed.** A native `EditText` ran 85/85 clean either way.

No docs update needed: `packages/docs` carries a one-line `keyboard` row and describes none of this.

<details>
<summary>Why not just slow the chord down</summary>

`-t 100` still missed 2/30 and `-t 300` missed 7/30 — worse, not better. There is no hold duration that makes it safe. The selection-free delete run cleared the same field 50/50.

Isolating the legs on the RN field: `MOVE_END` + one DEL, no chord — 30/30 correct, the DEL never misses. Chord then type, no DEL — 4/30 left the field untouched. Chord + DEL + text in one `adb shell` — 14/30 corrupt. Reading `mCursorSelStart/SelEnd` after the chord: 5/20 never selected, 11/20 only after 500ms.
</details>

<details>
<summary>What the read-back does NOT repair</summary>

Named on the `clear` parameter and the `cleared` docstring rather than left to be discovered:

- a screen the hierarchy cannot capture, or a clear whose earlier legs already spent the budget the read needs;
- a focused editable it cannot read — a password box, a WebView input, a custom widget — and one of those focused *anywhere* on screen discards the reading for the target too, since the walk cannot tell which field was meant;
- a field reported as holding exactly its own placeholder, which is read as empty;
- the repair itself deletes from end-of-LINE, so a multi-line field keeps what sits below the caret.
</details>

<details>
<summary>Reading the field back is a predicate, not a count</summary>

`measureFocusedTextLength` serves both, and the two want opposite errors. For the legacy sizing read an over-measurement costs a few no-op backspaces while an under-measurement truncates the field. For the read-back an over-measurement fires a delete run — or the length refusal — against a field the clear already emptied, and an under-measurement only leaves the fast path as it was.

So `readBack` flips three rules: the dump is not retried; any unmeasurable focused editable makes the whole reading undefined instead of contributing the blind floor; and a `text` equal to its own `hint` counts as empty. Without that last one a successful clear of a hinted field fires 24 backspaces into the empty box, and a placeholder past `MAX_DELETE_COUNT` turns the same successful clear into a 400 with the `text` never sent.
</details>

<details>
<summary>Messages that were asserting device state they cannot know</summary>

- The interrupted-DEL report enumerated two states. A delete already handed to the device keeps landing after adb is killed, so on a swallowed chord the field is one character short — a third state.
- The rescue-path over-length refusal opened "The select-all did not take on this field" and then conceded, three sentences later, that the count can be another window's field. It reports what it observed instead: the field read back non-empty.
- The same refusal said the DEL "removes one character where it lands" — true only where the chord was swallowed. A chord that took makes it remove the whole selection.
- The tool description promised "a rejected request never changes the field … a 400 leaves the field exactly as it was". Through the verify leg that same over-length 400 is raised after a select-all and a DEL have gone out; the paragraph names that one exception now.
</details>

<details>
<summary>Also</summary>

`keyboard-android.test.ts` left `adbExecOutBinary` unmocked while addressing `emulator-5554`. With an emulator attached those tests dumped a live device, so a clear's asserted command stream depended on whatever was on its screen — green in CI, red locally. Now mocked, matching `keyboard-clear.test.ts`.
</details>

## Checks

`vitest` (full tool-server suite, 346 files), `tsc`, `typecheck:tests`, `eslint`, `prettier`, `knip` all clean locally. Every new test is mutation-checked — each of `readBack`'s three rules, the placeholder rule's `readBack` gate and its `text === hint` compare, the hedged cause and state, the re-measure and the helper-read plumbing each fail their own test and nothing else.


latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: A review pass over this branch surfaced one real defect in the verify leg, now fixed in b6e903800.

**Blind floor read as residue on password fields (fixed).** `measureFocusedTextLength` floors an unmeasurable focused editable to `BLIND_DELETE_COUNT` — correct for the legacy sizing read, where over-deleting is a no-op. The select-all read-back inherited that floor, so every successful clear of an unmeasurable field (a focused password box) measured 150 back, passed the `residue > 0` gate, and fired a MOVE_END + 158-keyevent run into the credential box it had just emptied. Reproduced in the unit harness: an empty focused `password="true"` node after chord+DEL produced `[SELECT_ALL, DEL, RUN×158]`. The read-back now passes a flag that makes any unmeasurable focused editable render the whole reading undefined — evidence of nothing, like a refused dump — and the fast path stays untouched. Regression test added; mutation-checked (removing the flag fails exactly that test).

**Stale claims vs code (fixed).** The `clear` parameter description, the tool description body, and the `cleared` contract docstring all still said Android never reads the field back / only Chromium does. All three now describe the read-back-and-rescue behavior and its limits (uncapturable screen degrades to best-effort like iOS).

Verified: keyboard-clear + keyboard-android suites (202 tests), `tsc` incl. test config, eslint, prettier all green; knip output unchanged from the base commit.

latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Review converged after four passes. Remaining known-accepted limits, recorded here for the record (all LOW, none fixed by design):

- A placeholder reported as value can still trigger the rescue where the dump cannot tell them apart — it costs a delete run that removes nothing, which the code already documents and accepts.
- If the read-back ever misread a cleared field as >150 characters of residue, the rescue-path refusal would say "the field HAS been modified" over a field the DEL actually emptied. Requires a >150-char misreading; not observed.
- Every modern clear now consults the view-hierarchy reader, coupling clear latency to that RPC's worst case on a slow helper. Bounded by the existing 5s race; noted at `platforms/android.ts` without change.

latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Re-review pass over the branch after the four earlier passes. No reviewer threads existed to resolve; this pass surfaced one MEDIUM defect class, fixed in b605e1e5.

**Error messages asserting a device state the code cannot know (fixed).** Two messages stated as unconditional fact what this PR's own premise denies:

- The interrupted-DEL report said *"the select-all landed and survives, so the next character typed into it replaces the value."* But `input` exits 0 whether or not the selection took — that is the bug this branch fixes — so on a swallowed-chord widget a killed DEL reaches this message over an intact, UNselected field, and the promised replace-on-type becomes an append/splice. Now hedged ("selected or not") with both outcomes named.
- The rescue-path over-length refusal said *"The field HAS been modified… removed one character from it."* The reading behind that refusal is the longest focused `EditText` on screen, which can be another window's field or an empty field's placeholder (`measureFocusedTextLength` documents both conflations) — so what the DEL actually took is unknowable at that point, and the flat claim can fire over a field the chord emptied. Now hedged to "MAY have been modified"; the legacy arm keeps its flat "Nothing was modified", which IS provable there (the probe refuses before anything is sent).

**Doc scope fix.** The tool description scoped the over-150-character refusal to levels without `input keycombination`; through the verify leg the same refusal is reachable on a level that HAS the subcommand, and the sentence now says so.

Mutation check: restoring either flat claim fails exactly its test (`hedges the field's state…`, the interrupted-DEL rewrap test).

Everything else came back clean across claims-vs-code, reachability both ways, lifecycle-after-call, twins, inputs, and non-happy paths: budget/reserve arithmetic holds with the extra dump leg; no path composes sizing-dump + rescue; the rescue can never eat text typed in the same call (text is injected only after the clear resolves); `cleared: true` cannot survive a rescue failure; race timers are cleaned up; no other clear site in the repo sends a select-all chord.

Remaining known-accepted limits (all LOW, unchanged from the previous pass): placeholder-as-residue can cost one no-op delete run; clear latency now rides the view-hierarchy read (bounded by the existing 5s race); up to two abandoned helper reads per clear can hold the devtools chain slightly longer than base.

Verified: keyboard-clear + keyboard-android suites 202/202, `tsc`, eslint, prettier all green on b605e1e5.

packages/tool-server/src/utils/android-input.ts
```diff
@@ -384,6 +387,37 @@ export async function injectAndroidClear(
...
+  // treating it as failure would spend a blind BLIND_DELETE_COUNT run on every
+  // clear taken where `measureFocusedTextLength` cannot see. A measurement that
+  // is really a placeholder costs a run that deletes nothing.
+  //
+  // `unmeasurableIsUnreadable` carries that rule to the fields the dump cannot
+  // read: without it a focused password field floors to BLIND_DELETE_COUNT (see
+  // the measurement), so every successful clear of a credential box would read
+  // 150 back and fire a 158-keyevent run into the field just emptied.
+  const residue = await measureFocusedTextLength(serial, deadline, options.readHierarchy, 1, true);
+  if (residue !== undefined && residue > 0) {
```
latekvo, line 431: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: The rescue inherits the legacy delete run's line scope, so it cannot reach everything the read-back just counted. Reproduced against the unit harness: seeding the post-DEL read-back with a multi-line value (`AAAA\nBB\nC`, nine characters once the DEL has taken one) makes this leg emit `input keyevent 123` plus seventeen backspaces and complete without error — i.e. a successful clear. `KEYCODE_MOVE_END` is end-of-LINE rather than end-of-buffer (as this file documents below), so on a multi-line field whose chord was swallowed while the caret sat above the last line — reachable through the tap-to-focus path the tool description itself describes — everything below the caret survives the rescue while the call reports `cleared: true`, and an accompanying `text` appends to that survivor. That is the corruption shape this branch set out to eliminate, persisting on multi-line fields, while both the parameter description ("deletes whatever survived it, repairing the clear") and the `cleared` docstring state the repair without the line-scope qualifier the older-level path carries.

packages/tool-server/src/utils/android-input.ts
```diff
@@ -384,6 +387,37 @@ export async function injectAndroidClear(
...
+    await clearByDeleting(serial, deadline, options, residue);
```
latekvo, line 432: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: The measurement only counts focused nodes whose class contains "EditText", so any other editable gets no rescue. Reproduced: seeding the post-DEL read-back with a focused node of class `android.webkit.WebView` carrying readable text (`Monda`, the residue a swallowed chord leaves) sends only the chord and the DEL — no delete run follows — and reports `cleared: true` over the value minus one character. Inputs inside a WebView or a fully custom widget therefore keep exactly the failure this branch fixes, and the docstring's carve-outs for the read-back (a refused screen, a password field) do not cover them.

packages/tool-server/src/utils/android-input.ts
```diff
@@ -372,10 +374,11 @@ export async function injectAndroidClear(
     // retries against a field it believes is untouched.
     throw new FailureError(
       `keyboard clear: the delete did not finish on this device, so the focused field is ` +
-        `either empty or still holds its whole value with all of it SELECTED — the select-all ` +
-        `landed and survives, so the next character typed into it replaces the value. Nothing ` +
-        `was typed. Read the field's actual contents before continuing; do not treat it as ` +
-        `unchanged, and do not send a replacement that assumes it is empty.`,
+        `either empty or still holding its whole value — selected or not: \`input\` exits 0 ` +
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: This enumeration misses a third reachable state. On a widget that swallows the chord, the DEL removes exactly one character — and a DEL that was delivered but outlived its adb timeout still takes effect, since this file records elsewhere that deletes already handed over keep landing when adb is killed. That leaves the value minus one character, unselected: neither empty nor "still holding its whole value". A caller who trusts the enumeration instead of reading the field misjudges what a replacement will do to it.

packages/tool-server/src/utils/android-input.ts
```diff
@@ -513,24 +555,53 @@ async function clearByDeleting(
...
+    // remedy at all, since the level already has the subcommand. The mutation is
+    // hedged rather than asserted because the reading behind this refusal can be
+    // another window's field or a placeholder (see measureFocusedTextLength), so
+    // the code cannot know what the DEL actually took.
+    const why =
+      rescueFrom === undefined
+        ? `Without \`input keycombination\` (added after API 30) the only available clear is ` +
+          `one backspace per character, which is too slow to finish reliably past ` +
+          `${MAX_DELETE_COUNT}.`
+        : `The select-all did not take on this field, leaving one backspace per character as ` +
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: "The select-all did not take on this field" asserts as fact something the reading behind this refusal cannot establish — the rationale two sentences down concedes the count can be another window's field or an empty field's placeholder. Over such a spurious reading the chord may have taken and actually emptied the field, and this message tells the caller the opposite happened before directing them to read the field anyway.

packages/tool-server/src/tools/keyboard/index.ts
```diff
@@ -66,8 +66,8 @@ const zodSchema = z.object({
...
+        "On Android the read-back needs the screen's view hierarchy; where that cannot be captured, or where the focused field is one it cannot read (a password box), the clear stays best-effort like iOS. On levels older than `input keycombination` the clear deletes backwards from end-of-LINE, so a multi-line field keeps what sits below the caret, and a field over 150 characters is refused rather than partly deleted; the same refusal can fire on a level that HAS the subcommand when the read-back leaves more residue than backspaces are allowed to clear. A length that cannot be read at all (a password field, or a screen the device would not capture) falls back to a fixed 158 backspaces — more than every length that path accepts, so a field only keeps its head past 158 characters."
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Two sentences in this paragraph now give the same two triggers different outcomes. "Where that cannot be captured, or where the focused field is one it cannot read (a password box)" ends in best-effort with no delete run — but the final sentence says a length that cannot be read at all (a password field, or a screen the device would not capture) falls back to a fixed 158 backspaces. That fallback exists only behind the sizing read of a level without `keycombination`; on a level that has it, both triggers make the verify read return undefined and nothing further is sent. The final sentence's older-levels scoping was established two clauses upstream, but the inserted modern-refusal sentence between them lets a reader attach the 158-keyevent run to modern levels, where it never fires.

packages/tool-server/src/tools/keyboard/types.ts
```diff
@@ -49,11 +48,12 @@ export interface KeyboardResult {
...
    * - Android parses the `input keycombination` output, so a level without the
    *   subcommand takes the measured delete path instead of silently degrading
-   *   to a one-character backspace. It does not read the field back, though, so
-   *   a widget that swallows the select-all chord on a level that HAS the
-   *   subcommand leaves the following delete acting as a plain backspace: the
-   *   field is left one character shorter and reported as cleared. That is a
-   *   mutated field, not a no-op.
+   *   to a one-character backspace. On a level that HAS the subcommand it also
+   *   reads the field back afterwards and deletes whatever the select-all left
+   *   behind — a widget that swallows the chord does not leave the field one
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: This bullet contradicts itself across adjacent sentences: "a widget that swallows the chord does not leave the field one character shorter under a `cleared: true`" is stated absolutely, yet the next sentence scopes the same condition back to best-effort "like iOS" — under which a swallowed chord does leave the field one character shorter under `cleared: true` (helper down, dump refused or Killed, password box). It also does not hold for focused editables the hierarchy walk cannot measure at all. The parameter description phrases this correctly; the docstring's unconditional clause over-promises what android-input.ts implements.

packages/tool-server/test/keyboard-clear.test.ts
```diff
@@ -624,6 +624,115 @@ describe("keyboard clear — Android (adb input)", () => {
     expect(inputCmds().some((c) => c.includes("uiautomator"))).toBe(false);
   });
 
+  it("deletes what the select-all left behind, instead of trusting the chord", async () => {
```
latekvo, line 636: \[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Coverage gap: every positive-residue test seeds the read-back through the exec-out dump mock, so the production-primary source — `options.readHierarchy` from the devtools helper, which is how the ordinary describe → tap → keyboard order answers — is never exercised positively on this leg. A regression that stopped passing `options.readHierarchy` into the verify read would leave this suite green while disabling the repair whenever the helper holds the UiAutomation connection, which is the state the option exists for.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha\]: Ran a fresh full pass over b605e1e5: claims vs code across all three rewritten prose sites, nearest-twin comparison against the Chromium and iOS clear semantics, non-happy paths (residue undefined / zero / positive / over-limit, dump failure, deadline exhaustion mid-verify, interrupted rescue run), inputs (flag plumbing, call sites, the rescueFrom gate), reachability in both directions (legacy levels, TV targets, key-vs-text riders, repo-wide search for any other select-all sender), lifetime-after-call (race timers, abandoned helper reads), plus the absence pass over sibling comments and stale statements, and mutation spot-checks of the new tests (removing the flag and restoring the default dump retry each fail exactly their one guarding test). Two harness reproductions were executed against the mocked adb transport — a multi-line residue driven through the rescue leg, and a focused non-EditText editable with readable residue driven through the verify leg — reported inline below along with trace-checked message-truthfulness items. The four earlier passes' fixes were verified present at HEAD; their recorded accepted limits are not re-raised.

One further minor item, which the diff offers no line to anchor: the invocation-count comments at packages/tool-server/src/tools/keyboard/platforms/android.ts:91-93 and device-chain.ts:15 are now stale — with the verify leg the modern clear runs up to four adb invocations before any text (probe, DEL, read-back dump or helper RPC, optional rescue run), so "a third" understates the interleaving window those serialization notes cover.

packages/tool-server/src/utils/android-input.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Confirmed and fixed in 56d03bf78.

Reproduced independently in the unit harness before touching anything: a post-DEL read-back seeded with `text="AAAA&#10;BB&#10;C"` (nine characters, three lines) emits `input keycombination 113 29 2>&1`, `input keyevent 67`, then `input keyevent 123 67×17`, and resolves with `cleared: true`. `KEYCODE_MOVE_END` is 123 and is end-of-LINE, so everything below the caret's line survives that run.

Fixed as a docs finding rather than a behaviour change: the rescue IS `clearByDeleting`, and giving it a buffer-scoped clear the legacy caller does not get would split the two callers apart for a limit both carry. So the qualifier the older-level path already had is now carried everywhere the repair is claimed — the `clear` parameter ("repairing the clear on a single-line field it can read", plus "The repair itself deletes backwards from end-of-LINE, so a multi-line field keeps what sits below the caret"), the `cleared` docstring's third escape clause, `clearByDeleting`'s own known-limit paragraph ("The rescue does not escape it"), and the verify leg's comment.

packages/tool-server/src/utils/android-input.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Confirmed and fixed in 56d03bf78.

Reproduced: a focused `class="android.webkit.WebView"` node carrying `text="Monda"` after the chord and the DEL yields exactly `[input keycombination 113 29 2>&1, input keyevent 67]` and `cleared: true` — no rescue, because `measureFocusedTextLength` skips every focused node whose class does not match `/EditText/`, so `longest` never leaves `undefined`.

Left as a limit rather than widened: that `EditText` restriction is what makes a measured `0` trustworthy (a focused non-text container reports `text=""`, and taking it would issue `DELETE_MARGIN` backspaces against a field that is not empty), and it is the same predicate the TV focus walk uses for `isEditable`. So the carve-out is documented instead — the `clear` parameter now reads "measures focused `EditText` nodes only; where the screen cannot be captured, or the focused field is one it cannot read (a password box, a WebView input, a custom widget), the clear stays best-effort like iOS", the `cleared` docstring names it as the second of three escapes, and the verify leg's comment names it at the call site.

packages/tool-server/src/utils/android-input.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Confirmed and fixed in 56d03bf78.

The third state is reachable exactly as described, and this file establishes both halves itself: `input` exits 0 whether or not the select-all took (this branch's premise), and the delete-run rewrap below records that killing adb part-way does NOT stop the device — deletes already handed over keep landing. So a swallowed chord plus a delivered-but-timed-out DEL leaves the value minus one character, unselected, which the two-way enumeration did not admit.

The message now names all three: `empty, still holding its whole value — selected or not, since \`input\` exits 0 whether or not the select-all took — or one character shorter, since a delete already handed to the device keeps landing after adb is killed.` The comment above it no longer opens with "The select-all has already been applied" either — whether it applied is the one thing this leg cannot ask.

Mutation-checked: collapsing it back to two states fails `reports a killed modern DEL as INTERRUPTED…` and nothing else.

packages/tool-server/src/utils/android-input.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Confirmed and fixed in 56d03bf78.

Printed the assembled message to be sure the contradiction is in one string and not across two branches:

> …reports 200 characters, more than this Android level can clear. **The select-all did not take on this field**, leaving one backspace per character… The count comes from the screen's view hierarchy, which reports an empty field's placeholder in the same attribute as its value and covers every window… The field MAY have been modified: … **though a count this long can also be another window's field or an empty field's placeholder.**

The cause is now reported as what was observed rather than what is inferred from it: `The field read back non-empty after the select-all, leaving one backspace per character as the only clear available…`. The comment above the branch says so for both halves now — "Neither the cause nor the mutation is ASSERTED on that arm".

Mutation-checked: restoring the flat wording fails `hedges the field's state when the residue is too long to backspace away` and nothing else.

packages/tool-server/src/tools/keyboard/index.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Confirmed and fixed in 56d03bf78.

The scoping really was carried only by the semicolon that the inserted sentence broke: before this branch the 158-backspace clause hung off "On Android levels older than `input keycombination` …" inside one sentence, and splitting it into its own sentence with the modern-refusal sentence in between left nothing holding it to those levels.

The paragraph is now ordered so each sentence carries its own scope: the read-back's degradations, then "The repair itself deletes backwards from end-of-LINE…", then "Levels older than `input keycombination` take that same end-of-LINE delete from the start, and there a field over 150 characters is refused…", then "**On those older levels** a length that cannot be read at all … falls back to a fixed 158 backspaces". A modern level can no longer be read into the fallback.

packages/tool-server/src/tools/keyboard/types.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Confirmed and fixed in 56d03bf78.

The unconditional clause is gone. The bullet now scopes the guarantee to what the code actually delivers and then enumerates every way out, including the two the previous wording missed entirely:

> …reads the field back afterwards and deletes whatever the select-all left behind, so a swallowed chord **over a single-line field it can read** does not leave that field one character shorter under a `cleared: true`. Three things escape the repair: the read-back rides the screen's view hierarchy, so a screen it cannot capture degrades to best-effort like iOS; it measures focused `EditText` nodes only, so a password box, a WebView input or a custom widget gets no repair at all; and the repair deletes backwards from end-of-LINE, so a multi-line field keeps what sits below the caret.

The second and third clauses are the two threads above; this docstring, the `clear` parameter and the verify leg's comment now carry the same three.

packages/tool-server/test/keyboard-clear.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Confirmed and fixed in 56d03bf78.

Proved by mutation before fixing: replacing `options.readHierarchy` with `undefined` in the verify read left **202/202 green**, while that mutation disables the repair whenever the devtools helper holds the UiAutomation connection — the `describe` → tap → `keyboard` order, where the dump comes back a bare `Killed`. Every one of the helper-path tests in "reading the hierarchy from the connection's holder" opens with `seedLegacyLevel()`, so all of them exercise the sizing read, never this one.

Added `reads the select-all back through the helper, not only through the dump`: a live helper answering with a `Monda` residue, no legacy seed. It asserts the helper is consulted once with `{ clearCache: true }`, that no `exec-out` dump is raced against it, and that the rescue run of 5 + 8 backspaces follows the chord and the DEL. Re-ran the same mutation with the test in place — it fails, alone (1 failed | 162 passed).

latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Review converged after three passes; the seven open threads from the earlier pass are fixed and replied to individually. Marking ready.

**Round 1 — `56d03bf78`** — the seven standing threads. Two were behaviour the docs claimed away, both reproduced in the unit harness before being touched: a multi-line residue is repaired only down to the end of the caret's line (`input keyevent 123` + 17 backspaces, `cleared: true`), and a focused editable the dump reports as anything but an `EditText` — a WebView input — gets no rescue at all. Both are documented limits now, matching the qualifier the older-level path already carried. The rest were messages or docs asserting more than the code establishes. Also covered the verify leg's `options.readHierarchy` source: dropping it left all 202 tests green while disabling the repair in exactly the state the option exists for.

**Round 2 — `496b38c94`** — one MEDIUM the earlier passes had recorded as an accepted LOW, which turned out to be worse than recorded. uiautomator renders an empty field's placeholder into the same `text` attribute it uses for a value, and the read-back turned that measurement from a *count* into a *predicate*. On the API 36 shape this file documents (`text="Search settings" … hint="Search settings"`, on a level that HAS `keycombination`), a clear that fully worked fired `input keyevent 123` plus 23 backspaces into the empty box — and a 171-character placeholder turned the same successful clear into a 400 with the `{ clear, text }` call's text never sent. `hint` is the discriminator and `parseUiAutomatorXml` already surfaces it. The read-back now reads a `text` equal to its own `hint` as empty; the sizing read must not, since under-measuring there truncates the field. Three rules forking the same way collapsed `maxDumps` + `unmeasurableIsUnreadable` into one `readBack` flag.

Alongside it: the tool description still promised "a rejected request never changes the field … a 400 leaves the field exactly as it was", which the rescue arm's own refusal contradicts; that refusal said the DEL "removes one character where it lands", true only where the chord was swallowed; the `cleared` escape list was short by one and understated another (an unmeasurable focused editable *anywhere* on screen discards the reading, not only one in the target field); `devtoolsHierarchyReader` still described a legacy-only consumer; and the per-device chain's rationale claimed a selection the DEL has already consumed.

**Round 3 — `6068be415`, `17489b428`** — a dead `text !== ""` guard the mutation sweep caught (an empty `text` measures 0 down the other arm too), the `cleared` docstring calling the end-of-LINE limit a "skip" when that repair runs and merely stops at the caret's line, and the two prose sites enumerating different escape sets.

Every guard is mutation-checked. Each of `readBack`'s three rules, the placeholder rule's `readBack` gate and its `text === hint` compare, the `residue > 0` and `count > MAX_DELETE_COUNT` gates, and all three `rescueFrom` message forks fail their own test and nothing else. Thirteen dump shapes were driven through the new rule — hint absent, `hint=""`, whitespace- and case-differing, entities decoding to the same string, a hinted password node, `content-desc` without `hint`, two focused nodes where one is a placeholder, astral characters — and the safe-direction property holds throughout: where the two cannot be told apart, it repairs.

Verified on `17489b428`: full tool-server suite 4543 tests across 346 files green, `tsc`, `typecheck:tests`, `eslint`, `prettier`, `knip` all clean. No docs update needed — `packages/docs` carries a one-line `keyboard` row and describes none of this.

packages/tool-server/src/tools/keyboard/index.ts
```diff
@@ -66,8 +66,8 @@ const zodSchema = z.object({
...
+        "On Android the read-back needs the screen's view hierarchy and measures focused `EditText` nodes only; where the screen cannot be captured, or the focused field is one it cannot read (a password box, a WebView input, a custom widget), the clear stays best-effort like iOS. The repair itself deletes backwards from end-of-LINE, so a multi-line field keeps what sits below the caret. Levels older than `input keycombination` take that same end-of-LINE delete from the start, and there a field over 150 characters is refused rather than partly deleted; the same refusal can fire on a level that HAS the subcommand when the read-back leaves more residue than backspaces are allowed to clear. On those older levels a length that cannot be read at all (a password field, or a screen the device would not capture) falls back to a fixed 158 backspaces — more than every length that path accepts, so a field only keeps its head past 158 characters."
```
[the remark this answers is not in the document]
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Confirmed and fixed in c3d31f510.

Verified against the code rather than the wording: the unreadable-length run is the same `input keyevent ${KEYCODE_MOVE_END} ${dels}` invocation as every other delete run (`android-input.ts:630`), so the 158 backspaces walk backwards from the end of the caret's LINE. On a multi-line field everything below that line survives untouched — the sentence two clauses earlier says exactly this about the repair, and this one contradicted it.

Now: "falls back to a fixed 158 backspaces **from that same end-of-LINE start** — more than every length that path accepts, so a single-line field keeps only its head past 158 characters, and a multi-line one keeps that head plus every line below the caret."

Note the finding was raised against `56d03bf78`; the surrounding paragraph has since been rewritten twice, but the clause you flagged survived both and was still wrong at HEAD.

PR #829, opened 2026-08-18
Add remote app installation

szdziedzic:
## Summary

- add an `install-app` tool that downloads and installs Android or iOS simulator builds from remote URLs
- support direct APKs and ZIP/tar artifacts, including nested CI and GitHub artifact wrappers
- resolve Android package names and iOS bundle identifiers automatically, including binary Android manifests and binary plists on non-macOS hosts
- install in place on local Android/iOS and remote iOS simulators while preserving `reinstall-app` clean-reset semantics
- document the URL-first workflow in Argent's device interaction skills

## Artifact and network hardening

- pin each HTTP(S) socket to the exact public address returned by the validation DNS lookup, including every redirect hop
- drop all caller-provided headers when a redirect changes origin and destroy redirect/error response bodies
- reject private, local, reserved, documentation, and mixed public/private DNS destinations
- move ZIP safety policy into `@argent/archive`
- stream regular ZIP entries with actual-output bounds; tightly bound buffered manifest and symlink reads
- reject ZIP traversal, nested symlink writes, unsupported entries, deceptive sizes, excessive entries, and archive bombs
- count actual gzip output before extraction instead of trusting the wrapping ISIZE trailer
- support safe multi-entry tarballs such as app + dSYM or APK + Gradle metadata
- share cancellable adb/simctl install primitives with `reinstall-app`, including explicit kill signals

## Motivation and user impact

CI and EAS builds are often available only as remote artifacts. Agents currently need a local build path before Argent can install them. With this change, an agent can pass a device ID and artifact URL to `install-app`, optionally with authentication headers, and receive the resolved package or bundle ID for a subsequent launch. Existing app data is retained when the platform supports an in-place update.

## Validation

- `npm run build`
- `npm run lint`
- test typechecks for `@argent/archive`, `@argent/tool-server`, and `@argent/tools-client`
- `@argent/archive`: 15 tests passed
- focused install-app, catalog, and interaction-message coverage: 103 tests passed
- full `@argent/tool-server`: 4,311 passed, 1 skipped across 349 test files
- full `@argent/tools-client`: 191 passed across 12 test files


filip131311:
Could you move the new skill guidance out of the shared skills and into a small dedicated one (e.g. `argent-remote-tool-server-interact`)?

Right now the PR adds the `install-app` section, the "Choosing the Right Tool" row, the Android/iOS gotchas to `argent-device-interact/SKILL.md`, plus two table rows in `argent-react-native-app-workflow/SKILL.md`. Installing from a remote artifact URL is a niche flow (CI/EAS builds, usually with a remote tool-server) – for most agents working against a local checkout this is noise in skills they load on every session, and it pushes the already long `argent-device-interact` description/table further.

Suggestion:

- new skill `argent-remote-tool-server-interact` holding the `install-app` section as written (the `{ "udid", "url" }` example, in-place vs `reinstall-app` semantics, `headers` only for authenticated URLs, public-host requirement, the device-only IPA gotcha)
- keep `argent-device-interact` and `argent-react-native-app-workflow` unchanged, or at most a one-line pointer to the new skill

The tool description itself already carries enough for discovery, so nothing is lost for agents that need it.

PR #830, opened 2026-08-18
feat(cli): run every flow `argent flow list` prints when `argent flow run` is given no argument

j-piasecki:
Subdirectories included, because the two commands now walk one function:
what one covers the other cannot miss, and the agreement is structural
rather than two rule sets kept in step by hand.

The merge settles a skip set they had disagreed on in both directions.
`list` descended into dot-directories and `node_modules` but skipped
`__baselines__`; a recursive `run` did the opposite. All three are skipped
now. That narrows `list`: a flow under `.argent/flows/.hidden/` is no longer
printed, though a path typed straight at it still runs.

An empty argument is refused rather than read as no argument at all.
`argent flow run "$FLOW"` with the variable unset resolves to the current
directory, which would batch every flow found there.


packages/argent-cli/src/flow.ts
```diff
@@ -188,6 +197,17 @@ export function parseRunArgs(argv: string[]): {
...
     }
+    // An empty token names nothing, and it must not be read as "no argument":
+    // it would resolve to the current directory and batch every flow found
+    // there. Reachable as `argent flow run "$FLOW"` with the variable unset.
+    // Ordered after the duplicate check, which owns `run checkout ""`: a flow
+    // was named, so "omit the argument" would be the wrong correction.
+    if (tok === "") {
+      throw new FlagParseException(
+        `flow run requires a flow name, a YAML file path, or a directory path; ` +
+          `omit the argument to run every flow in ${FLOWS_DIR}`
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `argent flow run "" checkout` reaches this throw rather than the duplicate check above it: the empty token is the *first* positional, so `out.flowRef` is still undefined and the ordering this comment describes does not apply. The operator is then told to "omit the argument to run every flow in .argent/flows" for a command in which a flow *was* named — following that correction batches the whole saved suite instead of running `checkout`. The mirrored spelling gets the other message.

Observed against the built CLI at 759dc2f36:

```
$ argent flow run "" checkout
Error: flow run requires a flow name, a YAML file path, or a directory path; omit the argument to run every flow in .argent/flows
$ argent flow run checkout ""
Error: unexpected argument ""; flow run accepts one flow name, YAML file path, or directory path
```

Same result calling `parseRunArgs(["", "checkout"])` and `parseRunArgs(["checkout", ""])` directly against `packages/argent-cli/dist/flow.js`.

packages/argent-cli/src/flow.ts
```diff
@@ -1299,20 +1311,34 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
...
+        `Pass a flow name, a YAML file path, or a directory path to run a flow kept elsewhere.`
+    );
+    return exitAfterFlush(2);
+  }
   if (isDirectory) {
-    return runFlowDirectory(resolvedPath, args, projectRoot, options);
+    return runFlowDirectory(resolvedPath, args, recursive, projectRoot, options);
   }
+  // The flag as typed, not the effective mode: this explains --recursive, and
+  // a bare run never passed it.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `argent flow run --recursive` is an argument-less run that does pass the flag — the test added at `packages/argent-cli/test/flow.test.ts:454` calls exactly that argv (`flow(["run", "--recursive"], opts)`) and names it "a bare `--recursive` run" — so the reason given here is not the reason this read is safe.

What actually makes the two readings equivalent at this point is that a defaulted ref can never reach this line: it returns either at the `isDirectory` dispatch or at the `defaulted && !isDirectory` arm above it. That matches what the suite says — replacing `args.recursive` with `recursive` at this branch in a sandbox copy leaves all 104 tests in `packages/argent-cli/test/flow.test.ts` green, because the distinction is not observable from any input. The cost is that someone later moving or relaxing either of the two arms above would be leaning on a stated invariant that does not hold.

packages/skills/skills/argent-create-flow/references/flow-yaml.md
```diff
@@ -192,7 +192,7 @@ A `run:` target is a YAML path resolved against the directory of the flow file c
 
 ## Snapshots and standalone runs
 
-`argent flow run <name> [--device <id>] [--platform ios|android|chromium|vega] [--update-baselines] [--output <dir>] [--json]` runs without an LLM and exits non-zero on failure.
+`argent flow run <name> [--device <id>] [--platform ios|android|chromium|vega] [--update-baselines] [--output <dir>] [--json]` runs without an LLM and exits non-zero on failure. `<name>` also takes a YAML path or a directory, which runs every flow in it; add `-r` to include subdirectories. Omit it to run every flow `argent flow list` prints, which is all of `.argent/flows`, subdirectories included.
```
latekvo, line 201: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `packages/docs/docs/reference/flow-yaml.mdx` on `main` mirrors this same command surface and still describes the pre-change behaviour. It landed in 36aa14a5f, after this branch's merge base, which is why it is not in the diff — but it is in the tree this merges into.

From `git show origin/main:packages/docs/docs/reference/flow-yaml.mdx`:

```
249:| `argent flow run <dir>`       | Run every flow in the directory, one after the other |
250:| `argent flow list`            | List the flow files in `.argent/flows`               |
258:| `-r, --recursive`    | With a directory, also run the flows in subdirectories. Dot-directories and `node_modules` are skipped |
```

The forms table (247-250) has no argument-less row, so the shipped docs list three `run` forms where the CLI now has four. Line 258 states the skip set as two directories; after this change it is three, since `__baselines__` joins it — that sentence becomes untrue as written. And the `flow list` row does not mention that the command now skips anything at all, which it did not before this change: on the same fixture, `origin/main`'s `flow list` printed `.argent/flows/.hidden/hiddenflow.yaml` and `.argent/flows/node_modules/depflow.yaml`, and this branch's does not.

`packages/docs/docs/features/flows.mdx:56` carries the same three-form enumeration ("The command runs one flow, a flow file in a different location, or all flows in a directory").

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Reviewed 759dc2f364e885f2595d6178d52083cd5c8e4678 in a worktree with the workspace and the `@swmansion/argent` bundles built, driving the real `argent flow …` entry point (`packages/argent/dist/cli.js`) against a live iOS simulator through an auto-started tool-server.

What I ran: `flow list` and a bare `flow run` on a tree carrying nested flows at two depths plus `__baselines__/`, `.hidden/` and `node_modules/` — both produced the same three paths in the same order, and the batch dispatched exactly those to `flow-execute`; the same tree through `flow run .argent/flows -r` and `flow run .argent/flows/` (identical dispatch); a symlinked flow file, a directory-symlink cycle and a broken symlink (the walk terminates, the link runs under its own name, the broken one is dropped); `.argent/flows` missing, empty, holding only skipped directories, present as a file, and unreadable; `flow run ""` and `flow run -- ""`; a batch containing an invalid YAML (exit 1) against an all-pass batch (exit 0); `--json` and `--output` on a bare run with same-stem flows in three directories. Each was compared against `origin/main`'s `flow.ts` built in a second tree — `flow list` there printed 5 paths on the fixture where this branch prints 3, which matches the narrowing the description states.

Mutation pass in a sandbox copy: each arm of `isSkippedFlowDir`, `args.recursive || defaulted`, the `!defaulted` in the stat catch, the `-r` hint's effective-mode read, the empty-token guard, the `defaulted && !isDirectory` arm, and `?? FLOWS_DIR` each fail between 1 and 6 tests, so the new behaviour is pinned. `prettier --check`, `eslint --max-warnings 0`, `typecheck:tests`, `@argent/cli` (468 tests) and `@swmansion/argent` (80 tests) all pass.

Three notes below. One problem found outside this diff, present on `main` and on the merge base alike, is filed separately as #937.

packages/argent-cli/src/flow.ts
```diff
@@ -188,6 +197,17 @@ export function parseRunArgs(argv: string[]): {
         `unexpected argument ${JSON.stringify(tok)}; flow run accepts one flow name, YAML file path, or directory path`
       );
     }
+    // An empty token names nothing, and it must not be read as "no argument":
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The rationale here pairs the hazard with the wrong reading. Reading `""` as "no argument" is the one thing the rest of this PR makes safe: `out.flowRef` stays undefined, `defaulted` becomes true, and `flowRef = args.flowRef ?? FLOWS_DIR` sends the run to `.argent/flows`, recursively. The current-directory batch comes from the opposite reading - storing `""` as the ref, where `resolveFlowRef("")` hands back `suppliedPath = ""` and `path.resolve(projectRoot, "")` collapses to cwd. The comment attributes the outcome the guard exists to prevent to the one reading that cannot produce it, and it contradicts the message thrown two lines below, which tells the operator that omitting the argument runs every flow in `.argent/flows`.

The same inverted claim is the stated justification in two of the added tests - `packages/argent-cli/test/flow.test.ts:254` ("Nor does it mean \"no argument\": that reading resolves to the current directory and batches every flow under it") and `test/flow.test.ts:496` ("Read as \"no argument\", the empty string resolves to the current directory and batches every flow under it") - so all three sites tell a maintainer that this PR's defaulted path lands in the current working directory.

Observed at 759dc2f364e885f2595d6178d52083cd5c8e4678 by running the two readings side by side against one fixture holding `cwdflow.yaml` at the project root plus `.argent/flows/saved.yaml` and `.argent/flows/suite/nested.yaml`, driving `flow(["run", ""], opts)` with the tools client mocked:

```
guard deleted, so "" is STORED as the flow ref
    -> exit 0, dispatched ["cwdflow.yaml"]

guard replaced by `if (tok === "") return;`, so "" is genuinely READ AS "no argument"
    -> exit 0, dispatched [".argent/flows/saved.yaml", ".argent/flows/suite/nested.yaml"]
```

The guard itself lands on the right side; it is the reason recorded for it that is reversed. (The ordering sentence two lines below is a separate thread.)

packages/argent-cli/src/flow.ts
```diff
@@ -1236,21 +1246,23 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
...
     throw err;
   }
-  if (!args.flowRef) {
-    console.error(
-      "argent flow run <flow|flow.yaml|dir> requires a flow name, a YAML file path, or a directory path."
-    );
-    printHelp();
-    return exitAfterFlush(2);
-  }
+  // A bare `run` means the default flows directory, so "run everything I have
```
latekvo, line 1139: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: "the two spellings cannot drift; what it adds is recursion" - recursion is not all it adds. The same diff makes `defaulted` gate two operator-visible branches below this comment: the `!defaulted` term at line 1317 excludes the argument-less spelling from the "Flow directory not found" arm that a typed `.argent/flows -r` gets, and `if (defaulted && !isDirectory)` at line 1326 is an arm that exists only for it. For one and the same resolved target, the spellings print three different things.

From an empty directory with no `.argent`, at 759dc2f364e885f2595d6178d52083cd5c8e4678:

```
$ argent flow run
No .argent/flows directory in the current working directory.
Pass a flow name, a YAML file path, or a directory path to run a flow kept elsewhere.
EXIT=2

$ argent flow run .argent/flows -r
Flow directory not found: /…/c6fix/.argent/flows
EXIT=2

$ argent flow run .argent/flows
Flow path must end in .yaml: .argent/flows
EXIT=2
```

The divergence is deliberate and two of the added tests pin it ("exits 2 with a directory message when a bare run finds no .argent/flows", "keeps the missing-directory message on a bare --recursive run"), so it is the sentence that is wrong rather than the code. It matters because it is the sentence a maintainer reaches for when deciding whether the `!defaulted` term at 1317 is redundant - and the comment on line 1336 is currently justified by a different reading of the same question (separate thread on that one).

packages/argent-cli/src/flow.ts
```diff
@@ -1236,21 +1246,23 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
...
+  // saved" costs no argument. It resolves and dispatches exactly as a typed
+  // path does, so the two spellings cannot drift; what it adds is recursion,
+  // because a bare run is the run half of `flow list` and must cover the very
+  // set that command prints. Only this form implies it: a typed directory
+  // still runs what the operator spelled and nothing below it without -r.
+  const flowRef = args.flowRef ?? FLOWS_DIR;
```
latekvo, line 1146: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Making the positional optional applies to every flag combination, including the one destructive flag `run` has. `argent flow run --update-baselines` - the same line with the flow name dropped, or a CI invocation whose `$FLOW` expanded away unquoted - now walks `.argent/flows` recursively and sends `updateBaselines: true` for every flow it finds. `flow-visual.ts:294-296` then does an unconditional `fs.copyFile(currentPath, baselinePath)`, so whatever the app happens to show becomes the new baseline for the whole project. At the merge base that argv was refused before any tool call.

The quoted spelling of the same accident is guarded: `argent flow run "" --update-baselines` exits 2 with "omit the argument to run every flow in .argent/flows" and makes no tool call. The unquoted one reaches the device.

Observed at 759dc2f364e885f2595d6178d52083cd5c8e4678 against merge base 5d8fc107994f40664c78d21715dbe730f84fc6db, fixture `.argent/flows/{checkout,login}.yaml` plus `.argent/flows/suite/deep.yaml`, driving `flow()` with the tools client mocked:

```
MERGE-BASE 5d8fc1079   flow run --update-baselines  ->  exit 2, 0 tool calls
    argent flow run <flow|flow.yaml|dir> requires a flow name, a YAML file path, or a directory path.

HEAD 759dc2f36         flow run --update-baselines  ->  exit 0, 3 tool calls
    checkout.yaml     updateBaselines=true
    login.yaml        updateBaselines=true
    suite/deep.yaml   updateBaselines=true
```

`argent flow run .argent/flows -r --update-baselines` already reached exactly those three on the merge base, so what changes is not the blast radius but that it is now the zero-argument spelling.

packages/argent-cli/src/flow.ts
```diff
@@ -1299,20 +1311,34 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
...
+    if (recursive && !fromName && !defaulted) {
       console.error(`Flow directory not found: ${resolvedPath}`);
       return exitAfterFlush(2);
     }
   }
+  // A defaulted ref addresses that directory or nothing at all, while every
+  // message below is written for a path the operator typed — a missing flows
+  // directory would surface as "Flow path must end in .yaml: .argent/flows",
+  // complaining about a string this run never wrote. Say what happened instead.
+  if (defaulted && !isDirectory) {
```
latekvo, line 1223: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `isDirectory` is left false by any stat failure, not only ENOENT, so this guard asserts absence for a `.argent/flows` that exists and holds flows but could not be stat'd - EACCES on `.argent`, ELOOP on a self-referential link, ENOTDIR on an ancestor. Nothing in the output mentions a read failure, and the recovery line then sends the operator to look for their flows somewhere else.

The nearest twins in the same function do separate the two: the single-file arm splits "Flow file not found" from "Could not read flow file", and `runFlowDirectory` says "Could not read flow directory" for the case where the stat succeeded and the readdir failed.

Observed at 759dc2f364e885f2595d6178d52083cd5c8e4678 with the built CLI, in a project holding `.argent/flows/smoke.yaml` with `chmod 000 .argent`:

```
$ argent flow run
No .argent/flows directory in the current working directory.
Pass a flow name, a YAML file path, or a directory path to run a flow kept elsewhere.
EXIT=2

$ argent flow run smoke              # same project, same permissions
Could not read flow file: /.../.argent/flows/smoke.yaml
EXIT=2

$ chmod 755 .argent && ls .argent/flows
smoke.yaml
```

`ln -s flows .argent/flows` (ELOOP) produces the same "No .argent/flows directory" text. The `flow list` half of this conflation is pre-existing and is filed separately as #960.

packages/argent-cli/src/flow.ts
```diff
@@ -1299,20 +1311,34 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
...
+    console.error(
+      `No ${FLOWS_DIR} directory in the current working directory.\n` +
+        `Pass a flow name, a YAML file path, or a directory path to run a flow kept elsewhere.`
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: One of the three forms this line offers cannot work in the state that produced the message, and the message itself is what establishes that state. `resolveFlowRef` turns a bare name into `path.join(FLOWS_DIR, ${ref}.yaml)` (line 969), so a name addresses `.argent/flows` and nothing else - it can neither reach elsewhere nor resolve here, because the line above just said that directory is absent.

Observed at 759dc2f364e885f2595d6178d52083cd5c8e4678, standing in a package subdirectory of a repo whose flows live at the root (`<repo>/.argent/flows/smoke.yaml`):

```
$ argent flow run
No .argent/flows directory in the current working directory.
Pass a flow name, a YAML file path, or a directory path to run a flow kept elsewhere.
EXIT=2

$ argent flow run smoke          # following the advice
Flow file not found: /…/repo/packages/app/.argent/flows/smoke.yaml
No flow named "smoke" is saved there — run `argent flow list` to see the saved flows.
EXIT=2

$ argent flow list               # following that hint
No .argent/flows directory in the current working directory.
EXIT=0
```

Three commands, three dead ends, and the correction that does work - `cd` to the project root, or pass the root's `.argent/flows` - is never named. Same result in an ordinary empty directory, so it is not monorepo-specific.

The two nearest twins in this file both say the opposite, and both are unchanged context: the help at line 108 ("a path is what reaches flows kept elsewhere") and the sibling recovery at line 1400 ("Names run `.argent/flows/<name>.yaml`; pass a path ending in .yaml to run a flow file kept elsewhere"). This line is the only place a name is described as reaching elsewhere.

New here. The merge base rejected a bare run with a usage line that asserts nothing about the filesystem and printed `printHelp()` underneath it, so the reader was handed the "a name … resolved from the current directory" entry along with the complaint; this message asserts the directory is absent and then recommends the one form that requires it, with no help below.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Two further sweeps of 759dc2f364e885f2595d6178d52083cd5c8e4678, in a worktree at that commit with the workspace and the `@swmansion/argent` bundle built, driving both the real CLI (`packages/argent/dist/cli.js`) and `flow()` directly with the tools client mocked, against a live iOS simulator where a run had to reach the device.

What I ran. Claims vs code over every comment, help-text line, test name and test comment the diff adds, plus the skills reference. Nearest twin against the other error arms in `flow()` - the single-file `ENOENT`/`Could not read` split, `runFlowDirectory`'s directory message, `flow list`'s catch, and the `looksLikeName` recovery at line 1400. Non-happy paths: `.argent/flows` missing, empty, present as a file, mode 000, behind a mode-000 `.argent`, a self-referential symlink, and holding only skipped directories; a batch stopped mid-way by a transport throw (exit 1, `ok:false`, remaining flows counted skipped); `--json` and `--output` on a bare recursive run with same-stem flows in different subdirectories. Inputs: `""`, `" "`, `--`, `-- ""`, `-r ""`, `.`, `./`, `.argent/flows`, `.argent/flows/`, `-`, `-- --`, and duplicate positionals, each read off the built `parseRunArgs`. Reachability was taken as a cross-product of `defaulted` × `fromName` × `recursive` × stat outcome, every cell executed and its exit code and stderr recorded. Consumers of the unified walker and of `FLOWS_DIR` were swept across the monorepo, including the tool-server's own flow tooling, and the operator-facing chain was walked end to end: run the command the CLI prints, then the command that one prints.

On an adversarial fixture carrying `__baselines__`, `.hidden`, `node_modules`, a nested `__baselines__`, a directory named `dir.yaml`, `UPPER.YAML`, `other.yml`, `bad@name.yaml`, a symlinked flow file, a broken symlink, a symlinked directory and a `..` directory-symlink cycle, `flow list`, a bare `flow run` and `flow run .argent/flows -r` produced the same five paths in the same order and the cycle terminated:

```
.argent/flows/dir.yaml/inside.yaml
.argent/flows/linked.yaml
.argent/flows/suite/deep/deepest.yaml
.argent/flows/suite/nested.yaml
.argent/flows/top.yaml
```

Mutation pass, one at a time against the 468-test `@argent/cli` suite: dropping `isSkippedFlowDir`'s dot-directory arm fails 5, its `node_modules` arm 5, its `__baselines__` arm 2; `args.recursive || defaulted` collapsed to `args.recursive` fails 3; removing `!defaulted` from the stat catch fails 2; removing the `defaulted && !isDirectory` block fails 2. `prettier --check`, `eslint --max-warnings 0`, `typecheck:tests` and the 468 tests all pass at this commit.

Every candidate was then re-run against the merge base `5d8fc107994f40664c78d21715dbe730f84fc6db` and dropped where it reproduced there unchanged. That removed the unpinned final `.sort()` - both merge-base sorts are equally deletable with the suite green, and the merge base already carried the same "pure function of the set of paths, never of the walk order" prose - the unpinned symlink-cycle guarantee, the batch labels at line 1141, `--device ""` (#937), and `flow list` masking `EACCES`, which is outside the diff and is filed separately as #960.

Six notes below, on top of the three already open. The docs-site pages remain as described in the existing thread on the skills reference.

packages/skills/skills/argent-create-flow/references/flow-yaml.md
latekvo, line 201: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This line now points the reader at the directory and argument-less forms, and two sentences further down the same section are wrong for them.

Line 203 - "`--output <dir>` writes failed baseline, current, and diff images under `<dir>/<flow>/` for CI artifact upload." A batch keys nested flows by their subdirectory: `runFlowDirectory` hands `path.join(outputBase, path.dirname(rel))` to the exporter, so `.argent/flows/suite/nested.yaml` lands under `<dir>/suite/nested/`.

Line 201 - "Baselines live under `.argent/flows/__baselines__/<flow>/`." The tool-server anchors the store at the root flow file's own directory (`flowsDir = path.dirname(canonicalPath)` in `flow-run.ts`, then `<flowsDir>/__baselines__/<flowName>`), so that same flow keeps its baselines at `.argent/flows/suite/__baselines__/nested/` - which is what this PR's own `isSkippedFlowDir` comment states: "`__baselines__` (at any depth, since nested flows keep their baselines beside themselves)".

Observed at 759dc2f364e885f2595d6178d52083cd5c8e4678 - a bare run over `.argent/flows/suite/nested.yaml` whose snapshot step fails, with `--output artifacts`:

```
artifacts/suite
artifacts/suite/nested
artifacts/suite/nested/.argent-flow-source
artifacts/suite/nested/home__ios-390x844-baseline.png
artifacts/suite/nested/home__ios-390x844-current.png
artifacts/suite/nested/home__ios-390x844-diff.png
```

Scope, stated plainly: both sentences were already wrong for `flow run <dir> -r`, which predates this branch, and the `--output` entry in `argent flow run --help` already carried the second layout ("a directory run keys nested flows as `<dir>/<subdir>/<flow>/`") - that block is unchanged context in this diff. What changes here is that this document now sends its reader to those forms, and it is the document an agent authoring a nested suite follows when wiring CI.

packages/argent-cli/src/flow.ts
j-piasecki: [LLM generated, feel free to ignore]

Fixed in 62c8ca4d. The guard now runs after the positional count check, so an empty token only earns the "omit the argument" message when it is the sole positional. `run "" checkout` gets `unexpected argument "checkout"`, matching the mirrored spelling, and the comment's ordering sentence was corrected to say why the count goes first.

packages/argent-cli/src/flow.ts
j-piasecki: [LLM generated, feel free to ignore]

Correct, and fixed in 62c8ca4d. The comment now names the reading that actually reaches the current directory - `""` stored as the flow ref, where path.resolve collapses it to cwd - rather than the "no argument" reading, which lands in .argent/flows. Both test comments carrying the same inverted claim were fixed in the same commit.

packages/argent-cli/src/flow.ts
j-piasecki: [LLM generated, feel free to ignore]

Fixed in 84f0719d. Dropped "the two spellings cannot drift" and added that `defaulted` also aims the failure messages below, since a run that named no path cannot be told one was not found. That is the sentence a maintainer needs when weighing whether the `!defaulted` term is redundant.

packages/argent-cli/src/flow.ts
j-piasecki: [LLM generated, feel free to ignore]

Fixed in fd347721. A defaulted run with `--update-baselines` now exits 2 before any tool call and names the spelling that does work (`argent flow run .argent/flows -r --update-baselines`). `defaulted` is an exact signal, so no heuristic is involved. The reference page's flag row picked up the refusal in e60b47e5.

packages/argent-cli/src/flow.ts
j-piasecki: [LLM generated, feel free to ignore]

Fixed in 74398178. The stat error is now kept past the catch, and a non-ENOENT failure returns "Could not read flow directory: <path>" with no recovery line, since the flows have not moved. ENOENT keeps the original message, and `.argent/flows` present as a file keeps it too. Both cases are pinned by tests. `flow list` left alone, per #960.

packages/argent-cli/src/flow.ts
j-piasecki: [LLM generated, feel free to ignore]

Fixed in 1e38e208. The line now reads "Pass a YAML file path or a directory path to run a flow kept elsewhere, such as the project root's .argent/flows, or cd to that root first." The name form is gone, and the correction that does work is named.

packages/argent-cli/src/flow.ts
j-piasecki: [LLM generated, feel free to ignore]

Right - `flow run --recursive` is argument-less and does pass the flag. Fixed in b2e0926f with the reason that holds: a defaulted ref cannot reach that line, having returned at the directory dispatch or the `defaulted` arm before it. That is the invariant a later change to either arm has to preserve.

packages/skills/skills/argent-create-flow/references/flow-yaml.md
j-piasecki: [LLM generated, feel free to ignore]

Both pages are covered now. `reference/flow-yaml.mdx` picked up the argument-less row, the corrected `-r` description, the three-directory skip set and the `flow list` skip note in the PR commit itself. `features/flows.mdx:56` got the fourth run form in c5dad643.

j-piasecki: [LLM generated, feel free to ignore]

Both sentences fixed in 9825f5d4, at both sites - the same wrong claim was on `docs/reference/flow-yaml.mdx` too. Baselines are now described as living beside the flow file, and `--output` keeps `<dir>/<flow>/` as the base with the nested keying attached to directory and bare runs only, matching the wording `argent flow run --help` already used.

packages/argent-cli/src/flow.ts
```diff
@@ -108,23 +110,30 @@ multi-platform launch auto-detects a device instead. Pass --device to attach to
...
+  list                       List runnable YAML paths in .argent/flows,
+                             subdirectories included and the same directories
+                             skipped
 
 Options (run):
   --device <id>          Device id to run against (auto-detected when omitted)
   --platform <p>         ios | android | chromium | vega — narrow auto-detection
-  --update-baselines     Write/refresh screenshot baselines instead of diffing
+  --update-baselines     Write/refresh screenshot baselines instead of diffing.
+                         Refused on a bare run: pass .argent/flows -r to update
```
hubgan, line 135: The recovery this line hands out silently destroys a visual assertion when the tree keeps a `run:` fragment under `.argent/flows`.

Reproduced on iOS (booted sim `02835BA4`, app `com.argent.settlelab`). Tree:
```
.argent/flows/a-reset.yaml        # e2e, ends on the app's home screen
.argent/flows/z-nav.yaml          # e2e, navigates to the modal, then: run: shared/on-modal.yaml
.argent/flows/shared/on-modal.yaml  # fragment, executionPrerequisite + snapshot: modal-screen
```

1. `argent flow run z-nav --update-baselines` seeds `.argent/flows/__baselines__/z-nav/modal-screen__ios-750x1334.png` — the modal screen. `argent flow run z-nav` then passes.
2. `argent flow run` fails: the bare run also executes `shared/on-modal.yaml` standalone, and its snapshot resolves to `.argent/flows/shared/__baselines__/on-modal/`, which composed seeding never creates.
3. Running exactly what this line suggests — `argent flow run .argent/flows -r --update-baselines` — reports `PASS — 3 flows` and writes that missing baseline. But in sorted order the fragment runs immediately after `a-reset.yaml`, so what gets captured is the **home** screen, not the modal.

The two baselines for the same `snapshot: modal-screen` step now hold different screens:
```
27a208666d6e261214af8b9b9296076a  49712 bytes  .argent/flows/__baselines__/z-nav/modal-screen__ios-750x1334.png        (modal)
b58217db62336686dc6e5a3105c21776  76321 bytes  .argent/flows/shared/__baselines__/on-modal/modal-screen__ios-750x1334.png (home)
```
I opened both PNGs to confirm which screen each holds.

From then on `argent flow run` exits 0 forever, and `snapshot: modal-screen` compares the home screen against a home-screen baseline. The visual check no longer checks anything, and nothing warns at any point.

packages/argent-cli/src/flow.ts
```diff
@@ -1122,21 +1136,34 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
...
+  // A bare `run` means the default flows directory, so "run everything I have
+  // saved" costs no argument. It resolves and dispatches exactly as a typed
+  // path does; what it adds is recursion, because a bare run is the run half
+  // of `flow list` and must cover the very set that command prints. Only this
+  // form implies it: a typed directory still runs what the operator spelled
+  // and nothing below it without -r. `defaulted` also aims the failure
+  // messages below: a run that named no path cannot be told one was not found.
+  const flowRef = args.flowRef ?? FLOWS_DIR;
+  const defaulted = args.flowRef === undefined;
+  const recursive = args.recursive || defaulted;
```
hubgan, line 1148: Recursion by default pulls `run:` fragments into the batch, and a fragment runs against whatever the previous flow left on screen — so a bare run's verdict depends on the alphabetical filename of an unrelated flow.

Reproduced on iOS (sim `02835BA4`, app `com.argent.settlelab`). Tree: `a-reset.yaml` (ends on home), `m-nav.yaml` (navigates to the modal, then `run: shared/on-modal.yaml`), `shared/on-modal.yaml` (fragment awaiting an element that exists only on the modal). Each flow passes when run by name.

```
argent flow run                     -> exit 0
mv m-nav.yaml z-nav.yaml
argent flow run                     -> exit 1    ([2/3] shared/on-modal.yaml fails)
mv z-nav.yaml m-nav.yaml
argent flow run                     -> exit 0
```

The only change between runs is the name of a flow the fragment does not reference. Named `m-nav` it sorts before `shared/`, so the fragment inherits the modal screen; named `z-nav` it sorts after, so the fragment inherits the home screen left by `a-reset`.

Control: `argent flow run .argent/flows` (no `-r`) exits 0 under both names, and `argent flow run .argent/flows -r` exits 1 — so this arrives with the implied recursion, on the form that takes no arguments and is the one CI will use.

Two things visible in the same transcripts: in the passing arrangement the fragment passes only because an unrelated flow happened to sort first, so a green bare run is not evidence its assertion means anything; and the fragment executes twice per bare run — once inline through `run:`, once standalone — which for a fragment with side effects is a duplicated mutation per composing root.

packages/argent-cli/src/flow.ts
```diff
@@ -1122,21 +1136,34 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
...
-    printHelp(jsonStream);
-    return exitAfterFlush(2);
+  // A bare `run` means the default flows directory, so "run everything I have
+  // saved" costs no argument. It resolves and dispatches exactly as a typed
+  // path does; what it adds is recursion, because a bare run is the run half
+  // of `flow list` and must cover the very set that command prints. Only this
+  // form implies it: a typed directory still runs what the operator spelled
+  // and nothing below it without -r. `defaulted` also aims the failure
+  // messages below: a run that named no path cannot be told one was not found.
+  const flowRef = args.flowRef ?? FLOWS_DIR;
```
hubgan, line 1146: Replacing the missing-argument guard with a default also removes the backstop for argv accidents where a value-taking flag swallows the intended flow name. `parseCommandArgs` consumes the next non-dash token as the flag's value, so an unset **unquoted** variable placed before the positional leaves no positional at all — and the run now becomes a whole-suite run that reports success.

Reproduced against the branch build, project holding `a.yaml`, `b.yaml`, `sub/c.yaml`:
```
ART_DIR=            # unset, as in CI
argent flow run --output $ART_DIR checkout --platform ios --device <udid>
[1/3] a.yaml
[2/3] b.yaml
[3/3] sub/c.yaml
PASS — 3 flows: 3 passed, 0 failed, 0 skipped        exit 0
```
On the merge base the identical argv prints `requires a flow name, a YAML file path, or a directory path` and exits 2, running nothing.

Same shape for `--device $UDID checkout` (every flow is dispatched with `device: "checkout"`; the summary prints `PASS (started on checkout)`) and `--platform $P checkout`.

This is the accident class the empty-positional guard and the `--update-baselines` refusal were added for, reached from the other side of a flag. The destructive door does hold — I verified `run --output $UNSET checkout --update-baselines` is still refused — but flows perform real UI actions, and "run one flow" quietly becoming "run every saved flow" is reported as a pass.

packages/argent-cli/src/flow.ts
```diff
@@ -1073,12 +1090,9 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
...
-      // order.
-      const paths = (await collectRunnableFlowPaths(dir))
-        .sort()
-        .map((rel) => path.join(FLOWS_DIR, rel));
+      const paths = (await collectRunnableFlowPaths(dir, true)).map((rel) =>
+        path.join(FLOWS_DIR, rel)
+      );
       if (paths.length === 0) console.log("No flows found in .argent/flows");
       else console.log(paths.join("\n"));
     } catch {
```
hubgan, line 1098: This catch collapses every walker failure into "No `.argent/flows` directory in the current working directory." and exits 0, including the case the run side of this PR now distinguishes. With `.argent/flows` at mode 000 holding `smoke.yaml`:

```
argent flow list   -> No .argent/flows directory in the current working directory.   exit 0
argent flow run    -> Could not read flow directory: <abs>/.argent/flows             exit 2
```

The directory is right there; `list` sends the operator looking elsewhere, and a script reading its exit code sees success. Given the premise that a bare run is the run half of `flow list`, this is the one place the two halves still disagree about what "missing" means.

packages/argent-cli/src/flow.ts
```diff
@@ -1176,26 +1203,49 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
...
   }
+  // A defaulted ref addresses that directory or nothing at all, while every
+  // message below is written for a path the operator typed — a missing flows
+  // directory would surface as "Flow path must end in .yaml: .argent/flows",
+  // complaining about a string this run never wrote. Say what happened instead.
+  if (defaulted && !isDirectory) {
+    // A non-ENOENT stat read nothing: the flows may be saved right there, and
+    // the recovery below would send the operator hunting somewhere else.
+    if (statError !== undefined && statError.code !== "ENOENT") {
+      return fail(`Could not read flow directory: ${resolvedPath}`, 2);
```
hubgan, line 1227: Only EACCES matches this wording. ENOTDIR reaches it too: with `.argent` a regular file, `argent flow run` prints `Could not read flow directory: <cwd>/.argent/flows` (verified against the branch build). There no flows directory can exist at all, so the situation is structural rather than a read that failed — and unlike the ENOENT arm just below, this one carries no recovery line.

packages/argent-cli/src/flow.ts
```diff
@@ -191,6 +202,17 @@ export function parseRunArgs(argv: string[]): {
       `unexpected argument ${JSON.stringify(positionals[1])}; flow run accepts one flow name, YAML file path, or directory path`
     );
   }
+  // An empty token names nothing, and it must not be stored as the flow ref:
+  // path.resolve collapses "" to the current directory, and the run batches
+  // every flow found there. Reachable as `argent flow run "$FLOW"` with the
+  // variable unset. Checked after the count so `run "" checkout`, which did
+  // name a flow, is reported as one argument too many instead.
+  if (positionals[0] === "") {
```
hubgan, line 210: The same unset-variable accident is unguarded for the value-taking flags, and there it is silently accepted rather than refused. `--output ""`, `--device ""` and `--platform ""` all parse, and every consumer downstream is a truthiness test.

Verified against the branch build:
- `argent flow run a --output "$UNSET" --platform ios --device <udid>` runs, exits 0, and writes no failure artifacts at all — a CI upload step then finds nothing.
- `argent flow run a --device "$UNSET" --platform ios` drops the device pin the docs tell operators to always set and falls back to auto-detection; the summary prints `Flow "a"` with no device.
- `--platform` is declared `{ kind: "value" }` with no `choices`, though the help enumerates `ios | android | chromium | vega` and the other enum options in this CLI declare theirs.

hubgan (commented):
Reviewed the CLI dispatch, the merged directory walker, and the docs, and exercised the branch build end to end on a booted iOS simulator. The findings are inline. The two that matter both come from the bare run executing `run:` fragments standalone; the rest are message and coverage notes.

packages/argent-cli/src/flow.ts
```diff
@@ -1176,26 +1203,49 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
...
+    return fail(
+      `No ${FLOWS_DIR} directory in the current working directory.\n` +
+        `Pass a YAML file path or a directory path to run a flow kept elsewhere, ` +
+        `such as the project root's ${FLOWS_DIR}, or cd to that root first.`,
+      2
+    );
+  }
   if (isDirectory) {
     if (args.jsonStream) {
       return fail("--json-stream supports a single flow; directory runs are not supported.", 2);
```
hubgan, line 1238: A bare `argent flow run --json-stream` lands here and is told "directory runs are not supported" for an invocation that named no directory — the thing the `defaulted` arm below was added to avoid ("a missing flows directory would surface as ... complaining about a string this run never wrote"). It also offers no recovery, unlike the sibling `--update-baselines` refusal, which names the bare form and hands back a command. Verified: both the NDJSON error record and the stderr line say "directory runs" for an argument-less run.

packages/argent-cli/src/flow.ts
```diff
@@ -95,6 +95,8 @@ Run a YAML flow without an LLM in the loop. \`run\` takes any of these forms:
            path instead); a path is what reaches flows kept elsewhere, and
            what \`argent flow list\` prints for nested ones
   a dir    Every flow in that directory, run sequentially
+  nothing  Every flow in .argent/flows, subdirectories included: exactly the
```
hubgan, line 98: This promises the bare run covers exactly what `flow list` prints, but the batch labels those flows with a spelling that is not re-runnable. A bare run prints `[1/2] suite/nested.yaml`; `flow list` prints `.argent/flows/suite/nested.yaml`. Pasting the batch label back gives `Flow file not found: <cwd>/suite/nested.yaml`, exit 2.

For a typed `run <dir> -r` the operator supplied the base and can reconstruct the path. A bare run supplies none, and that label is the only identifier a failing flow gets — `renderSummary` emits no path.

packages/argent-cli/src/flow.ts
```diff
@@ -1122,21 +1136,34 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
...
+  const defaulted = args.flowRef === undefined;
+  const recursive = args.recursive || defaulted;
+
+  // parseRunArgs refuses the quoted spelling; unquoted, `run $FLOW
+  // --update-baselines` with the variable unset arrives here as a bare run.
+  if (defaulted && args.updateBaselines) {
+    return fail(
+      `flow run --update-baselines requires a flow name, a YAML file path, or a directory path; ` +
+        `it overwrites the baselines of every flow it runs, so the target must be one you named.\n` +
+        `Run \`argent flow run ${FLOWS_DIR} -r --update-baselines\` to update every flow saved there.`,
```
hubgan, line 1156: This is the only "run this command" hint in the file that interpolates a path without going through `shellQuoteArg` (the ones at the `..` guard and the saved-flow hint all do), and what it interpolates is `FLOWS_DIR` — `.argent\flows` on Windows, a separator `SHELL_SAFE_ARG` deliberately excludes. Under Git Bash / MSYS with a win32 node, the suggested `argent flow run .argent\flows -r --update-baselines` reaches argv as `.argentflows` and lands on the `Flow directory not found` arm, so the tool's own suggested fix dead-ends.

Two smaller things about the same message: the guard fires before any stat, so in a directory with no `.argent/flows` the suggested command fails as well; and a run that already passed `-r` is told to add `-r`.

packages/argent-cli/src/flow.ts
```diff
@@ -726,42 +748,61 @@ async function isRunnableFlowFile(filePath: string): Promise<boolean> {
 }
 
 /**
- * Every runnable flow path under `dir`, relative to it. `flow run` binds its
- * name contract to the filename alone — intermediate directory names are
- * unconstrained — so a YAML at any depth is runnable and a non-recursive
- * listing would hide paths `run` accepts. Two deliberate exclusions:
+ * Directories a recursive walk never enters, none of them a place flows are
```
hubgan, line 751: "at any depth" is stronger than what the code does: the skip list is applied only to entries discovered during the walk, never to the walk root. `argent flow run .argent/flows/__baselines__ -r` prints `[1/1] checkout/sneaky.yaml` and executes it (verified against the branch build).

Defensible as "run what the operator spelled" — it is the rule dot-directories always had — but the comment asserts the stronger invariant and no test pins either reading.

packages/docs/docs/reference/flow-yaml.mdx
```diff
@@ -234,7 +234,7 @@ A fragment whose `run:` chain reaches a launch cannot declare `executionPrerequi
 
 A snapshot compares the current screen, or the frame of `cropOn`, with a stored baseline. A missing baseline fails the step. A mismatch above `maxMismatch` percent fails the step. The default `maxMismatch` is 0.5. A `cropOn` element whose size changed fails the step.
 
-Baselines live in `.argent/flows/__baselines__/<flow>/`. Argent keys a baseline by platform and capture geometry, plus the selector for `cropOn`. Run `argent flow run <name> --update-baselines` to write the baselines from a known-good state. Review each baseline before you commit it.
+Baselines live in `__baselines__/<flow>/` next to the flow file. Argent keeps the baselines of a nested flow in `.argent/flows/<subdir>/__baselines__/<flow>/`. Argent keys a baseline by platform and capture geometry, plus the selector for `cropOn`. Run `argent flow run <name> --update-baselines` to write the baselines from a known-good state. Review each baseline before you commit it.
```
hubgan, line 237: "a nested flow" is ambiguous here, and false under the meaning this page established fifteen lines above: the Composition section calls a `run:` target "a nested fragment or end-to-end flow". A `run:`-composed fragment's baselines anchor to the **root** flow, not beside the fragment.

Verified on iOS: with `.argent/flows/checkout.yaml` holding `run: shared/login.yaml` and `shared/login.yaml` holding `snapshot: home`, `argent flow run checkout --update-baselines` wrote `.argent/flows/__baselines__/checkout/home__ios-750x1334.png` — the root's directory and the root's stem. Nothing was written under `shared/`.

The sentence holds only for a flow file in a subdirectory that is itself the run root. That distinction is also the mechanism behind the bare-run failure I filed on `flow.ts`, so it is worth being unambiguous about.

packages/skills/skills/argent-create-flow/references/flow-yaml.md
```diff
@@ -198,15 +198,15 @@ A `run:` target is a YAML path resolved against the directory of the flow file c
...
```
hubgan, line 201: The synopsis carries `[--update-baselines]` and the next clause says the argument may be omitted, but that combination now exits 2. The docs-site copy records the refusal; this agent-facing reference does not, so an agent following it hits an unexplained exit 2 on the highest-blast-radius flag in the CLI.

The same sentence also says the bare form is "all of `.argent/flows`, subdirectories included" without naming the directories now skipped — and dot-directories and `node_modules` are a change to what `flow list` prints, not just to what a run covers.

Related, outside this diff: `packages/skills/skills/argent-qa-flows/SKILL.md:100` still names only `argent flow run <name> --platform <platform>` for CI, in the skill whose entire subject is a saved regression suite — the case this PR's new form exists for.

packages/skills/skills/argent-create-flow/references/flow-yaml.md
```diff
@@ -198,15 +198,15 @@ A `run:` target is a YAML path resolved against the directory of the flow file c
...
 
 Do not use a snapshot as the only proof of navigation, persistence, data, accessibility state, logs, or network behavior. Avoid unstable timestamps, live data, ads, animation, and device drift. First establish deterministic state, identity, and readiness.
 
-Baselines live under `.argent/flows/__baselines__/<flow>/` and are keyed by platform and full-capture geometry; `cropOn` also contributes its selector. Seed from a known-good state with `--update-baselines`. Inspect every baseline and require user review. Do not commit it yourself. Baseline creation or update is not a test pass. Never update a baseline only to make a diff pass. The default `maxMismatch` is 0.5 percent.
+Baselines live in `__baselines__/<flow>/` beside the flow file, so a nested flow keeps its own under `.argent/flows/<subdir>/__baselines__/<flow>/`. They are keyed by platform and full-capture geometry; `cropOn` also contributes its selector. Seed from a known-good state with `--update-baselines`. Inspect every baseline and require user review. Do not commit it yourself. Baseline creation or update is not a test pass. Never update a baseline only to make a diff pass. The default `maxMismatch` is 0.5 percent.
```
hubgan, line 207: Same wording as the docs-site copy: this file also uses "nested" for `run:` composition (lines 195-196), and a composed fragment's baselines key under the root flow rather than beside the fragment. Under that reading the sentence is wrong.

packages/argent-cli/test/flow.test.ts
```diff
@@ -1168,6 +1385,45 @@ describe("argent flow run", () => {
     expect(logs.join("\n")).toBe("No flows found in .argent/flows");
   });
 
+  it("runs exactly the set `flow list` prints, from one shared walker", async () => {
```
hubgan, line 1388: This is the only place the run-side `__baselines__` narrowing is pinned, and it pins it through the bare run.

The spelling whose behavior actually changed for existing users is the typed `argent flow run <dir> -r`: before this PR the directory-run walker had no `__baselines__` rule at all, so a YAML under `__baselines__` was executed. The `argent flow run <dir>` fixture (line ~1826) contains no `__baselines__` directory, so nothing asserts the new skip at that entry point.

packages/argent-cli/test/flow.test.ts
```diff
@@ -417,14 +425,223 @@ describe("argent flow run", () => {
...
+      `Pass a YAML file path or a directory path to run a flow kept elsewhere, ` +
+        `such as the project root's ${path.join(".argent", "flows")}, or cd to that root first.`
+    );
+    expect(errs.join("\n")).not.toContain("must end in .yaml");
+    expect(toolsClientMock.callTool).not.toHaveBeenCalled();
+  });
+
+  // Skipped as root / on Windows, where a mode-000 directory is still readable
+  // (see canDenyRead).
+  it.skipIf(!canDenyRead)(
```
hubgan, line 497: `canDenyRead` is `platform !== "win32" && getuid() !== 0`, so this is the only coverage of the new non-ENOENT stat arm and it vanishes on Windows and on a root CI runner — common in Docker images. `.argent` as a regular file reaches the same arm with ENOTDIR and needs no chmod; the test below it covers `.argent/flows` being a file, where stat succeeds and the arm is not taken at all.

Separately on this file: the bare-run tests reuse the `beforeAll` `tempRoot` and assert exact `callTool` counts (`[1/1] saved.yaml`), and a bare run now walks `tempRoot/.argent/flows` recursively. Nothing wrote there before; any future fixture placed under it silently breaks those assertions from 900 lines away.

PR #843, opened 2026-08-19
feat: support iOS launch arguments

szdziedzic:
## What changed

- add an optional `launchArgs` array to the `launch-app` and `restart-app` tool schemas
- forward those arguments as discrete argv values after the bundle ID for local and remote Apple simulator launches
- cover launch and restart behavior on both `xcrun simctl` and `sim-remote simctl`

## Why

Some simulator workflows need process arguments at launch time, for example to override `UserDefaults` without modifying or rebuilding an app:

```json
{
  "udid": "remote:<udid>",
  "bundleId": "dev.example.app",
  "launchArgs": [
    "-EXDevMenuIsOnboardingFinished",
    "1",
    "-EXDevMenuShowsAtLaunch",
    "0"
  ]
}
```

The remote simctl helper already accepted trailing launch arguments, but the public lifecycle tools did not expose or forward them. This made the capability unavailable through Argent, including hosted EAS Simulator sessions.

## Impact

Callers can now supply launch-time feature flags and preference overrides through both `launch-app` and `restart-app`. Existing calls remain unchanged when `launchArgs` is omitted. Non-Apple platforms ignore the field as documented.

## Validation

- `npm run build`
- `npm run lint`
- `npm test -w @argent/tool-server -- --run test/ios-launch-args.test.ts test/launch-restart-system-app.test.ts test/launch-restart-tvos.test.ts test/restart-app-android-activity.test.ts` (20 tests)

### DIFF
--- a/packages/tool-server/src/tools/launch-app/index.ts
+++ b/packages/tool-server/src/tools/launch-app/index.ts
@@ -41,6 +41,12 @@ const zodSchema = z.object({
     .describe(
       "Android-only: fully-qualified Activity name (e.g. `.MainActivity` or `com.example/com.example.MainActivity`). If omitted on Android, the app's default launcher activity is used. Ignored on iOS / Chromium."
     ),
+  launchArgs: z
+    .array(z.string())
+    .optional()
+    .describe(
+      'Apple simulator-only: process arguments forwarded verbatim after the bundle id in `simctl launch`. Useful for launch-time feature flags and UserDefaults overrides such as `["-FeatureFlag", "YES"]`. Ignored on Android / Chromium / Vega.'
+    ),
 });
 
 type Params = z.infer<typeof zodSchema>;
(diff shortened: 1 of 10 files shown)


filip131311 (approved):
Tested this live on a merged-with-main build of the branch tool-server, against a booted iPhone 17 (iOS 26.5) and Apple TV (tvOS 26.5) sim:

- `restart-app` with `["-AppleLanguages","(fr)"]` → process argv is `Preferences -AppleLanguages (fr)` and Settings renders as "Réglages", so the UserDefaults override really lands. Same on tvOS.
- Everything after the bundle id goes to the app as argv: `--stdout=/tmp/x` / `--console` are *not* parsed as simctl flags (no file written, no hang). Spaces, unicode, `--k=v` and empty-string args are preserved as discrete argv entries.
- `launchArgs: []` / omitted is byte-identical to today's behaviour; non-array is rejected by zod; Android and Chromium ignore the field as documented; schema is exposed on both tools. The 20 unit tests pass.

Two findings, neither blocking:

**1. `launch-app` silently drops `launchArgs` when the app is already running.**
`simctl launch` on a running process only foregrounds it. With Settings already open, `launch-app … launchArgs:["-AppleLanguages","(de)"]` returned `{launched: true}` in ~130 ms and the process kept its old argv. Given the schema text says "forwarded verbatim", an agent will believe the flags applied — and "re-launch the running app with flags" is the common path for this feature. simctl has the knob for it: `--terminate-running-process` (placed before `<device>`). Suggest adding it whenever `launchArgs` is non-empty, in `localSimctl.launch` / `makeIosImpl` and in `sim-remote`'s `simctlLaunch` (worth confirming the remote helper accepts it); failing that, the description should say to use `restart-app` to apply args to a running app.

**2. Flows can't carry `launchArgs`.**
The flow `launch:` directive calls `restart-app` with `{bundleId}` only (`flow-run.ts` `runLaunch`), so a YAML e2e flow can't pass iOS launch args, while the chromium form already has `launch: {path, args}`. A generic `tool: restart-app` step still works. Fine as a follow-up.

Nit: `restart-app` says "Ignored on Android / Vega" while `launch-app` says "Android / Chromium / Vega" — both are correct for their own platform sets, just inconsistent.

PR #817, opened 2026-08-17
fix(installer): say why a global install cannot work on a Nix toolchain

latekvo:
## The report

A colleague on a Nix-managed project:

```
●  Running: npm install -g @swmansion/argent@0.20.0
npm error code EACCES
npm error syscall rename
npm error path /nix/store/04fc23…-nodejs-22.14.0/lib/node_modules/@swmansion/argent
■  Update failed: Error: Command failed: npm install -g @swmansion/argent@0.20.0
```

npm derives its global prefix from the node binary's own location. Under Nix that is a store path - mode `0555`, root-owned, and a read-only *mount* on NixOS and nix-darwin - so `npm install -g` there dies with EACCES for the developer running it, and what root writes Nix undoes at the next rebuild or garbage-collect. `argent init --global` and `argent update` ran the install anyway and handed the user npm's stack trace with no way forward.

## The preflight

`global-prefix.ts` answers "where would a global install land, and can this user write there", by asking the package manager itself (`npm root -g`, `pnpm root -g`, `yarn global dir`, `bun pm bin -g`) and probing the nearest existing ancestor of the path the install has to create.

That path is `<global root>/@swmansion/argent`, so what gets probed is its **parent** - the `@swmansion` scope directory once anything is installed under it, the global root before that. It is the directory npm renames inside, which is the syscall the bug report failed on. Probing the global root alone passes a prefix whose `@swmansion/` came from an earlier `sudo npm i -g`:

```
$ ls -ld lib/node_modules lib/node_modules/@swmansion
drwxr-xr-x  …/lib/node_modules
dr-xr-xr-x  …/lib/node_modules/@swmansion
$ npm install -g @types/semver@7.5.1        # same layout, unrelated scope
npm error code EACCES
npm error syscall rename
```

`init` and `update` preflight it and print the cause plus the two ways out:

```
npm cannot update @swmansion/argent globally: its global package directory is
inside the read-only Nix store.
  /nix/store/04fc23…-nodejs-22.14.0/lib/node_modules
Nix owns that directory — a sudo install into it is undone by the next rebuild or garbage-collect.

  Point npm at a writable prefix, then retry:
    npm config set prefix "$HOME/.npm-global"
    export PATH="$HOME/.npm-global/bin:$PATH"  (add to your shell profile)

  Use @swmansion/argent per project instead — no global directory needed:
    argent init --local
```

`update` fails **before** the confirm prompt and before the tool-server teardown, so no server is stopped for an install that was never going to happen.

[...]

PR #844, opened 2026-08-19
fix(installer): refuse a run that cannot ask instead of exiting 0 having done nothing

### LINKED ISSUE #842: argent init: a prompt with no terminal on stdin exits 0 without installing or configuring anything
`argent init` with no terminal on stdin (a Dockerfile line, a CI step, a `RUN` layer, an agent driving it with stdin closed) renders the install-mode menu, never receives an answer, and the process **exits 0 having done nothing** — no install, no configs, no install record. A caller that checks the exit code is told the setup succeeded.

### Repro

A writable global prefix (nothing Nix-specific — this is the plain path), a project with a `package.json`, stdin closed:

```
$ cd /private/tmp/writable-prefix/proj
$ env -i HOME=$H PATH=$PREFIX/bin:/usr/bin:/bin TERM=dumb node .../dist/cli.js init --no-telemetry < /dev/null
●  Telemetry disabled (--no-telemetry).
◆  How should argent be installed?
│  ● Globally (recommended) (Installs the argent command on your PATH; shared across every project)
│  ○ This project only
│  ↑/↓ to navigate • Enter: confirm
└
$ echo $?
0
$ ls ~/.argent .mcp.json node_modules 2>&1   # nothing was written or installed
```

With stdin an open-but-silent pipe the run hangs instead, indefinitely.

### Cause

`@clack/prompts`' `select` promise never settles on EOF, so `init` unwinds through no branch at all: node drains the event loop and exits 0. `parseInitArgs` (`packages/argent-installer/src/init-args.ts`) derives `nonInteractive` from `--yes`/`-y` only; nothing in `packages/argent-installer/src` consults `process.stdin.isTTY`, so every prompt in the flow (`promptInstallMode`, `chooseAdapters`, `chooseScope`, `configureAllowlist`, the skills step) has the same exposure.

### Impact

The failure is silent and reports success. Anyone scripting `argent init` (image builds, provisioning, agents) gets a green step and an unconfigured machine; the correct outcome is either to run the `--yes` defaults or to fail with a message naming `--yes`.

Found while reviewing #817, which adds a `process.stdin.isTTY` guard to the one prompt it introduces on the `--global` path. The prompts listed above are untouched by it.

latekvo:
Closes #842.

`@clack/prompts` reads keystrokes straight from stdin. Answers piped in do settle its prompts — the moment that input runs out, the promise never settles. A command that asks anyway unwinds through no branch at all: node drains the event loop and the process **exits 0 having done nothing** — no install, no configs, no install record. A caller that checks the exit code is told the setup succeeded.

```
$ node dist/cli.js init --no-telemetry < /dev/null
●  Package: @swmansion/argent@0.20.0
◆  How should argent be installed?
│  ● Globally (recommended)  ○ This project only
└
$ echo $?
0
$ ls ~/.argent .mcp.json node_modules   # nothing was written or installed
```

`argent uninstall` does the same at its "Remove argent configuration from this workspace?" confirmation, `argent update` at three more prompts, and `argent link` / `argent unlink` at four.

## What changed

`canPromptUser()` is checked before the first question each command cannot skip, and the refusal names the flags that answer it.

| Command | Prompt | Without a terminal |
| --- | --- | --- |
| `init` | all of them (mode, editors, scope, allowlist, skills, first-run telemetry notice) | refuses at the top, exit 2, before any install or config write |
| `uninstall` | the workspace confirmation | refuses, exit 2 |
| `update` | the update confirm | fails that target, exit 1 |
| `update` | which install to act on, when a global and a local coexist | fails, exit 1 |
| `update` | the stale-config sweep's cross-project removals | no confirmer: the sweep reports the entries instead of removing them, exactly as `--yes` already did |
| `link` | host and port | refuses, exit 2 |
| `link` | replacing an existing link | refuses, exit 2 |
| `link` | the retry menu after a failed pre-flight | takes the `--yes` branch: fails on the unreachable server, exit 1 |
| `unlink` | the removal confirmation | refuses, exit 2 |

A run that needs no answer is untouched: `argent update` with nothing to install still refreshes configuration and exits 0, and `link --host h --port n --no-verify` still saves and exits 0.

`--yes` is not the same choice at every command — for `uninstall` it removes a global install the interactive default *keeps* — so each refusal says what taking it will do there. The help lines match: `--yes` is unconditionally required with no terminal for `init`, only when there is something to confirm for `update`.

Two telemetry codes carry the refusals that happen with telemetry live: `UPDATE_NEEDS_TERMINAL`, `UNINSTALL_NEEDS_TERMINAL`. `init` refuses before consent is resolved, so it tracks nothing.

Docs: `fundamentals/installation.mdx` no longer promises a terminal-less `init` the read-only-prefix remedies it now refuses before reaching.

Stacked on #817, whose per-prompt terminal check this replaces — the one check now sits at init's entry, so the recovery prompt in `install-runner.ts` is back to a plain `nonInteractive` test.

<details>
<summary>Verified</summary>

Sandboxed HOME + npm prefix, real CLI bundle (`packages/argent/dist/cli.js`), stdin closed:

- `init`, `init --global`, `uninstall` → exit 2, one-line message, `~/.argent` and the project left as they were.
- `init --local --yes --from <tarball>` → exit 0, devDependency materialized, MCP configs written for 7 clients.
- `uninstall --yes` → removes the package and its configs, exit 0.
- `update --version 0.20.0` on an up-to-date install → exit 0, "Update complete", refresh ran.
- `update` with an update actually pending → exit 1 and the message, nothing installed.
- `link`, `link --port 3001`, `link --host h` (port still unanswered), `unlink` → exit 2, `~/.argent/link.json` untouched.
- `link argent://tok@h:3001`, `link https://host`, `link --host h --port n --no-verify`, `link --host h --yes --no-verify`, `unlink --yes` → exit 0, link written / removed.
- `link http://unreachable:9999` with verification on → exit 1 on the pre-flight, no link written.
- Under a real pty (`expect`), `init` renders the install-mode menu as before.

Unit: the installer, cli, argent, telemetry and registry suites (1941 tests), plus the full root suite in six shards. Every guard is mutation-checked — deleting it, deleting either arm of `link`'s condition, flipping an exit code, dropping the `--yes` escape, reverting the per-site remedy, moving `uninstall`'s install-mode resolution back below the confirm, or making `canPromptUser()` constant each fails at least one test. Scenario A2 of the Nix suite asserts the refusal (exit 2, no state left behind); scenario A covers the same command with `--yes`, which still reaches the store and reports the Nix block.

Note: this PR is stacked, so `unit-tests`, `lint`, `format`, `knip` and `nix-install-e2e` are all `branches: [main]` and do not run on it. Those were run locally instead — all clean, plus `npx docusaurus build`.

</details>


packages/argent-installer/src/terminal.ts
```diff
@@ -0,0 +1,15 @@
+import pc from "picocolors";
+
+// clack reads keystrokes straight from stdin: behind a pipe or a closed
+// descriptor its promise never settles, so a command that asks anyway unwinds
+// through no branch at all and exits 0 having done nothing.
+export function canPromptUser(): boolean {
+  return process.stdin.isTTY === true;
+}
```
latekvo, line 11: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: This guard is consulted only by the installer commands — `argent link` and `argent unlink` still ask their questions without it and hit exactly the failure mode described in the comment above. Reproduced on this PR's head (e652cd98) against a bundled `dist/cli.js` with stdin closed: `link` renders its host prompt and exits 0 with no `~/.argent/link.json` written (`packages/argent-cli/src/link.ts:237`, also :258, :411, :449), and `unlink` prints its "Remove link to …?" confirm and exits 0 leaving the link intact (`packages/argent-cli/src/link.ts:525`). A scripted call to either is told it succeeded while nothing happened — the same silent-success shape #842 describes, surviving one package over from where it is now fixed.

latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Ran the review moves on the three commits (claims vs code, nearest twin, non-happy paths, inputs, reachability both ways, what outlives the call, plus the absence pass), ran the argent-installer/telemetry/registry suites (996/997 — the one failure is the opencode `globalPath` probe picking up this machine's own ~/.config/opencode/opencode.jsonc), and replayed the flows E2E against a bundled dist/cli.js with stdin closed: `init` refuses at the top with exit 2 and writes nothing, matching every claim in the description. One finding from the absence pass, posted inline.

packages/argent-installer/src/terminal.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 332de93c8.

Reproduced both on 2f1974879 through a driver shaped like `packages/argent/src/cli.ts` (`main().catch()`, not top-level await, so the loop drains instead of tripping node's unsettled-TLA exit 13): `link` exit 0 with no `~/.argent/link.json`; `unlink` exit 0 with the link intact.

The repro also turned up a fourth site the comment did not name: `link --host h --no-verify` still asks for the port (`flags.port === null && !flags.yes` falls through to `promptPort`), so it hangs the same way. Guarded sites are now the host/port pair (one check before the intro), the replace-existing confirm, and `unlink`'s confirm — all exit 2 — plus the failed pre-flight, which now takes the `--yes` branch and fails on the unreachable server rather than opening a retry menu nobody can answer.

Same reproductions after the fix: exit 2, nothing written, nothing removed; `link --host 10.0.0.9 --port 3001 --no-verify` (asks nothing) still saves and exits 0, and `unlink --yes` still removes. `packages/argent-cli/test/link-no-terminal.test.ts` pins all four — every guard deleted, `canPromptUser()` pinned to a constant, and exit 2 changed to 1 each kill at least one test.

packages/argent-installer/src/update.ts
```diff
@@ -623,6 +639,16 @@ export async function update(args: string[]): Promise<void> {
...
     let targets: InstallMode[];
     if (decision.kind === "prompt") {
+      if (!canPromptUser()) {
+        p.log.error(
+          noTerminalMessage("argent update", "to take the defaults without being asked.")
+        );
+        await trackPackageAction("update_failed", updateStartTime, false, UPDATE_NEEDS_TERMINAL);
+        await failUpdateTelemetry(UPDATE_NEEDS_TERMINAL);
+        p.outro(pc.red("Update failed."));
+        process.exit(1);
```
latekvo, line 649: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The same condition and the same remedy exit 2 from `init` (init.ts:66) and from `uninstall` (uninstall.ts:377), but 1 from both of update's refusal sites. Ran all three against the built `dist/cli.js` with stdin closed in a sandboxed HOME: `init --global` -> 2, `uninstall` -> 2, `update` -> 1.

For `update` that status is also what UPDATE_REGISTRY_CHECK_FAILED, UPDATE_GLOBAL_PREFIX_UNWRITABLE, UPDATE_TOOLSERVER_STOP_FAILED and a partially failed multi-target run produce, so a wrapper that wants "this run needs --yes, retry with it" can key on the status for two of the three commands and has to parse stderr for the third.

packages/argent-installer/src/init.ts
```diff
@@ -52,6 +58,14 @@ export async function init(args: string[]): Promise<void> {
     let version = getInstalledVersion() ?? "unknown";
     p.log.info(`${pc.dim("Package:")} ${PACKAGE_NAME}@${version}`);
 
+    // Without --yes every path below has a prompt in it, so a run with nobody
+    // to ask is refused before it installs or configures anything — and before
+    // consent is resolved, which is why the refusal itself is not tracked.
+    if (!parsed.nonInteractive && !canPromptUser()) {
```
latekvo, line 64: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `init` is the command #842 is filed about, and its refusal is the one of the three that emits nothing: the guard sits above `track("installation:cli_init_start")` at init.ts:88, and there is no INIT_NEEDS_TERMINAL in FAILURE_CODES, while `uninstall` and `update` each emit a dedicated code from the same condition.

The ordering reason is real — the refusal precedes `resolveTelemetryConsent`, so on a genuine first run there is no consent to collect under. But on every later run consent is already recorded (`hasShownFirstRunNotice()` true, or DO_NOT_TRACK / ARGENT_TELEMETRY set), `resolveTelemetryConsent` returns `{kind:"skipped"}` without asking, and the repeated Dockerfile/CI invocation still reports nothing. The funnel therefore shows the two rarer instances of this failure and not the common one.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Second pass over the branch after 332de93c8 and 58883e843. Two findings left, both low: an exit-code split between the three refusals, and `init`'s refusal being the only one that emits no telemetry.

PR #851, opened 2026-08-19
feat: support named screenshots

szdziedzic:
## Summary

- add an optional human-readable `name` to the `screenshot` tool
- include named screenshots in lifecycle event messages
- stage the captured PNG under its normalized name so both local and remote/EAS Simulator artifact paths are readable
- validate names against path traversal and control characters
- document named capture usage in the device interaction skill

## Why

Screenshot artifacts currently inherit opaque backend-generated filenames such as `624543000-1786034937633.png`. Those names also appear in Argent's event logs, which makes captures difficult to identify in EAS Simulator runs.

Named captures now produce readable files such as `checkout-complete.png` while preserving the existing behavior for callers that omit `name`.

## Validation

- `npm run build`
- `npm test -w @argent/tool-server -- --run test/screenshot-tool.test.ts test/interaction-messages.test.ts test/tool-input-schema-contract.test.ts` (87 tests)
- targeted ESLint and Prettier checks

### DIFF
--- a/packages/tool-server/src/tools/screenshot/index.ts
+++ b/packages/tool-server/src/tools/screenshot/index.ts
@@ -1,9 +1,10 @@
 import { execFile } from "node:child_process";
+import { copyFile, mkdtemp } from "node:fs/promises";
 import { promisify } from "node:util";
 import * as os from "node:os";
 import * as path from "node:path";
 import { z } from "zod";
-import type { Registry, ToolCapability, ToolDefinition } from "@argent/registry";
+import type { ArtifactStore, Registry, ToolCapability, ToolDefinition } from "@argent/registry";
 import { simulatorServerRef, type SimulatorServerApi } from "../../blueprints/simulator-server";
 import { chromiumCdpRef, type ChromiumCdpApi } from "../../blueprints/chromium-cdp";
 import { resolveDevice } from "../../utils/device-info";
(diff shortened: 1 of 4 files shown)


packages/tool-server/src/tools/screenshot/index.ts
```diff
@@ -63,6 +76,51 @@ interface Result {
...
+
+/**
+ * A filename override on the artifact handle is enough for a remote download,
+ * but a co-located client uses `hostPath` directly. Stage named captures under
+ * their real basename too, so both paths (including EAS Simulator collection)
+ * observe the human filename instead of the backend's timestamp/UUID.
+ */
+async function stageNamedScreenshot(capturedPath: string, name?: string): Promise<string> {
+  if (!name) return capturedPath;
+  const dir = await mkdtemp(path.join(os.tmpdir(), "argent-named-screenshot-"));
```
filip131311, line 106: **Leaked temp dirs.** This `mkdtemp`s a fresh dir per call and nothing ever removes it — after 5 named captures I had 5 orphaned `argent-named-screenshot-*` dirs in `os.tmpdir()`. The sim-server's own `simserver-*/media` dir is lifecycle-managed; this one isn't. Suggest one staging dir per tool-server process that is cleaned on shutdown, or registering the dir with a cleanup hook.

packages/tool-server/src/tools/screenshot/index.ts
```diff
@@ -63,6 +76,51 @@ interface Result {
...
+  const namedPath = path.join(dir, namedScreenshotFilename(name));
+  await copyFile(capturedPath, namedPath);
```
filip131311, line 108: **`copyFile` doubles disk.** A `scale: 1.0` capture is several MB and now exists twice. `rename` with a `copyFile` fallback (for the cross-device case) would avoid that — the sim-server doesn't need its copy after export.

filip131311 (commented):
Built the tgz from this branch and ran it live (iOS 26.5 sim, tool-server + `argent mcp` stdio, event log on): named captures land as `Checkout-complete.png` in both `filename` and `hostPath`, `.png` isn't doubled, traversal/control-char names 400 at zod, the no-name path is unchanged, and the lifecycle messages show up as expected. Two non-blocking nits inline.

PR #855, opened 2026-08-19
feat(flow): stream run progress as NDJSON

schicaelmott:
## Summary

- add `argent flow run --json-stream`
- stream flow progress as JSON lines
- finish with a JSON result or error

## Why

`--json` prints only after a flow finishes. `--json-stream` lets scripts and CI see live progress.

## Why this is a separate PR

This is useful without the custom-tools feature in #856.

## Validation

- `npm test -w @argent/cli -- flow.test.ts` — 106 tests passed
- `npm run typecheck:tests -w @argent/cli`
- `npm run build -w @argent/cli`

### DIFF
--- a/packages/argent-cli/src/flow.ts
+++ b/packages/argent-cli/src/flow.ts
@@ -1568,8 +1568,8 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
       baseUrl
     );
   } catch (err) {
-    if (args.jsonStream) writeJsonStreamError(err);
-    throw err;
+    if (!args.jsonStream) throw err;
+    return fail(err instanceof Error ? err.message : String(err), 2, err);
   }
 
   if (args.jsonStream) {
--- a/packages/argent-cli/src/flow.ts
+++ b/packages/argent-cli/src/flow.ts
@@ -1029,12 +1029,12 @@ async function collectFlowFiles(dir: string, recursive: boolean): Promise<string
  * remotely callable, but reject CLI routing that cannot guarantee the
  * shared filesystem. This deliberately rejects even single-file flows that
  * could run remotely — the CLI cannot tell them apart without parsing the
- * flow. Prints the recovery hint and returns false when remote routing is
+ * flow. Returns the refusal with its recovery hint when remote routing is
  * configured.
  */
-async function requireLocalToolServer(): Promise<boolean> {
+async function requireLocalToolServer(): Promise<string | undefined> {
   const routing = await getResolvedToolsUrl();
-  if (routing.source === "none") return true;
+  if (routing.source === "none") return undefined;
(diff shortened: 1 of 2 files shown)


packages/argent-cli/src/flow.ts
```diff
@@ -1479,6 +1511,10 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
   let liveIndex = 0;
   const onStepReport = (event: unknown): void => {
     const s = event as StepReport;
+    if (args.jsonStream) {
+      writeJsonStreamRecord({ event: "progress", data: s });
```
hubgan, line 1526: This record contains the step report before the artifact fields get their final form.

`exportAndResolveArtifacts` changes each artifact to a path string. It operates only on the final report at line 1558. As a result, `artifacts.current` is an object in this record and a string in the `result` record. A program cannot use one accessor for the field.

The object also contains a `hostPath`. This path points to the temporary cache of the tool-server. If the user gives `--output`, only the `result` record contains the durable copy. A program that acts on the progress records can get a path to a file that the server deletes.

The comment at line 1088 says that every output mode must show the durable location. `renderArtifactLines` keeps only string values for the same reason.

One snapshot step on an iOS simulator gives these two values:

```
progress  artifacts.current -> object  /var/folders/.../simserver-nHC8GF/media/99374000-....png
result    artifacts.current -> string  /.../out/snap/home__ios-1206x2622-current.png
```

packages/argent-cli/src/flow.ts
```diff
@@ -1506,23 +1542,34 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
     // --output copies are fetched, below.
     report = resp.data as FlowReport;
   } catch (err) {
+    if (args.jsonStream) writeJsonStreamError(err);
```
hubgan: The error record covers three failures only. They are this failure, the invalid report at line 1552, and the artifact export failure at line 1565.

Thirteen other failures write text to stderr and write no record. They are at lines 1263, 1272, 1317, 1334, 1340, 1351, 1384, 1416, 1424, 1432, 1452, 1497 and 1500. A wrong subcommand at line 1243 does the same.

These failures include a wrong flow name, a missing file, a directory path, a bad file extension, and the refusal of remote routing. The combination of `--json` and `--json-stream` also does this.

A program that reads the NDJSON output sees an empty stream for these failures. The program must read the text on stderr to find the cause.

hubgan (commented):
Thank you for this contribution. The new `--json-stream` flag fills a real gap. `--json` writes nothing until the run stops, so a program could not read step data while a run is in progress.

I tested the flag against a real iOS simulator with a local tool-server. The records stay valid JSON on stdout in each condition that I tested, and the envelope agrees with the NDJSON format of the tool-server.

I found two possible problems with the new flag. Both comments are on the lines below.

PR #934, opened 2026-08-23
test(run-sequence): pin the one observation a sequence returns

latekvo:
## What

`run-sequence` tells agents:

> One screenshot is captured automatically after the whole sequence (not per step)

The MCP-side half of that promise is already pinned — argent-mcp's auto-screenshot tests name `run-sequence` explicitly, twice. The tool-server half is pinned by nothing: this adds the gate that pins it, plus the renderer-side pin in argent-mcp. **No behavior change.**

## Why it is unpinned

The steps stay silent only because no tool `ALLOWED_TOOLS` admits returns an image or an artifact handle. Nothing checks:

- which tools the allow-list admits,
- that every admitted name still resolves to a registered tool (a stale name still reads as allowed and stays advertised in the description, but every step naming it dies on `ToolNotFoundError` and halts the sequence),
- that none of them declares `outputHint: "image"`.

Admitting a frame-producing tool would go unnoticed: a sequence result reaches `toMcpContent` whole, and its `materializeArtifacts` walk collects handles from anywhere inside it, `steps[].result` included, appending one image block per handle. That walk recurses into arrays with `Promise.all`, so those frames arrive in **completion order, not step order** — the agent could not even tell which one is the screen the sequence ended on.

Both halves are verified against the code, not assumed:

- Putting an artifact handle in one step's result makes `toMcpContent` emit an image block for it, confirming the walk reaches `steps[].result`.
- Materializing three step handles whose fetches complete fastest-last yields `[s2, s1, s0]` — exactly reversed from step order.

## What is gated

**tool-server** — `test/run-sequence-observation-gate.test.ts`, against the real registry (`createRegistry()`, no device). `ALLOWED_TOOLS` is exported for it:

- the allow-list admits exactly the reviewed set of 14 step tools
- every name still resolves to a registered tool
- none of them declares `outputHint: "image"`

**argent-mcp** — `test/content.test.ts`:

- a multi-step sequence result renders with no frame of its own, against a `fetchImpl` that resolves and is asserted never to be called (without a resolving fetch the assertion would hold equally for a sequence whose step frames merely failed to download, since `materializeArtifacts` swallows a failed download and renders the handle as nothing)
- the positive control that pin rests on: a handle on `steps[].result` does surface as an image block, so the walk ceasing to reach steps cannot turn the pin vacuous

The membership of `run-sequence` in `AUTO_SCREENSHOT_TOOLS` was already pinned by existing assertions and needed no new test.

## Verification

Every test was mutation-checked — each fails on the broken code, not just passes on the good:

| Mutation | Fails |
| --- | --- |
| add `screenshot` to `ALLOWED_TOOLS` | reviewed-set + `outputHint` checks |
| rename an allow-list entry the registry lost | reviewed-set + registration checks |
| give a step result an image artifact handle | the frameless-sequence check, on both its fetch-call and its image-block assertion |
| stop `materializeArtifacts` recursing into arrays | the step-carried-handle check, and nothing else in the file |

Locally green: `prettier --check`, `tsc --build`, `typecheck:tests` for both touched packages, the full argent-mcp suite (82 tests), and the gate test (3 tests).

## Scope note

The allow-list is spelled out rather than counted, so adding a step tool means answering the question the file exists to ask — does it put a frame of its own in front of the agent? The gate deliberately sits on the allow-list rather than on the renderer: a tool that would break the guarantee is caught before it can be dispatched as a step, rather than having its frames filtered afterwards.

### DIFF
--- a/packages/argent-mcp/test/content.test.ts
+++ b/packages/argent-mcp/test/content.test.ts
@@ -255,6 +255,31 @@ describe("toMcpContent with artifact ctx", () => {
     expect((result[1] as { text: string }).text).toMatch(/^Saved: .*shot\.png$/);
   });
 
+  // The renderer's half of the one-observation guarantee; the allow-list half
+  // is tool-server's run-sequence-observation-gate.test.ts. The result is
+  // walked whole, steps included, and every artifact handle found becomes its
+  // own image block — so a sequence holds at one frame only while its steps
+  // report none.
+  it("renders a multi-step sequence result without a frame of its own", async () => {
+    const result = await toMcpContent(
+      {
+        completed: 3,
+        total: 3,
+        steps: [
+          { tool: "gesture-swipe", result: { swiped: true, timestampMs: 1 } },
+          { tool: "keyboard", result: { typed: "hello", keys: 0 } },
+          { tool: "gesture-tap", result: { tapped: true, timestampMs: 2 } },
+        ],
+      },
+      undefined,
+      { toolsUrl: "http://remote:3001", deviceId: "DEV-1" }
+    );
+
+    expect(result.filter((b) => b.type === "image")).toEqual([]);
+    expect(result).toHaveLength(1);
+    expect((result[0] as { text: string }).text).toContain("gesture-tap");
+  });
+
   it("rewrites non-image artifacts to local paths inside the JSON result", async () => {
     const result = await toMcpContent(
       { exportedFiles: { cpu: artifactHandle("cpu1", "cpu.xml", "application/xml") } },
--- a/packages/argent-mcp/test/content.test.ts
+++ b/packages/argent-mcp/test/content.test.ts
@@ -261,6 +261,11 @@ describe("toMcpContent with artifact ctx", () => {
   // own image block — so a sequence holds at one frame only while its steps
   // report none.
   it("renders a multi-step sequence result without a frame of its own", async () => {
+    // A fetch that resolves. Left to the global one, a step-carried handle
+    // would fail to download and be swallowed, so the assertions below would
+    // hold just as well for a sequence whose frames were merely unreachable.
+    const fetchImpl = vi.fn(fetchReturning([...PNG_SIGNATURE, 0x42]));
+
     const result = await toMcpContent(
       {
         completed: 3,
@@ -272,12 +277,34 @@ describe("toMcpContent with artifact ctx", () => {
         ],
       },
       undefined,
-      { toolsUrl: "http://remote:3001", deviceId: "DEV-1" }
+      { toolsUrl: "http://remote:3001", deviceId: "DEV-1", fetchImpl }
     );
 
+    expect(fetchImpl).not.toHaveBeenCalled();
     expect(result.filter((b) => b.type === "image")).toEqual([]);
     expect(result).toHaveLength(1);
     expect((result[0] as { text: string }).text).toContain("gesture-tap");
+  });
+
+  it("materializes an artifact handle carried on a step result", async () => {
+    // Keeps the pin above honest: the walk really does reach steps[].result,
+    // so a step tool that returned a frame would be inlined mid-sequence.
+    const pngBytes = [...PNG_SIGNATURE, 0x42];
+    const result = await toMcpContent(
+      {
+        completed: 1,
+        total: 1,
+        steps: [
+          { tool: "gesture-tap", result: { image: artifactHandle("s0", "shot.png", "image/png") } },
+        ],
+      },
+      undefined,
+      { toolsUrl: "http://remote:3001", deviceId: "DEV-1", fetchImpl: fetchReturning(pngBytes) }
+    );
+
+    expect(result.filter((b) => b.type === "image")).toEqual([
+      { type: "image", data: Buffer.from(pngBytes).toString("base64"), mimeType: "image/png" },
+    ]);
   });
 
   it("rewrites non-image artifacts to local paths inside the JSON result", async () => {
--- a/packages/tool-server/test/run-sequence-observation-gate.test.ts
+++ b/packages/tool-server/test/run-sequence-observation-gate.test.ts
@@ -1,0 +1,70 @@
+/**
+ * Gate on the guarantee run-sequence advertises to agents: "One screenshot is
+ * captured automatically after the whole sequence (not per step)".
+ *
+ * Nothing in run-sequence enforces it. The capture is appended once, by the MCP
+ * layer, because `run-sequence` is in that layer's AUTO_SCREENSHOT_TOOLS
+ * (pinned in argent-mcp's auto-screenshot test). The steps stay silent only
+ * because no tool the allow-list admits returns an image or an artifact handle.
+ *
+ * A step that returned one would be inlined: the sequence result reaches
+ * `toMcpContent` whole, and its `materializeArtifacts` walk collects handles
+ * from anywhere inside it — `steps[].result` included — appending one image
+ * block per handle. That walk recurses into arrays with `Promise.all`, so those
+ * frames arrive in completion order rather than step order, leaving the agent
+ * unable to tell which is the screen the sequence ended on. Hence the gate sits
+ * on the allow-list: such a tool has to be caught before it can be a step.
+ */
+import { describe, it, expect } from "vitest";
+import { ALLOWED_TOOLS } from "../src/tools/run-sequence";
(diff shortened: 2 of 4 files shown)


packages/tool-server/test/run-sequence-observation-gate.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+  });
+
+  it("admits no tool that renders as a frame of its own", () => {
+    const registry = createRegistry();
+    for (const id of ALLOWED_TOOLS) {
+      // `outputHint: "image"` declares a result that IS a picture — screenshot
+      // alone today. Such a tool also hands back the artifact behind it, so
+      // admitting one puts a mid-sequence frame in front of the agent.
+      expect(registry.getTool(id)?.outputHint, `${id}.outputHint`).not.toBe("image");
```
latekvo, line 67: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Low: this checks `outputHint !== "image"`, but the render walk collects any `__argentArtifact`-marker object from anywhere in the result regardless of hint (`packages/argent-mcp/src/content.ts:73`, `packages/argent-tools-client/src/artifacts.ts:508-514`). So the gate is a proxy for "admits no tool that returns a handle", not proof of it. Verified today's members are safe: none of the 14 imports `requireArtifacts` (only screenshot, screenshot-diff, screen-recording, profiler/*, flow-visual do), and step failure paths emit plain strings, so no reachable consequence now. Impact is limited to a future admission that returns a non-image-hint handle - which the name-spelled-out expectation above forces through review anyway.

latekvo (commented):
Verification run for this PR: all nine code claims checked against source (all verified); four mutation checks each caught by the intended assertion; completion-order claim proven empirically; knip re-run clean.

One residual finding (low): this assertion checks `outputHint !== "image"`, but the frame walk collects any `__argentArtifact`-marker object from anywhere in the result regardless of hint (`packages/argent-mcp/src/content.ts:73`, `packages/argent-tools-client/src/artifacts.ts:508-514`). So the gate is a proxy for "returns no handle", not proof of it. Today the claim holds - none of the 14 members mints handles (`requireArtifacts` appears only in screenshot, screenshot-diff, screen-recording, profiler/*, flow-visual), and error paths push plain strings - so this only bites a hypothetical future admission, which the spelled-out expectation array forces through review at that moment.

packages/tool-server/test/run-sequence-observation-gate.test.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Re-screened on the branch. The mechanism claim is accurate — `materializeArtifacts` keys on the `__argentArtifact` marker regardless of hint, and `toMcpContent` (content.ts:73) walks the whole sequence result including `steps[].result`. But a grep of tool-server confirms handle-minting is confined to 7 modules (screenshot, screenshot-diff, flow-visual, screen-recording-stop, profiler/*), none in ALLOWED_TOOLS, and screenshot alone carries the image hint — so there is no reachable consequence for any admitted tool today. Dismissing as an accepted gate-strength limitation rather than fixing: any static strengthening (source-text import scanning) would be more brittle than the risk it covers, and the spelled-out EXPECTED_ALLOWED_TOOLS array already forces every future admission through review where this exact question gets asked.

latekvo:
[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free]: Full review pass complete; branch rewritten to `5a728e09`.

**Existing finding (outputHint vs `__argentArtifact` walk):** re-screened on the code — mechanism accurate, no reachable consequence today (handle-minting confined to screenshot/screenshot-diff/screen-recording-stop/profiler/*/flow-visual, none admitted). Dismissed on-thread as an accepted gate-strength limitation; the spelled-out expectation array forces future admissions through review.

**New finding this pass (MEDIUM, reproduced before fixing):** the argent-mcp addition claimed dropping `run-sequence` from `AUTO_SCREENSHOT_TOOLS` was caught only by a loop that iterates that same set. Falsified by mutation: with the entry removed, the pre-existing "returns true for unified interaction tools" test (which hard-codes `"run-sequence"` at auto-screenshot.test.ts:214) and the containsSecretPlaceholder assertion (:279) both fail. The added test was a third duplicate resting on a false premise. Fix: removed it and rewrote the branch history so neither commit message nor PR body carries the falsified coverage claim anymore.

Also verified: all six claims-vs-code checks (AUTO_SCREENSHOT_TOOLS membership + prefix normalization; steps dispatch via in-process `registry.invokeTool` with no path to the MCP capture layer; none of the 14 allowed tools declares image hint or mints handles; `Promise.all` array recursion → completion-order frames; exact description string at run-sequence/index.ts:107; `createRegistry()` side-effect-free); absence pass (flow-execute intentionally escapes the guarantee and docs agree; deleting the gate file leaves nothing else pinning the allow-list); second full pass over the final tree came back clean.

Final state: tool-server gate (3 tests) + renderer pin green, argent-mcp suite 81/81, prettier/typecheck clean.

packages/argent-mcp/test/content.test.ts
```diff
@@ -255,6 +255,31 @@ describe("toMcpContent with artifact ctx", () => {
     expect((result[1] as { text: string }).text).toMatch(/^Saved: .*shot\.png$/);
   });
 
+  // The renderer's half of the one-observation guarantee; the allow-list half
+  // is tool-server's run-sequence-observation-gate.test.ts. The result is
+  // walked whole, steps included, and every artifact handle found becomes its
+  // own image block — so a sequence holds at one frame only while its steps
+  // report none.
```
latekvo, line 262: [[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: This check does not fail under the mutation the PR's verification table claims for it. I replaced the `gesture-tap` step's result with an image artifact handle (`{ [ARTIFACT_MARKER]: true, id: "mut", filename: "shot.png", mimeType: "image/png", size: 0 }`) and ran this test — it still passes.

The reason: the ctx passed here declares no `fetchImpl`, so `materializeArtifacts` attempts a real fetch of `http://remote:3001/artifacts/mut` during the walk over `steps[].result`. That fetch fails, the error is swallowed by the catch-and-return-null path in `packages/argent-tools-client/src/artifacts.ts` (~line 556), and no image block is produced — the handle just disappears from the rendered JSON. The walk does reach the step result (the fetch attempt fires), but any step-carried frame whose download fails renders as nothing, so this assertion cannot distinguish a silent sequence from one whose step frames all failed to materialize. As committed, the renderer half of the one-observation pin only bites for handles that actually resolve in this environment.

latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat), Ox Alpha]: Ran the sweep on 5a728e09: claims vs code (all three gate assertions, the render path mcp-server.ts:304 → content.ts:72-91 → artifacts.ts walk), non-happy paths, inputs, reachability both ways, lifetime, absence pass, plus mutation runs of the new tests locally (gate suite green; allow-list mutation correctly caught by the reviewed-set and outputHint checks). One finding below; everything else held up.

packages/argent-mcp/test/content.test.ts
latekvo: Confirmed, and fixed in f5cbf76fc.

Reproduced exactly as described: with the `gesture-tap` step's result replaced by `{ image: artifactHandle("mut", "shot.png", "image/png") }` the test still passed, and its runtime jumped from 4 ms to 289 ms - the real `fetch` to `http://remote:3001` going out and failing. The handle was swallowed and rendered as nothing, so the assertion could not tell a silent sequence from one whose step frames all failed to download.

The ctx now carries a `fetchImpl` that resolves to PNG bytes, and the test asserts it is never called. Both assertions now bite under your mutation, independently: with the fetch-call assertion removed the image-block assertion still fails (`expected [ { type: 'image', ... } ] to deeply equal []`).

Also added the positive control the pin rests on - a handle on `steps[].result` does surface as an image block. Mutating `materializeArtifacts`' array recursion to `return value` fails that test and nothing else in the file, which is what keeps the frameless assertion from going vacuous again if the walk ever stops reaching steps.

latekvo:
@coderabbitai how do i change your review profile from `CHILL` to whatever is the max so that you stop being by far the most useless integration we've tried in this repo so far?

latekvo:
@coderabbitai can you open a PR that'll do exactly just that?

PR #940, opened 2026-08-24
feat(flows): record a script step by running it

hubgan:
#864 (the executor) and #865 (the `script:` step) are merged. This branch now
sits directly on `main` as a single commit, so the diff and the commit list are
this change alone.

A `script:` step could only enter a flow by being typed into the YAML after the
recording ended. The authoring skill said so in as many words, and listed it
among the legal unrecorded insertions. That made it the one directive exempt
from every rule the recorder exists to enforce, and it cost two things:

- **The script never ran while the flow was authored.** It first executed during
  the closing replay, where a failure reads as a flow that was never right
  rather than a script that needs one more attempt. Worse, the device steps
  recorded after it were walked against backend state the author had established
  by hand — so the recording proved nothing about whether the script establishes
  that same state.
- **The author never saw what the script returned.** A later release reads
  `{{output:user.id}}`; the only way to know that path exists is to have seen the
  document.

This adds one MCP tool, `flow-add-script`, that runs a local `.mjs` and records
the step it ran. **No new flow syntax** — the `script:` step is unchanged, and
this is a second way for one to enter a flow file.

```text
flow-add-script { name: "checkout", project_root: "/Users/dev/shop",
                  path: "../../scripts/seed-order.mjs", timeout: 60000 }
```

## What it adds

- **One shared execution path.** Path resolution, the on-disk casing guard, the
  executor call and the verdict were private to `flow-run.ts`. They are now
  `runFlowScriptStep` in `flow-script-step.ts`, with the file-ref helpers in
  `flow-file-refs.ts`, and `runScriptStep` is that call plus its `ExecState`
  plumbing. This is the load-bearing change: "the recording runs the script
  exactly as replay will" is one function, not two code paths that agree today.
- **A failure records nothing.** The opposite of `flow-add-step`, deliberately.
  A failed script did not establish the state the rest of the walkthrough would
  then be recorded against, so an appended red step costs every step after it,
  not just itself. The `message` also says whether a child actually *ran* — a
  path that resolved to no file forked nothing, while a script that threw
  halfway did not roll back — because the author's next move is a retry or a
  cleanup and they cannot pick without knowing.
- **The returned document, as JSON text.** Text rather than an object because
  this is the first tool result carrying bytes the server did not author: every
  result is deep-walked for `__argentClientFile` and `__argentArtifact`
  directives, which match on shape alone and write or fetch files on the agent's
  machine. As a string it inherits none of that. Bounded at 64 KiB with
[...]

PR #865, opened 2026-08-21
feat(flows): run an external script as a flow step

hubgan:
Stacked on #864, which adds the executor. This branch adds the `script:` step
itself: the YAML directive, the runner integration, the CLI and MCP rendering,
and the authoring reference.

```yaml
- script: { path: ../../scripts/seed-order.mjs }
- script: { path: ../../scripts/seed-order.mjs, timeout: 60000 }
```

A `script:` step runs an external `.mjs` file from the project, in a fresh Node
process. What the script does is its own business — seeding an order, creating
a test account, cleaning up what a run left behind are the reasons flows have
wanted one. It drives no device, so a flow of nothing but scripts runs with
nothing booted.

## What it adds

- **The directive.** The value is always a map; `path` is required and obeys
  the same name rules a `run:` target does, spelled for `.mjs`. It resolves
  against the directory of the flow file containing the step, so a fragment
  reaches the same script whichever flow composed it. A mis-cased path is
  refused rather than run, because macOS and Windows would open it and Linux CI
  would not.
- **The report.** The script's stdout and stderr come back on the step and
  print under the step line, on a pass as well as a failure — it is the only
  record of what the step did to the backend. Both renderers (CLI, MCP) carry
  it, live and buffered.
- **The fail/error split.** A `fail` is the script's own answer; an `error` is
  the host's. That is what lets CI tell a regression from the machine it ran
  on.
- **Two boundaries.** An uploaded flow carrying a `script:` step is refused
  before anything runs — its `.mjs` never left the client. A flow whose script
  step sits beside a `run:` step still resolves a device.

## Review follow-ups

The last three commits answer a review of the three above them.

- A launch behind a leading `script:` was not read as the flow's leading
  launch, so a Chromium flow that seeded a backend first hoisted no boot, bound
  to whatever browser happened to be up, and passed a launch it never
  performed. The same lead-in walked a fragment past the
  `executionPrerequisite` refusal its `echo:`-led twin is given. Both readers
  now share one exhaustively-switched predicate.
- Coverage: the whole failure-kind table, the mis-cased arm CI never ran, the
  two unreached `scriptFileProblem` arms, symlink resolution for a script path,
  and the CLI live-progress closure that no test had executed.
- Docs: the reference claimed the log was secret-redacted (it is not), never
  mentioned the environment allowlist, the working directory, the time-limit
  ceiling or the log caps, stated a fail/error rule that was not the code's,
  and was missing from its own table of contents. The skills stated a closed
[...]

PR #954, opened 2026-08-25
optimization: simplify flow-run description

kasperski95:
# Why

It is reported that benchmark results are getting worse and the `flow-execute` description has the largest optimization potential. The description field of MCP tools should describe what the tool does and when to call it, not reveal implementation details.

Screenshot 2026-08-25 at 12 45 30

# Warning

Not tested. I'd have to run this change twice against the benchmark repo but that feels like overkill. I am not sure about behavioral regressions, relying on CI here.

## Summary by CodeRabbit

- **Documentation**
  - Updated the Flow YAML reference to streamline directive documentation.
  - Revised the flow execution tool’s description with a concise overview of supported flow sources and per-step reports.
  - Clarified documentation around flow behavior, including warning handling and project-root configuration.
  - Removed outdated or overly detailed statements from the published guidance to keep the reference accurate and easier to scan.

### DIFF
--- a/packages/skills/skills/argent-create-flow/references/flow-yaml.md
+++ b/packages/skills/skills/argent-create-flow/references/flow-yaml.md
@@ -109,7 +109,7 @@ Scopes can combine and nest, with at most six scope keys. Use strict selectors f
 
 ## Directives
 
-Directives stop the flow on failure and skip later steps. `flow-execute` documents their shapes. The available directives are `launch`, `tap`, `long-press`, `swipe`, `type`, `scroll-to`, `pinch`, `rotate`, `await`, `assert`, `wait`, `snapshot`, `run`, `script`, `when`, `echo`, and `tool`.
+Directives stop the flow on failure and skip later steps. The available directives are `launch`, `tap`, `long-press`, `swipe`, `type`, `scroll-to`, `pinch`, `rotate`, `await`, `assert`, `wait`, `snapshot`, `run`, `script`, `when`, `echo`, and `tool`.
 
 Use the launch map for cross-platform flows. A bare launch applies everywhere and becomes an app path on Chromium. The map takes `native:`, `ios:`, `android:`, `vega:`, and `chromium:`. `native:` is one id shared by iOS, Android, and Vega, and a per-platform key overrides it for that platform. `chromium:` accepts a relative or absolute app path. A launch that declares no id for the run's platform is an error, not a cue to switch platforms. On iOS, a successful launch also pins later tree reads to that app until the next raw `tool:` step, so read [The runner tree is not the discovery tree](#the-runner-tree-is-not-the-discovery-tree) when a read describes the wrong screen.
 
--- a/packages/tool-server/src/tools/flows/flow-run.ts
+++ b/packages/tool-server/src/tools/flows/flow-run.ts
@@ -1223,7 +1223,7 @@ export function createRunFlowTool(
       failedMsg: ({ params, failureSignal }) =>
         `Failed to run flow ${displayFlowName(params)}: ${failureSignal.error_code}`,
     },
-    description: `Run a saved flow — a YAML script of UI steps — end to end against a booted device. Use when
+    description: `Run a saved YAML flow end to end. Use when
 asked to replay a recorded path, re-run a QA regression, or check that a known journey still passes; for a
 one-off interaction use the gesture tools instead, and to author a flow use flow-start-recording. Pass
 exactly one flow source: name (under project_root) or flow_path.
--- a/packages/tool-server/src/tools/flows/flow-run.ts
+++ b/packages/tool-server/src/tools/flows/flow-run.ts
@@ -1226,9 +1226,8 @@ export function createRunFlowTool(
     description: `Run a saved flow — a YAML script of UI steps — end to end against a booted device. Use when
 asked to replay a recorded path, re-run a QA regression, or check that a known journey still passes; for a
 one-off interaction use the gesture tools instead, and to author a flow use flow-start-recording. Pass
-exactly one of name (under project_root) or flow_path.
-Returns a per-step report: the first failure stops the run and the rest report as skipped; a flow with an unacknowledged executionPrerequisite returns a
-notice instead of running; and a step that passes carrying a \`warning\` was sent, not verified.`,
+exactly one flow source: name (under project_root) or flow_path.
+Returns a per-step report: the first failure stops the run and the rest report as skipped.`,
     longRunning: true,
     zodSchema,
     fileInputs,
--- a/packages/tool-server/src/tools/flows/flow-run.ts
+++ b/packages/tool-server/src/tools/flows/flow-run.ts
@@ -1223,85 +1223,9 @@ export function createRunFlowTool(
       failedMsg: ({ params, failureSignal }) =>
         `Failed to run flow ${displayFlowName(params)}: ${failureSignal.error_code}`,
     },
-    description: `Run a saved flow from the .argent/flows/ directory, or an explicit boundary-managed flow_path.
-Use when a scenario is already authored as YAML and the whole of it should replay in one call with a
-per-step verdict; reach for the individual gesture tools when nothing is authored yet, and for
-run-sequence when the steps are an ad-hoc list rather than a stored flow.
-Steps run in order: \`launch\` starts an app from scratch (terminate + relaunch) and waits until it is
-ready (on iOS it also pins later element lookups to that app rather than auto-detecting the frontmost
-one); \`tool\` calls dispatch through the registry (a raw \`tool\` step ends that iOS pin, so lookups
-auto-detect again until the next \`launch\`, though a tool that cannot change the foreground app leaves the
-launched id as a fallback for a timed-out auto-detect, and \`launch-app\`/\`restart-app\` leave the id they
-started as that fallback instead); \`tap\`/\`long-press\`/\`type\` resolve a selector to an
-element and act on it (\`tap: { on, times: 2 }\` double-taps; \`long-press: { on, duration }\` presses and
-holds; \`tap\`/\`long-press\` alternatively take a raw normalized point — bare \`{ x, y }\` or \`on: { x, y }\`;
-any selector may scope its matches geometrically, the CSS combinators read off frames: \`within: <selector>\`
-(descendant — inside that container's frame), \`after: <selector>\` (CSS \`~\` — following it in reading
-order), \`next: <selector>\` (CSS \`+\` — the nearest such follower, which unlike CSS reaches past a
-non-matching neighbour rather than failing), plus \`any: true\` (CSS \`*\` — legal only WITH a scope and
-never beside text/id/role). Scopes nest to disambiguate — \`within: { id: card, within: { id: list } }\`
-reads "inside card inside list", each container's frame inside the next);
-\`swipe\` performs one finger flick (\`swipe: left\`, or \`swipe: { from?, direction|to|by, momentum?, duration? }\` —
-direction is the FINGER's travel, the opposite sense of scroll-to's content direction; \`by: { x?, y? }\` — signed
-0–1 screen fractions, combined length at least 0.03 (a diagonal clears it where neither axis does); duration in ms,
-default 300, minimum 150, maximum 10000; each bound is a parse error that rejects the file before any step runs;
-\`momentum: false\` lands exactly where the finger lifts instead of flinging);
-\`scroll-to\` scrolls (momentum-free) until a target is visible; \`pinch\` zooms
-(\`pinch: { on?, scale }\` — scale > 1 in, < 1 out; screen center when \`on\` is omitted); \`rotate\` is the
-two-finger rotation gesture (\`rotate: { on?, by }\` — degrees, + clockwise, within ±3000°; screen center
-when \`on\` is omitted; distinct from the \`rotate\` tool, which changes device orientation); \`await\` waits
-for a UI condition, and additionally takes the one condition that has no selector: \`idle: true\` waits
-until the screen has content and stops moving in BOTH the UI tree and the rendered pixels (it never
-fails a run — a screen that never settles passes carrying a \`warning\`, which is what makes it safe to
-persist; the one idle outcome that does stop the run is an \`error\` for a tree source THIS step could not
-read at all — a broken window rather than a verdict about the app, which leaves the run not-ok and skips
-every later step; it says nothing about WHICH screen settled — a dropped tap leaves the source screen
-perfectly idle — so pair it with the element check that names the destination); \`wait\` pauses for a fixed number of milliseconds; \`assert\` checks one now; \`snapshot\`
-diffs a screenshot — or, with \`cropOn: <selector>\`, one element's cropped region — against a stored
-baseline (a missing baseline fails the step — set updateBaselines to adopt the current screen; a
-cropped element whose size drifted fails on dimensions); \`echo\` annotates; \`run\` executes another flow
-inline — a YAML path resolved against the directory of the flow file that references it (co-located
-runs only); \`script\` runs a local .mjs file in a fresh Node process for setup, cleanup, or any work a device step cannot do
-(\`script: { path: ../../scripts/seed.mjs, timeout?: <ms> }\` — always a map, never a bare path; the path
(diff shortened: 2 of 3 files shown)


packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -1041,70 +1041,12 @@ export function createRunFlowTool(
...
-
-If a fragment has an execution prerequisite and prerequisiteAcknowledged is not set to true, the tool
-returns a notice with the prerequisite instead of running.
-Pass exactly one flow source: name for a saved flow under project_root, or flow_path for an explicit YAML — both together, or neither, fails the call.`,
+    description: `Run a saved flow — a YAML script of UI steps — end to end against a booted device. Use when
+asked to replay a recorded path, re-run a QA regression, or check that a known journey still passes; for a
+one-off interaction use the gesture tools instead, and to author a flow use flow-start-recording. Pass
+exactly one of name (under project_root) or flow_path.
+Returns a per-step report: the first failure stops the run and the rest report as skipped; a flow with an unacknowledged executionPrerequisite returns a
+notice instead of running; and a step that passes carrying a \`warning\` was sent, not verified.`,
```
latekvo:
```suggestion
Returns a per-step report: the first failure stops the run and the rest report as skipped; and a step that passes carrying a \`warning\` was sent, not verified.`,
```

This fact doesn't need to be mentioned in description in my opinion.
The `prerequisiteAcknowledged` argument is self-explanatory as long as the agent read the prerequisites. In worst case the agent will not provide `prerequisiteAcknowledged: true` as a parameter, in which case there will be a warning ensuring the agent is aware of said prerequisites.

Pre-informing the agent about this needing to be `true` somewhat reduces reliability of these prerequisites being read and enforced with little to no benefit (in my opinion).

latekvo: >and a step that passes carrying a \`warning\` was sent, not verified.

This sentence is very vague. It is not clear what a step being `sent` but not `verified` means in this context, because neither the concepts of step `verification` or `sending` are well defined. 

While i suspect `sent` means `executed`, i have no clue what `verified` could be referring to.

latekvo (changes requested):

packages/tool-server/src/tools/flows/flow-run.ts
kasperski95: > While i suspect sent means executed, i have no clue what verified could be referring to.

Removed that. An event/command was sent/dispatched to the device but the result of that event wasn't verified.

kasperski95: Removed. LLM added line that because tool-description-quality checks for has_error_guidance.

filip131311 (approved):

latekvo (changes requested):
`flow-yaml.md:106` says "flow-execute documents their shapes." which isn't true anymore due to these changes.

Please run cross-reference checks to make sure no similar issues occur after these changes. If they do, i think it'd be reasonable to remove them altogether rather than fill in the gaps, since the agent doesn't need to know the output shapes or similar technicalities beforehand.

That plus the conflicts pls

kasperski95:
> `flow-yaml.md:106` says "flow-execute documents their shapes." which isn't true anymore due to these changes.
> 
> Please run cross-reference checks to make sure no similar issues occur after these changes. If they do, i think it'd be reasonable to remove them altogether rather than fill in the gaps, since the agent doesn't need to know the output shapes or similar technicalities beforehand.
> 
> That plus the conflicts pls

I don't see a way to "run cross-reference checks" other than the one in the CI pipeline, unless you meant asking the AI to do so. If so, the AI didn't find any more issues.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -1197,78 +1197,11 @@ export function createRunFlowTool(
...
-off <id>\`, or \`retired <id> (same app relaunched)\` when the instance it left was the one killed —
-a relaunch that retired an older owned instance names both.
-
-If a fragment has an execution prerequisite and prerequisiteAcknowledged is not set to true, the tool
-returns a notice with the prerequisite instead of running.
-Pass exactly one flow source: name for a saved flow under project_root, or flow_path for an explicit YAML — both together, or neither, fails the call.`,
+    description: `Run a saved flow — a YAML script of UI steps — end to end against a booted device. Use when
+asked to replay a recorded path, re-run a QA regression, or check that a known journey still passes; for a
+one-off interaction use the gesture tools instead, and to author a flow use flow-start-recording. Pass
+exactly one flow source: name (under project_root) or flow_path.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This branch forks at `6e9e45c9f` (Sep 4), before `script:` flow steps landed (`9c4bc9ef5` #865, Sep 7 12:08). So the text being replaced is not the text on main, and two facts about `script:` steps live in it and nowhere else an agent can see:

- that the step needs no device — `The step needs no device, so a script-only flow runs with nothing booted — but a \`run\` or a \`when\` step beside it still resolves one`
- that a chromium launch is hoisted `following a leading \`run:\`, \`echo:\` or \`script:\``

I swept all 18 dirs under `packages/skills/skills/`, every `description:` and `.describe()` in `packages/tool-server/src/tools/**`, `packages/registry/**` and `packages/argent-mcp/**`, and `argent flow --help`. Each fact appears in exactly one agent-facing string: this one. `references/flow-yaml.md` §Local scripts covers the map form, `.mjs`, path-relative-to-the-flow-file and `timeout`, and says nothing about devices.

The second fact is the sharper one, because the surfaces that remain state the **narrower, now-wrong** rule rather than being merely silent:

- `references/flow-yaml.md:229` — `The leading launch — the flow's own, or the one its leading \`run:\` chain reaches — boots before step 1`. An exhaustive-sounding enumeration that omits `script:`.
- `packages/argent-cli/src/flow.ts:105` (`argent flow --help`) — `a fragment whose first step \`run:\`s a chromium e2e flow boots that flow's app before step 1`. Same omission.

Both survive otherwise only in `packages/docs/docs/reference/flow-yaml.mdx:277,320`, which is `private: true`, is excluded from the root `workspaces` list, is not among `packages/argent`'s published `files`, and is linked from no skill — so it never reaches an agent.

#865 also added `describe("create-flow script docs")` to `packages/tool-server/test/flows/flow-skill-docs.test.ts`, whose docblock states the same thing: `The \`script\` account in the flow-execute description is the only agent-facing statement of what the step needs and where its path points, so the two claims an author acts on are pinned here rather than left to prose review.`

Repro, on built trees. `origin/main` alone, `npx vitest run test/flows` in `packages/tool-server`: `Test Files 70 passed (70)`, `Tests 2063 passed (2063)`. Then main's `flow-run.ts` with only the description literal swapped for this branch's: `Tests 3 failed | 2060 passed (2063)` — the delta is exactly those three, no collateral, identical across repeated runs.

```
 FAIL  create-flow script docs > names every step that still resolves a device beside a deviceless script
   AssertionError: to match /a script-only flow runs with nothing booted[^.]*`run`/
 FAIL  create-flow script docs > shows a script path that reaches the directory scripts really live in
   AssertionError: to contain 'script: { path: ../../scripts/seed.mjs'
 FAIL  create-flow script docs > lets the chromium hoist follow every step a leading launch can sit behind
   AssertionError: to match /following a leading[^.]*`run:`/
```

Left exactly as git auto-merges it, CI stops one step earlier than that: the merged test file keeps main's two uses of `createRunFlowTool` but takes this PR's deletion of its import, and `unit-tests.yml` runs `Typecheck tests` before `Run tests` — `tsc --noEmit -p tsconfig.test.json` exits 2 with `test/flows/flow-skill-docs.test.ts(234,35): error TS2304: Cannot find name 'createRunFlowTool'`.

Worth flagging because git gives no signal here: trial-merging `origin/main` conflicts only in `flow-run.ts` and `flow-yaml.md`; the test file merges clean. The only signal is a red CI whose cheapest-looking fix is deleting the three tests — which this PR already does once, to the four `description` assertions in `create-flow idle docs`. And the eight green checks now on the PR were computed at `2026-09-07T09:40`, before #865 (12:08) and #940 (12:59), so they are not evidence about the merged state.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -1197,78 +1197,11 @@ export function createRunFlowTool(
...
+Returns a per-step report: the first failure stops the run and the rest report as skipped.`,
```
latekvo, line 1230: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `failure` is the only outcome word this line gives the reader, but the report it describes carries four step statuses and two counters the sentence never accounts for. A step that throws reports `error`, not `fail`, and `summarize` counts the two separately, so a run that stopped dead reports `failed: 0`.

Hit it on a booted simulator without trying to — an `await: { idle: true }` whose tree source could not be read:

```
{"flow":"e2e-settings","device":"770F654F-…","ok":false,
 "passed":1,"failed":0,"skipped":3,"errored":1,
 "steps":[{"kind":"launch","status":"pass"},
          {"kind":"idle","status":"error","reason":"could not read the UI tree while waiting …"},
          {"kind":"echo","status":"skip"},{"kind":"assert","status":"skip"},…]}
```

Both renderers print `errored` as its own counter (`argent-mcp/src/content.ts`, `argent-cli/src/flow.ts`), so an agent that learned only the word `failure` here reads `failed: 0` beside a stopped run.

`skipped` has the same gap in the other direction: an unmet `when:` guard skips its block and the run keeps going, which `flow-run.ts:1727-1731` states outright — `A skip by itself is NOT a failure: an unmet \`when:\` guard skips its block as a successful omission.` A regression flow whose checks all sit behind one false guard comes back `ok: true` with skips and no failure — and re-running a QA regression is a use case this description names. The deleted text covered that case (`a skipped block reports distinctly`); nothing else does now.

`aborted` is the third: it is set on cancellation, and after this it appears in no agent-facing surface at all (0 hits across `packages/skills` and `packages/docs`).

packages/skills/skills/argent-create-flow/references/flow-yaml.md
```diff
@@ -103,7 +103,7 @@ Scopes can combine and nest, with at most six scope keys. Use strict selectors f
 
 ## Directives
 
-Directives stop the flow on failure and skip later steps. `flow-execute` documents their shapes. The available directives are `launch`, `tap`, `long-press`, `swipe`, `type`, `scroll-to`, `pinch`, `rotate`, `await`, `assert`, `wait`, `snapshot`, `run`, `when`, `echo`, and `tool`.
+Directives stop the flow on failure and skip later steps. The available directives are `launch`, `tap`, `long-press`, `swipe`, `type`, `scroll-to`, `pinch`, `rotate`, `await`, `assert`, `wait`, `snapshot`, `run`, `when`, `echo`, and `tool`.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Dropping the pointer is right — it was false. It does leave this sentence naming directives whose shapes are then on no surface an agent receives, which is the cross-reference sweep asked for above.

Three of them, checked one at a time rather than as a class (`pinch`'s `scale` and `rotate`'s `by` are fine — `references/live-authoring.md:172-173` gives both conversions):

- `tap: { on, times: 2 }`, the only way to author a double-tap. `git grep -E "times: *[0-9]|double-tap" packages/skills/` → 0 hits.
- `long-press: { on, duration }`. → 0 hits for `long-press: {` in `packages/skills/`; this file's `duration` at line 142 is `swipe`'s.
- `any: true`, the one selector with no `id`/`text`/`role` of its own. → 0 hits in `packages/skills/`.

None appears in any other tool `description` or `.describe()` either. All three survive only in `packages/docs/docs/reference/flow-yaml.mdx:66,116-117`, which is `private: true`, is not among `packages/argent`'s published `files`, and is linked from no skill — so it does not reach an agent's context.

`long-press` is the sharpest, because `flow-add-step` actively routes the author here: `"long-press" is a flow directive, not a tool, and no tool records one — there is no gesture-long-press. Record the rest of the path, then add the \`long-press:\` step by hand during polish` (`flow-add-step.ts:776-778`). The author is sent to hand-write a step whose shape nothing agent-facing states, and `parseFlow` rejects the whole file before step 1 on a wrong guess. Those parse errors do name the allowed keys, so the cost is a round trip rather than a wrong result.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -1197,78 +1197,11 @@ export function createRunFlowTool(
...
-Returns a structured report ({ flow, device, executionPrerequisite, ok, aborted?, passed, failed,
-skipped, errored, steps }) — \`device\` is the device the run STARTED on; when launches moved it onto
-runner-booted instances, each names its instance in that step's reason and marks the move — \`run moved
-off <id>\`, or \`retired <id> (same app relaunched)\` when the instance it left was the one killed —
-a relaunch that retired an older owned instance names both.
-
-If a fragment has an execution prerequisite and prerequisiteAcknowledged is not set to true, the tool
-returns a notice with the prerequisite instead of running.
-Pass exactly one flow source: name for a saved flow under project_root, or flow_path for an explicit YAML — both together, or neither, fails the call.`,
+    description: `Run a saved flow — a YAML script of UI steps — end to end against a booted device. Use when
```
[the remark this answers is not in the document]
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Still present as of `b39f62c76`, and the runner really does run device-free flows.

Drove the built tool-server over `POST /tools/flow-execute` with an `echo` / `wait` / `echo` flow and `platform: android`, on a host with no Android device attached:

```
{"flow":"deviceless","device":"","executionPrerequisite":"","ok":true,
 "passed":1,"failed":0,"skipped":0,"errored":0,
 "steps":[{"kind":"echo","status":"pass"},{"kind":"wait","status":"pass"},{"kind":"echo","status":"pass"}]}
```

`resolveRunDevice` returns `{ device: null, booted: null }` before it lists anything when `flowRequiresDevice` is false (`flow-run.ts:1431`), and `test/flows/flow-deviceless.test.ts` exists for this case — 45 tests including `runs with nothing booted, and never looks for a device`.

Two more shapes the wording excludes. A leading `launch: { chromium: … }` has the runner **boot** the Electron app itself, so nothing needs to be up beforehand. And on current main `stepRequiresDevice` returns `false` for `script` too (`flow-device.ts:198-227`), with `flows/script/flow-script-step-run.test.ts` asserting `ok === true`, `device === ""` and that `list-devices` is never invoked for a script-only flow.

The same sentence's other half has the same problem: `a YAML script of UI steps`. `echo`, `wait` and `script` steps drive no UI at all — a `script:` step runs a local `.mjs` in a fresh Node process. That half was already loose before main moved (`echo`/`wait` were device-free and UI-free at this branch's fork point); #865 is what turns both halves from pedantry into a documented capability the sentence contradicts.

latekvo:
>I don't see a way to "run cross-reference checks" other than the one in the CI pipeline, unless you meant asking the AI to do so. If so, the AI didn't find any more issues.

Yes, asking AI to do so is what I meant. It's pretty good at spotting discrepancies between skills, docs and the code.
I see the automated review left some issues, i think we can merge after they're resolved.

packages/tool-server/src/tools/flows/flow-run.ts
kasperski95: Fixed by removing those (imo pointless AI generated) tests. CI didn't catch failures because this branch wasn't rebased to the latest main branch.

packages/tool-server/src/tools/flows/flow-run.ts
kasperski95: So, AI here complains that the description mentions that first failure stops the run, but "error" or "skip" status can result in similar behavior. Is this a problem?

packages/skills/skills/argent-create-flow/references/flow-yaml.md
kasperski95: What is the problem here in plain language?

latekvo (approved):

latekvo:
Thanks, merge when ready.
The AI automated reviews sweep over every open PR periodically, apologies for their unreadability - they aren't really meant to be read by humans. I'll make changes to prevent/limit this occuring in the future. Feel free to ignore both the comments left by Diplomat.

PR #958, opened 2026-08-25
feat(mcp): auto-describe — element tree after interaction tools + relaxed tapping rule

filip131311:
## Why

The rule's `<tapping_rule>` ("Before **every** tap, you MUST call a discovery tool") makes agents loop `tap → describe → tap → describe`: over 35 baseline runs, `describe` was called 90 times against 86 taps — one round-trip per tap, each a full model turn. The screen after an action is already captured by the auto-screenshot; the element tree is the only thing the agent still had to fetch.

## What

- **MCP** (`packages/argent-mcp`): after the action settles, call `describe` and append `--- Elements after action (describe) ---` + the tree to the tool result. Follows the same tool allow-list as the auto-screenshot, skips `describe` itself, and sits inside the non-secret branch so a `{{secret:…}}` value never re-enters context as text. Spy-logged as `auto_describe` (duration, chars). MCP `instructions` reworded.
- **Flag** `disable-auto-describe` (opt-out, mirrors `disable-auto-screenshot`). Auto-describe has its own tool list (`AUTO_DESCRIBE_TOOLS`) and its own branch, so the two features are independent; the secret-placeholder skip and the screen-idle wait cover both.
- **Rule + skill**: `tapping_rule` now says take coordinates from the returned tree, call a discovery tool only when there is no fresh tree for the screen (first action on a screen, after a wait, target not listed); the two-failed-taps rule and the "never from pixels" prohibition stay. `argent-device-interact` best-practice #6 and the general-rules line updated to match.
- **Docs**: `reference/configuration.mdx` flag row, `features/interacting-with-apps.mdx` discovery section.
- **Tests**: `autoDescribeEnabled` (global / project scope) and `shouldAutoDescribe`.

## Evidence

A/B with `claude -p --model sonnet`, one build (variant env-gated during the experiment), two sandbox projects differing only in rule/skill text and the toggle; 7 tasks × 2 arms × 5 reps = 70 runs on the iPhone 17 Pro simulator (iOS Settings ×3, Expo `my-app` ×2, Contacts create verified in sqlite) and the Electron testbed.

| | main | this PR | Δ |
|---|---|---|---|
| Task success | 35/35 | 35/35 | = |
| Turns / run | 10.4 | 8.2 | −21% |
| Wall / run | 36.5 s | 28.0 s | −23% (new won 28/34 pairs) |
| Cost / run | $0.28 | $0.23 | −17% (27/34) |
| `describe` calls / run | 2.2 | 0.24 | −89% |
| Cache-read tokens / run | 622k | 421k | −32% |

(34 clean pairs; one pair excluded for external interference on the simulator — with it included: wall −5%, cost −10%, turns −16%.) Every iOS task moved 15–30% in favour of the new arm; Electron was neutral (both arms describe once after `boot-device`, then a single `run-sequence`). The appended tree costs ≈2.4k chars (~700 tokens) and a median 73 ms per action; per-turn latency was unchanged, the saving is the removed model round-trip.

Full report: https://claude.ai/code/artifact/279528cf-ea92-4f52-9276-4f0475ba3e3c

## Caveats

- Sonnet only, short tasks, small screens. On feed-heavy apps with 10–20k-char trees the per-action overhead grows; a follow-up could cap or filter the appended tree.
- Flags are read once at MCP start (same as `disable-auto-screenshot`), so toggling needs an MCP restart.

## Checks

`vitest` argent-mcp (85) + configuration-core (122) green; `eslint --max-warnings 0` on changed packages; `prettier --check .`; `npx docusaurus build` green.

## Summary by CodeRabbit

- **New Features**
  - Interaction tools now return an accessibility element tree alongside the latest screenshot.
  - Agents can use fresh element data for follow-up actions without an additional discovery step.

- **Configuration**
  - Added the `disable-auto-describe` feature flag to turn off automatic element trees.

- **Documentation**
  - Updated interaction guidance and configuration references to explain the new behavior and flag.

- **Bug Fixes**
  - Secret-placeholder handling now correctly accounts for both screenshots and element trees.


packages/argent-mcp/src/auto-capture.ts
```diff
@@ -56,6 +57,42 @@ export function autoScreenshotEnabled(options?: FlagsPathOptions): boolean {
   return !isFlagEnabled("disable-auto-screenshot", options);
 }
 
+/**
+ * Tools whose result gets the accessibility element tree appended, so the
+ * agent has fresh tap frames without a separate `describe` round-trip.
+ * Independent of AUTO_SCREENSHOT_TOOLS; `describe` is absent because its own
+ * result already is the tree.
+ */
+export const AUTO_DESCRIBE_TOOLS = new Set([
```
latekvo, line 66: Why can't we just rename the auto screenshot tools to `AUTO_CAPTURE_TOOLS` and use this list for both `auto-screenshot` and `auto-describe`? Is there a valid reason?

latekvo (changes requested):

packages/skills/skills/argent-device-interact/SKILL.md
```diff
@@ -32,7 +32,7 @@ Use `list-devices` to get a target id. Results are tagged with `platform` (`ios`
 3. **Use `gesture-swipe` for lists/scrolling**, not `gesture-custom`, unless you need non-linear movement. On Chromium use `gesture-scroll` instead — `gesture-swipe` is touch-only. Consider whether you need multiple swipes, if yes - use `run-sequence`.
 4. **Tap a text field before typing**, then use `keyboard` to enter text.
 5. **Coordinates are normalized** — always 0.0–1.0, not pixels.
-6. **For app navigation, prefer `describe` first.** It works on any screen without app restart. Do not navigate from screenshots on regular in-app screens unless `describe` failed to expose a reliable target. Use `native-describe-screen` only when you need app-scoped UIKit properties.
+6. **For app navigation, use the element tree returned after each action** (`--- Elements after action (describe) ---`); call `describe` only when no fresh tree is available for the current screen. It works on any screen without app restart. Do not navigate from screenshot pixels on regular in-app screens unless the tree failed to expose a reliable target. Use `native-describe-screen` only when you need app-scoped UIKit properties.
```
latekvo, line 35: How would the agent ever know whether a fresh tree is available to the Argent internals or not? This is not the agent's decision to make.

latekvo, line 35:
```suggestion
6. **For app navigation, use the element tree returned after each action** (`--- Elements after action (describe) ---`); call `describe` only when no fresh tree is available for the current screen. It works on any screen without app restart. Do not guess interactive element positions unless; use Argent tooling for determining precise element positions. Use `native-describe-screen` only when you need app-scoped UIKit properties.
```

My suggestion. I think it's better.

latekvo (commented):
Good overall, 3 comments in total and i think other than that we should merge it asap.

packages/argent-mcp/src/auto-capture.ts
filip131311: we would need to extend AUTO_CAPTURE_TOOLS to AUTO_SCREENSHOT_TOOLS anyway because describe is an auto screenshot tool and i didn't want to play with inheritance

PR #959, opened 2026-08-25
fix(cli): accept --help after a telemetry subcommand

latekvo:
Found while differentially regression-testing the 0.22.1 release against a bundle built from `v0.22.0`.

## The problem

`argent telemetry status --help` exits 2 with `Error: Unknown flag: --help`. So do `-h`, and the same after `enable` / `disable`. On 0.22.0 all of them exited 0.

```
$ argent telemetry status --help
0.22.0   telemetry: / state: enabled / device id: not created     (rc=0)
0.22.1   Error: Unknown flag: --help                              (rc=2)
```

`telemetry` is the only subcommand that behaves this way. Checked against every sibling on both versions — `server status|logs|start`, `config list|get|set`, `flow list|run`, `secrets`, `flags`, `enable`, `link`, `unlink` all return 0 on both. And the top-level help tells the user to run exactly this: "Run `argent <command> --help` for command-specific help."

## Cause

`TELEMETRY_OPTIONS` declares only `scope`, and `parseCommandArgs(args.slice(1), …)` reads everything after the subcommand — so `--help` arrives as an unknown flag. The `case "--help"` arm in the switch only fires when help is `args[0]`, which is why bare `argent telemetry --help` still works.

Introduced when `telemetry` was converted to the declarative parser (#952, extended by #953).

## The fix

Hoist the help check above the parser, matching what `config.ts:220` and `flags.ts:92` already do; `server`, `link` and `lens` instead declare `help` in their option spec. Telemetry did neither. Ordering of `telemetryInit` / `telemetryShutdown` around the help path is unchanged, so a help invocation behaves exactly as `argent telemetry --help` did before.

## Tests

Six cases added to `telemetry-command.test.ts` covering `--help` / `-h`, bare and after each subcommand, asserting usage is printed and no config file is written.

Mutation-checked: with the source change reverted, 4 of the 6 fail (the 2 bare-help cases pass either way, since those already worked) — so the test pins the actual defect rather than the behaviour around it.

`@argent/cli`: 498 passed, 3 skipped. `tsc --build` clean, prettier clean.

## Docs

No docs update needed — this restores previously documented behaviour rather than changing a documented surface.

### DIFF
--- a/packages/argent-cli/src/telemetry.ts
+++ b/packages/argent-cli/src/telemetry.ts
@@ -20,19 +20,30 @@ const TELEMETRY_OPTIONS = {
 export async function telemetry(args: string[]): Promise<void> {
   const sub = args[0];
   let scope: FlagScope = "global";
-  try {
-    const { positionals, options } = parseCommandArgs(args.slice(1), TELEMETRY_OPTIONS);
-    if (positionals.length > 0) {
-      throw new UsageError(`Unexpected argument "${positionals[0]}".`);
+  // Help is accepted in any position, including after a subcommand; the option
+  // parser below knows only `--scope` and would reject it as an unknown flag.
+  const wantsHelp = args.includes("--help") || args.includes("-h");
+  if (!wantsHelp) {
+    try {
+      const { positionals, options } = parseCommandArgs(args.slice(1), TELEMETRY_OPTIONS);
+      if (positionals.length > 0) {
+        throw new UsageError(`Unexpected argument "${positionals[0]}".`);
+      }
+      if (options.scope !== undefined) scope = options.scope as FlagScope;
+    } catch (err) {
+      if (!(err instanceof UsageError)) throw err;
+      console.error(`Error: ${err.message}`);
+      printUsage();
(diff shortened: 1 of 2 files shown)


filip131311 (approved):
Verified locally on the PR head (f2e883a56):

- 12/12 tests pass; with `src/telemetry.ts` swapped back to main, the 4 after-subcommand cases fail and the 2 bare-help cases pass — matches the mutation claim in the description.
- Live on a build of this head: `telemetry --help`, `-h`, `status --help`, `enable -h`, `disable --scope project --help`, `--scope bogus --help` all exit 0 with usage; `status` still prints state; `enable --bogus` still exits 2; `nope` still exits 1.
- Confirmed independently that no other subcommand has this problem (25-case `--help` matrix across config/enable/disable/flags/server/link/unlink/flow/lens on main).

Hoisting the check above the parser is the right shape given `config`/`flags` already do it, and keeping init/shutdown around the help path preserves the old `telemetry --help` behaviour exactly. LGTM.

PR #864, opened 2026-08-21
feat(flows): run a local script in a bounded child process

hubgan:
Adds the executor a flow `script:` step runs on: a fresh Node child per step,
a protocol over its IPC channel, and the two watchdogs that make a hung or
runaway script the runner's problem rather than the host's.

The step itself is not here. This branch stops at the executor and its unit
tests; `feat/flow-script-step` stacks the YAML directive, the runner
integration, and the reference docs on top.

## What it does

- **One child per step.** Spawned with an old-space heap limit, an explicit
  working directory, and an environment built from an allowlist rather than
  copied from the tool server.
- **A deadline and a lifeline.** A separate watchdog holds the deadline, so a
  script that wedges the event loop still dies on time; a second one reaps the
  whole process tree when the parent goes away, so a grandchild cannot outlive
  the run.
- **A concurrency queue.** Slots are bounded per server. A step that never gets
  one is refused with a message that says so, and an aborted run frees its slot
  immediately.
- **Log budgets.** 64 KiB per step and 256 KiB per run, counted on the bytes
  the report keeps: redaction and V8 frame collapsing both run before anything
  is counted, so what the limits bound is the size of the report rather than
  the size of the script's writes.
- **Secret hold-back.** The scrub walks a chunk and stops where a value could
  still begin, so neither half of a value split across two chunks is released
  on its own, and a shorter value is never taken where the longer one
  containing it has not arrived yet.
- **A failure taxonomy.** Twelve kinds, split into what the script did (it
  threw, it did not load, it exited non-zero, it wrote an unusable `output`)
  and what the host did to it (a limit, a signal, a spawn that failed, a queue
  slot it never got).

## Docs

`packages/docs/docs/reference/configuration.mdx` lists the two configuration
keys this branch adds, `scripts.maxTimeoutMs` and `scripts.heapLimitMb`. The
`script:` step itself is documented with the step, on `feat/flow-script-step`.

## Verification

`npm run build`, `npx eslint . --max-warnings 0`, `npx prettier --check .`,
`npm run knip`, `npm run typecheck:scripts`, the test typecheck across every
workspace, the tool-server suite (4719 passed, 1 skipped) and `npm run
test:scripts` (92 tests) are green, as is `npx docusaurus build`. The executor's
own behaviour is covered by the eight `test/flows/script/` files added here,
against real child processes, and the compiled `dist/` executor was driven
against its copied runner assets over every path the review reached.

Each review fix carries the run that reproduced it before the change and the
[...]

PR #970, opened 2026-08-27
fix(flows): stop reporting what a script prints, on both sides

hubgan:
The last three PRs of this stack - #864 (the executor), #865 (the `script:` step)
and #940 (the recorder) - have landed, so this now sits directly on `main`:
one commit, and the diff is its own change alone.

A `script:` step used to bring its stdout and stderr back with it. The recorder
returned `log`/`logTruncated`, the step report carried
`scriptLog`/`scriptLogTruncated`, and the CLI and the MCP content builder each
rendered them.

Both of those surfaces land in an agent's context, and a script is a plausible
place for a connection string, a bearer token or customer data to be printed -
by the script itself, or by a CLI it shells out to. Nothing in the flow file
declares what a script prints, so nothing can decide in advance whether that
text is safe to forward.

An opt-in parameter was the other option, and it is not a control: the agent is
the one who would set it. So the channel goes rather than being gated - the
recorder returns none, a replay returns none, and there is no parameter to ask
for them.

## What a script step answers with now

Its verdict and its `reason`. `reason` is the script's own message for three of
the nine failure kinds - a thrown error, a bad output document, a load failure -
and the runner's account of how the process stopped for the other six (an exit
code, a signal, a time limit, the heap limit, a cancellation). On a PASS it
carries the executor's notes, the clamped time limit among them.

A script with something to say has to `throw` it or return it in `output`. Both
of those texts are still redacted against the flow's secrets; that half is
unchanged.

## What goes with the channel

The capture is deleted rather than hidden behind a flag, so everything that
existed to bound it has no subject left: the 64 KiB per-step cap, the 256 KiB
run-wide budget, the V8 frame collapser, the partial-marker guard, and the
streaming half of the secret scrub. `scrubSecretValues` absorbs the chunk walk
it was the only caller of and goes back to being a whole-text function.

## What the streams still owe

Three things, and only these:

- **stderr is still scanned for V8's heap-exhaustion banner.** It is the only
  account of why an abort was an out-of-memory rather than a plain signal.
  `StderrSignalWatch` matches over a sliding window and keeps no text beyond it.
- **stderr is still scanned for a watchdog that never armed.** That pipe is not
  only the script's: `flow-script-runner.mjs` and the lifeline thread each
  report their own failure to arm on it, and those lines are the only account of
  a step that ran with the group kill unarmed. A match sets a flag and earns a
  fixed sentence argent wrote, on `notes` beside the clamped-timeout and
  wrong-directory notes - never a byte the script chose, so a script forging the
  banner buys itself a false note rather than a channel. The report is only as
  available as the emission: `new Worker` rejects through an `error` event, so a
  script that returns immediately can exit before it fires.
- **Both pipes are still drained.** A paused pipe fills at 64 KiB and blocks the
  child from ever reaching its own time limit, which turns a passing step into a
  timeout. `child.stdout?.resume()` reads as a no-op and is not - see the A/B in
  Verification.

`StderrSignalWatch.end()` (then `HeapFatalWatch.end()`) is dropped along the way: it scanned what
`StringDecoder.end()` returns, which is either the empty string or a replacement
character for a trailing incomplete sequence, and the banner regex is ASCII-only.
It could never match anything the per-chunk scans had missed. The settle wait
above it is what actually puts the last chunks in before the streams are
destroyed, and that stays.

## One reworded reason

The exit-code failure said "no output was captured", meaning the output
document. With the console channel gone that reads as "the script printed
nothing", so it now names the document and points at `throw`.

## Tests

`flow-script-logs.test.ts` (660 lines) goes; its redaction and heap coverage
moves to `flow-script-redaction.test.ts` against the two texts that still carry
script-controlled bytes. Six tests whose names or comments claimed more than they
proved are fixed rather than deleted:

- the only guard on stdout draining was an integration test that failed by suite
  timeout, over a flood sized against this host's pipe buffer. The executor's own
  suite now has one, at 4 MiB;
- the straddle test did not straddle - the clamp landed 14 characters before the
  secret, so both assertions passed with `partialSecretTail` stubbed out;
- the nested-secret test did not exercise longest-first ordering;
- the stdout-close test described a write the fixture never made;
- the reworded exit-code reason had no assertion of its own;
- the recorder's chatty script flooded 112 KiB, below the threshold it claimed.

The renderers keep `scriptLog` on the wire in their fixtures on purpose: the
assertion is that a report carrying one renders nothing from it.

## Documentation

**No docs update is needed.** `packages/docs` and `packages/skills` are untouched
by this PR. An earlier revision added a paragraph to five of them saying Argent
discards what a script prints; a reader who never saw the log channel learns
nothing from being told it is absent, so all five went back to their base content.

What did have to change is the prose that still *promised* the removed fields,
and there the edit is a deletion rather than a replacement: `flow-add-script`
drops the `log`/`logTruncated` paragraph, `flow-run` drops "Its stdout and stderr
come back on the step report", and `argent flow --help` drops "each script step's
output" from what a directory run prints.

## Verification

`npm run build`, `npx eslint . --max-warnings 0`, `npx prettier --check .`,
`npm run typecheck:tests --workspaces`, `npm run knip`, `npm run test:scripts` and
the repo-hygiene checks are all green. (Only *Tool Description Quality* runs on
this PR — every other workflow is gated on `pull_request: branches: [main]` — so
these were run locally.)

- `npm test -w @argent/tool-server`: 378 files, 5025 passed, 1 skipped
- `npm test -w @argent/cli`: 27 files, 508 passed
- `npm test -w @argent/mcp`: 6 files, 82 passed

Exercised end to end against a tool server built from this branch, device-free
(a `script:` step needs none), through all three entry points. Fixtures print a
connection string, a token-shaped value and a plain canary to stdout and stderr:

`argent flow run` over a flows directory - a passing script contributes nothing,
and the three failures carry only their reason:

```
[1/6] a-chatty.yaml
  PASS — 1 passed, 0 failed, 0 errored, 0 skipped
[2/6] b-throws.yaml
  ✗  1 script ../../scripts/throws.mjs — seed API returned 500
  FAIL — 0 passed, 1 failed, 0 errored, 0 skipped
[3/6] c-oom.yaml
  ✗  1 script ../../scripts/oom.mjs — The script exceeded its 32 MiB heap limit.
  FAIL — 0 passed, 0 failed, 1 errored, 0 skipped
[4/6] d-flood.yaml
  PASS — 1 passed, 0 failed, 0 errored, 0 skipped
[5/6] e-exits.yaml
  ✗  1 script ../../scripts/exits.mjs — The script stopped its own process with exit code 3 instead of returning, so it left no output document behind: throw the reason instead of exiting on it.
  FAIL — 0 passed, 1 failed, 0 errored, 0 skipped
[6/6] f-clamp.yaml
  PASS — 1 passed, 0 failed, 0 errored, 0 skipped

FAIL — 6 flows: 3 passed, 3 failed, 0 skipped
```

A single run still prints a PASS note - `f-clamp.yaml` asks for a 166.7m limit:

```
Flow "f-clamp"
  ✓  1 script ../../scripts/chatty.mjs — The requested 166.7m time limit is above this host's maximum of 5m; the step ran with the maximum.
```

`flow run --json`, the raw report the CLI renders from, carries no `scriptLog`:

```json
{ "index": 0, "kind": "script", "flow": "a-chatty",
  "target": "../../scripts/chatty.mjs", "status": "pass" }
```

`flow-add-script` over HTTP, recording the same chatty script - no `log`, no
`logTruncated`, on a pass and on a failure:

```json
{"status":"pass","durationMs":41,
 "message":"Script step added to \"scriptrec\" flow — it ran here exactly as it will at replay. …",
 "outputJson":"{\"seeded\":true,\"orderId\":\"ord_4711\"}",
 "stepCount":1,"recorded":"1. script: ../../scripts/chatty.mjs"}
{"status":"fail","reason":"seed API returned 500","durationMs":119,
 "message":"The script \"../../scripts/throws.mjs\" failed — nothing was recorded in \"scriptrec\", …"}
```

The MCP stdio adapter, replaying that recording - no `script output:` block:

```
[1] ✓ script ../../scripts/chatty.mjs
[2] ✓ script ../../scripts/flood.mjs — The requested 166.7m time limit is above this host's maximum of 5m; the step ran with the maximum.
```

None of `hunter2`, `sk-live-…`, `DATABASE_URL` or any canary appears in the CLI
output, the `--json` report, the HTTP payload or the MCP content blocks.

**The drain, A/B against the built bundle.** `d-flood.yaml` writes 4 MiB to
stdout - past any host's pipe buffer - under a 5s limit. With
`child.stdout?.resume()` patched out of the compiled tool-server it fails at the
script's own limit, and passes again the moment the line is restored:

```
✗  1 script ../../scripts/flood.mjs — The script did not finish within its 5s time limit and its process tree was stopped.
✓  1 script ../../scripts/flood.mjs        (line restored)
```

The heap verdict in `c-oom.yaml` is the matching proof for the other stream:
stderr is still scanned even though nothing it carries is reported.

---

## Refreshed onto current main (2026-09-09)

Rebased from `0f90d182b` onto `02fd22e9d`. Still one commit, 21 files, +703 -1380.

One conflict, in `flow-run.ts`: main's #954 replaced the whole `flow-execute`
description with a short one, and this branch's only edit inside that block was
deleting the sentence "Its stdout and stderr come back on the step report."
Main's rewrite drops the sentence along with the paragraph that held it, so the
resolution takes main's text whole. `git diff <old-head> <new-head>` is exactly
main's own delta apart from that one line, so nothing else on the branch moved.

Re-verified after the rebase, not carried over:

- `npm run build`, `npm run lint`, `npx prettier --check .`, `npm run knip`, and
  `tsc --noEmit -p tsconfig.test.json` in tool-server / argent-cli / argent-mcp /
  argent - all clean.
- `npm test --workspaces`: 8467 passed, tool-server 433/433 files green. The two
  failures are pre-existing machine state (an installer test that hardcodes
  `opencode.json` where this host has `opencode.jsonc`, and a load-dependent
  SIGTERM dispatcher test that passes in isolation).
- The four-surface E2E re-run against a tool-server built from this commit: a
  script writing `DATABASE_URL`, an `AWS_SECRET_ACCESS_KEY` and three `CANARY_`
  markers to stdout and stderr, on a pass, a throw and a non-zero exit. None of
  those bytes appears in the CLI report, `--json`, the HTTP payload or the MCP
  content blocks, while the thrown message and its frames still land:
  `✗ 1 script ../../scripts/noisy-then-throw.mjs — seed API returned 500 at scripts/noisy-then-throw.mjs:3:7`.


latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Reviewed `10a0fb6e6` against its base `a7e236acf`, with both branches built and driven end to end rather than read.

What I ran:

- Both branches built and served: a tool-server from each on its own port, driven over HTTP (`flow-execute`, `flow-add-script`, `flow-start-recording` / `flow-finish-recording`), plus the real `argent flow run` binary across the single-flow, directory, `--json` and `--help` surfaces, and `flowRunToMcpContent` called on the server's own reports.
- The leak question directly: fixtures printing `CANARY_STDOUT_LINE_ONE`, `DATABASE_URL=postgres://user:hunter2@db/prod` and `CANARY_STDERR_LINE sk-live-…` on both pipes, on passing, failing, throwing, timed-out and cancelled steps. No script-printed byte reaches any of the four report surfaces on any status.
- Drain correctness: 4 MiB and 256 MiB stdout floods both pass (256 MiB in 203 ms); a heap-exhausting script still classifies off the stderr window as "The script exceeded its 512 MiB heap limit."
- Mutation testing on the changed lines. Removing `child.stdout?.resume()`, swapping the stderr `heapWatch` wiring for a bare `resume()`, dropping `this.tail = ""`, shrinking `HEAP_FATAL_WINDOW_CHARS` to 4, reverting the exit-code wording, dropping the notes in `scriptVerdict`, and re-adding each deleted renderer are all killed by the suites in this PR. Only `SETTLE_TIMEOUT_MS = 0` survives, which is a pre-existing timing race rather than anything this diff introduces.
- `scrubSecretValues` before vs after, differentially fuzzed over 200,014 cases (marker-shaped text, nested and prefix values, lone surrogates, astral characters, empty and duplicate values): 0 divergences.
- What outlives the call: fd count flat at 22 across 150 executions, and peak RSS 162 MB here against 207 MB on base.
- `prettier --check .`, `npm run knip`, `npm run typecheck:tests --workspaces`, and the tool-server script suites (358 tests), argent-cli (115) and argent-mcp (36) are green.

The deleted `flow-script-logs.test.ts` checks out: everything dropped is about the removed channel, and the non-log pins moved to `flow-script-redaction.test.ts`. I also could not find a docs or skills page that mentions script stdout/stderr, so the "no docs update" call matches the tree.

Three notes below. None of them is about the leak this PR closes.

packages/argent-cli/src/flow.ts
```diff
@@ -302,24 +287,16 @@ export function renderArtifactLines(report: FlowReport): string[] {
...
   const lines: string[] = [];
   let n = 0;
   for (const s of report.steps) {
     if (s.kind === "echo") continue;
     n++;
-    const scriptLog = renderScriptLogLines(s, n);
-    if (s.status !== "fail" && s.status !== "error" && !s.warning && scriptLog.length === 0) {
-      continue;
-    }
+    if (s.status !== "fail" && s.status !== "error" && !s.warning) continue;
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: With the log-based admission gone, this gate drops every PASSING `script` step from a directory run — and it can never be rescued by the `!s.warning` arm, because `runScriptStep` returns `Pick<StepReport, "status" | "reason">` and `warning` is only ever set on a directive step (`flow-run.ts:2286`). So after this change a passing script step is structurally unprintable on that surface, whatever it has to say.

`script` is the one kind that carries information in `reason` on a pass. `scriptVerdict` puts the executor's notes there on every outcome — "pass included" — and its own comment states the stake: *"They are how the executor says a time limit was clamped to the host's maximum, or that the working directory it was given did not exist — and dropping them on a pass is how a script that silently ran somewhere else stays silent."* `flow-script-verdict.test.ts` pins that with a test named "an executor note on the step report › rides into the reason of a step that PASSED", asserting `reason: "timeout clamped to 300000ms. project_root did not exist."`.

Reproduced with the real `argent flow` command against a tool-server built from each branch, over a one-flow directory holding `script: { path: ../scripts/clamped.mjs, timeout: 10000000 }` (host max 5m):

`a7e236acf` (base):

```
[1/1] g-clamp.yaml
  ✓  1 script ../scripts/clamped.mjs — The requested 166.7m time limit is above this host's maximum of 5m; the step ran with the maximum.
       │ CANARY_CLAMPED_PRINTED
  PASS — 1 passed, 0 failed, 0 errored, 0 skipped
```

`10a0fb6e6` (this PR):

```
[1/1] g-clamp.yaml
  PASS — 1 passed, 0 failed, 0 errored, 0 skipped
```

The same report still carries the note everywhere else — `--json`, a single-flow run (`✓  1 script ../../scripts/clamped.mjs — The requested 166.7m …`), and the MCP content block (`[1] ✓ script ../../scripts/clamped.mjs — The requested 166.7m …`, checked by calling `flowRunToMcpContent` on the server's actual report). Only `argent flow run <dir>` loses it, which is the surface the deleted test called "the surface CI reads" and the one the docs point at for CI. The `--help` text in this same diff was updated to match ("each failure and each warning"), so the narrowing is deliberate at the level of the sentence — but what it narrows away is not log text, it is the executor's verdict-adjacent notes. The `resolveWorkingDirectory` note — `project_root <path> does not exist; the script ran in <other path> instead.` — passes and disappears by the same route, so a seed script that ran against the wrong tree reports a bare green in CI.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
```diff
@@ -592,8 +570,13 @@ export class FlowScriptExecutor {
...
 
-    child.stdout?.on("data", (chunk: Buffer) => capture.push("stdout", chunk));
-    child.stderr?.on("data", (chunk: Buffer) => capture.push("stderr", chunk));
+    // Drained, never kept: what a script prints is nobody's to read here, but a
+    // pipe left paused fills its buffer and blocks the child from ever reaching
+    // its own time limit. `resume()` discards stdout as it arrives; stderr goes
+    // through the one reader that survives, which holds nothing beyond the
+    // window it matches against and reports none of it.
+    child.stdout?.resume();
+    child.stderr?.on("data", (chunk: Buffer) => heapWatch.push(chunk));
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The prose sweep this description accounts for ("the prose that still *promised* the removed fields") reached the three files it edited, but two comments in the same subsystem still describe the channel this commit deletes, and neither file is in the diff (`git diff a7e236acf...HEAD --stat` lists 20 files; neither is among them):

- `packages/tool-server/src/tools/flows/script/flow-script-protocol.ts:6-9` — *"Script logs never travel here — they ride stdout/stderr, so that console text and any subprocess the script starts land in one stream in written order, and so that a limit can apply while draining rather than after a whole message has been serialized."* There is no "one stream" any more: `ScriptLogCapture`, the class that merged the two pipes into one buffer in arrival order, goes in this commit, and so do both limits (`SCRIPT_STEP_LOG_LIMIT_BYTES`, `SCRIPT_RUN_LOG_LIMIT_BYTES`). `git blame` dates that paragraph to `a5ee1b4021` on the base branch, so the one file that documents the executor↔runner contract now explains its shape by a mechanism that no longer exists — and reads as an instruction not to move logging onto IPC, on grounds that are gone.

- `packages/tool-server/src/tools/flows/script/flow-script-runner.mjs:766-768` — `flushStream`'s *"writing to an ended stream raises an unhandled `error` event that would land in the step's own log."* There is no step log for it to land in. The test that pinned this guard was reworked here (`does not put a stream crash into the log of a passing step` → `passes a script that ends its own stdout mid-run`), so the assertion moved off the log while the comment beside the code did not. `finish`'s doc at `:639-641` names the same vanished thing — "a bare `process.exit` would discard buffered log output" is now the only stated reason for waiting on both stream flushes before the child exits.

Checked by reading each against the code at `10a0fb6e6`; `/usr/bin/grep -rn "stdout\|stderr" packages/tool-server/src/tools/flows` returns `flow-script-protocol.ts:6` as the only surviving prose in the flows tree that says logs travel on the pipes.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
```diff
@@ -592,8 +570,13 @@ export class FlowScriptExecutor {
       });
     };
 
-    child.stdout?.on("data", (chunk: Buffer) => capture.push("stdout", chunk));
-    child.stderr?.on("data", (chunk: Buffer) => capture.push("stderr", chunk));
+    // Drained, never kept: what a script prints is nobody's to read here, but a
+    // pipe left paused fills its buffer and blocks the child from ever reaching
+    // its own time limit. `resume()` discards stdout as it arrives; stderr goes
+    // through the one reader that survives, which holds nothing beyond the
+    // window it matches against and reports none of it.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: "what a script prints is nobody's to read here" holds for stdout, but it over-scopes stderr: that pipe is not only the script's. It is also the only channel argent's own child-side code has, and those lines now go into the discard with everything else.

Two first-party writers use it, both reporting that a watchdog — the second line of defence for a parent that is gone or stalled — never armed:

- `flow-script-runner.mjs:370-374` — `[argent] script watchdog <name> unavailable: <err>`, when either watchdog `Worker` fails to construct or errors.
- `flow-script-watchdog-lifeline.mjs:61` — `[argent] script lifeline unavailable: <err>`, whose own comment spells out the state it announces: *"the run finished with the group kill never armed, and descendants outlived a tool server that died afterwards."*

After this change nothing carries either line: `HeapFatalWatch` regex-tests each chunk for V8's banner and keeps nothing, and there is no `notes` entry, no server log and no telemetry standing in. The result is byte-identical to a healthy run, so a step that ran with neither watchdog armed reports a clean pass with the warning existing nowhere.

A/B against the built executor, same script, `runnerDir` pointing at a tree holding only `flow-script-runner.mjs` so both `Worker`s fail with MODULE_NOT_FOUND:

`a7e236acf`:

```json
{"ok":true,"output":{"ok":true},
 "log":"[argent] script watchdog flow-script-watchdog-lifeline.mjs unavailable: Cannot find module '…/flow-script-watchdog-lifeline.mjs'\n[argent] script watchdog flow-script-watchdog-deadline.mjs unavailable: Cannot find module '…/flow-script-watchdog-deadline.mjs'\n",
 "logTruncated":false,"durationMs":835,"queuedMs":0,"notes":[]}
```

`10a0fb6e6`:

```json
{"ok":true,"output":{"ok":true},"durationMs":844,"queuedMs":0,"notes":[]}
```

Being precise about how far that repro reaches: `runnerDir` is a test-only seam (`/usr/bin/grep -rn runnerDir packages/*/src` finds only the field, its one read, and `resolveRunnerPath`), and a mis-packaged publish is separately gated by `copy-build-assets.mjs`, `bundle-tools.cjs`'s `required: true`, and the "published layout" test — so I am not claiming a common path to a missing asset. What the repro does establish is the wiring: any `Worker` construction failure at runtime lands on the same `reportWatchdogProblem`, and there is no longer anywhere for it to surface. I confirmed the lines really are emitted by forking the real runner with this file's exact stdio (`['ignore','pipe','pipe','ignore','pipe','ipc']`) and reading the child's stderr directly. The description's account of what the streams still owe — "Two things, and only these" — does not cover this third one.

hubgan: Fixed in cd44bbe0.

Confirmed: `child.stderr?.on("data", …)` was stderr's only consumer, and the watch it fed keeps a boolean and a sliding tail, so both `[argent]` lines were irrecoverable.

stderr is now scanned for them the same way it is scanned for the heap banner - a flag, never text. `HeapFatalWatch` became `StderrSignalWatch` and tracks both patterns; its window is dropped only once neither is still looking, since one shared `found` would have ended the scan at whichever banner arrived first. A match earns a fixed sentence argent wrote, on `notes` beside the clamped-timeout and wrong-directory notes.

The flag is forgeable, as the heap flag already was. What forging it does not buy is a way to put the script's own bytes back on the report - verified against a script printing the banner alongside a connection string and a canary, neither of which reaches the report.

Verified end to end against a tool server built from this branch, driving `flow-execute` over HTTP with both watchdog assets removed so the real `Worker`s fail:

    watchdogs present: {"kind":"script","status":"pass"}
    watchdogs missing: {"kind":"script","status":"pass",
      "reason":"A watchdog for this step did not arm, so a process the script
                started can outlive a tool server that dies afterwards."}

One limit worth stating rather than hiding: `new Worker` rejects through an `error` event, so a script that returns immediately can exit before it fires - measured absent at 0ms and present from ~200ms. Making the emission deterministic would mean awaiting both worker spawns ahead of every script, a cost on the hot path this change does not take.

Three tests cover it, in the `flow script watchdogs` block: the real path through a runner directory holding only the runner, the forgery case, and one pinning that each banner is still looked for after the other has matched. The PR description's "Two things, and only these" now says three.

packages/argent-cli/src/flow.ts
hubgan: Fixed in db2d31c6.

Confirmed all three, and a sweep of the flows tree for surviving log prose returns only these (the other hits are a recording log, a `"Clear logs"` example selector and `Math.log`).

`flow-script-protocol.ts` - rewritten. You are right that the grounds were gone, and that the paragraph read as a standing instruction not to move logging onto IPC. The reason that actually holds is not about transport at all, so the comment no longer argues one: the executor drains stdout and stderr so console text never reaches the report, because nothing in a flow file declares what a script prints is safe to forward. What a script has to say still travels here, via `throw` or `output`, and the limits bounding both are in that same file.

`finish` - the wait survives, its stated reason does not. There is no buffered log output to discard, but stderr is still scanned - for V8's banner, and now for a watchdog that never armed (cd44bbe0) - so a write dropped at exit costs a signal rather than console text. Also worth stating and now stated: the verdict is not at stake either way, since `sendToParent` puts it on IPC before the flush starts.

`flushStream` - an unhandled `error` event is an uncaught exception, not something that lands anywhere. The comment now names what the guard actually buys, and points at the test you noted had moved off the log: a script closing its own stdout still passes.

Comment-only, so the evidence is that nothing keyed off the text: build, prettier, eslint and knip clean, the 12 `flows/script` suites 372/372, and the full tool-server run 5298 passed with only the two failures this branch already had before any of my commits (`flow-add-step`'s own description assertion, and a Lens platform-gate 5s timeout that flakes in and out run to run).

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
hubgan: Fixed in db2d31c6.

Confirmed all three, and a sweep of the flows tree for surviving log prose returns only these (the other hits are a recording log, a `"Clear logs"` example selector and `Math.log`).

`flow-script-protocol.ts` - rewritten. You are right that the grounds were gone, and that the paragraph read as a standing instruction not to move logging onto IPC. The reason that actually holds is not about transport at all, so the comment no longer argues one: the executor drains stdout and stderr so console text never reaches the report, because nothing in a flow file declares what a script prints is safe to forward. What a script has to say still travels here, via `throw` or `output`, and the limits bounding both are in that same file.

`finish` - the wait survives, its stated reason does not. There is no buffered log output to discard, but stderr is still scanned - for V8's banner, and now for a watchdog that never armed (cd44bbe0) - so a write dropped at exit costs a signal rather than console text. Also worth stating and now stated: the verdict is not at stake either way, since `sendToParent` puts it on IPC before the flush starts.

`flushStream` - an unhandled `error` event is an uncaught exception, not something that lands anywhere. The comment now names what the guard actually buys, and points at the test you noted had moved off the log: a script closing its own stdout still passes.

Comment-only, so the evidence is that nothing keyed off the text: build, prettier, eslint and knip clean, the 12 `flows/script` suites 372/372, and the full tool-server run 5298 passed with only the two failures this branch already had before any of my commits (`flow-add-step`'s own description assertion, and a Lens platform-gate 5s timeout that flakes in and out run to run).

packages/argent-cli/src/flow.ts
hubgan: Correction: my previous reply on this thread was meant for the stale-comments thread on `flow-script-executor.ts` and has been posted there. Please disregard it here.

For this finding: fixed in b546c637, before you raised it - the thread is marked outdated for that reason.

`renderFailedSteps` now admits a passing `script` step that carries a reason:

    const scriptNote = s.kind === "script" && Boolean(s.reason);
    if (s.status !== "fail" && s.status !== "error" && !s.warning && !scriptNote) continue;

restricted to `script` for the reason you identified - a passing `when` guard, snapshot or chromium launch also sets `reason` and narrates a result the summary already counts. `flow-script-render.test.ts` pins it with "carries a passing script's executor note into batch mode", over a clamped time limit, asserting both the step line and the note text.

Your reading of the mechanism was right on both counts, including that `!s.warning` could never rescue it because `runScriptStep` returns only `status` and `reason`. Worth noting the surface now carries one more note than when you filed this: cd44bbe0 adds a watchdog-never-armed note on the same `notes` -> `reason` path, so a seed script that lost its group kill also reaches a directory run through the gate you got restored.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: The log reference is gone here, confirmed. One thing on the replacement wording, since it now points at a test: `flushStream`'s comment (runner.mjs:773-775) says the early return and the one-shot `error` listener "are what keep a script that closed its own stdout on the passing path, which `passes a script that ends its own stdout mid-run` pins."

That test does not pin the guard. The passing path is held one layer up, by the verdict reaching IPC before the flush: `finish` calls `sendToParent(bounded, flush)`, so the terminal message is already on the channel before `flushStream` ever runs, and the parent classifies on that message rather than on the child's exit. I stripped both the guard and the listener from `flushStream` (leaving just `stream.write("", done)`), rebuilt the tool-server at db2d31c6, and re-ran `passes a script that ends its own stdout mid-run`: still green. So the named test neither depends on the guard nor would fail if it were removed — the write-after-end at most raises an unhandled `error` after the verdict is already delivered.

The guard is still worth keeping as exit-time hygiene against that unhandled `error` event — but its stated causal role in the passing path, and the test attribution, both claim more than the code and that test actually establish.

latekvo (commented):
Hmm i really don't like this. This completely blackholes `stdout` and `stderr`, well the latter is technically compressed to booleans but that's "dropping it" for all intents and purposes.

I understand we've talked about this and the Maestro compliant replacement is there in form of the return documents that the scripts are supposed to return, but that approach really seems to require too much from the user - basically catching all errors globally and writing them to a file right before exiting, it's just so weird.

I think it'd make sense to at least store the stdout and stderr in some file, since a regular CI nearly always **does** return failures, stderr et cetera for debugging purposes, there isn't a good reason not to do that.

I don't know man, after re-reading and re-considering everything here, i really dislike this approach but i could be missing something.

PR #972, opened 2026-08-27
fix: skip writing to allowlist when permissions.json is empty

magdalipka:
Previously when Argent was installed or updated it would write to Cursor's `permissions.json` file to add argent to mcp allowlist.
The issue is that when this file is used to determine [run mode](https://cursor.com/docs/agent/security/run-modes). When it is empty the cursor allows the user to use 'Run Everything (unsandboxed), but when it is not empty the cursor will keep asking for permission every time it does something.
So when Argent adds itself to empty file it effectively removed permission for other mcps and commands.

This PR updates the Argent to not write to the `permissions.json` file when it is empty.

### DIFF
--- a/packages/argent-installer/src/init-allowlist.ts
+++ b/packages/argent-installer/src/init-allowlist.ts
@@ -54,8 +54,16 @@ export async function configureAllowlist(args: {
       continue;
     }
     try {
-      adapter.addAllowlist!(effectiveRoot, scope);
-      lines.push(`${pc.green("+")} ${adapter.name}`);
+      // `false` means the adapter skipped on purpose: writing the rule would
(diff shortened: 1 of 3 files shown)


magdalipka:
#787

filip131311:
Thanks for the fix and the clear diagnosis, @magdalipka! We took it forward in https://github.com/software-mansion/argent/pull/986 — your commit is cherry-picked as-is, with the guard tightened from "file is empty" to "`mcpAllowlist` key exists" (a permissions.json holding only `terminalAllowlist` also constrains the run mode when the key gets added), plus a recorded `allowlist.enabled` config so `argent update` only re-applies allowlists for users who opted in.

PR #973, opened 2026-08-27
optimization: clean flow related skills

hubgan:
## Flow skills: load by phase and platform, cut the bloat

`argent-create-flow` + `argent-qa-flows` were 10,675 words across 5 files, and `SKILL.md` told the agent to read almost all of it on every recording (~8.9k words per session, ~10.7k for QA).

### What changed

- New layout: `SKILL.md` (router + 6 rules) -> one platform file (`references/platforms/{ios,android,chromium}.md`) -> phase files `record.md`, `polish.md` + `yaml.md`, `replay.md`, and `warnings.md` as the only home of every warning list. `flow-yaml.md`, `live-authoring.md`, `reliability-and-recovery.md` are removed.
- The skills say what to do; the tool descriptions already say how the runtime works.
- Flows are documented for iOS, Android, and Chromium only. Vega, Apple TV, and Android TV are declared unsupported in both skills, `rules/argent.md`, and `argent-screen-recording`.
- `long-press:` is now a permitted hand-added step, matching what `flow-add-step` already tells the agent.

### Size

| | Before | After |
| --- | --- | --- |
| Corpus | 10,675 words / 73 KB | 8,084 words / 50 KB |
| Record + polish + replay, iOS | ~8,900 | ~5,500 |
| QA flow, Android | ~10,700 | ~6,500 |
| Replay-only | 2,600-5,800 | ~1,400 |

## Summary by CodeRabbit

* **Documentation**
  * Updated flow guidance to clarify support for iOS, Android, and Chromium.
  * Clarified that saved-flow recording is not yet supported for Vega, Apple TV, or Android TV, while hand-written Vega flows can still be replayed.
  * Added platform-specific recording and replay guidance, plus expanded references for YAML authoring, polishing, warnings, recovery, and troubleshooting.
  * Revised related skill documentation and links to reflect the updated workflow.


packages/skills/skills/argent-create-flow/references/platforms/chromium.md
```diff
@@ -0,0 +1,43 @@
...
+
+```yaml
+steps:
+  - launch:
+      chromium:
+        path: ../../app
+        args: ["--enable-feature-flag"]
+```
+
+The path is relative to `.argent/flows/`. Write the boot arguments accurately. If the boot had no arguments, do not add `args`. If the file has an `executionPrerequisite`, remove it. This is the only launch that you can add by hand.
```
j-piasecki: The path is relative to the flow file location. This is true when an agent is recording, but maybe it should be clarified?

packages/skills/skills/argent-create-flow/references/yaml.md
```diff
@@ -0,0 +1,130 @@
...
+| `snapshot`   | `<name>` or `{ name, maxMismatch, cropOn }`                        | See [Snapshots](#snapshots)                               |
+| `run`        | `<path>`                                                           | See [Composition](#composition)                           |
+| `when`       | `{ <condition> }` or `{ platform }`, with `steps:`                 | See [Optional steps](#optional-steps)                     |
+| `echo`       | `<message>`                                                        | Printed in the report                                     |
+| `tool`       | `<tool name>`, with `args:` and optional `delayMs:`                | One of the Argent tools                                   |
+
+`tap`, `type`, and `long-press` do not scroll. When the target can be off-screen, put `scroll-to` before them. A `type` step can contain `{{secret:NAME}}`. The runner reads it from the environment variable `ARGENT_SECRET_NAME`, from `.argent/secrets.env` or `~/.argent/secrets.env` as `NAME`, or from `.env.local` or `.env` as `ARGENT_SECRET_NAME`. The report redacts the value, so do not use a secret for text that the report must show.
```
j-piasecki:
```suggestion
A `type` step can contain `{{secret:NAME}}`. The runner reads it from the environment variable `ARGENT_SECRET_NAME`, from `.argent/secrets.env` or `~/.argent/secrets.env` as `NAME`, or from `.env.local` or `.env` as `ARGENT_SECRET_NAME`. The report redacts the value, so do not use a secret for text that the report must show.
```

I _think_ this should be safe to remove

packages/skills/skills/argent-create-flow/references/yaml.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Launch map
+
+```yaml
+- launch: { native: com.acme.app, chromium: ../../app }
+- launch: { ios: com.acme.app, android: com.acme.app.android, chromium: ../../app }
+```
+
+`native` is one id for iOS and Android. A per-platform key replaces it. `chromium` takes an app path or `{ path, args }`. A launch with no id for the platform of the run is an error.
```
j-piasecki, line 73: `native` should also count Vega, but given that TVs are not supported by the recorder, it may be ok to omit it here.

packages/skills/skills/argent-create-flow/references/yaml.md
```diff
@@ -0,0 +1,130 @@
...
+Use `await` for a result that comes after an interval. Increase its timeout only after the default expires. Use `assert` for a settled state. A `hidden` check also passes before the element shows, for a typo, and on the incorrect screen. Record it in three steps ([Record: Absence](record.md#absence)).
+
+### Navigation checks
+
+```yaml
+- await: { visible: { id: profile-screen } } # identity
+- await: { idle: true } # readiness
+```
+
+The identity selector must be only on the destination screen. `idle` waits until the screen has content and does not move in the tree or in the pixels. The two checks do not replace each other.
```
j-piasecki: > `idle` waits until the screen has content

How does it determine whether the screen has content vs not?

packages/skills/skills/argent-create-flow/references/yaml.md
```diff
@@ -0,0 +1,130 @@
...
+
+```yaml
+- await: { idle: true, stableFor: 400, timeout: 9000 }
+```
+
+`stableFor` (default 250 ms) is how long the screen must not move. `timeout` (default 7500 ms) is the maximum time of the full wait. The smallest `timeout` is 600 ms. When `stableFor` is more than 400 ms, the smallest `timeout` is `stableFor` + 200 ms.
+
+`idle` does not fail a run. When the screen does not settle, the step passes with a warning. Read [Idle warnings](warnings.md#idle-warnings).
+
+`idle` has no `assert` form and no `when` form. Add it after each screen change, not after each step.
```
j-piasecki, line 106: Hmm, do you think it would be worth it to add "assert: { idle }" that would behave the same as await, but fail instead of warn?

packages/skills/skills/argent-create-flow/references/yaml.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Optional steps
+
+```yaml
+- when: { visible: { text: Got it } }
+  steps:
+    - tap: { text: Got it }
+```
+
+The guard is one `exists`, `visible`, `hidden`, or `text` condition, or `{ platform: ios | android | chromium }`. It uses the assert grace and rejects `timeout`. There is no `else`. Put separate behavior paths in separate flows. Use `when:` only for optional setup that goes back to the necessary path. Do not put a necessary check in `when:`.
```
j-piasecki, line 116: It can also be vega, but that comes down to the same argument as before.

packages/skills/skills/argent-create-flow/references/record.md
```diff
@@ -0,0 +1,83 @@
...
+
+1. Find the element without a change of state: `describe`, the platform discovery tool, `debugger-component-tree`, or `screenshot`. Do not record discovery calls or `debugger-*` calls.
+2. Select a stable target (`argent-create-flow` SKILL.md, rule 3).
+3. Add an echo with the current state, the action, and the result that must show.
+4. Do the action through `flow-add-step`. Read the `recorded` line.
+5. When its state shows, record the result check.
+
+### Navigation
+
+After each screen change, record `await-ui-element` with `visible` on an element that is only on the destination screen. A tab bar, a shared header, or a positional id does not identify the screen. If a stable control (not a data value) shows that the data is loaded, record a second wait on it. In the polish pass, add `await: { idle: true }` after the identity check. Go to a fixed destination when possible. A back button or a back swipe pops one stack entry, so its destination can change between visits. If the test is about back navigation, use back. Then check its result like each other screen change.
```
j-piasecki: > If the test is about back navigation, use back.

I don't think flow has `back` built in, does it?

packages/skills/skills/argent-create-flow/references/warnings.md
```diff
@@ -0,0 +1,35 @@
+# Warnings
+
+Read this file when a tool result or a replay report contains a warning. A warning does not reject a step. Find which warning it is. Then do the action for it.
+
+## Capture warnings
+
+The recorder kept raw coordinates for a tap. If the warning shows a tree error, the tree was unreadable. Repair the tree (platform file). Then record the tap again. For all other causes, do the [coordinate fallback gate](record.md#coordinate-fallback-gate). A `tap: { role: ... }` with no other key gets no warning. Do the same for it.
```
j-piasecki, line 7: > A `tap: { role: ... }` with no other key gets no warning. Do the same for it.

Since it's being treated as if it did, maybe it should warn?

packages/skills/skills/argent-create-flow/SKILL.md
```diff
@@ -1,49 +1,42 @@
...
+description: Create, record, edit, replay, or repair reusable Argent flow YAML files. Use when the user asks to record or replay a repeatable device path, set up profiling or an A/B comparison, or invoke the authoring engine behind argent-qa-flows. Also use before repeating three or more interactions. Supports iOS, Android, and Chromium. For one-off UI checks, acceptance-driven regression tests, or screen video, use argent-test-ui-flow, argent-qa-flows, or argent-screen-recording respectively.
 ---
 
 # Create an Argent flow
 
-An Argent flow is a replayable sequence in `.argent/flows/<name>.yaml`.
+A flow is a list of steps in `.argent/flows/<name>.yaml` that Argent can replay. Flows run on iOS, Android, and Chromium. Vega, Apple TV, and Android TV are not supported at this time. If the device is a TV, tell the user. Then stop.
```
j-piasecki: In another file, it says that recording doesn't work on TVs, but replay of manually defined flows does. It looks like it contradicts itself.

> The agent does not record flows on TV targets (Vega, Apple TV, Android TV) yet, but `argent flow run` still replays a hand-written Vega flow.

j-piasecki (commented):
If the TV flow playback is going to be supported now, I fear that omitting (or saying it doesn't work) it from skills would make the agent claim it won't work when asked for help - with syntax for example.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Reviewed the skill rewrite by running it rather than reading it. I built the tool-server from this branch and followed the new corpus exactly as an agent would — `list-devices`, then `platforms/ios.md`, `record.md`, `polish.md`, `replay.md` — recording a sign-in flow against Bluesky on an allocated iOS 18.6 simulator. The authored flow replays green both ways (`flow-execute` -> `ok: true, passed: 4, failed: 0, errored: 0`; `argent flow run e2e-signin --device <udid>` -> `PASS — 4 passed, 0 failed, 0 errored, 0 skipped`), and the prose about echoes not counting as verdicts and the `assumes:` line rendering both held live. The structure of the rewrite is a real improvement over the three files it replaces — the platform split and the warning routing table are much easier to follow from inside a recording.

Moves covered: claims vs code, nearest twin (each rule against its sibling platform file and against the pre-PR text at `2740a3ec3`), non-happy paths, inputs, reachability both ways, what outlives the call, plus the absence pass against the deleted corpus. All 14 fenced and 29 inline YAML snippets were parsed through the real `parseFlow`; all 35 internal links and anchors resolve; the six `rg` audit recipes were run against fixtures under `rg`, GNU `grep -E` and BSD `grep -E`.

Two things worth flagging above the line comments.

First, the docs are now unpinned. `test/flows/flow-skill-docs.test.ts` is deleted, and nothing replaces it: `scripts/grade-skills.mjs` scores `SKILL.md` shape only and never opens `references/**`, and `skill-description-quality.yml` is path-filtered to `**/SKILL.md`, so the nine added and three deleted reference files trigger no job. I measured this rather than assuming it — I planted four defects in the docs at once on a throwaway worktree (the idle floor changed 600 -> 900 ms, a `within:` selector changed to an unparseable `inside:`, the idle verdict inverted to "`idle` FAILS on timeout", and a warning bullet deleted) and prettier, `tsc --build`, knip, eslint, `grade-skills.mjs`, the repo-hygiene greps, `argent-installer` and the full tool-server suite (363 files / 4587 tests) all stayed green. Under the deleted file, the inverted-verdict mutation would have failed by name.

Second, the branch is behind `main` in a way that changes the diff's meaning. `gh pr view` reports `mergeable: CONFLICTING`, and `git merge-tree origin/main HEAD` gives `CONFLICT (modify/delete)` on all three deleted reference files plus a content conflict in `argent-qa-flows/SKILL.md`. No CI run has ever evaluated this diff against current `main`, and `swipe` landed on `main` in `e314b3ef4` after this branch's merge-base — so a couple of the line comments below are about what the merge produces, not about what the branch says today.

CI is green at `fe549682f` (9/9). Everything below is a line comment; I have not attached a verdict, and the approve or changes-requested call is not mine to make.

packages/skills/skills/argent-create-flow/references/record.md
```diff
@@ -0,0 +1,83 @@
+# Record
+
+Read this file before you record a flow or record steps again. Do each action through the recorder. Edit the YAML only after `flow-finish-recording`.
+
+## Recorder contract
+
+- Give the same `name` and the same absolute `project_root` to each recording tool. Use a name that is unique to your task. Give each concurrent recording its own device.
+- `flow-start-recording` erases the named YAML, also a complete flow. Make a copy of the previous flow first. If a call says that the recording is inactive, record with a new name.
+- After each `flow-add-step` call, read `recorded`, `message`, and `toolResult`. The recorder also records a call that returns an unmet condition.
+- `flow-add-step.command` is a tool name, and `args` is a JSON string. A directive name gets guidance and records nothing.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: "A directive name gets guidance and records nothing" is not true of the two directive names the recorder actually treats this way.

`UNHINTED_DIRECTIVE_KEYS = ["rotate", "tool"]` (`packages/tool-server/src/tools/flows/flow-add-step.ts:744-749`) is the set with no recording tool, and neither is inert when passed as `command`. `rotate` resolves to the registered device-orientation tool: I called `flow-add-step` with `command: "rotate"` against a booted iOS 18.6 simulator on this branch and the simulator physically rotated, with `tool: rotate` written into the flow. `tool` returns `REGISTRY_TOOL_NOT_FOUND`.

So for `rotate` the sentence is wrong in the direction that costs something: it says nothing happens, and what happens is a real device state change plus a recorded step, mid-recording, on the device the flow is being captured on.

packages/skills/skills/argent-create-flow/references/warnings.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Recorded wait warnings
+
+`flow-add-step` and `flow-finish-recording` give a warning about a recorded `await-ui-element`:
+
+- The live wait failed. Read `cause` ([Record: Live waits](record.md#live-waits)).
+- The condition did not hold on the runner tree. For `text`, first make sure that the selector matches only one element. Then make sure that the screen did not change. If the two trees are different, use a selector from the runner tree. Then replay it.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: "If the two trees are different, use a selector from the runner tree" is the opposite of what the code advises for a `hidden` wait, and the warning being routed from says so in the same breath.

`retargetRemedy` (`flow-add-step.ts:129-141`) carries the header comment "The advice inverts on `hidden`", and `runnerSideReadClause` splices that inverted branch into every determinate re-probe warning.

Reproduced on a booted iPhone against a React Native app: recording `await-ui-element {condition: "hidden", selector: {identifier: "chooseAccountBtn-alice.test"}}` (an id the flow tree carries and the accessibility tree does not) returned the divergence warning ending `…but this verdict says that tree still HAS the element, so retargeting at an `id` it definitely carries is the wrong direction — either narrow the selector until it matches only what you expect to leave, or gate on something that does leave, and prove it with `flow-execute`; or keep the step raw`.

The trigger is one this skill mandates, not a corner case: `record.md:35` (Absence) and `record.md:39` (Overlays) both require recording a `hidden` wait. Following this line on such a wait retargets at something the runner tree definitely holds, which makes the `hidden` directive fail harder at replay.

Separately in the same clause: "use a selector **from** the runner tree" names an action with no tool behind it — the same warning states "No read-only tool reports the runner's projection on iOS", and `runnerSideReadClause` says the equivalent for Android and Chromium, deliberately ("naming one of them would point the author at the wrong tree").

packages/skills/skills/argent-create-flow/references/warnings.md
```diff
@@ -0,0 +1,35 @@
...
+## Idle warnings
+
+`await: { idle: true }` does not fail a run. Each result other than a clean settle passes with one of these warnings:
+
+- **The screen did not stop.** A video, a shimmer, or a carousel is healthy. A load that did not finish looks the same. Look at the screen. Make sure that the next step targets a stable element.
+- **A small part continued to move.** A spinner, a caret, or a progress dot moved. Look at the screen.
+- **The screen did not move for the last N ms.** The wait ran out during the hold. Increase the `timeout:` of this step.
+- **The tree stayed empty.** The screen had no accessible content. Make sure that the screen is shown.
+- **Settled on the UI tree alone.** The runner could not read a screenshot pair. Examine the capture path.
+- **Too few reads.** The step ended with no evidence. Set the `timeout:` to at least three times the read time that the warning names, or repair the tree source.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: No idle warning names a read time, so "at least three times the read time that the warning names" cannot be evaluated by the agent that is told to evaluate it.

The five idle warnings are built at `packages/tool-server/src/tools/flows/flow-actions.ts:1847, 1868, 1892, 1913, 1924`, plus the localized-motion one at `:1730`. The "too few reads" text this bullet is about (`:1868`) reports a **count** — `contentReads`, `timeoutMs`, and `IDLE_POLL_MS` — and no per-read duration. `slowestAnsweredReadMs` is tracked (`:1559, :1602`) but is only ever compared against a budget at `:1619`; it never reaches a message. The one read *duration* in the file is `HUNG_TREE_READ_MS = 2000` (`:1820`), which belongs to the `ok: false, indeterminate: true` path — the errored case the next line of this file covers separately.

The reading an agent can act on makes it worse rather than merely unusable: taking "read time" as the 200 ms poll gives a `timeout:` of 600 ms, which is exactly the parser's floor (`idleMinimumTimeoutMs(250) = max(400, 250) + 200 = 600`, the number `yaml.md:97` quotes). That is a 12x *reduction* from the 7500 ms default, on a step that just reported it had too little time — the opposite of the remedy. The pre-PR text carried no multiplier at all (`2740a3ec3:packages/skills/skills/argent-create-flow/references/flow-yaml.md:200`: "a settle needs three reads across two intervals and this step got fewer"), so the quantity was introduced here.

packages/skills/skills/argent-create-flow/references/yaml.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Directives
+
+| Directive    | Shape                                                              | Notes                                                     |
+| ------------ | ------------------------------------------------------------------ | --------------------------------------------------------- |
+| `launch`     | `<app id>` or a [launch map](#launch-map)                          | Restarts the app and waits until the app is available     |
+| `tap`        | `<selector>`, `{ on, times }`, or `{ x, y }`                       | `times: 2` double-taps. `x` and `y` are normalized        |
+| `long-press` | `<selector>` or `{ on, duration }`                                 | `duration` in ms                                          |
+| `type`       | `{ into, text, submit }`                                           | Presses Enter unless `submit: false`                      |
+| `scroll-to`  | `{ target, direction, within }`                                    | `direction` is `down` (default), `up`, `left`, or `right` |
```
latekvo, line 47: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `within` on `scroll-to` is not the match scope it is ten lines above, and the two spellings that result are both accepted with different meanings.

`yaml.md:37` defines `within` as a selector scope ("visual containment"). On `scroll-to` it is neither: `packages/tool-server/src/tools/flows/flow-actions.ts:714-736` reads `const region = within ? flowSelectorToFrame(tree, within) : FULL_SCREEN;` and then matches `const frame = flowSelectorToFrame(tree, target);` — `target` is resolved against the whole screen, and `within` only selects the surface the gesture is applied to and the clip window.

Confirmed with the real parser (`parseFlow` / `serializeFlow` via `npx tsx`), both forms parse and they are not the same step:

```
scroll-to: { target: { text: Delete }, within: { id: card } }
  -> { kind: "scroll-to", target: { text: "Delete" }, within: { identifier: "card" } }
scroll-to: { target: { text: Delete, within: { id: card } } }
  -> { kind: "scroll-to", target: { text: "Delete", within: {...} } }
```

The first scrolls the `card` surface while hunting for any `Delete` on screen; only the second constrains which `Delete` counts. Nothing marks the difference. And the mismatch is not caught downstream: at `:757-762`, once the fingerprint stops changing, `if (frame) return { frame }` accepts a match found outside the container, so a flow that scrolls the wrong element to a same-named row elsewhere on screen reports success. The pre-PR corpus gave the row a gloss that at least fixed the intent (`2740a3ec3:…/references/flow-yaml.md:117`, "Use `within` for a nested scroller"); nothing in the new corpus or in `flow-execute`'s own description replaces it.

packages/skills/skills/argent-create-flow/references/polish.md
```diff
@@ -0,0 +1,96 @@
...
+- If the recorder wrote `text:` for an element that has a stable id, change the selector to the id. The replay proves it.
+- Convert `textMatch: equals` to `equals:`. Convert other text checks to `contains:`.
+- A focus tap plus one text-only `keyboard` call becomes `type:` with `submit: false`.
+- If the conversion changes the behavior, keep the raw tool step. Examples are a point-anchored pinch, a swipe with a tested speed, and a rotation with a tested start angle. A wait with `pollIntervalMs` or `bundleId` also stays raw.
+- Keep screenshots as evidence for a person. Use `snapshot:` when the runner must compare the screens.
+- Replay the full flow after the conversion ([Replay](replay.md)).
+
+## Steps that you can add by hand
+
+You can add only these steps by hand, at states that you saw live:
```
latekvo, line 36: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This list is normative — `SKILL.md:27` reads "You can add by hand only the [steps that you can add by hand]" — and it omits two directives that no tool records and that the rest of the corpus instructs adding by hand.

`when:` — `flow-add-step`'s own guidance, read live from the running tool-server, is: `"when" is a flow directive, not a tool, and no tool records one — it GUARDS the steps nested under it, so there is no action of its own to run. Record those steps, then wrap them in the `when:` block by hand during polish and prove both branches with the replay.` Meanwhile `yaml.md:104-111` documents `when:` blocks in full, `polish.md:91` audits `when:` guards, and `argent-qa-flows/SKILL.md:64` ("Do not put acceptance evidence in `when:`") presupposes QA flows contain them. There is no conversion-table row for `when:` either, so an agent obeying rule 5 can never produce one.

`wait:` — same shape. `flow-add-step`: `"wait" is a flow directive, not a tool, and there is no tool that records one`. Yet `platforms/ios.md:37` ("Put a `wait:` or a raw wait before a gesture that follows a screen change") and `warnings.md:35` ("Put a `wait:` before a gesture that follows a screen change") both instruct the insertion, `yaml.md:52` documents the directive and `polish.md:92` audits for it.

The PR description says `long-press:` was added here "matching what `flow-add-step` already tells the agent"; by that same test `when:` and `wait:` belong on the list.

packages/skills/skills/argent-create-flow/references/yaml.md
```diff
@@ -0,0 +1,130 @@
...
+| `await`      | `{ <condition>, timeout }` or `{ idle: true, stableFor, timeout }` | Default `timeout` 7500 ms                                 |
+| `assert`     | `{ <condition> }`                                                  | Fixed 1000 ms grace. Rejects `timeout`                    |
+| `wait`       | `<ms>`                                                             | Give it an echo and a hard check after it                 |
+| `snapshot`   | `<name>` or `{ name, maxMismatch, cropOn }`                        | See [Snapshots](#snapshots)                               |
+| `run`        | `<path>`                                                           | See [Composition](#composition)                           |
+| `when`       | `{ <condition> }` or `{ platform }`, with `steps:`                 | See [Optional steps](#optional-steps)                     |
+| `echo`       | `<message>`                                                        | Printed in the report                                     |
+| `tool`       | `<tool name>`, with `args:` and optional `delayMs:`                | One of the Argent tools                                   |
```
latekvo, line 58: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This table reads as the complete directive list, and it is missing one that `origin/main` ships.

`swipe` entered `STEP_DIRECTIVE_KEYS` in `e314b3ef4` (2026-08-28), after this branch's merge-base `2740a3ec3` (2026-08-25): main's list has 16 keys, this head's has 15. So the omission is a rebase artifact rather than an authoring slip — but it lands as a real gap, because the deleted `references/flow-yaml.md` was `swipe`'s only skill documentation and `gh pr view 973` currently reports `mergeable: CONFLICTING`, with `git merge-tree origin/main HEAD` giving `CONFLICT (modify/delete)` on exactly those deleted files. Resolving that conflict the way the PR intends — taking the delete — removes the documentation without this table replacing it.

The loss is partial, which is what makes it easy to miss: `origin/main:flow-run.ts:1201-1205` (the `flow-execute` description) still carries the shape, "direction is the FINGER's travel", the `by` bounds and the 300/150/10000 durations. What exists nowhere else is "Never use `swipe` to scroll", the preset magnitudes, the OS-edge-margin rule and the Chromium drag-and-drop trap (`2740a3ec3:…/flow-yaml.md:123, 137, 140, 144`). `record.md:57` and `polish.md:28` route every movement to `scroll-to`, which is correct at the merge-base but contradicts main's explicit `swipe` scroll rule once merged.

packages/skills/skills/argent-create-flow/references/platforms/ios.md
```diff
@@ -0,0 +1,40 @@
...
+4. If the failure continues, report that the instrumentation stops the flow. A missing tree does not let you use coordinates in a QA flow.
+
+## System apps
+
+A `com.apple.*` app does not get a flow tree. Use this form only for a `com.apple.*` app:
+
+- Keep `launch:`. It passes after about 16 seconds.
+- Use raw `tool: await-ui-element` checks and point taps from `describe`, each with an echo.
+- For text, use a point focus tap, a raw `keyboard` with `text` and `delayMs: 500`, then a raw `keyboard` with `key: "enter"`.
+- Use raw swipes with `settle: true`, because the tree is necessary for `scroll-to`.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `settle` no longer exists on `origin/main`, and this is the one place the corpus says coordinates are legitimate, so every raw swipe in the `com.apple.*` form is written in a spelling that fails.

`d4a712c40` renamed the flag to `momentum` with the opposite sense after this branch forked. On main, `gesture-swipe` declares `settle: z.never({ error: "gesture-swipe's `settle` was renamed to `momentum`, with the opposite sense — write `momentum: false` for the momentum-free swipe that `settle: true` used to mean…" })`, declared that way deliberately so "this non-strict object refuses it instead of stripping it and flinging - the exact inverse of the gesture the caller asked for." The flow parser refuses `swipe.settle` by name too (`flow-utils.ts`, "never aliased: `settle: true` maps to `momentum: false`, so a silent rewrite would invert what the author wrote").

The companion constraint also has no home anywhere in the new corpus: `momentum: false` requires `durationMs >= 150` and is rejected below it, because a shorter ease-out flings harder than a plain swipe and on Android flings backwards. The deleted `reliability-and-recovery.md:63` carried both the flag and the floor.

packages/skills/skills/argent-create-flow/references/polish.md
```diff
@@ -0,0 +1,96 @@
...
+  - await: { idle: true }
+```
+
+## Audit
+
+Run these searches before the replay (`rg`, or `grep -nE` if `rg` is not installed):
+
+```text
+rg -n '(\{ *x:|^ +(x|centerX|fromX|toX):|gesture-(tap|swipe|scroll|drag|pinch|rotate|custom))' .argent/flows/<name>.yaml
+rg -n -B2 '^ +role:' .argent/flows/<name>.yaml
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This search only matches block-style YAML, so it misses the form the same file's own example writes.

Run against a file containing `- tap: { role: Button }` and `- long-press: { role: Link }`, `rg -n -B2 '^ +role:'` returns nothing (rc=1); it fires only on the block form `tap:\n      role: Button`. The converted example at lines 61-69 is entirely flow style (`- tap: { id: settings-tab }`), and this is the hand-conversion pass, so a role-only selector introduced or left here is exactly what escapes.

The audit bullet at line 88 ("No `tap:` or `long-press:` has `role:` as its only key") therefore reports clean on those selectors, and line 77's coordinate search does not cover them either. The identical regex was in the deleted `live-authoring.md:224`, but that file also carried the premise it depends on — "Recorder output is always block style" — which is not in the new file.

packages/skills/skills/argent-create-flow/references/polish.md
```diff
@@ -0,0 +1,96 @@
...
+rg -n '(udid|device_id|serial)' .argent/flows/<name>.yaml
+rg -n '(-selector-\d+|selector-\d+\b)' .argent/flows/<name>.yaml
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `\d` is not POSIX ERE, so the `grep -nE` fallback this block sanctions on line 74 silently finds nothing.

Against a file containing `settings-selector-3` and `android-selector-12`:

- `ggrep -nE '(-selector-\d+|selector-\d+\b)'` (GNU grep 3.12) → `warning: stray \ before d` twice, rc=1, no matches
- `rg -n '(-selector-\d+|selector-\d+\b)'` → both lines
- macOS `/usr/bin/grep -nE` (BSD) → both lines

So this passes on a dev Mac and returns clean on any GNU-grep box, including Linux CI. It is the only search in the block guarding `SKILL.md:25`'s ban on positional ids; the `\s` and `\b` used elsewhere in the block are GNU extensions that do work under both greps.

packages/skills/skills/argent-create-flow/references/polish.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+## Audit
+
+Run these searches before the replay (`rg`, or `grep -nE` if `rg` is not installed):
+
+```text
+rg -n '(\{ *x:|^ +(x|centerX|fromX|toX):|gesture-(tap|swipe|scroll|drag|pinch|rotate|custom))' .argent/flows/<name>.yaml
+rg -n -B2 '^ +role:' .argent/flows/<name>.yaml
+rg -n '(udid|device_id|serial)' .argent/flows/<name>.yaml
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This search cannot match the one device key the recorder is documented to leave in the file, so the audit rule it feeds passes on the case it exists for.

`replay.md:11` instructs the opposite of what `:90` then asserts: "A recorded `stop-all-simulator-servers` keeps its `devices` list … so record the scope or pass `device`." That is deliberate in the code — `DEVICE_BIND_KEYS = ["udid", "device_id", "device"]` are stripped from every recorded step (`flow-device.ts:23`, applied at `flow-add-step.ts:1251`), while `DEVICE_BIND_LIST_KEYS = ["devices"]` is deliberately *not*, with the comment at `flow-device.ts:40-46` explaining that "a SCOPE survives into the YAML".

So the flow the corpus tells the agent to produce contains a literal UDID under `devices:`, `:90` asserts "No device id or literal credential is in the file", and this search matches neither `devices` nor anything else that would surface it. An agent following `replay.md:11` and then running this audit is told its flow is clean while the id is still there — and because the recorded scope is the thing that keeps `stop-all-simulator-servers` from becoming the machine-wide sweep, the id is one an agent may well try to remove once it notices, which reintroduces the sweep.

packages/skills/skills/argent-create-flow/SKILL.md
```diff
@@ -1,49 +1,42 @@
...
-   - **e2e:** the first non-echo step is `launch:`. The flow controls process start.
-   - **fragment:** there is no leading launch. Declare a precise `executionPrerequisite`.
-2. Follow [Live authoring](references/live-authoring.md): start, record one verified step at a time, finish, polish, audit, and replay.
-3. Report the file, replay command, result, prerequisite or side effects, and every coordinate or raw-gesture exception.
-
-## Proactive recording
-
-Before repeating three or more interactions, tell the user and start a recording. Record that run and replay it afterward. A completed path cannot be recorded retroactively.
+1. Select the flow type. An **e2e** flow starts with `launch:`. A **fragment** has no launch at the start and declares an `executionPrerequisite`. If the app has no control for the action in the task, report it. Then stop.
+2. Do the record, polish, audit, and replay phases. Each phase file gives the steps.
```
latekvo, line 38: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: "audit" is named as one of four phases and "Each phase file gives the steps", but the router at lines 15-18 lists three phase files and none of them is an audit file.

The audit lives at `references/polish.md:72`. `grep -rn 'polish.md#' packages/skills packages/docs` returns exactly two hits, both `#steps-that-you-can-add-by-hand` — nothing links `polish.md#audit`. `references/replay.md:56` orders "After each correction, audit and replay from the declared start" without a link, and `argent-qa-flows/SKILL.md:86` says "Audit again." without one either.

An agent on the repair path takes router step 2 to `replay.md` only, hand-edits YAML per `replay.md:38`, and is then told to audit with no route to the six searches and eleven checks that catch stored device ids, literal credentials, coordinate and role-only selectors, and a missing `idle`. Before this change `SKILL.md` sent the agent to read `live-authoring.md` completely, and its `## Blocking audit` was reachable from there.

packages/tool-server/test/flows/flow-skill-docs.test.ts
```diff
@@ -1,205 +0,0 @@
...
-        `steps:\n  - await: { idle: true, stableFor: ${stableFor}, timeout: ${t} }\n`;
-      expect(() => parseFlow(step(smallest)), `${stableFor}`).not.toThrow();
-      expect(() => parseFlow(step(smallest - 1)), `${stableFor}`).toThrow(
-        new RegExp(`at least ${smallest}ms`)
-      );
-    }
-  });
-});
-
-describe("create-flow directive-answer docs", () => {
```
latekvo, line 165: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: These two tests read no skill markdown at all, so the prose rewrite does not invalidate them — they pin `flow-add-step`'s `description` and its `command` parameter description against `STEP_DIRECTIVE_KEYS` / `directiveCommandHint`, and they still pass at this head. They go out with the file.

After this PR nothing pins any flow doc to the code. I checked what is left and then measured it. `scripts/grade-skills.mjs` scores `SKILL.md` shape only and never opens `references/**`, and `skill-description-quality.yml` is path-filtered to `**/SKILL.md`, so the nine added/deleted reference files trigger no job at all; the only surviving markdown-to-code test in the repo is `test/ios-instruments/degraded-xcode-doc-consistency.test.ts`, for an unrelated profiler doc.

Measured by mutating the docs only, on a throwaway worktree at this head, four planted defects at once: `yaml.md:97` "smallest `timeout` is 600 ms" → 900 ms; `yaml.md:32` `within:` → `inside:` (which `parseFlow` rejects — "selector has unknown key `inside`"); `yaml.md:99` inverted to "`idle` FAILS on timeout"; and the `**Too few reads.**` bullet removed from `warnings.md`. Result: prettier, `tsc --build`, knip, eslint, `grade-skills.mjs` (10.0/10), the repo-hygiene greps, `argent-installer` (555 tests) and the full tool-server suite (`363 files / 4587 tests`) all stayed green. Under the deleted file, mutation 1 alone would have failed by name — it asserted `not.toMatch(/FAILS on timeout/i)` on exactly that wording, and pinned `600` and the `stableFor + 200` arithmetic to `idleMinimumTimeoutMs`.

For contrast, mutating `IDLE_POLL_MS` 200 → 250 in the source does fail 12 tests — but not one of them names a document, so the doc half of each of those constants is now free to drift alone. The comment this file carried on the directive test says what it was for: "names every directive it answers, so a new one cannot go unmentioned" — and `swipe` was added on main after this branch forked and is absent from the new corpus.

packages/skills/skills/argent-create-flow/references/warnings.md
```diff
@@ -0,0 +1,35 @@
+# Warnings
+
+Read this file when a tool result or a replay report contains a warning. A warning does not reject a step. Find which warning it is. Then do the action for it.
+
+## Capture warnings
```
latekvo, line 5: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This section only covers tap-coordinate warnings, but `flow-add-step` emits a second capture-warning family on the same `message` channel and the file offers no bucket for it: seven `kept the raw flow-execute step — …` texts at `flow-add-step.ts:992, 996, 1028-1031, 1056-1060, 1084, 1089-1093, 1099` (no flow name; a remote recording, where `run:` composition is host-resolved; a non-absolute `project_root`; the target not resolving to the recording's sibling; and so on).

The trigger is in-skill: `polish.md:22` lists "sibling `tool: flow-execute` → `run:` (the recorder writes it)", and `argent-qa-flows/SKILL.md:33` tells the agent to "record a safe reset or seed flow, then record a `flow-execute` of it (the recorder writes `run:`)". Line 3 frames this file as the routing table — "Find which warning it is. Then do the action for it" — so an agent that hits one of these lands in a section prescribing the coordinate fallback gate, which does not apply, and the parenthetical in `polish.md:22` has already told it the recorder handled the rewrite.

packages/skills/skills/argent-create-flow/references/platforms/android.md
```diff
@@ -0,0 +1,18 @@
...
+## Discovery and selectors
+
+- Keep the id form that the recorder wrote. An unqualified id also matches the qualified resource id.
+- `describe` shows trimmed interactable nodes, and it can fall back to `uiautomator`. Thus a healthy `describe` is no proof of the flow tree. A recorded `await-ui-element` that passes with no cross-tree warning is the proof. An id that `describe` does not show can work. Do a test of it in a scratch fragment ([Record](../record.md#coordinate-fallback-gate)).
+- The platform discovery tool is `debugger-component-tree` for a React Native app and `describe` for other apps. Do not record `debugger-*` calls.
+- `describe` does not report focus on Android. After the `keyboard` step, make sure that the field has the typed value.
+
+## Missing flow tree
+
+If a result says that the android devtools helper is unavailable, unlock the device. Make sure that the device permits `adb install -t`, because Argent installs the helper APK itself. Then call `restart-app`. Record those steps again.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `restart-app` is not a remedy here. The Android flow tree is served by a separate `am instrument` helper process that the registry spawns, independent of the app under test — `flow-run.ts:479-488` notes "Unlike iOS's native-devtools … the Android helper is a separate `am instrument` process", and `restart-app` on Android is only `am force-stop` + `am start` (`restart-app/platforms/android.ts:19-28`), which touches nothing the helper depends on.

The runner's own remedy text names no relaunch — `flow-android-tree.ts:209-210`: "the android devtools helper is unavailable (…) confirm the device is unlocked and the helper can be installed (`adb install -t`)" — and the deleted `reliability-and-recovery.md:79` said the same without a restart. Following this line spends a relaunch, and a green `restart-app` reads as evidence the tree is repaired when it is not.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -3132,7 +3132,7 @@ export async function writeNewFlowFile(filePath: string, content: string): Promi
  * and this host cannot read it at all, so the in-memory copy is both the take
  * and the only thing countable — the guarantee below does not carry across that
  * boundary. The agent-facing statement of it lives in
- * `packages/skills/skills/argent-create-flow/references/live-authoring.md`.
+ * `packages/skills/skills/argent-create-flow/references/record.md`.
```
latekvo, line 3552: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `record.md` does not state the guarantee this now points at. The JSDoc below names it precisely — the file rather than the in-memory `flow` is the take in "host" mode, `appendStep` re-reads it before every append, and `flow-finish-recording` reads it back for its summary, none of which carries into "client" mode. `record.md` never mentions host versus client mode, re-reading, or the in-memory copy; its closest sentence is line 3, "Edit the YAML only after `flow-finish-recording`", which is the derived instruction rather than the guarantee.

The previous target did carry it: `live-authoring.md:31` had "To see the whole file mid-recording, read it at `savedTo`. A `savedTo` that comes back `null` means the write failed on your side. The step is still in the recording, so continue: the next step rewrites the whole file, and `flow-finish-recording` returns `flowFile` regardless." That is gone — `record.md:9` lists only `recorded`, `message` and `toolResult`, and `flow-add-step`'s own description says only "Returns { message, toolResult, stepCount, recorded, savedTo }" without the semantics (those live on `flow-start-recording`'s description). An agent that sees `savedTo: null` mid-recording now has no statement that the take is intact, and the intuitive repair — restarting the recording — truncates it, per `record.md:8`.

packages/skills/skills/argent-qa-flows/SKILL.md
```diff
@@ -1,123 +1,92 @@
...
+1. Select one runner for the two passes: `flow-execute`, or `argent flow run <name> --platform <platform>` for CI. A change of runner resets the streak.
+2. Write and examine the snapshot baselines. Then freeze them. A baseline update is not a pass.
+3. Before pass 1 on iOS or Android, call `stop-all-simulator-servers` with `devices: [<device>]`, or `argent run stop-all-simulator-servers --devices <device>` from the standalone runner. A call without `devices` stops the devices of all agents on this machine. On Chromium, the restarted service is the instance that the runner boots itself, so do not pass `device`.
+4. Run from the launch, without baseline-update mode. When the report shows `PASS` and each acceptance check ran, count a pass. An `errored` step does not increase the streak. If the step could not run, repair the environment. Then run again. A failed `launch:` is a verdict about the app. Report it.
+5. Resolve each warning before the flow is complete. Read [Warnings](../argent-create-flow/references/warnings.md).
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Step 5 requires resolving every warning, but the file it links to defines a class of warnings whose documented outcome is acceptance, not resolution.

`references/warnings.md:22` opens the idle section with "`await: { idle: true }` does not fail a run. Each result other than a clean settle passes with one of these warnings", `:24` says "A video, a shimmer, or a carousel is healthy", and `:32` closes with "Report each idle warning that you accept, with its cause." `platforms/ios.md:38` states it outright for a recorded wait: "This warning is expected." The selector-less gesture section is the same shape — `warnings.md:35`, "Accept the warning only for an app with no tree".

The trigger is this skill's own item 4 at `:17`, which makes `await: { idle: true }` mandatory after every screen change. On a screen with a legitimately animating element — the carousel and shimmer that `warnings.md:24` calls healthy — that mandatory step passes carrying the warning on every run, so it is not resolvable. The completion gate at `:12` then cannot be satisfied for such a flow, and the two routes out are to weaken or delete the mandatory `idle` check, which `references/replay.md:56` and `argent-create-flow/SKILL.md:42` both forbid.

`:90`'s report list is aligned with the same reading: it has a slot for "the resolved warnings" and none for accepted ones, while `warnings.md:32` requires the accepted ones be reported. The pre-PR file carried the acceptance path — `2740a3ec3:packages/skills/skills/argent-qa-flows/SKILL.md:104`, "Inspect the screen, disclose the cause, and verify that surrounding acceptance checks use stable elements rather than stillness" — and it has no home in the new corpus.

packages/skills/skills/argent-qa-flows/SKILL.md
```diff
@@ -1,123 +1,92 @@
...
 
-A snapshot has no recorder form. Inspect its stable state during the walkthrough, then add the planned snapshot during polish. If direct recovery changes state, re-record the affected behavior. A recovered walkthrough is not proof.
+## 3. Make the evidence specific
 
-## 3. Make evidence discriminating
+- **State change:** show the new state. When the two can match, also show that the previous state is not shown.
+- **Cancel or persistence:** go across the commit boundary. After cancel or save, go out of the screen. Go into it again. Then make sure of the kept state.
+- **Absence:** show the screen that contains the element. Record the same selector as `visible`, then the action, then `hidden`. In a collection, absence in the viewport is not global absence. Use a fixed seeded position, a count, or an empty state.
+- **Repeated controls:** use an id. If there is none, use `within` with a stable container and `text.in` for membership.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `text.in` is not a membership test, so this prescribes a check that does not answer the question the bullet poses.

This PR's own `references/yaml.md:80` defines it as single-element: "`text` compares the text of the element in `in` with one comparator", and `yaml.md:27` says "`text` reads the first visible match". The implementation agrees — `packages/tool-server/src/utils/ui-tree-match.ts:518`, `const first = firstInReadingOrder(matches.filter(isVisible)) ?? firstInReadingOrder(matches);` — the `text` arm reduces the match set to one node and compares only that node's text, unlike the `exists` / `visible` arms just above it (`:511-515`), which quantify over the set.

Run against the real matcher with three rows Alice / Carol / Bob inside a `profile-list` container:

```
text.in{any, within: profile-list} contains "Alice" -> true   (matches=3)
text.in{any, within: profile-list} contains "Carol" -> false  (matches=3)
text.in{any, within: profile-list} contains "Bob"   -> false  (matches=3)
```

So for the case the bullet is about — repeated controls with no id — a QA agent encoding "this row is present in the list" as instructed gets a verdict about row 1 only. It reports a false failure for a member that is genuinely rendered, and its result flips on row-order changes that the requirement does not care about. Because the surrounding skill treats these as hard acceptance checks (`:14` item 3), the flow certifies the wrong thing rather than erroring.

packages/skills/skills/argent-qa-flows/SKILL.md
```diff
@@ -1,123 +1,92 @@
...
-6. Run the same YAML again immediately with the same runner. Do not manually reset app or account data.
-7. Reset the streak after any failure, edit, re-recording, baseline update, or state-changing manual recovery. Repair through `argent-create-flow`, audit again, and restart with fresh services.
+1. Select one runner for the two passes: `flow-execute`, or `argent flow run <name> --platform <platform>` for CI. A change of runner resets the streak.
+2. Write and examine the snapshot baselines. Then freeze them. A baseline update is not a pass.
+3. Before pass 1 on iOS or Android, call `stop-all-simulator-servers` with `devices: [<device>]`, or `argent run stop-all-simulator-servers --devices <device>` from the standalone runner. A call without `devices` stops the devices of all agents on this machine. On Chromium, the restarted service is the instance that the runner boots itself, so do not pass `device`.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The closing clause reads as an instruction about `stop-all-simulator-servers`, which is the step's subject and whose scoping parameter differs from the named one only by a trailing `s`.

The step's imperative is "call `stop-all-simulator-servers` with `devices: [<device>]`", the sentence immediately before defines the unscoped form as the machine-wide sweep, and then the clause says "do not pass `device`". The intended referent is the runner's parameter documented two files away at `references/platforms/chromium.md:41-42` ("With `flow-execute`, pass `platform: chromium`. Do not pass `device`."), but nothing local marks the switch of subject.

Read as the Chromium variant of the step's own imperative, it licenses exactly the call the tool's schema warns about — `packages/tool-server/src/tools/stop-all-simulator-servers.ts:18`, "an unscoped stop also kills devices another agent is mid-session on" (the implementation gates on `const scoped = devices !== undefined;`). On a machine running several agents, which the repo's device-allocator rule treats as the normal case, that terminates other sessions' simulators.

The pre-PR text could not be read this way, because the guard sat in the same breath as the instruction: `2740a3ec3:packages/skills/skills/argent-qa-flows/SKILL.md:102`, "Never omit the scope — a bare call is the machine-wide sweep." That guard is not present anywhere in the new corpus.

packages/skills/skills/argent-qa-flows/SKILL.md
```diff
@@ -1,123 +1,92 @@
...
-6. The unchanged YAML passes twice with the same runner. Pass 1 starts with fresh mobile Argent services, and pass 2 follows immediately.
+1. The first step after an echo is `launch:` (on Chromium, the launch that you add in the polish pass). In-flow setup shows a deterministic data baseline. Subsequent runs do not collect data, and manual cleanup is not necessary.
+2. The first walkthrough recorded each action and each structural check live. Only the [steps that you can add by hand](../argent-create-flow/references/polish.md#steps-that-you-can-add-by-hand) are unrecorded.
+3. Each requirement maps to a hard `await:`, `assert:`, or examined `snapshot:`. Echoes and screenshots are not verdicts.
+4. Each screen change has an identity check and then `await: { idle: true }`.
+5. Each target obeys the stable-selector rule and the coordinate fallback gate. A QA flow keeps a coordinate only for an element with no label.
+6. The unchanged YAML passes two times with the same runner. Pass 1 starts with restarted Argent services. Pass 2 follows immediately.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Item 6 requires "restarted Argent services" without qualification, but the only procedure for it is restricted to two platforms: `:82` scopes the restart to "Before pass 1 on iOS or Android", and tells a Chromium run to skip it. A Chromium QA flow therefore has a definition-of-done item whose procedure resolves only in the final clause of a step three sections later. The pre-PR wording did not have the gap — `2740a3ec3:packages/skills/skills/argent-qa-flows/SKILL.md:23` said "fresh **mobile** Argent services".

Related to the same item: `2740a3ec3:…:102` also carried "The reset must not change app or account data.", which has no home in the new corpus — I grepped this file, all eight `argent-create-flow` reference files and `rules/argent.md`. `:85`'s "Do not reset app or account data by hand" is scoped to the interval *between* the two passes, not to the pre-pass-1 service reset, so nothing now bounds what "restarted Argent services" is allowed to touch.

packages/skills/skills/argent-qa-flows/SKILL.md
```diff
@@ -1,123 +1,92 @@
...
-6. Run the same YAML again immediately with the same runner. Do not manually reset app or account data.
-7. Reset the streak after any failure, edit, re-recording, baseline update, or state-changing manual recovery. Repair through `argent-create-flow`, audit again, and restart with fresh services.
+1. Select one runner for the two passes: `flow-execute`, or `argent flow run <name> --platform <platform>` for CI. A change of runner resets the streak.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `--platform` alone does not identify a device when more than one of that platform is booted, which is the condition this repo's own multi-agent setup produces.

`packages/tool-server/src/tools/flows/flow-device.ts:115-128` filters the booted devices by platform and then, for `scoped.length > 1`, throws `"${scoped.length} booted devices matched — pass --device or --platform to disambiguate."` The CLI hands the flag straight to that resolver (`packages/argent-cli/src/flow.ts:913`), so the recommended CI invocation aborts on device resolution rather than running the flow.

It bites precisely where the skill needs it: this step is the runner for both proof passes, and `:82` one line later already hands the agent the `<device>` id it would need. The new corpus states the qualification only for one platform — `references/replay.md:9` and `references/platforms/ios.md:3` — with nothing for Android, whereas the pre-PR corpus stated it unconditionally at `2740a3ec3:packages/skills/skills/argent-create-flow/references/flow-yaml.md:209`, "Pin `--platform` and `--device` for iOS, Android, or Vega".

packages/skills/skills/argent-create-flow/references/yaml.md
```diff
@@ -0,0 +1,130 @@
...
+`run: ../shared/login.yaml` resolves against the directory of the flow file. The `.yaml` suffix is optional. A nested fragment or e2e flow runs inline, and a nested launch restarts the app. If the `run:` chain of a fragment gets to a launch, the fragment cannot declare `executionPrerequisite`.
+
+## Snapshots
+
+```yaml
+- snapshot: checkout-summary
+- snapshot: { name: price-card, cropOn: { id: price-card }, maxMismatch: 0.2 }
+```
+
+A snapshot compares the screen, or the frame of `cropOn`, with a baseline in `.argent/flows/__baselines__/<flow>/`. A missing baseline, a mismatch above `maxMismatch` percent (default 0.5), or a `cropOn` dimension change fails the step. Write baselines from a known-good state with `--update-baselines` (`updateBaselines: true` in `flow-execute`). Examine each baseline. Let the user examine it. Do not commit it. A baseline update is not a test pass. Do not update a baseline only to make a diff pass.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: "Do not commit it" contradicts the runner's own instruction for the same file, and breaks the CI path a sibling skill recommends.

`packages/tool-server/src/tools/flows/flow-visual.ts:307-314` returns, on a missing baseline: `no baseline for "<name>" on this device class — expected <path>, nothing was compared. Run with updateBaselines (--update-baselines) to adopt the current screen, then review and commit it`. The comment directly above it (`:303-306`) says the branch fails *without* seeding deliberately, "a workspace that never persists baselines (ephemeral CI) would gate nothing forever".

So an uncommitted baseline is not a neutral state: every CI run of a snapshot-bearing flow takes that fail branch, and it never self-heals. `argent-qa-flows/SKILL.md:80` recommends `argent flow run … --platform <platform>` "for CI" and `:81` requires writing and freezing baselines, so following both files as written yields a flow that cannot pass in CI.

The pre-PR wording carried the qualifier that made it coherent — `2740a3ec3:packages/skills/skills/argent-create-flow/references/flow-yaml.md:207`, "Do not commit it **yourself**" — i.e. do not commit unreviewed, which is the same rule the runner states. The compressed form reverses it into a prohibition.

packages/skills/skills/argent-create-flow/references/polish.md
```diff
@@ -0,0 +1,96 @@
...
+rg -n '(-selector-\d+|selector-\d+\b)' .argent/flows/<name>.yaml
+rg -n '(visible|hidden|exists) *: *["'"'"'A-Za-z0-9]' .argent/flows/<name>.yaml
+rg -n '^\s*- wait:|open-url' .argent/flows/<name>.yaml
+```
+
+Correct each hit. Then make sure that:
+
+- Each action uses a stable selector, or the fallback gate let it through and you documented the step.
+- No `tap:` or `long-press:` has `role:` as its only key.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This audit item is absolute where the rule it audits has an exemption, so it fails flows the skill's own target ladder tells the agent to write.

`record.md:71` ranks a role-only target as a legitimate tier — "a role that is unique on the screen" — above the coordinate fallback, and the line directly above this one (`:87`) carries the matching exemption for selectors generally ("or the fallback gate let it through and you documented the step"). This line carries none: "No `tap:` or `long-press:` has `role:` as its only key."

Trigger: a screen with exactly one Button and no label or id — the case `record.md:71` is written for. The agent records `tap: { role: Button }` as instructed, reaches the polish audit, and this item fails a step that the recording rules explicitly sanctioned. The two ways out are to demote the step to the coordinate fallback that `record.md:71` ranks strictly lower, or to declare the audit passed against its literal text.

packages/skills/skills/argent-create-flow/references/record.md
```diff
@@ -0,0 +1,83 @@
...
+Do not remove a step during the recording. A wait in `run-sequence` gets no recorder warning, so read the nested result. The tool schema gives the selector fields. If `message` has a warning about the runner tree, read [Recorded wait warnings](warnings.md#recorded-wait-warnings).
+
+## Coordinate fallback gate
+
+Target sequence: a strict stable id, then narrow stable text or an accessibility label, then a role that is unique on the screen. Use `scroll-to` plus one of these for an off-screen target. Use raw coordinates only after the checks below.
+
+Do the gate when capture keeps a raw point or only a role. Keep the source screen open:
+
+1. Find candidates with the platform discovery tool.
+2. Do a test of each candidate in a scratch fragment with `assert: { visible: <candidate> }` on the correct screen. A scratch fragment is a temporary YAML file with an `executionPrerequisite` and only the assert step. Run it with `flow-execute` and `flow_path`. Then remove it. Examine each failure before you try a better id, label, or container.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This recipe returns a notice instead of a verdict on the first call, so the gate it implements never actually decides anything.

It prescribes a scratch fragment that carries an `executionPrerequisite`, then says to "Run it with `flow-execute` and `flow_path`". `packages/tool-server/src/tools/flows/flow-run.ts:1175-1183` returns before the run in exactly that case:

```
if (flow.executionPrerequisite && !params.prerequisiteAcknowledged) {
  return { flow: flowName, notice: "This flow has an execution prerequisite that must be fulfilled before it can run. …", executionPrerequisite: flow.executionPrerequisite };
}
```

That return precedes `resolveRunDevice`, so the `assert` never executes and no `ok`/`status` comes back. `prerequisiteAcknowledged` appears nowhere in this file; it is documented only at `replay.md:8`, which `SKILL.md:15-18` routes to in the *replay* phase, not the record phase where this gate lives. The requirement is new — `2740a3ec3:…/references/reliability-and-recovery.md:31` prescribed the same test with no `executionPrerequisite` on the fragment.

The consequence is that the candidate test reads as inconclusive-or-failed on every candidate, which is the input to "Keep coordinates only for an element with no id, text, or label" two lines down — so the gate resolves toward keeping the coordinate, which is the outcome it exists to prevent.

The same recipe also meets a second constraint this corpus dropped. `replay.md:7` specifies "the absolute `project_root`" but says only "`flow_path` for a file elsewhere", while `flow-run.ts:2574-2585` throws `FLOW_FILE_INVALID` for a relative `flow_path`, with the source comment naming this trap: "`argent flow list` prints relative paths, so this is the spelling an agent is most likely to pass back." The deleted `live-authoring.md:254` stated it ("`flow_path`, an absolute path to any flow `.yaml`").

packages/skills/skills/argent-create-flow/references/replay.md
```diff
@@ -0,0 +1,56 @@
...
+- If the first step after an echo is not `launch:`, the flow is a fragment. Read its prerequisite with `flow-read-prerequisite`. Put the app in that state. Then pass `prerequisiteAcknowledged: true`.
+- When more than one device is booted, pass `device` (on Chromium, obey the platform file). Run a flow with `snapshot:` steps on the device that wrote its baselines.
+- Without an agent, run `argent flow run <name> [--device <id>] [--platform ios|android|chromium] [--update-baselines] [--output <dir>] [--json]`. It exits non-zero on failure. `--output` writes the failed baseline, current, and diff images for CI.
+- A recorded `stop-all-simulator-servers` keeps its `devices` list, and only an explicit `device` replaces it. A step with no list is narrowed to the run device. When the run resolved no device, the step stops the devices of all agents on this machine, so record the scope or pass `device`.
+
+## Read the report
+
+- A pass shows `PASS` in the report (`ok: true` in the JSON report), with each step done. The report does not count echo steps. Manual help during the run voids the pass.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: "The report does not count echo steps" is true of the tallies and of the CLI, but not of the report an MCP agent actually reads — and this file is written for that agent.

The tallies do skip echo (`flow-run.ts:1601-1604`) and so does the CLI (`packages/argent-cli/src/flow.ts:270, 294, 694`). The MCP renderer does not: `packages/argent-mcp/src/content.ts:283` prints the header as `` `Running flow "${result.flow}"… (${result.steps.length} steps)` `` over the unfiltered array, and the loop at `:286-297` emits a line for every entry, numbering with `step.index + 1` where `index` is assigned `state.reports.length` (`flow-run.ts:1836`) — a counter echo reports increment too. `stepLabel` even has a dedicated echo arm at `content.ts:256`.

So a flow of 3 actions and 2 echoes renders its header as "(5 steps)", numbers the actions `[2] [4] [5]`, and closes with "3 passed". `replay.md:30` then tells the agent to write down the index of the first failure; that index does not match the number the CLI prints for the same step in the same flow, which is the comparison `qa-flows/SKILL.md:80` sets up by running the two runners against one another.

Worth noting the code comment at `flow-run.ts:1602-1603` asserts the same thing this line does — "counting it would let the summary disagree with the renderers' step numbering (which skips echo too)" — and is accurate only for the CLI.

packages/skills/skills/argent-create-flow/references/polish.md
```diff
@@ -0,0 +1,96 @@
...
+| Recorded form                             | Finished form                                       |
+| ----------------------------------------- | --------------------------------------------------- |
+| focus tap + `tool: keyboard`              | `type:`                                             |
+| text `keyboard` + `key: enter` `keyboard` | `type:` with the default submit, no Enter in `text` |
+| `tool: await-ui-element`                  | `await:` or `assert:`                               |
+| a movement that finds an element          | `scroll-to:`                                        |
+| a coordinate tap or long-press            | a strict selector, after the fallback gate          |
+| `tool: gesture-pinch`                     | `pinch:` with `scale = endDistance / startDistance` |
+| `tool: gesture-rotate`                    | `rotate:` with `by = endAngle - startAngle`         |
+| sibling `tool: flow-execute`              | `run:` (the recorder writes it)                     |
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This row is the only account of how a `run:` step comes to exist, and it describes a path the corpus never tells the agent how to take.

`run:` is not in the hand-addable list (`SKILL.md:27` and `polish.md:32-39` permit only `snapshot:`, `await: { idle: true }`, `long-press:`, and the Chromium `launch:`), so by this skill's own rules the only route is recording the sibling `flow-execute` call. Nothing in the new corpus says how to make that call: it needs `args.name` to name the sibling flow while the session's own `name` identifies the recording in progress, and `flow-add-step.ts:1240-1243` (`RUN_TARGET_COMMAND = "flow-execute"` → `captureRunTarget`) → `:1279-1280` is what turns it into `{ kind: "run", flow: … }`. The deleted file carried exactly that disambiguation — `2740a3ec3:…/references/live-authoring.md:22`, "A recorded `flow-execute` has two names. The top-level `name` identifies the recording. `args.name` identifies the sibling flow captured as `run:`."

The parenthetical also states unconditionally what the code does conditionally: `flow-add-step.ts:991-1101` keeps the call raw with a warning in six situations — no flow name, a remote recording, a non-absolute `project_root`, a case-folded sibling, an unparseable sibling, and a live call that resolved a different file. None of those is described anywhere in the new corpus.

`yaml.md:113-115` documents composition as a supported feature, so the net effect is a directive that is specified but unauthorable: an agent guessing the two `name` fields gets "No active recording for …" (`flow-utils.ts:478`) or silently records against the wrong flow.

packages/skills/skills/argent-create-flow/references/platforms/chromium.md
```diff
@@ -0,0 +1,43 @@
+# Chromium
+
+Read this file one time, before you record or replay against an Electron app or a Chromium browser with CDP.
+
+## Boot
+
+1. Boot the app with `boot-device` and `electronAppPath`. If arguments are necessary for the app, add `electronArgs`.
+2. Take one `screenshot`. Write down its pixel dimensions. The app sets its window dimensions, and no boot argument changes them. Those dimensions key the snapshot baselines, so CI must reproduce them. If the app sizes the window from host or session state, say so in the report.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This is stated for the whole file, but `:3` scopes the file to "an Electron app **or a Chromium browser with CDP**", and it only holds for the first.

`packages/tool-server/src/tools/devices/boot-electron.ts:143-153` (`sanitizeExtraArgs`) strips only `--remote-debugging-port`; every other entry in `electronArgs` reaches the binary (`:314-319`), and `resolveLauncher` (`:74-90`) accepts any `.app` bundle. So `boot-device` with `electronAppPath: "/Applications/Google Chrome.app"` and `electronArgs: ["--window-size=1024,768"]` boots at that size — a boot argument changing exactly what this line says no boot argument changes. The pre-PR text was correctly scoped to the case where it is true (`2740a3ec3:…/references/live-authoring.md:50`: "Electron ignores `--window-size` — the size comes from the app's own `BrowserWindow` options").

It matters because the next sentence makes the claim load-bearing: "Those dimensions key the snapshot baselines, so CI must reproduce them." An agent that believes the window size is not controllable at boot has no reason to pin it, and for the browser half of the file's scope pinning it is both possible and the thing that makes the baselines reproducible.

Related gap in the same section: `:7-9` gives a boot procedure only for launching the app, with nothing for attaching to an already-running CDP browser. The only mention of that case is at `:43`, two lines after `:41`'s unqualified "Do not pass `device`."

packages/skills/skills/argent-create-flow/references/yaml.md
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Still present as of `fe549682f`, and the suggested replacement carries the same ordering, so it survives applying the suggestion.

The sentence groups `~/.argent/secrets.env` with the project's `.argent/secrets.env` and puts both ahead of `.env.local` / `.env`. `secretSources` (`packages/configuration-core/src/secrets.ts:164-190`) appends the global file **after** the shared ones: env, then `<project>/.argent/secrets.env`, then `.env.local`, then `.env`, and `~/.argent/secrets.env` last. `lookupSecret` is first-match-wins (`:194-199`), so a project `.env` beats the user's global file, not the other way round.

Reproduced on this branch, with the tool-server built from it. `~/.argent/secrets.env` holding `DEMO=from-global-file` and the project's `.env` holding `ARGENT_SECRET_DEMO=from-project-dotenv`, then a one-step flow typing `{{secret:DEMO}}` into the Bluesky sign-in field: the field received `from-project-dotenv`. Deleting the project `.env` and replaying the same flow unchanged then yielded `from-global-file`, so the project shared file is genuinely ahead of the global dedicated file.

The impact is on the case the ordering exists for: a developer who keeps a real credential in `~/.argent/secrets.env` and whose repo also has a committed `.env` carrying a placeholder under the same name will have the placeholder typed, and the report redacts it, so the value that was actually used is not visible in the failure.

packages/skills/skills/argent-create-flow/SKILL.md
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Still present as of `fe549682f`. Tracing which half is wrong: the "not supported" half is, and this PR is where it is introduced.

Vega is wired end to end at this head, none of it touched here. `LAUNCH_PLATFORMS = ["ios", "android", "chromium", "vega"]` (`packages/tool-server/src/tools/flows/flow-utils.ts:1859`), `WhenPlatform` derives from that tuple (`:642`), `flow-tree.ts:52` dispatches `vega` to `queryVegaTree`, `flow-vega-tree.ts` is present, `flow-run.ts` carries vega branches at `:525`, `:1145`, `:1205`, `flow-pixels.ts:270-272` captures vega snapshots, and `argent-cli/src/flow.ts:126` advertises `--platform ios | android | chromium | vega`. Only the six touch directives are refused (`flow-actions.ts:786-798`).

The recording half is wrong too, which is the part worth catching, because `flows.mdx:71` is where this PR rewrites it. Before: "The agent records a launch step for iOS, Android and Vega apps." After: "for iOS and Android apps … The agent does not record flows on TV targets (Vega, Apple TV, Android TV) yet." The code disagrees on every clause — `restart-app` supports vega (`restart-app/index.ts:44` `vega: { vvd: true }`, `:89` `vega: vegaImpl`); the `restart-app` → `launch:` rewrite gate has no platform condition at all (`flow-add-step.ts:1251-1256`); `flow-start-recording.ts` has no platform check, and no `RECORDABLE_PLATFORMS`-style allowlist exists anywhere in `packages/tool-server/src/`; the recorder carries dedicated vega guidance at `flow-add-step.ts:202` and `:269`, and its own comment at `:120-121` says a determinate verdict "answers only on ios, android, chromium and vega". Vega recording is unit-tested — `test/flows/flow-record-cross-tree.test.ts:65`, `const VEGA = "amazon-4a27df03c9777152";`, used through `:1136`.

So the contradiction resolves against the skills, not against `flows.mdx`. The refusals are `SKILL.md:8` here ("If the device is a TV, tell the user. Then stop."), `rules/argent.md:128` and `:170`. `packages/docs/docs/reference/flow-yaml.mdx` is untouched by this PR and still documents Vega throughout (`:114`, `:145`, `:210`, `:224`, `:255`, `:271`), so the docs site now disagrees with itself as well.

This also reaches beyond prose: against `2740a3ec3` the same rewrite deletes Vega support from `argent-qa-flows/SKILL.md` — the frontmatter clause "Supports iOS, Android, Chromium, and Vega (Fire TV), where recorded tv-remote steps replace touch directives", the `:10` paragraph mapping the QA contract onto a remote-driven target, item 5's "Vacuous on Vega, which has no coordinate targets", and step 3's "Vega owns no recyclable Argent services". Apple TV and Android TV were already excluded there before this PR, so Vega is the only platform whose support is withdrawn, and it is withdrawn in the skills and user-facing docs while the flow path serving it stays live and tested.

packages/skills/skills/argent-create-flow/references/replay.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Run the flow
+
+- Use `flow-execute` with the absolute `project_root`. Use `name` for a flow in `.argent/flows/`, or `flow_path` for a file elsewhere. One shared filesystem for the agent and the tool server is necessary for `flow_path`, `run:`, and `snapshot:`.
+- If the first step after an echo is not `launch:`, the flow is a fragment. Read its prerequisite with `flow-read-prerequisite`. Put the app in that state. Then pass `prerequisiteAcknowledged: true`.
+- When more than one device is booted, pass `device` (on Chromium, obey the platform file). Run a flow with `snapshot:` steps on the device that wrote its baselines.
+- Without an agent, run `argent flow run <name> [--device <id>] [--platform ios|android|chromium] [--update-baselines] [--output <dir>] [--json]`. It exits non-zero on failure. `--output` writes the failed baseline, current, and diff images for CI.
+- A recorded `stop-all-simulator-servers` keeps its `devices` list, and only an explicit `device` replaces it. A step with no list is narrowed to the run device. When the run resolved no device, the step stops the devices of all agents on this machine, so record the scope or pass `device`.
```
latekvo, line 11: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: This bullet carries the recorded-`devices`-list rule and the no-device-resolved sweep, but drops the cross-host branch the deleted `references/flow-yaml.md` covered. When a flow whose teardown recorded a `devices` list is replayed on a different machine, those ids match nothing there, so `stop-all-simulator-servers` reaps nothing and reports them `unmatched` (`packages/tool-server/src/tools/flows/flow-device.ts:144` and `:298`: recorded ids "on another host come back in `unmatched`"; the tool itself surfaces only "N supplied ids matched no service", `stop-all-simulator-servers.ts:55`). An agent replaying a recorded teardown on a second host then keeps the stale scope and gets a green, no-op cleanup, and nothing in the new corpus tells it that an `unmatched` teardown means the recorded scope no longer applies on that host and must be re-established there — routing the deleted `flow-yaml.md` carried.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Reviewed the reworked flow skills at 885a713c: claims-vs-code across the new prose (swipe magnitudes and edge geometry, secret-source precedence, `flow-execute` params, idle timeout floors, assert grace, Chromium gesture mapping), the absence pass against the three deleted reference files, nearest-twin across the platform files, and every intra-doc anchor link (all resolve). The earlier round's findings are addressed. One inline note below, plus a reply on the still-open doc-test thread.

packages/tool-server/test/flows/flow-skill-docs.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Still present as of 885a713c. The file is deleted at head and nothing replaces the doc-to-code pinning: the PR adds no test (only the deletion + the two `9ac147706`/`885a713c` prose commits), `argent-installer/test/skills-frontmatter.test.ts` validates SKILL.md frontmatter only, and no test under `packages/tool-server/test` reads a reference `.md`. The two `flow-add-step` description/`command` assertions this file held still pass at head and go out with it.

PR #831, opened 2026-08-18
feat(keyboard): clear Android fields atomically

hubgan:
# Clear Android fields atomically, and say so when we can't

Depends on this [PR](https://github.com/software-mansion/argent-private/pull/31).

## Why

`keyboard { clear: true }` on Android was a select-all chord plus a delete, sent
as separate `adb shell input` invocations. Two things are wrong with that:

1. **The field is observed empty in between.** `input` costs ~250ms fixed plus
   ~89ms per character on a Pixel 3a (API 36) - it injects one KeyEvent pair per
   character and waits for the app to handle each. Typing `Friends` is ~850ms of
   wall clock with the app's main thread running seven times inside it. An app
   that reacts asynchronously to its field becoming empty lands that reaction
   mid-typing and wipes what was already sent; the survivor was always a proper
   suffix of the request.
2. **A swallowed chord is invisible.** The chord reports nothing and exits 0, so
   a widget that ignores it (a Flutter `TextField` does) left the following delete
   acting as a plain backspace - a field one character shorter, reported as
   `cleared: true`. A mutated field, not a no-op.

PR #580, opened 2026-07-28
feat(tool-server): `clear` on the keyboard tool

hubgan:
> Stacked on #579 (`fix/android-keyboard-key-order`) - **base that branch, not `main`**.

## Problem

argent can type into a field but never empty one. Every text-entry path appends, so a flow replaying against a non-empty field - a remembered login, a restored draft, a fragment re-run without a `launch`, a retry after a failed step - types on top of the old value. It does not fail at that step; it fails later at an `assert`, pointing at the wrong place.

Counting characters and pressing backspace N times is not a general substitute: what Android uiautomator reports for a password field is not its value - on API 36 it is the masked rendering (a 35-character password dumps as 35 bullets), on other levels it can be empty - so its length is not readable even to the caller.

## Design

`clear` is a parameter, not a new tool. Ordering falls out for free (`{ clear, text }` is replace-a-value in one device round trip), and it inherits registration, the run-sequence allowlist, auto-screenshot and secret handling untouched. `{ udid, clear: true }` alone stays valid.

`clear` runs **first** within a call, on every backend, ahead of whichever of `text` / `key` accompanies it. Those two cannot be combined with each other (#579), so submitting after a replace is a second call.

| Platform | Implementation |
| --- | --- |
| iOS | Cmd+A then backspace over the simulator-server HID transport |
| Android (with `input keycombination`) | `input keycombination 113 29` then `input keyevent 67` |
| Android (without it) | measured delete run - see below |
| Chromium | `rawKeyDown` carrying `commands: ["selectAll","deleteBackward"]` **and** the platform select-all modifier, with the field read back afterwards |
| Vega, TV | rejected before any injection |

### The Chromium form: `commands` for the edit, the modifier as well

Measured on a **macOS** Chrome 150.0.7871.187, controlled input seeded `hello123`:

| variant | chars selected | final value |
| --- | --- | --- |
| `modifiers: 4` (Meta+A) | 0 | `hello12` |
| `modifiers: 2` (Ctrl+A) | 0 | `hello12` |
| `commands: [selectAll]` | 8 | `""` |
| `commands: [selectAll, deleteBackward]` | - | `""` |

[...]

PR #976, opened 2026-08-28
feat(keyboard): clear the focused field

hubgan:
Empties the focused text field in one call:

```
keyboard { udid, clear: true }
```

Exactly one of `text` / `key` / `clear` per call; combinations go through `run-sequence`. Works for native, React Native and Flutter apps on Android, iOS and Chromium.

Supersedes #580 -> #581 -> #831 / #821 (and argent-private #31), which together came to ~6.1k production lines, ~11.6k test lines, 0 docs and ~15 tuning constants. That size was a symptom of the mechanism, not of sloppiness: every layer there existed to defend a **select-all chord that a widget may silently ignore**, and once the primitive can no-op you need read-backs, length measurement, budgets, fallbacks and outcome taxonomies. This replaces the primitive instead.

## Why the chord cannot be the primitive

- Flutter on Android swallows `Ctrl+A` from `input keycombination` deterministically - the trailing `DEL` then removes one character.
- React Native on Android misses it intermittently (#821 measured 15/85 corrupt on API 34).
- On API 31/32 `keycombination` sends no `metaState`, so `TextView.onKeyShortcut` never sees a shortcut at all.
- On iOS the chord holds Left-GUI across awaits, so a concurrent `{ text: "w" }` becomes Cmd+W and a throw in between latches Command.
- On Chromium a modifier-only `Meta+A` / `Ctrl+A` selects nothing on macOS builds.

## What it does instead

**iOS and Android** send **100 backspaces interleaved with 100 forward-deletes** to whatever holds keyboard focus - one bounded burst, no caret move, no selection, no chord, no read of the field, no per-framework branch. Backspace at a line start joins lines and forward-delete at a line end does too, so a multi-line field empties from wherever the tap left the caret; pressing either key on an empty side is a no-op, so over-sending is harmless. Android sends the whole burst as ONE `adb shell input keyevent 67 112 ...` (multi-code `keyevent` has been accepted since API 19, so the API-31/32 metaState bug and the API-30 no-`keycombination` fallback simply do not arise). iOS writes it over simulator-server HID at a 2 ms cadence with a 300 ms settle, holding no modifier at any point.

**Chromium** gets the DOM equivalent - one `Runtime.evaluate` that finds the focused editable (descending through shadow roots), then `execCommand('selectAll'); execCommand('delete')`. It is one round trip, it delivers no keydown at all so a page shortcut cannot cancel it, and the DOM tells us up front whether anything editable has focus - so a clear aimed at nothing 400s with `KEYBOARD_CLEAR_NO_EDITABLE_FOCUS` instead of deleting from whatever the page focuses by default.

It also reads what `execCommand('delete')` answers, which is the one place this could still have reported a failure as a success. Chromium's five date/time input types (`date`, `datetime-local`, `month`, `week`, `time`) pass every editability signal the script can read - not in the denylist, not readonly or disabled, plain `<input>` - and keep their value anyway, because it is structured rather than text. `delete` returns false for exactly those five and true for everything that ends up empty, including a field that was already empty (where `selectAll` returns false). A refused delete therefore raises its own code, `KEYBOARD_CLEAR_UNSUPPORTED_FIELD`, with its own repair - one `key: "backspace"` on the already-focused field empties it - rather than the "tap the field first" advice, which would loop an agent forever here.

**TV targets and Vega** reject `clear` before touching the device.

Documented limit: a field holding more than 100 characters on either side of the caret keeps the remainder; call `clear` again. That is the same contract Maestro ships (50 by default), and it is stated in the parameter description, the tool description, the skill and the docs.

## Deliberately not here

- No select-all chord on any platform.
- No field read-back or verification loop - the result reports what was sent, not what the field holds (a cleared field may have held a secret). Flows that need proof `assert` the field or its consequence, as they already do for `type`.
- No length measurement: `uiautomator dump` races the Android helper for the single UiAutomation connection.
- No Android helper `setText` RPC (argent-private #31) - the only exact/atomic option, but it costs a cross-repo protocol + versionCode bump, a spawn/idle-respawn window, and a fallback for devices where the helper cannot install. It stays parked as the future exact-clear path; nothing here depends on it.
- No `{ clear, text }` in one call - that shape is what forced #580 to pre-validate the whole request and detect focus loss between the halves. Replace-a-value is a `run-sequence`.
- No flow `type: { clear }` - follow-up.

## API

- `clear?: boolean`. Only `true` acts; `false` reads as absent, like `{}`.
- More than one of `text` / `key` / `clear` -> `InvalidToolInputError` before anything is sent, keeping the `KEYBOARD_TEXT_AND_KEY_COMBINED` code and naming which two the request carried.
- Result `{ typed: "", keys: <key events sent>, cleared: true }` - no field value, no length.
- Interaction messages: "Clearing the field" / "Cleared the field".
- `delayMs` does not apply to the burst (it has its own cadence).
- Two new failure codes, both Chromium-only and both with distinct repairs: `KEYBOARD_CLEAR_NO_EDITABLE_FOCUS` (tap the field) and `KEYBOARD_CLEAR_UNSUPPORTED_FIELD` (the focused field kept its value).

## Verified on device

Driven through a branch build of the tool-server over HTTP.

| target | app | result |
| --- | --- | --- |
| Pixel 3a API 36 | native `EditText` | cleared from the end and with the caret mid-text (26 chars, both sides); password field cleared; the unfocused field untouched |
| Pixel 3a API 36 | Flutter `clearlab` | single-line, multi-line (caret mid-block, all three lines gone) and password all cleared |
| Pixel 3a API 36 | RN Bluesky search | 5/5 type -> clear clean; `run-sequence` clear -> text -> enter left exactly `argent`, not spliced |
| Pixel 3a API 36 | native `EditText` | 191-char field -> one clear leaves 91 -> second clear empties it (the documented 100-per-side bound, exactly) |
| iPhone 16 Pro 18.3 | Settings search (UIKit) | screenshot: "Wifi settings" -> placeholder back |
| iPhone 16 Pro 18.3 | Expo dev-launcher field (RN) | screenshot: value gone, Connect button disabled |
| iPhone 17 Pro 26.4 | Settings search (UIKit) | screenshot: "Bluetooth options" -> placeholder back |
| iPhone SE 18.3 | Flutter `clearlab` | single, multi-line and password cleared |
| Chrome 151 | fixture page | plain / textarea / contenteditable / password / number / controlled input all cleared, each firing one `input` with `inputType: deleteContentBackward` (React's value tracker); readonly and disabled refused and untouched; nothing-focused refused; an already-empty field still reports success; an `<input>` inside a shadow root cleared |
| Chrome 151 | date input | refused with `KEYBOARD_CLEAR_UNSUPPORTED_FIELD`, value left at `2020-01-02`; the message's own repair (`key: "backspace"`) then emptied it, driven through the tool |
| Chrome 151 | bsky.app (RN-web) | screenshot: controlled field cleared; `run-sequence` clear -> text left exactly `argent.example` |
| Apple TV tvOS 26.5 | - | `clear` rejected with `TOOL_CAPABILITY_UNSUPPORTED_OPERATION`; `text` still works |

Not re-driven on hardware here: Android TV and Vega (no AVD / VVD available on this host) - Android TV goes through the same `typeTv` rejection that was verified live on Apple TV, and both are unit-tested; and the API 24 / API 30 emulators, which the measurement pass covered (a multi-code `input keyevent` is accepted on API 24, and a 60-80 key burst clears a native `EditText` and a Flutter single- and multi-line `TextField` on API 30 - the level where `keycombination` Ctrl+A is swallowed). The burst uses no `keycombination` at all, which `keyboard-android.test.ts` pins on the exact argv, so the API-31/32 missing-metaState bug and the API-30 fallback cannot arise.

Suites: `npm test -w @argent/tool-server` (4883 passed, 1 skipped), `-w @argent/registry` (120), `-w @argent/mcp` (80), `npm run lint`, `npm run build`, `npm run typecheck:tests -w @argent/tool-server`, `npm run typecheck:scripts`, `npm run test:scripts` (92), `npm run knip`, `node scripts/grade-skills.mjs` (16/16 at 10.0), `npx docusaurus build`, `prettier --write .`.

## Docs and skills

`reference/tools.mdx` keyboard row; `features/interacting-with-apps.mdx` (a paragraph under "Gestures and text", an example instruction, two limits). `argent-device-interact` gains a "Clearing a field" section, a capability row, and loses the triple-tap-to-select-all advice; `argent-tv-interact`, `argent-test-ui-flow` (login example clears before typing), `argent-create-flow` live-authoring + flow-yaml updated.

## E2E harness

`scripts/e2e-full`: the Chromium tier gets before/after `describe` evidence around a clear on a dedicated fixture field, the nothing-focused refusal, the exclusivity rejection and a clear -> text `run-sequence`; the Android tier gets the contract cases (a headless tier has no focused text field to observe, so the per-framework evidence lives in the device matrix above).

## Follow-ups

- Flow `type: { into, text, clear: true }` - dispatch `{ clear: true }` before `{ text }` in `runType`; the existing `waitForFocus` gate is enough. ~40 lines + docs.
- `clear: <n>` if a real long-field case shows up.
- Android TV and Vega support if requested.
- #580, #581, #831, #821 and argent-private #31 can be closed as superseded once this lands.


latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Reviewed the keyboard `clear` feature and the new per-device serialization layer across claims-vs-code, nearest-twin / symmetry, non-happy paths (abort / partial burst / transport-death), inputs, reachability, and call lifetime, plus a mutation-minded look at the new constants and an absence sweep of the `awaitDeviceHold` focus-mover contract. Individual notes are inline.

packages/tool-server/src/utils/device-serial.ts
```diff
@@ -0,0 +1,240 @@
...
+ * is open. This map is empty the rest of the time, and `awaitDeviceHold` then
+ * costs one lookup.
+ */
+const activeHolds = new Map<string, Promise<unknown>>();
+
+/**
+ * Wait out another session's hold on this device, if there is one.
+ *
+ * Called by the tools that move keyboard focus (the gestures, `button`,
+ * `tv-remote`, `launch-app`, `restart-app`, `open-url`). A call made INSIDE the
```
latekvo, line 217: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: This caller list (and the actual wiring — `awaitDeviceHold` is imported into exactly these twelve tools and no others) omits two tools that also move keyboard focus:

- `chromium-tabs` re-points the shared per-device CDP that a `keyboard { clear }` evaluates against. Its own description says selecting a tab makes it "the active one … every other tool then acts on it," and `chromium-server` re-points `cdp` in place on a tab switch "so every subsystem that captured `cdp` follows the new page automatically" — which includes `server.evaluate`, the path `clearChromium` runs on. `chromium-tabs.execute` calls no `awaitDeviceHold`, so a second session's `chromium-tabs { action: "select", tab: "t2" }` landing inside a first session's `run-sequence` [tap t1's field, `keyboard { clear }`] hold switches the active tab before the clear evaluates: the clear then runs against t2 and, when t2 has a focused editable, empties it and returns `cleared: true` / `clearVerified: true` while the field the sequence tapped keeps its value. Its nearest twin `open-url` (a strictly smaller active-tab operation) does wait the hold out.
- `reinstall-app` uninstalls the running app first (dropping keyboard focus) — the same terminate/focus-move class as `restart-app` and `launch-app`, both listed here — but calls no `awaitDeviceHold`, so it can tear the focused app down inside another session's keyboard hold and leave that session's clear running against the home screen.

Traced at 63bc63ef: `chromium-tabs/index.ts` execute → `chromium-server/index.ts` (`cdp` re-point + `evaluate`), `grep -rl awaitDeviceHold` (twelve tools, neither of these), against `open-url` / `restart-app`.

packages/tool-server/src/tools/keyboard/platforms/chromium.ts
```diff
@@ -7,12 +7,782 @@ import type { KeyboardParams, KeyboardResult } from "../types";
...
+    //
+    //   * \`execCommand\` acts on the document's SELECTION, not on the focused
+    //     element. Measured on Chrome 151 with the standard rich-text toolbar
+    //     shape (\`mousedown\` + \`preventDefault\` + \`focus()\`, which keeps the
+    //     editor's selection alive): focus on the button, selection still in a
+    //     neighbouring \`contenteditable\` — \`selectAll\` + \`delete\` emptied THAT
+    //     editor and answered true.
+    //   * \`delete\` answers true whether or not it removed anything, so its
+    //     return value is not evidence, and an opaque host cannot be read back
+    //     to get any. \`cleared\` on this backend means the field was SEEN empty;
```
latekvo, line 156: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: This says that on the Chromium backend `cleared` "means the field was SEEN empty," but in the shipped result that is `clearVerified`'s meaning, not `cleared`'s. `types.ts` documents Chromium `cleared` as "the delete was ACCEPTED," and `clearChromium`'s unverified return emits `{ cleared: true }` with no `clearVerified` in the three read-back-declined cases (field replaced, `window` sealed, target went read-only → `remaining: null`) — `cleared: true` while nothing saw the field empty. The parallel wording further down ("`cleared` alone means … 'read back empty' here") reads the same way and is contradicted a few lines later by "the delete was still accepted, so `cleared` stands — but nothing saw the field empty." Compared against the `cleared` / `clearVerified` docs in `types.ts` and the verified / unverified returns in this file.

packages/tool-server/src/tools/keyboard/platforms/ios.ts
```diff
@@ -1,27 +1,163 @@
...
+        : "the clear then works on either kind."),
+    {
+      error_code: FAILURE_CODES.KEYBOARD_TARGET_KIND_UNKNOWN,
+      failure_stage: "keyboard_ios_runtime_kind",
+      failure_area: "tool_server",
+      // Not "timeout": a listing that came back without this UDID answers the
+      // same undefined, and no probe timed out there. What both causes share is
+      // that the device is not in the listing.
+      error_kind: "not_found",
+      failure_command: "xcrun_simctl",
```
latekvo, line 75: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: `failure_command: "xcrun_simctl"` is stamped unconditionally, but `refuseUnknownKind` is also called from `makeIosRemoteImpl` (with `remote = true`), whose probe is `getRemoteSimulatorRuntimeKind` → `simctlListDevices` → `execFileAsync("sim-remote", …)` — the `sim-remote` binary, not `xcrun` / CoreSimulator. So `keyboard { clear }` on an `ios-remote` UDID whose sim-remote listing omits it (or misses its budget) raises a `KEYBOARD_TARGET_KIND_UNKNOWN` failure whose telemetry attributes it to an `xcrun` command that never ran; the human-readable message is correct (it interpolates the `probe` arg) but the structured field ignores the `remote` parameter. `FAILURE_COMMANDS` has no sim-remote entry but does carry `"unknown"`. Checked against `sim-remote.ts` and `registry/src/errors.ts`.

packages/docs/docs/reference/tools.mdx
```diff
@@ -50,12 +50,42 @@ Run `argent tools` to list them from the terminal, `argent tools describe <name>
...
+| Vega                               | 200 backspaces, because the platform has no forward delete | 200 characters before the cursor          | 200    | never                                    |
+| Chromium                           | Selects the focused editable and deletes the selection     | none                                      | 0      | when the read-back found the field empty |
+
+Send `clear` again for each further 100 characters the field holds. Argent reports what it sent, so read the field to get proof.
+
+Chromium is the only target that checks focus, and it fails the clear when no editable element holds focus. The other five send the keys wherever focus is, and they report success.
+
+A remote Apple TV (an `ios-remote` tvOS id) refuses `clear` and `key`. The tvOS daemons drive a simulator in the Argent host's own CoreSimulator set, and they do not reach a device behind sim-remote. Use a local Apple TV simulator instead.
+
+A failed clear carries one of five codes:
```
latekvo, line 77: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: "A failed clear carries one of five codes" and the table lists five, but a clear can fail with a sixth. `keyboard`'s `execute` routes the whole clear dispatch through `serializedPerDevice`, whose `guarded()` throws `KEYBOARD_DEVICE_BUSY` when the queue wait exceeds `DEVICE_QUEUE_MAX_WAIT_MS` (30s — reachable, since the same file budgets an Android clear up to 90s) and when the request aborts while queued. With one shared tool-server and two sessions (the documented default), session B's `keyboard { clear }` behind session A's >30s hold fails with `KEYBOARD_DEVICE_BUSY` — a code this table, and the docs generally, never mention. Traced `keyboard/index.ts` → `device-serial.ts` (`guarded` / `deviceBusyError`); grep of `packages/docs` shows the code is absent.

Issue #932, opened 2026-08-23
iOS sim HID injection silently no-ops on macOS 27 beta 26A5416b (CoreDevice dtuhidd owns sim HID); #547 killall workaround ineffective


mrmeg:
## Summary

All iOS-simulator HID injection (`gesture-tap`, `gesture-swipe`, hardware `button` presses including Home) silently no-ops on the current macOS 27 beta: the tools return success (`{ tapped: true }` / `{ pressed: "home" }`) but no event ever reaches the simulator. Screenshot capture, the AX `describe` tree, `launch-app`, `open-url`, and the Metro debugger all keep working through the same simulator-server session.

Looks like the systemic version of #547 — but unlike #547, the `killall com.apple.CoreSimulator.CoreSimulatorService` workaround does **not** recover it (tested with both SIGTERM and SIGKILL, fresh daemon pid confirmed, sim re-booted through argent afterwards).

## Environment

- Argent 0.21.0 and 0.22.0 (reproduced identically on both; 0.22.0 verified via `argent --version` after `update-argent`)
- macOS 27.0 beta, build **26A5416b** (Darwin 27.0.0)
- Xcode 27.0 beta (27A5228h), installed Jul 17 — unchanged across the regression window
- iOS 27.0 simulators (iPhone 17 Pro Max, iPhone 17 Pro), default device set
- Node/npm global install at `/opt/homebrew/lib/node_modules/@swmansion/argent`

## Timeline — this is an OS-update regression

- Argent taps worked on these same iOS 27 sims + this same Xcode 27 beta on **Aug 16**.
- The boot volume was re-blessed **Aug 18 17:16** (macOS beta update to 26A5416b).
- On **Aug 21**, taps silently no-op on every simulator, fresh boots included.

## Evidence the sim HID stack moved to CoreDevice in this build

- This Xcode beta has **no `Simulator.app`** — the GUI is `Contents/Applications/DeviceHub.app`.
- Inside a booted sim, HID is served by CoreDevice daemons: `launchctl print system` (via `simctl spawn`) shows **`com.apple.coredevice.dtuhidd`** running and `com.apple.coredevice.dthidd` registered.
- `xcrun devicectl list devices` now lists **simulators** (`Reality: simulated`) alongside physical devices.
[...]

PR #985, opened 2026-08-31
fix(tool-server): await the simulator-server command ack so a lost tap fails

latekvo:
## The defect

`sendCommand` returns `void`. It writes the JSON command onto the WebSocket — or queues it behind `once("open")` — and returns without ever reading the reply:

```ts
export function sendCommand(api, cmd): void {
  const ws = getOrCreateWs(api);
  const payload = JSON.stringify({ id: String(++cmdId), ...cmd });
  if (ws.readyState === WebSocket.OPEN) ws.send(payload);
  else ws.once("open", () => ws.send(payload));   // queued, never awaited
}
```

simulator-server *does* acknowledge every command; argent just never looked. Probing the shipped binary's `/ws` endpoint:

| sent | reply |
|---|---|
| `{"id":"1","cmd":"touch","type":"Down",…}` | `{"id":"1","status":"ok"}` |
| `{"id":"7","cmd":"button","button":"notARealButton"}` | `{"status":"error","message":"parse error: unknown variant `notARealButton`…"}` |

Because the return type was `void` there was no ack to check and no error to surface, so `gesture-tap` ended in an unconditional `return { tapped: true, timestampMs }`.

**Repro:** shut a simulator down underneath an open session. The server sends no reply at all and `gesture-tap` still reports `{ tapped: true }`. Every transport failure — that one, a dead socket, a wedged device — looks identical to success from the agent's side.

This is argent's own bug, and it is what turns an Apple regression into a silent one: awaiting the reply would have made #932 an actionable failure on day one instead of a silent no-op.

## The fix

`sendCommand` now returns a promise that resolves on `{"status":"ok"}` and rejects when:

| condition | `error_code` |
|---|---|
| server answers `status: "error"` | `SIMULATOR_COMMAND_REJECTED` |
| socket fails or closes with the command in flight | `SIMULATOR_COMMAND_TRANSPORT_FAILED` |
| reachable server never answers (5s) | `SIMULATOR_COMMAND_ACK_TIMEOUT` |

Every WebSocket call site awaits it: `gesture-tap`, `gesture-swipe`, `gesture-custom`, `gesture-pinch`, `gesture-rotate`, `button`, `rotate`.

Two details of the wire protocol shaped the implementation:

- **A success reply echoes the command id; an error reply carries no id at all.** So a rejection can only be matched positionally, against the oldest outstanding command. Since every send is awaited, that is normally the only one in flight.
- **The abort paths in `gesture-swipe` / `gesture-rotate` lift the finger before unwinding, and that lift is now fallible.** It is therefore best-effort: a refused lift rides along as the `AbortError`'s `cause` rather than replacing it — matching the existing pattern in `gesture-drag`, because `format-error.ts` and `cdp-session.ts` key on `err.name === "AbortError"`.

## Verified on a real simulator

Before, after shutting the device down mid-session: `{"tapped": true}`. After:

```
result: {"tapped":true,"timestampMs":…}          ← healthy tap, unchanged

--- simulator shut down underneath the open session ---
tap FAILED as it should:
   code: SIMULATOR_COMMAND_TRANSPORT_FAILED
   msg : simulator-server did not accept the 'touch' command: connect ECONNREFUSED
         127.0.0.1:62818. The command was NOT delivered to the device. …
```

The failure signal is fully populated for telemetry (`error_kind: "network"`, `network_failure: "connection_reset"`, `failure_command: "simulator_server"`).

## Cost

An ack is a localhost round-trip behind an already-open socket: **0.06ms p50 / 0.17ms max**, measured over 200 sends against a booted iOS sim. Awaiting one per event adds a flat **~30ms to a swipe of any duration** — against the 36–142ms of `setTimeout` skew the 16ms frame loop already carries:

| `durationMs` | wall before | wall after | added |
|---|---|---|---|
| 200 | 236ms | 266ms | +30ms |
| 500 | 559ms | 591ms | +31ms |
| 1000 | 1142ms | 1178ms | +35ms |

Flagging it explicitly since the swipe cadence is deliberately tuned. If reviewers would rather keep the cadence byte-identical, the alternative is to pipeline within a gesture and await all acks at the end; that costs more code and defers failure to the end of the gesture, so I kept the simpler per-send await.

## Deliberately out of scope

- **The MoQ (`ios-remote`) transport** is unchanged — it has no equivalent ack.
- **The stdin `pressKey` channel**, used by `keyboard` and paste's ⌘V, is untouched: simulator-server never replies on stdin, so it has the same blind spot with no local remedy. The WebSocket protocol *does* expose an acked `key` command (`{"cmd":"key","direction":"Down","code":4}` → `{"status":"ok"}`), so moving those onto it would close the gap — a separate, riskier change to the text-entry path that deserves its own PR.

## Tests

`packages/tool-server/test/simulator-command-ack.test.ts` — 8 tests against a fake `/ws` server that mirrors the real wire format, **all 8 fail on `main`**: ack resolves, `status: error` rejects, socket death rejects in flight, silent server times out, no phantom success, `gesture-tap` throws instead of claiming a rejected tap, and the aborted swipe still reports `AbortError`.

Full suite: 369 files / 4840 tests pass, `npm run lint` clean, `tsc --build` clean.

https://claude.ai/code/session_011bEvNDV7o3CpKdFmzjVBYF

## Summary by CodeRabbit

* **New Features**
  * Simulator commands now confirm successful delivery before completing.
  * Added clear errors for rejected commands, connection failures, and acknowledgement timeouts.
* **Bug Fixes**
  * Button, rotation, tap, swipe, pinch, and custom gestures now report failures instead of silently succeeding.
  * In-progress gestures handle connection interruptions and cleanup failures more reliably.
* **Tests**
  * Added coverage for command acknowledgements, rejections, timeouts, transport failures, and gesture error handling.


latekvo:
Thanks — took one of the three, and it was a real bug in this branch.

**Fixed (82f7f9e): late ack credited to the wrong command.** The correlation had a hole that reintroduced exactly what this PR removes. A reply naming an id no longer in `pending` — the late ack for a command that already timed out — fell through to the positional fallback and settled whatever command was in flight *now*:

```
A (id=1) times out at 5s        → A rejects, pending.delete("1")
B (id=2) sent                   → pending = {2}
server's late {"id":"1","ok"}   → id 1 not pending → fell back to oldest → settled B
```

B reported the success that was really A's. Now an id'd reply is matched *only* by id and dropped when argent isn't waiting on it; positional matching is reserved for the id-less replies it was meant for (errors carry no id). A timeout additionally drops the connection, since the answer to the timed-out command is still owed and once it lands nothing can tell it apart from a later one. Regression test added — it resolves the phantom success without the fix.

**Skipped: cleanup Ups in `button` / `gesture-custom` / `gesture-pinch`.** Not a regression from this change. When the Up is refused, it did not reach the device before this PR either — the send was fire-and-forget, so the finger/button was left held exactly the same way; the only difference now is that argent says so. And the usual reason an Up fails is a transport that is already gone, which a cleanup Up would not survive either. Adding lift-on-failure to three tools that have no cancellation handling today is a device-state-hygiene change worth doing on its own terms, not folded into the ack fix.

**Skipped: cancellation-aware ack waits.** The abort is checked once per frame, and an ack is a 0.06ms p50 / 0.17ms max localhost round-trip (measured over 200 sends against a booted sim), so on a healthy server an abort is already noticed within its own frame. The 5s exposure only appears when the server is wedged — the exact condition the timeout exists to report — and closing it would mean threading an `AbortSignal` through `sendCommand` into every call site, which is a wider change than the defect warrants.

filip131311 (approved):
perfectr pr

PR #740, opened 2026-08-05
fix(screenshot-diff): describe the live capture the tool actually takes

latekvo:
## What the descriptions promised

```
description        "…one saved PNG plus one live full-resolution capture from a device."
captureBaseline    "Capture the baseline screenshot live at full resolution before diffing."
captureCurrent     "Capture the current screenshot live at full resolution before diffing."
```

`captureLiveInput` has fallen back since #307, and says so in a comment right above the code.
Measured on a Pixel_9a (1080x2424) through the real production path:

```
$ argent run screenshot-diff --udid emulator-5554 --baselinePath <1069x2400>.png --captureCurrent true
- size_normalized: baseline=1,069x2,400 current=324x727 compared_at=324x727
```

324x727 is the fallback scale, not full resolution. The behaviour is deliberate; the descriptions
were the agent-facing contract and they were wrong.

## "the server's default scale" was wrong twice

The first pass at this replaced "full resolution" with "the server's default scale". Driving
`executeScreenshotDiffTool` against a live simulator-server, that is also false:

```
ARGENT_SCREENSHOT_SCALE unset  ->  current=324x727
ARGENT_SCREENSHOT_SCALE=0.6    ->  current=648x1,454
ARGENT_SCREENSHOT_SCALE=1.0    ->  THREW  wrong data size, expected 7853760 got 17627328
```

The retry resolves `getScreenshotScale()`, which returns the *configured* scale and falls back to
`DEFAULT_SCREENSHOT_SCALE` only when unset or unparseable — so it is not a default. And when that config is `1.0`,
`httpScreenshot` omits an in-band 1.0, making the retry byte-identical to the request that just
failed: both attempts throw. The descriptions name the env var and the literal default, and both
live flags carry the dead end.

## What a normalized comparison costs

The resize block warned in one direction only — differences below "may include resampling
artifacts". The downscale erases as readily as it invents. Same regression, two capture outcomes:

```
=== full-resolution live capture ===        === fallback live capture ===
- status: changed                           - status: resized_no_change
- pixel_mismatch: 0.04%                     - pixel_mismatch: 0%
- text_analysis: status=ok                  - text_analysis: status=skipped
differentPixels = 1179                      differentPixels = 0
```

Measured while the fallback was 0.3; #878 has since lowered it to 0.25. Only the scale moved -
the two outcomes are what any downscale produces.

A 1px hairline shifting shade vanishes, and the OCR pass goes with it — text analysis is gated on a
[...]

PR #994, opened 2026-08-31
feat(screenshot-diff): stage a live baseline instead of hand-carrying its path

latekvo:
Part of the "guidance mandates a call whose inputs already live server-side" audit (see #984, #990, #992, #993).

> **Merged with `main`.** Since this branch was cut, `main` gave `screenshot-diff` a physical-iPhone capture backend (`iosDeviceRunner`) selected by a per-device dispatch. Both the staging branch and the two-sided comparison now route through one shared `captureLiveSide()` helper, so a `captureBaseline` staging call on a physical iPhone reaches the runner `services()` resolves for it instead of failing for a missing simulator-server. Covered by three new tests in `screenshot-diff-staged-baseline.test.ts`.

## The defect

`argent-screenshot-diff`'s Deterministic Flow, step 2:

> Capture a baseline PNG with `screenshot` using `scale: 1.0` and `includeImageInContext: false`; keep the returned `path`.

Two magic parameters the agent has to remember on a separate tool call, and a path it then hand-carries across however many turns the change under test takes. Both parameters are load-bearing and both fail quietly:

- forget `scale: 1.0` and the baseline is downscaled
- forget `includeImageInContext: false` and a full-resolution PNG lands in the model's context

`argent-test-ui-flow` repeats the same instruction in three places.

Meanwhile `screenshot-diff` already owns a live capture path - that is what `captureBaseline` / `captureCurrent` do. The one thing it could not do was capture a baseline **now** and diff against it **later**.

## The change

`captureBaseline: true` with **no current side** (no `currentPath`, no `captureCurrent`) captures the baseline, stages it for that `udid`, and returns without comparing. A later call with a current side and **no baseline side** resolves it.

```
screenshot-diff { captureBaseline: true, udid }                      # stage
...change the code, reload, navigate...
screenshot-diff { captureCurrent: true, udid, outputDir }            # compare
```

Two `screenshot-diff` calls. No `screenshot` call, no magic parameters, no path bookkeeping.

Every previously valid shape still works unchanged: `baselinePath` + `captureCurrent`, `baselinePath` + `currentPath`, and `captureBaseline` + `currentPath` (a current side makes it an immediate comparison, not a staging call).

## Guarding the stale baseline

A silently stale baseline is the obvious hazard, so it is made impossible to miss:

- **Every comparison against a staged baseline names it**, with `udid`, `captured_at` and `age_seconds`, and that block **leads** the summary rather than trailing it - the staged baseline is the one input the call did not name, so its age is read before the figures measured against it.
- **Keyed by device.** One device's staged baseline can never answer for another; a call for an unstaged `udid` fails even when another device has one.
- **A missing or reaped baseline is a hard failure**, never a silent live re-capture. A baseline captured after the change is not a baseline. The staged PNG sits in the staging call's `outputDir`, which defaults to a temp directory the OS can reap, so the file is re-checked at use.

## Why in-process, not on disk

Module-level `Map` keyed by device, the idiom `app-session.ts` and `screen-recording-reminder.ts` already use. Two reasons, both in the header comment: the staged PNG lives in a temp directory the OS is free to reap, so a durable index would have to survive a dangling path anyway; and an entry that cannot outlive the tool-server can never be served as the known-good screen after a restart that rebuilt the app underneath it. Bounded lifetime is the safety property here, not a limitation.

## The validation rewrite

`validateInputSources` (reject-only) becomes `classifyInputSources` (returns `"stage-baseline"` or `"diff"`, rejecting everything that is neither). The four conflict checks still run **first**, and the comment says why: a call that names a baseline BOTH ways still has no current side, so testing for staging ahead of them would accept it and silently ignore the `baselinePath` it was given. That ordering is mutation-tested.

`services()` was already conditional - it deliberately does not resolve the simulator-server for pure static-PNG diffs, because that fails on tvOS simulators. A staging call sets `captureBaseline`, so it lands in the resolving branch; pinned by a test.

## Note on #740

#740 is open and shows, with measurements, that `captureLiveInput` has fallen back since #307 - 324x727 on a 1080x2424 Pixel 9a - so the "full resolution" in these descriptions is false. Two consequences here:

1. The sentences this change rewrites carried that phrase forward. Second commit drops it rather than re-asserting it. The descriptions this change does **not** touch are #740's to correct; they are deliberately left alone rather than duplicating that PR.
2. It resolves the fidelity question this change raises. Staging captures through `captureLiveInput` rather than through `screenshot { scale: 1.0 }`, so the staged baseline is not full resolution - but #740's own output shows the old flow already downscaled the full-resolution baseline to meet the live current side (`baseline=1,069x2,400 current=324x727 compared_at=324x727`). The comparison happened at the live-capture resolution either way. Staging both sides through one path makes them match rather than mismatch.

## Tests

`packages/tool-server/test/screenshot-diff-staged-baseline.test.ts` - 11 tests:

- staging returns a summary and no diff images
- a later live capture compares against it, and the summary carries the capture time
- a saved `currentPath` compares against it too, without resolving the simulator server
- a staging call **does** resolve the simulator server
- staged per device: a second device's call is refused while the first stays staged
- nothing staged -> refused, naming how to stage
- staged file reaped -> refused, naming when it was staged
- re-staging replaces the entry
- one staged baseline serves repeated comparisons
- `captureBaseline` + `currentPath` still diffs immediately instead of staging
- every previously invalid combination is still refused

Mutation-checked - each leaves the file red:

| mutation | failures |
|---|---|
| never classify a call as staging | 8 |
| drop the reaped-file check | 1 |
| ignore the staged baseline and use `baselinePath` | 4 |
| classify staging *before* the conflict checks | 1 |
| drop the provenance block from the diff summary | 3 |

## Docs and skills

- `argent-screenshot-diff` - Capture Rules, Parameters, Deterministic Flow and both JSON examples.
- `argent-test-ui-flow` - the three places it repeated the `scale: 1.0` / `includeImageInContext: false` instruction.
- `docs/features/visual-regression.mdx` - how a comparison works, and what the agent no longer holds.
- `docs/reference/tools.mdx` - the `screenshot-diff` row.
- `screenshot`'s `includeImageInContext` description - it pointed at this workflow as the reason to set it false.

## Verification

- `npm run build`, `npm run lint`, `npm run knip`, `npm run format`, `typecheck:tests -w @argent/tool-server`
- `npx docusaurus build` in `packages/docs`
- full `packages/tool-server` suite in 6 shards: 4852 passed, 1 skipped, **zero failures**.


latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Reviewed after merging `main` (which had added the physical-iPhone capture backend since this branch was cut). Resolved the one conflict and fixed the integration it exposed: the `captureBaseline` staging branch now routes through the shared `captureLiveSide()` dispatch, so a staging call on a physical iPhone reaches the `iosDeviceRunner` that `services()` resolves for it, instead of throwing for a missing simulator-server. Added three tests that fail against the pre-merge simulator-only staging path. Full CI is green.

A multi-dimension review (control flow, state lifecycle, claims-vs-code, tests, inputs/nearest-twin, absence pass) found **no High or Medium** issue. `classifyInputSources` routes all input shapes correctly, `services()` and `captureLiveSide` derive the backend from the same pure functions so they cannot disagree, and the five mutations the description claims are each pinned by a discriminating test.

Residual **Low**, left as-is:

- **Cross-agent stale baseline (feature-level).** The staged map is keyed by `udid` with no agent identity — the same idiom as `app-session.ts` / `screen-recording-reminder.ts`. If a device is freed and reallocated, an out-of-flow `captureCurrent` (one that skips staging its own baseline) would diff against the previous holder's baseline. Mitigated: every comparison leads with `age_seconds`, so a stale baseline is conspicuous in the result.
- **Pre-existing, not introduced here (noted for the record):** `captureBaseline`/`captureCurrent` use `z.coerce.boolean()`, so a stringified `"false"` coerces to `true` (every other boolean param uses plain `z.boolean()`); an otherwise-invalid call still creates an empty temp dir before validation throws; physical-iOS live capture does not forward the abort signal. None sit on the staging feature's critical path.

Kept as a draft: no High/Medium, but that residual Low means it is not flip-to-ready clean.

PR #1015, opened 2026-09-01
feat(screenshot): add `out` so an agent can keep the PNG

latekvo:
## Human written:

Argent's CLI exposed `--out` flag on `screenshot` while the MCP command had no such option.
Thus all MCP-originated screenshots used to end up in tmp.

This PR fixes that by adding `out` param to MCP.

---

Slop:

<details>

## The symptom

`screenshot` returns an artifact handle and nothing else, so the client materializes the PNG into a session-scoped temp directory that goes away with the session. An agent that captures a full-resolution baseline for `screenshot-diff` (`scale: 1.0`, `includeImageInContext: false`) has no way to keep it.

The CLI already had this: `argent run screenshot --udid <UDID> --out ./screen.png`. MCP had no equivalent, and MCP is the path agents actually use.

## What now happens

`screenshot` takes an optional `out`. With it, the PNG is written where the caller asked and the result reads `Saved: <that path>`. Without it **nothing changes** - the capture stays in the temp directory, as it does today.

```json
{ "udid": "<UDID>", "scale": 1.0, "includeImageInContext": false, "out": "/tmp/baseline.png" }
```

Relative paths resolve against the agent's working directory, `~` is expanded (no shell stands between the model and the arg), missing parents are created, and an existing file is overwritten - the caller named the path.

## Where the write happens, and why

Client-side, in `argent-mcp`'s content builder, next to the temp copy it replaces. `out` names the **agent's** filesystem, which under `argent link` is a different machine from the tool-server, and it is the filesystem `screenshot-diff` resolves `baselinePath` against (`fileInputs`: "The saved PNGs live on the AGENT's machine"). A server-side write would put the baseline on the wrong host and, because `outputHint: "image"` renders only the image plus a `Saved:` line, the agent would never see the path it asked for.

So `out` is input-only on the tool, exactly like the neighbouring `includeImageInContext` - declared in the schema because that is what the agent reads, consumed by the adapter.

## The volume question

The MCP layer appends an automatic screenshot after most interaction tools. It calls `toMcpContent(result, "image", ctx)` with **no args**, so it structurally cannot carry an `out` and cannot write into the working tree. That is pinned by the existing test that asserts the no-args render reports the temp path.

## Non-happy path

Every failure mode of the write is reported and none of them costs the capture: on `EACCES`, `ENOTDIR`, `EISDIR` or a full disk the result is

```
Saved: <temp path>
Could not save to <out>: <errno message>
```

with the image block still attached. The capture is the expensive part and it already succeeded.

## CLI help

`argent run screenshot --help` printed `--out` twice: once in the per-tool **Flags** block (rendered from the schema) and once under **Global flags**, where `splitOptions` consumes `--out` before the schema parser ever runs. The schema `out` is therefore unreachable through its own flag, so the Flags block now drops any property a global flag shadows (`out`, `json`) and `--out` prints once.

## Verification

Gates from the worktree root, all clean: `npx tsc --build`, `npx prettier --check .`, `npx eslint . --max-warnings 0`, `typecheck:tests` in tool-server / mcp / tools-client, and `npx docusaurus build` in `packages/docs`. Suites: **tool-server 4845** (6 shards, 1 skipped), **mcp 92**, **tools-client 196**, **cli 527**, **registry 120** - all passing. No dependency change, so no lockfile regen.

**E2E through the real MCP stdio server** - worktree `startMcpServer` over JSON-RPC, `ARGENT_TOOLS_URL` at a worktree tool-server on `:3197`, live iOS 18.6 simulator `770F654F-…`:

| case | result |
| --- | --- |
| `tools/list` | `screenshot.inputSchema.properties.out` present |
| absolute `out`, nested parent missing | file created, 535 991 B, magic `89504e47`, real Settings capture; `Saved: <out>` |
| relative `out` | resolved against the MCP process cwd, file present |
| no `out` | `Saved: /var/…/simserver-…/media/936187000-….png` - unchanged |
| `out` under an existing file | image block kept, temp path kept, `Could not save to …: EEXIST …` |

**Mutation checks** - each revert turns tests red, so nothing here is pinned by accident:

| reverted | result |
| --- | --- |
| `savedText` ignores `out` | 5 failed |
| tilde expansion | 1 failed (and left a literal `~/` directory behind, which is the point) |
| `mkdir` no longer recursive | 2 failed |
| failure path returns the `out` path anyway | 1 failed |
| `out` dropped from the zod schema | 1 failed |
| CLI Flags block lists the reserved global flag | 1 failed |

## Docs and skills

- `argent-screenshot-diff`, `argent-test-ui-flow`, `argent-device-interact` SKILL.md: the baseline recipes said "keep the returned `path`", which is the temp path. They now pass `out` and say why.
- `docs/features/visual-regression.mdx`: one paragraph saying a screenshot goes to a temporary directory and that you can ask for a path. (`features/` deliberately carries no parameter names; the parameter itself is documented by its schema description, which `argent tools describe screenshot` prints.)

## Relationship to #590

This supersedes #590, which made `.argent/screenshots` durable by default for every explicit capture. That needed a client-side allowlist, a `transient` opt-out for the auto-screenshot, a carve-out for flow-visual's internal capture, a Lens preview-route fix, and a `.gitignore` entry - all of it to keep a default nobody asked for from filling the working tree. Naming the file is the thing the agent actually wanted, and it costs one parameter.

## Where `out` is and is not honored

Both clients write it, over one `resolveOutPath` / `writeOutFile` in `@argent/tools-client`, so the parameter cannot mean two things depending on which one reads it: trim, `~` expansion, a directory-shaped path refused, the resolved absolute path reported back, and the bytes staged through a sibling `.part` file and renamed into place so `out` only ever holds a whole capture.

- **MCP direct call, CLI `--out`, CLI payload `out`** (`--args` / `--out-json`): honored, on the artifact-handle shape and on a pre-artifact-handle tool-server's legacy `{ url, path }` alike.
- **A `tool: screenshot` step inside a flow**: not honored, and it says so. A step's args are the flow file's, not the caller's, so relaying them into a client-side write would let a YAML file pick any path on this machine. The step now carries a `Could not save to <out>: …` line rather than reporting a pass and a `Saved:` line naming a scratch path.
- **The automatic post-interaction screenshot**: structurally cannot write - it calls `toMcpContent` with no args at all.

## Left out

- `argent flow run` never materializes tool-step results, so it prints no `out` diagnostic of its own. Pre-existing, unchanged by this PR.

</details>

## Summary by CodeRabbit

* **New Features**
  * Added an optional screenshot output path for saving PNG files outside temporary session storage.
  * Supports relative paths, home-directory expansion, whitespace trimming, automatic directory creation, and explicit CLI path precedence.
  * Save results report the resolved file location and provide fallback details if writing fails.

* **Bug Fixes**
  * Prevented global options from appearing redundantly as per-tool flags.
  * Clarified that output paths supplied within flow steps are not used on the agent machine.

* **Documentation**
  * Updated screenshot and visual regression guidance for preserving images and using unique output paths.


packages/docs/docs/features/visual-regression.mdx
```diff
@@ -20,6 +20,8 @@ Use the comparison after a change to the layout, the spacing, the colors, the ty
 
 The agent can compare two saved screenshots, or one saved screenshot and a live capture of the device. The agent can also capture both sides live, before and after an interaction.
 
+The agent puts a screenshot in a temporary directory. The system deletes this directory later. Tell the agent where to put the file when you want to keep it, for example a baseline for a later session.
```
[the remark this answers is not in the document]
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in f2e15b83c — split the 22-word instruction into two sentences, each one idea and under the docs' 20-word instruction limit (`packages/docs/CLAUDE.md`). I kept it free of parameter names, since the `features/` pages deliberately carry none: the reader is told to name a path without the page naming `out`.

packages/skills/skills/argent-test-ui-flow/SKILL.md
```diff
@@ -47,7 +47,7 @@ Goal: Test [feature name]
 Steps:
 1. Classify expected result: visual / structural / runtime-log-network / mixed → choose evidence
 2. [Navigate / tap / type to reach stable comparable starting point] → verify auto-screenshot
-3. screenshot { scale: 1.0, includeImageInContext: false } → save baseline path when visual or mixed evidence needs diffing
+3. screenshot { scale: 1.0, includeImageInContext: false, out } → save baseline path when visual or mixed evidence needs diffing
```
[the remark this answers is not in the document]
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 4.8\]: Fixed in f2e15b83c — gave `out` a concrete value (`/tmp/baseline.png`) in both template steps (lines 50 and 88), matching the other skill examples in this PR so the call is copyable and the placeholder is unambiguous.

hubgan (commented):
Reviewed the `out` parameter end to end. The client-side placement is the right call and I confirmed the premise it rests on: `baselinePath` is a file-boundary input, so a client-written path really is the correct side of the boundary.

Everything below was reproduced against running software - the real MCP stdio server over JSON-RPC against a live tool-server, on an iOS simulator, an Android emulator and a Chromium (CDP) target, plus the real `argent run` CLI. The first comment is the one I would want resolved before merge.

For reference, the gates are green here too: prettier, eslint, argent-mcp 93, argent-cli 569, tool-server screenshot + flows 1599, and `docusaurus build`.

packages/argent-mcp/src/content.ts
```diff
@@ -90,6 +93,38 @@ export async function toMcpContent(
...
+ * against. A failed write is reported and the temp path is still handed back,
+ * so a bad `out` costs the agent a copy, never the capture.
+ */
+async function savedText(image: MaterializedImage, args: unknown): Promise<string> {
+  const out = isRecord(args) && typeof args.out === "string" ? args.out.trim() : "";
+  if (!out) return `Saved: ${image.localPath}`;
+  const target = resolve(expandTilde(out));
+  try {
+    await mkdir(dirname(target), { recursive: true });
+    await writeFile(target, image.data);
```
hubgan: The destination of this write is not always a path the agent chose.

`toMcpContent` is also reached from `flowRunToMcpContent`, which passes `step.args`. For a `tool: screenshot` step that is the `args:` mapping copied verbatim out of the flow YAML - `flow-run.ts` echoes the pre-zod args back in the step report - and under `argent link` it is whatever the remote tool-server chooses to put there. So the path can come from a file in the repo rather than from the caller, and the write is unconfined: `mkdir -p` creates the parents and `writeFile` truncates whatever was at the target. Because fragment steps share one flat report list, a `run:`-composed flow reaches the same code, so it is not only top-level steps.

Reproduced against a booted iOS simulator. A flow at `<project>/.argent/flows/poison.yaml`:

```yaml
executionPrerequisite: ""
steps:
  - tool: screenshot
    args:
      out: /abs/path/to/an/unrelated/notes.txt
```

`flow-execute` with `{ name: "poison", project_root: "<project>", device: "<UDID>" }` returns `PASS - 1 passed` and `Saved: /abs/path/to/an/unrelated/notes.txt`; the 39-byte text file is now 173 KB of PNG. Rebuilding `argent-mcp` from the merge base and replaying the identical flow leaves the file untouched and reports the temp path, so the primitive is new here.

Worth noting how the same input class is treated a few files away: `applyClientFileDirectives` refuses anything outside `.argent/flows/*.yaml`, with a comment saying that is specifically so a misbehaving tool-server cannot write elsewhere, and `durableSaveTarget` requires an allowlisted relative directory plus a post-`mkdir` realpath check against symlinked segments. This is the only write in `argent-mcp` that takes its path straight off the wire.

packages/tool-server/src/tools/screenshot/index.ts
```diff
@@ -56,6 +56,17 @@ const zodSchema = z.object({
...
+  // Input-only, like `includeImageInContext`: the write happens on the CLIENT
+  // (argent-mcp's content.ts), because the path names the agent's own
+  // filesystem - a different host under `argent link`, and the one where
+  // `screenshot-diff` resolves a baseline path.
+  out: z
+    .string()
+    .min(1)
+    .optional()
+    .describe(
+      "Save the PNG here on YOUR machine instead of leaving it in a session temp directory that is deleted later. Relative paths resolve against your working directory; missing parent directories are created, and an existing file is overwritten. Pass it whenever the file has to outlive this session - notably a baseline or current image you will hand to `screenshot-diff`."
```
hubgan: "Relative paths resolve against your working directory" is true of the process that writes the file, but the same description then tells the agent to hand the path to `screenshot-diff`, and those two sentences do not compose.

`baselinePath` is a `kind: "file"` boundary input. Co-located - the default local setup - the client sends the raw string with no bytes and the tool-server stats it against its own cwd. The daemon is reused across projects (keyed on bundle path alone, spawned with no explicit cwd), so a long-lived daemon and the current MCP client routinely sit in different directories.

Reproduced with an autospawned daemon whose cwd is `dirA` and an MCP client running in `dirB`. `screenshot { out: "shots/base.png" }` correctly writes `dirB/shots/base.png` and reports it. Then `screenshot-diff { baselinePath: "shots/base.png", captureCurrent: true }` fails with:

> File "shots/base.png" was not found on the tool-server host and the client did not upload its content. Either the file does not exist, or it changed since it was referenced - re-create it (or re-run the producing tool) and try again.

The file does exist, so that advice cannot succeed. Control: running both calls from `dirA` with the identical relative path clears the file boundary and proceeds.

This describe text is `alwaysLoad: true`, so the sentence is permanently in every agent's context, while the corrective guidance ("the reported `Saved:` path is the one to pass on") only arrives if the screenshot-diff skill happens to be loaded.

packages/argent-cli/src/run.ts
```diff
@@ -97,6 +103,15 @@ function printToolHelp(meta: ToolMeta): void {
   console.log("  --help, -h             Show this help");
 }
 
+/** Drop schema properties whose name a global flag shadows, so the per-tool Flags
```
hubgan: The filter is applied to one of the two renderers. `formatSchemaUsage` has exactly two call sites, and `tools.ts` still calls it on the unfiltered schema, so `argent tools describe screenshot` continues to print `--out <value>` with the new description - and that is the surface `argent run --help` points people at to see a tool's flags. `argent tools describe --json` emits the raw schema too.

The asymmetry underneath is that on the CLI the schema `out` never does anything at all. `splitOptions` intercepts only the exact `--out` / `--out=` tokens, so both remaining routes reach the payload, are stripped server-side, and write nothing:

```
$ argent run screenshot --args '{"udid":"<UDID>","out":"/tmp/x.png"}'
Saved screenshot: /var/folders/.../simserver-.../media/784535000-....png
$ ls /tmp/x.png
ls: /tmp/x.png: No such file or directory      # exit 0, no warning
```

`--out-json '"/tmp/x.png"'` behaves identically, which also makes the comment above this function inaccurate: it says a tool property of the same name "can't be reached through its flag", but `--out-json` is not one of the tokens `splitOptions` consumes and does reach the payload as `out`.

packages/argent-mcp/src/content.ts
```diff
@@ -90,6 +93,38 @@ export async function toMcpContent(
   return [{ type: "text" as const, text: stringifyForText(result) }];
 }
 
+/**
+ * The `Saved:` line for an image result, honoring the caller's `out` path.
+ *
+ * A materialized PNG sits in a temp directory that goes away with the session;
```
hubgan: "sits in a temp directory that goes away with the session" does not match what happens, and the same claim is repeated in the two describe strings, the three skills and the docs page.

Two things are off. First the directory: the session-scoped `argent-artifacts/<project>/<session>/<device>` cache is only populated on the download branch, i.e. against a genuinely remote tool-server. Co-located, `resolveLocalFile` returns the tool-server's own `hostPath` and it is used in place with no copy - so the reported path is the simulator-server's media directory, exactly as the PR description's own no-`out` row shows.

Second the lifetime: nothing in argent deletes either location. I captured a screenshot, then stopped that simulator-server through `stop-simulator-server`, and the PNG was still there byte for byte afterwards. `artifactsRoot()` has no GC anywhere in the repo. The file goes away only when the OS reclaims its temp tree.

That matters most where the skill text pushes it furthest - `argent-screenshot-diff/SKILL.md` now says "so a baseline can be gone before you diff against it", which will not happen inside a session.

packages/argent-mcp/src/content.ts
```diff
@@ -90,6 +93,38 @@ export async function toMcpContent(
...
+ * `out` asks for a copy the agent keeps. It is written HERE rather than by the
+ * tool because the path names the agent's filesystem - a different host under
+ * `argent link`, and the one `screenshot-diff` resolves a `baselinePath`
+ * against. A failed write is reported and the temp path is still handed back,
+ * so a bad `out` costs the agent a copy, never the capture.
+ */
+async function savedText(image: MaterializedImage, args: unknown): Promise<string> {
+  const out = isRecord(args) && typeof args.out === "string" ? args.out.trim() : "";
```
hubgan: Two shapes of `out` are accepted and then quietly do something other than what was asked. Both confirmed on a device.

`"   "` clears the schema's `.min(1)`, trims to empty here, and becomes indistinguishable from no `out` at all: the result reports the temp path and nothing says the requested save was dropped.

A path ending in a separator (`/…/shots/`) has the separator stripped by `resolve`, so `shots` is created as a regular file rather than the directory that was named, and the result reports `Saved: /…/shots`. "Missing parent directories are created" reads as though a directory-shaped path would go the other way, and the next call that wants `shots/` to be a directory now fails against a file.

packages/argent-mcp/src/content.ts
```diff
@@ -67,7 +70,7 @@ export async function toMcpContent(
 
     if (outputHint === "image") {
       if (images.length > 0) {
-        const saved: ContentBlock = { type: "text", text: `Saved: ${images[0]!.localPath}` };
+        const saved: ContentBlock = { type: "text", text: await savedText(images[0]!, args) };
```
hubgan: When the image artifact cannot be materialized, `out` is dropped without a word. Two ways in: an older tool-server returning the `{ url, path }` shape, and any download failure (remote 404/401, truncated body), which leaves `images` empty. Both fall through to `legacyImageContent`, which never receives `args`.

The agent then gets either a `Saved:` line naming a path it did not ask for - on a different host entirely under `argent link` - or a bare `{"image": null}`, with no mention of the `out` it requested. With `includeImageInContext: false`, which is the recipe the updated skills prescribe, that is the only signal it has. A stale file left at `out` by an earlier run also survives to be diffed against as if it were fresh.

packages/skills/skills/argent-screenshot-diff/SKILL.md
```diff
@@ -23,10 +23,10 @@ Use `screenshot-diff` when pixel comparison can answer the verification question
 
 ## 3. Capture Rules
 
-Use normal downscaled `screenshot` calls for UI context and state checks. Use full-resolution screenshots only when saving baseline/current PNG files for visual regression comparison. Suppress the image block so the full-size PNG is not loaded into context:
+Use normal downscaled `screenshot` calls for UI context and state checks. Use full-resolution screenshots only when saving baseline/current PNG files for visual regression comparison. Suppress the image block so the full-size PNG is not loaded into context, and name the file with `out` - without it the PNG lands in a session temp directory that is deleted later, so a baseline can be gone before you diff against it:
 
 ```json
-{ "udid": "<UDID>", "scale": 1.0, "includeImageInContext": false }
+{ "udid": "<UDID>", "scale": 1.0, "includeImageInContext": false, "out": "/tmp/baseline.png" }
```
hubgan: This turns a fixed filename into a write instruction. `/tmp/baseline.png` already appeared in this file as an illustrative `baselinePath` value, but nothing wrote there; now three skills tell the agent to actually write to that exact literal, and `savedText` uses a plain truncating `writeFile`.

One tool-server is shared by every agent on an install, so two sessions following this skill race on one name: A's baseline can be replaced by B between A's capture and A's `screenshot-diff`, and the diff then silently compares the wrong pair - a false verdict in the QA-regression case the skill exists for. Argent's own client-side durable writer takes the opposite stance a few files away, using `wx` / `COPYFILE_EXCL` and a `name (2).ext` fallback explicitly so concurrent materializations cannot clobber each other.

Related, in `argent-lens/SKILL.md`, which was not updated: `propose_variant.previewImage` is read on the tool-server host and confined to realpath'd `os.tmpdir()`, the tool-server's `process.cwd()` and `/tmp`. An agent that now habitually keeps captures via `out` somewhere else and passes that path through gets a silent "No preview".

packages/argent-mcp/test/content.test.ts
```diff
@@ -307,6 +307,128 @@ describe("toMcpContent with artifact ctx", () => {
...
+
+    expect(await fs.readFile(out)).toEqual(Buffer.from(pngBytes));
+    expect(result[0]?.type).toBe("image");
+    expect(result[1]).toEqual({ type: "text", text: `Saved: ${out}` });
+  });
+
+  it("resolves a relative `out` against the working directory", async () => {
+    const pngBytes = [...PNG_SIGNATURE, 0x43];
+    const cwd = process.cwd();
+    process.chdir(root);
```
hubgan, line 343: These two cases are silently tied to the default `forks` pool. Running the file with `--pool=threads` produces 2 failures: the relative-path case throws `process.chdir() is not supported in workers`, and the `~` case is the worse of the two - in a worker `process.env` is a JS-side copy that `os.homedir()` never reads, so the PNG is written into the developer's real home at `~/shots/base.png` before the assertion fails. Confirmed by running the file both ways.

Separately, this block covers `out` only on the direct call path. Nothing drives it through `flowRunToMcpContent`, which is where `out` arrives from wire data, and nothing covers the legacy path where `out` is dropped.

packages/tool-server/test/screenshot-tool.test.ts
```diff
@@ -102,6 +102,18 @@ describe("screenshot tool", () => {
     expect(result).not.toHaveProperty("includeImageInContext");
     expect(result).not.toHaveProperty("url");
   });
+
+  it("advertises `out` and passes it through as an input-only path the client writes", () => {
```
hubgan: The title says `out` is "passed through as an input-only path the client writes", but the body only calls `zodSchema.parse`. It never runs `execute`, so nothing here would notice if the tool started reading `params.out` - which is the invariant the title claims. The sibling assertion a few lines above does exactly that check for `includeImageInContext`.

packages/argent-mcp/src/content.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 19ce3914e. Confirmed the primitive first: a synthetic step report with `args: { udid, out: <victim> }` driven through `flowRunToMcpContent` truncated a 39-byte text file to PNG bytes, exactly as you described.

The fix is at the wire boundary rather than in the writer. `flowRunToMcpContent` now launders a step's args through `renderArgsOnly()`, which passes `includeImageInContext` and nothing else, so `out` is honored only when it came from the MCP caller's own `params.arguments`. `toMcpContent`'s fourth parameter is documented as the caller's own args for that reason.

I went with dropping flow-origin `out` rather than confining it, because `argent flow run` never honored it either — so this removes the asymmetry the PR body listed under "Left out" instead of adding a second confinement scheme next to `durableSaveTarget`'s.

Re-verified end to end on iOS 18.6 (`770F654F-…`), through the real MCP stdio server against a worktree tool-server: a flow at `.argent/flows/poison.yaml` with `out: <cwd>/victim-notes.txt` returns `PASS — 1 passed`, reports `Saved: /var/…/simserver-lcBjNV/media/933641000-….png`, and the victim file still reads `important user notes that must survive`. Regression test in `packages/argent-mcp/test/content.test.ts`; reverting the launder turns it red.

packages/tool-server/src/tools/screenshot/index.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 48a036e48. Reproduced without a device by driving the real `prepareFileInputs`/`resolveFileInputs` pair with screenshot-diff's own spec: client cwd `dirB` sends `{"__argentFileInput":true,"path":"shots/base.png",…}` verbatim, the server re-stats it against `dirA`, and `file-inputs.ts:256` throws the exact text you quoted. Control with both cwds equal resolves fine.

You are right that the two sentences name different processes' cwds — `content.ts`'s `resolve()` uses the MCP client's, `tool-server/src/file-inputs.ts:86` uses the daemon's — and `flow-run.ts:2646` already hard-rejects a relative `flow_path` for exactly this reason, which makes the new describe the odd one out.

The describe now reads:

> Save the PNG at this path on YOUR machine instead of the scratch temp path the capture produced. Missing parent directories are created, and an existing file is overwritten. A relative path resolves against your working directory - but `screenshot-diff` resolves one against the tool-server's, so hand it the absolute path reported back in `Saved:`.

`argent-screenshot-diff/SKILL.md` step 2 carries the same correction. I kept relative `out` working rather than rejecting it — unlike `flow_path` it resolves on the side that writes, so it is correct on its own; only handing the relative spelling onward is not.

The code comments you flagged as "right about the host, wrong about the cwd" are reworded too: `screenshot/index.ts` no longer mentions baseline resolution at all, and `content.ts` says "the host `screenshot-diff` reads a `baselinePath` from".

packages/argent-cli/src/run.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 19ce3914e. All three sub-claims reproduced through the real `run()` with a mocked client, printing the payload `callTool` receives:

| argv | payload | file written |
| --- | --- | --- |
| `--out <p>` | `{udid}` | yes (global flag) |
| `--out-json '"<p>"'` | `{udid, out}` | **no**, exit 0, silent |
| `--args '{…,"out":"<p>"}'` | `{udid, out}` | **no**, exit 0, silent |

Rather than filter harder, I made the schema `out` real on the CLI: `imageOut = outPath ?? outFromPayload(payload)`, so a tool's own `out` names the image destination and an explicit `--out` still wins. That makes `argent tools describe screenshot`'s `--out <value>` row truthful, so `tools.ts` needs no filtering — the two renderers now disagree about text, not about reality — and it stops the CLI ignoring the `out` guidance this PR wrote into three skills.

One correction to the mechanism: the tool-server does not strip `out`. `zodSchema` declares it, so zod keeps it in `params`; `execute` simply never reads it. The outcome is as you described.

Both inaccurate comments are rewritten — `GLOBAL_FLAG_NAMES` now says these are flags that already do what a same-named property asks, so listing the property would print the flag twice in one help screen, which is the true reason.

Verified against the live simulator through the real CLI: `--out`, `--out-json` and `--args` each write a 59 957-byte PNG, and with both spellings present only the `--out` path is written. Three regression tests in `run-artifacts.test.ts`; dropping the fallback fails two, inverting the precedence fails the third.

Not actioned: an object-typed `out` would render as `--out-json` and still be filtered from the Flags block. No tool declares one (screenshot is the only `out`, nothing declares `json`), so there is no trigger — flagging it here rather than adding unreachable code.

packages/argent-mcp/src/content.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 48a036e48. Both halves confirmed.

The directory: `materializeArtifacts` reuses `hostPath` in place when the size/mtime gate passes (`artifacts.ts:504-509`) and only populates `artifactDir()` on the download branch, so co-located the reported path is the backend's own file. The E2E on iOS 18.6 shows it plainly — a no-`out` capture reports `Saved: /var/…/T/simserver-lcBjNV/media/588707000-….png`.

The lifetime: `artifactsRoot` has three references repo-wide and none of them sweeps, every `rm`/`unlink` in `packages/*/src` targets something else, and `stop-simulator-server` ends at a bare `proc.kill()` (`blueprints/simulator-server.ts:387`) with no fs call. Empirically this machine still holds an `argent-artifacts` PNG from Sep 2 and 38 orphaned `simserver-*` dirs going back to Aug 23. It goes away when the OS reclaims its temp tree, on a multi-day horizon.

All six sites now say the file sits on "a scratch temp path you did not choose" rather than a directory something deletes. The screenshot-diff line you singled out no longer claims a baseline can vanish mid-session — it says the scratch path is not somewhere to build a baseline, which is the true reason to pass `out`. The docs page keeps its own register: "Without a target path, the agent leaves the screenshot in the temporary directory of the system", matching `profiling.mdx:59`.

One thing worth recording that neither of us had: radon's session dir is a `tempfile::TempDir`, so its `Drop` does wipe `simserver-*/media/` on a graceful sim-server exit (stdin EOF, killswitch, device-stream end) — but not on the SIGTERM `stop-simulator-server` sends, which is why your byte-for-byte check passed. Undocumented anywhere; I have not written it into this PR since it is not what any of these sentences was claiming.

packages/argent-mcp/src/content.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 48a036e48. Both shapes reproduced.

Whitespace-only: `z.string().min(1)` counts the spaces, so `"   "` parsed clean and then trimmed to nothing client-side. The schema is now `z.string().trim().min(1)`, so it is rejected at the boundary before any capture is spent. Live: `{ out: "   " }` returns ```out`: Too small: expected string to have >=1 characters``` instead of a silent temp path. The client-side `.trim()` stays load-bearing — `savedText` reads the caller's raw args, not the server's parsed copy, so `" /tmp/x.png "` still lands at the clean path.

Trailing separator: confirmed a regular 9-byte file named `shots` where a directory was asked for, and confirmed it is sticky — the next save under it fails EEXIST at the immediate level (thanks for the ENOTDIR correction going the other way; `mkdir(dirname)` trips before `writeFile` is reached, and ENOTDIR appears one level deeper). `savedText` now refuses a directory-shaped `out` up front, at the cost of the copy rather than the capture, matching the existing catch:

```
Saved: /var/…/simserver-lcBjNV/media/626911000-….png
Could not save to /…/shotsdir/: out names the file to write, not a directory.
```

and `ls` shows no `shotsdir` entry afterwards. Both verified on the live simulator; regression tests in `content.test.ts` and `screenshot-tool.test.ts`, and each dies when its guard is reverted.

packages/argent-mcp/src/content.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 48a036e48. Reproduced all three: the legacy `{url, path}` shape with `out` set wrote nothing and never mentioned `out`; a `{ok:false}` download reduced the whole response to `{"image": null}`; and with `includeImageInContext: false` that was the agent's only signal.

The stale-file half is the one I treated as the serious one, since it turns into a false PASS in exactly the regression check the feature exists for. Both are closed:

- `legacyImageContent` now takes `args` and routes the bytes it already fetches through the same writer, so `out` is honored on that shape too — the CLI's `--out` has always handled it via `fetchImageToFile`, so the MCP side was the outlier. It also fetches when the image is suppressed but `out` is set, which is the recipe the skills prescribe and previously returned early without fetching.
- When there are no bytes at all, `unsavedBlocks()` appends: `Could not save to <out>: no image came back, so there was nothing to write. Any file already at that path is stale - do not diff against it.`

That also makes the `savedText` docblock true again rather than needing narrowing, which was the right catch. Regression tests cover the legacy write, the suppressed-plus-`out` fetch, and the stale-file case; neutering `unsavedBlocks` fails six, and making the legacy path ignore `args` fails two.

This supersedes the second bullet of the PR body's "Left out" section, which I have updated.

packages/skills/skills/argent-screenshot-diff/SKILL.md
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 48a036e48 for the race; the lens half is real but out of scope and now has its own issue.

The race stands on its own — one tool-server serves every agent on an install, and three skills had just turned one literal into a write instruction. All the examples across the three files are now run-unique (`/tmp/<run-id>-baseline.png`, `/tmp/<run-id>-current.png`), and the capture rules say why: a shared name lets another session overwrite your baseline between capture and diff.

I kept the plain truncating `writeFile` rather than adopting `durableSaveTarget`'s `wx`/`COPYFILE_EXCL` + `name (2).ext`. Those two writers are answering different questions: `saveDir` is a destination the tool-server suggests and several artifacts can land in at once, so a unique suffix is right there; `out` is a path the caller typed, and re-capturing a baseline to the same path has to overwrite or the second capture silently lands somewhere the caller is not looking. The collision to remove was the shared *name*, which is what the skills now handle.

I also kept the examples under `/tmp` deliberately, because of your lens note: `preview.ts:530-543` allows `os.tmpdir()`, the tool-server's `process.cwd()` and `/tmp`, so steering `out` toward a project directory would have newly broken `propose_variant.previewImage` for anyone mixing the two skills. `argent-lens/SKILL.md` still says to pass the returned path straight through and does not mention the confinement, which is the actual gap — filed as its own issue rather than folded in here, since it is a skill this PR does not touch.

packages/argent-mcp/test/content.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Coverage half fixed in 19ce3914e and 48a036e48; the pool half I am declining, with the mechanism confirmed.

Coverage first, since that was the substantive part: `flowRunToMcpContent` now has a case driving `out` through wire args (it asserts the named file is untouched), and the legacy path has three — `out` honored on `{url, path}`, the suppressed-plus-`out` fetch, and the no-bytes case with a stale file at `out`.

On the pool: I reproduced it exactly as you did. Under `--pool=threads` the file goes 2 failed / 41 passed, and `~/shots/base.png` really is created in the developer's home, 9 bytes ending `0x44` — that test's own payload. In a worker `process.env` is a JS-side copy and `os.homedir()` reads libuv's `environ`, so the redirect is invisible to it. Deleted afterwards; it had not existed before.

What stopped me acting on it is reachability and blast radius. No `vitest.config.ts` in the repo sets `pool`, no script or workflow passes `--pool`, so nothing reaches this except a hand-typed flag. And it is not this PR's property: `process.chdir` appears at 40+ pre-existing test sites and `process.env.HOME` at 38+, including the shared `tool-server/test/helpers/temp-home.ts`. The genuinely destructive instance is elsewhere — `argent-installer`'s tests redirect only `HOME` while `mcp-configs.ts` resolves `~/.claude.json` and `~/.config/zed/settings.json` through bare `os.homedir()`, so under threads those clobber real editor config. Fixing one of 79 sites would not make the suite threads-safe and would leave the worse ones standing.

So: correct, and worth knowing, but it belongs to a suite-wide decision (pin `pool` in config, or move the env redirects behind an injectable resolver) rather than to this diff.

packages/tool-server/test/screenshot-tool.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Fixed in 48a036e48. Confirmed by mutation before touching it: with `return { image, out: params.out };` substituted at all five exit points the file still went 7/7 green, and the same mutation with a spread was caught — by the `includeImageInContext` sibling at line 102, not by this test. So the title's second clause was entirely unpinned, exactly as you said.

The test now stubs `fetch` and the registry the way the sibling does, calls `execute`, and asserts `expect(result).not.toHaveProperty("out")`. Re-running the leak mutation now fails it. The schema assertions stay, and picked up a third case in the process — `parse({ udid: "ABC", out: "   " })` must throw, which pins the `.trim()` added for the whitespace finding.

packages/argent-cli/src/run.ts
```diff
@@ -368,15 +391,20 @@ Examples:
     process.exit(1);
   }
 
+  // `--out` wins; without it a tool's own `out` property names the destination, so
+  // the spellings that reach the payload (`--args`, `--out-json`) do what the
+  // schema advertises rather than passing a path nothing on this side reads.
+  const imageOut = outPath ?? outFromPayload(payload);
```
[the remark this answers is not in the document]
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: Fixed in 5c1def4e4. Reproduced first, through the real `run()` against a stub tool-server:

| argv | before | after |
| --- | --- | --- |
| `--args '{"out":"<p>"}' --out=` | exit 0, no `Wrote:`, **neither** destination written | exit 2, `Error: --out requires a path` |
| `--out '   '` | exit 0, `Wrote:    `, a file literally named `   ` created in the cwd | exit 2, same error |
| `--out ' <abs> '` | `Wrote:  <abs> `, and the PNG buried under a directory named `" "` because `path.resolve` reads the leading space as relative | writes `<abs>`, `Wrote: <abs>`, no stray directory |

The third row is the one that made me take this beyond the empty case: `--out` was the only spelling left untrimmed once this PR added `outFromPayload`'s `.trim()` beside it, so one variable meant two things. `splitOptions` now trims both `--out <v>` and `--out=<v>` and refuses an empty result, which lands on the same `FlagParseException` the missing-value case already used.

I did not reject whitespace *inside* a path - a filename may legitimately contain spaces, and only the surrounding padding was ambiguous.

Three regression tests in `run-artifacts.test.ts`; reverting the trim fails two of them.

packages/argent-cli/src/run.ts
```diff
@@ -368,15 +391,20 @@ Examples:
...
+
   // Prefer the bytes the materializer already resolved; fall back to fetching the
   // legacy `{ url }` for older tool-servers that don't emit artifact handles.
-  if (outPath && meta.outputHint === "image") {
+  if (imageOut && meta.outputHint === "image") {
```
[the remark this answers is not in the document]
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: Fixed in 5c1def4e4, and the fix went wider than the branch you named.

Reproduced verbatim first. With `data: null` and with `data: "no-image-here"`, `argent run screenshot --out <p>` printed:

```
stdout: null
stdout: Wrote: /var/.../never.png
exit: 0        # file does not exist
```

So the `Wrote:` line named a path holding nothing - or, worse, whatever an earlier run had left there.

Rather than add an `else` beside the two existing arms I collapsed the whole block into one `saveImageTo`, because the same audit turned up two more defects in the same seven lines: the CLI never expanded `~` (the MCP twin does, deliberately), and a failed write did `process.exit(1)` *before* `renderResult`, so a bad `out` also discarded the only line naming a capture that had already succeeded. All three are the same bug - the CLI writer had drifted from the MCP writer for one shared parameter - so both now go through `resolveOutPath` in `@argent/tools-client`.

After:

```
stdout: {"image": null}
stderr: Could not save to /var/.../never.png: no image came back, so there was nothing to write. Any file already at that path is stale - do not diff against it.
exit: 1
```

One correction to the finding's shape: `{ image: null }` - the reachable case, an artifact handle that failed to materialize - is an *object*, so it never took the branch you describe; it went through `fetchImageToFile` and did exit 1, just without warning about the stale file. The genuinely silent case needed `result` to be `null` or a primitive. Both are closed by the same guard.

Five regression tests in `run-artifacts.test.ts`. Mutation-checked: making the no-bytes path return a write again fails three, and restoring the old print-after-exit ordering fails two.

packages/argent-tools-client/src/out-path.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * `~` never reaches an argument through a shell — an MCP arg is JSON, and the
+ * CLI's only route to a tool's own `out` is JSON inside `--args`/`--out-json` —
+ * so expand it here rather than creating a directory literally named `~` in the
+ * agent's project.
+ */
+function expandTilde(p: string): string {
+  if (p === "~") return homedir();
+  return p.startsWith("~/") || p.startsWith(`~${sep}`) ? join(homedir(), p.slice(2)) : p;
```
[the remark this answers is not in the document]
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: Fixed in db44a03d3, and the fix is wider than bare `~`.

Reproduced against the built dist before touching it:

```
resolveOutPath("~")          -> {"path":"/Users/ignacylatka"}
resolveOutPath("~/")         -> {"path":"/Users/ignacylatka"}
resolveOutPath("~/shots/.")  -> {"path":"/Users/ignacylatka/shots"}
writeOutFile("~", png)       -> EACCES ... open '/Users/ignacylatka.95427-ebaorf.part'
```

So bare `~` was one of three: the guard ran on the expanded path, and `expandTilde` calls `join`, which collapses a trailing separator and a `.` segment before the guard ever looks. `~/` and `~/shots/.` reached `resolve` looking like filenames for the same reason a bare `~` did.

The guard now runs on the trimmed, pre-expansion spelling, with bare `~` named explicitly since it never carries a directory shape at all. Same four inputs on the rebuilt dist:

```
"~"           {"refusal":"out names the file to write, not a directory."}
"~/"          {"refusal":"out names the file to write, not a directory."}
"~/shots/."   {"refusal":"out names the file to write, not a directory."}
"~/base.png"  {"path":"/Users/ignacylatka/base.png"}
writeOutFile("~") -> {"failure":"Could not save to ~: out names the file to write, not a directory."}
```

Covered by an `it.each` over `~`, `~/`, `~/shots/`, `~/shots/.` and `~/shots/..`. Two mutations confirm the tests discriminate: dropping the bare-`~` case fails 1, moving the guard back after expansion fails 4.

packages/argent-mcp/src/content.ts
```diff
@@ -298,7 +363,10 @@ export async function flowRunToMcpContent(
     });
 
     if (step.result !== undefined) {
-      blocks.push(...(await toMcpContent(step.result, step.outputHint, ctx, step.args)));
+      blocks.push(
+        ...(await toMcpContent(step.result, step.outputHint, ctx, renderArgsOnly(step.args)))
+      );
+      if (step.outputHint === "image") blocks.push(...unhonoredOutBlocks(step.args));
```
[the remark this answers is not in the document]
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: Addressed in db44a03d3, though not by moving the call - `unhonoredOutBlocks` is gone.

The premise held: the block was inside `if (step.result !== undefined)`. But the client was the wrong place for it. `argent flow run` renders the same report through `packages/argent-cli/src/flow.ts` and never had the diagnostic at all, so a CLI user got a green step and a stale baseline no matter what `step.result` held. Driving the real `flow(argv, options)` against a stub tool-server proved it: buffered and `--json` both printed a plain pass on a step carrying `out`.

The runner now raises it as a step `warning` instead, which is the field already meant for "passed in a way that weakens it as proof". That renders in all four places for free - the CLI report, the CLI failing-steps renderer, `--json`, and the MCP step line at content.ts:348 - and the MCP step line is emitted **before** the `step.result` guard, so the case you named is covered as a side effect.

Verified on the built dist, driving the real runner and then the real MCP renderer over its report:

```
--- result undefined: status=pass result=undefined warning=YES
    MCP: [1] ✓ screenshot ⚠ `out` was not written: a flow step's arguments come from the flow file, ...
--- tool throws:      status=error result=undefined warning=no
    MCP: [1] ✗ screenshot — device vanished | FAIL — 0 passed, 0 failed, 1 errored, 0 skipped
```

The second line is the deliberate limit: `warning` is documented pass-only, and a step that errored already renders red with nothing claiming a capture was saved anywhere, so there is no green step to correct.

E2E through `argent flow run`:

```
  ⚠  1 tool screenshot
       ⚠ `out` was not written: ... Anything already at /Users/ignacylatka/baselines/home.png is from
         an earlier run - do not diff against it. Call `screenshot` directly to keep a capture.

PASS — 1 passed, 0 failed, 0 errored, 0 skipped, 1 warning
```

Covered by two tests in `flow-tools.test.ts` (warned with `out`, unwarned without, including a whitespace-only `out`); three mutations - dropping the spread, dropping the empty guard, dropping the `.trim()` - each fail. `flow-yaml.mdx` claimed this behaviour already and was false when written; it is now true.

latekvo (commented):
Verified against `e5acc559759c`. Gates locally: `tsc --build` clean; all workspace suites green (6169 passed, 1 skipped). New findings below as inline comments; each of the three substantive ones reproduced against the built `dist/` at head. Everything else came back clean: the flow-step write confinement (wire args cannot trigger a client write), `.part` cleanup on every catchable failure path, symlink/special-file refusals, `~user` non-expansion, the auto-screenshot no-args path, and the schema-vs-resolver agreement (trim, `~`, overwrite of regular files) all hold as described.

Also worth noting, outside the diff: `argent tools describe screenshot` still lists `--out <value>` from the unfiltered schema at `packages/argent-cli/src/tools.ts:52` — the second `formatSchemaUsage` call site (the first was filtered at `run.ts:102`). Unlike `run --help`, the describe page has no Global flags block to reconcile it, so the same property renders once with a flag spelling that `splitOptions` intercepts before the schema parser runs. The invariant the comment above `GLOBAL_FLAG_NAMES` states ("never prints a second row for a flag the Global flags block lists") holds at `run.ts` only; this renderer never learned the filter.

packages/argent-tools-client/src/out-path.ts
```diff
@@ -0,0 +1,120 @@
...
+ * would otherwise push a filename the caller may legally use (up to 255 bytes)
+ * over the limit, failing a write that the target path alone permits.
+ */
+function stagingPath(target: string): string {
+  const suffix = `.${process.pid}-${Math.random().toString(36).slice(2, 8)}.part`;
+  const base = basename(target);
+  const room = 255 - Buffer.byteLength(suffix);
+  const trimmed =
+    Buffer.byteLength(base) <= room ? base : Buffer.from(base).subarray(0, room).toString();
+  return join(dirname(target), `${trimmed}${suffix}`);
```
latekvo, line 89: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]: A legal, byte-length-compliant name fails at the staged `.part` write: the truncation cuts on a byte boundary, so a basename of multi-byte characters can end up longer than the target's.

The comment on this function promises the staging path "cannot itself exceed NAME_MAX" and that a legal 255-byte name must not "fail[] a write that the target path alone permits". `Buffer.from(base).subarray(0, room)` cuts mid-codepoint and `toString()` re-encodes each broken sequence as U+FFFD (2 bytes), so the staged name inflates instead of shrinking. Reproduced against the built `dist/out-path.js` (target basename 252 bytes of `é`, legal on the host):

```
resolveOutPath -> {path: ".../é×124.png"}
writeOutFile -> {failure: "Could not save to .../é×124.png:
  ENAMETOOLONG: name too long, open '.../é…é\uFFFD.2429552-rcu0lx.part'"}
```

A single-byte basename of the same length is trimmed correctly, so the existing 251-`x` test does not cover this shape. The failure also names the `.part` sibling, the exact wording `occupantRefusal` exists to keep out of these messages. Consequence: a legal `out` the staging design exists to support is refused, and the refusal text references a path the caller never asked about.

packages/argent-mcp/src/content.ts
```diff
@@ -105,25 +155,29 @@ function stringifyForText(value: unknown): string {
...
+    return [{ type: "text", text: JSON.stringify(result, null, 2) }, ...unsavedBlocks(args)];
   }
-  return [{ type: "text", text: JSON.stringify(result, null, 2) }];
+  const r = result as { url: string; path?: string };
+  // Suppressing the image normally spares the fetch, but `out` still needs the
+  // bytes — `includeImageInContext: false` plus `out` is the baseline recipe.
+  const buf = suppressImage && !requestedOut(args) ? null : await fetchPngBytes(r.url);
+  if (!buf) {
+    const head: ContentBlock = suppressImage
+      ? { type: "text", text: `Saved: ${r.path ?? ""}` }
```
latekvo, line 172: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]: When the legacy fetch yields no bytes and the image is suppressed, the head block is `Saved: ${r.path ?? ""}` — the stale host path — while `unsavedBlocks` below it reports the write failed and calls the file at `out` stale. The result carries a success-shaped `Saved:` line and a failure in the same payload.

Reproduced against the built `dist/content.js` with a legacy result `{ url, path: "/tmp/old-capture.png" }`, `includeImageInContext: false`, `out` set, and a 404 media URL:

```json
[
  { "type": "text", "text": "Saved: /tmp/old-capture.png" },
  { "type": "text",
    "text": "Could not save to /tmp/…/m2-baseline.png: no image came back, so there was nothing to write. Any file already at that path is stale - do not diff against it." }
]
```

An agent following the baseline recipe reads the first block and continues with `/tmp/old-capture.png` as its fresh capture. The non-suppressed arm of this same branch does not have the issue — its head block is the "(Screenshot unavailable: …)" placeholder — so the failure shape depends on whether the image was suppressed.

packages/argent-tools-client/src/out-path.ts
```diff
@@ -0,0 +1,120 @@
...
+  const resolved = resolveOutPath(out);
+  if ("refusal" in resolved) return { failure: `Could not save to ${out}: ${resolved.refusal}` };
+  const target = resolved.path;
+  const occupied = await occupantRefusal(target);
+  if (occupied) return { failure: `Could not save to ${target}: ${occupied}` };
+  const staged = stagingPath(target);
+  try {
+    await mkdir(dirname(target), { recursive: true });
+    await writeFile(staged, data);
+    await rename(staged, target);
```
latekvo, line 112: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]: `writeOutFile` silently replaces a read-only file and resets its permissions; the sibling staged-rename writer in this repo refuses that exact case.

`occupantRefusal` passes on any regular file, so a `chmod 0444` baseline is swapped away by `rename` and the staged file's umask mode (0644) rides over — reproduced against `dist/out-path.js`: `writeOutFile` on a 0444 baseline returns `{wrote: ...}`, the content is replaced, and the mode becomes 644.

The monorepo's twin, `writeFlowFile` (`packages/tool-server/src/tools/flows/flow-utils.ts:3811-3845`), implements the identical sibling-temp + rename pattern and refuses the same condition explicitly, with its own comment explaining why: rename needs permission on the directory rather than the file, so a write-through would fail with EACCES while the swap silently succeeds — "refused explicitly rather than quietly overwriting a flow you made read-only" — and it re-applies `previousMode` to the staged file before renaming. This writer has neither the refusal nor the mode re-application.

packages/argent-mcp/src/content.ts
```diff
@@ -84,12 +93,53 @@ export async function toMcpContent(
...
+ * rather than by the tool because the path names the agent's filesystem - a
+ * different host under `argent link`, and the host `screenshot-diff` reads a
+ * `baselinePath` from. A failed write is reported and the scratch path is still
+ * handed back, so a bad `out` costs the agent a copy, never the capture.
+ */
+async function savedText(scratchPath: string, data: Buffer, args: unknown): Promise<string> {
+  const out = requestedOut(args);
+  if (!out) return `Saved: ${scratchPath}`;
+  const saved = await writeOutFile(out, data);
+  return "wrote" in saved ? `Saved: ${saved.wrote}` : `Saved: ${scratchPath}\n${saved.failure}`;
```
latekvo, line 117: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]: An agent following the updated `argent-screenshot-diff` skill scans for the path "reported in `Saved:`" (`SKILL.md`, step 2), but on the CLI the honored destination prints as `Wrote:` (`run.ts:473`), preceded by a `Saved screenshot: <scratch path>` line. Both clients share `writeOutFile`, yet one reports the kept file as `Saved:` and the other as `Wrote:` — an agent driving the same recipe through `argent run` finds `Saved` pointing at the scratch path and uses it.

packages/argent-mcp/src/content.ts
```diff
@@ -105,25 +155,29 @@ function stringifyForText(value: unknown): string {
...
+  if (!(result && typeof result === "object" && "url" in result)) {
+    // Not a renderable screenshot at all — an artifact handle that failed to
+    // materialize leaves `{ image: null }` here.
+    return [{ type: "text", text: JSON.stringify(result, null, 2) }, ...unsavedBlocks(args)];
   }
-  return [{ type: "text", text: JSON.stringify(result, null, 2) }];
+  const r = result as { url: string; path?: string };
+  // Suppressing the image normally spares the fetch, but `out` still needs the
+  // bytes — `includeImageInContext: false` plus `out` is the baseline recipe.
+  const buf = suppressImage && !requestedOut(args) ? null : await fetchPngBytes(r.url);
```
latekvo, line 169: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]: Legacy `{url, path}` results are fetched on every failed write to `out`, including when the failure happens before `writeOutFile` is ever reached — so a wedged media host holds a caller's whole turn open.

`resolveOutPath` refusals return synchronously, but the code that feeds bytes in fetches first and resolves later: in `argent-mcp`'s `legacyImageContent` (`content.ts:169`) the `fetchPngBytes(r.url)` call happens before `savedText` ever touches `writeOutFile`, and `fetchPngBytes` runs a bare `fetch(url)` with no `AbortSignal` — the 30s `FETCH_TIMEOUT_MS` in `mcp-server.ts` covers only the tool-server POST, not this direct fetch. The CLI's `fetchLegacyImage` (`run.ts:138`) has the same shape. With `includeImageInContext: false` + `out` (the recipe the updated skills prescribe) that fetch runs for every capture, on a shape the newer artifact path avoids — so a legacy-shape server plus a hung media host leaves the MCP tool call (or the `argent run` process) hanging with no timeout at any layer. Consequence: an agent's turn stalls indefinitely on a tool whose modern path would have answered in seconds.

packages/argent-mcp/src/content.ts
latekvo, line 117: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]: `writeOutFile` throws when its argument is not a string, violating its `OutWriteResult` contract — a malformed `out` degrades the whole render instead of degrading to a failure block.

`requestedOut` duck-types the caller's args (`typeof args.out === "string"`), which protects the normal path, but `writeOutFile` itself starts with `out.trim()`, so anything non-string throws `TypeError` rather than returning `{failure}` like every other bad input it accepts. The type contract (`OutWriteResult = {wrote} | {failure}`) promises failures are returned, not thrown. Today the only reachable caller is the validated one, so this is a contract gap rather than a live bug — but every sibling input shape (empty, whitespace, directory-shaped, occupied) is reported, and a future raw caller (the MCP adapter does not zod-validate tool args itself; only the tool-server's HTTP layer does) turns the same input into a thrown render instead of a refusal block.

PR #1061, opened 2026-09-08
feat(flows): environment values for a script step

hubgan:
Lets a flow `script` step take input from environment values: defaults in the flow file, values for one run from `argent flow run --env` or `flow-execute`'s `env`, and values for one step. A credential is passed as `{{secret:NAME}}`.

## Why

A script step has no input channel today. It gets argent's host allowlist and nothing else, so a base URL, a build number or a credential has to be hard-coded in the script or read from a file, and a CI job that needs a different value has to edit the YAML. Now the flow file holds the defaults a project checks in, and the run supplies what changes:

```yaml
env: # a default for every script in this file
  API_URL: https://api.example.com
steps:
  - script:
      path: scripts/create-user.mjs
      env: # this script only
        API_KEY: "{{secret:API_KEY}}"
```

```bash
argent flow run checkout --env API_URL=https://staging.example.com
```

An `.mjs` script reads `process.env.API_URL` and a `.sh` script reads `$API_URL`.

## Where a value comes from

Each layer replaces the names it sets in the layers before it:

1. The host allowlist, extended by the new `scripts.env.allow` config key
2. The root flow file's `env:`
3. Each nested `run:` fragment's `env:`, outermost first
4. The run's values: `argent flow run --env` or `flow-execute`'s `env`
5. The `script` step's `env:`

So `--env` overrides the file's defaults at any depth, but not a step's own `env:`. A `run:` fragment inherits the values and can override them inside itself. A `tool: flow-execute` step starts a separate run and does not inherit, so it passes what it needs in `args.env`.

`scripts.env.allow` (project and global, unioned) lists extra variables a script inherits from the tool-server's environment. Invalid entries, reserved names and names starting with `ARGENT_` are ignored, so a checked-in config file, which an agent can write, cannot hand a script argent's own variables. A note, once per run, names each ignored entry and any config file that is not valid JSON.

## Secrets

`{{secret:NAME}}` resolves on the tool-server from the existing secret sources. For a script, the project's secret files are read from the run's `project_root`, not from the tool-server's working directory, which an editor sets to `/` or `$HOME`. A missing secret in a flow file's or a step's `env` errors the script step before its script starts. One in the run's values prevents the run from starting.

A resolved value is replaced by its placeholder in what a script step reports: the failure message and stack (a `.sh` step's stderr reason included), the log, and the notes that quote an error. A script that throws `` `401 from ${process.env.API_URL} for key ${process.env.API_KEY}` `` reports:

```
401 from https://api.example.com for key {{secret:API_KEY}}
```

Argent matches the value as written. For a value of six characters or more it also matches the forms that the new `@zapier/secret-scrubber` dependency lists: `encodeURIComponent` (also with `+` for a space), the JSON-escaped body and the base64 of the whole value. Other encodings, such as hex, are not matched. The reference says so and tells authors not to print a credential.

Every other `env` value is flow data, shown as written in the flow file, reports and tool logs. The output document a script returns is not scrubbed either: it is data for a later step, where a placeholder would be useless. No step reads it yet, but `flow-add-script` returns it to the recording agent, so the docs say not to put a credential there. The base's `scrubDocument`, which no caller reached, is removed.

## What is refused

Refused wherever a value can arrive (the flow file, `flow-execute`, `flow-add-script`, `--env`), because most of these would otherwise reach the script silently dropped or changed:

- a name that does not match `[A-Za-z_][A-Za-z0-9_]*`, and on Windows two names that differ only in case
- a reserved name: one that steers the runner's own process, such as `NODE_OPTIONS` or npm's spellings of `node-options`, plus `ARGENT_OUTPUT` and `__proto__`
- a value that is not a string, a YAML-tagged map, a NUL or an unpaired surrogate
- a `{{output:...}}` reference

## How it works

- `FlowFile` gains an optional `env` next to `steps`. `appendStep` rewrites the file from that type, so without the field any `flow-add-step` call would delete a hand-written `env:`.
- `flow-execute` validates its `env` and checks that the secrets in it resolve, then keeps the unresolved map on the run state. A `run:` fragment merges its `env:` into the step scope, and `runScriptStep` merges the scope, the run's values and the step's `env`, in that order.
- `runFlowScriptStep` resolves the placeholders against `project_root` and passes the executor the resolved map and the list of secrets. The executor already accepted both, but no caller passed them. `buildChildEnv` adds the `scripts.env.allow` names.
- Redaction keeps the existing scrubbers: `scrubSecretValues` for whole text, and `scrubSecretChunk` for the live log, which holds back a tail across pipe chunks. `secretForms` expands each secret with `makeSensitiveBank`. The package's own `scrub()` is not used: run after a plain pass, it breaks a secret nested inside another secret's encoding, and its `:censored:` tokens hash with a public salt.
- The reserved-name table moved from the executor to `configuration-core`, so the CLI checks `--env` against the same table as the tool-server. It also gains `NODE_CHANNEL_SERIALIZATION_MODE`.
- `serializeFlow` writes every scalar on one line, because a long multi-line value, such as a key or a JSON blob, came back changed when it held a whitespace-only line. A long `echo` message is no longer wrapped at 80 columns as a result. `parseFlow` reads every map key as a string, so a name such as `TRUE` or `NULL` keeps its spelling, and no longer strips trailing Unicode whitespace, such as a no-break space, from the last value in the file.

## Also in this change

- `flow-add-script` takes an `env`, runs the script with the flow file's `env:` plus that map, and records the map as the step's `env`. It reads the flow file before running the script, so a missing or broken file now refuses the call without running it. If the file's `env:` changes during the run so that the recorded step would replay with different values, the result says so.
- When a `flow-execute` call is recorded as a `run:` step, the recorder warns if the call passed an `env` (a `run:` step has none), or if the recording's `env:` would reach the fragment at replay but did not reach the live call. `flow-finish-recording` repeats the warning below the step.
- Recording tools never add a top-level `env:`, and `flow-start-recording` still clears the whole file. Its description and the skills say to save `env:` first.
- `argent flow run --env NAME=value` can be repeated. The value is everything after the first `=`, and the last value for a name wins. A bad name is refused before the tool-server is contacted. A directory run stops on a refusal about the call itself, such as a missing secret in `--env`, instead of repeating it for every flow.
- A failure that reads as a missing command gets a note: the tool-server keeps the environment it started with, `PATH` included, so a later `export` never reaches a script. If the run sets `PATH` through `env`, the note says that instead.
- A directory run's `re-run:` command (from #1184 on main) carries the run's `--env` values, quoted for the shell, so the printed command runs the failed flow with the same values.
- Smaller fixes: a spawn that fails with `E2BIG` names the environment's size and the ARG_MAX limit. The Bash version probe runs under the step's environment, so it now drops `BASH_ENV`, `ENV`, `SHELLOPTS` and `BASHOPTS`. On Windows, exit code 134 with V8's heap-fatal message is reported as a heap exhaustion, not as the script exiting with 134.

## Docs and tests

- Docs: `reference/flow-yaml.mdx` (an Environment values section, the step's `env` key, `--env`), `reference/configuration.mdx` (`scripts.env.allow`, where a script's secrets resolve), `features/flows.mdx`, and the `argent-create-flow` and `argent-qa-flows` skills.
- Tests: new suites for env values (`flow-script-env.test.ts`), log redaction (`flow-script-log-redaction.test.ts`) and the reserved-name table (`script-env-names.test.ts`), plus new cases in the parser and serializer, redaction, environment, recording, CLI and HTTP suites. The Windows job now runs the env, environment and redaction suites.


latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: Reviewed at `4b1636fa2db378089f07a1f535c30273f5f1816e` against its stacked base `f42bb797c3dd0b2e26a45e47303f65522a0934cf`, in a built worktree of the branch (root `npm run build` clean, the PR's suites green at 1761 tests). Everything below was run through the real entry points - `createRunFlowTool` for `flow-execute`, the recording tools from `dist` for `flow-start-recording` / `flow-add-script`, `parseFlow` / `serializeFlow` for the YAML round trip, and `argent flow run --env` for the CLI channel - not read off the diff.

What I exercised: the five-layer precedence and the `run:`-fragment override, `{{secret:NAME}}` resolution against `project_root`, redaction on both the `.mjs` and `.sh` failure paths, the reserved-name / `__proto__` / malformed-name / NUL / case-fold refusals on every channel, the command-not-found note, `scripts.env.allow`, the flow-file round trip under the new `yaml` options, ARG_MAX behaviour on both interpreters, and a mutation pass over the new env and redaction code (each mutation applied in a scratch worktree, restored immediately, with the full suite run against it).

The comments fall into two groups.

The first is the failure-text redaction the description calls out as the guarantee - `{{secret:NAME}}` taking the value's place in a failed script step's message and stack. It holds for the shapes the new tests cover, but five distinct renderings get a resolved credential into the step reason, the `--json` report and `~/.argent/mcp-calls.log` in the clear. Each of those threads carries its own run and, where the neighbouring shape redacts correctly, the control that discriminates them.

The second is coverage. Seven lines added by this change can be mutated with the whole suite still green, and in four of those cases a credential is then printed or an entry silently dropped; twice the test named for the very shape passes under the mutation. Those threads give the mutation, the suite result, and the probe output either side of it.

A few smaller threads sit outside both groups - a docs claim, a CLI verdict, prose that narrates the change rather than the state.

I stopped after a sweep that returned nothing above LOW.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
```diff
@@ -1139,38 +1260,860 @@ function findOwnProtoKey(root: Record<string, unknown>): string | undefined {
...
+  const runs: ByteRun[] = [];
+  let run: ByteRun = { tokens: [], cut: false };
+  let end = -1;
+  const close = (cut: boolean) => {
+    if (run.tokens.length > 1) runs.push({ tokens: run.tokens, cut });
+    run = { tokens: [], cut: false };
+  };
+  for (const token of text.matchAll(numbers)) {
+    const gap = end < 0 ? "" : text.slice(end, token.index);
+    if (/[A-Za-z]/.test(gap)) close(false);
```
latekvo, line 1981: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: `byteRuns` closes a numeric run only on an ASCII letter, so the first word after a `<Buffer …>` rendering keeps the run open when its leading hex-character prefix is odd-length (`d` of "did", `e` of "expected", `a` of "and", `f` of "from"). `decodeByteRun` then rejects the whole run at radix 16, `byteRunSpans` yields no spans, and `repairByteRenderings` replaces nothing — so the full hex of the value a `{{secret:NAME}}` resolved to reaches the step reason, the `--json` report and `~/.argent/mcp-calls.log` in the clear.

Ran on a built worktree at 4b1636f, project holding `.argent/secrets.env` with `ARGENT_SECRET_API_KEY=sec-9d3f-TOPSECRET`, step `env: { API_KEY: "{{secret:API_KEY}}" }`, and a `.mjs` throwing ``util.inspect(Buffer.from(process.env.API_KEY)) + " " + WORD + " not match the stored digest"``. Driving the real `flow-execute` path, one run per WORD:

```
LEAK      WORD="did"       <Buffer 73 65 63 2d 39 64 33 66 2d 54 4f 50 53 45 43 52 45 54> did not match the stored digest
LEAK      WORD="expected"  <Buffer 73 65 63 2d 39 64 33 66 2d 54 4f 50 53 45 43 52 45 54> expected not match the stored digest
LEAK      WORD="and"       <Buffer 73 65 63 2d 39 64 33 66 2d 54 4f 50 53 45 43 52 45 54> and not match the stored digest
LEAK      WORD="from"      <Buffer 73 65 63 2d 39 64 33 66 2d 54 4f 50 53 45 43 52 45 54> from not match the stored digest
redacted  WORD="failed"    <Buffer {{secret:API_KEY}}> failed not match the stored digest
redacted  WORD="to"        <Buffer {{secret:API_KEY}}> to not match the stored digest
```

`73 65 63 2d 39 64 33 66 2d 54 4f 50 53 45 43 52 45 54` decodes to `sec-9d3f-TOPSECRET`, the resolved secret. This is the case `flow-yaml.mdx` states as "Argent replaces each value that a `{{secret:NAME}}` resolved to with that same placeholder, in the message and the stack of the failure", and the shape `repairByteRenderings`' own doc block names as covered. Nothing pins this pass: every `Buffer` in `flow-script-redaction.test.ts` is a `.toString("base64"/"hex")`, i.e. `repairEncodedRuns` coverage - no test anywhere under `packages/tool-server/test/flows/script/` produces the numeric `<Buffer 73 6b …>` / `Uint8Array(n) [ … ]` rendering this function exists for, and dropping `repairByteRenderings` from `SCRUB_REPAIRS` (executor.ts:1528) leaves all 1761 tests in the PR's suites green while the full credential prints.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
```diff
@@ -1139,38 +1260,860 @@ function findOwnProtoKey(root: Record<string, unknown>): string | undefined {
...
+
+/**
+ * A value RENDERED AS BYTES, which is neither a spelling of it nor a cut of it.
+ *
+ * `util.inspect` prints a `Buffer` or a `TypedArray` as its NUMBERS, so nothing
+ * a whole-value or per-spelling search looks for is in the text at all — and
+ * the disclosure is total and lossless, character for character. Two shapes
+ * reach a step reason, both from one line a seeding script plausibly writes:
+ *
+ *   assert.deepStrictEqual(Buffer.from(process.env.API_KEY), expected)
```
latekvo, line 1754: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This comment names `assert.deepStrictEqual(Buffer.from(process.env.API_KEY), expected)` as a shape the pass covers, but it is not covered whenever the assert diff interleaves the expected side's byte lines into the actual side. `byteRuns` then builds one run holding the foreign bytes, `byteRunSpans` finds no whole value in it, and `stitchedSpans` only helps when a run's tail plus the next run's head reconstruct the value exactly — which the interleaving breaks. Every actual-side byte still prints, in order.

Ran on a built worktree at 4b1636f, same project and secret as the `byteRuns` thread, with `assert.deepStrictEqual(Buffer.from(process.env.API_KEY, "utf8"), Buffer.from("expected", "utf8"))`. Real step reason:

```
Expected values to be strictly deep-equal:
+ actual - expected

+ Buffer(18) [Uint8Array] [
+   115,
- Buffer(8) [Uint8Array] [
    101,
-   120,
…
+   69,
+   84
  ]
```

Keeping the lines that are not `-` prefixed and decoding them recovers the credential exactly:

```
bytes kept  : [115,101,99,45,57,100,51,102,45,84,79,80,83,69,67,82,69,84]
RECOVERED   : "sec-9d3f-TOPSECRET"
EXACT MATCH : true
```

The gap is easy to miss because the neighbouring shapes all redact correctly — `Buffer.alloc(0)` as the expected value, and `util.inspect` of a `Uint8Array` on its own, both come back as `{{secret:API_KEY}}` in the same harness. It is only the two-non-empty-buffers diff that interleaves.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -134,6 +144,13 @@ const zodSchema = z
...
+          "Put a credential behind `{{secret:<NAME>}}` rather than in the clear: a plaintext value in a tool call enters your context and ~/.argent/mcp-calls.log, which records every call whole. The placeholder is resolved on the machine running the tool-server, from the same sources `keyboard` uses (`ARGENT_SECRET_<NAME>`, the project's `.argent/secrets.env`, its `.env.local`/`.env` ARGENT_SECRET_-prefixed keys, then `~/.argent/secrets.env`). A name no source defines refuses THIS CALL before the run starts, unlike the same name in a flow file's own `env:`, which errors the step that reads it: this map is one argument rather than one flow's own text, so a run over a directory stops at the first flow instead of repeating the refusal once per file. " +
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This description says the placeholder resolves "from the same sources `keyboard` uses", but after this change they are not the same sources. `resolveScriptEnvSecrets` is called with `{ cwd: projectRoot }` (here, and in `flow-script-step.ts`), while `keyboard` (`tools/keyboard/index.ts:200`) and `paste` (`tools/paste/index.ts:76`) still call `resolveSecretPlaceholders(params.text)` with no options, so they resolve the project scope against the tool-server's own `process.cwd()`. Before this change every channel used the same anchor; now one run can resolve one name two different ways.

That is the case this PR's own prose calls the motivating one — `configuration-core/src/secrets.ts` now reads "an editor sets it to `/` or `$HOME`", and `flow-yaml.mdx` says the tool-server cwd is "usually your editor, and often `/` or your home directory". Ran exactly that, against `packages/tool-server/dist/utils/secrets.js` at 4b1636f, with `process.chdir("/")` and a project holding `.argent/secrets.env`:

```
tool-server cwd = /
project root    = …/e2eproj   (holds .argent/secrets.env with ARGENT_SECRET_API_KEY)

script env  ({ cwd: project_root }) -> RESOLVED, 1 secret(s)
keyboard/paste (no cwd)             -> FAILED: Unknown secret "API_KEY" — no source on the
                                       machine running the tool-server defines API_KEY.
                                       Available secrets: (none).
```

So a flow using `{{secret:APP_PASSWORD}}` in both a `script:` step's `env` and a `type:` step has the script step succeed and the type step fail on that host. The docs' single Secrets table (`configuration.mdx`) says "`<project-root>`" for both without saying which root, even though the same PR spells that distinction out explicitly for `scripts.env.allow` a few lines above it.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
```diff
@@ -1565,6 +2522,48 @@ export function buildChildEnv(overrides: Record<string, string> | undefined): No
...
+ * Why the process would not start, with the one refusal that names nothing on
+ * its own spelled out.
+ *
+ * The operating system caps the block of arguments and environment a new
+ * process is handed (`ARG_MAX`), and Node reports the refusal as a bare
+ * `spawn E2BIG`. Every other environment failure in this file is diagnosed by
+ * name; this one would point the author nowhere, and an `env` map is the one
+ * part of that block a flow controls.
+ */
+function spawnFailureMessage(err: unknown, env: NodeJS.ProcessEnv): string {
```
latekvo, line 2806: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This ARG_MAX diagnosis only reaches a `.mjs` step. A `.sh` step touches the child environment earlier, in the bash version probe inside `resolveBashInterpreter`, and that probe's own spawn is the one refused with `E2BIG`. Every candidate then fails the probe, the "no bash this host offers" message is returned instead, and the step's reason condemns the host's bash and points the author at `scripts.bash` — a different subsystem, and one that is working.

Ran both on the same built worktree at 4b1636f, same `env` map (`flow-execute` `env: { BIGVALUE: "x".repeat(3 * 1024 * 1024) }`), one flow per interpreter:

```
.sh   status=error  reason="No bash this host offers could run the script: the executor looked at
      PATH, /bin/bash and /usr/bin/bash, and refused what it found — /opt/homebrew/bin/bash is not
      a bash: running it printed no $BASH_VERSION, so a `.sh` step would report the document it was
      seeded with rather than the one the script writes (spawn E2BIG); /bin/bash is not a bash: …
      Fix the candidate above, or set scripts.bash to the absolute path of a bash that answers."

.mjs  status=error  reason="Could not start the script process: spawn E2BIG — the environment it
      would carry is 3146901 bytes, past this operating system's limit for one process (ARG_MAX).
      Shorten the `env` values, or write the payload to a file and pass its path."
```

`env` is the channel this PR adds, so it is also the first thing that can make an environment big enough to hit this. The size note at the `!startedSeen` branch is not reached either, because the probe returns before the fork. I did check that it does not persist: a healthy `.sh` step run after the oversized one in the same process still passes, so the probe result is not cached across steps.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -3567,18 +3700,58 @@ export function validateFlow(flow: FlowFile): void {
...
+  // a second top-level node at column 1 and the file stopped parsing AT ALL:
+  // not the `env:` block, the whole thing, for every caller of this function.
+  // The paste artefact the line above exists for is the same artefact that
+  // lands there.
+  //
+  // Nothing at the leading edge can be part of a value either, because the top
+  // level of a flow file is a map, so that trim costs nothing and keeps what it
+  // always covered: a file whose first line opens with a TAB, which YAML refuses
+  // as indentation and this accepted before.
+  const body = content.replace(/^\s+/, "").replace(/\n\s+$/, "\n");
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: The trailing half of this trim only strips a whitespace run that FOLLOWS a line break, so a file with no final newline whose last character is whitespace keeps that character — where `content.trim()` removed it. This applies to every flow file, not only ones that use `env`.

Compared the two readings directly against `packages/tool-server/dist/tools/flows/flow-utils.js` at 4b1636f, using `parseFlow(content)` (this change) versus `parseFlow(content.trim())` (what the line did before), on an `echo` step so the field exists on both sides:

```
echo, CRLF block style, final LF lost
   base (content.trim()) : {"kind":"echo","message":"hello"}
   this change           : {"kind":"echo","message":"hello\r"}

echo, CRLF flow style, final LF lost
   base (content.trim()) : {"kind":"echo","message":"hello"}
   this change           : THROWS: Invalid flow file: Unexpected scalar at node end at line 2, …

echo, trailing U+00A0, no final newline
   base (content.trim()) : {"kind":"echo","message":"hello"}
   this change           : {"kind":"echo","message":"hello "}
```

The same on a `script` step's own map: `…\n        TOK: abc\r` (no final newline) parses to `env: { TOK: "abc\r" }`, and the flow-style spelling of the same file stops parsing altogether. A CRLF-authored flow that lost its final LF is the ordinary way to reach this. The test added beside the change (`flow-utils.test.ts`, "still parses a file whose last line holds nothing but a stray whitespace character") only exercises the stray AFTER a `\n`, which is the half that still works.

packages/tool-server/src/tools/flows/flow-add-script.ts
```diff
@@ -232,17 +387,19 @@ export const flowAddScriptTool: ToolDefinition<z.infer<typeof zodSchema>, FlowAd
...
-      // did nothing wrong, and never name the edit that has to be undone.
-      const refusedAnEarlierStep = getFailureSignal(err)?.failure_stage === "flow_output_reference";
+      // scan that refuses an output reference judges what is already in the file
+      // as well — a step from an earlier call, or the file's own top-level
+      // `env:`, both of which a mid-recording hand edit can put there. The step
+      // just run is not among them: its own `env` is refused before the script
+      // starts, and a `script` step spells no other field that scan reads.
+      // Saying "recording it failed" would send the author back over the one
+      // call that did nothing wrong, and never name the edit to undo.
+      const refusedTheFile = getFailureSignal(err)?.failure_stage === "flow_output_reference";
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This append error path still tests only `flow_output_reference`, while the same PR taught both sibling recorders to also recognise `flow_file_parse` / `flow_file_parse_step` as "already in the file, not in this call" — `flow-add-step.ts` (`const fromTheFile = stage === "flow_file_parse" || stage === "flow_file_parse_step"`) and `flow-insert-echo.ts` (the same two lines). This is the third recorder of the three, and the one where the message costs most: the script has already run and nothing is rolled back.

The window is real and this file already accounts for it — `flowEnvOnDisk` reads the file before the run, the append re-reads it after, and the `envDrifted` check below exists because "the file's `env:` was read before a run that may have taken minutes".

Ran it on a built worktree at 4b1636f: `flow-start-recording`, then `flow-add-script` on a 4-second `.mjs`, with an external hand edit adding a top-level `env: { NODE_OPTIONS: … }` to `rec.yaml` 1.5 s into the run:

```
[external hand edit applied to rec.yaml at t=1.5s]

THROWN MESSAGE:
Script "../../scripts/slow.mjs" passed, but the step was not recorded. Check the script's changes
before you retry. Invalid flow file: `env` holds NODE_OPTIONS, which steers the runner's own
process and cannot be set for a script (reserved names: …)
```

The script did nothing wrong and has already had its effect, and "Check the script's changes before you retry" is the sentence both siblings were changed to stop producing for this class.

packages/tool-server/src/tools/flows/script/flow-script-env.ts
```diff
@@ -0,0 +1,321 @@
...
+
+/**
+ * The zod shape of a tool's `env` parameter.
+ *
+ * `z.record` REBUILDS the map, and a JSON body can carry `__proto__` as an own
+ * property — `JSON.parse` puts it there without invoking the accessor — so the
+ * rebuild dropped it before any rule of argent ran and the call passed with
+ * that entry silently gone. Both descriptions said so, and one CLI channel
+ * refused the same name outright: `argent flow run --env __proto__=v` exits 2
+ * with a paragraph, while `--env-json` reached this schema and exited 0
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This rationale contrasts `argent flow run --env __proto__=v` with `--env-json`, but there is no `--env-json` flag. `RUN_OPTIONS` in `packages/argent-cli/src/flow.ts` gains only `"env": { kind: "values" }`, and grepping the repo for `env-json` finds this comment and its own `dist` copy and nothing else.

The guard itself is right and the paragraph names the real channel two sentences earlier (a JSON tool-call body, where `JSON.parse` puts `__proto__` on as an own property). It is the concrete example that does not exist — which is the part a maintainer would reach for to re-test the guard.

packages/tool-server/src/tools/flows/script/flow-script-env.ts
```diff
@@ -0,0 +1,321 @@
...
+ * drifted and to delete a step whose script may already have had its effect.
+ *
+ * Read at call time, like the four other copies of this rule in this module
+ * family, so a test can fake the platform around it.
+ */
+export function envNameKey(name: string): string {
+  return process.platform === "win32" ? name.toLowerCase() : name;
+}
+
+export function mergeScriptEnv(
```
latekvo, line 245: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: `mergeScriptEnv` has no doc comment. The block describing the four layers opens above it, but a second block (for `envNameKey`) opens immediately after it, and the next declaration is `envNameKey` — so both comments attach to `envNameKey` and `mergeScriptEnv` is declared bare.

The module header at the top of this file says "the precedence the runner builds the child environment in is written down once, in {@link mergeScriptEnv}", so the one symbol a reader is sent to for the precedence rule is the one that documents nothing. The generated `.d.ts` shows the same stacking: both comments sit above `envNameKey`, and `export declare function mergeScriptEnv(…)` follows with none.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -3109,6 +3160,13 @@ function* outputReferenceFields(step: FlowStep): Generator<StepField> {
       if (step.cropOn) yield* selectorFields(step.cropOn, "snapshot.cropOn");
       return;
     case "script":
+      // An `env` value is where a `{{output:` reference will belong once PR 4
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: "once PR 4 lands" dates a permanent source comment to a position in a stacked-PR sequence. After the stack merges there is no "PR 4" a reader can resolve, and the sentence stops explaining why `script.env` is scanned here. The rest of the comment already carries the reason on its own.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -3539,20 +3609,83 @@ function fromYamlStep(raw: YamlStep, blockDepth = 0): FlowStep {
...
   if (flow.executionPrerequisite) doc.executionPrerequisite = flow.executionPrerequisite;
   // blockQuote: false — a block scalar is not round-trip-safe for our free-text
   // fields: whitespace-only lines inside a multi-line value are silently
-  // stripped on re-parse (" \n" comes back as "\n"), and a block scalar at the
-  // document tail exposes its raw last line to parseFlow's content.trim(), so
-  // parseFlow(serializeFlow(x)) was not the identity. Disabling it emits
+  // stripped on re-parse (" \n" comes back as "\n"), and a block scalar's own
+  // chomping decides what its last line keeps, so a value at the document tail
+  // came back changed. Either way parseFlow(serializeFlow(x)) was not the
+  // identity. (The second half used to name `parseFlow`'s `content.trim()`;
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This parenthetical describes how the comment itself changed rather than what the code does. "The second half" refers to a sentence in the pre-change version of this block that a reader with the diff hidden cannot see, so it has no referent, and "used to name" / "now trims" narrate the transition rather than the state.

The same block closes with "a long `echo` message or `executionPrerequisite` is no longer wrapped at 80 columns" — "no longer" reads against a previous behaviour that is not in the file either.

.github/workflows/windows-e2e.yml
```diff
@@ -61,8 +75,32 @@ on:
...
+      - "packages/tool-server/test/flows/script/flow-script-env.test.ts"
+      - "packages/tool-server/test/flows/script/flow-script-environment.test.ts"
+      # The redaction cases. A stack frame is a `file://` URL that the reason
+      # decodes back into a path, and a Windows path is where the two spellings
+      # differ most — a drive letter and backslashes, percent-encoded on the way
+      # in. The bash cases here are guarded the same way the file above them is.
+      - "packages/tool-server/test/flows/script/flow-script-redaction.test.ts"
+      # The tools `flow-script-env.test.ts` drives, for the same reason as the
+      # helpers below: editing one changes what this job asserts without
+      # touching any file listed above it. The env cases there record through
```
latekvo, line 97: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This comment states the rule the surrounding `paths:` entries exist for — "editing one changes what this job asserts without touching any file listed above it" — and lists the three recording tools the env cases record through. `packages/tool-server/test/flows/script/flow-script-env.test.ts`, which this same change adds to the job's test list, also imports `../../../src/tools/flows/flow-add-step`, and that file is in no `paths:` entry in this workflow.

This PR changes `flow-add-step.ts` itself (+106), including the new `run:`-drops-`env` warning that `flow-script-env.test.ts` covers, so a follow-up touching only that file would not run this job and the Windows-specific cases driving it would go unexercised.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
```diff
@@ -1139,38 +1260,860 @@ function findOwnProtoKey(root: Record<string, unknown>): string | undefined {
...
+ * takes one to three digits below 256; hex takes exactly the two a byte is
+ * always written as, so a decimal run is not read as hex by accident.
+ */
+function decodeByteRun(run: ByteRun, radix: number): Buffer | undefined {
+  const codes: number[] = [];
+  for (const { text } of run.tokens) {
+    if (radix === 16 && text.length !== 2) return undefined;
+    if (radix === 10 && text.length > 3) return undefined;
+    const code = parseInt(text, radix);
+    if (!(code >= 0 && code <= 255)) return undefined;
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: A `Buffer`/`TypedArray` rendering of a value longer than 255 bytes prints its own element count immediately before the byte run, and `(`/`)`/`[` are not run terminators here — only an ASCII letter is. The count joins the run, `decodeByteRun` rejects the whole run at decimal radix because that element is above 255, no span is produced, and nothing is replaced.

Ran on a built worktree at 4b1636f, `.argent/secrets.env` holding a 298-byte `ARGENT_SECRET_BIGKEY=sk-live-AAA…`, step `env: { BIGKEY: "{{secret:BIGKEY}}" }`, script throwing ``"digest mismatch " + util.inspect(new Uint8Array(Buffer.from(process.env.BIGKEY)), { maxArrayLength: Infinity })``:

```
placeholder present : false
reason head : digest mismatch Uint8Array(298) [
  115, 107, 45, 108, 105, 118, 101, 45, 65, 65, 65, 65,
   65,  65, 65,  65,  65,  65,  65, 65, 65, 65, 65, 65, …
```

`115, 107, 45, 108, 105, 118, 101, 45, …` decodes to `sk-live-AAA…`, the resolved secret, whole and in order. Any credential over 255 bytes — a PEM block, a service-account JSON, a long JWT, all shapes `env` is documented to carry — takes this path, and the same count appears in the assert-diff spelling (`Buffer(298) [Uint8Array] [`).

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
```diff
@@ -1139,38 +1260,860 @@ function findOwnProtoKey(root: Record<string, unknown>): string | undefined {
...
+        decoded.compare(bytes, 0, bytes.length, at, at + bytes.length) === 0
+    );
+    if (!hit) {
+      at += 1;
+      continue;
+    }
+    spans.push(span(at, hit.bytes.length, hit.name));
+    at += hit.bytes.length;
+  }
+  if (spans.length > 0 || !run.cut) return spans;
```
latekvo, line 2024: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: The cut-repair branch here is gated on `run.cut`, which `byteRuns` sets only when an ellipsis appears in the gap between two number tokens — a foreign one, written by whoever truncated the text. But `redactTruncated` scrubs `text.slice(0, marker.index)`, i.e. the head with argent's own `… [N more characters omitted]` marker already removed, so the head it hands the scrub contains no ellipsis, `cut` is never set on it, and the branch cannot fire on the one cut argent itself makes.

Ran both cuts on the same value and rendering, built worktree at 4b1636f, secret `sk-live-9d3f2a7c41b8e05f6a2d`:

```
### cut by ARGENT's clamp
  [X-RUN] Uint8Array(28) [
  115, 107, 45, 108, 105, 118, 101,  45,
   57, 100, 51, 102,  50,  97,  55,  99,
  … [65 more characters omitted]
  decoded visible bytes: "sk-live-9d3f2a7c"
  placeholder present: false

### cut by NODE (maxArrayLength: 16 — a foreign ellipsis, which survives into the head)
  boom Uint8Array(28) [
  115, 107, 45, 108, 105, 118,
  101,  45, 57, 100,  51, 102,
   50,  97, 55,  {{secret:LIVEKEY}},
  ... 12 more items
] tail
  placeholder present: true
```

Identical rendering, identical 16-byte visible prefix: repaired when Node cut it, reported in the clear when argent cut it. `partialSecretTail` does not cover this either — it compares text-space suffixes against spellings, and a byte rendering is not a spelling.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
```diff
@@ -1139,38 +1260,860 @@ function findOwnProtoKey(root: Record<string, unknown>): string | undefined {
...
+ * with it; over-redacting argent's own text is the lesser fault. Case comes for
+ * free, which is what covers a hex signature printed upper-case.
+ *
+ * Floored at {@link ENCODED_RUN_MIN_BYTES}, unlike every other pass here. A run
+ * of ordinary letters decodes to bytes too, and nothing about it says it was
+ * ever an encoding, so a short value would be found in the noise: four bytes
+ * puts a chance hit past one in four billion per position, and a credential
+ * shorter than that is not one.
+ */
+function repairEncodedRuns(text: string, secrets: readonly FlowScriptSecret[]): string {
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: `repairEncodedRuns` only ever matches a WHOLE value inside a decoded run — unlike `repairQuotedCuts` and `byteRunSpans`, it has no prefix/cut branch. The only cut guard on the path is `partialSecretTail`, which searches the head for a prefix of a SPELLING, and that works only while the encoding is character-local. Base64 is exactly the encoding that is not, which is the reason this pass exists — so the idiom its own doc block names is the one a cut leaves standing.

Ran on a built worktree at 4b1636f, secret `sk-live-9d3f2a7c41b8e05f6a2d`, a `.mjs` throwing ``"x".repeat(8130) + " Basic " + Buffer.from("api:" + k).toString("base64") + " and more text"`` so argent's 8 KiB message ceiling cuts inside the payload:

```
REASON: [X-RUN] Basic YXBpOnNrLWxpdmUtOWQzZjJhN… [33 more characters omitted]

base64 fragment left standing: YXBpOnNrLWxpdmUtOWQzZjJhN
decodes to: "api:sk-live-9d3f2a"
placeholder present: false
```

Control — the same payload short enough not to be clamped, so the value stays whole:

```
CONTROL: Basic YXBp{{secret:LIVEKEY}}= rejected
```

So the pass works on the whole value and leaves a decodable prefix of the credential whenever argent's own ceiling cuts it. The `.sh` channel reaches the same place through the `$ARGENT_REASON` budget.

packages/docs/docs/reference/flow-yaml.mdx
```diff
@@ -355,6 +359,84 @@ Rules for the script file:
...
+
+The map is a plain map. A YAML tag that builds something else — `!!omap` in its sequence form builds a Map, `!!set` a Set, `!!timestamp` a Date — is refused, because Argent reads no entries from any of them and the script would run without every value, silently.
+
+Argent refuses a value that holds a NUL character or an unpaired surrogate, because an environment carries neither. Argent applies the same rule to the value a `{{secret:NAME}}` resolves to.
+
+No recording tool writes the top-level `env:`. `flow-start-recording` resets the file to an empty flow, and it does not keep that header. Write the header into the YAML yourself. Write it again each time you record the flow again.
+
+On Windows, Argent refuses two names of one map that differ only in case. Windows reads them as one variable, so a flow file that parses on macOS is refused there.
+
+`BASH_ENV`, `SHELLOPTS` and the other names that steer bash are **not** refused. They steer the interpreter of the script, not the runner, and an author who sets one meant to.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This sentence blesses `BASH_ENV` as steering "the interpreter of the script, not the runner", but it also steers argent's own bash-version probe. `runOne` hands the child environment to `resolveBashInterpreter`, so a `BASH_ENV` preamble that ends the shell aborts `BASH_PROBE_COMMAND` before it prints its marker; `notBashProblem` then rejects every candidate and the step errors blaming the host's bash installation. The one value in the author's own flow file that caused it is named nowhere in that message.

This PR is also the first thing that can set it - the base `flow-script-step.ts` passes no `env` to a script at all.

Ran on a built worktree at 4b1636f, driving the real `flow-execute` path, flow file `env: { BASH_ENV: <preamble> }` where the preamble is the ordinary `set -e` + a probe shape:

```sh
set -e
export GREETING=hello
command -v some-optional-tool >/dev/null
```

```
ok: false
status: "error"
reason: "No bash this host offers could run the script: the executor looked at PATH, /bin/bash and
  /usr/bin/bash, and refused what it found - /opt/homebrew/bin/bash is not a bash: running it
  printed no \$BASH_VERSION, so a `.sh` step would report the document it was seeded with rather
  than the one the script writes; /bin/bash is not a bash: ... Fix the candidate above, or set
  scripts.bash to the absolute path of a bash that answers."
```

Controls in the same harness, on both `.sh` and `.mjs`: a `BASH_ENV` preamble that only echoes, and `SHELLOPTS=xtrace`, `IFS=x`, `PATH=""`, `POSIXLY_CORRECT=1`, `LD_PRELOAD=/nope.so` all pass. Of the names this paragraph groups together, `BASH_ENV` is the one that reaches argent's own machinery, and `RESERVED_ENV_NAMES` covers the node interpreter's steering variable for exactly the reason given here.

packages/skills/skills/argent-create-flow/references/live-authoring.md
```diff
@@ -30,16 +30,18 @@ command: "gesture-tap"
...
 
-When the user requests a local `.mjs` or `.sh` script, call `flow-add-script` at the point where it must run. Read [Flow YAML: Local scripts](flow-yaml.md#local-scripts) first. If the call fails, check its changes before you retry.
+When the user requests a local `.mjs` or `.sh` script, call `flow-add-script` at the point where it must run. Read [Flow YAML: Local scripts](flow-yaml.md#local-scripts) first. If the call fails, its own message says whether the script ran: `was NOT run and nothing was recorded` (the pre-run read of the flow file, and the refusal of the `env` argument) and `did not run` leave nothing behind; every other wording (`failed`, `may have run`, `passed, but the step was not recorded`) leaves changes to check before you retry.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This decision rule tells an agent that the refusal of the `env` argument reads "was NOT run and nothing was recorded", and that every other wording leaves changes to check. None of the three `env` refusals produces any of the five wordings the rule enumerates: `describeScriptEnvProblem` (flow-add-script.ts:228), `assertNoEnvOutputReferences` (:265) and `scriptEnvParameter`'s `__proto__` preprocess all throw an `InvalidToolInputError` whose message begins "This call's `env` ...". Only the pre-run read of the flow file (:308) says the promised sentence.

Ran the real tools from dist on a built worktree at 4b1636f - `flow-start-recording`, then `flow-add-script` with each refused `env`:

```
### reserved name  -> This call's `env` holds NODE_OPTIONS, which steers the runner's own process
                      and cannot be set for a script (reserved names: ...)
### malformed name -> This call's `env` holds "not a name", which is not an environment variable
                      name - a name starts with a letter or "_" ...
### output ref     -> This call's `env.A` uses unsupported template syntax. Replace it with the
                      literal value the script needs: "{{output:x}}"
```

An agent applying the rule literally puts all three in the "every other wording" bucket and goes looking for device or database changes a call that never spawned a process cannot have made - and the same rule is what is supposed to stop it retrying a side-effecting script.

packages/tool-server/src/tools/flows/flow-finish-recording.ts
```diff
@@ -158,7 +166,7 @@ export const flowFinishRecordingTool: ToolDefinition<
...
...y while the step it judges is still identifiable by its number: hand-editing the .yaml during the recording moves the steps, so those warnings are DROPPED rather than pinned on whichever step inherited the number, and \`message\` says how many were dropped. A step that carries a cross-tree warning was re-probed against the runner's tree: read it before converting that wait to \`await:\`/\`assert:\`, which is what the verdict is about and what this moment is for. A step that recorded a wait which did not pass was never probed at all, and its own warning names the CAUSE, because only one of them judges the condition: an unmet wait was read and found false, and it stops the run at replay; a wait whose tree source could not be read, or one that was cancelled, observed nothing and leaves the condition UNKNOWN rather than known-bad. Read those before replaying. A step that lost its \`env\` is the third kind and is about neither: a recorded \`flow-execute\` becomes a \`run:\` step, which carries no env of its own, so the values that call passed are not in the flow and the replay runs without them — write them into the target flow's own \`env:\`, or into this recording's, before replaying.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This remedy - "write them into the target flow's own `env:`, or into this recording's" - drops the qualification the warning it explains carries. flow-add-step.ts:1119-1121 says "only for a name that fragment does not itself declare, since a fragment's `env:` layers OVER the flow that runs it", and `execRunStep` does exactly that: `childScope(scope, { env: mergeScriptEnv(scope.env, fragment.env) })`.

So for a name the fragment declares, the second remedy silently does nothing: a recorded `flow-execute` passing `env: { BUILD: "1421" }` to a fragment that declares `env: { BUILD: "0" }` gets rewritten to `run:`, the env is dropped, the warning fires, and writing `BUILD: "1421"` into the recording's top-level `env:` still leaves the script reading `BUILD=0` - the outcome the warning exists to prevent. The branch's own pinned test states the precedence ("lets a fragment's own env beat the parent's default for a recorded step", flow-script-env.test.ts:246-268; passes at 4b1636f).

The same unqualified remedy is in `packages/skills/skills/argent-create-flow/references/live-authoring.md:33`.

packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -1102,14 +1103,58 @@ async function captureRunTarget(
           `step would replay a different flow than the one that just ran`,
       };
     }
-    return { flow: `${name}.yaml` };
+    // A `run:` step carries no environment of its own, so whatever this call
+    // passed the sub-run is not part of what was recorded — the one lossy
+    // rewrite in this function that used to say nothing.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: "the one lossy rewrite in this function that used to say nothing" describes how this function changed rather than what it does. A reader on main with the diff hidden is told about a behaviour that is not in the file.

The same in the append `catch` below (~1471-1475): "Every other `env:` fault ... arrives as `flow_file_parse` and reached the agent as a bare 'Invalid flow file', AFTER the device action had already run. An agent reading that has no reason not to retry, and runs the action a second time." Past tense for the pre-change behaviour, sitting directly above the code that now handles it, so it reads as a live hazard in the current file.

packages/tool-server/src/tools/flows/script/flow-script-env.ts
```diff
@@ -0,0 +1,321 @@
...
+  if (!isPlainMap(raw)) {
+    return (
+      `is ${describeTaggedMap(raw)} rather than a plain map, so argent reads no entries from ` +
+      "it — every name would be dropped and the script would run without them, silently. A " +
+      "YAML tag is what builds one; remove it, e.g. " +
+      "`env: { API_URL: https://api.example.com }`"
+    );
+  }
+  for (const [name, value] of Object.entries(raw as Record<string, unknown>)) {
+    if (!SCRIPT_ENV_NAME_PATTERN.test(name)) {
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This pattern test runs before `reservedScriptEnvName`, so `npm_config_node-options` - the spelling npm itself documents - never reaches the reserved branch and is refused as malformed instead. The same module advertises that exact spelling as reserved in every other refusal, and the docs table this PR adds (flow-yaml.mdx:402) lists it with the reason "Each one reaches `NODE_OPTIONS`". The message's own claim is untrue as well: `-` is legal in a POSIX environment name.

Ran against the built dist at 4b1636f:

```
npm_config_node-options => Invalid flow file: `env` holds "npm_config_node-options", which is not
  an environment variable name - a name starts with a letter or "_" and continues with letters,
  digits or "_"
npm_config_node_options => Invalid flow file: `env` holds npm_config_node-options, which steers the
  runner's own process and cannot be set for a script (reserved names: ..., npm_config_node-options,
  npm_config_userconfig, npm_config_globalconfig)
advertised reserved list: ..., npm_config_node-options, npm_config_userconfig, npm_config_globalconfig
```

So the author who writes the documented hyphenated spelling is told the name is invalid, while the underscore spelling produces a message naming the hyphenated one - a name the first message calls not-a-name. `scripts.env.allow` takes the same path (flow-script-executor.ts:2766 tests the pattern first) and drops it into the `malformed` note bucket.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
```diff
@@ -1139,38 +1260,860 @@ function findOwnProtoKey(root: Record<string, unknown>): string | undefined {
...
+ * What runs after the whole-value scrub, in order. Each one answers a rendering
+ * that leaves no spelling of the value in the text: a prefix another process
+ * cut, a rendering of the bytes, an escaper's backslashes, a re-encoding into
+ * another alphabet. Each reads the text the ones before it left, so a value
+ * that two of them rewrote is still taken.
+ */
+const SCRUB_REPAIRS = [
+  repairQuotedCuts,
+  repairByteRenderings,
+  repairBackslashEscapes,
```
latekvo, line 1552: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: `repairBackslashEscapes` is not pinned by anything, including the test written for it. Dropping it from `SCRUB_REPAIRS` in a scratch worktree leaves all 1761 tests in this PR's suites green, and `flow-script-redaction.test.ts`'s own "replaces a value an escaper backslashed" still passes on its own - its fixture value happens to be caught by a different repair in the list.

`printf %q` is one of the three escapers this function's JSDoc names as its reason for existing, and it is the one that then leaks. Secret `Q = "sk live 9d3f0a1bcdef"`, bash step `printf %q "$Q" > "$ARGENT_REASON"; exit 1`:

```
unmutated : ... exited with code 1 (bash: /opt/homebrew/bin/bash). {{secret:Q}}
mutated   : ... exited with code 1 (bash: /opt/homebrew/bin/bash). sk\ live\ 9d3f0a1bcdef
```

So the suite cannot tell whether this line is doing its job.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
```diff
@@ -1139,38 +1260,860 @@ function findOwnProtoKey(root: Record<string, unknown>): string | undefined {
...
+
+/**
+ * Every spelling of every resolved value this report may hold, for one scrub.
+ *
+ * {@link scrubSecretValues} takes them as one list and orders it longest first,
+ * so a value that contains another — or a spelling that contains the value it
+ * came from — is still taken whole.
+ */
+function scriptSecretSpellings(raw: readonly FlowScriptSecret[]): FlowScriptSecret[] {
+  const spellings = withTrimmedSpellings(raw);
```
latekvo, line 1467: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: `withTrimmedSpellings` is not pinned either. Replacing this call with `[...raw]` leaves all 1761 tests green - including `flow-script-redaction.test.ts:110`, "replaces a secret whose own trailing newline the reason trim ate", the test written for exactly this shape (it passes under the mutation, 35/35 in that file).

The function exists because the runner trims `$ARGENT_REASON`, so a secret whose value ends in whitespace never appears whole in the text. Secret `TS = "sk-live-9d3f0a1bcdef "` (trailing space), bash step `printf %s "$TS" > "$ARGENT_REASON"; exit 1`:

```
unmutated : The script exited with code 1 (bash: /opt/homebrew/bin/bash). {{secret:TS}}
mutated   : The script exited with code 1 (bash: /opt/homebrew/bin/bash). sk-live-9d3f0a1bcdef
```

A trailing space or newline is the ordinary shape of a pasted credential or a PEM blob, so this is the case the function was added for and the one the suite does not discriminate.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
```diff
@@ -1613,6 +2630,210 @@ function configuredNumber(key: string): number | undefined {
...
+ *
+ * The whole prefix, not a list of the names known today: the set this must keep
+ * out — the bearer token, the port, every `ARGENT_SECRET_*` value, the telemetry
+ * ingest token a release build exports — is exactly the set that grows without
+ * this file being touched, which is the argument the built-in allowlist is
+ * built on. A script that needs one of these values is asking for argent's own
+ * configuration; pass what it needs under a name of the project's own.
+ */
+function argentOwnedEnvName(name: string): boolean {
+  return name.toUpperCase().startsWith(ARGENT_ENV_PREFIX);
```
latekvo, line 3171: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: The case-fold here is not pinned. Making it case-sensitive (`name.startsWith`) leaves all 1761 tests green, while `scripts.env.allow: ["Argent_Auth_Token"]` then widens the child allowlist for that name - the host value reaches the script and the run reports nothing:

```
unmutated : script saw {"Argent_Auth_Token": null}
            note: scripts.env.allow names Argent_Auth_Token, which argent keeps out of the copy it
                  takes from its own environment; that entry was ignored. ...
mutated   : script saw {"Argent_Auth_Token": "host-token"}
            note: undefined
```

The twin rule in `reservedNameFor` IS pinned - "boots the child as Node when the server's environment carries Electron_Run_As_Node" - so the two case-insensitive rules added by this change are covered asymmetrically. It matters most on Windows, where environment names are case-insensitive and every `ARGENT_*` value is reachable by this spelling; the run above sets the host spelling explicitly to show the path.

packages/tool-server/src/tools/flows/script/flow-script-env.ts
```diff
@@ -0,0 +1,321 @@
...
+ * JSON Schema this parameter publishes is unchanged — `whose` names the map,
+ * as every other `env` refusal does.
+ */
+export function scriptEnvParameter(whose: string) {
+  return z.preprocess(
+    (raw, ctx) => {
+      if (
+        raw !== null &&
+        typeof raw === "object" &&
+        Object.getOwnPropertyNames(raw).includes(PROTO_ENV_NAME)
```
latekvo, line 151: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This preprocess is the only thing that refuses `__proto__` on the two tool channels - `z.record` provably rebuilds the map without the key, which the test file's own comment says (flow-script-env.test.ts:692-696):

```
$ node -e 'const {z}=require("zod");const raw=JSON.parse(String.raw`{"__proto__":"x","A":"y"}`);
           console.log(Object.getOwnPropertyNames(z.record(z.string(),z.string()).parse(raw)))'
[ 'A' ]
```

But both tests in "a tool env parameter that carries `__proto__`" call `createRunFlowTool(registry).execute({}, {...} as never)` and `flowAddScriptTool.execute({}, {...} as never)`, which never runs the schema - so the refusal they observe comes from the `describeScriptEnvProblem` backstop, not from this line. Turning the preprocess into a pass-through leaves all 1761 tests green:

```
unmutated : PARSE REFUSED: This call's `env` holds __proto__, which argent cannot carry: ...
mutated   : PARSE OK, env own props = [ 'A' ]
```

The mutated behaviour - the entry silently gone on a real MCP or HTTP call, whose JSON body is the only path that runs the schema - is exactly the regression this preprocess was added to prevent, and the suite would not catch it.

packages/tool-server/src/tools/flows/script/flow-script-env.ts
```diff
@@ -0,0 +1,321 @@
...
+ * script run with a value silently missing or replaced. None of them is about
+ * what a value holds — an `env` value is ordinary flow data, and argent does
+ * not judge whether one looks like a credential.
+ */
+export function describeScriptEnvProblem(raw: unknown): string | null {
+  // `env:` with nothing under it is a map with no entries, not a malformed one:
+  // every entry commented out is an ordinary authoring state, and `env: {}`
+  // beside it is accepted. Nothing is lost by taking it, which is the question
+  // every other rule in this function answers.
+  if (raw === null) return null;
```
latekvo, line 39: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This guard carries a five-line comment about `env:` with every entry commented out being an ordinary authoring state, but nothing pins it. Deleting the line leaves all 1761 tests green, and `parseFlow` on such a file then throws out of `Object.getPrototypeOf(null)` inside `isPlainMap`:

```
probe: parseFlow("env:\nsteps:\n  - echo: hi\n")
unmutated : parsed, env={}
mutated   : THREW: Cannot convert undefined or null to object
```

A bare `env:` header - written before the values, or with the values commented out - is the shape the comment itself describes, and the failure mode is an unclassified `TypeError` rather than a flow-file refusal, for every caller of `parseFlow`.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -1303,6 +1332,57 @@ Pass exactly one flow source: name for a saved flow under project_root, or flow_
...
+      // mistyped name. Raised as caller input, the run stops at the first flow,
+      // which is what the two checks above it already do for a bad NAME.
+      //
+      // The step resolves again and is still the authority: this reads the
+      // run's own map only, holds nothing it resolved, and passes the same
+      // project anchor the step will. A flow file's own `env:` is a fault in
+      // that FILE and is left to the step, where it stays one flow's problem.
+      if (params.env && Object.keys(params.env).length > 0) {
+        try {
+          resolveScriptEnvSecrets(params.env, { cwd: params.project_root });
```
latekvo, line 1308: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This anchor is not covered. Changing `{ cwd: params.project_root }` to `{}` here leaves all 1761 tests green; the identical mutation on the step-level twin (`flow-script-step.ts:115`) fails 12 tests. Nothing exercises a `{{secret:NAME}}` supplied through `flow-execute`'s `env` or `argent flow run --env`.

The comment on this line claims it passes "the same project anchor the step will", and the anchor is the point of the change - the PR description calls the tool-server cwd "whatever spawned the server, and an editor sets it to `/` or `$HOME`". A probe on that exact host shape, secret defined only in the project's `.argent/secrets.env` and passed through the run-time parameter:

```
unmutated : PASS  the run-time env param resolves a project secret
mutated   : FAIL  This run's env value API_KEY: Unknown secret "API_KEY" - no source on the machine
                  running the tool-server defines API_KEY. Available secrets: (none).
```

So a regression here refuses the whole run up front on the very host the feature targets, and the suite stays green.

packages/tool-server/src/tools/flows/flow-utils.ts
```diff
@@ -3539,20 +3609,83 @@ function fromYamlStep(raw: YamlStep, blockDepth = 0): FlowStep {
...
 /** Serialize a full flow file to YAML, omitting empty/defaulted fields. */
 export function serializeFlow(flow: FlowFile): string {
-  const doc: YamlFlowFile = { steps: flow.steps.map(toYamlStep) };
+  // `env` first, because it is a header the whole file reads under. Emitted
+  // whenever the flow carries one, empty included: `parseFlow` sets `env` from
+  // the KEY's presence, not from its size, so dropping an empty map is what
+  // breaks parseFlow(serializeFlow(x)) — and deletes an `env: {}` an author
+  // wrote as soon as the next step is appended.
+  const doc: YamlFlowFile = {
+    ...(flow.env ? { env: { ...flow.env } } : {}),
```
latekvo, line 3618: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: Only the STEP's empty `env` map is pinned; the file-level one is not. Requiring a non-empty map in this spread leaves all 1761 tests green - including "round-trips a flow whose env holds an empty map and a multi-line value", which asserts only the step's `env: {}` - while `parseFlow(serializeFlow(x))` loses the file's own key:

```
unmutated : "env: {}\nsteps:\n  - script:\n      path: seed.mjs\n      env: {}\n"
mutated   : "steps:\n  - script:\n      path: seed.mjs\n      env: {}\n"
```

The same mutation applied to the step-level `if (step.env)` in `toYamlStep` fails that test immediately, so the two halves of one round-trip rule are covered asymmetrically. The reachable consequence is a top-level `env: {}` disappearing from the author's file on the next recorder append.

packages/argent-cli/src/flow.ts
```diff
@@ -1080,7 +1150,16 @@ async function runFlowDirectory(
...
+      // A refusal about the CALL is not a refusal about this file. Every flow
+      // in a batch is invoked with the same arguments apart from its own path
+      // — `--env` most of all — so a tool-input refusal repeats verbatim for
+      // each of them: `--env NODE_OPTIONS=x` over eleven flows printed one
+      // message eleven times and ended `0 passed, 11 failed`, with nothing to
+      // say the fault was one argument rather than eleven files. That stops the
+      // batch, like every other failure the server did not tie to this flow.
+      const rejectedThisFlowOnly =
+        toolErr?.errorKind === "validation" &&
+        toolErr.errorCode !== FAILURE_CODES.TOOL_INPUT_INVALID;
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: Excluding `TOOL_INPUT_INVALID` here is what stops the batch on a call-level refusal, but `rejectedThisFlowOnly` drives a second decision as well - the stdout verdict at :1172 - so the same refusal now prints differently in the two runners. The single-flow path (:1548) still keys its verdict on `errorKind === "validation"` alone.

Ran both against a real tool-server from the built branch, temp HOME, a project holding three one-step flows:

```
$ argent flow run one --env NODE_OPTIONS=x
Flow "one"
  x not run (rejected)

$ argent flow run .argent/flows --env NODE_OPTIONS=x
[1/3] one.yaml
  x did not finish (run error)
[2/3] three.yaml
  . not run (batch stopped)
[3/3] two.yaml
  . not run (batch stopped)
```

Same for `--env S={{secret:NOPE}}`. The comment just above this line says "Verdict before detail, as the single-flow runner prints them, so a merged log reads the same way", and the single-flow runner's own comment says "The same stdout verdict a directory run gives every entry" - both now describe something the code does not do.

"did not finish (run error)" is also not what happened: `flow-execute` throws the env refusal before `resolveFlowSource`, so nothing ran. It is the one remaining line that pins a bad `--env` on `one.yaml`, which is the misattribution the surrounding comment says this change was written to remove. A per-file validation refusal still routes through `rejectionVerdict` in the batch ("not run (invalid flow)"), so the two vocabularies disagree only on the class this change introduces.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
```diff
@@ -1613,6 +2630,210 @@ function configuredNumber(key: string): number | undefined {
...
+    }
+    const raw =
+      typeof document === "object" && document !== null && !Array.isArray(document)
+        ? getAtPath(document as Record<string, unknown>, SCRIPT_ENV_ALLOW_KEY)
+        : undefined;
+    if (Array.isArray(raw)) {
+      // Trimmed, because `asStringArray` trims before the names are judged, so
+      // an untrimmed key would never be found again.
+      for (const entry of raw) {
+        if (typeof entry !== "string" || entry.trim() === "") continue;
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: This `continue` drops a non-string or blank `scripts.env.allow` entry in exactly the silence the four notes around it exist to end - the raw array is in hand here, so the name could be reported.

The `scripts.env.allow` schema description added by this change gives a reason that this file falsifies: "An entry that is not a string, or a blank one, goes when the file is read, before the run can name it" (config-schema.ts:256). The run re-reads the raw JSON itself at :2686-2715 - that is the only reason the `is not valid JSON` and `is not a list` notes can exist - so the entry is present at this line when it is discarded.

Ran four allow lists on the built branch, one flow and one `.mjs` step each:

```
["DB_URL", ["AWS_PROFILE","AWS_REGION"]]  ok=true   reason: (none)
["DB_URL", null]                          ok=true   reason: (none)
["DB_URL", 42]                            ok=true   reason: (none)
["9LIVES", 42, null]                      ok=true   reason: scripts.env.allow names "9LIVES", which
    is not an environment variable name - a name starts with a letter or "_" and continues with
    letters, digits or "_". That entry was ignored.
```

The last is the discriminator: a malformed *string* is named while the `42` and the `null` beside it are not. In the first case `AWS_PROFILE` and `AWS_REGION` never reach the script and the run says nothing - and a nested array is the natural mis-grouping for that pair. For contrast, from the same loop: `allow: "DATABASE_URL"` gives "is not a list, so argent read no names from it and the script ran without them", and a trailing comma gives "is not valid JSON ... read nothing from it at all".

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
hubgan: Fixed in 4f461f3c4

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
hubgan: Fixed in 4f461f3c4

Same root cause as the `<Buffer ...>` thread, so one commit answers both: a number that is no byte at the radix now ends the run and is dropped rather than sinking it.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
hubgan: Fixed in 1839fb9ec

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
hubgan: Fixed in caf7803e5

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
hubgan: Fixed in ab8319b67

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
hubgan: Fixed in ca761d68d

packages/tool-server/src/tools/flows/script/flow-script-env.ts
hubgan: Fixed in 70a222396

packages/tool-server/src/tools/flows/script/flow-script-env.ts
hubgan: Fixed in 2d2f7646c

packages/tool-server/src/tools/flows/flow-run.ts
hubgan: Fixed in 2ba97f7d2

packages/tool-server/src/tools/flows/flow-utils.ts
hubgan: Fixed in d7c64ecde

packages/tool-server/src/tools/flows/script/flow-script-env.ts
hubgan: Fixed in 8d221b94c

packages/tool-server/src/tools/flows/script/flow-script-env.ts
hubgan: Fixed in 2ee65f791

packages/tool-server/src/tools/flows/flow-utils.ts
hubgan: Fixed in 13a87263d

packages/tool-server/src/tools/flows/flow-utils.ts
hubgan: Fixed in 450e2c8bd

packages/tool-server/src/tools/flows/flow-add-step.ts
hubgan: Fixed in e09d277a3

packages/tool-server/src/tools/flows/flow-utils.ts
hubgan: Fixed in 43f405cf3

.github/workflows/windows-e2e.yml
hubgan: Fixed in 9f6a0ff48

packages/tool-server/src/tools/flows/flow-add-script.ts
hubgan: Fixed in b0c6c05f5

packages/tool-server/src/tools/flows/flow-finish-recording.ts
hubgan: Fixed in f0500acd4

packages/skills/skills/argent-create-flow/references/live-authoring.md
hubgan: Fixed in 2f442fc14

packages/tool-server/src/tools/flows/script/flow-script-env.ts
hubgan: Fixed in 7f3257b44

The ordering is fixed on both channels. The message's second claim - that `-` is not legal in an environment name - is left as it stands: the same sentence is in `argent-cli/src/flow.ts`, which cannot import the reserved list, and the three channels are kept word-for-word identical.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
hubgan: Fixed in 14991ad28

The note names the entry and points at the nested-list mis-grouping. The `scripts.env.allow` schema description and the configuration reference, which both claimed the run could not name it, are corrected in the same commit.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
hubgan: Fixed in 8c9b03d7f

Two further rules the flag needed: the cut is tried at both `decoded.length` and `decoded.length - 1`, because a character ceiling can stop mid-number where a renderer stops between elements; and the prefix is floored at `CUT_MIN_PREFIX_CHARS`, because a one-byte match grew the message past the child's ceiling and the placeholder itself came back cut.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
hubgan: Fixed in 16daadeeb

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
hubgan: Fixed in 5ad553482

packages/docs/docs/reference/flow-yaml.mdx
hubgan: Fixed in f7d40c4d7

Fixed by taking the four startup-steering names out of the PROBE's environment only, so the reference's claim stays true for the step. The reference now says the probe ignores them.

packages/tool-server/src/tools/flows/flow-run.ts
hubgan: Fixed in 2fed9c008

The behaviour itself is older than this stack - both device tools are byte-identical to the base and `secretSources` is unchanged - so what this commit fixes is the claim: the parameter and the configuration reference now say which root each channel means, and what makes every channel agree. Closing the asymmetry needs a `project_root` on `keyboard`/`paste` or an `ARGENT_PROJECT_ROOT` fallback in the resolver, which is outside this PR.

packages/argent-cli/src/flow.ts
hubgan: Fixed in a8d008a94

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
hubgan: Follow-up in 6b6d415ef: the drop was placed before the gap test, so a token that is no byte closed the run with `cut: false` and threw away an ellipsis that had just closed it with `cut: true`. The token after an ellipsis is a COUNT, which is no byte whenever it has the wrong number of digits - so `util.inspect`'s own `<Buffer … ... N more bytes>` left the first fifty bytes of any credential over fifty standing. Reading the gap first restores that, and also closes the `... N more items` shape at Node's default `maxArrayLength`, which the base leaked for every value over 260 bytes.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
hubgan: Two follow-ups in 4fdf59ed3: the cut search returned on the first NEEDLE that answered at any length rather than the longest prefix across all of them, and the `CUT_MIN_PREFIX_CHARS` floor now applies only to argent's own cut - a renderer's ellipsis is nowhere near the ceiling, and flooring it there left a byte standing the base branch took. Swept against the pre-PR base across renderings, value lengths 6 to 1600 and clamped/unclamped: better in 65 cases, worse in none.

packages/tool-server/src/tools/flows/flow-utils.ts
hubgan: Follow-up in 611e4914c: the rule is now `\r+$`, since the same conversion run twice ends a file in two; and the comment's justification was wrong - this parser keeps a raw CR as content. What makes the trim safe is that `serializeFlow` escapes a CR inside a value and ends the document with a newline, so no file argent writes can end in one.

packages/tool-server/src/tools/flows/script/flow-script-env.ts
hubgan: Follow-up in 46d135f73: `argent flow run --env` was a third site asking this question, with its own copy of the name pattern and no reserved check, so it still answered "not an environment variable name" for npm's hyphenated spelling. The table now lives in `configuration-core` beside `SCRIPT_ENV_NAME_PATTERN` and the CLI reads it, so all three spellings give one answer.

packages/tool-server/src/tools/flows/script/flow-script-executor.ts
hubgan: Follow-up in 6b6d415ef: the drop was placed before the gap test, so a token that is no byte closed the run with `cut: false` and threw away an ellipsis that had just closed it with `cut: true`. The token after an ellipsis is a COUNT, which is no byte whenever it has the wrong number of digits - so `util.inspect`'s own `<Buffer … ... N more bytes>` left the first fifty bytes of any credential over fifty standing. Reading the gap first restores that, and also closes the `... N more items` shape at Node's default `maxArrayLength`, which the base leaked for every value over 260 bytes.

packages/tool-server/test/windows-e2e-paths.test.ts
```diff
@@ -106,15 +110,30 @@ describe("the Windows job's path filter", () => {
     // line so the two lists cannot drift apart. The seven other files that
     // line names reach most of the tool server between them, and covering
     // those is not this rule's job — the filter never claimed them.
```
latekvo, line 112: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]: This pin no longer matches the workflow it reads, so the test fails at this branch's head. The `vitest run` line of `windows-e2e.yml` names six `test/flows/script/*.test.ts` files here (`flow-script-interpreter`, `flow-script-bash`, `flow-script-bash-sweep`, `flow-script-env`, `flow-script-environment`, `flow-script-redaction` — the last three were added by this branch's commits), while the assertion below still pins three.

Reproduced at `71fca6ec` on a clean checkout of this branch, Ubuntu, GNU bash 5.3:

```
npx vitest run test/windows-e2e-paths.test.ts
  × names every file the .sh tests it runs import
  AssertionError: expected [ …(6) ] to have a length of 3 but got 6
  Test Files  1 failed (1)
  Tests  1 failed | 1 passed (2)
```

The workflow on the stacked base names three seed files, where the pin passes; the additions that pushed it to six are all from this PR. The file carries no platform skip, so `npm test --workspaces --if-present` — the command the `Unit Tests` job runs on every PR — reports the failure on any host. The graph-coverage assertion in the same test still holds; only the count is stale, but the count is what fails the suite.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]: Reviewed at `71fca6ec5cd12cb8706f00b79a3510ae27281611` against its base, in a clean clone of the branch (`npm ci`, `npm run build` clean). I went through the claims vs code, the nearest twins of every new rule, the non-happy paths, the inputs, the reachability of each new branch in both directions, and what each call leaves behind; then a pass for absences against siblings, prose and symmetry, including the previously-raised threads re-checked at this head. Ran the real entry points end to end where a finding needed it: `describeScriptEnvProblem` / `reservedScriptEnvName` / `parseRunArgs` against the built `dist`, `parseFlow` / `serializeFlow` round trips, `FlowScriptExecutor.execute` for the bash paths, and the full `test/flows` suite (2348 tests). One finding below; everything else I probed held up, including all twenty-eight earlier threads, which I re-verified against this head rather than assuming.

PR #1062, opened 2026-09-08
fix(flows): run a cross-platform launch map on a remote simulator

hubgan:
## The problem

A flow that names its app per platform could not start it on a remote (cloud) simulator:

```yaml
steps:
  - launch: { ios: com.acme.app, android: com.acme.app.android }
```

```
no app id declared for platform "ios-remote" - add a launch entry for it
```

The advice could not be followed. `ios-remote` is not a launch-map key, so adding one fails
parsing as a misspelled platform. The only way out was to rewrite the flow for the cloud.

## The fix

A platform string answers two different questions, and they had been sharing one answer:

- **Which machine am I driving?** `ios-remote`. Device selection, service refs, transports.
- **What did the flow author write?** `ios`. Launch maps, `when:` guards.

A remote simulator is an iOS simulator on another Mac: same OS, same app, same UI. A flow file
names neither a host nor a device, so it must not need an edit to run in the cloud.

This adds one named helper, `authoringPlatform()`, folding `ios-remote` to `ios`, and applies it
where the author's own words are read:

- the launch map now resolves the `ios` entry on a remote run;
- the "no app id declared" error now names `ios`, a key you can actually write.

The fold already existed, inlined in one place - the `when: { platform: ios }` guard. That call
site now uses the helper, with identical behaviour, so the rule has one spelling instead of two.

`ios-remote` stays unwritable in a flow file. It selects a machine; it is not something a flow
can name.

### The recorder's advice

The recorder's post-step advice reads the platform too, to name which two trees disagree and
how to retarget a selector. Read raw, `ios-remote` matched no clause, so a remote simulator got
"no read-only tool is known to report the runner's projection on this platform" - wrong, since
it reads the same iOS trees a local sim does. That advice now folds as well.

One clause there is **not** an authoring question and is guarded instead. An indeterminate
verdict means the tree source did not answer, and the remedy depends on which silence it was:
a source that is DOWN can be relaunched, one that does not exist cannot. `supportsFlowTree`
already draws exactly that line, so the repair advice now turns on it rather than on the
platform name. Today that keeps "relaunch the app" off a remote simulator, which has no flow
tree source yet; when it gains one, the advice follows with no further edit. Since #721 the iOS
remedy itself travels in the reason `queryFullHierarchyTree` writes rather than in the
recorder's caveat, so what the guard now keeps off a remote simulator is the caveat's note
about `bundleId`.

The same line picks the sentence that closes the warning. A source that failed is told to
re-probe once that source is back. A device whose platform has no tree source at all is told
that its runner has no tree to probe, so the conversion stays unverified, instead of being
sent to wait for a source that cannot come back.

## Verification

End-to-end against a booted cloud simulator (`remote:73A22194-…`, iPhone 17 Pro / iOS 26.5) driven through the branch tool-server, with the test app installed on it:

| Flow | Result |
| --- | --- |
| `launch: { ios: …, android: … }` | passes, and a screenshot confirms the app is up. The `android` value names an id that is not installed, so the `ios` key is what resolved |
| `launch: { native: … }` | passes |
| `when: { platform: ios }` | `condition met (platform ios)` |
| map with no `ios` key | `no app id declared for platform "ios"` |
| `launch: { ios-remote: … }` | still a parse error, listing the allowed keys |

On the parent commit the first row stops on step 1 with `no app id declared for platform "ios-remote"`.

The recorder's advice stays covered by unit tests rather than a live run: it needs a probe against a real remote tree, which does not exist until that source lands. Both new cases were checked by reverting their production line and confirming the test fails.

`npm run build`, `npm run lint`, `npm test --workspaces --if-present` and the docs build all pass.

## Docs

Two sentences in the flow-yaml reference, naming which launch-map key a remote run uses and
that it falls back to `native`, and one clause on the `when:` guard. The same file already documents `--platform ios-remote` as a run
target while both key lists exclude it, so a reader comparing them has a question this answers.

The `argent-create-flow` skill is deliberately untouched. Authoring a flow against a remote
simulator does not work yet - selectors, idle waits and snapshots all still fail there - so a
skill that walked an agent into it would break on the step after the launch.


packages/tool-server/src/tools/flows/flow-add-step.ts
```diff
@@ -381,6 +399,10 @@ function unmetWaitWarningFor(cause: UnmetUiWaitCause): string {
 // than honour it.
 function indeterminateReasonCaveat(udid: unknown): string {
   if (platformOf(udid) !== "ios") return "";
+  // Every remedy below repairs a source that is down. A platform with no flow
+  // tree source at all is not down: no relaunch can produce a tree there, and
+  // "once that tree source is back" is nonsense for one that never left.
+  if (!hasRunnerTree(udid)) return "";
```
latekvo, line 415: [[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: The `hasRunnerTree` gate added in `indeterminateReasonCaveat` suppresses only the caveat tail — this base sentence is still emitted ungated, so on `ios-remote` the author is told to wait for a source to come "back" when it never existed. This is exactly the case the gate's own rationale names: the comment above it ("A timeout and an outage need different next moves: 'once that tree source is back' is nonsense for a source that never left") and the `hasRunnerTree` docstring ("a source that is DOWN can be brought back, one that does not exist cannot") both describe the fix, but the gate stops one clause short of the sentence it motivates.

Reproduction: with the recorder running against a `remote:` device, a step whose runner-tree probe hits the not-supported throw records with the message ending:

`...UNKNOWN, not known-bad — re-probe once that tree source is back before trusting the conversion`

(captured live through `flow-start-recording` → `flow-add-step` on the built dist). The PR's own new test pins the sibling clause — its "offers no repair on a platform whose tree source does not exist" case asserts the `bundleId`/`launch-app` remedies are gone — but its comment ("Still recorded, still honestly labelled UNKNOWN — just not repaired") is falsified by this sentence still being present on the same path: the step is still told to repair. The mutation test confirms the test's negative assertions are genuinely sensitive to the gate line (removing it fails the test on `no directive takes a bundleId`), so the sentence above simply sits outside what the gate guards.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -2113,11 +2116,11 @@ async function execWhenStep(
...
-    // spelling, so without this fold iOS-only blocks would silently skip there.
+    // A guard names an authoring platform ({@link authoringPlatform}): the
+    // parser rejects "ios-remote" as a guard spelling, so without the fold
+    // every iOS-only block would silently skip on a remote simulator.
     const guardEnv = deviceEnv(state);
-    const platform = guardEnv.device.platform === "ios-remote" ? "ios" : guardEnv.device.platform;
+    const platform = authoringPlatform(guardEnv.device.platform);
     met = platform === step.condition.platform;
   } else {
     const probe = await probeWhenCondition(deviceEnv(state), step.condition);
```
latekvo, line 2063: [[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: On `ios-remote`, a UI guard (`visible`/`exists`/`text`) now errors and stops the whole run, and the reason quotes `ui-tree matching is not supported on platform "ios-remote"` — the unwritable spelling, in the same run path where this PR folds the platform arm of the guard so `ios` works. Reproduced end-to-end through the real `flow-execute` path with the docs' own example:

```
steps:
  - when:
      visible: "Got it"
    steps:
      - tap: { text: "Got it" }
```

on device `remote:…` → `when:error`, `tap:skip`, `ok:false`, reason: `could not evaluate when guard (visible text="Got it"): could not read the UI tree: ui-tree matching is not supported on platform "ios-remote"`.

The message is honest about the machine, but the fold exists precisely so author-facing text names a key the author can act on, and the run-stop itself sits next to the new docs line (flow-yaml.mdx:239, "A run on a remote simulator matches `ios`") without any warning that the other half of the guard stops the run there.

latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: Ran the full sweep on this PR: claims vs code (all nine claims in the description verified against the branch, including re-deriving the guard call-site swap against the parent commit), nearest-twin/absence pass over every author-facing platform read, non-happy paths, hostile inputs, and reachability-both-ways over every `device.platform` read in the changed files, plus a mutation on the new `hasRunnerTree` gate and a live run of the recorder's re-probe against a synthetic `remote:` device. Replayed the launch/`when`-guard matrix end-to-end through the real `flow-execute` path (22 scenarios: cross-platform map, `native` shorthand, bare-string launch, parser rejections, guard met/skip, restart-app failure modes, malformed map values). `npm run build`, full test suite (6179), lint, knip, prettier, docs build and docs format all pass in a clean worktree at head 467e75c59. One main finding (the ungated repair sentence in the recorder's indeterminate warning, with a live reproduction) plus two run-path siblings and one docs wording item, all left as inline comments.

packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -585,9 +586,11 @@ async function runLaunch(state: ExecState, app: Launch): Promise<DirectiveOutcom
 
   const bundleId = appIdForPlatform(app, device.platform);
   if (!bundleId) {
```
latekvo, line 599: [[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: `ios-remote` falls through every arm here, so a launch on a remote simulator reports `launch:pass` while `fetchFlowTree` throws on every read there (`ui-tree matching is not supported on platform "ios-remote"`) — the first selector directive of the same flow then hard-stops the run with that raw message. That is the failure mode this function's own contract says it exists to prevent ("an actionable, platform-specific reason instead of letting the first directive surface a raw tree-source error").

This is the limitation the PR description acknowledges ("selectors, idle waits and snapshots all still fail there"), so to be clear: the report is not about the limitation existing, but about the launch step reporting **pass** on the exact platform where it cannot be followed by a selector step — the run fails one step later, in a message that never mentions the launch succeeded or that this platform is tree-blind, while `treeSourceGate` exists to surface exactly this at the launch step. On a real remote device the launch can also fail in `restart-app`'s native-devtools precheck before any of this, which the mocked test does not exercise.

packages/docs/docs/reference/flow-yaml.mdx
```diff
@@ -147,6 +147,8 @@ Use the map form for a flow that runs on more than one platform:
 
 `native` is one id for iOS, Android and Vega. A per-platform key replaces it for that platform. `chromium` accepts a relative or absolute app path, or `{ path, args }`. A launch that has no id for the platform of the run is an error.
 
+A run on a remote simulator uses the `ios` entry.
```
latekvo, line 150: [[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: As an unconditional sentence this is not quite true: `appIdForPlatform` resolves the remote run's entry as `launch["ios"] ?? launch.native`, so a `{ native: … }` launch — which this PR verifies passing on a remote run — uses the `native` entry, not `ios`. As written, the sentence implies a launch map must carry an `ios` entry to run remotely, which a native-only flow does not need.

PR #1063, opened 2026-09-08
feat(flows): read the UI tree on a remote simulator

hubgan:
## What

Flows can now read the UI tree on a remote (cloud) iOS simulator.

Selectors, `await: { idle: true }`, `snapshot: { cropOn: ... }` and selector capture during recording all work there, and a `launch:` waits for the app's tree source before the next step runs - exactly as they do on a local simulator.

## Why

A flow run against a remote simulator was never refused, it just degraded. The flow tree source table had no `ios-remote` entry, so every read fell through to a helper that refuses the platform by name:

```
could not read the UI tree: ui-tree matching is not supported on platform "ios-remote"
```

That failed every selector step, errored the run on `await: { idle: true }` and `snapshot: { cropOn }`, and made the recorder keep bare coordinates instead of a selector.

The quiet half was worse. With no tree source, a coordinate gesture skipped its settle entirely and warned about nothing. In a measured run, four taps went out 50ms apart into a still-launching app and every one reported `pass`:

```
11:32:52.166  restart-app
11:32:54.823  gesture-tap      <- 2.6s after launch, app still cold-starting
11:32:55.297  gesture-tap
11:32:55.347  gesture-tap
11:32:55.398  gesture-tap
=> ok=True passed=6 failed=0 errored=0
=> screen: still on the login page. Every tap was lost.
```

A second identical run reached the target screen. Same flow, same device, same green report.

## How

A remote simulator is an iOS simulator reached over the sim-remote tunnel. The native-devtools blueprint already routes `ViewHierarchy.getFullHierarchy` to it over TCP, so the source is the local simulator source with no physical-device arm - `ios-remote` is always a simulator.

Four lines of behaviour, plus the comments that asserted the old state:

- **the tree source** - `ios-remote` gets an entry in the flow tree source table, onto the same iOS full hierarchy.
- **the launch gate** - `launch:` now waits for native devtools on a remote simulator too. This ships in the same PR on purpose: the source alone leaves the first directive after a launch racing the cold start, which is the window that lost those four taps. The gate alone is inert.
- **`fetchTree`** - the same arm, so the helper stops disagreeing with the `describe` tool, which has always served `ios-remote` through `describeIos`.
- **the terminate advice** - two iOS targeting reasons quote a command that clears a competing native-devtools-connected app. With the tree source, those reasons now reach a remote simulator, which is in no local device set, so an `xcrun simctl terminate` there answers "Invalid device". A remote simulator now gets `sim-remote simctl terminate <udid> <bundleId>`, with its udid spelled out.

## Tests

Two existing tests asserted the *absence* of support. They are flipped rather than deleted, so they now guard the fix:

- a remote device settles like a local one, and an outage on it warns in the same words iOS gets
- the recorder's advice on a remote device offers the iOS repair

New coverage in `flow-remote-tree.test.ts` and `flow-remote-launch.test.ts`, none of it stubbing the tree fetch, so it pins which *source* is dispatched to:

- a flow read on a remote device asks native devtools for the full hierarchy, at the same depth and fields a local simulator asks for, and returns the same tree
- a recorded tap writes a selector, not the coordinates it tapped, and a recorded wait's re-probe reads the full hierarchy and returns a determinate verdict
- the launch gate fails a remote launch when native devtools never comes up, in the same words a local one reports
- a targeting reason on a remote device quotes the `sim-remote` terminate command, and a local one still quotes `xcrun simctl`

Every one of these fails on the parent commit.

## Verification

End-to-end against a booted cloud simulator (`remote:73A22194-…`, iPhone 17 Pro / iOS 26.5) driven through the branch tool-server. Every row fails on the parent commit:

| Flow | Result |
| --- | --- |
| `await: { visible: { text: … } }` | passes - was `ui-tree matching is not supported on platform "ios-remote"` |
| `await: { idle: true }` | passes - was an errored run |
| `snapshot: { cropOn: … }` | passes - was an errored run |
| selector `tap` → `await` → `assert` | 4/4: resolves the element, taps it, navigates, asserts the new screen |
| recorded tap | writes a selector, not the coordinates it tapped |

What the gate buys is the settle. A flow of `launch:` then an immediate coordinate tap, with an `assert` that fails if the tap is lost, ran 6/6. On the parent commit the same flow alternated between landing the tap and silently losing it, reporting green either way.

`npm run build` and the tool-server suite pass.

## Docs

Three sentences in the create-flow YAML reference, each of which changes what an author writes or types: which launch key a remote run reads and its `native` fallback, which `when: { platform }` guard it matches, and that `--platform` accepts `ios-remote`. The first two are the notes the parent PR deferred until the reads worked.

Nothing else. A remote simulator behaves as iOS everywhere the docs already describe it, so "this now works" is a release note, not reference material - and a platform with no exceptions does not belong in a list of platform exceptions.


packages/tool-server/src/tools/flows/flow-run.ts
```diff
@@ -533,7 +533,10 @@ async function treeSourceGate(
       );
     }
   }
-  if (device.platform === "ios" && !signal?.aborted) {
+  // Both iOS simulator platforms gate the same way: `waitForNativeDevtools`
+  // resolves the service through `nativeDevtoolsRef(device)`, which the
+  // blueprint serves over TCP for a remote sim.
+  if ((device.platform === "ios" || device.platform === "ios-remote") && !signal?.aborted) {
```
latekvo, line 550: [[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: On a remote simulator this timeout verdict always measures `indeterminate`, and the remedy it quotes cannot repair the remote failure that produces it.

`remoteIosHost.inspectRunningApp` returns `process: null` by construction (packages/tool-server/src/utils/ios-host.ts:604-610 — app processes live on the orchestrator, out of reach of the local process table), and `appConnectionState` maps `process: null` to `indeterminate` (packages/tool-server/src/blueprints/native-devtools.ts:1047), as it does the failed `reverifyEnv`/`inspectRunningApp` calls a dead tunnel also produces (lines 1029, 1034-1045). The process-shaped states whose measured remedies are local-repair-free (`stale_process`, `unregistered`, `connecting`) are unreachable for a remote app: they all require `inspection.process`.

So when the sim-remote tunnel drops after a successful `restart-app` (the app connected, then the tunnel died — the one remote-specific failure mode this gate now serves), the launch fails after 16.5s with:

> could not connect to native devtools. Native devtools are not connected to com.acme.app, and its process could not be inspected to tell whether it is injected. Call restart-app then retry. If it is still not connected after that restart, the native-devtools service is stale rather than the app being uninjected — do not keep restarting the app; restart the tool-server (`argent server stop && argent server start --detach`) and retry. This step already performed that one restart, so re-run the flow at most once more before restarting the tool-server rather than the app.

Reproduction (real `runLaunch` → `treeSourceGate` → `waitForNativeDevtools` → `appConnectionState`, only the service stubbed; ran against this head in a scratch worktree):

```ts
// device: remote:<UDID>; restart-app returns success
resolveService: async () => ({
  isConnected: () => false,
  listConnectedBundleIds: () => [],
  appConnectionState: async () => "indeterminate", // what the remote measurement collapses to
})
// → launch:error after the real 15s budget; reason quoted above
```

The remedy named is a tool-server restart, which rebinds local listeners and cannot re-establish a remote tunnel — following it re-runs the flow into the identical state, twice. The actual repair (reconnect the tunnel / re-run `sim-remote` setup, whose failure stderr the init path does capture) is never named author-facing. The comment at flow-run.ts:536-538 ("the blueprint serves over TCP for a remote sim") covers the transport but not the verdict text, and the new pin at flow-remote-launch.test.ts:137-146 locks this same wording for remote on purpose ("the author cannot act on differently anyway") — but the remote action set here genuinely does differ: restarting the tool-server cannot reconnect what the tunnel severed.

The mid-run counterpart has the same shape one level down: a tunnel drop after launch surfaces as `${bundleId} lost its devtools connection after launch (the app crashed, was terminated, or its socket closed)` (packages/tool-server/src/tools/flows/flow-ios-tree.ts:369), which attributes to an app crash the failure whose likely cause on this platform is the transport.

packages/tool-server/test/flows/flow-record-cross-tree.test.ts
```diff
@@ -2101,22 +2099,38 @@ describe("a recorded wait is re-probed against the runner's tree", () => {
...
+  // hierarchy. So the verdict there is determinate, not UNKNOWN.
   it("is reachable on ios-remote: await-ui-element accepts the device", () => {
     const tool = createAwaitUiElementTool(registryWhereWaitSucceeds());
     expect(tool.capability?.appleRemote).toEqual({ simulator: true });
     expect(() =>
       assertSupported("await-ui-element", tool.capability, resolveDevice(`remote:${IOS}`))
     ).not.toThrow();
   });
+
+  it("returns a determinate verdict on ios-remote, not an UNKNOWN one", async () => {
```
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: This new test's claim of base-failure does not hold: I ran it against the parent commit (467e75c5, PR test copied over the base source) and it passes there — the mock at the top of the file stubs `fetchFlowTree` wholesale, so at base the re-probe's throw is caught and translated into the same determinate "does NOT hold against the tree the runner resolves" branch this test asserts. It passes on base-src, on head-src, and even with the new `ios-remote` arm in flow-tree.ts mutated out, so it pins nothing about the arm this PR adds (unlike its sibling at line 1284, whose `supportsFlowTree` assertion does pin the table). The PR body states "Every one of these fails on the parent commit"; this one doesn't, so the coverage claim for the determinate-verdict behavior is narrower than described — the behavior change itself is real, but this test doesn't guard it.

packages/tool-server/src/tools/flows/flow-run.ts
hubgan: Checked this on a cloud simulator (iPhone 17 Pro, iOS 26.5) through this branch's build: a passing `launch:`, then `sim-remote reverse stop <udid> <port>` on the native-devtools tunnel, then the same flow again.

The first half holds. Every unconnected state reads `indeterminate` there, and the next `launch:` fails the gate with exactly the quoted reason (after 23-25s rather than 16.5s, because the remote env re-verify is slow). Re-running the flow fails the same way.

The named remedy does repair it, though. The tunnel is not outside the tool-server: the native-devtools service creates it (`host.startProxy`, which is `sim-remote reverse start`, in the factory, and `reverse stop` in `dispose`). A restarted tool-server bound a new port, opened a new tunnel to it and re-pointed `NATIVE_DEVTOOLS_IOS_CDP_PORT`, and the same flow passed in 7.2s (9.2s on a second cycle). So the escalation the verdict names - re-run at most once, then restart the tool-server - ends in a passing run. The re-run is still the right first move on a remote sim, where a cold start slower than the gate reads `indeterminate` too.

The mid-run message behaves the same way: a drop mid-flow surfaces as `lost its devtools connection after launch (... or its socket closed)`, and following it (relaunch, the gate's verdict, then a tool-server restart) also ends in a passing run.

So I'm leaving the wording as it is.

packages/tool-server/test/flows/flow-record-cross-tree.test.ts
hubgan: Fixed in f1451a6ff.

Confirmed first: with the five source files reverted to the parent commit, the test still passed, because this file stubs `fetchFlowTree`. It now lives in `flow-remote-tree.test.ts`, which stubs nothing below the native-devtools service, and it also asserts that the re-probe issued `ViewHierarchy.getFullHierarchy`. It fails on the parent commit, and with only the `ios-remote` arm removed (`expected [] to include 'ViewHierarchy.getFullHierarchy'`).

The same pattern in `flow-gesture-settle.test.ts` ("resolves a selector step there instead of refusing the platform" also passes on the parent commit) is corrected in 74ba56b0b.

packages/tool-server/src/tools/flows/flow-tree.ts
```diff
@@ -46,24 +46,32 @@ const FLOW_TREE_SOURCES: Partial<
...
       : queryFullHierarchyTree(registry, device, target),
-  android: (registry, device) => queryAndroidFullHierarchy(registry, device),
-  chromium: (registry, device) => queryChromiumTree(registry, device),
-  vega: (_registry, device) => queryVegaTree(device),
+  // A remote sim is an iOS simulator reached over the sim-remote tunnel, and
+  // the native-devtools blueprint routes `getFullHierarchy` over TCP for one.
+  // So it is the local simulator source with no `kind === "device"` arm:
+  // `ios-remote` is always kind "simulator" (utils/device-info.ts) and has no
+  // physical-device variant.
+  "ios-remote": (registry, device, target) => queryFullHierarchyTree(registry, device, target),
```
latekvo, line 58: [[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: This arm sends a remote device into the unpinned auto-target path of `queryFullHierarchyTree`, whose failure reasons quote a terminate remedy built for the LOCAL host — a command that cannot reach a cloud simulator.

`explainTargetingFailure` (flow-ios-tree.ts:595 and :654-655) embeds `terminateCommand`, which resolves the command through `simctlTargetForUdid(device.id)`. For a `remote:` id that call passes straight through the entitlement gate (a raw udid no provider claims is not refused — external-devices.ts:473-481) and, with no additional device sets configured, resolves to the default set, so the clause is kept and the quoted remedy is:

> `xcrun simctl terminate <udid> <bundleId>`

A cloud simulator is addressable only through the `sim-remote` CLI (`simctlTerminate`, utils/sim-remote.ts:126-128); no local CoreSimulator set contains it, so following the remedy fails with simctl's not-found error and the loop continues. The `<udid>` placeholder does not save it: substituting the device id embeds `remote:<uuid>`, which local simctl rejects, and stripping the prefix leaves a UUID with the same shape as a local iOS UDID (device-info.ts:37-39), so the only host where the command "succeeds" is a UUID collision — terminating an app on the wrong machine.

Reproduction (run against this head, calling the real `simctlTargetForUdid` the way `terminateCommand` does):

```ts
const { prefix } = await simctlTargetForUdid("remote:00000000-0000-0000-0000-0000000000ab");
// → prefix is ["simctl"], no throw — so `terminateCommand` quotes
//   `xcrun simctl terminate <udid> <bundleId>` for a device only
//   `sim-remote simctl terminate <udid> <bundleId>` can act on.
```

Trigger on a real run: flow on a `remote:` sim with two native-devtools-connected apps (the injection is simulator-wide, so any dev-client app connected alongside the launched one) and no unique frontmost — every failing `await:`/selector step re-quotes the unusable command, and the same clause rides inside `unsettledGestureWarning` and the recorder's per-tap warning. Nothing downstream rewrites the command for remote devices, and `terminateCommand`'s doc comment (flow-ios-tree.ts:693-714) is silent on `ios-remote`.

latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: Re-ran the full sweep at 0a12387c (claims vs code, nearest twin, non-happy paths, inputs, reachability, lifetime, absence pass, plus a mutation run that removed the three `ios-remote` arms — 12 of the new tests fail without them). Both earlier threads verified at head. One new finding inlined below. One note, prose-only, in a file this diff does not touch: the comment above `UNSUPPORTED_PLATFORM` in flow-add-step.ts:133-136 still says `fetchFlowTree` "answers only on ios, android, chromium and vega" — the `ios-remote` arm this PR adds makes that "only" false, and the sibling PR that reworks this file (#1064) leaves the sentence standing.

PR #1064, opened 2026-09-08
refactor(flows): make the flow tree source table total

hubgan:
Stacked on #1063.

## What

Deletes the code written to handle "this platform has no UI tree source at all":

- `FLOW_TREE_SOURCES` becomes `Record<Platform, ...>` instead of `Partial<Record<...>>`
- `fetchFlowTree` dispatches straight through the table - the `fetchTree` fallthrough is gone
- `fetchTree` in `ui-tree-match.ts` is deleted too - that fallthrough was its last caller, and knip's unused-export gate fails on an export nothing imports
- `supportsFlowTree` is deleted, along with its two callers (the guard in `settleForGesture` and `hasRunnerTree` in the recorder's advice, which also chose the no-tree wording #1062 gives the re-probe warning)
- comments that still described a platform with no tree source are corrected

## Why

The remote simulator was the last platform without a tree source, and #1063 gives it one. Every `Platform` now has a source, so those branches can never run.

Leaving them in is worse than not having them, because they degrade quietly. The guard in `settleForGesture` is the clearest case: on a platform with no source it skipped the settle and returned no warning, so a coordinate gesture fired into whatever motion was in flight and the run reported a clean pass. That is exactly what a cloud run did before #1063.

With `Partial` gone, adding a platform to `Platform` without giving it a tree source is a compile error rather than a run that silently reads nothing.

## Behaviour

No change. `supportsFlowTree` already returned true for `ios`, `android`, `chromium` and `vega`, and #1063 makes it true for `ios-remote` too - so every caller took the branch that is now unconditional.

## Verification

- `npm run build` - compiles with no `as` casts
- Guard proven: temporarily adding a member to `Platform` fails `tsc` on `FLOW_TREE_SOURCES`, then reverted
- End-to-end against a booted cloud simulator (`remote:73A22194-…`, iPhone 17 Pro / iOS 26.5) - the platform whose deleted branch was the only live one, so it is where a behaviour change would show. A selector `tap` → `await` → `assert` flow, a selector `await` and `await: { idle: true }` all pass unchanged, and the launch-then-coordinate-tap race ran 5/5
- `npm test --workspaces --if-present` - suites pass, apart from a pre-existing `argent-installer` failure that also fails on the base branch: it reads the developer's real `~/.config/opencode` directory

## Docs

No user-facing change, so no docs update.


latekvo (approved):

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]: Ran the full sweep on this one: claims vs code (every statement in the PR description checked against the source), nearest twin, non-happy paths (abort/timeout/outage paths through `settleTree`), inputs, reachability both ways, what outlives the call, plus the absence pass and mutation checks across the new total table and the unwrapped `settleForGesture`. Verified the totality guarantee by hand: adding a `"tvos"` member to `Platform` makes `tsc` flag `FLOW_TREE_SOURCES` — and restoring the old `Partial` table with its `fetchTree` fallthrough and `supportsFlowTree` under the same new member still compiles clean, so it is this PR's shape that turns the silent runtime read into a compile error. Confirmed all three callers previously took the now-unconditional branch (`supportsFlowTree` returned true on every `Platform`), that the deleted-guard semantics (abort checkpoint, outage warning, settle skip memo) are unchanged, that the probe advice prose now matches the total table, and that the removed exports leave no orphaned imports. Ran `npm run build` (no `as` casts introduced), the full tool-server flow suite (72 files, 2090 tests including `flow-gesture-settle` and both updated test files), and Prettier — all pass. Returned completely clean. Thank you for contributing!

PR #1065, opened 2026-09-08
feat(flows): make a snapshot on a remote simulator comparable

hubgan:
Flow snapshots did not work the same way on a cloud simulator as on a local one, so a visual run there proved less than it looked.

**Baselines were never shared.** The baseline filename carried the host, so a cloud run of an iPhone 17 Pro looked for `home__ios-remote-1206x2622.png` while a local run of the same iPhone had committed `home__ios-1206x2622.png`. A missing baseline is a hard step failure, so every baseline had to be captured and reviewed twice, and the two copies could then drift apart with nothing to catch it. The key now names the device class rather than the machine: a remote simulator uses the `ios` key. The `WxH` component still does the real work of catching a wrong-simulator or rotated run.

A shared baseline also needs a visible writer. A remote run with `--update-baselines` rewrites the same file a local run compares against, so its step reason now says `baseline updated from a remote simulator`. Exported failure artifacts go the other way: a remote run's files are named `<key>-remote-<role>.png`, so a local and a remote run exported into one `--output` directory keep their evidence apart.

**The clock kept moving.** `pinStatusBar` freezes the bar at 9:37 so it cannot drive a diff, but it had no remote arm and quietly returned false. A full-screen snapshot got away with it because the differ masks the top band; a `cropOn` region overlapping the bar has no such mask and leans on the pin directly, so it would have been flaky. Both `pinStatusBar` and `restoreStatusBar` now drive a remote simulator through a new `simctlStatusBar` wrapper. The override values live in one constant shared by both iOS arms, so local and remote cannot pin to different pixels, which would fail any `cropOn` snapshot that includes the bar.

Adding that arm surfaced a second question the local code never had to answer: what to do when the pin fails *and* cannot be undone. `pinStatusBar` returns false there, and the caller schedules no teardown after a false. That is right for local iOS, where one `xcrun` command either applied or did not. It is wrong across a network, where one dead tunnel fails both calls and the far host may already have applied the request whose response never came back - leaving a shared cloud simulator pinned after the run, with nothing scheduled to clear it. A remote pin now joins Android in the retry set.

A network also changes what a hung call costs. Each remote `status_bar` call now gives up after 5s instead of the CLI's 30s default, so an unresponsive tunnel no longer holds a run for the override, its undo and the teardown restore at 30s each. A run that was cancelled before the pin pins nothing, and a remote tvOS simulator, which has no status bar, is skipped.

**The capture budget was sized for localhost.** Every warm settle capture got a flat 2s. On a cloud device each capture is a round trip to another machine, and the settle needs two comparable captures per interval, so one dropped read costs the whole round. A warm remote capture now gets a 4s ceiling. It is a ceiling, not a wait, so it costs nothing unless it is actually needed. The first capture keeps the wider first-capture ceiling, which it does not use: a remote capture is one MoQ request and never polls for a first frame.

## Verification

- Baseline key on a cloud simulator (`remote:73A22194-…`, iPhone 17 Pro / iOS 26.5): a `snapshot:` step writes `shared__ios-1206x2622.png`. The parent commit writes `shared__ios-remote-1206x2622.png`.
- Cross-host reuse, which is the point of the change: that cloud-written baseline was then compared by a run on a local iPhone 17 Pro, reporting `diff 4.96% > 0.5% (shared__ios-1206x2622.png)` - a pixel comparison, not `missing baseline`. On the parent commit the two hosts write different filenames, so this cannot happen at all.
- The cloud baseline, captured mid-run, shows the status bar pinned to 09:37 with full bars and a charged battery, so the remote pin applies and is what a later run compares against.
- A local run is unchanged: wrote `home__ios-1206x2622.png`, and a second run compared against it at `diff 0.00%`.
- The compiled `status-bar` module driven against a stub `sim-remote` on `PATH`, with no test mocks: it emits `simctl status_bar <udid> override --time 9:37 …` and `simctl status_bar <udid> clear`. The teardown bug was reproduced the same way - with both calls failing, a remote pin reported unpinned where Android in the identical state reported pinned - and the same run now reports pinned.
- Unit tests for every behavior here. Each was confirmed to fail when its production change is reverted.
- `npm run build`, `npm run lint`, `npm run knip`, `npm test --workspaces`, and the docs build all pass. One unrelated failure in `argent-installer` (`opencode adapter > globalPath`) reproduces on the base branch untouched.

## Docs

`reference/flow-yaml.mdx` said a baseline is keyed by platform and capture geometry, which no longer told a reader whether a cloud run needs baselines of its own. It now says a remote run uses `ios` in the key and shares the baseline of a local iOS run with the same capture geometry, and that a remote `--update-baselines` rewrites that file. The create-flow skill reference says the same.


packages/tool-server/src/utils/sim-remote.ts
```diff
@@ -155,6 +155,15 @@ export async function simctlPrivacy(
...
 }
 
+/**
+ * Remote analogue of `xcrun simctl status_bar <udid> <action> ...` — overrides
+ * or clears the remote simulator's status bar. `sim-remote simctl` forwards its
+ * arguments verbatim, so the argv is the local one minus the udid.
+ */
+export async function simctlStatusBar(udid: string, args: string[]): Promise<void> {
+  await run(["simctl", "status_bar", stripRemotePrefix(udid), ...args]);
+}
```
latekvo, line 165: [[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: Every other tool that narrows an already-supported remote device by runtime reaches for `isRemoteTvOsSimulator` (shake/platforms/ios.ts:115, paste/platforms/ios.ts:81) — the remote tvOS probe this repo already maintains. The flows path never does, so a remote **tvOS** simulator also classifies as `ios-remote` and is treated as having a status bar throughout the flows path: `statusBarMaskFraction` masks the top 6% of its full-screen capture unconditionally (flow-pixels.ts:156), and `pinStatusBar`/`restoreStatusBar` now override a bar that doesn't exist there. Impact: the settle's idle/pixel comparison is blinded to real motion in that band on a remote TV sim — the exact harm the local arm's `isTvOsSimulator` probe exists to prevent — and the tvOS runtime lookup that would distinguish them is one memoized call away. Repro'd by reading the classification path: `simctlListDevices()` iterates every runtime the orchestrator reports, including tvOS ones, and `listRemoteIosSimulators` (tools/devices/list-devices.ts:154-174) tags them all `platform: "ios-remote"`, so a remote tvOS device reaches these arms as `ios-remote` with no runtime check. Note the unconditional mask arm itself predates this PR; flagging it here because this PR's new test comment (flow-pixels.test.ts:403-407, "sim-remote drives an ordinary iOS simulator, so it has the same status bar to mask") states it as true of every remote sim when it is only true of the iOS ones.

packages/tool-server/src/tools/flows/flow-pixels.ts
```diff
@@ -187,6 +193,14 @@ export function pixelCaptureTimeoutMs(device: ActionEnv["device"], firstCapture:
   if (isIosPhysicalDevice(device)) {
     return IOS_DEVICE_PIXEL_CAPTURE_TIMEOUT_MS;
   }
+  // A remote simulator warms its stream like a local one, so its first capture
+  // keeps the cold-stream wait; every later capture still crosses the network,
+  // which the localhost warm bound does not allow for. `waitForIdle` needs two
+  // comparable captures per interval, so one timed-out read costs a whole
+  // settle round.
+  if (device.platform === "ios-remote" && !firstCapture) {
```
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: The comment's mechanism doesn't hold for the remote route. A local capture "keeps the cold-stream wait" because `httpScreenshot` polls for the stream's first frame under `FIRST_FRAME_WAIT_MS` (simulator-client.ts:474-510). A remote capture never reaches that loop: `httpScreenshot` returns via `api.transport.screenshot` before it (simulator-client.ts:462-468), the MoQ transport's `screenshot` is a single awaited request (simulator-client.ts:618-622), and the cloud SDK's `ScreenshotChannel` carries its own internal 30s request timeout (`DEFAULT_REQUEST_TIMEOUT_MS = 3e4` in @swmansion/argent-cloud-sdk), not a first-frame poll. So for `ios-remote`, `firstCapture=true` doesn't preserve any wait — the 6.5s budget is just unused headroom, like the adjacent tvOS case the function's doc already acknowledges. Behavior is safe (it's a ceiling, not a wait), but the rationale as written would steer a future reader into "the 6.5s is load-bearing for the remote first capture" when nothing on this route consumes it; if the MoQ session setup ever exceeds 6.5s, the round loses its pixel evidence entirely rather than being bounded by this branch.

latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: Ran the full sweep on this PR: claims vs code, nearest twin, non-happy paths, inputs, reachability both ways, lifetime/teardown, plus the absence pass (docs coverage, sibling platform arms, test-mutation pinning) across the new key fold, the remote status-bar arms and the capture-timeout branch. Reproduced the baseline-key fold and both arms' argv locally against the unit suites (108 tests across the three touched files pass; tsc and eslint clean). Every HIGH/MEDIUM candidate that surfaced was traced to a guard or pre-existing behavior and dropped; two LOW findings posted inline. Event is COMMENT only — the approve/changes call is yours.

packages/tool-server/src/utils/sim-remote.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: Still present as of 6f9e5e8 — re-verified at head: `listRemoteIosSimulators` still tags every runtime `ios-remote` (list-devices.ts:158-168), `isRemoteTvOsSimulator` remains used only by paste/shake, and the new `pinStatusBar`/`restoreStatusBar` arms now drive a bar for any `ios-remote` device, tvOS included.

packages/tool-server/src/tools/flows/flow-pixels.ts
latekvo: [[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: Still present as of 6f9e5e8 — re-verified at head: the first-capture branch still returns 6.5s for `ios-remote`, while `httpScreenshot` still returns through `api.transport.screenshot` (simulator-client.ts:462-468) before any first-frame polling, the MoQ transport's `screenshot` remains a single awaited request, and `DEFAULT_REQUEST_TIMEOUT_MS = 3e4` in @swmansion/argent-cloud-sdk remains the only bound on that route. The 6.5s stays unused headroom rather than a load-bearing wait.

latekvo (commented):
[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash]: Second full pass at head 6f9e5e8, after the earlier review: claims vs code on every bullet of the PR description, nearest-twin checks (simctlPrivacy, the Android retry arm, the physical-iOS capture ceiling), non-happy paths (partial pin, dead tunnel, restore failure, dropped read in the settle loop), inputs, reachability both ways, teardown/lifetime, plus the absence pass and mutation checks — reverting the retry return, the key fold, and the remote capture ceiling each breaks exactly the new tests, and reverting the remote pin return to the pre-PR form reproduces the described Android-vs-remote divergence (unpinned where Android reports pinned). 108 tests across the three touched suites pass and tsc is clean. No new HIGH or MEDIUM findings; the two open threads from the earlier review still hold at this head and were answered there rather than duplicated. Event is COMMENT only — the approve/changes call is yours.

packages/tool-server/src/tools/flows/flow-pixels.ts
hubgan: Fixed in e9fc07a22

packages/tool-server/src/utils/sim-remote.ts
hubgan: Not changing this in this PR. The part that causes the harm predates it.

- The top-band mask on `ios-remote` (`statusBarMaskFraction`, flow-pixels.ts:156) has been on main since #728, so a remote tvOS capture was masked before this PR. The "ordinary iOS simulator" wording in the test comment comes from the same commit.
- The new pin/restore arms do nothing on tvOS. On a tvOS 26.5 simulator, `simctl status_bar <udid> override ...` and `simctl status_bar <udid> clear` both exit 45 with "Status bar overrides not supported on this platform". I ran this branch's `pinStatusBar` against that simulator through a `sim-remote` stub. Nothing was overridden and nothing threw. The only difference from the local arm is one extra failed `clear` at run end.
- No remote tvOS device reaches this path today. The cloud orchestrator lists only an iOS 26.5 runtime (iPhone 17 Pro, iPhone SE, iPad Air).

A remote tvOS check would belong in `statusBarMaskFraction`, which this PR does not touch.

packages/tool-server/src/tools/flows/flow-visual.ts
```diff
@@ -215,7 +215,12 @@ export async function runSnapshot(
     opts.cropOn === undefined
       ? ""
       : `-crop-${createHash("sha256").update(cropIdentity(opts.cropOn)).digest("hex").slice(0, 8)}`;
-  const snapshotKey = `${opts.name}__${env.device.platform}-${w}x${h}${cropSuffix}`;
+  // Keyed on the AUTHORING platform: the key names a device class, not a host.
+  // A remote simulator of the same model renders the same pixels at the same
+  // geometry, so it must reuse the baseline a local run committed rather than
+  // demand a second copy that can drift. `WxH` still separates genuinely
+  // different device classes, which is the check the key exists for.
+  const snapshotKey = `${opts.name}__${authoringPlatform(env.device.platform)}-${w}x${h}${cropSuffix}`;
```
latekvo, line 223: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Driving the real `runSnapshot`, the fold makes a remote run and a local run resolve to one filename, so `argent flow run <x> --device remote:<uuid> --update-baselines` now rewrites the committed local baseline in place:

```
LOCAL  seeded:  baseline written (home__ios-390x844.png)   size=2510
REMOTE seeded:  baseline updated (home__ios-390x844.png)
baseline dir after: [ 'home__ios-390x844.png' ]   local baseline bytes changed: true
```

Pre-fold the two hosts wrote separate files and that was not reachable. Nothing records which host produced the bytes: the key carries no host, no iOS-runtime-build and no `kind` component, `VisualOutcome.snapshotKey` is already folded by the time it reaches the run report, and the reason line reads `baseline updated` either way - so a reviewer about to commit the PNG cannot tell which lane it came from.

The comment above states the premise as settled ("renders the same pixels at the same geometry"), but the only cross-host measurement in this PR is `diff 4.96% > 0.5%` - roughly 10x the default `maxMismatch`, against a `0.00%` local-to-local control. I checked what could absorb that and nothing does: `normalizeToCommonSize` returns the identity pair when the geometry already matches (it only resamples on a size difference), `DEFAULT_THRESHOLD = 0.1` is applied per-pixel before the count, there is no colour management in `diffPngFiles`, and no committed flow YAML raises `maxMismatch` off 0.5. If that 4.96% is host divergence rather than app content, the mixed-host workflow this enables cannot be green on both lanes at once, and the only escape is the overwrite above - which then breaks the other lane.

Two related gaps. `remoteEnv` (`flow-visual.test.ts:94`) is never passed `updateBaselines: true`: both new tests exercise the compare direction, so the write direction this happens on is uncovered. And the collision reaches `--output` too - `exportFailureArtifacts` (`argent-cli/src/flow.ts:660`) writes `<key>-<role>.png`, so a local and a remote run of one flow now both produce `home__ios-390x844-current.png` where they previously produced two names:

```
this PR:  [ 'home__ios-390x844-current.png' ]
parent:   [ 'home__ios-390x844-current.png', 'home__ios-remote-390x844-current.png' ]
```

packages/tool-server/src/utils/status-bar.ts
```diff
@@ -4,13 +4,34 @@ import type { DeviceInfo } from "@argent/registry";
...
 
 /** Pins the status bar to fixed values so its clock / battery / signal never drive a screenshot diff. */
 
 const DEMO_BROADCAST = "am broadcast -a com.android.systemui.demo";
 
+/**
+ * The overridden values, shared by the local and remote iOS arms. A remote
+ * simulator shares its baselines with a local one of the same model, so the two
+ * must pin the bar to the same pixels — a second literal here would let them
+ * drift and fail every shared snapshot on the clock alone.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: "a second literal here would let them drift and fail every shared snapshot on the clock alone" - measured, a clock drift cannot fail a full-screen snapshot at all. `runSnapshot` passes `topMask: cropFrame === undefined ? "status-bar" : "none"` (`flow-visual.ts:340`), and that mask covers the band the clock sits in. Real `diffPngFiles` on a 1320x2868 capture taken from a booted simulator, with the whole top band replaced between baseline and current:

```
topMask=status-bar   mismatch=0.0000%   vs maxMismatch 0.5  => PASS
topMask=none         mismatch=5.9972%   vs maxMismatch 0.5  => FAIL
```

So the exposure is a `cropOn` region overlapping the bar, not "every shared snapshot". The diff states this correctly 34 lines below, on the `ios-remote` pin arm ("a `cropOn` there is not masked"), so the file now carries both versions of the claim - and this is the one a reader meets first, on the constant whose existence it justifies.

packages/docs/docs/reference/flow-yaml.mdx
```diff
@@ -263,7 +263,7 @@ A fragment whose `run:` chain reaches a launch cannot declare `executionPrerequi
 
 A snapshot compares the current screen, or the frame of `cropOn`, with a stored baseline. A missing baseline fails the step. A mismatch above `maxMismatch` percent fails the step. The default `maxMismatch` is 0.5. A `cropOn` element whose size changed fails the step.
 
-Baselines live in `.argent/flows/__baselines__/<flow>/`. Argent keys a baseline by platform and capture geometry, plus the selector for `cropOn`. Run `argent flow run <name> --update-baselines` to write the baselines from a known-good state. Review each baseline before you commit it.
+Baselines live in `.argent/flows/__baselines__/<flow>/`. Argent keys a baseline by platform and capture geometry, plus the selector for `cropOn`. A run on a remote simulator uses `ios` in the key. It compares against the baseline that a local run of the same device model writes. Run `argent flow run <name> --update-baselines` to write the baselines from a known-good state. Review each baseline before you commit it.
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: This sentence has a second copy the change did not reach. `packages/skills/skills/argent-create-flow/references/flow-yaml.md:258` still reads "Baselines live under `.argent/flows/__baselines__/<flow>/` and are keyed by platform and full-capture geometry". That file is the reference the `argent-create-flow` skill ships to an authoring agent, and line 252 of the same file already advertises `--platform ios|android|chromium|vega|ios-remote` - so an agent is offered `ios-remote` as a target and told six lines later that a remote run needs its own baselines, which is the belief this PR exists to remove. These are the only two copies of the sentence in the repo, and the two preceding ios-remote commits on this branch (`677a322bf`, `9e3f7b7a7`) updated both together, so it reads as an omission rather than a deliberate split.

Two gaps in the new wording itself. It describes only the compare direction, while a remote `--update-baselines` overwrites the local baseline in place - stated nowhere. And "the same device model" overstates what the key enforces: the key is geometry, and iPhone 15, 15 Pro and 14 Pro all capture at 1179x2556.

packages/tool-server/test/flows/flow-pixels.test.ts
```diff
@@ -400,10 +400,11 @@ describe("statusBarMaskFraction", () => {
   });
 
   it("masks the band on a remote iOS simulator too", async () => {
-    // sim-remote drives an ordinary iOS simulator, so its status bar ticks
-    // like a local one — and the run-level `pinStatusBar` does not cover the
-    // platform either. The tvOS probe reads the local simulator list, which
-    // cannot see another machine's device, so it is not asked.
+    // sim-remote drives an ordinary iOS simulator, so it has the same status
+    // bar to mask. The run-level `pinStatusBar` covers it too, but the mask is
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: The replacement rationale does not hold. `statusBarMaskFraction` has exactly one production caller - `flow-actions.ts:1817`, inside `waitForIdle`, which is an `idle:` step within a flow run, i.e. precisely where `pinStatusBar` does apply. There is no "snapshot outside a flow run" leaning on this mask: `runSnapshot` never calls it, and the top band a snapshot ignores comes from `diffPngFiles`' own `topMask` / `DEFAULT_IGNORE_TOP_NORMALIZED_Y`, a separate constant.

The previous comment was true before this PR and the PR correctly falsified it. The sentence that replaced it is what a future reader would use to decide whether the `ios-remote` mask arm is still load-bearing now that the pin covers the platform, and it points at a consumer that does not exist - the reasons the arm is still needed are the ones already documented at `flow-pixels.ts:123-130` (the pin lands after the run starts; a nested `flow-execute` clears it).

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Second full sweep at head `92cdda20`, independent of the earlier run on `6f9e5e8`: claims vs code, nearest twin, non-happy paths, inputs, reachability both ways, lifetime/teardown, plus the absence pass and mutation testing.

Ran it E2E through the real entry point: `flow-execute` against a booted iPhone (1320x2868) on a tool-server built from this head. `--update-baselines` wrote `settings-home__ios-1320x2868.png`; the captured baseline shows the bar pinned to 09:37 with full bars and a charged battery; a second run compared at `diff 0.00%`; the teardown restore put the live clock back. The local arm is unregressed. For the remote arms I drove the compiled `status-bar` module against a stub `sim-remote` on `PATH` with no test mocks, and measured the timings in the thread below.

Mutation-tested the diff: 23 mutations across the override values, both new platform arms, the retry predicate, the capture ceiling and the key fold - 22 killed. The two survivors are benign (a duplicated-but-identical literal; and hardcoding the key to `ios`, which the pre-existing `flow-snapshot-collision.test.ts` catches). The new tests discriminate.

Checked and dropped: device sets (`simctlArgsForUdid` adds nothing a remote host needs, and no `sim-remote` wrapper passes `--set`); pin/restore symmetry; the artifact/cache/report sweep (the fold reaches every consumer of `snapshotKey` - no surface still spells `ios-remote`); the recorder path (`flow-add-step` refuses `snapshot` outright); two-concurrent-run races on the baseline file (real, but two local runs already collided - the fold widens rather than introduces it); and the remote capture-timeout rationale, which now holds - the MoQ route really does return before the first-frame poll, and 6500 > 4000, so the "unused headroom" wording is accurate.

The two open threads from the earlier run were not duplicated here. The remote-tvOS one is unreachable on the runtimes the orchestrator currently offers, as answered there.

Event is COMMENT only - the approve / changes call is the maintainer's.

packages/tool-server/src/utils/sim-remote.ts
```diff
@@ -155,6 +155,15 @@ export async function simctlPrivacy(
   await run(["simctl", "privacy", stripRemotePrefix(udid), action, service, bundleId]);
 }
 
+/**
+ * Remote analogue of `xcrun simctl status_bar <udid> <action> ...` — overrides
+ * or clears the remote simulator's status bar. `sim-remote simctl` forwards its
+ * arguments verbatim, so the argv is the local one minus the udid.
+ */
+export async function simctlStatusBar(udid: string, args: string[]): Promise<void> {
+  await run(["simctl", "status_bar", stripRemotePrefix(udid), ...args]);
```
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `run()` applies `DEFAULT_TIMEOUT_MS = 30_000` and takes no abort signal, and `pinStatusBar` runs unconditionally before step 1 of every flow run (`flow-run.ts:1358`) whether or not the flow has a `snapshot:` step. Against an unresponsive tunnel that is three sequential 30s waits: the override, the `clear` the catch issues, and - because the double failure now returns `true` - the run-end restore. Measured against the compiled module with a stub `sim-remote` that never exits:

```
pinStatusBar(ios-remote) -> true   [60013ms]
  simctl status_bar <udid> override --time 9:37 --batteryState charged --batteryLevel 100 --wifiBars 3 --cellularBars 4
  simctl status_bar <udid> clear
```

60s before the first step, then up to 30s more at teardown. Neither `pinStatusBar` nor `restoreStatusBar` takes the run's signal and `run()` sets no `killSignal`, so a client disconnect neither cancels nor shortens it; progress events fire only per completed step, so the stall is indistinguishable from a slow device. On the parent commit an `ios-remote` device fell through to `return false` here and spent nothing. The local `ios` arm cannot reach this - it is `xcrun` on localhost - and `sim-remote`'s own siblings that carry a real cost override the default (`bootstatus` 5min, `openUrl` 60s) while this one takes it.

PR #1066, opened 2026-09-08
test(flows): stop pinning idle and type-focus waits to exact wall-clock boundaries

### LINKED ISSUE #795: flow-idle-run: four wall-clock assertions fail whenever the flows directory is run on its own
`packages/tool-server/test/flows/flow-idle-run.test.ts` asserts absolute wall-clock windows with real timers and no slack. Run the flows directory as a batch and four of them fail; run the file alone and all 46 pass.

## Repro

```
npx vitest run packages/tool-server/test/flows
#  Test Files  1 failed | 42 passed (43)
#       Tests  4 failed | 1039 passed (1043)

npx vitest run packages/tool-server/test/flows/flow-idle-run.test.ts
#  Test Files  1 passed (1)
#       Tests  46 passed (46)
```

Reproduced twice, on two different working trees, on an otherwise idle machine.

## What fails

- `settles inside the smallest timeout the parser allows for the hold` (line 268) — `AssertionError: expected 1671 to be less than 1000` at line 286. The step is a `timeout: 1000, stableFor: 800` idle wait whose settle normally lands at ~820ms.
- `does not call a slow but answering source a wedged one` (line 536) — `Test timed out in 5000ms`. The fixture serves 2300ms reads inside a 4600ms wait, so it needs almost the whole default test timeout even when nothing is loaded.
- `does not report a tree-only settle when the hold was never served` (line 634)
- `still warns, rather than erroring, when a slow tree source keeps changing` (line 1294)

All four measure `Date.now()` deltas across a real-timer poll loop and compare them to the flow's own budget. Scheduling 43 flow files together is enough to double the elapsed time of a sub-second wait.

## Scope

Not a regression from anything in flight: the failures reproduce on `feat/harmonyos-emulator`, whose diff against `origin/main` touches neither the test nor the code under it — `git diff origin/main...HEAD -- packages/tool-server/test/flows/flow-idle-run.test.ts packages/tool-server/src/tools/flows/flow-actions.ts packages/tool-server/src/utils/timing.ts` is empty.

CI is not red today: `unit-tests.yml` runs `npm test --workspaces`, i.e. the whole tool-server package at once (349 files), which passes — the failure needs the narrower selection. It hits anyone who narrows a run to the flows suite, or who runs `npx vitest run` from the repo root, which is a common way to check a change.

## Impact

Four red tests unrelated to whatever the developer or agent is working on, with failure text ("expected 1671 to be less than 1000") that reads like a real timing regression in the idle wait. Every reader has to re-derive that it is a load artifact.

### LINKED ISSUE #1000: flow-type-focus: a 500ms wall-clock assertion fails at 499 under full-suite load
`packages/tool-server/test/flows/flow-type-focus.test.ts:165` asserts a wall-clock gap with no slack:

```ts
// The fixed settle still applies even without a focus-reporting source.
expect(keys[0]!.t - tap!.t).toBeGreaterThanOrEqual(500);
```

`t` is `Date.now()` recorded in the mock registry when each tool call arrives (`:48`). The gap it measures is one `sleepOrAbort(TYPE_FOCUS_SETTLE_MS, …)` — `flow-actions.ts:1499`, `TYPE_FOCUS_SETTLE_MS = 500` at `:221` — which is a bare `setTimeout(…, 500)` (`utils/timing.ts:6-19`). So the assertion is that a 500ms `setTimeout` plus two `Date.now()` reads spans at least 500 whole milliseconds, and it fails at 499 whenever the two reads land on either side of a millisecond boundary the timer did not clear.

## Repro

Six runs of the shard that contains it, on an otherwise-normal machine:

```
npx vitest run --shard=3/3     # from packages/tool-server
```

Two of the six failed, four passed. The failure is always the same:

```
 × skips the focus poll on a source that can't report focus 765ms
 FAIL  test/flows/flow-type-focus.test.ts > type directive focus wait > skips the focus poll on a source that can't report focus
AssertionError: expected 499 to be greater than or equal to 500
      Tests  1 failed | 1709 passed (1710)
```

The file passes on its own every time; it needs the load of a full shard.

## Why it matters

A full `packages/tool-server` run fails intermittently for a reason unrelated to whatever change is being tested, and the failure text names a timing constant, so the first reading is that the change under test slowed something down. `Unit tests` in CI runs the whole package, so this can redden an unrelated PR.

## Not the same as #795

[#795](https://github.com/software-mansion/argent/issues/795) is the same class — wall-clock assertions with no slack — but names `flow-idle-run.test.ts` and a different trigger (running the flows directory as a batch, four failures, one of them a 671ms overshoot). This is one assertion in a different file, off by a single millisecond, under full-suite load. Both are instances of "a real-timer assertion pinned to an exact boundary"; a fix for one does not reach the other.

## What the assertion is actually for

The comment says it: the fixed settle still applies on a source that cannot report focus, i.e. the code does not skip the sleep along with the poll. That is a claim about a branch being taken, and it survives a small tolerance — `toBeGreaterThanOrEqual(450)` still fails if the sleep is skipped (the gap collapses to single-digit ms) while passing on a boundary landing. The nearby assertions in the same file pin call counts and ordering rather than durations.

latekvo:
Real-timer assertions in the flows suite were pinned to exact wall-clock
boundaries, so a batch run reddened them for reasons unrelated to whatever was
being tested. Each now pins the same property with slack, or with no clock at
all — every one was mutation-tested to confirm it still fails loudly when the
behaviour it guards is removed.

Test-only; no docs change.

Fixes #1000
Fixes #795

<details>
<summary>Per-assertion: what it pins, and what changed</summary>

| Site | Pinned property | Before | After |
| --- | --- | --- | --- |
| `flow-type-focus` "skips the focus poll…" | the fixed 500ms settle still runs on a source that cannot report focus | `gap >= 500` | `gap >= 450` |
| `flow-type-focus` "waits for the tapped field…" | settle **plus** at least one 300ms focus poll | `gap >= 800` | `gap >= 720` |
| `flow-idle-run` "holds a still screen…" | `stableFor` is served, not ignored | run timed, `[750, 2400)` | span between first and last tree read, `>= 750` |
| `flow-idle-run` "settles inside the smallest timeout…" | the hold is counted across the polls, so the parser's floor is servable | run timed, `[750, 1000)` | read span `>= 750`; "inside the wait" carried by `warning === undefined` |
| `flow-idle-run` "does not call a slow but answering source a wedged one" | an abandoned read is not a wedge when the source's own slowest answer explains it | 2300ms reads / 4600ms wait, 5s default test timeout | 2800ms / 5400ms, own 20s timeout |
| `flow-idle-run` "does not report a tree-only settle…" | the tree-only settle owes the same hold | churn window clocked from before the run | clocked from the first read |
| `flow-idle-run` H1/H2 "settled early and then moved / went blank" | the tree-only verdict is not a write-once latch | 4 quiet reads in a 2500ms wait | 3 quiet reads in a 6000ms wait, reads asserted |

The two focus bounds had zero slack in the one direction that matters: a 500ms
`setTimeout` measured on `Date.now()` spans 499ms roughly once in 200 runs
(400 samples, idle machine: `499ms x2, 500ms x48, 501ms x289, 502ms x55,
503ms x2, 505ms x1`). 10% of slack keeps them failing by hundreds of
milliseconds if either wait is dropped.

The two idle `elapsed` bounds timed the whole run, so the flow's start-up —
parse, device resolve — counted against the step's own budget. Timing between
the reads the hold is actually measured across removes that, and the upper
bounds went with it: every way out of the wait except the settle carries a
warning, so `warning === undefined` already says it landed inside.

The remaining three are fixtures, not assertions: each assumed a number of poll
rounds, or a wall-clock window opened before the run, that a batch run does not
serve. They now count reads, or anchor the window on the first read.

</details>

<details>
<summary>Repro — before</summary>

`npx vitest run packages/tool-server/test/flows` on `origin/main`:

```
 FAIL  flow-idle-run.test.ts > settles inside the smallest timeout the parser allows for the hold
AssertionError: expected 2439 to be less than 1000

 FAIL  flow-idle-run.test.ts > does not call a slow but answering source a wedged one
Error: Test timed out in 5000ms.

 FAIL  flow-idle-run.test.ts > does not report a tree-only settle when the hold was never served
AssertionError: expected 'settled on the UI tree alone — this s…' not to contain 'UI tree alone'

 FAIL  flow-idle-run.test.ts > does not pass on a screen that settled early and then started moving again
AssertionError: expected 'settled on the UI tree alone — this s…' to contain 'never held still'
```

#1000's own intermittency did not surface in 10 rounds of
`npx vitest run --shard=3/3` (4 idle, 6 under a 10-way CPU load), so it was
forced instead: the `setTimeout`-vs-`Date.now()` measurement above is the whole
mechanism, and the assertion had no margin over it.

</details>

<details>
<summary>Repro — after</summary>

Six batch rounds of `npx vitest run packages/tool-server/test/flows`, on a
machine carrying other work (load average 22-30 throughout). Rounds 2-6 were
additionally narrowed to `--maxWorkers=3`, which is harsher than CI: the 13
`flow-composition` cases of #1067 pin one of the three workers for ~13 minutes
and starve the rest.

| round | workers | result for the assertions this PR touches |
| --- | --- | --- |
| 1 | default | all clean |
| 2 | 3 | H1 starved to 3 reads |
| 3 | 3 | all clean |
| 4 | 3 | all clean |
| 5 | 3 | all clean |
| 6 | 3 | H2 starved to 3 reads |

Every other failure in those rounds was the same 13 `flow-composition`
timeouts (#1067), plus load artifacts in `flow-selector-scopes`,
`flow-tool-step-tree-target-e2e`, `flow-launch-pins-tree-target` and
`flow-gesture-settle` — none of whose assertions this branch touches.

The two starved rounds were a real gap, not noise: H1 and H2 each need a
fourth read to reach a verdict at all, and
a 3500ms wait serves eighteen on an idle machine but three on a starved one.
The case then saw a screen that never moved and passed the wait cleanly, so the
assertion ran against `undefined`:

```
AssertionError: the given combination of arguments (undefined and string) is
invalid for this assertion.
```

Each case now budgets for the starved run and asserts the reads it was served,
so an environment that cannot reach the verdict says so:

```
AssertionError: reads served — under four, nothing moved to be judged:
expected 3 to be greater than or equal to 4
```

Confirmed by stretching each read to 1800ms, which starves the wait to three
reads without touching the wait itself.

</details>

<details>
<summary>Mutation tests (source broken, assertion must fail)</summary>

Each mutation was applied to `flow-actions.ts`, the test run, then reverted.

| Mutation | Assertion under test | Result |
| --- | --- | --- |
| delete `sleepOrAbort(TYPE_FOCUS_SETTLE_MS, …)` | both focus gaps | FAIL — `expected 0 to be greater than or equal to 450`; `expected 303 to be greater than or equal to 720` |
| focus poll sleeps `0` instead of `POLL_INTERVAL_MS` | `>= 720` | FAIL — `expected 504 to be greater than or equal to 720` |
| drop `&& heldForMs >= stableFor` from the settle | both idle read spans | FAIL — `expected 403 to be greater than or equal to 750` (both) |
| drop `slowestAnsweredReadMs +` from the hung check | "slow but answering" | FAIL — step errors, `expected false to be true` on `r.ok` |
| drop `&& now - treeSince >= stableFor` from `treeSettledAtLastRead` | "tree-only settle" | FAIL — `expected 'settled on the UI tree alone…' not to contain 'UI tree alone'` |
| `if (lastRead === "timeout")` without `&& treeReadHung` | "still warns, rather than erroring" | FAIL — `expected 'error' to be 'pass'` |
| make `treeSettledAtLastRead` a write-once latch | H1 "settled early then moved again" | FAIL — reports the tree-only settle |
| drop `treeSettledAtLastRead = false` from the blank-read branch | H2 "settled early then went blank" | FAIL — reports the tree-only settle |

The `treeReadHung` one did **not** fail at `timeout: 3000`: with the round
period dividing the wait exactly, drift decided whether the closing read was
abandoned at all. `3200` leaves the seventh read 200ms — less than the source
needs — so the case ends on an abandoned read by arithmetic.

</details>

<details>
<summary>Why the latch waits stop at 6000ms, and H3 is untouched</summary>

The first attempt at these two used a 12000ms wait, which passed CI and broke
the budget-tail case below them in the same file — that one has to fit three
reads into a 2000ms window, and with 24s of polling ahead of it it saw two:

```
expected 'the screen came back with content on 2 reads in 2000ms…'
  to contain 'never held still'
```

Deterministic, not load: 4/4 runs on the branch, 0/4 on `origin/main`, and it
passes in isolation. Bisected across the file:

| latch wait | whole file |
| --- | --- |
| 2500 (main) | 46 passed |
| 3500 | 46 passed |
| 6000 | 46 passed |
| 12000 | budget-tail fails every run |

So the latch cases take 6000ms and no more. H3 itself is left exactly as it is
on `main`: widening its wait scores its abandoned closing read as a wedged
source (`HUNG_TREE_READ_MS`) rather than a step that ran out of time, and
slowing its source starves it below the three reads a verdict takes — both
were tried against CI and both moved the failure rather than removing it.

</details>

<details>
<summary>Deliberately not changed</summary>

- `flow-idle-run`'s `elapsed < 3000` bounds at "honours its timeout", "rides
  over a dropped capture" and "stops on abort": 1.8–2.2s of slack over a
  sub-second expectation already clears a scheduling delay of hundreds of
  milliseconds.
- `flow-composition.test.ts` times out 13 cases at 60s each in the same batch
  run, on `origin/main` as well as here, while passing in 27s on its own. It is
  a batch artifact of a different kind and belongs in its own issue, not
  papered over from here.

</details>

### DIFF
--- a/packages/tool-server/test/flows/flow-idle-run.test.ts
+++ b/packages/tool-server/test/flows/flow-idle-run.test.ts
@@ -1274,10 +1274,10 @@ steps:
   it("does not pass on a screen that settled early and then started moving again", async () => {
     currentFrame = () => undefined; // force the tree-only path
     // Three still reads set the verdict — two agreeing intervals is what it
-    // takes — and everything after them moves. Counted in reads and given a
-    // wait that fits well over four of them: the case needs a moving read after
-    // the quiet stretch, and a loaded suite that fits only the quiet ones into
-    // the wait would hand it a screen that never moved at all.
+    // takes — and everything after them moves, so the case needs a fourth read
+    // to show the movement. A healthy 6000ms wait serves about thirty; the
+    // assertion on `reads` below is what separates a starved run, which saw no
+    // movement to judge, from a latch that failed to clear.
     let reads = 0;
     currentTree = () => {
       reads += 1;
@@ -1287,19 +1287,22 @@ steps:
       "ready",
       `executionPrerequisite: ""
 steps:
-  - await: { idle: true, timeout: 3500, stableFor: 0 }
+  - await: { idle: true, timeout: 6000, stableFor: 0 }
 `
     );
     const r = await run("ready");
+    expect(reads, "reads served — under four, nothing moved to be judged").toBeGreaterThanOrEqual(
+      4
+    );
     expect(r.steps.at(-1)!.warning).toContain("never held still");
-  });
+  }, 15_000);
 
   // H2: the same latch let a screen that had gone BLANK by the deadline report
   // ready. A blank tree is an observation, not a gap — it clears the verdict.
   it("does not pass on a screen that settled early and then went blank", async () => {
     currentFrame = () => undefined;
-    // Three still reads, then blank, and a wait sized for them the same way the
-    // case above sizes its own.
+    // Three still reads, then blank, sized and guarded the same way as the
+    // case above.
     let reads = 0;
     currentTree = () => {
       reads += 1;
@@ -1309,11 +1312,16 @@ steps:
       "ready",
       `executionPrerequisite: ""
 steps:
-  - await: { idle: true, timeout: 3500, stableFor: 0 }
+  - await: { idle: true, timeout: 6000, stableFor: 0 }
 `
     );
-    expect((await run("ready")).steps.at(-1)!.warning).toContain("never held still");
-  });
+    const r = await run("ready");
+    expect(
+      reads,
+      "reads served — under four, nothing went blank to be judged"
+    ).toBeGreaterThanOrEqual(4);
+    expect(r.steps.at(-1)!.warning).toContain("never held still");
+  }, 15_000);
 
   // H3: bounding the tree read by the remaining budget made the LAST read
   // time out on every run, which turned every honest timeout into an
--- a/packages/tool-server/test/flows/flow-idle-run.test.ts
+++ b/packages/tool-server/test/flows/flow-idle-run.test.ts
@@ -243,7 +243,18 @@ steps:
   // first read must still be held for it before the step returns. Without that
   // the option means nothing, since three reads take about 400ms whatever it
   // is set to.
+  //
+  // Timed between the READS the hold is measured across, not around the run:
+  // timing the run charges the hold with the flow's start-up as well, which a
+  // loaded suite stretches by seconds. That the settle landed inside the wait
+  // needs no clock at all — every way out of it except the settle carries a
+  // warning.
   it("holds a still screen for the whole requested hold before passing", async () => {
+    const readsAt: number[] = [];
+    currentTree = () => {
+      readsAt.push(Date.now());
+      return screenWith("Home");
+    };
     await writeFlow(
       "ready",
       `executionPrerequisite: ""
@@ -251,13 +262,10 @@ steps:
   - await: { idle: true, timeout: 2500, stableFor: 800 }
 `
     );
-    const started = Date.now();
     const r = await run("ready");
-    const elapsed = Date.now() - started;
     expect(r.steps.at(-1)).toMatchObject({ kind: "idle", status: "pass" });
     expect(r.steps.at(-1)!.warning).toBeUndefined();
-    expect(elapsed).toBeGreaterThanOrEqual(750);
-    expect(elapsed).toBeLessThan(2_400);
+    expect(readsAt.at(-1)! - readsAt[0]!).toBeGreaterThanOrEqual(750);
   });
 
   // The floor the parser enforces has to be one the runner can serve, or the
@@ -266,6 +274,11 @@ steps:
   // 1400ms, while the settle lands at ~820ms because the hold is counted
   // across the polls rather than after them.
   it("settles inside the smallest timeout the parser allows for the hold", async () => {
+    const readsAt: number[] = [];
+    currentTree = () => {
+      readsAt.push(Date.now());
+      return screenWith("Home");
+    };
     await writeFlow(
       "ready",
       `executionPrerequisite: ""
@@ -274,16 +287,16 @@ steps:
   - echo: reached
 `
     );
-    const started = Date.now();
     const r = await run("ready");
-    const elapsed = Date.now() - started;
     expect(r.ok).toBe(true);
     const step = r.steps.find((s) => s.kind === "idle")!;
     expect(step.status).toBe("pass");
-    // A clean settle, not a step that ran out of budget and warned its way out.
+    // A clean settle, not a step that ran out of budget and warned its way out
+    // — which is also what says it fitted inside the 1000ms wait.
     expect(step.warning).toBeUndefined();
-    expect(elapsed).toBeGreaterThanOrEqual(750);
-    expect(elapsed).toBeLessThan(1_000);
+    // And the hold was served rather than skipped; timed between the reads for
+    // the reason the case above gives.
+    expect(readsAt.at(-1)! - readsAt[0]!).toBeGreaterThanOrEqual(750);
     expect(r.steps.at(-1)).toMatchObject({ kind: "echo", status: "pass" });
   });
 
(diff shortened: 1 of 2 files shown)


latekvo:
@coderabbitai open a PR that modifies your config to stop appending this section to PR descriptions: "Summary by CodeRabbit"

PR #1084, opened 2026-09-09
fix(ios-remote): report a MoQ input send that never reached the device

hubgan:
## What

On the cloud (MoQ) path, every input send threw its promise away:

```ts
touch(opts)    { void moq.sendControl(encodeTouch({ … })); }
button(opts)   { void moq.sendControl(encodeButton({ … })); }
rotate(dir)    { void moq.sendControl(encodeRotate(dir)); }
pressKey(d, k) { void moq.sendControl(encodeKey({ … })); }
```

Those sends now return their promise, and `sendCommand` awaits it. A send the
session refuses becomes the same `SIMULATOR_COMMAND_TRANSPORT_FAILED` failure
the local WebSocket path already produces, so a caller reads one contract
rather than two. A command the MoQ transport does not implement rejects the
returned promise instead of throwing out of it, and is not reported as a refused
send. `screenshot` awaited already.

The remote instance also built its own `pressKey` straight onto
`moq.sendControl`, bypassing `sendCommand`, so `keyboard` on a dead session
threw a bare SDK error with no failure signal while a tap on the same session
reported a classified one. It now goes through `sendCommand`, like the attached
instance next to it. So does the Cmd+V chord `paste` sends after filling the
pasteboard, which awaited its sends but failed with the same bare error.

A screenshot on a dead session killed the server too, and that one could not be
fixed by awaiting: `MoqDeviceSession.screenshot` subscribes a track on first
use, guarded only by a `disposed` flag that an explicit `close()` sets - which a
released machine never does. `@moq/net` runs that subscribe as
`void #runSubscribe(...)`, so the `InvalidStateError` it throws is born detached
and no caller can attach a handler. Argent does not own the subscribe, but it
owns whether one is asked for: the session now closes itself when its `closed`
promise settles, so the SDK's own guard rejects the screenshot before any stream
is opened.

## Why

A dropped promise cost twice.

**The tool reported a success that never happened.** `gesture-tap` on a
released cloud session answered `{ tapped: true }` for an input that was never
written. That is the phantom success #932 removed from the WebSocket path,
reintroduced on MoQ: a QA flow can run green against a device that is no longer
there.

**And it killed the tool server.** With no handler, the rejection reached the
process-level `unhandledRejection` hook, which `index.ts` treats as fatal. The
server died asynchronously, detached from the call that caused it - the tap
reported success in the same millisecond the process began shutting down.

## Verification

E2E against a live cloud simulator: boot, tap, type, release the machine
underneath the open session, wait for the MoQ session to actually close, then
drive it again. Same script and same steps on both builds.

| | tap on the dead session | tool server | unhandled rejections |
| --- | --- | --- | --- |
| before | `{"tapped": true}` | dead | 3 |
| after | `SIMULATOR_COMMAND_TRANSPORT_FAILED` | alive, answered the next call | 0 |

The full sequence - tap, release the machine, wait for the session to close,
screenshot, tap - now runs to the end with the server still serving. It used to
die three ways: two discarded input sends and the detached screenshot subscribe.

`keyboard` on that dead session now fails the same way, where it used to throw
an unclassified `track is closed`. Typing on a live cloud session is
unaffected: `{"typed": "hi", "keys": 2}`.

The failure the tools now return:

```
The remote simulator did not accept the 'touch' command: track is closed.
The command was NOT delivered to the device. Call stop-simulator-server for this
device and retry - the next call opens a new cloud session. If the machine itself
was released, acquire one again before retrying.
```

A dead session recovers on the same machine: after `stop-simulator-server`, the
next input opens a new cloud session and reaches the device.

Tests: `moq-transport-errors.test.ts` covers a rejecting send per command,
`gesture-tap` failing instead of answering `{ tapped: true }`, the remote
instance routing its keys and its paste chord through `sendCommand`, an unknown
command rejecting unclassified, and - as the regression guard for the crash -
that no unhandled rejection is produced. `moq-session-dispose.test.ts` covers
the screenshot half, modelling the SDK's subscribe-on-first-use. The WebSocket
ack tests are unchanged and still pass.

Not addressed: the dead-session screenshot reports
`REGISTRY_TOOL_EXECUTION_FAILED` rather than a classified failure, exactly as it
did before this branch. Giving the screenshot path its own failure codes is a
separate change.

## Docs

No user-facing surface changes, so no docs update.


packages/tool-server/src/utils/moq-client.ts
```diff
@@ -35,3 +36,29 @@ async function openMoqClientFromInfo(info: MoqInfo): Promise<MoqClient> {
...
+function disposeWhenTransportDies(session: MoqDeviceSession): void {
+  const dispose = () => {
+    try {
+      session.close();
+    } catch {
+      // Tearing down a transport that is already gone. Nothing left to close.
+    }
+  };
+  // Both arms: `closed` rejects when the transport failed rather than ended.
+  void session.closed.then(dispose, dispose);
```
latekvo, line 63: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  The rejection arm of this handler is untested. `session.closed` is `connection.closed`, which maps the WebTransport session's `closed`, and the Node polyfill *rejects* it when the session closes before reaching `connected` (`@fails-components/webtransport/lib/session.js` `onClose`: `closedReject(new WebTransportError('Opening handshake failed.'))`; the Node client's `startUpConnection` also `closedReject`s on a failed connect). The fake in `moq-session-dispose.test.ts` models `closed` as a promise that only resolves (`killTransport` resolves), so every test in the file exercises the resolve arm only.

I confirmed the gap by mutation: changing this line to `void session.closed.then(dispose);` leaves all 13 tests across both new files green, while a probe fake whose `closed` rejects shows both consequences this PR fixes come back - `dispose` never runs, so `disposed` stays false and the first screenshot takes the unguarded subscribe path, and the derived `then(dispose)` rejection itself fires as an unhandled rejection, which `index.ts` treats as fatal. With the shipped code the same probe reports `disposed: true` and zero unhandled rejections, so the behavior is right today - nothing in the suite would notice if it stopped being.

packages/tool-server/src/utils/moq-client.ts
```diff
@@ -35,3 +36,29 @@ async function openMoqClientFromInfo(info: MoqInfo): Promise<MoqClient> {
...
+/**
+ * A cloud transport can die under a session nobody closed - the machine is
+ * released, the network drops - and the SDK does not mark the session disposed
+ * for it. Its first screenshot after that subscribes a track on a dead
+ * WebTransport, and `@moq/net` runs that subscribe detached (`void
+ * #runSubscribe`), so the `InvalidStateError` it throws lands as an unhandled
+ * rejection no caller can catch. `index.ts` treats one of those as fatal, so
+ * asking a released machine for a screenshot killed the tool server.
+ *
+ * Closing the session when its transport closes takes the SDK's own guarded
```
latekvo, line 49: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  This closes the crash for a screenshot issued after `dispose` has run, but a screenshot reaching `session.screenshot()` in the same tick the transport dies still passes the `disposed` guard and subscribes on the dead transport. I reproduced it with the real `@moq/net` `Subscriber` over a dead QUIC session and the real `MoqDeviceSession`: settle `closed` and call `screenshot()` in the same tick, and the subscribe is dequeued (`subscribe start: id=0 broadcast=simulator track=screenshot`), `Stream.open` throws `InvalidStateError` outside `#runSubscribe`'s try (`lite/subscriber.js` puts `const stream = await Stream.open(this.#quic)` before the method's `try`), and that rejection fires unhandled even with `disposeWhenTransportDies` in place - the SDK's own `closed` arm rejects the in-flight request ("MoQ connection closed"), but the detached subscribe rejection has no handler. One macrotask of separation (`setImmediate`) before the call is enough for the guard to hold.

So the fix narrows the crash window from "any time between transport death and explicit close" to a single microtask hop; it does not remove it. The verification's "0 unhandled rejections" holds for its sequence, which waits for the session to close before screenshotting - a screenshot racing the death tick can still kill the tool server.

packages/tool-server/test/moq-transport-errors.test.ts
```diff
@@ -0,0 +1,233 @@
...
+
+  it("still refuses a command the transport does not implement", async () => {
+    const transport = createMoqTransport(closedMoqClient(), { pasteText: () => Promise.resolve() });
+    await expect(async () => sendCommand(apiWith(transport), { cmd: "wiggle" })).rejects.toThrow(
+      /does not implement sendCommand cmd 'wiggle'/
+    );
+  });
+});
+
+describe("every send routed onto MoQ reports its own failure", () => {
```
latekvo, line 175: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  Paste is the fifth send on this transport and none of the cases here cover it. `transport.paste` awaits `options.pasteText` (which is the four-awaited `sendControl` Cmd+V chord in `buildRemoteInstance`, simulator-server.ts:100-110), and the `paste` tool's ios-remote branch awaits `transport.paste` (tools/paste/platforms/ios.ts:89). Every transport built from `closedMoqClient` in this file stubs `pasteText` to `() => Promise.resolve()`, so the refusing-chord shape is never driven through paste.

I checked what a regression there would ship: with the transport's `paste` re-dropping the chord's sends, one `paste` call on a dead session produces two unhandled rejections - the same crash class this PR removes from touch/button/rotate/key - with zero test failures. The fix itself is sound (the real path rejects with the chord's send failure; I probed that directly), but the suite's "every send routed onto MoQ" claim currently covers four of the five sends.

packages/tool-server/test/moq-transport-errors.test.ts
```diff
@@ -0,0 +1,233 @@
...
+
+/**
+ * Node emits `unhandledRejection` once the microtask queue has drained with no
+ * handler attached, which is exactly when the tool server's own handler would
+ * fire. Watching for it is the regression guard for the crash itself: the
+ * rejection being reported to the caller and the rejection not killing the
+ * process are two different fixes, and only this one catches a send that is
+ * awaited somewhere but still dropped on another branch.
+ */
+async function watchUnhandledRejections(run: () => Promise<unknown>): Promise<{
```
latekvo, line 79: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  These local `unhandledRejection` listeners are load-bearing in a way the comment doesn't say: vitest 4's worker handler withdraws while any other listener is attached (`vitest/dist/chunks/init.d4hAcNdp.js` checks `processListeners(event).length > 1` and returns - "assume that it's handled by user code"). I probed both directions: with a local listener attached, an escaping rejection is reported by nothing (the run exits 0); with the listener removed, vitest's own net reports "Unhandled Rejection" and fails the run.

The current assertions are adequate - both helpers record every reason and assert emptiness, and the `finally` removal re-arms vitest's net for anything escaping after the window. But any future test in this suite that attaches a listener like this and asserts something other than emptiness, or forgets the removal, silently disables vitest's global crash net for the whole file while it is attached.

packages/tool-server/src/utils/simulator-client.ts
```diff
@@ -209,15 +233,21 @@ function getOrCreateConnection(api: SimulatorServerApi): Connection {
...
+ * (a closed track, a released machine) rejects, while a send that goes out is
+ * reported as delivered. That is strictly narrower than the WebSocket ack, and
+ * it is still the difference between a failed gesture and `{ tapped: true }`.
  */
 export function sendCommand(api: SimulatorServerApi, cmd: Record<string, unknown>): Promise<void> {
+  const cmdName = typeof cmd.cmd === "string" ? cmd.cmd : "unknown";
   if (api.transport) {
-    routeViaTransport(api.transport, cmd);
-    return Promise.resolve();
+    return Promise.resolve(routeViaTransport(api.transport, cmd)).catch((cause: unknown) => {
```
latekvo, line 245: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  `routeViaTransport`'s default arm throws for an unknown `cmd.cmd`, and because it is evaluated inside `Promise.resolve(...)`'s argument, that throw escapes `sendCommand` synchronously instead of returning a rejected promise. I confirmed the bare-call shape: `void sendCommand(api, { cmd: "wiggle" })` throws `MoQ transport does not implement sendCommand cmd 'wiggle'` at the call site, where `sendCommand(api, cmd).catch(...)` without `await` would throw rather than route through the rejection path (and skip any recovery a rejection flows through).

No current caller can trigger it - every live caller of `sendCommand` and `api.pressKey` awaits, and no call site produces a cmd outside touch/button/rotate/key - so this is latent. The branch contracts now differ, though: the WebSocket branch always returns a pending promise; this branch can throw.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  Ran the full sweep: claims vs code, nearest twin (attached/spawned/remote pressKey paths), non-happy paths, inputs, reachability both ways, what outlives the call, and the absence pass (promise-dropping siblings, stale prose, routeViaTransport/transport symmetry), plus mutation passes across the new branches and the dispose hook.

Reproductions run:
- The PR's two new test files replayed onto the parent commit (6f9e5e868): 7/10 and 2/3 fail there, matching the PR's discrimination claims; 13/13 pass on the head.
- Mutations against the head: sendCommand dropping the await (7 tests fail), routeViaTransport dropping the touch arm (3 fail), dropping the button arm (2 fail), the remote pressKey reverted to a direct `moq.sendControl` (1 fail) - each caught by specific assertions.
- Probes against the real `@swmansion/argent-cloud-sdk` 0.1.3, `@moq/net` `Subscriber`, and the `@fails-components` WebTransport polyfill for the dispose-on-death behavior: steady state (screenshot after the dispose microtask) is guarded, `disposed: true`, zero unhandled rejections; a same-tick racing screenshot still arms the detached `#runSubscribe` rejection (details on that thread).

Five findings below (no high-severity issues: the core fix holds, and every mutation that matters is caught). The tool-server suite was also run on the head (433 files / 6209 tests pass; the 3 flow-script failures under my environment are pre-existing and fail identically with that setup on main), and typecheck is clean apart from one pre-existing unrelated error present on the base branch.

packages/tool-server/src/utils/moq-client.ts
latekvo, line 63: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  Still present as of 7bf7f3eb9 - re-verified with a fresh repro on the current head: a probe fake whose `closed` rejects (matching the @fails-components behavior where a session that dies pre-connect closes with an error, @fails-components/webtransport/lib/session.js `onClose`) shows that with the shipped code `dispose` still runs and the screenshot stays guarded - so the code is correct today, and nothing else in this suite would notice if that stopped being true. The fake in moq-session-dispose.test.ts still models `closed` as resolve-only (killTransport resolves), so the reject arm remains untested.

packages/tool-server/src/utils/moq-client.ts
latekvo, line 49: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  Still present as of 7bf7f3eb9 - re-reproduced on the current head with an asserting probe against the real SDK structure (`void #runSubscribe` with `Stream.open` before its try): killTransport() then session.screenshot() in the same synchronous tick leaves `disposed` false at the guard, arms the detached subscribe, and records exactly 1 unhandled `InvalidStateError: Session is failed or closed and can not open streams` - the error index.ts fatalizes. Separating the call by one macrotask (setImmediate) yields 0. The fix still narrows the window to a microtask hop rather than removing it.

packages/tool-server/test/moq-transport-errors.test.ts
latekvo, line 175: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  Still present as of 7bf7f3eb9, with a sharper number: the real paste chord at simulator-server.ts:100-110 is four sequential awaited sends, so the claimed regression shape (drops the awaits, reverts to raw `void` sends) produces **4** unhandled rejections per paste, not 2 - and the awaited shape yields 0 because each await propagates the rejection. Both measured by probe on this head. The COMMANDS list here still covers touch/button/rotate/key only and every transport stub in this file still pins `pasteText: () => Promise.resolve()`, so that regression would still ship with zero test failures.

packages/tool-server/test/moq-transport-errors.test.ts
latekvo, line 79: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  Still present as of 7bf7f3eb9 - re-confirmed against vitest 4.1.9 on this head: with a local `unhandledRejection` listener attached (as both helpers here do), a genuinely escaping rejection is seen only by the local listener and the run exits 0; with the listener removed, vitest's net catches the identical rejection and fails the run (node_modules/vitest/dist/chunks/init.d4hAcNdp.js withdraws its worker handler when another listener exists). These helpers are therefore load-bearing as written - the concern about future misuse stands.

packages/tool-server/src/utils/simulator-client.ts
latekvo, line 245: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  Still present as of 7bf7f3eb9 - re-reproduced with a direct call (no async arrow, plain try/catch at the call site): `sendCommand(api, { cmd: "wiggle" })` throws synchronously (`syncThrew: true`, `promiseRejected: false`) out of routeViaTransport's default arm, and `getFailureSignal(err)` on the thrown error is null - the rejection path's classification never runs. Latent (no current caller can hit it) but the branch contracts still differ from the WebSocket arm, and the existing test here masks the shape by wrapping the call in an async arrow.

latekvo (commented):
\[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  Second sweep on this PR. The five threads from the earlier review were re-checked against the current head (7bf7f3eb9) with fresh reproductions, plus a full re-run of the review moves - claims vs code, nearest twin, non-happy paths, inputs, reachability both ways, lifetime, and the absence pass - and mutation probes across the new branches and the dispose hook.

What was run this time:
- Both new test files on the head (13/13 green) and the dispose hook probed with a REJECTING `closed` fake against both the head and the parent commit (6f9e5e868): the shipped code disposes on the reject arm too (disposed: true, screenshot guarded, zero unhandled rejections) and 2/2 fail on the parent - so Finding 1's code-correctness question is settled, and the reject-arm coverage gap it reported is confirmed still open (details on that thread).
- The same-tick racing screenshot from Finding 2 re-reproduced with an asserting probe: exactly 1 unhandled InvalidStateError escapes when the screenshot lands in the same tick as transport death, 0 when separated by a macrotask - still open as written (thread unchanged).
- Finding 3 re-verified on the current head: the paste chord at simulator-server.ts:100-110 still awaits four raw `moq.sendControl` calls, the COMMANDS list still omits paste, and the regression probe now counts 4 unhandled rejections for the dropped-await shape (the awaited shape yields 0 - the awaits are load-bearing, so the missing coverage matters more than the original comment's "2").
- Finding 4's suppression mechanics re-confirmed against vitest 4.1.9 on this head: with a local `unhandledRejection` listener attached, an escaping rejection is reported by nothing and the run exits 0; without it, vitest's net catches the same rejection and fails the run.
- Finding 5 re-reproduced with a direct (non-async-arrow) call: `sendCommand(api, { cmd: "wiggle" })` throws synchronously out of `routeViaTransport`, `getFailureSignal(err)` is null, and `promiseRejected` is false - the error bypasses `remoteTransportError` entirely.

New on this pass: the tool-server suite is green on the head (436 files / 6212 tests pass, 1 skipped), the PR's five files pass prettier, and the one `tsc --noEmit -p tsconfig.test.json` error (flow-remote-tree.test.ts:85, `probeAnswered` missing) is present on the parent commit and on the base branch, so it is pre-existing and not from this PR. No new findings beyond the threads below; the prior verification's mutation matrix (sendCommand await, touch/button/router arms, remote pressKey) was not re-run and those threads already carry the mutations.

packages/tool-server/src/utils/moq-client.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  Still present as of 7bf7f3eb9. Re-ran the probe on the current head: a fake whose `closed` rejects (the @fails-components pre-connect shape) shows `dispose` still runs with the shipped code - `disposed: true`, screenshot guarded, zero unhandled rejections - so the code is correct today, and replaying both new test files against the parent commit 6f9e5e868 fails 2/2 there. The coverage gap stands: the fake in moq-session-dispose.test.ts is still resolve-only.

packages/tool-server/src/utils/moq-client.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  Still present as of 7bf7f3eb9. Re-reproduced on the current head with an asserting probe: same-tick `killTransport()` then `screenshot()` records exactly 1 unhandled `InvalidStateError: Session is failed or closed and can not open streams` (the error index.ts fatalizes); one macrotask of separation records 0. The window is unchanged.

packages/tool-server/test/moq-transport-errors.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  Still present as of 7bf7f3eb9, with a corrected count. The real chord at simulator-server.ts:100-110 is four sequential awaited sends, and a probe of the regression shape (awaits dropped back to raw `void` sends) measures **4** unhandled rejections per paste on a dead session - the original comment's "2" undercounts, and the awaited shape measures 0 because each await propagates the rejection. Coverage is still absent: no test here drives pasteText with a rejecting sendControl.

packages/tool-server/test/moq-transport-errors.test.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  Still present as of 7bf7f3eb9. Re-confirmed against vitest 4.1.9 on this head: with a local `unhandledRejection` listener attached, an escaping rejection is seen only by the local listener and the run exits 0; with the listener removed, vitest's net catches the identical rejection and fails the run.

packages/tool-server/src/utils/simulator-client.ts
latekvo: \[[Diplomat](https://github.com/latekvo/Diplomat), GLM 5.3 Flash\]:  Still present as of 7bf7f3eb9. Re-reproduced with a direct call (plain try/catch at the call site): `syncThrew: true`, `promiseRejected: false`, and `getFailureSignal` on the thrown error is null - the classification never runs, and the existing test still masks the shape by wrapping the call in an async arrow.

packages/tool-server/src/utils/moq-client.ts
hubgan: Fixed in 5ac546c73. The fake session can now fail its transport, and a new test asserts that the session closes on the reject arm with no unhandled rejection. With the line mutated back to `.then(dispose)`, that test fails.

For context: on a live cloud machine, `closed` resolved when the machine was released (`{"closeCode":0,"reason":"relay closed"}`), so this arm protects against a transport that follows the WebTransport spec, not the Node polyfill in use today.

hubgan: Fixed in 5ac546c73. The fake session can now fail its transport, and a new test asserts that the session closes on the reject arm with no unhandled rejection. With the line mutated back to `.then(dispose)`, that test fails.

For context: on a live cloud machine, `closed` resolved when the machine was released (`{"closeCode":0,"reason":"relay closed"}`), so this arm protects against a transport that follows the WebTransport spec, not the Node polyfill in use today.

packages/tool-server/test/moq-transport-errors.test.ts
hubgan: Fixed in d9db4cc7c. Both helpers now say that vitest's own net stands down while they are attached, so they must record every reason, the caller must assert the list is empty, and `finally` must detach them.

Reproduced the mechanism first against vitest 4.1.9: with a local listener that asserts nothing, an escaping rejection leaves the run green (exit 0). Without the listener, vitest reports "Unhandled Rejection" and the run fails.

hubgan: Fixed in d9db4cc7c. Both helpers now say that vitest's own net stands down while they are attached, so they must record every reason, the caller must assert the list is empty, and `finally` must detach them.

Reproduced the mechanism first against vitest 4.1.9: with a local listener that asserts nothing, an escaping rejection leaves the run green (exit 0). Without the listener, vitest reports "Unhandled Rejection" and the run fails.

packages/tool-server/src/utils/simulator-client.ts
hubgan: Not changed in this PR: the behavior predates it. On the parent commit 6f9e5e868 the MoQ branch was `routeViaTransport(api.transport, cmd); return Promise.resolve();`, which throws synchronously for an unknown cmd too. The same direct-call probe gives `syncThrew: true`, `promiseRejected: false` and a `null` signal on both commits.

Routing that throw through the new `.catch` would also be wrong for this case: it would label a programming error as `SIMULATOR_COMMAND_TRANSPORT_FAILED` and tell the caller to acquire a new machine.

hubgan: Not changed in this PR: the behavior predates it. On the parent commit 6f9e5e868 the MoQ branch was `routeViaTransport(api.transport, cmd); return Promise.resolve();`, which throws synchronously for an unknown cmd too. The same direct-call probe gives `syncThrew: true`, `promiseRejected: false` and a `null` signal on both commits.

Routing that throw through the new `.catch` would also be wrong for this case: it would label a programming error as `SIMULATOR_COMMAND_TRANSPORT_FAILED` and tell the caller to acquire a new machine.

packages/tool-server/src/utils/moq-client.ts
hubgan: Not changed in this PR: the same-tick case predates it. The same probe on the parent commit 6f9e5e868 also subscribes once and leaks one `InvalidStateError`. This PR narrows the window; it does not open it.

On a live cloud machine a tool call cannot reach this window. The close arrives from `UDP.onMessage`, and the polyfill's `onClose` sets the state to `closed` and resolves `closed` in that same callback. `dispose` runs two microtasks later, before any other macrotask starts. Every screenshot argent asks for starts from an HTTP request or a timer, so it runs at least one macrotask later. You measured that separation as enough.

E2E on a released cloud machine, same script on both builds, with the first screenshot of the session taken after the transport died:

- parent commit: the screenshot leaked `InvalidStateError: Session is failed or closed and can not open streams`, and the run had 5 unhandled rejections in total. `index.ts` treats each one as fatal.
- this PR: the screenshot was refused with `MoQ session is closed`, with 0 unhandled rejections.

One gap remains, and the parent commit has it too: a first screenshot whose subscribe is still in flight when the transport dies. `Stream.open` and the writes before the `try` in `@moq/net`'s `#runSubscribe` can then reject with no handler. Closing that needs a change in the library.

hubgan: Not changed in this PR: the same-tick case predates it. The same probe on the parent commit 6f9e5e868 also subscribes once and leaks one `InvalidStateError`. This PR narrows the window; it does not open it.

On a live cloud machine a tool call cannot reach this window. The close arrives from `UDP.onMessage`, and the polyfill's `onClose` sets the state to `closed` and resolves `closed` in that same callback. `dispose` runs two microtasks later, before any other macrotask starts. Every screenshot argent asks for starts from an HTTP request or a timer, so it runs at least one macrotask later. You measured that separation as enough.

E2E on a released cloud machine, same script on both builds, with the first screenshot of the session taken after the transport died:

- parent commit: the screenshot leaked `InvalidStateError: Session is failed or closed and can not open streams`, and the run had 5 unhandled rejections in total. `index.ts` treats each one as fatal.
- this PR: the screenshot was refused with `MoQ session is closed`, with 0 unhandled rejections.

One gap remains, and the parent commit has it too: a first screenshot whose subscribe is still in flight when the transport dies. `Stream.open` and the writes before the `try` in `@moq/net`'s `#runSubscribe` can then reject with no handler. Closing that needs a change in the library.

packages/tool-server/test/moq-transport-errors.test.ts
hubgan: Not changed in this PR: the gap predates it. This PR does not touch the paste chord in `buildRemoteInstance` or `transport.paste` (blame: 20f5a1282 and b23211442). Both already awaited every send on the parent commit, and no test covered them there either. The `COMMANDS` table covers what `sendCommand` routes through `routeViaTransport`: the four sends this PR changed.

On a live cloud machine the chord was never reached. `paste` failed first in `sim-remote simctl pbcopy`, on both builds, before and after the machine was released. This PR's run had 0 unhandled rejections.

hubgan: Not changed in this PR: the gap predates it. This PR does not touch the paste chord in `buildRemoteInstance` or `transport.paste` (blame: 20f5a1282 and b23211442). Both already awaited every send on the parent commit, and no test covered them there either. The `COMMANDS` table covers what `sendCommand` routes through `routeViaTransport`: the four sends this PR changed.

On a live cloud machine the chord was never reached. `paste` failed first in `sim-remote simctl pbcopy`, on both builds, before and after the machine was released. This PR's run had 0 unhandled rejections.

PR #1085, opened 2026-09-09
fix(windows): pass windowsHide on every console-subsystem spawn the tool-server makes

anasalaqeel:
## Summary

On Windows, every Android tool call — gesture-tap, screenshot, describe, list-devices, boot-device — pops a console window. Root cause: the tool-server is a detached background process with no console of its own, and it spawns console-subsystem binaries (`adb.exe`, `emulator.exe`) without `windowsHide`, which is exactly the parent/child shape Windows allocates a fresh popup console for.

- `utils/adb.ts` is the chokepoint nearly all Android tools route through (`runAdb`, `runAdbBinary`, `listAvds`, `checkSnapshotLoadable`, `emulatorSupportsFlag`) — none of its 5 `execFileAsync` call sites passed `windowsHide`.
- Three more spots use `spawn(..., { detached: true, ... })`: `boot-device.ts` (emulator launch), `boot-electron.ts` (Electron/Chromium launch), `update-argent.ts` (self-update). Node's own docs note a detached child gets its own console window on Windows unless `windowsHide` is also set — same bug, different call shape.
- `android-devtools.ts`'s `am instrument` spawn (the accessibility-helper injection behind every Android `describe`) and `android-profiler/capture.ts`'s raw adb spawn (bypasses `runAdb` for its stdin/stdout streaming) both spawn `adb` directly, outside `adb.ts`.

`windowsHide` is a documented no-op on macOS/Linux, so this only changes behavior on win32.

## Test plan

- [x] `npm run build` — clean
- [x] `npm test -w @argent/tool-server -- adb-hardening android-adb` — 21/21 pass, including 3 new tests
- [x] Added regression tests (`windowsHide (Windows console-popup fix)` in `adb-hardening.test.ts`) pinning that `runAdb`, `listAvds`, and `checkSnapshotLoadable` pass `windowsHide: true` to `execFile`, so a future edit can't silently drop it
- [x] Full `npm test -w @argent/tool-server` — same 180 pre-existing failures on this branch as on a clean `main` checkout (verified via `git stash` A/B on the same machine — all Windows-environment issues unrelated to this change, mostly Unix-domain-socket tests that don't run on win32); zero new failures
- [x] `npx prettier --check` on all changed files — clean
- [x] `npm run knip` (both passes) — clean
- [x] Manually reproduced against a real Android emulator on Windows 11 (Claude Code + this tool-server) before the fix — every `gesture-tap`/`screenshot`/`describe` call visibly flashed a console window and stole focus

Claude-Session: https://claude.ai/code/session_015SGhQu3bo4AetFZaxMpxER

### DIFF
--- a/packages/argent-tools-client/src/launcher.ts
+++ b/packages/argent-tools-client/src/launcher.ts
@@ -310,6 +310,12 @@ export function spawnToolsServer(
       detached: true,
       stdio: ["ignore", "pipe", logFd],
       env: buildToolsServerEnv(paths, port, process.env, options),
+      // `detached: true` on Windows gives node.exe (a console-subsystem
+      // binary) its own console window — this is the ONE persistent popup
+      // that stays open for the tool-server's whole lifetime, distinct from
+      // the many short-lived adb.exe popups fixed in @argent/tool-server.
+      // No-op off Windows.
+      windowsHide: true,
     });
 
     child.unref();
--- a/packages/tool-server/src/blueprints/android-devtools.ts
+++ b/packages/tool-server/src/blueprints/android-devtools.ts
@@ -88,7 +88,14 @@ async function spawnHelper(serial: string): Promise<SpawnedHelper> {
   const proc = spawn(
     adbPath,
     ["-s", serial, "shell", "am", "instrument", "-w", manifest.instrumentationRunner],
-    { stdio: ["ignore", "pipe", "pipe"] }
+    {
+      stdio: ["ignore", "pipe", "pipe"],
+      // No-op off Windows. On Windows this is the adb spawn behind every
+      // Android `describe` call (this helper backs the accessibility tree) —
+      // without it, each one pops its own console window. See utils/adb.ts
+      // for the same fix on the rest of the adb call sites.
+      windowsHide: true,
+    }
   );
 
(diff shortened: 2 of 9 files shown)


anasalaqeel:
Pushed a second fix: the tool-server's own process launch — `spawnToolsServer` in `packages/argent-tools-client/src/launcher.ts` — spawns `node.exe` with `detached: true` and no `windowsHide`, which is a separate instance of the same bug class. Found this from live testing: after installing the first fix, the many short-lived adb.exe popups were gone, but one persistent console window remained for the tool-server's own lifetime — traced it to this spawn.

This function backs both the MCP auto-spawn path (bundled into `mcp-server.mjs`) and `argent server start --detach` (via `runDetached` in `argent-cli/src/server.ts`), so one fix covers both.

Caveat, for reviewers testing this locally: an already-running `argent mcp` bridge process has the old `spawnToolsServer` cached in its module memory (Node doesn't hot-reload), so installing the fix doesn't change behavior for that process until it's restarted — only a fresh `argent mcp` start, or `argent server start --detach` run directly, exercises the new code path. I haven't yet gotten a clean-process confirmation that this second fix actually stops the persistent window; noted honestly in the commit message.

PR #1092, opened 2026-09-09
fix(tool-server): refuse a non-iOS device id passed to boot-device as `udid`

### LINKED ISSUE #1078: boot-device accepts a chromium-cdp-<port> udid through its declared chromium capability and boots it as an iOS simulator
`boot-device` declares `chromium: { app: true }` (`packages/tool-server/src/tools/devices/boot-device.ts:1305-1311`) so the capability gate accepts a Chromium device id, but the `execute` dispatch never tests for one. With a `udid`, the chain is:

```ts
if (isExternalId(...) || claim) { ...refuse... }
if (classifyDevice(params.udid!) === "ios-remote") return bootIosRemote(...);
return bootIos(params.udid!, registry, params.force, params.headless);
```

`classifyDevice("chromium-cdp-9222")` returns `"chromium"`, not `"ios-remote"`, so the call falls through to `bootIos` and reaches `xcrun simctl boot chromium-cdp-9222` on macOS. The declared capability exists for the `electronAppPath` form, which carries no `udid` at all.

**Impact.** The agent gets a raw simulator error ("Invalid device") for a Chromium id, which reads as a broken simulator rather than as "boot-device does not take a Chromium id". It is also the one tool an agent reaches for after a Chromium app is gone, so the wrong-looking error lands exactly where the recovery starts.

**Repro**

```bash
curl -s -X POST localhost:$ARGENT_PORT/tools/boot-device \
  -H 'content-type: application/json' \
  -d '{"udid":"chromium-cdp-9222"}'
```

The capability gate passes (chromium is declared), and the failure comes back from the iOS boot path.

**Expected.** A `udid` whose `classifyDevice` is `"chromium"` should be refused by `boot-device` with a message naming `electronAppPath` as the Chromium form, in the same shape as the external-provider refusal directly above it.

Found while reviewing #703; unrelated to that PR's diff.

latekvo:
Fixes #1078

**Cause.** `boot-device`'s capability declaration covers Chromium/Vega/Android because those platforms have their own parameters, so the capability gate lets their device ids through on `udid` - and `execute` only special-cased `ios-remote`, sending everything else to `simctl boot`.

**Fix.** Classify the `udid` first and refuse anything that is not an iOS simulator UDID, naming the parameter that boots that platform (`electronAppPath` / `vvdImage` / `avdName`). 13 lines, no new failure code - it reuses the `BOOT_DEVICE_TARGET_SELECTION_INVALID` shape already used for the exactly-one-selector check.

The issue reports only the Chromium id; Vega (`amazon-*`) and Android serials reached `bootIos` the same way and are refused together.

**Docs.** No change needed: no tool, CLI, config key or flow-file surface moved - only the error text on a rejected input.

<details>

The new cases in `packages/tool-server/test/boot-device.test.ts` fail on the unfixed source - with the source change stashed, a `chromium-cdp-9222` udid runs all the way through the mocked `xcrun simctl boot` / `bootstatus` and dies later in native-devtools priming:

```
 × refuses chromium-cdp-9222 and names the parameter that boots it
 × refuses amazon-4a27df03c9777152 and names the parameter that boots it
 × refuses emulator-5554 and names the parameter that boots it

AssertionError: expected [Function] to throw error matching
  /`udid` takes an iOS simulator UDID/ but got 'ndApi.reverifyEnv is not a function'
```

With the fix restored the same tests pass and `execFile` is never called.

Ran from the worktree root:

- `npx tsc --build` - clean
- `npm run typecheck:tests -w packages/tool-server` - clean
- `npm test -w packages/tool-server` - 368 files, 4835 passed, 1 skipped
- `npx prettier --check` on both changed files - clean
- `npx eslint` on both changed files - no findings (the only output is the known `packages/docs` tsconfig resolution error present on `main`)

</details>


latekvo:
Waiting for coderabbit merging once it's done.

PR #1096, opened 2026-09-09
fix(cli): keep the batch tally when a directory run's artifact export throws

### LINKED ISSUE #1026: A directory run's artifact export sits outside the loop's try, so an export throw loses the batch summary, the skip entries and the --json aggregate
`runFlowDirectory` catches everything the `flow-execute` call can raise, and guards a reply that is not a report — so a bad flow gets its outcome line, the batch either continues or stops deliberately, and the run always ends with a tally. One call sits outside that protection.

`packages/argent-cli/src/flow.ts:1146`:

```ts
    // Key exports by the flow's subdirectory so recursive same-stem flows
    // cannot clobber each other (exportFailureArtifacts keys by stem only).
    await exportAndResolveArtifacts(
      report,
      outputBase ? path.join(outputBase, path.dirname(rel)) : undefined,
      path.join(dir, rel),
      baseUrl
    );
```

It is after the `try`/`catch` that wraps `callTool` and after the `!report` guard, so a throw from `baseUrl()` or `exportFailureArtifacts` escapes `runFlowDirectory` entirely. The loop never reaches `results.push`, the remaining flows are never marked skipped, and `exitAfterFlush` never runs, so the batch summary and the `--json` aggregate are both lost.

## Repro

Stub `exportFailureArtifacts` to throw and run a two-flow directory with `--output`:

```
stdout: ["[1/2] a-login.yaml"]
stderr: []
rejection: artifact fetch died
```

With `--json`, stdout is empty - a consumer piping into `jq` gets a parse error rather than the aggregate. The error surfaces at `packages/argent/src/cli.ts:160` as `console.error(err)` and `exit(1)`, the same exit code an ordinary flow failure produces, but with a stack trace where the ledger should be.

## The sibling that handles it

The single-flow runner wraps the identical call, `packages/argent-cli/src/flow.ts:1517`:

```ts
  try {
    await exportAndResolveArtifacts(...);
  } catch (err) {
    if (!args.jsonStream) throw err;
    return fail(err instanceof Error ? err.message : String(err), 2, err);
  }
```

## Trigger

`argent flow run <dir> --output <dir>` where the artifact fetch fails: the tool-server dies between the run and the export, the output directory is not writable, or a handle 404s. The failure lands on the CI path that `--output` exists to serve.

latekvo:
Fixes #1026

**Cause:** the artifact export in `runFlowDirectory` sat outside the loop's `try`, so a throw from `baseUrl()`/`exportFailureArtifacts` escaped the function - the batch lost its skip entries, its summary and the `--json` aggregate.

**Fix:** wrap the call the way the single-flow runner already does; the flow is recorded `fail`, the batch stops (an export failure is the same wall for every remaining flow) and the loop still emits the tally.

Docs: none needed - `reference/flow-yaml.mdx` already states "An infrastructure error stops the batch and counts the remaining flows as skipped"; the code now matches it.

<details>

Two tests in `packages/argent-cli/test/flow.test.ts` make `baseUrl` reject during a `--output` run. Reverting only the source fix:

```
FAIL  test/flow.test.ts > stops the batch on an artifact-export throw instead of losing the tally
  expected [Function] to throw error including 'process.exit:1' but got 'artifact fetch died'
FAIL  test/flow.test.ts > emits the --json aggregate when artifact export throws
  expected [Function] to throw error including 'process.exit:1' but got 'artifact fetch died'
```

With the fix restored:

```
npx tsc --build                      # clean
npm run typecheck:tests -w @argent/cli  # clean
npm test -w @argent/cli              # 27 files, 529 tests passed
npx prettier --write <changed files>
npx eslint packages/argent-cli/src/flow.ts  # clean
```

Class check: `exportAndResolveArtifacts` has exactly two production call sites - this one and the single-flow runner, which was already wrapped.
</details>

### DIFF
--- a/packages/argent-cli/src/flow.ts
+++ b/packages/argent-cli/src/flow.ts
@@ -963,8 +963,8 @@ interface BatchFlowResult {
  * Run every discovered flow in `dir` sequentially. Reports failures only (no
  * live step lines), then a flow-level summary. A flow failing its steps — or
  * one the tool-server rejects as invalid — lets the batch continue, while an
- * infra error (transport throw, unclassified failure, non-report result) stops
- * it and counts the remaining flows skipped.
+ * infra error (transport throw, unclassified failure, non-report result, failed
+ * artifact export) stops it and counts the remaining flows skipped.
  */
 async function runFlowDirectory(
   dir: string,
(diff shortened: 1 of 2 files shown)


packages/argent-cli/src/flow.ts
```diff
@@ -1107,46 +1013,45 @@ async function runFlowDirectory(
...
+      // A dead tool-server or an unwritable --output directory is the same
+      // wall for every remaining flow, so stop the batch rather than run them
+      // into it.
+      const message = err instanceof Error ? err.message : String(err);
+      console.error(message);
+      results.push({ path: rel, status: "fail", error: message });
+      stopped = true;
+      continue;
+    }
     results.push({ path: rel, status: report.ok ? "pass" : "fail", report });
```
latekvo, line 1055: This looks like a duplicate. Won't this yield two identical `fail` results?

PR #1098, opened 2026-09-09
fix(cli): keep flow help off stdout under --json

### LINKED ISSUE #1025: argent flow run --json prints the usage text on stdout for a flag error, while --json-stream keeps stdout machine-readable
`argent flow run --json` promises a machine-readable stdout: on a successful run it holds the report and nothing else, and on a directory run the aggregate document alone. Three early-exit paths break that. `printHelp(jsonStream)` is called with the *streaming* flag only, so with `--json` the usage text goes to `console.log`:

- `packages/argent-cli/src/flow.ts:1210` - `--help`/`-h`
- `packages/argent-cli/src/flow.ts:1226` - a `FlagParseException`
- `packages/argent-cli/src/flow.ts:1236` - a missing flow reference

`printHelp` (`packages/argent-cli/src/flow.ts:85`) writes to `console.log` unless `toStderr` is set.

## Repro

Driving `flow()` from the built CLI, capturing stdout only:

```
$ node drive.mjs run --json --platfrom ios 2>/dev/null | wc -c
    3500
$ node drive.mjs run --json --platfrom ios 2>/dev/null | head -1
Usage: argent flow <subcommand> [options]
```

Exit code 2, and the actual diagnostic (`Error: unknown flag --platfrom`) is on stderr. `| jq` gets 3.5 KB of prose.

## The sibling that gets it right

`--json-stream` on the same input:

```
$ node drive.mjs run --json-stream --platfrom ios 2>/dev/null
{"event":"error","error":"unknown flag --platfrom"}
```

stdout is one NDJSON record; the 3532-byte help goes to stderr. The comment at `packages/argent-cli/src/flow.ts:1188` states the rule - "Once streaming is requested stdout belongs exclusively to NDJSON, so help goes to stderr as the diagnostic it is" - and `--json` makes the same promise about stdout without getting the same treatment.

## Impact

A CI job that pipes `argent flow run <dir> --json` into a parser and mistypes a flag gets a parse error whose text is a usage screen, not the flag error. Any wrapper that treats stdout as JSON has to special-case it.

latekvo:
Fixes #1025

**Cause:** the three early-exit paths in `argent flow run` passed only the `--json-stream` flag to `printHelp`, so under `--json` the 3.4 KB usage screen went to `console.log`.

**Fix:** compute the "help is a diagnostic" flag from `--json` as well, and pass it to all three `printHelp` calls.

**Verified:** driving the built CLI, stdout for `run --json --platfrom ios` goes from 3404 bytes of usage text to 0; four new tests cover `--help`, a flag-parse error, `--json=` carrying a value, and a missing flow reference, and all four fail on the unfixed code.

Sibling `--json` commands (`config`, `flags`, `secrets`, `run`) already print usage only on an explicit `--help` and send errors to stderr, so nothing else had this shape. No docs change: `flow-yaml.mdx` documents `--json` as "print the raw JSON report", which is what it now does.

<details>
<summary>Repro and checks</summary>

Before (dist patched back to `printHelp(jsonStream)`):

```
$ node drive.mjs run --json --platfrom ios 2>/dev/null | wc -c
    3404
$ node drive.mjs run --json --platfrom ios 2>/dev/null | head -1
Usage: argent flow <subcommand> [options]
```

After:

```
$ node drive.mjs run --json --platfrom ios 2>/dev/null | wc -c
       0
$ node drive.mjs run --json --platfrom ios 2>&1 1>/dev/null | head -1
Error: Unknown flag: --platfrom
```

Test discrimination - with `printHelp(helpToStderr)` reverted to `printHelp(jsonStream)`:

```
 × keeps help off stdout for a bad flag under --json
 × keeps help off stdout for --json= carrying a value
 × keeps help off stdout for --json --help
 × keeps help off stdout for a --json run given no flow
 Tests  4 failed | 1 passed
```

Checks run from the worktree root:

- `npx tsc --build` - clean
- `npm run typecheck:tests -w @argent/cli` - clean
- `npm test -w @argent/cli` - 27 files, 531 tests passed
- `npx prettier --write` on both changed files
- `npx eslint packages/argent-cli/src/flow.ts` - clean

</details>

### DIFF
--- a/packages/argent-cli/src/flow.ts
+++ b/packages/argent-cli/src/flow.ts
@@ -1092,16 +1092,18 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
     return exitAfterFlush(2);
   }
 
-  // Once streaming is requested stdout belongs exclusively to NDJSON, so help
-  // goes to stderr as the diagnostic it is.
+  // Either flag claims stdout for machine-readable output, so help goes to
+  // stderr as the diagnostic it is.
   const jsonStream = rest.some(
     (tok) => tok === "--json-stream" || tok.startsWith("--json-stream=")
   );
+  const helpToStderr =
+    jsonStream || rest.some((tok) => tok === "--json" || tok.startsWith("--json="));
   // Checked before parseRunArgs so --help wins even when it trails a
   // value-taking flag (`--device --help` would otherwise throw "requires a
   // value" instead of printing help).
   if (rest.includes("--help") || rest.includes("-h")) {
-    printHelp(jsonStream);
+    printHelp(helpToStderr);
     return;
   }
   const fail = (message: string, code: number, err: unknown = message): Promise<never> => {
@@ -1117,7 +1119,7 @@ export async function flow(argv: string[], options: FlowCommandOptions): Promise
     if (err instanceof FlagParseException) {
       if (jsonStream) writeJsonStreamError(err);
       console.error(`Error: ${err.message}\n`);
-      printHelp(jsonStream);
+      printHelp(helpToStderr);
       return exitAfterFlush(2);
     }
     throw err;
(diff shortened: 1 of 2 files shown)


latekvo:
To re-validate

PR #1100, opened 2026-09-09
fix(settings-permissions): 400, not 500, for a permission the platform cannot map

### LINKED ISSUE #1032: settings-permissions returns 500 for a permission the platform has no equivalent for, so the MCP client retries something no retry can fix
\[[Diplomat](https://github.com/latekvo/Diplomat), Fable 5.1\]: Found while reviewing #566, which hits the same fork and resolves it the other way. Out of scope there.

Both `settings-permissions` platform arms reject an unmappable permission with a bare `FailureError` carrying `error_kind: "unsupported"`:

- `packages/tool-server/src/tools/settings-permissions/platforms/android.ts:113` — `Permission '<p>' has no Android runtime-permission equivalent…`
- `packages/tool-server/src/tools/settings-permissions/platforms/ios.ts:115` — `Permission '<p>' cannot be changed on the iOS simulator…`

`http.ts` keys the status on the **error class**, never on `error_kind` (`packages/tool-server/src/http.ts:889-919`): `UnsupportedOperationError` and `InvalidToolInputError` map to 400, `NotImplementedOnPlatformError` to 501, and everything else — including a `FailureError` whose signal says `unsupported` — falls through to **500**.

So a permission this build can never change is reported to the agent as an internal server fault. It is caller input: no retry, no different device and no wait satisfies it, and the agent's own error handling reads 5xx as the retryable class.

**Repro**
```
POST /tools/settings-permissions {"udid":"<android serial>","action":"grant","permission":"notifications"}
→ 500 {"error":"Permission 'notifications' has no Android runtime-permission equivalent, so there is nothing to grant.", "error_kind":"unsupported"}
```

**Same shape elsewhere:** `packages/tool-server/src/tools/boot-device.ts:391` and `:1201`.

**The pattern that gets it right:** `keyboard/simulator-server-keys.ts:41-44` ("Caller input error → 400 via the error class, so the granular … code survives (#420)") and, on #566's branch, `system-settings` — both throw `InvalidToolInputError` / `UnsupportedOperationError` with the granular `error_code` on the signal, which lands 400 while keeping the telemetry bucket.

latekvo:
Fixes #1032

**Cause** - both `settings-permissions` platform arms rejected an unmappable permission with a bare `FailureError`. The HTTP dispatcher keys the status on the error *class*, never on `error_kind`, so it fell through to **500** and the agent read caller input as the retryable class.

**Fix** - throw `InvalidToolInputError` instead (same message, same `SETTINGS_PERMISSION_UNSUPPORTED` signal), which the dispatcher maps to **400** while keeping the granular telemetry bucket. Same shape as the keyboard backends' "no keycode for this character" arm.

Both sites of this class inside `settings-permissions` are fixed (iOS + Android; there is no third platform arm). The two sites the issue also names in `devices/boot-device.ts` are left alone: `BOOT_IOS_UNSUPPORTED_HOST` (iOS asked for on a non-macOS host) and `VEGA_ALREADY_RUNNING` (a device-state conflict a `force:true` retry does resolve) each need a different class - a design call in its own right, not this fix.

No docs change: `packages/docs` documents what the tool does, not per-tool HTTP status codes.

<details>

Discriminating test - reverted just the two source files to `main` and re-ran the suite:

```
FAIL test/settings-permissions.test.ts > settings-permissions iOS branch > notifications is rejected as unsupported without calling simctl
  expected FailureError: Permission 'notifications' … to be an instance of InvalidToolInputError
FAIL test/settings-permissions.test.ts > settings-permissions Android branch > reminders is rejected as unsupported (no Android equivalent)
  expected FailureError: Permission 'reminders' has … to be an instance of InvalidToolInputError
Tests  2 failed | 59 passed (61)
```

With the fix restored: `61 passed (61)`.

Also run from the worktree root:

- `npx tsc --build` - clean
- `npm test -w @argent/tool-server` - full package suite green
- `npx prettier --check` on the three changed files - clean
</details>

## Summary by CodeRabbit

- **Bug Fixes**
  - Unsupported Android and iOS permissions are now reported as invalid input errors, providing more accurate error classification while preserving existing error details.
  - Existing rejection behavior and error codes remain unchanged.

- **Tests**
  - Added coverage to verify that unsupported Android and iOS permissions return the expected invalid-input error type.

### DIFF
--- a/packages/tool-server/src/tools/settings-permissions/platforms/android.ts
+++ b/packages/tool-server/src/tools/settings-permissions/platforms/android.ts
@@ -1,4 +1,5 @@
 import { FAILURE_CODES, FailureError, getFailureSignal } from "@argent/registry";
+import { InvalidToolInputError } from "../../../utils/capability";
 import type { PlatformImpl } from "../../../utils/cross-platform-tool";
 import { adbShell, isTerminalAdbError, shellQuote } from "../../../utils/adb";
 import type {
@@ -136,7 +137,7 @@ export const androidImpl: PlatformImpl<
 
     const permissions = permissionsFor(permission, action);
     if (permissions.length === 0) {
-      throw new FailureError(
+      throw new InvalidToolInputError(
         `Permission '${permission}' has no Android runtime-permission equivalent, so there is nothing to ${action}.`,
         {
           error_code: FAILURE_CODES.SETTINGS_PERMISSION_UNSUPPORTED,
--- a/packages/tool-server/src/tools/settings-permissions/platforms/ios.ts
+++ b/packages/tool-server/src/tools/settings-permissions/platforms/ios.ts
@@ -1,6 +1,7 @@
 import { execFile } from "node:child_process";
 import { promisify } from "node:util";
 import { FAILURE_CODES, FailureError, subprocessFailureMetadata } from "@argent/registry";
+import { InvalidToolInputError } from "../../../utils/capability";
 import type { PlatformImpl } from "../../../utils/cross-platform-tool";
 import { simctlArgsForUdid } from "../../../utils/ios-device-sets";
 import { simctlPrivacy as remoteSimctlPrivacy } from "../../../utils/sim-remote";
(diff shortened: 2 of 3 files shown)


hubgan (approved):

latekvo:
Nitpick, minimally better alignment to the available errors.

PR #1101, opened 2026-09-09
fix(tool-server): admit a digit-leading bundleId in the four app-lifecycle schemas

### LINKED ISSUE #1024: bundleId schemas reject a digit-leading id with a message that says digits are allowed, in four copies of the same regex
`launch-app`, `restart-app`, `reinstall-app` and `settings-permissions` each carry their own copy of

```js
const BUNDLE_ID_PATTERN = /^[A-Za-z_][A-Za-z0-9._-]*$/;
```

and reject anything whose **first** character is a digit, with the message

> bundleId may only contain letters, digits, '.', '_' and '-'

which says digits are allowed. An agent handed `9gag.app` reads a rule its input satisfies and has nothing to act on.

Observed, running the four regexes as written:

```
9gag.app             REJECTED -> "bundleId may only contain letters, digits, '.', '_' and '-'"
3d.tools.app         REJECTED -> same
com.example.app      ACCEPTED
_foo.bar             ACCEPTED
```

The head restriction has a stated reason, at `launch-app/index.ts:15`:

> The head is restricted so a bundleId like `--user` can't masquerade as a flag inside `am start -n …` / `cmd package resolve-activity …`.

That reason requires excluding `-`. It does not require excluding digits, which cannot begin a flag. Android package names cannot start with a digit (each segment is a Java identifier), but iOS bundle ids can: `CFBundleIdentifier`'s documented charset is alphanumerics, hyphen and period, with no constraint on the first character. So on iOS the four tools refuse such an app outright — including `restart-app`, which is the remedy several native-devtools states prescribe by name.

Two separate things to fix:

1. The head class excludes digits for no reason the comment gives. `[A-Za-z0-9_]` keeps the flag-injection guard intact. (`native-devtools`'s handshake validator already uses that class, deliberately, so the socket admits ids these four schemas reject — noted while reviewing #712, which is where this surfaced.)
2. The message describes the alphabet but not the head rule, so any rejection it produces is unactionable. It should name the constraint it actually enforces.

The regex and its message are also duplicated verbatim across the four tools rather than shared, which is what let the message and the pattern drift apart in the first place. Related: #901, on the same constant being enforced on Chromium where the description calls the value an arbitrary tag.

latekvo:
Fixes #1024

**Cause.** `launch-app`, `restart-app`, `reinstall-app`, `settings-permissions` and `open-url` each carried their own copy of `/^[A-Za-z_][A-Za-z0-9._-]*$/`, which rejects an iOS bundle id such as `9gag.app` - with a message that says digits are allowed, so the agent reading it has nothing to act on.

**Fix.** One shared `BUNDLE_ID_PATTERN` / `BUNDLE_ID_MESSAGE` in `src/utils/bundle-id.ts`. The head class becomes `[A-Za-z0-9_]` (the same class `native-devtools`' handshake validator already uses); `-` and `.` stay out, which is the whole of the flag-injection guard the old comments cited. The message names the head rule it enforces.

**Class check.** No copy of the pattern remains (`grep BUNDLE_ID_PATTERN`). `open-url` was a fifth copy that landed on main while this branch was open - the catalogue-wide test caught it. `ACTIVITY_PATTERN` in `launch-app` / `restart-app` keeps its letter-only head deliberately: an activity is a Java class name and cannot start with a digit.

**Docs.** No docs change needed - `reference/tools.mdx` lists these tools by one-line summary and documents no parameter constraints.

<details>

New `test/bundle-id-pattern.test.ts` walks the whole catalogue rather than one tool, so a sixth `bundleId` constraint has to satisfy the same rules. Discrimination proved by restoring the old pattern and message in the shared module and re-running:

```
 × derives a sane MCP JSON schema with bundleId required for every action
 × launch-app: admits a digit-leading bundle id and still refuses a flag
 × launch-app: names the head rule when it rejects one
 × restart-app / reinstall-app / settings-permissions / open-url: both, each
 Tests  11 failed | 61 passed (72)
```

With the fix in place, from the worktree root:

```
npx tsc --build                                  OK
npm run typecheck:tests -w @argent/tool-server   OK
npm test -w @argent/tool-server                  434 files, 6174 passed | 1 skipped
npx prettier --write <changed files>             unchanged
npx knip                                         bundle-id.ts absent from the report
```

`npx eslint` could not run in this checkout (`packages/docs/tsconfig.json` cannot resolve `@docusaurus/tsconfig` without a separate `npm ci` in that workspace, a pre-existing local-environment issue); the ESLint CI job passes.

</details>


latekvo:
Simple hardcoded string extraction into global const.

Issue #631, opened 2026-07-31
debugger-component-tree: `maxNodes` is not enforced and `onScreenOnly: false` returns the same tree as the default


filip131311:
## Two related gaps

**`maxNodes` not enforced.** `{maxNodes: 5}` against a 20-node tree returned the full tree. This is
consistent with the documented mechanism (collapse single-child wrapper chains — there were none here),
but the schema promises "Maximum total nodes to include", so a caller sizing a response cannot rely on it.

**`onScreenOnly: false` is a no-op.** The response was **byte-identical** to the default
(`onScreenOnly: true`) on a screen with a scrollable list, where off-screen content demonstrably exists —
the same session's `includeSkipped: true` run reported 314 fibers walked and 207 JS-skipped.

## Expected

Either honour both parameters, or document the conditions under which they don't apply (and say so in the
response).

## Environment

Expo SDK 54 / RN 0.81.5, Hermes, bridgeless, React Compiler enabled.

---
*Found during a full-surface QA sweep of **0.18.1** (`swmansion-argent-0.18.1.tgz`; the global install
is byte-identical to the tarball). Matrix: iPhone 17 Pro + iPhone 17 / iOS 26.5, Pixel_9 / Android 16
(API 36), Apple TV 4K / tvOS 26.5, Electron 42 (Chrome 148), Linux aarch64 (Lima) incl. the remote-link
leg, and the deviceless CLI. Host: macOS Darwin 25.5.0, Node v22.23.1.*

filip131311:
Worked through both halves on the same stack (Expo SDK 54 / RN 0.81.5, Hermes, bridgeless, React Compiler on). **The first is real; the second isn't — `onScreenOnly` works.**

### `maxNodes` — confirmed

`{maxNodes: 5}` returned a 24-node tree here too. `debugger-component-tree.ts:316-354` only ever collapses single-child wrapper chains of length ≥ 2 — it never drops a node and never truncates, so with no qualifying chain it does nothing at all.

There's a self-defeating interaction worth noting: the TS-side filter already removes content-free wrappers *before* this runs (the same run reports `Content-free wrapper: 47`), so the very nodes `maxNodes` wants to collapse have largely been deleted by an earlier pass. That's why it so reliably no-ops. Your reading of the schema is right — "Maximum total nodes to include" promises a hard cap, and the implementation is a best-effort compaction hint.

### `onScreenOnly: false` — not reproducible; the parameter is working

[...]

PR #1111, opened 2026-09-09
docs(skills): split includeSkipped guidance per tool in argent-metro-debugger

### LINKED ISSUE #928: argent-metro-debugger documents includeSkipped as the way to find a component missing from debugger-component-tree, where it only appends a bounded per-name count block

`packages/skills/skills/argent-metro-debugger/SKILL.md:68-72`:

> ### `includeSkipped` guidance
> **Applies to both** `debugger-component-tree` and `debugger-inspect-element`. Set to `true` only when debugging filter behavior — e.g., **an expected component is missing from output**, or you need to **inspect a very specific branch of the tree** (not just an overview).
> > **Warning:** Output can be very large. Always combine with `maxNodes` (component-tree) … Do not use `includeSkipped` without a limit on large apps.

It holds for `debugger-inspect-element`, which really does keep the filtered items (`packages/tool-server/src/tools/debugger/debugger-inspect-element.ts:35-58`, `skipped: true` + `skipReason`). It is wrong for `debugger-component-tree` in both halves.

## What component-tree's `includeSkipped` actually does

Nothing is added to the tree. `packages/tool-server/src/tools/debugger/debugger-component-tree.ts:429-483` appends a fixed counts-only block, and the JS side never emits a skipped fiber at all — `packages/tool-server/src/utils/debugger/scripts/component-tree.ts:322-323` increments `skippedCounts[name]` in the `else` arm of the `if (!skip)` that does the pushing, and `:438-441` returns only the totals.

## Repro

Same synthetic screen, `includeSkipped: false` then `true`:

```
--- includeSkipped: true ---
Screen: 393x852

HomeScreen (tap: 0.50,0.47)
  SubmitButton "Submit" [testID=submit] (tap: 0.50,0.38)

--- Filtered ---
Total fibers walked: 314
JS-side skipped: 207 (View: 120, RCTView: 61, Provider: 26)

(includeSkipped block adds 4 lines)
```

Four lines. The block is bounded by construction: one header, one `Total fibers walked`, one `JS-side skipped` capped at the top 10 names (`:448-450`), one `TS-side removed` and at most seven detail lines — ~12 lines at the absolute maximum, regardless of app size. So the "output can be very large" warning and the "do not use it without a limit" rule describe a risk that cannot occur on this tool, while the stated use cases (find the missing component, inspect a specific branch) are exactly the two things it cannot do.

## The knob that does control this is never named

The parameter that decides what is in the tree is `onScreenOnly` (`debugger-component-tree.ts:495-502`, default `true`). It appears nowhere in `SKILL.md` and nowhere in the tool's own description (`:529-545`) — only in its zod `.describe()`. Note it is separately reported as a no-op in practice by #631, so an agent that does find it has a second problem.

## Expected

Split the section: for `debugger-component-tree`, say `includeSkipped` appends a bounded per-name summary of what was pruned and names no node, point "a component is missing" at `onScreenOnly: false`, and drop the size warning. Keep the current text for `debugger-inspect-element`.

An agent that cannot find a component in the tree does what line 70 says: re-runs with `includeSkipped: true`, plus a `maxNodes` it was told to pair with it. It gets the identical tree back with four extra lines saying `View: 120`. The component it is looking for is not named, its position is not given, and the counts tell it nothing actionable.

From there it concludes the component is not mounted — the one wrong conclusion the flag was documented to prevent — and starts debugging the app instead of the filter. The `maxNodes` it added on the strength of the size warning can additionally collapse wrapper chains (`:316-356`) and shrink the tree it was trying to widen.

latekvo:
Fixes #928

**Cause** - one `includeSkipped` section covered both tools, but on `debugger-component-tree` the flag only appends a counts-only "--- Filtered ---" block: it adds no node and names no component, so "an expected component is missing" sent agents to a flag that cannot surface it (and paired with a `maxNodes` that shrinks the tree further), while the "output can be very large" warning describes a block capped at ~12 lines by construction.

**Fix** - split the section. `component-tree`: says the summary is bounded and names nothing, and routes a missing component to `onScreenOnly: false` (or, for JS-side pruning, a `testID` / `inspect-element`). `inspect-element`: keeps the existing retained-items text and the size warning, now scoped to `maxItems`.

**Verified** - claims re-read against source; `npx tsc --build` and `prettier --check` green.

<details>
<summary>Source backing each claim</summary>

- counts-only block, no node added: `packages/tool-server/src/tools/debugger/debugger-component-tree.ts:402-455` builds `--- Filtered ---` from `totalFibers` / `skippedCounts` / `filterStats` only.
- bounded: header + `Total fibers walked` + `JS-side skipped` (top 10 names on one line, `:425-427`) + `TS-side removed` + at most 7 detail lines.
- `TS-side removed -> Off-screen` is the one category `onScreenOnly` controls: `:90` gates the off-screen subtree prune on `opts.onScreenOnly` (default `true`, `:468-475`).
- JS-side pruning is unreachable from any parameter: `packages/tool-server/src/utils/debugger/scripts/component-tree.ts:262-275` (inactive `activityState` / `isPageFocused` subtrees), `:296-300` (host fibers with no `testID`), `:322-323` (the `else` arm that only increments `skippedCounts`).
- `inspect-element` really does retain filtered items: `packages/tool-server/src/tools/debugger/debugger-inspect-element.ts:32-34` (`skipped: true` + `skipReason`); `maxItems` default is 35 (`:137-141`).

No test: the change is prose in a skill file - `packages/skills` has no test suite and no code reads `SKILL.md`.

No docs update needed: no tool, CLI, config key, flow file or user-facing capability changed. `includeSkipped` appears in no page under `packages/docs/`.

Class check: `grep -rn includeSkipped --include='*.md'` matches only this file, so there is no sibling copy of the wrong guidance.

Out of scope, left alone: #631 (`onScreenOnly: false` reported as a no-op on a screen whose pruning was all JS-side) and the fact that the tool description does not name `onScreenOnly` outside its zod `.describe()`.
</details>

### DIFF
--- a/packages/skills/skills/argent-metro-debugger/SKILL.md
+++ b/packages/skills/skills/argent-metro-debugger/SKILL.md
@@ -67,9 +67,16 @@ Both can point to source files, but `inspect-element` is purpose-built for sourc
 
 ### `includeSkipped` guidance
 
-Applies to both `debugger-component-tree` and `debugger-inspect-element`. Set to `true` only when debugging filter behavior — e.g., an expected component is missing from output, or you need to inspect a very specific branch of the tree (not just an overview).
+**`debugger-component-tree`** — appends a bounded summary (about a dozen lines, whatever the app size) of what was pruned: total fibers walked, JS-side skip counts by component name, and TS-side filter-pass totals. It adds no node to the tree and never names the component you are looking for, so it tells you which filter removed things, not what was removed.
 
-> **Warning:** Output can be very large. Always combine with `maxNodes` (component-tree) or `maxItems` (inspect-element) and increase it incrementally (e.g., start at 50, then grow). Do not use `includeSkipped` without a limit on large apps.
+When a component you expect is missing from the tree:
+
+- scrolled or otherwise positioned off-canvas (`TS-side removed → Off-screen`) — re-run with `onScreenOnly: false`, which is the parameter that decides what the tree contains.
+- pruned JS-side (host views with no `testID`, framework wrappers, inactive navigation stacks) — no parameter brings it back. Give it a `testID`, or use `debugger-inspect-element` at its coordinates.
+
+**`debugger-inspect-element`** — filtered items stay in the result carrying `skipped: true` and a `skipReason`. Set to `true` only when debugging filter behavior — e.g., an expected component is missing from output, or you need to inspect a very specific branch of the hierarchy (not just an overview).
+
+> **Warning:** Output can be very large. Always combine with `maxItems` and increase it incrementally (e.g., start at 35, then grow). Do not use `includeSkipped` without a limit on large apps.
 
 ---
 


latekvo:
Severe bloat. Will open a separate PR in the future, skipping for now.

PR #1112, opened 2026-09-09
docs(skills): describe what debugger-component-tree actually renders

### LINKED ISSUE #927: argent-metro-debugger promises debugger-component-tree returns bounding rects; the renderer emits only a normalized centre point

`packages/skills/skills/argent-metro-debugger/SKILL.md:50`:

> `debugger-component-tree` | Full React fiber tree (names, depth, **bounding rects**, tap coordinates).

Two of the four are wrong.

## No bounding rects are emitted

`packages/tool-server/src/tools/debugger/debugger-component-tree.ts:365-376` is the whole renderer for a node:

```ts
function formatLabel(c: RawEntry): string {
  let label = c.name;
  const displayText = c.text ?? c.accLabel;
  if (displayText) label += ` "${displayText}"`;
  if (c.testID) label += ` [testID=${c.testID}]`;
  if (c.rect && canNormalize) {
    const tapX = ((c.rect.x + c.rect.w / 2) / screenW).toFixed(2);
    const tapY = ((c.rect.y + c.rect.h / 2) / screenH).toFixed(2);
    label += ` (tap: ${tapX},${tapY})`;
  }
  return label;
}
```

`c.rect` is read only to compute a normalized centre point. Its `x`/`y`/`w`/`h` reach the output nowhere. Elsewhere in the file `rect` is used purely for filtering — overlap dedup (`:28-38`), off-screen pruning (`:112-124`), full-screen-wrapper collapse (`:162-176`), content-free-wrapper collapse (`:207-219`). The only geometry in the entire response is the header at `:360-363`, `Screen: <W>x<H>`.

## "Full" is also wrong

Seven TS-side filter passes plus the JS-side skip rules prune the tree; the tool's own description says so (`:530-531`: "off-screen (scrolled) content, full-screen transparent wrappers, and implementation-detail components are pruned").

## Repro

Driving `buildTextTree` with a two-node screen:

```
Screen: 393x852

HomeScreen (tap: 0.50,0.47)
  SubmitButton "Submit" [testID=submit] (tap: 0.50,0.38)
```

Each node carried `rect: { x: 40, y: 300, w: 313, h: 48 }` on the way in. Nothing of it survives except the centre.

## Expected

The row should describe what is rendered: a pruned tree of on-screen components with name, text/accessibilityLabel, testID and a normalized tap centre — and no per-node geometry. `depth` is conveyed by indentation and is fine.

Any question about size or overlap routes here: "is that button clipped at the bottom", "do the two cards overlap", "how wide is the header". `SKILL.md:63` reinforces it — component-tree is "Best for: **Layout overview**". An agent reads line 50, calls the tool expecting rects, and gets a list of centre points.

Two outcomes, both bad: it back-computes dimensions it does not have (a centre point plus `Screen: WxH` cannot yield width or height), or it reports the layout question as unanswerable while `screenshot` + `describe` would have answered it. There is no signal in the response that the promised field is simply absent rather than empty for this screen.

latekvo:
Fixes #927

`argent-metro-debugger/SKILL.md` sold `debugger-component-tree` as a "Full React fiber tree (names, depth, **bounding rects**, tap coordinates)". `formatLabel` reads each node's `rect` only to derive a normalized center - `x`/`y`/`w`/`h` reach the output nowhere - and seven filter passes prune the tree. Layout questions ("is this clipped", "do these overlap") were routed to a tool that cannot answer them.

Prose only, one file.

| | Before | After |
| --- | --- | --- |
| Tool table | Full React fiber tree (names, depth, bounding rects, tap coordinates). | Pruned tree of the on-screen React components: name, text/accessibilityLabel, testID, and a normalized tap center. Carries no bounding rects; for size, overlap or clipping use `describe`, which does return them. |
| Best-for row | Layout overview; finding tap targets; user-defined component hierarchy | Finding tap targets; user-defined component hierarchy |
| Quick reference | Full component tree | On-screen component tree |

**Docs:** no update needed - `packages/docs/docs/reference/tools.mdx` already says "compact React component text tree", and no tool, CLI, config key or capability changed.

<details>
<summary>Repro and verification</summary>

Drove `buildTextTree` with two nodes carrying `rect: {x:40, y:300, w:313, h:48}` on a 393x852 screen:

```
Screen: 393x852

HomeScreen (tap: 0.50,0.38)
  SubmitButton "Submit" [testID=submit] (tap: 0.50,0.38)
```

Nothing of `w`/`h` survives; the only geometry in the whole response is the `Screen: WxH` header. `describe` by contrast renders `(x, y, width, height)` per node (`packages/tool-server/src/tools/describe/format-tree.ts` `fmtFrame`), so the redirect is real.

`packages/skills` ships markdown only and has no test suite, so no discriminating test applies. `npx prettier --check` and `npx tsc --build` green.

**Class check:** every other mention of the tool (`argent-device-interact`, `argent-test-ui-flow`, `argent-react-native-app-workflow`, `rules/argent.md`, the tool's own `description`) already says names/text/testID/tap coords. This row was the only one claiming geometry.

</details>

## Summary by CodeRabbit

- **Documentation**
  - Updated the debugger component tree documentation to describe its pruned on-screen tree output.
  - Clarified that tap centers are normalized and bounding rectangles are not included.
  - Recommended using `describe` for size, overlap, and clipping details.
  - Updated comparison and quick-reference tables to reflect the revised behavior.

### DIFF
--- a/packages/skills/skills/argent-metro-debugger/SKILL.md
+++ b/packages/skills/skills/argent-metro-debugger/SKILL.md
@@ -45,12 +45,12 @@ With two or more devices on one Metro, `debugger-connect` refuses a udid/serial 
 
 ### Inspection & console
 
-| Tool                       | Purpose                                                                                                                                                                                                              |
-| -------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `debugger-component-tree`  | Full React fiber tree (names, depth, bounding rects, tap coordinates).                                                                                                                                               |
-| `debugger-inspect-element` | Inspect at (x, y) using **logical pixel coordinates** (not normalized 0-1): component hierarchy with source file:line and code fragment. See `references/source-maps.md`.                                            |
-| `debugger-log-registry`    | Get log summary (counts, clusters, file path). Then use `Grep`/`Read` on the flat log file for details. If it returns `status: "not_connected"`, there is **no** `file` — follow its `guidance` instead of grepping. |
-| `debugger-evaluate`        | Run a JS expression in the app runtime.                                                                                                                                                                              |
+| Tool                       | Purpose                                                                                                                                                                                                                       |
+| -------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `debugger-component-tree`  | Pruned tree of the on-screen React components: name, text/accessibilityLabel, testID, and a normalized tap center. Carries no per-node geometry; for size, overlap or clipping use `describe`, which returns a bounding rect. |
+| `debugger-inspect-element` | Inspect at (x, y) using **logical pixel coordinates** (not normalized 0-1): component hierarchy with source file:line and code fragment. See `references/source-maps.md`.                                                     |
+| `debugger-log-registry`    | Get log summary (counts, clusters, file path). Then use `Grep`/`Read` on the flat log file for details. If it returns `status: "not_connected"`, there is **no** `file` — follow its `guidance` instead of grepping.          |
+| `debugger-evaluate`        | Run a JS expression in the app runtime.                                                                                                                                                                                       |
 
 ---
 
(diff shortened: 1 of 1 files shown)


latekvo:
Re-verified and it holds up, but will have to shorten it.

PR #1114, opened 2026-09-09
fix(tool-server): give the CDP binding timeout the same recovery guidance as its sibling

### LINKED ISSUE #947: cdp-client: DEBUGGER_CDP_BINDING_TIMEOUT carries no recovery guidance while its sibling timeout got the dual-branch essay
\[[Diplomat](https://github.com/latekvo/Diplomat), X Preview F Free\]: Found during review of #703 (pre-existing on main; adjacent to the REQUEST_TIMEOUT rewrite the PR lands).

**The defect.** `cdp-client.ts` rewrote the `REQUEST_TIMEOUT` message into a full paused-vs-hung recovery with per-platform relaunch branches. The same file's other timeout surface, `DEBUGGER_CDP_BINDING_TIMEOUT` (~line 381: `Binding response for requestId=... timed out`), stays bare - and by the PR's own model it is the one timeout where *paused* is genuinely possible: `evaluateWithBinding` evaluates with `awaitPromise:false`, which a paused V8 answers, and then the binding never fires.

So the surface where "ask the user to resume" matters most is the one carrying zero guidance, and `DEBUGGER_CDP_BINDING_TIMEOUT` is unmapped in `NOT_CONNECTED_CODE_MAP`.

**Impact:** an agent hitting a binding timeout on a paused renderer gets a dead-end error string instead of the resume/check-first advice the sibling timeout now gives.

latekvo:
Fixes #947

**Cause.** `DEBUGGER_CDP_BINDING_TIMEOUT` rejected with a bare `Binding response for requestId=... timed out`, while the sibling `DEBUGGER_CDP_REQUEST_TIMEOUT` in the same file carries a full paused-vs-frozen recovery. It is the timeout where *paused* matters most: `evaluateWithBinding` dispatches the script with `awaitPromise: false`, so a runtime stopped at a breakpoint accepts the evaluate and never runs the callback.

**Fix.** Give the binding timeout the same shape of guidance - check/resume the debugger, else restart the app, do not retry in a loop (each attempt burns the full timeout).

**Deliberately not done:** the issue also suggests mapping the code into `NOT_CONNECTED_CODE_MAP`. That map is only consulted by `debugger-status` / `debugger-log-registry`, and neither calls `evaluateWithBinding` (only `debugger-component-tree` and `debugger-inspect-element` do), so the entry would be unreachable. The message is the only surface an agent actually sees.

**Class check:** `evaluateWithBinding` is the only other timeout surface in `cdp-client.ts`; `chromium-server/cdp-session.ts` has no bare timeout message. The wording stays platform-neutral ("restart the app", not `restart-app`) because `cdp-client` is shared with the Chromium blueprint.

**Docs:** none needed - no tool, CLI flag, config key or flow directive changed; this is an error-message string.

<details>

Test added to `packages/tool-server/test/metro/cdp-client.test.ts`: the fake CDP server answers the evaluate and never fires the binding, then the rejection is asserted for the code and each guidance clause.

Discriminates - with the source change stashed:

```
FAIL test/metro/cdp-client.test.ts > CDPClient > failure signal classification >
     an unanswered binding rejects with DEBUGGER_CDP_BINDING_TIMEOUT and paused-vs-frozen recovery
AssertionError: expected 'Binding response for requestId=req-t …' to match /paused at a breakpoint/
+ Received: "Binding response for requestId=req-t timed out"
```

Ran from the worktree root:

- `npx tsc --build` - clean
- `npm test -w @argent/tool-server` - 368 files, 4833 passed, 1 skipped
- `npx prettier --write` on both changed files - unchanged
- `npx eslint` on the changed src file - clean (the test file only hits the known local `packages/docs` tsconfig resolution error)

</details>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved debugger timeout messages with clearer guidance when execution is paused or frozen.
  - Added recommendations to resume the debugger or restart the app, along with a warning against repeated retries.
- **Tests**
  - Added coverage to verify timeout messaging and failure metadata for unanswered debugger evaluation requests.

### DIFF
--- a/packages/tool-server/src/utils/debugger/cdp-client.ts
+++ b/packages/tool-server/src/utils/debugger/cdp-client.ts
@@ -340,12 +340,20 @@ export class CDPClient {
       const timer = setTimeout(() => {
         this.pendingBindings.delete(id);
         reject(
-          new FailureError(`Binding response for requestId=${id} timed out`, {
-            error_code: FAILURE_CODES.DEBUGGER_CDP_BINDING_TIMEOUT,
-            failure_stage: "debugger_cdp_binding",
-            failure_area: "tool_server",
-            error_kind: "timeout",
-          })
+          new FailureError(
+            `Binding response for requestId=${id} timed out — the runtime took the script ` +
+              `but never called back over the binding. It may be paused at a breakpoint ` +
+              `(the script is dispatched without awaiting it, so a paused runtime still ` +
+              `accepts it and never runs the callback), or frozen. Check the debugger and ` +
+              `resume it; if nothing is paused, restart the app. Do not retry in a loop — ` +
+              `each attempt waits out the full timeout.`,
+            {
+              error_code: FAILURE_CODES.DEBUGGER_CDP_BINDING_TIMEOUT,
+              failure_stage: "debugger_cdp_binding",
+              failure_area: "tool_server",
+              error_kind: "timeout",
+            }
+          )
         );
       }, timeout);
 
--- a/packages/tool-server/test/metro/cdp-client.test.ts
+++ b/packages/tool-server/test/metro/cdp-client.test.ts
@@ -287,6 +287,35 @@ describe("CDPClient", () => {
       await client.disconnect();
     });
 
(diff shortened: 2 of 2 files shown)


latekvo:
Only merging due to consistency across the same file. I'm not a big fan of this prose as is.

PR #1119, opened 2026-09-09
docs(skills): debugger-reload-metro reloads one app, not every connected one

### LINKED ISSUE #922: argent-metro-debugger says debugger-reload-metro reloads "all connected apps"; the CDP path modern RN takes reloads exactly one

`packages/skills/skills/argent-metro-debugger/SKILL.md:43` documents the tool as:

> `debugger-reload-metro` | Reload **all connected apps** (like pressing "r" in Metro terminal). Needs a CDP target.

The tool's own description says the opposite — singular, one target:

`packages/tool-server/src/tools/debugger/debugger-reload-metro.ts:35-36`
> Restart the Metro JS bundle in **the connected React Native app** … Returns `{ reloaded, port, method, deviceName, appName, logicalDeviceId }` indicating which reload path was used and **which device/app was targeted**.

## What the code does

`debugger-reload-metro.ts:64-70` — the primary path is `Page.reload` on `api.cdp`, the single CDP session the URN `JsRuntimeDebugger:<port>:<device_id>` resolved (`:44-46`). `Page.reload` is a per-target CDP command; it reaches exactly the one runtime that session is attached to.

Only the fallback broadcasts. `debugger-reload-metro.ts:74-76` posts to Metro's `/reload`, which is the "press r in the terminal" behaviour the skill describes — and it runs *only* when `Page.reload` throws (`:68-70`).

Which path runs is decided by the RN version, and the file says so itself at `:56-57`: *"Primary: CDP Page.reload — works reliably with RN 0.76+ (Fusebox/Bridgeless)"*, versus `:72-73` for the fallback: *"Metro's HTTP /reload endpoint (RN CLI classic, older Expo setups). Not always present — Expo SDK 52+ / RN 0.76+ may not expose it."*

So on every modern RN the tool reloads **one** app; the skill's description only ever matched the legacy fallback.

## Expected

The skill row should say what the tool description says: one app — the one `device_id` resolved. The scope difference between the two paths is also invisible today except through the returned `method` field (`"cdp"` vs `"http"`), which the skill does not mention.

Section 2 of the same skill (`SKILL.md:28`) is entirely about one Metro port serving several devices. An agent that has just edited JS and needs both a simulator and an emulator to pick it up reads line 43, calls `debugger-reload-metro` once, and treats both devices as reloaded. On RN 0.76+ the second device is still running the old bundle. Every subsequent `debugger-evaluate`, `debugger-log-registry` and screenshot on that device then describes pre-edit code, and the agent attributes the stale behaviour to the change it just made.

latekvo:
Fixes #922

**Cause:** the skill's tool row said `debugger-reload-metro` reloads "all connected apps", but its primary path is CDP `Page.reload` on the single session `device_id` resolved. Only the HTTP `/reload` fallback broadcasts.

**Fix:** the row now describes the per-target behaviour, the fallback, the returned `method` that distinguishes them, and says to call it once per device when one Metro port serves several.

**Verified:** read the tool source; `npx tsc --build` and `npx prettier --check` green.

Docs: no change needed - `packages/docs/docs/reference/tools.mdx` already describes the tool singularly ("Restart the Metro bundle without restarting the native process").

Tests: none - prose-only change in a package (`@argent/skills`) that ships no test suite.

<details>
<summary>Class check: other references to the tool</summary>

`grep -rn "reload-metro"` across the repo - every other mention is singular or scope-neutral, so nothing else needed the same edit:

- `packages/docs/docs/reference/tools.mdx:69` - "Restart the Metro bundle without restarting the native process"
- `packages/skills/skills/argent-metro-debugger/SKILL.md:129` - "Reload JS (already connected)"
- `packages/skills/skills/argent-react-native-app-workflow/SKILL.md:96,132,226` - "Reload tool" / "Reload JS bundle"
- `packages/skills/skills/argent-lens/SKILL.md:48` - "Reload the RN bundle"
- `argent-tv-interact` / `argent-metro-debugger` availability lines - platform gating only

</details>

<details>
<summary>Source the wording came from</summary>

`packages/tool-server/src/tools/debugger/debugger-reload-metro.ts`

```ts
try {
  await api.cdp.send("Page.reload");
  void disableLogBox();
  return { reloaded: true, port, method: "cdp", ...context };
} catch {
  /* fall through */
}

const res = await fetch(`http://127.0.0.1:${port}/reload`, { method: "POST" });
```

The CDP session is `JsRuntimeDebugger:${port}:${canonicalDeviceId(device_id)}` - one runtime.

</details>

### DIFF
--- a/packages/skills/skills/argent-metro-debugger/SKILL.md
+++ b/packages/skills/skills/argent-metro-debugger/SKILL.md
@@ -38,10 +38,10 @@ With two or more devices on one Metro, `debugger-connect` refuses a udid/serial 
 
 ### Reload & recovery
 
-| Tool                    | Purpose                                                                                       |
-| ----------------------- | --------------------------------------------------------------------------------------------- |
-| `debugger-reload-metro` | Reload all connected apps (like pressing "r" in Metro terminal). Needs a CDP target.          |
-| `restart-app`           | Terminate and relaunch the app by device id and bundleId. Use when app lost Metro connection. |
+| Tool                    | Purpose                                                                                                                                                                                                                                                                                                                                                                  |
+| ----------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `debugger-reload-metro` | Reload the JS bundle in the one app `device_id` resolved, via CDP `Page.reload`. Needs a CDP target. Falls back to Metro's `/reload` broadcast (the "press r" behaviour, hitting every connected app) when the runtime has no `Page.reload`; the returned `method` (`"cdp"` / `"http"`) says which ran. With several devices on one Metro port, call it once per device. |
+| `restart-app`           | Terminate and relaunch the app by device id and bundleId. Use when app lost Metro connection.                                                                                                                                                                                                                                                                            |
 
 ### Inspection & console
 


latekvo:
Bloat. Will include this in a separate PR, skipping for now.

PR #1120, opened 2026-09-09
docs(skills): say the [L:<id>] log anchor must be matched literally

### LINKED ISSUE #916: argent-metro-debugger documents `[L:42]` as a grep anchor, but it is a regex character class that matches every line in the log file
`packages/skills/skills/argent-metro-debugger/SKILL.md:100-106` documents the flat-log format and names the id field the search key:

| Field | Example | Notes |
| --- | --- | --- |
| `[L:<id>]` | `[L:42]` | Unique grep anchor |

and §5 then tells the agent to search the file with `Grep` / `grep` (`:92`, `:114-115`), defaulting to `-m 50`. Nothing in the skill says the anchor must be escaped or matched literally.

`[L:42]` is not a literal — it is a POSIX/PCRE character class matching a single `L`, `:`, `4` or `2`. Every line the writer emits starts with `[L:` and contains a digit, so the anchor matches **every line in the file**.

### Repro

Write a log through the real `LogFileWriter` and grep it exactly as the skill says:

```bash
node_modules/.bin/esbuild packages/tool-server/src/utils/debugger/log-file-writer.ts \
  --bundle --platform=node --format=cjs --outfile=/tmp/lfw.cjs
cat > /tmp/mk.cjs <<'EOF'
const os = require("node:os"); os.homedir = () => "/tmp/anchorhome";
const { LogFileWriter } = require("/tmp/lfw.cjs");
const w = new LogFileWriter(8081);
for (let id = 0; id <= 45; id++)
  w.write({ id, timestamp: "2026-03-17T14:31:00.000Z", level: id === 42 ? "error" : "log",
            message: id === 42 ? "TARGET ENTRY FORTY TWO" : id === 3 ? "state update: [object Object]" : "filler " + id });
console.log(w.getFilePath()); w.close({ keepFile: true });
EOF
F=$(node /tmp/mk.cjs)
wc -l < "$F"                  # 46 lines
grep -c '[L:42]' "$F"         # the documented anchor
rg  -c '[L:42]' "$F"          # the Grep tool's engine
grep -cF '[L:42]' "$F"        # what was meant
grep -c  '[object Object]' "$F"
grep -cF '[object Object]' "$F"
```

### Observed vs expected

| command | observed | expected |
| --- | --- | --- |
| `wc -l` | 46 | - |
| `grep -c '[L:42]'` | **46** | 1 |
| `rg -c '[L:42]'` | **46** | 1 |
| `grep -cF '[L:42]'` | 1 | 1 |
| `grep -c '[object Object]'` | **46** | 1 |
| `grep -cF '[object Object]'` | 1 | 1 |

With the skill's own recommended `-m 50`, `grep -m 50 '[L:42]' "$F"` returns the first 46 lines of the file — i.e. the head of the log, silently, with entry 42 buried in it.

### Impact on an agent

The anchor is exactly what the tool answer hands the agent to search with: `debugger-log-registry` returns `clusters[].firstId` / `lastId` (`packages/tool-server/src/utils/debugger/log-file-writer.ts:178-189`), and the documented next step is to grep `[L:<that id>]`. Following the documentation as written returns the whole file (or its first `-m N` lines) instead of one entry, and the agent has no signal that anything went wrong — it gets plausible-looking log lines back, just not the ones it asked for, and reasons from the head of the session about an entry near the end.

`[object Object]` is the same shape and is extremely common in RN console output (any object logged through a template string): a search for it matches every line, so "how many times did we log an unserialized object" comes back as the file's line count.

Two ways out, both cheap: document the anchor as a literal (`grep -F '[L:42]'`, or `grep '\[L:42\]'`) in the format table and in the "When reading from the log file" bullets, and/or pick an anchor that is regex-inert.

### Provenance

All line references are against `origin/main` (`366f17c61`). Not caused by #705: the format table (`SKILL.md:96-109`) and the grep bullets (`:112-117`) are byte-identical there. Found while reviewing that PR. Adjacent to #912 (`<LEVEL>` column documented as `WARN`, written as `WARNING`) but a different column and a different failure mode.

latekvo:
Fixes #916

**Cause:** the flat-log table calls `[L:42]` a "unique grep anchor" and never says to match it literally - unescaped, `[L:42]` is a character class, and since every log line starts with `[L:`, it matches the whole file.

**Fix:** table cell points at the search rule, plus one bullet telling the agent to use `grep -F` (or `\[L:42\]`) for bracketed text like `[L:<id>]` and `[object Object]`.

**Verified:** wrote 46 entries through the real `LogFileWriter` and ran both engines against the file - `grep`/`rg` with the bare anchor return 46 lines, `-F` returns 1. Prose-only change to one SKILL.md, so no test applies; `npx tsc --build` and `prettier --check` are green. No docs-site page documents the log format or the anchor, so `packages/docs` needs no update.

<details>
<summary>Repro output</summary>

```
$ head -2 repro.log
[L:0] 2026-03-17T14:31:00.000Z LOG   - | filler 0
[L:1] 2026-03-17T14:31:00.000Z LOG   - | filler 1

lines: 46
grep -c '[L:42]'          -> 46
grep -cF '[L:42]'         -> 1
grep -c '\[L:42\]'        -> 1
rg   -c '[L:42]'          -> 46
rg   -cF '[L:42]'         -> 1
grep -c '[object Object]' -> 46
grep -cF '[object Object]'-> 1
```

Entry 42 is the only `TARGET ENTRY FORTY TWO` line; `[object Object]` appears in exactly one message.

</details>

## Summary by CodeRabbit

- **Documentation**
  - Clarified the meaning of the `[L:<id>]` field in log registry output.
  - Added guidance for searching bracketed log text, including examples such as `[L:42]` and `[object Object]`.
  - Explained how to use fixed-string or escaped-bracket searches to avoid pattern-matching issues.

### DIFF
--- a/packages/skills/skills/argent-metro-debugger/SKILL.md
+++ b/packages/skills/skills/argent-metro-debugger/SKILL.md
@@ -99,7 +99,7 @@ One entry per line — fields (whitespace-separated, `|` delimiter before messag
 
 | Field         | Example                     | Notes                                               |
 | ------------- | --------------------------- | --------------------------------------------------- |
-| `[L:<id>]`    | `[L:42]`                    | Unique grep anchor                                  |
+| `[L:<id>]`    | `[L:42]`                    | Unique anchor; search it literally (see below)      |
 | `<timestamp>` | `2026-03-17T14:30:00.000Z`  | ISO 8601                                            |
 | `<LEVEL>`     | `ERROR`, `WARN `, `LOG  `   | Uppercase, padded to 5 chars                        |
 | `<source>`    | `src/api/user.ts:42` or `-` | Relative path from source map; `-` if unavailable   |
@@ -115,6 +115,7 @@ When reading from the log file:
 - Default to `-m 50` unless you need more.
 - Use `tail -N` recent entries.
 - `clusters[].message` gives you the exact text which you may look for
+- Search bracketed text such as `[L:42]` or `[object Object]` with `grep -F`, or escape the brackets (`\[L:42\]`). Unescaped, `[...]` is a character class: `grep '[L:42]'` matches every line in the file.
 
 > **If the file is too large** Delegate to an `Explore` subagent with the file path, the format spec above, the specific patterns you need, and Golden Rule 4's untrusted-data caveat.
 


latekvo:
I'd call this an opinionation to our documentation style, but i don't see why this wouldn't be a good clarification. Merging.

PR #1123, opened 2026-09-09
docs(skills): list the log-file level tokens the writer actually emits

### LINKED ISSUE #912: argent-metro-debugger documents the log file level column as WARN, but console.warn writes WARNING
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `SKILL.md:107` documents the flat log format's level column as:

| field | values | note |
|---|---|---|
| `<LEVEL>` | `ERROR`, `WARN `, `LOG  ` | Uppercase, padded to 5 chars |

`LogFileWriter` writes `LEVEL_DISPLAY[entry.level] ?? entry.level.toUpperCase().padEnd(5)` (`packages/tool-server/src/utils/debugger/log-file-writer.ts:154`), and `LEVEL_DISPLAY` (`:53-59`) is keyed `log` / `warn` / `error` / `info` / `debug`. But the CDP `Runtime.consoleAPICalled` type for `console.warn` is **`warning`**, not `warn` — the file's own comment says so four lines up (`:149-153`: *"CDP emits levels longer than 5 chars (e.g. `warning` from `console.warn`, `assert` from `console.assert`)"*), which is why that line deliberately does not truncate.

So `warning` misses the lookup table and falls through to `"warning".toUpperCase().padEnd(5)` → `WARNING` (7 chars, unpadded). Driving the real writer under a scratch `$HOME` confirms it:

```
[L:2] 2026-08-22T… WARNING app.js:10 | a warning line
[L:3] 2026-08-22T… ASSERT  app.js:11 | an assert line
```

`WARN ` is only produced for the literal level `warn`, which CDP does not emit.

### Why it matters

The documented column width is what the skill tells the agent to grep with, and §5's whole workflow is "get the summary, then `Grep` the flat file". An agent that greps for the documented `WARN` token finds nothing and reports that there were no warnings before the crash — a false negative on the exact output the file exists to preserve. It is also the last step of every crash-recovery route in #705, which is where I hit it.

`ASSERT` is the same shape: `console.assert` emits `assert`, also absent from `LEVEL_DISPLAY`, and also unlisted in the table.

Either the table should list the values CDP actually produces, or `LEVEL_DISPLAY` should be keyed by the CDP type names so the column matches the documentation.

Not caused by #705 — `SKILL.md:107` and `LEVEL_DISPLAY` are both unchanged by it (`git diff origin/main...HEAD` touches neither). Found while reviewing it.

latekvo:
Fixes #912

**Cause:** the flat-log table listed `WARN `, which `LogFileWriter` only writes for the level `warn` - a level CDP never sends. `console.warn` arrives as `warning` (written `WARNING`), `console.assert` as `assert` (`ASSERT`), neither documented. An agent grepping §5's documented token finds nothing and reports no warnings.

**Fix:** the `<LEVEL>` row now lists the tokens the writer emits and states the rule (uppercased CDP level, padded to at least 5 chars, never truncated). `LEVEL_DISPLAY` is gone: every key mapped to exactly what the `toUpperCase().padEnd(5)` fallback already computes, so it changed no output - it only kept `warn: "WARN "` alive as the token this table was copied from.

| before | after |
| --- | --- |
| `` `ERROR`, `WARN `, `LOG  ` `` - Uppercase, padded to 5 chars | `` `ERROR`, `WARNING`, `LOG  `, `INFO `, `DEBUG`, `ASSERT` `` - Uppercased CDP level, padded to at least 5 chars, never truncated |

**Verified:** new test drives the real `LogFileWriter` and pins the table to what lands on disk; it fails on the old row (3/3) and passes on the new one. Full `@argent/tool-server` suite, `tsc --build`, `typecheck:tests`, prettier all green.

Docs site needs no update - the level column is documented only in this skill; `argent-react-native-app-workflow` §4.2 defers to it rather than restating it.

<details>
<summary>Checks</summary>

Discrimination (docs change stashed, test kept):

```
FAIL  documents the token console.warn produces
  expected [ 'ERROR', 'WARN ', 'LOG  ' ] to include 'WARNING'
FAIL  documents the token console.assert produces
  expected [ 'ERROR', 'WARN ', 'LOG  ' ] to include 'ASSERT'
FAIL  documents no token a CDP level cannot produce
  `WARN ` is documented but never written
Tests  3 failed (3)
```

Restored: `Tests  3 passed (3)`.

Package suite: `Test Files 1 failed | 368 passed`, `Tests 1 failed | 4834 passed` - the one failure is `flow-type-focus.test.ts` asserting a 500 ms settle and measuring 499 ms; it passes on its own re-run and touches nothing in this change.

```
npx tsc --build                                   # clean
npm run typecheck:tests -w packages/tool-server   # clean
npx prettier --write <changed files>              # clean
```

</details>


latekvo:
Re-verified and it holds up.

PR #1124, opened 2026-09-09
fix(screen-recording): put the pid and a random suffix in the mp4 name

### LINKED ISSUE #914: Screen-recording output name has no pid and a non-injective device sanitizer, so two tool-servers recording one device share a path

```ts
// tools/screen-recording/capture.ts:307-310
const outputFile = path.join(
  os.tmpdir(),
  `argent-screen-recording-${api.deviceId.replace(/[^A-Za-z0-9._-]/g, "-")}-${Date.now()}.mp4`
);
```

Two separate problems in one name.

**No process discriminator.** In-process this is guarded by the per-device session plus `assertNoActiveRecording` (`tools/screen-recording/session-guards.ts:49-51`) — but that guard is in-memory, so it says nothing about a second tool-server. Two tool-servers told to record the same device in the same millisecond produce one path and two ffmpeg encoders writing it. Two tool-servers on one host is routine here (a global install alongside per-worktree dev servers).

**The sanitizer is not injective.** `replace(/[^A-Za-z0-9._-]/g, "-")` maps `emulator:5554` and `emulator-5554` onto the same segment, so two genuinely different adb serials can collapse into one name even across different milliseconds if the timestamps coincide.

**Nearest twin, three files away, gets it right:** `tools/screen-recording/watermark.ts:143-152` uses `argent-watermark-${process.pid}-${Date.now()}-${Math.random().toString(36).slice(2,8)}.png`, with a comment explaining that the suffix "keeps two recordings started in the same millisecond from sharing (and deleting) one another's file". The mp4 — the artifact the user actually keeps — has neither component.

Lower severity than #913 because the in-process guard covers the common single-server case, and because a millisecond is a much narrower window than that issue's one second. Filed separately since the fix is in a different file and the sanitizer half is independent of the pid half.

Found while reviewing #705, which hit the same class in the console-log writer and fixed it there by adding pid and a per-process sequence.

latekvo:
Fixes #914.

**Cause** - `startCapture` named its mp4 `argent-screen-recording-<sanitized device>-<ms>.mp4`. No pid, so two tool-servers on one host (a global install next to a dev server) recording the same device in the same millisecond derive one path and point two ffmpeg encoders at it; and `replace(/[^A-Za-z0-9._-]/g, "-")` is not injective, so `emulator:5554` and `emulator-5554` collapse onto one segment.

**Fix** - one line: `argent-screen-recording-${process.pid}-${sanitized}-${Date.now()}-${rand}.mp4`, the same shape `writeLogoTemp` in `watermark.ts` already uses for the watermark temp file.

**Class** - the only other artifact temp names of this shape are left alone: `chromium-server/screenshot.ts` already carries `${Date.now()}-${process.pid}`, and `tools/screenshot/index.ts` (`tvScreenshot`) stamps `process.hrtime.bigint()`, a host-wide monotonic nanosecond clock, so a collision there needs the same nanosecond rather than the same millisecond.

**Docs** - none needed: the name is a tool-server temp path, and nothing in `packages/docs/` or the skills documents its shape.

<details>

Both new tests fail on the unfixed source (`git stash` of `capture.ts` only):

```
 × names the output file after the recording process, not the device alone
   expected 'argent-screen-recording-6DBF83B4-0000-0000-0000-000000000000-1788975122156.mp4'
   to contain '-11818-'
 × keeps two device ids that sanitize alike on separate output paths
   expected '/var/folders/.../argent-screen-recording-emulator-5554-1700000000000.mp4'
   not to be the same path
```

With the fix restored, both pass. Also run from the worktree root:

```
npx tsc --build                              exit 0
npm run typecheck:tests -w @argent/tool-server   exit 0
npm test -w @argent/tool-server              368 files, 4834 passed, 1 skipped
npx prettier --write <changed files>         unchanged
```

`npx eslint` on the changed files could not run locally: `packages/docs/tsconfig.json` fails to resolve `@docusaurus/tsconfig` because the docs workspace is not installed here. Unrelated to this diff.

</details>

### DIFF
--- a/packages/tool-server/src/tools/screen-recording/capture.ts
+++ b/packages/tool-server/src/tools/screen-recording/capture.ts
@@ -287,9 +287,13 @@ async function startCaptureLocked(
     );
   }
 
+  // The pid and the random suffix are what keep two captures apart: the
+  // in-process session guard says nothing about a second tool-server on the
+  // same host, and the device sanitizer is not injective (`emulator:5554` and
+  // `emulator-5554` collapse onto one segment).
   const outputFile = path.join(
     os.tmpdir(),
-    `argent-screen-recording-${api.deviceId.replace(/[^A-Za-z0-9._-]/g, "-")}-${Date.now()}.mp4`
+    `argent-screen-recording-${process.pid}-${api.deviceId.replace(/[^A-Za-z0-9._-]/g, "-")}-${Date.now()}-${Math.random().toString(36).slice(2, 8)}.mp4`
   );
 
   const stream = await openMjpegStream(params.streamUrl, STREAM_CONNECT_TIMEOUT_MS);
--- a/packages/tool-server/test/screen-recording.test.ts
+++ b/packages/tool-server/test/screen-recording.test.ts
@@ -192,12 +192,8 @@ const androidDevice: DeviceInfo = {
   kind: "emulator",
 } as DeviceInfo;
 
-// startCapture names its output join(os.tmpdir(),
-// `argent-screen-recording-${deviceId}-${Date.now()}.mp4`). The device ids are
-// fixtures, and vi.useFakeTimers() seeds its clock from the real time rather
-// than a constant, so the millisecond two runs start in is the only thing
-// separating their paths — and each test ends by deleting the path it derived.
-// Scoping the tmpdir per test removes that window instead of leaving it narrow.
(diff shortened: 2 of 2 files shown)


latekvo:
I get the date, but why would ever need to disciminate PIDs when we already have a random suffix... `${Math.random().toString(36).slice(2, 8)}`

Skipping for now because this does not fix anything real and has potential to convolute things needlessly.

PR #1125, opened 2026-09-09
docs(skills): name restart-app's real parameters in argent-metro-debugger

### LINKED ISSUE #921: argent-metro-debugger says "All tools accept port AND device_id", but restart-app - its prescribed recovery - takes udid and rejects both

`packages/skills/skills/argent-metro-debugger/SKILL.md:26` opens section 2 with a blanket claim over every tool in the tables that follow:

> All tools accept `port` (default 8081) AND `device_id` …

`restart-app` is in one of those tables (`SKILL.md:44`, "Reload & recovery"). It accepts neither. `packages/tool-server/src/tools/restart-app/index.ts:20-37` declares `{ udid, bundleId, activity? }` — and `udid` is required, `port` and `device_id` are not in the shape at all. The word `udid` appears nowhere in the skill: the `restart-app` row says "by device id and bundleId", and `SKILL.md:130` (Quick Reference) says only "`restart-app`".

This is the skill's most-prescribed recovery action. `SKILL.md:79` (Golden Rule 2) routes the not-connected path to it — *"**`reason: "no_app_connected"` → get the app to connect to Metro** — use `restart-app` on the device, then retry `debugger-status` once."* — and `references/failure-scenarios.md:9,11,14` send the agent there from three more states.

## Repro

Parse the tool's real schema with the arguments the skill describes:

```ts
const def = createRestartAppTool(registry);
def.zodSchema.safeParse({ port: 8081, device_id: "6DBF83B4-…", bundleId: "com.example.app" });
```

Observed:

```
restart-app declared params: [ 'udid', 'bundleId', 'activity' ]
success: false
error.message:
[
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [ "udid" ],
    "message": "Invalid input: expected string, received undefined"
  }
]
```

`z.object` strips the unknown `device_id`/`port` keys, so the required `udid` reads as absent. That message is what reaches the caller verbatim — `packages/tool-server/src/http.ts:755-768` returns `res.status(400).json({ error: parseResult.error.message })`, and `packages/registry/src/registry.ts:128-131` wraps the same string for every non-HTTP dispatch path.

## Expected

Either the section-2 claim is scoped to the `debugger-*` tools it is actually true of, or the `restart-app` row names `udid` explicitly. As written the two contradict each other and only one of them is the tool.

An agent working from this skill has just been told by `debugger-status` that the app is gone, and Golden Rule 2 tells it to `restart-app`. It carries the `device_id` it has been passing to every debugger call and sends `{ port, device_id, bundleId }`. It gets back a raw Zod dump naming a parameter the skill never mentions, on the one step the skill says to take. The recovery it was mid-way through does not happen.

The failure is also mis-shaped for retry: the parameter it names (`udid`) is not the one the agent's error was about, so a retry that "fixes" `device_id` fails identically.

latekvo:
Fixes #921

**Cause** - `SKILL.md` section 2 opened with "All tools accept `port` ... AND `device_id`" over tables that include `restart-app`, which takes `udid` + `bundleId` and rejects both. `restart-app` is the recovery Golden Rule 2 and three `failure-scenarios.md` rows prescribe, so an agent following the skill sends `{port, device_id, bundleId}` and gets a raw Zod dump naming `udid`, a parameter the skill never mentioned.

**Fix** - scope the claim to the tools it holds for, and name `udid` / `bundleId` in the `restart-app` row and in the failure-scenarios row that spelled out its arguments. A new test derives both from the tool's real zod schema, so the claim cannot drift again.

No docs-site change: `packages/docs/docs/reference/tools.mdx` lists `restart-app` without naming parameters.

<details>

Discrimination, run twice against the unfixed prose:

```
# claim reverted to "All tools accept ..."
AssertionError: restart-app takes no port/device_id yet the claim covers it:
  expected 'All tools accept `port` (default 8081...' to contain '`restart-app`'

# row reverted to "by device id and bundleId"
AssertionError: restart-app's row does not name its `udid` parameter:
  expected '| `restart-app`           | Terminate...' to contain '`udid`'
```

Checks, from the worktree root:

```
npx tsc --build                                    # clean
npm run typecheck:tests -w @argent/tool-server     # clean
npx vitest run --shard=N/6   (packages/tool-server, N=1..6)
  -> 369 files, 4834 passed, 1 skipped, 0 failed
npx prettier --write <changed files>
```

Class check: `restart-app` is named in `argent-tv-interact` (already `{udid, bundleId}`), `argent-device-interact`, `argent-react-native-app-workflow`, `argent-settings-permissions` and the `argent-create-flow` references. Only `argent-metro-debugger` pairs it with a blanket `port`/`device_id` claim, so only that skill contradicted the schema; the others say "device id" with nothing nearby to read it as `device_id`.

</details>


packages/skills/skills/argent-metro-debugger/references/failure-scenarios.md
```diff
@@ -6,7 +6,7 @@ When a debugger tool fails, use **`debugger-status`** first to diagnose. Note: `
...
+| **App not connected**              | Error (or `not_connected` `detail`) contains: `Metro at port 8081 has no CDP targets — is a React Native app connected?`                           | `debugger-status` / `debugger-log-registry` return `{ status: "not_connected", reason: "no_app_connected", guidance }` here instead of erroring; other debugger tools still throw. 1) Confirm the app is running on the device. 2) Use `restart-app` with the app's `udid` and `bundleId` to relaunch so it connects to Metro. 3) Wait a few seconds for the bundle to load. 4) Retry `debugger-status` once. Do **not** use `debugger-reload-metro` to fix this — it also requires at least one target.                                                                                                                                                             |
```
latekvo, line 9: Weird, this PR is right but it should actually be `device_id` in source not `udid`, not sure why it isn't.

latekvo (commented):
Leaving this PR open for now. Do not merge yet, i'll have to check the scale of this issue and make adjustments accordingly.

PR #1126, opened 2026-09-09
fix(debugger): stop LogFileWriter retaining console lines no reader can reach

### LINKED ISSUE #892: LogFileWriter buffers console lines that nothing can read or free when its file could not be created
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: When `LogFileWriter`'s file cannot be created, every console line it captures is buffered in memory and nothing can ever read it or drop it.

`open()` is called once, from the constructor, and it swallows its own failure:

```ts
// packages/tool-server/src/utils/debugger/log-file-writer.ts
private open(): void {
  try {
    this.fd = fs.openSync(this.filePath, "w");
    this.ready = true;
    this.flushBuffer();
    ...
  } catch {
    // Nothing reopens the file: `write` buffers instead, and `hasFile` is how
    // a caller finds out there is nothing on disk to read.
  }
}
```

With `ready` false, `write()` takes the buffering branch (`this.writeBuffer.push(line)`), and both routes out of that buffer are closed to it:

- `flushBuffer()` returns at `if (!this.ready || this.fd === null) return;`
- `readAll()` returns `[]` at `if (this.closed || !this.ready) return [];` before it would call `flushBuffer()`

Nothing else reopens the file or clears the array, so the buffer grows for the life of the session — up to `MAX_ENTRIES` (50 000) lines — holding entries no tool can retrieve. `getStats().totalEntries` counts them, so `debugger-log-registry` reports a number whose entries cannot be read; that mismatch is what the `note` about a missing `file` exists to explain.

**Trigger**

`~/.argent/tmp` unwritable (a read-only home, a permissions repair, an `.argent` that is a file), then a debugger session that logs. The tool-server's own suite already builds this state - `packages/tool-server/test/debugger/log-file-writer.test.ts` chmods the log directory to `0o555` and asserts `totalEntries: 1` with `hasFile() === false`.

**Impact**

Bounded by `MAX_ENTRIES`, so it is a ceiling rather than an unbounded leak: one wedged session holds up to 50 000 formatted lines. Several such sessions on one long-lived tool-server multiply it. Nothing crashes; the memory is simply never returned and never useful.

**Fix shape**

Either stop buffering when there is no file to flush to (the entry counts and clusters are kept separately and stay correct), or make the buffer readable — `readAll()` could serve it directly rather than returning `[]` — and cap it. Both predate this PR; `open()`'s catch, `write()`'s branch and `readAll()`'s guard are unchanged from `main` (`36aa14a5f`).

latekvo:
Fixes #892

**Cause** - when `LogFileWriter`'s file cannot be created, `ready` stays false for the writer's life (nothing reopens it), so `write()` pushed every formatted line into `writeBuffer`. Both drains are closed in that state: `flushBuffer()` returns on `!this.ready`, and `readAll()` returns `[]` before reaching it. Up to `MAX_ENTRIES` (50 000) lines per wedged session, unreadable and never freed.

**Fix** - drop the dead buffering path; a line with no file is not retained. Entry counts, level tallies and clusters are tracked separately, so `getStats()` / `getClusters()` / `debugger-log-registry` are unchanged, as is `readAll()`, which already returned `[]` in this state.

`writeBuffer` / `flushBuffer` existed only in this file - no sibling writer has the same shape. No docs change: no tool, CLI flag, config key, flow directive or user-visible behaviour changed.

<details>

New test in `packages/tool-server/test/debugger/log-file-writer.test.ts` chmods a temp-`HOME` `~/.argent/tmp` to `0555`, writes one entry, and asserts the writer holds no line array while the counts and clusters still record it (skipped on Windows and under uid 0).

Without the source fix:

```
FAIL  packages/tool-server/test/debugger/log-file-writer.test.ts > LogFileWriter whose log file cannot be created > keeps no copy of a line it could not write
AssertionError: expected [ '[L:0] 2024-03-09T16:00:00.000Z LOG…' ] to deeply equal []
Tests  1 failed | 19 passed (20)
```

With it:

```
npx tsc --build                                  # clean
npm run typecheck:tests -w @argent/tool-server   # clean
npm test -w @argent/tool-server                  # 368 files, 4833 passed | 1 skipped
npx prettier --write <changed files>             # unchanged
```
</details>

## Summary by CodeRabbit

- **Behavior Changes**
  - Log entries written before the log file is ready are no longer retained in the file.
  - Entry counts and in-memory clusters continue to reflect those writes.
  - Reading a log no longer triggers pending data to be flushed first.

- **Tests**
  - Added coverage for unwritable log directories, including platform- and privilege-aware handling.

### DIFF
--- a/packages/tool-server/src/utils/debugger/log-file-writer.ts
+++ b/packages/tool-server/src/utils/debugger/log-file-writer.ts
@@ -64,7 +64,6 @@ const LINE_RE = /^\[L:(\d+)\] (\S+) (\S+)\s+(\S+) \| (.*)$/;
 export class LogFileWriter {
   private filePath: string;
   private fd: number | null = null;
-  private writeBuffer: string[] = [];
   private bytesWritten = 0;
   private entryCount = 0;
   private levelCounts: Record<string, number> = {};
@@ -84,19 +83,9 @@ export class LogFileWriter {
     try {
       this.fd = fs.openSync(this.filePath, "w");
       this.ready = true;
-      this.flushBuffer();
     } catch {
       // ignore
     }
-  }
-
-  private flushBuffer(): void {
-    if (!this.ready || this.fd === null) return;
-    for (const line of this.writeBuffer) {
-      const buf = Buffer.from(line);
-      fs.writeSync(this.fd, buf);
-    }
-    this.writeBuffer = [];
   }
 
   write(entry: Omit<RichLogEntry, "marker">): RichLogEntry {
@@ -123,11 +112,11 @@ export class LogFileWriter {
     const levelDisplay = LEVEL_DISPLAY[entry.level] ?? entry.level.toUpperCase().padEnd(5);
     const line = `[L:${entry.id}] ${entry.timestamp} ${levelDisplay} ${source} | ${flatMessage}\n`;
 
+    // Nothing reopens the file, and readAll() serves it alone, so a line with
+    // no file has no reader: the entry lives on only in the counts and clusters.
     if (this.ready && this.fd !== null) {
       const buf = Buffer.from(line);
       fs.writeSync(this.fd, buf);
-    } else {
-      this.writeBuffer.push(line);
     }
 
     this.bytesWritten += Buffer.byteLength(line);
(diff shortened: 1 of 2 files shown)


latekvo (commented):
Took a while to make sure this does not break anything. If i understand correctly, the merit of this PR is: "The buffer is only used when the session is wedged, and it'll be only flushable when the session becomes un-wedged, which will never happen, thus filling up RAM needlessly"

This merit makes sense.

I assume we could also detect and handle the wedge-itself, but that's a research+fix for a separate PR.

Issue #885, opened 2026-08-21
android-devtools leaves an adb forward, and tv-control an ax daemon, when their factories throw


latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Two smaller instances of the leak-on-throw class, both in blueprints that already carry the right rollback one frame away.

## 1. `android-devtools`: the adb forward outlives the throw

```ts
// packages/tool-server/src/blueprints/android-devtools.ts
const { stdout } = await runAdb(["-s", serial, "forward", "tcp:0", `tcp:${devicePort}`], { timeoutMs: 5_000 });
const lpMatch = ADB_FORWARD_PORT_MARKER.exec(stdout.trim());
if (!lpMatch) {
  throw new FailureError(`adb forward returned unexpected output: ${stdout.trim()}`, { ... });
}
...
} catch (err) {
  settle(() => reject(...), () => proc.kill());   // kills the helper; never removes the forward
}
```

Trigger: `adb forward` **succeeds** - the host-side forward now exists - but prints something `ADB_FORWARD_PORT_MARKER` does not match (an extra warning line, a locale or adb-version difference). The helper process is killed and the forward is left on the device for the life of the adb server. The factory's own rollback a frame up does call `removeAdbForward`; this path does not.

## 2. `tv-control`: the ax daemon is orphaned when the hid daemon fails to spawn

```ts
// packages/tool-server/src/blueprints/tv-control.ts
let axProc = spawnAxDaemon(udid, axSock);
const hidProc = spawnHidDaemon(udid, hidSock);   // throws synchronously when the binary is missing
```

Trigger: the tvOS ax binary present and the hid binary absent (a partial build) - `spawnHidDaemon` throws out of `requireBinIn`/`requireDarwin` while the ax daemon is already running inside the simulator, holding its socket for its `--timeout 3600`. The factory throws, so `node.instance` is never set and the file's own careful rollback never runs.

Both found while reviewing #705, which fixes the same class in the two JS-runtime-debugger blueprints. Filed separately per blueprint: #882 (native-devtools), #883 (ax-service), #884 (chromium-server), and these two.

Issue #884, opened 2026-08-21
createChromiumServer leaks the browser CDP WebSocket when priming the page throws


latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `createChromiumServer` opens the browser's CDP WebSocket and then runs three unguarded awaits:

```ts
// packages/tool-server/src/chromium-server/index.ts
const { cdp, wsUrl, target } = await connectCdp(opts.port);   // socket is OPEN from here
await primePageSession(cdp);                                  // <- unguarded
let viewport: ViewportSize = await readViewport(cdp);         // <- unguarded
...
await network.reattach();                                     // <- unguarded
```

This is the acquisition behind the `ChromiumCdp` blueprint's factory, so a rejection there means `node.instance` is never set and `_teardown` skips `dispose` - the socket is never closed.

## Trigger

A wedged renderer: those sends reject with `DEBUGGER_CDP_REQUEST_TIMEOUT`, which `not-connected.ts` maps to `runtime_unresponsive`. That reason's own guidance tells the agent to check the app and **retry once**, so the documented recovery path leaks a second live WebSocket, and each further attempt another - together with the `disconnected` listener registered just below.

## Impact

One live browser WebSocket per attempt for the life of the tool-server, against a browser the user is still using.

Found while reviewing #705, which fixes the same class in the two JS-runtime-debugger blueprints (a rejected setup send used to leave the CDP socket holding a Metro debugger target). Filed separately per blueprint: #882 (native-devtools), #883 (ax-service), this one, plus `android-devtools` and `tv-control`.

Issue #883, opened 2026-08-21
ax-service's factory leaks its listener when startProxy or spawnAxDaemon throws


latekvo:
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `ax-service`'s factory already carries the rollback this needs - it is just wrapped around the wrong span.

```ts
// packages/tool-server/src/blueprints/ax-service.ts
const server = await startListener(endpoint, ...);   // LISTENING (unix socket file, or an ephemeral TCP port)
...
await host.startProxy(udid, endpoint.port!);         // <- outside the try below
const proc = host.spawnAxDaemon(udid, endpoint);     // <- outside it too, and throws SYNCHRONOUSLY
...
try {
  daemonSocket = await waitForDaemonConnection(server, proc, 10_000);
} catch (err) {
  if (!proc.killed) proc.kill("SIGTERM");
  server.close();
  if (endpoint.transport === "unix") fs.unlinkSync(endpoint.socketPath);
  if (endpoint.transport === "tcp") await host.stopProxy(udid, endpoint.port!);
  throw err;
}
```

## Triggers

- `startProxy`: an `ios-remote` device whose tunnel cannot be started - `proxyStart` re-throws anything not matching `/already/i`.
- `spawnAxDaemon`: **no remote sim needed**. It is declared synchronous (`ios-host.ts`) and its bodies throw out of `requireBinIn` / `requireDarwin` when the ax artifacts are missing or the platform is wrong - a partially built checkout on a local sim reaches it.

## Impact

A listening server plus its fd leak per attempt, and on the unix transport the socket file at `/tmp/ax-<udid8>.sock` is left behind - which matters more than a stray file here, since a later instance binding that path is the contention #775 describes. The registry never calls `dispose`, because `node.instance` is only set once the factory resolves.

Found while reviewing #705, which fixes the same class in the two JS-runtime-debugger blueprints. Filed separately per blueprint: #882 (native-devtools), this one, plus `chromium-server`, `android-devtools` and `tv-control`.

PR #1127, opened 2026-09-09
fix(native-devtools): close the TCP listener when startProxy throws

### LINKED ISSUE #882: native-devtools' factory leaks its listening server when startProxy throws
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `Registry` sets `node.instance` only after `factory()` resolves, and `_teardown` is gated on `if (node.instance)`. Anything a factory acquires before it throws is therefore never disposed - it leaks for the life of the tool-server.

`native-devtools`'s factory binds its listener and then awaits an unguarded call:

```ts
// packages/tool-server/src/blueprints/native-devtools.ts
await new Promise<void>((resolve, reject) => {
  server.listen(endpoint.port ?? 0, "127.0.0.1", () => { ... endpoint.port = addr.port; resolve(); });
});
await host.startProxy(udid, endpoint.port!);   // <- outside any try/catch, server is LISTENING
```

## Trigger

An `ios-remote` device. `pickIosHost` returns the remote host, whose `startProxy` is `proxyStart` (`src/utils/sim-remote.ts`), which re-throws anything that does not match `/already/i` - the remote daemon being down or refusing the tunnel is enough.

## Impact

The listening TCP server (and its fd, and its realized ephemeral port) stays bound forever. `dispose` would have closed it (`server.close()`) and stopped the proxy, but it never runs. Every retry of the failing call leaks another.

## What makes this a clear miss rather than a judgement call

The statement immediately after states the rule verbatim, for a different call:

```ts
// Tolerate ensureEnv failure: throwing here would leak `server` — the
// registry's `_teardown` skips dispose when `node.instance` is never set.
await ensureEnvReady().catch(() => {});
```

Found while reviewing #705, which fixes the same class in the two JS-runtime-debugger blueprints (a rejected CDP setup send used to leave the socket holding a Metro debugger target). Filed separately per blueprint: this one, plus `ax-service`, `chromium-server`, `android-devtools` and `tv-control`.

latekvo:
Fixes #882

**Cause** - the TCP branch of the `native-devtools` factory binds an ephemeral 127.0.0.1 listener and then `await`s `host.startProxy(...)` outside any try/catch; on an `ios-remote` device that call rethrows when the tunnel is refused, and the registry sets `node.instance` only after `factory()` resolves, so `dispose` (which would `server.close()`) never runs. The listener, its fd and its port leak, once per retry.

**Fix** - close the server and rethrow, matching what `dispose` does and what the unix branch (`bindNativeDevtoolsUnixSocket`) already does on a bind error.

**Verified** - new test drives the factory with a remote host whose `startProxy` throws, then rebinds the port it was handed: EADDRINUSE before the fix, free after. Full `@argent/tool-server` suite green (434 files / 6164 tests) on the branch with main merged in, `tsc --build` and `typecheck:tests` clean.

Siblings from the same review are filed and left to their own issues: #883 (ax-service), #884 (chromium-server), #885 (android-devtools, tv-control).

No docs change: no tool, CLI flag, config key or flow-file surface moves.

<details>
<summary>Repro on the unfixed source</summary>

```
$ git stash push -- packages/tool-server/src/blueprints/native-devtools.ts
$ npx vitest run packages/tool-server/test/blueprints/native-devtools-tcp-proxy-leak.test.ts

 FAIL  native-devtools-tcp-proxy-leak.test.ts > closes the listener when startProxy throws
AssertionError: promise rejected "Error: listen EADDRINUSE: address already… { …(5) }" instead of resolving
Caused by: Error: listen EADDRINUSE: address already in use 127.0.0.1:51306
```

With the fix restored: 1 passed.
</details>


latekvo:
Makes sense.

PR #1131, opened 2026-09-09
fix(tool-server): stop boot-electron piping the app's stdout into nothing

### LINKED ISSUE #905: boot-electron pipes the Electron app stdout and never reads it, so a chatty renderer wedges on a full pipe buffer
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: `bootElectronApp` spawns the app with stdout **piped** and never reads it:

```ts
child = spawn(launcher.command, args, {
  detached: true,
  stdio: ["ignore", "pipe", "pipe"],
  env: electronGuiChildEnv({ ELECTRON_ENABLE_LOGGING: "1" }),
});
```
(`packages/tool-server/src/tools/devices/boot-electron.ts:372-380`)

Only `stderr` gets a consumer:

```ts
// Forward Electron stderr to our stderr so launch failures are visible to
// the user / agent. Drop stdout (renderer chatter) to keep tool-server logs clean.
child.stderr?.on("data", (chunk: Buffer) => { … });
```
(`:450-455`)

"Drop stdout" is what the comment intends, but a piped fd with no reader is not dropped — it is buffered. Once the OS pipe buffer fills (64 KiB on Linux, 16–64 KiB on macOS), the child's next `write(1, …)` blocks, and Chromium's logging is on the renderer's own thread.

The same call sets `ELECTRON_ENABLE_LOGGING: "1"` three lines above, which is precisely what makes the app chatty on stdout.

### Why it matters

A booted Electron app is expected to stay up for the whole session. An app that logs more than one pipe buffer's worth to stdout wedges: CDP stops answering, `debugger-status` reports `runtime_unresponsive` or `cdp_unreachable`, and every recovery surface then tells the agent to have the user quit and relaunch an app that only ever needed its output read.

`child.unref()` and `detached: true` do not help — the pipe is owned by the parent for as long as the tool-server lives.

### Fix shape

Either genuinely drop it (`stdio: ["ignore", "ignore", "pipe"]`) or attach a consumer that discards (`child.stdout?.resume()`). The first matches the comment; the second keeps the fd available if stdout is ever wanted.

Not measured against a real chatty app — reported from the spawn options and the missing consumer. Found while reviewing #703, which does not touch this file's runtime behaviour.

latekvo:
Fixes #905

**Cause:** `bootElectronApp` spawned the app with `stdio: ["ignore", "pipe", "pipe"]` and never read stdout. An unread pipe is buffered, not dropped, so once the OS pipe buffer fills the app blocks on its next `write(1, ...)` and CDP stops answering. The same spawn sets `ELECTRON_ENABLE_LOGGING: "1"`, which is what makes it write there.

**Fix:** `stdio: ["ignore", "ignore", "pipe"]`. stderr keeps its pipe - the existing forwarding listener drains it.

**Verified:** new `boot-electron-stdio.test.ts` asserts the spawned stdio and then spawns a real child under those exact options that writes 2 MiB to stdout; it must exit. Both cases fail on the unfixed line and pass with it. Full `@argent/tool-server` suite green (369 files, 4834 tests).

No docs change: this is the tool-server's internal spawn, no tool, CLI, config key or flow-file surface moves.

<details>
<summary>Discrimination proof and sibling spawn sites</summary>

Reverting only the source line, both cases fail:

```
FAIL  test/boot-electron-stdio.test.ts > leaves stdout unpiped and keeps stderr piped
  expected [ 'ignore', 'pipe', 'pipe' ] to deeply equal [ 'ignore', 'ignore', 'pipe' ]
FAIL  test/boot-electron-stdio.test.ts > lets a child that floods stdout run to completion
  AssertionError: expected false to be true   (the 2 MiB child never exits)
Tests  2 failed (2)
```

With the fix: `Tests  2 passed (2)`; package suite `Test Files 369 passed, Tests 4834 passed | 1 skipped`.

Also ran: `npx tsc --build`, `npm run typecheck:tests -w @argent/tool-server`, `npx prettier --write` on both files. Root eslint could not run in this checkout (`packages/docs/tsconfig.json` needs `@docusaurus/tsconfig`, which the root install skips).

Other spawns that pipe a stream, and why they are left alone:

- `argent-tools-client/src/launcher.ts` already calls `child.stdout?.resume()` once readline stops consuming - the precedent this fix follows.
- `blueprints/simulator-server.ts` and `blueprints/android-devtools.ts` read stdout with readline and `rl.close()` it at settle, leaving a piped stdout unread for the rest of a long-lived child. Same end state, but whether either child writes to stdout after its ready marker is unverified without a device, so it is not changed here.
- `boot-device.ts` (Windows) points stdout at a real log fd, `preview-window.ts` ignores stdout, and the `execFile`-style spawns are short-lived with a reader.
</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Electron app startup reliability by preventing unread standard output from blocking the app when it produces large amounts of output.
  * Continued forwarding standard error output for diagnostics.

* **Tests**
  * Added regression coverage for Electron process output handling, including high-volume output scenarios.

### DIFF
--- a/packages/tool-server/src/tools/devices/boot-electron.ts
+++ b/packages/tool-server/src/tools/devices/boot-electron.ts
@@ -322,7 +322,10 @@ export async function bootElectronApp(options: BootElectronOptions): Promise<Ele
   try {
     child = spawn(launcher.command, args, {
       detached: true,
-      stdio: ["ignore", "pipe", "pipe"],
+      // stdout is discarded, not piped: nothing reads it, and an unread pipe
+      // blocks the child's writes once the OS buffer fills. The
+      // ELECTRON_ENABLE_LOGGING below is what keeps it writing.
+      stdio: ["ignore", "ignore", "pipe"],
       // Strip ELECTRON_RUN_AS_NODE (see electronGuiChildEnv): inherited from an
       // Electron-based MCP host it would boot the binary in Node mode with no
       // CDP endpoint, failing boot-device instead of bringing the app up.
--- a/packages/tool-server/test/boot-electron-stdio.test.ts
+++ b/packages/tool-server/test/boot-electron-stdio.test.ts
@@ -1,0 +1,137 @@
+/**
+ * Regression test for the stdio boot-electron gives its child. The Electron app
+ * is spawned with ELECTRON_ENABLE_LOGGING=1 and outlives the boot call, but
+ * nothing in the tool-server ever reads its stdout. Piping stdout therefore
+ * buffers it in the kernel and blocks the app's next write once that buffer
+ * fills, wedging CDP; only stderr, which has a forwarding listener, may be a
+ * pipe.
+ */
+
+import { describe, it, expect, vi, beforeEach, beforeAll, afterAll } from "vitest";
+import { EventEmitter } from "node:events";
+import * as fs from "node:fs";
+import * as http from "node:http";
+import * as os from "node:os";
+import * as path from "node:path";
+import type { StdioOptions } from "node:child_process";
+import { scopeTempHome } from "./helpers/temp-home";
+
+scopeTempHome("argent-boot-electron-stdio-home-");
+
+const spawnMock = vi.fn();
+
+vi.mock("node:child_process", async () => {
+  const actual = await vi.importActual<typeof import("node:child_process")>("node:child_process");
+  return {
+    ...actual,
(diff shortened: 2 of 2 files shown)


latekvo:
Verified, makes sense. We encountered many such "buffer that eventually runs out of space" issues so far, perhaps worth to sweep the entire repo for this category of issue.

PR #1133, opened 2026-09-09
fix(tool-server): prune a Chromium CDP port only when nothing is listening

### LINKED ISSUE #886: One list-devices probe permanently forgets a running Chromium app that has no window open
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Found while reviewing #703 and set aside as out of scope - `utils/chromium-discovery.ts` is untouched by that PR and identical on `origin/main`.

## The defect

`probePort` prunes a port from both the tracked set and the persisted file whenever the probe throws:

```ts
// chromium-discovery.ts:106-129
const version = await ensureCdpReachable(port, ctrl.signal);
const target  = await discoverPrimaryPage(port, ctrl.signal);   // <-- also required
...
} catch {
  // Drop dead tracked ports so list-devices doesn't keep probing a closed app.
  TRACKED_PORTS.delete(port);
  if (loadPersistedPorts().includes(port)) {
    persistPorts((ports) => ports.delete(port));
  }
  return null;
}
```

The comment says "dead", but the `try` block requires `discoverPrimaryPage` to succeed as well - so a **live** app throws here whenever it has no drivable page. That is not an exotic state: an Electron app or browser whose last window was closed keeps its process and its CDP endpoint alive (`/json/version` answers fine) while `/json/list` reports no `page` target. One `list-devices` call in that window erases the port permanently.

This inverts the file's own stated purpose, four lines up at `:43-49`:

> "Booted Chromium apps are detached and deliberately outlive the tool-server ... so without persistence every restart makes running apps invisible to `list-devices` - **the agent then boots a duplicate instance**. Dead ports are pruned on probe failure, so the file self-heals after the app quits."

The pruning is meant to fire *after the app quits*. It fires while the app is running, and the documented consequence - a duplicate instance - is exactly what follows.

## Repro

No browser needed; a stub CDP endpoint reproduces it, since the only thing that matters is `/json/list` having no `page` entry. `ARGENT_CHROMIUM_PORTS_FILE` redirects the persisted file.

```js
import * as http from "node:http";
import * as fs from "node:fs";

let hasWindow = false;                       // the app starts with its window closed
const server = http.createServer((req, res) => {
  const port = server.address().port;
  res.writeHead(200, { "Content-Type": "application/json" });
  if (req.url.startsWith("/json/version"))
    return res.end(JSON.stringify({ Browser: "Electron/1", webSocketDebuggerUrl: `ws://127.0.0.1:${port}/b` }));
  return res.end(JSON.stringify(hasWindow
    ? [{ id: "p", type: "page", title: "App", url: "http://app/", webSocketDebuggerUrl: `ws://127.0.0.1:${port}/devtools/page/p` }]
    : []));
});
await new Promise((ok) => server.listen(0, "127.0.0.1", ok));
const { port } = server.address();

const f = `${process.env.TMPDIR}/ports-${port}.json`;
fs.writeFileSync(f, JSON.stringify([port]));
process.env.ARGENT_CHROMIUM_PORTS_FILE = f;

const { discoverChromiumDevices } = await import("./packages/tool-server/dist/utils/chromium-discovery.js");
console.log(await discoverChromiumDevices(), fs.readFileSync(f, "utf8"));  // probe while windowless
hasWindow = true;                            // the user brings the window back
console.log(await discoverChromiumDevices(), fs.readFileSync(f, "utf8"));  // app is healthy again
server.close();
```

Observed:

```
start: app alive, tracked          file=[53717]
  probe #1 (window closed)         -> 0 device(s)
after one failed probe             file=[]
  probe #2 (window restored)       -> 0 device(s)  <-- app is healthy again
after recovery                     file=[]
```

The app recovers completely and never comes back to `list-devices`. Nothing re-adds the port: `getCandidateChromiumPorts` unions 9222, `ARGENT_CHROMIUM_PORTS`, `TRACKED_PORTS` and the persisted file, and `trackChromiumPort`'s only caller is `boot-electron.ts:543`. So the port returns only if the user boots the app again - the duplicate.

## Impact

An agent that calls `list-devices` at the wrong moment loses the device permanently, and the natural recovery is the duplicate boot the persistence layer was built to avoid. Driving the app by its known `chromium-cdp-<port>` id still works (`resolveDevice` is shape-based), so this bites discovery and any flow that re-lists devices between steps, not a session that holds its id.

Narrowing the prune to failures that mean the endpoint is gone - `CHROMIUM_CDP_UNREACHABLE`, i.e. the `fetch` rejection in `fetchJson` - would keep the self-healing described at `:43-49` while leaving a windowless app tracked. `CHROMIUM_CDP_NO_PAGE_TARGET` and `CHROMIUM_CDP_INVALID_RESPONSE` both prove something is still listening on that port.

latekvo:
Fixes #886

**Cause:** `probePort` pruned a port from `TRACKED_PORTS` and the persisted file on *any* probe failure - including `CHROMIUM_CDP_NO_PAGE_TARGET`, which a live Electron app raises whenever its last window is closed (`/json/version` answers, `/json/list` has no `page`). One `list-devices` call in that window erased the port for good; the only recovery is booting a duplicate instance, exactly what the persistence layer exists to avoid.

**Fix:** prune only when the failure is `CHROMIUM_CDP_UNREACHABLE` - the code `fetchJson` raises when the connection itself fails. Any other failure proves something is still listening.

**Verified:** the issue's repro (a stub CDP endpoint with an empty `/json/list`), before vs after:

| | probe #1 (window closed) | probe #2 (window restored) |
| --- | --- | --- |
| `main` | 0 devices, file `[]` | **0 devices**, file `[]` |
| this branch | 0 devices, file `[61110]` | **1 device**, file `[61110]` |

Class check: `chromium-discovery.ts` holds the only prune-on-probe-failure path in the repo; `trackChromiumPort` / `untrackChromiumPort` have one caller each (`boot-electron.ts`, `flow-run.ts`) and are unaffected.

Docs: no change needed - no tool, CLI flag, config key or flow directive changes, and the `ARGENT_CHROMIUM_PORTS_FILE` row in `reference/configuration.mdx` still describes the file correctly.

<details>
<summary>Tests and checks</summary>

Two tests added to `packages/tool-server/test/chromium-discovery.test.ts`; both fail on the unfixed source:

```
 FAIL  keeps a running app tracked while it has no page target
 AssertionError: expected [ 9222 ] to include 51375
 FAIL  keeps a tracked port whose endpoint answers with a non-2xx status
 AssertionError: expected [ 9222 ] to include 51378
```

The first also asserts the app is rediscovered once its window returns; the second covers `CHROMIUM_CDP_INVALID_RESPONSE`. The existing "untracks a port after it stops responding" and "a dead persisted port is pruned" tests still pass - a closed listener yields `ECONNREFUSED`, so self-healing is intact.

Ran from the worktree root:

- `npx tsc --build` - clean
- `npm test -w @argent/tool-server` - 368 files, 4834 passed, 1 skipped (baseline on `main`: 4832 passed)
- `npx prettier --write` on both changed files - unchanged

</details>

### DIFF
--- a/packages/tool-server/src/utils/chromium-discovery.ts
+++ b/packages/tool-server/src/utils/chromium-discovery.ts
@@ -1,6 +1,7 @@
 import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
+import { FAILURE_CODES, getFailureSignal } from "@argent/registry";
 import { CHROMIUM_ID_PREFIX, chromiumIdFromPort } from "./device-info";
 import { ensureCdpReachable, discoverPrimaryPage } from "../blueprints/chromium-cdp";
 
@@ -43,7 +44,8 @@ const TRACKED_PORTS = new Set<number>();
  * Tracked ports are mirrored to a file: booted apps are detached and outlive
  * the tool-server (which auto-exits on idle), so without persistence a restart
  * hides running apps from `list-devices` and the agent boots a duplicate.
- * Failed probes prune the file, so it self-heals after the app quits.
+ * Probes that find nothing listening prune the file, so it self-heals after the
+ * app quits.
  */
 function portsFilePath(): string {
   return (
@@ -81,7 +83,7 @@ export function trackChromiumPort(port: number): void {
   persistPorts((ports) => ports.add(port));
 }
 
-/** Remove a port. Optional: a failed probe prunes it anyway. */
+/** Remove a port. Optional: an unreachable probe prunes it anyway. */
 export function untrackChromiumPort(port: number): void {
   TRACKED_PORTS.delete(port);
   persistPorts((ports) => ports.delete(port));
@@ -112,13 +114,18 @@ async function probePort(port: number, timeoutMs: number): Promise<ChromiumDevic
       browser: version.Browser ?? null,
       state: "Running",
     };
-  } catch {
-    // Drop dead tracked ports so list-devices stops probing a closed app.
-    TRACKED_PORTS.delete(port);
-    // Only touch the file when this port was persisted — a failed probe of
-    // 9222 or an env port must not create or rewrite it.
-    if (loadPersistedPorts().includes(port)) {
-      persistPorts((ports) => ports.delete(port));
+  } catch (err) {
+    // Drop the port only when nothing answers on it. Any other failure means
+    // something is still listening: an Electron app whose last window closed
+    // answers /json/version but lists no page target, and pruning it there
+    // hides a running app from list-devices for good.
+    if (getFailureSignal(err)?.error_code === FAILURE_CODES.CHROMIUM_CDP_UNREACHABLE) {
+      TRACKED_PORTS.delete(port);
+      // Only touch the file when this port was persisted — a failed probe of
+      // 9222 or an env port must not create or rewrite it.
+      if (loadPersistedPorts().includes(port)) {
+        persistPorts((ports) => ports.delete(port));
+      }
     }
     return null;
   } finally {
--- a/packages/tool-server/test/chromium-discovery.test.ts
+++ b/packages/tool-server/test/chromium-discovery.test.ts
@@ -20,7 +20,8 @@ interface FakeCdpServer {
 async function startFakeCdpServer(options?: {
   responses?: {
     version?: number | object;
-    list?: number | object;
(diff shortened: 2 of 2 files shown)


latekvo:
Will have to re-verify this against edge-cases, leaks. Proper no-activity timeouts **must** be present before getting this merged.

PR #1137, opened 2026-09-09
fix(describe): name the gesture tools Chromium accepts

### LINKED ISSUE #873: describe tells every Chromium caller to use gesture-swipe / gesture-pinch, which the capability gate rejects
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Found while reviewing #703 and set aside as out of scope - it lives in `describe`, which #703 does not touch. Present on `main` today.

## The defect

Every `describe` call against a Chromium target ends its header by naming three tools to pass the coordinates to. Two of them are rejected on Chromium at the capability gate.

`packages/tool-server/src/tools/describe/format-tree.ts:205-207`:

```ts
    header.push(
      "Pass them straight to gesture-tap / gesture-swipe / gesture-pinch, which expect this same space."
    );
```

Only `gesture-tap` declares chromium support:

| tool | `capability` | on Chromium |
|---|---|---|
| `gesture-tap` (`gesture-tap/index.ts:53-57`) | apple, appleRemote, android, **`chromium: { app: true }`** | works |
| `gesture-swipe` (`gesture-swipe/index.ts:45-48`) | apple, appleRemote, android | **rejected** |
| `gesture-pinch` (`gesture-pinch/index.ts:67-70`) | apple, appleRemote, android | **rejected** |

The branch is reachable on every Chromium describe: the walker sets `source: "cdp-dom"` (`describe/platforms/chromium.ts:544`), `isVega` is `opts.source === "vega-automation"` (`format-tree.ts:190`), so the `else` arm at `:204` always fires.

```
Tool 'gesture-swipe' is not supported on chromium app (no chromium support declared).
```

## Why this one matters more than its siblings

`describe` is the tool the skills tell an agent to call **first**, and this text is in the response body, so it arrives on every read of the screen - it does not require anyone to have opened a skill file. `gesture-swipe` is also the natural next call after describe shows content below the fold.

## The nearest twin already does it right

The Vega arm immediately above (`format-tree.ts:198-203`) is platform-corrected and names the tool that does work there:

> "Vega is remote-driven, not touch - there is no tap. Use the frames as spatial hints to plan D-pad moves with the `tv-remote` tool..."

Chromium gets no equivalent, though it has one: `gesture-scroll` is the wheel-based replacement for `gesture-swipe` (`gesture-scroll/index.ts:63-65`, `chromium: { app: true }`), and `gesture-swipe`'s own description already says so - *"Not supported on Chromium - use gesture-scroll there instead."* The correction exists at the tool level and is missing at the point of use.

## Same string, second surface

`describe/index.ts:172-173` repeats it in the tool's own `description`: *"the same space as gesture-tap / gesture-swipe / gesture-pinch"*. That one is `alwaysLoad`, so it reaches an agent with no skill loaded.

Verified by reading source at `fda49d690`; not executed.

latekvo:
Fixes #873

**Cause:** `describe`'s response header and tool description tell every caller to pass frames to `gesture-tap / gesture-swipe / gesture-pinch`, but only `gesture-tap` declares `chromium` support - swipe and pinch are rejected at the capability gate on a Chromium target.

**Fix:** on `source: "cdp-dom"` the header names `gesture-tap / gesture-scroll / gesture-drag` (the tools with `chromium: { app: true }`) and says swipe/pinch are not supported on Chromium. The tool description gets the same qualifier, plus the physical-iOS `gesture-pinch` gap the header already covers. Two skill lines that called `gesture-swipe` "touch-only" now say it is not supported on Chromium.

**Not "touch-only":** Chromium accepts CDP touch input (a touch swipe scrolls, a two-finger spread zooms - see details). The rejection is argent's: it drives Chromium with mouse events, where a drag selects text instead of scrolling - hence `gesture-scroll`.

**Docs:** none needed - `reference/tools.mdx` already labels `gesture-scroll` / `gesture-drag` as Chromium; no tool, CLI, config key or capability changed.

<details>

Chromium touch probe - headless Chrome 153 over raw CDP, no touch emulation (`navigator.maxTouchPoints === 0`):

| input | result |
|---|---|
| `Input.dispatchTouchEvent` tap | `pointerdown:touch`, `touchstart`, `touchend`, `click:touch` |
| `Input.dispatchTouchEvent` swipe up 400px | `scrollY` 0 → 484 |
| `Input.dispatchTouchEvent` two-finger spread | `visualViewport.scale` 1 → 3 |
| `Input.dispatchMouseEvent` drag up 400px (argent's Chromium path) | `scrollY` 0, 424 chars selected |

`test/describe-format-tree.test.ts` - "names the Chromium-capable gesture tools in the cdp-dom header" pins the exact line. Each mutation fails it (`Tests 1 failed | 23 passed (24)`):
- restore the "touch-only and are rejected on Chromium" wording
- drop the `cdp-dom` arm

From the worktree root:
- `npx tsc --build` - clean
- `npm test -w @argent/tool-server` - `Test Files 441 passed (441)`, `Tests 6241 passed | 1 skipped`
- `npx eslint --max-warnings 0` + `npx prettier --check` on the changed files - clean

E2E - the server's own `createRegistry()` + `createHttpApp()` served in-process, `POST /tools/<name>` against a live headless Chrome (`chromium-cdp-9853`), same entry bundled against `main` and this branch:

```
[after]  Pass them straight to gesture-tap / gesture-scroll / gesture-drag, which expect this same space. gesture-swipe and gesture-pinch are not supported on Chromium.
[before] Pass them straight to gesture-tap / gesture-swipe / gesture-pinch, which expect this same space.
```

Gate on that device: `gesture-tap` / `gesture-scroll` / `gesture-drag` 200; `gesture-swipe` / `gesture-pinch` 400 `Tool 'gesture-swipe' is not supported on chromium app (no chromium support declared).`

</details>

### DIFF
--- a/packages/skills/rules/argent.md
+++ b/packages/skills/rules/argent.md
@@ -17,7 +17,7 @@ Use cases:
 - Any request to execute manual QA, UI QA, or visual behavior validation for a mobile app
 - Running, debugging, or testing a React Native app (iOS, Android or Vega)
 - Profiling performance or diagnosing re-renders in a React Native app (iOS or Android)
-- Running, debugging, or testing a Chromium (CDP) app — an Electron app (boot with `boot-device` + `electronAppPath`) or a Chromium browser exposing CDP (auto-discovered on port `9222` / `ARGENT_CHROMIUM_PORTS`); on Chromium scroll with `gesture-scroll` and drag with `gesture-drag` — `gesture-swipe` is touch-only
+- Running, debugging, or testing a Chromium (CDP) app — an Electron app (boot with `boot-device` + `electronAppPath`) or a Chromium browser exposing CDP (auto-discovered on port `9222` / `ARGENT_CHROMIUM_PORTS`); on Chromium scroll with `gesture-scroll` and drag with `gesture-drag` — `gesture-swipe` is not supported there
   </description>
 
 <availability_check>
--- a/packages/skills/skills/argent-device-interact/SKILL.md
+++ b/packages/skills/skills/argent-device-interact/SKILL.md
@@ -21,7 +21,7 @@ Use `list-devices` to get a target id. Results are tagged with `platform` (`ios`
 
 1. **Always refer to tapping_rule** from your argent.md rule before tapping.
 2. Before performing interactions, consider whether they can be **dispatched sequentially** - more on that in `run-sequence`.
-3. **Use `gesture-swipe` for lists/scrolling**, not `gesture-custom`, unless you need non-linear movement. On Chromium use `gesture-scroll` instead — `gesture-swipe` is touch-only. Consider whether you need multiple swipes, if yes - use `run-sequence`. Pass `momentum: false` when the swipe should decelerate before ending for a precise movement.
+3. **Use `gesture-swipe` for lists/scrolling**, not `gesture-custom`, unless you need non-linear movement. On Chromium use `gesture-scroll` instead — `gesture-swipe` is not supported there. Consider whether you need multiple swipes, if yes - use `run-sequence`. Pass `momentum: false` when the swipe should decelerate before ending for a precise movement.
 4. **Tap a text field before typing**, then use `keyboard` to enter text.
 5. **Coordinates are normalized** — always 0.0–1.0, not pixels.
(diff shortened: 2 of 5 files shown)


hubgan (approved):

latekvo:
I feel like this isn't factually correct. Does chromium not support touches? I think it does. Re-verify.

latekvo:
Yup, touch works perfectly well on chromium. I feel like these changes would require more research, and instead the time would be better used just adding chromium support instead of verifying and documenting lack thereof.

PR #1139, opened 2026-09-09
fix(tool-server): absorb the deferred spawn error after boot-electron's no-pid throw

### LINKED ISSUE #866: boot-device: an unresolvable electronAppPath crashes the whole tool-server (unhandled ENOENT 'error' event)
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Found while reviewing #703 (a docs-only skill change) and set aside as out of scope - it is present on `main` today and #703 touches no code.

## The defect

One `boot-device` call with an `electronAppPath` whose Electron binary cannot be resolved kills the **entire tool-server process**, taking every other session's device connections down with it. The tool first returns a clean, correct 500 - and about 3 ms later the process exits on an uncaught exception.

```
$ curl -s -w "\nHTTP %{http_code}\n" -X POST http://127.0.0.1:39704/tools/boot-device \
    -H 'content-type: application/json' -d '{"electronAppPath":"/tmp/noelectron"}'
{"error":"[Tool:boot-device] Electron boot: spawn returned without a pid (binary: electron).",
 "error_code":"CHROMIUM_ELECTRON_SPAWN_FAILED","error_kind":"subprocess"}
HTTP 500

$ ps -p <server pid>
SERVER DEAD (crashed)
```

tool-server log:

```
[registry] toolFailed boot-device: [Tool:boot-device] Electron boot: spawn returned without a pid (binary: electron).
    at bootElectronApp (packages/tool-server/dist/tools/devices/boot-electron.js:421:15)
[tool-server] Uncaught exception: Error: spawn electron ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:286:19)
    at onErrorNT (node:internal/child_process:484:16)
[registry] serviceStateChange NativeDevtools:770F654F-...: RUNNING -> TERMINATING
```

`/tmp/noelectron` is just a directory with a `package.json` (`"main": "main.js"`) and a `main.js`, with no `electron` in a local `node_modules` and none on `PATH`. That is the ordinary shape of the mistake: `electronAppPath` is documented as "a project directory whose package.json points the Electron binary at the entry script" (`boot-device.ts:98-103`), so pointing it at a project that has not run `npm i` yet - or at the wrong directory - is a one-character slip, not an exotic input.

## Root cause

`packages/tool-server/src/tools/devices/boot-electron.ts:430-449`. The `'error'` listener is attached at line 428, then the no-pid guard **removes it** before throwing:

```ts
  child.once("error", spawnErrorListener);

  if (!child.pid) {
    // No pid + no async error yet is still possible on some platforms when
    // spawn fails very early. Detach the error listener before throwing so a
    // deferred `'error'` event delivered after this synchronous throw doesn't
    // resolve onto an orphan promise (which Node would surface as an
    // UnhandledPromiseRejection ...).
    child.removeListener("error", spawnErrorListener);
    spawnErrorReject = null;
    throw new FailureError(`Electron boot: spawn returned without a pid ...`);
  }
```

The removal was added to avoid an unhandled *rejection*. But it leaves the child with **no `'error'` listener at all**, so the deferred ENOENT escapes as an unhandled `'error'` event - which is precisely the crash the listener was introduced to prevent. The comment fourteen lines above it (`boot-electron.ts:395-402`) already spells this out:

> EventEmitter convention: an unhandled `error` event escapes as an uncaught exception - here that would crash the entire tool-server every time someone called boot-device with `electronAppPath` on a host that doesn't have electron on PATH.

The guard's premise - "No pid + **no async error yet**" - does not hold on this platform. Node emits both:

```
$ node -e 'const {spawn}=require("node:child_process");
  const c=spawn("definitely-not-a-real-binary-xyz",[],{detached:true});
  console.log("sync: child.pid =", c.pid);
  c.once("error",e=>console.log("async: error event ->", e.code));
  setTimeout(()=>{},300)'
sync: child.pid = undefined
async: error event -> ENOENT
```

Node v24.14.1, macOS 24.5.0 (darwin arm64). `pid` is `undefined` **and** the `'error'` event fires. So the branch that assumes the two are mutually exclusive is the branch that runs, every time.

## Why nothing caught it

`packages/tool-server/test/boot-electron-spawn-error.test.ts:201-213` covers this exact path but asserts only the rejection:

```ts
it("still rejects when spawn returns a child with no pid (early-fail path)", async () => {
  // Some platforms produce a child without a pid AND no async error event.
  // The synchronous "no pid" guard catches that case.
  spawnMock.mockReturnValue(makeFakeChild({ pid: undefined }));
  await expect(bootElectronApp({...})).rejects.toThrow(/spawn returned without a pid/);
});
```

The fake child never emits `'error'` after the no-pid throw, so the uncaught-exception window is never entered. The test's own comment encodes the assumption the run above disproves. The sibling test at line 79 (`registers an 'error' listener on the spawned electron child`) asserts `listenerCount("error") > 0` - but only on the **pid-bearing** child, so it does not cover this branch either.

## Suggested shape of a fix

Keep a listener attached across the throw instead of removing it - e.g. swap `spawnErrorListener` for a no-op absorber (`child.once("error", () => {})`) before throwing, so the deferred ENOENT is swallowed rather than escaping, and the caller still gets the synchronous `CHROMIUM_ELECTRON_SPAWN_FAILED`. A discriminating test would emit `'error'` on the fake child *after* awaiting the no-pid rejection and assert the process survives (the current test passes with or without the removal).

## Impact

`boot-device` with `electronAppPath` is the documented way to start a Chromium/CDP target, and the argent-metro-debugger skill (see #703) now names it as the recovery for a crashed Chromium app. A wrong path there does not just fail the call - it terminates the shared tool-server, so every concurrently connected simulator/emulator session on that host loses its services.

## Environment

- `main` at `c821a463b`; `boot-electron.ts` is byte-identical between `origin/main` and #703's head, and #703 changes no code.
- macOS 24.5.0, Node v24.14.1.

latekvo:
Fixes #866

**Cause:** an unresolvable electron binary makes `spawn()` report failure twice - `child.pid` is `undefined` *and* an ENOENT `'error'` event lands on the next tick. `bootElectronApp`'s no-pid guard removed the only `'error'` listener before throwing, so that event escaped as an uncaught exception and killed the whole tool-server ~3 ms after `boot-device` had already returned its clean 500.

**Fix:** attach a no-op absorber (`child.on("error", () => {})`) in place of the rejecting listener. The deferred error is swallowed instead of escaping, the caller still gets `CHROMIUM_ELECTRON_SPAWN_FAILED`, and no orphan promise is rejected.

**Verified:** `npx tsc --build`, `npm run typecheck:tests -w @argent/tool-server`, and the full `@argent/tool-server` suite (4831 passed; one unrelated wall-clock flake, `flow-idle-run` "settles inside the smallest timeout", green in isolation). The rewritten no-pid test fails on the unfixed line.

No docs change: no tool, CLI, config key, flow file or user-facing capability moves.

<details>
<summary>Repro, discrimination proof, and the sibling guards left alone</summary>

Node reports both signals, so the guard's old premise ("no pid **and** no async error yet") never holds - Node v24.14.1, darwin arm64:

```
$ node -e 'const {spawn}=require("node:child_process");
  const c=spawn("definitely-not-a-real-binary-xyz",[],{detached:true});
  console.log("sync: child.pid =", c.pid);
  c.once("error",e=>console.log("async: error event ->", e.code));
  setTimeout(()=>{},300)'
sync: child.pid = undefined
async: error event -> ENOENT
```

`test/boot-electron-spawn-error.test.ts` previously asserted the crash as the expected behaviour (`expect(() => child.emit("error", err)).toThrow(/late ENOENT/)` on a child with zero listeners). That assertion is inverted, and the test now also keeps its original `unhandledRejection` check so both failure modes stay pinned. Removing only `child.on("error", () => {})` from the source:

```
FAIL  test/boot-electron-spawn-error.test.ts > swallows the deferred 'error' that follows the no-pid throw
  expected function not to throw an error but "Error: late ENOENT" was thrown
Tests  1 failed | 8 passed (9)
```

With the fix: `Tests  9 passed (9)`.

Sibling spawn-failure paths checked and left alone:

- `utils/ios-profiler/startup.ts` (`waitForXctraceReady`) already keeps its `'error'` listener attached for the child's lifetime, and says why - this fix follows that precedent.
- `platforms/ios.ts:647` throws on a pidless `xctrace` child while that listener is still attached, so the deferred error is handled.
- `blueprints/tv-control.ts` and `screen-recording/capture.ts` detach `'error'` only after a child that *did* get a pid settles, where no deferred spawn error is in flight.
- `boot-electron.ts`'s own `detachBootListeners()` is the same pid-bearing case: the child is intentionally long-lived and its later exit belongs to the next owner.

Root `npm run lint` could not run in this checkout - `packages/docs/tsconfig.json` needs `@docusaurus/tsconfig`, which the root install skips.
</details>

### DIFF
--- a/packages/tool-server/src/tools/devices/boot-electron.ts
+++ b/packages/tool-server/src/tools/devices/boot-electron.ts
@@ -374,10 +374,12 @@ export async function bootElectronApp(options: BootElectronOptions): Promise<Ele
   child.once("error", spawnErrorListener);
 
   if (!child.pid) {
-    // No pid and no async error yet is still possible on some platforms when
-    // spawn fails very early. Detach first so an `error` event delivered after
-    // this throw can't reject an orphan promise and crash the tool-server.
+    // An unresolvable binary reports both ways: spawn() returns without a pid
+    // AND emits ENOENT on the next tick. Swap the rejecting listener for an
+    // absorber so that event neither rejects the promise this throw orphans nor
+    // escapes as an uncaught `error` event and kills the tool-server.
     child.removeListener("error", spawnErrorListener);
+    child.on("error", () => {});
     spawnErrorReject = null;
     throw new FailureError(
       `Electron boot: spawn returned without a pid (binary: ${launcher.command}).`,
--- a/packages/tool-server/test/boot-electron-spawn-error.test.ts
+++ b/packages/tool-server/test/boot-electron-spawn-error.test.ts
@@ -227,8 +227,9 @@ describe("bootElectronApp — spawn error handling", () => {
   });
 
   it("still rejects when spawn returns a child with no pid (early-fail path)", async () => {
-    // Some platforms produce a child without a pid AND no async error event.
-    // The synchronous "no pid" guard catches that case.
+    // `spawn()` reports an unresolvable binary synchronously as a missing pid,
+    // before any deferred `error` event can arrive. The "no pid" guard turns
+    // that into the rejection.
     spawnMock.mockReturnValue(makeFakeChild({ pid: undefined }));
 
     await expect(
@@ -382,12 +383,13 @@ describe("bootElectronApp — spawn error handling", () => {
     }
   });
 
-  it("detaches the error listener after the no-pid throw — a deferred 'error' must not become an unhandled rejection", async () => {
-    // Real-world regression scenario: a hostile platform returns a Child with
-    // no pid AND fires a deferred 'error' event after spawn returns. Before
-    // the fix, the error listener would still be attached and would call
-    // reject() on a promise that nobody is awaiting — Node's default
-    // --unhandled-rejections=throw would then crash the tool-server.
+  it("swallows the deferred 'error' that follows the no-pid throw — no uncaught exception, no unhandled rejection", async () => {
+    // Node reports an unresolvable binary both ways: `spawn()` returns a child
+    // with no pid AND emits ENOENT on the next tick, after the "no pid" guard
+    // has already thrown. That event must reach a listener that neither
+    // rejects the promise the throw orphaned nor lets EventEmitter escalate it
+    // to an uncaught exception — either one kills the whole tool-server,
+    // taking every other session's device connections with it.
     const child = makeFakeChild({ pid: undefined });
     spawnMock.mockReturnValue(child);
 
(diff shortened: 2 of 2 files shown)


latekvo:
Sus, this has to be re-verified before merging.

PR #1141, opened 2026-09-09
fix(flows): put the chromium carve-out on flow-start-recording's description

### LINKED ISSUE #874: flow-start-recording's description tells every flow to record restart-app first, without the Chromium carve-out its own param states
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Found while reviewing #703 and set aside as out of scope - it lives in the flow tools, which #703 does not touch. Present on `main` today.

## The defect

`flow-start-recording`'s `description` instructs the agent to record a `restart-app` as the first step, with no platform qualifier:

`packages/tool-server/src/tools/flows/flow-start-recording.ts:93-98`:

> After starting, use flow-add-step to append tool calls [...] For a self-contained e2e flow, **record a restart-app of the app under test as the FIRST step** (captured as the flow's `launch` step); for a reusable fragment, skip that and pass executionPrerequisite instead.

`restart-app` declares no chromium capability (`restart-app/index.ts:41-46`), so on a Chromium target that first step is rejected:

```
Tool 'restart-app' is not supported on chromium app (no chromium support declared).
```

## The carve-out already exists sixty lines up, on a param

The same file's `executionPrerequisite` `.describe()` (`flow-start-recording.ts:31-37`) states it in full:

> "restart-app has no chromium support, so a chromium flow records as a fragment; add the `launch: { chromium: <app path> }` line to the YAML afterward, deleting the executionPrerequisite line if you passed one - a flow that starts with a launch must not declare it."

`flow-add-step.ts:485` carries it too. So the project knows the rule; the one string that tells the agent what to do first does not.

## Why the param is not a substitute

A tool `description` is loaded whenever the tool is; a param `.describe()` is read when the agent is filling that argument - and an agent recording a self-contained e2e flow is explicitly told **not** to pass `executionPrerequisite`, so it has no reason to read the field that carries the warning. The instruction and its exception are on opposite sides of a branch the agent takes before it sees the exception.

Flow tools declare no `capability` at all, so nothing gates this by device context.

Verified by reading source at `fda49d690`; not executed.

latekvo:
Fixes #874

**Cause:** `flow-start-recording`'s tool description tells every recording to capture a `restart-app` as the FIRST step, with no platform qualifier. `restart-app` declares no chromium capability, so on a Chromium target that first step is rejected. The carve-out existed only on the `executionPrerequisite` param `.describe()` - which an agent recording a self-contained e2e flow is explicitly told *not* to pass, so it never reads it.

**Fix:** append the exception, in the wording the param already uses, to the instruction itself.

**Class:** `flow-start-recording`'s description is now the only place that gives the instruction - `flow-add-step`'s description dropped it on main (#940). The `argent-create-flow` skill scopes it under an "iOS, Android, and Vega e2e flows" heading and gives Chromium its own section, so it needs nothing.

**Docs:** none needed - `packages/docs/docs/reference/tools.mdx` lists only the description's first sentence, unchanged, and no tool, CLI flag, config key or flow-file capability moved.

<details>

Test: `flow-start-recording's record-a-restart-app-first instruction` in `packages/tool-server/test/flows/flow-tools.test.ts`.

With the description reverted to main's wording:

```
 FAIL  test/flows/flow-tools.test.ts > flow-start-recording's record-a-restart-app-first instruction > states the chromium exception alongside it
AssertionError: expected 'Start recording a new flow, resetting…' to contain 'restart-app has no chromium support, …'
```

Restored, it passes.

Ran on the merged branch, from the worktree root:

- `npx tsc --build` - clean
- `npx vitest run` over `packages/tool-server`, sharded 1/3, 2/3, 3/3 - 434 files, 6169 passed, 1 skipped
- `npx prettier --check` on both changed files - clean

</details>


latekvo:
Skip for now, excess bloat for a rare edge-case

PR #1145, opened 2026-09-09
fix(cli): reject a --platform value flow run cannot use

### LINKED ISSUE #823: argent flow run prints a raw Zod dump for an invalid --platform, and aborts the batch
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Found while reviewing #739; out of that PR's scope (both halves predate it and live in files it does not touch), so filing separately.

`argent flow run` never validates `--platform` against the four values its own help text lists, so a casing slip reaches the tool server and comes back as a raw Zod dump. This is the class #645 reported and #648 fixed for `argent run` — `flow run` was not covered.

## Repro

Built CLI from `main`, three ordinary flows in `./batch`:

```
$ argent flow run ./batch --platform iOS > out.txt 2> err.txt ; echo $?
1

-- out.txt --
[1/3] a-login.yaml
  ✗ did not finish (run error)
[2/3] b-checkout.yaml
  · not run (batch stopped)
[3/3] c-logout.yaml
  · not run (batch stopped)

FAIL — 3 flows: 0 passed, 1 failed, 2 skipped

-- err.txt --
[
  {
    "code": "invalid_value",
    ...
```

Single-flow form is the same:

```
$ argent flow run ./batch/a-login.yaml --platform iOS
Flow "a-login"
  ✗ did not finish (run error)
```

Any value outside `ios|android|chromium|vega` does it — `iOS`, `macos`, `web`, `android-tv`.

## Impact

1. The operator gets a JSON array of Zod internals instead of a sentence naming the flag and its accepted values. `--platform`'s own help line already lists them (`packages/argent-cli/src/flow.ts:121-131`).
2. In a directory run it costs the whole batch: the flag is per-run, not per-flow, so every remaining flow is marked `not run (batch stopped)` for a typo the CLI could have caught before the first request.

## Mechanism

Two independent gaps.

**The CLI does not check the value.** `packages/argent-cli/src/flow.ts:257` — `else if (flag === "--platform") out.platform = takeValue("--platform");` — stores any string, and `buildRunPayload` (`flow.ts:1058`) forwards it verbatim: `if (args.platform) payload.platform = args.platform;`. The enum exists (`LAUNCH_PLATFORMS`, `packages/tool-server/src/tools/flows/flow-utils.ts`) and the server's schema is `platform: z.enum(LAUNCH_PLATFORMS)`.

**The server classifies the rejection but drops the classification from the response.** `packages/tool-server/src/http.ts:750-768`:

```js
const parseResult = def.zodSchema.safeParse(bodyArgs);
if (!parseResult.success) {
  emitHttpFailure(
    { error_code: FAILURE_CODES.HTTP_ZOD_VALIDATION_FAILED, ..., error_kind: "validation" },
    ...
  );
  res.status(400).json({ error: parseResult.error.message });
  return;
}
```

`emitHttpFailure` is the telemetry sink; the response body is `{ error }` alone. Every failure raised from *inside* tool execution reaches the client through `errorSignalFields`, so a client can classify those — but this pre-invoke 400 carries neither `error_code` nor `error_kind`. A client therefore cannot tell a schema rejection from a transport death, which is why the batch treats it as an infra failure and stops.

Fixing the CLI side alone resolves the reported symptom. The second gap is worth deciding on separately: any future pre-invoke rejection is equally unclassifiable to every client.

## Aside, same repro

When `--platform` *is* valid but several devices of that platform are booted, the message still says to pass the flag the caller already passed — `packages/tool-server/src/tools/flows/flow-device.ts`:

```js
const scoped = opts.platform ? booted.filter((d) => d.platform === opts.platform) : booted;
...
throw deviceResolutionError(
  `${scoped.length} booted devices matched — pass --device or --platform to disambiguate.`, scoped);
```

With `--platform ios` and two booted iOS simulators, `--platform` cannot disambiguate further; only `--device` can.

latekvo:
Fixes #823

**Cause:** `flow run`'s `--platform` spec was a bare `{ kind: "value" }`, so any string was stored and forwarded to `flow-execute`, whose `z.enum(LAUNCH_PLATFORMS)` answered with a raw Zod issue array. The flag is per-run, so a directory run spent its first flow on the rejection and marked every remaining one `not run (batch stopped)`.

**Fix:** give the spec `choices: ["ios", "android", "chromium", "vega"]` - the shared parser's existing facility, matching the four the `--platform` help line already lists and the server's enum exactly. The run now fails before the first request with `--platform must be "ios", "android", "chromium" or "vega", got "iOS"`, exit 2.

**Class:** `--platform` was the last fixed-set value option in the CLI without `choices` (`config`/`flags`/`telemetry --scope` and `lens --terminal` already have it). Left alone: `--device`, `--output`, `link --host/--port/--token`, `server start --port/--host/--idle-timeout` are free-form, and `lens --agent` is resolved against an interactive picker.

**Out of scope, from the same issue:** the tool-server's pre-invoke 400 carries neither `error_code` nor `error_kind` (`http.ts:750-768`), so no client can classify a schema rejection - that is a server-side design call, not this fix. Likewise the `flow-device.ts` disambiguation message that suggests `--platform` to a caller who already passed it.

**Docs:** none needed. `reference/flow-yaml.mdx` already documents the four accepted values; this only rejects an invalid invocation earlier, with no change to the tool, CLI surface, config keys or flow-file behaviour.

<details>

Repro on `main`'s build - `parseRunArgs` stores anything:

```
iOS        => {"flowRef":"./batch","platform":"iOS", ...}
macos      => {"flowRef":"./batch","platform":"macos", ...}
android-tv => {"flowRef":"./batch","platform":"android-tv", ...}
```

Same call on this branch:

```
"iOS"        => THROW: --platform must be "ios", "android", "chromium" or "vega", got "iOS"
"macos"      => THROW: --platform must be "ios", "android", "chromium" or "vega", got "macos"
"web"        => THROW: --platform must be "ios", "android", "chromium" or "vega", got "web"
"android-tv" => THROW: --platform must be "ios", "android", "chromium" or "vega", got "android-tv"
ios          => {"flowRef":"./batch","platform":"ios", ...}
```

Through the built `flow()` on a directory, no tool server reachable:

```
$ node -e "flow(['run','/tmp/i823batch','--platform','iOS'], ...)"
Error: --platform must be "ios", "android", "chromium" or "vega", got "iOS"
Usage: argent flow <subcommand> [options]
...
exit code: 2
```

Discriminating tests - `packages/argent-cli/test/flow.test.ts`: "throws on a --platform value outside the four the help lists" and "rejects a misspelled --platform before running any flow in the batch" (asserts `callTool` was never called). With `src/flow.ts` reverted both fail:

```
FAIL > parseRunArgs > throws on a --platform value outside the four the help lists
  expected [Function] to throw an error

FAIL > argent flow run <dir> > rejects a misspelled --platform before running any flow in the batch
  expected [Function] to throw error including 'process.exit:2' but got 'process.exit:0'

Tests  2 failed | 108 passed (110)
```

Ran from the worktree root: `npx tsc --build`, `npm run typecheck:tests -w @argent/cli`, `npm test -w @argent/cli` (27 files, 529 tests passed), `npx prettier --write` on both changed files (unchanged), `npx eslint packages/argent-cli/src/flow.ts` (clean).

</details>

### DIFF
--- a/packages/argent-cli/src/flow.ts
+++ b/packages/argent-cli/src/flow.ts
@@ -154,7 +154,10 @@ const RUN_OPTIONS = {
   "json-stream": { kind: "boolean" },
   "recursive": { kind: "boolean", alias: "r" },
   "device": { kind: "value" },
-  "platform": { kind: "value" },
+  // Constrained here because the tool-server's own rejection is a raw Zod issue
+  // dump — and in a directory run it lands only after the first flow has run,
+  // taking the rest of the batch down with it.
+  "platform": { kind: "value", choices: ["ios", "android", "chromium", "vega"] },
   "output": { kind: "value" },
 } as const satisfies OptionSpecs;
 
--- a/packages/argent-cli/test/flow.test.ts
+++ b/packages/argent-cli/test/flow.test.ts
@@ -163,6 +163,21 @@ describe("parseRunArgs", () => {
     );
   });
 
+  it("throws on a --platform value outside the four the help lists", () => {
+    for (const bad of ["iOS", "macos", "web", "android-tv"]) {
+      expect(() => parseRunArgs(["checkout.yaml", "--platform", bad])).toThrow(FlagParseException);
+      expect(() => parseRunArgs(["checkout.yaml", "--platform", bad])).toThrow(
+        `--platform must be "ios", "android", "chromium" or "vega", got "${bad}"`
+      );
+      expect(() => parseRunArgs(["checkout.yaml", `--platform=${bad}`])).toThrow(
+        FlagParseException
+      );
+    }
+    for (const good of ["ios", "android", "chromium", "vega"]) {
+      expect(parseRunArgs(["checkout.yaml", "--platform", good]).platform).toBe(good);
+    }
+  });
+
   it("treats a following flag as a missing value, not as the value", () => {
     expect(() => parseRunArgs(["checkout.yaml", "--device", "--json"])).toThrow(
       "--device requires a value"
@@ -1642,6 +1657,17 @@ describe("argent flow run <dir>", () => {
     ]);
     expect(errs.join("\n")).toContain(
       "--json-stream supports a single flow; directory runs are not supported"
+    );
+  });
+
+  it("rejects a misspelled --platform before running any flow in the batch", async () => {
+    await expect(flow(["run", flowsDir, "--platform", "iOS"], opts)).rejects.toThrow(
+      "process.exit:2"
+    );
+
+    expect(toolsClientMock.callTool).not.toHaveBeenCalled();
+    expect(errs.join("\n")).toContain(
+      '--platform must be "ios", "android", "chromium" or "vega", got "iOS"'
     );
   });
 


packages/argent-cli/src/flow.ts
```diff
@@ -161,7 +154,10 @@ const RUN_OPTIONS = {
   "json-stream": { kind: "boolean" },
   "recursive": { kind: "boolean", alias: "r" },
   "device": { kind: "value" },
-  "platform": { kind: "value" },
+  // Constrained here because the tool-server's own rejection is a raw Zod issue
+  // dump — and in a directory run it lands only after the first flow has run,
+  // taking the rest of the batch down with it.
+  "platform": { kind: "value", choices: ["ios", "android", "chromium", "vega"] },
```
latekvo, line 160: 1. This is missing `ios-remote`
2. Double-check whether we want to duplicate this source of truth, there 99% are better ways to do this.

PR #1146, opened 2026-09-09
fix(tools-client): kill the wedged server before respawning it on Windows

### LINKED ISSUE #832: Windows: ensureToolsServer never kills the wedged server before respawning it (identity guard unguarded at one of three call sites)
\[[Diplomat](https://github.com/latekvo/Diplomat), Opus 5\]: Found while reviewing #713, which hardens the same guard. Out of that PR's scope, filing separately.

## The asymmetry

`processCommandMatches` cannot work on Windows — there is no `ps`, and the repo says so in three places, including a test that skips itself to avoid SIGTERMing the runner (`packages/argent-tools-client/test/launcher-state.test.ts:241`). Two of its three call sites account for that and fall back to an unguarded kill:

`sweepDeadStateFiles`, launcher.ts:560-563
```ts
// Windows `ps` is unavailable and the check always fails, so the kill
// stays unguarded there rather than never retiring dead-bundle servers.
const guarded = process.platform !== "win32";
if (guarded && !processCommandMatches(fresh.pid, fresh.bundlePath)) continue;
```

`killToolServerForInstallDir`, launcher.ts:685-687 — same pattern.

`ensureToolsServer`'s kill-before-respawn, launcher.ts:931-938 — no guard:
```ts
if (
  state &&
  state.managed === "autospawn" &&
  state.bundlePath === paths.bundlePath &&
  isProcessAlive(state.pid) &&
  processCommandMatches(state.pid, state.bundlePath)
) {
  await terminatePid(state.pid, () => processCommandMatches(state.pid, state.bundlePath));
}
```

Blame: the unguarded site is `5f9dd48e2a` (2026-06-29); the two `guarded` siblings came later, in #510 (`07427917e2`, 2026-07-15). The guard was added to two of the three.

## Consequence on Windows

`processCommandMatches` returns false unconditionally, so the `if` never fires and the wedged autospawned server is not terminated. `clearToolsServerState(paths.bundlePath)` then unlinks its record and a replacement spawns on a fresh port. The old process keeps its listening port and is untracked — `sweepDeadStateFiles` cannot retire it because there is no longer a record to find, and `argent server stop`/status cannot reach it.

`PS_BIN` (added in #713) does not change this: neither `/bin/ps` nor `/usr/bin/ps` exists on Windows, so it falls back to bare `"ps"`. A Git-Bash/MSYS2 `ps.exe` on PATH does not help either — it rejects `-o command=`.

Reachability: the block needs the fast path to miss, the double-checked `reusableHandle` to return null (a health failure **or** a version mismatch — the version gate returns null even for a healthy server), `managed === "autospawn"` (written unconditionally, platform-independent), matching `bundlePath`, and `isProcessAlive` (`process.kill(pid, 0)`, which Node supports on Windows).

## Why this is LOW and not worse

The orphan reaps itself. `AUTOSPAWN_IDLE_TIMEOUT_MINUTES = 30` reaches the server as `ARGENT_IDLE_TIMEOUT_MINUTES` and drives `utils/idle-timer.ts`. Verified against the real built server:

```
$ ARGENT_IDLE_TIMEOUT_MINUTES=1 ARGENT_PORT=39117 node packages/tool-server/dist/index.js start
[16:42:05] Tools server listening on http://127.0.0.1:39117
[16:42:05]   Idle timeout: 1min
[16:43:05] [argent] No activity for 1min — shutting down
→ process gone
```

`/health` does not touch the idle timer (only `/tools`, `/tools/:name`, `/upload`, `/chromium-server/*` do), so nothing keeps an abandoned server warm. And `AUTOSPAWN_IDLE_TIMEOUT_MINUTES` was introduced in the same commit as the `managed` field (`8efd7c634`, #182), so any record that can match `managed === "autospawn"` was written by a version that also armed the timer — the reap is guaranteed by construction, not incidental.

So the real cost is one extra node process plus one ephemeral port for up to ~31 minutes, not a leak until reboot. The residual gap is a server whose event loop is hard-blocked, where the interval cannot run — exactly the case the kill exists for, and where Windows `process.kill(pid, "SIGTERM")` is an unconditional terminate rather than a handleable signal.

## Fix

Mirror the siblings:

```ts
const guarded = process.platform !== "win32";
if (
  state &&
  state.managed === "autospawn" &&
  state.bundlePath === paths.bundlePath &&
  isProcessAlive(state.pid) &&
  (!guarded || processCommandMatches(state.pid, state.bundlePath))
) {
  await terminatePid(
    state.pid,
    guarded ? () => processCommandMatches(state.pid, state.bundlePath) : undefined
  );
}
```

This site is already narrower than either sibling — it also requires `managed === "autospawn"` and `bundlePath === ours` — so the pid-reuse risk it takes on is strictly smaller than the one they already accept. The comment above the block enumerates "four guards"; it needs to say the command-line match applies only where `ps` exists.

Wants verification on a real Windows host, which is why it is not folded into #713.

latekvo:
Fixes #832

**Cause:** `ensureToolsServer`'s kill-before-respawn gates on `processCommandMatches`, which shells out to `ps` - absent on Windows, so it returns false for every pid. The wedged autospawned server is never terminated, its record is then unlinked, and it keeps its listening port with nothing left to find it by (`server stop`/status and the sweep both need the record).

**Fix:** drop the command-line check on win32, exactly as its two siblings already do (`sweepDeadStateFiles`, `killToolServerForInstallDir`). This site keeps its two narrower guards - `managed === "autospawn"` and `bundlePath === ours` - so it takes on less pid-reuse risk than either sibling.

**Class:** those two were the only other `processCommandMatches` call sites, and both are already guarded. `killToolServer` also calls `terminatePid` unguarded, but on every platform by design - it is the explicit `argent server stop` kill of a record the user named, with no identity check to conditionalize.

**Docs:** none needed - no tool, CLI, config key, flow file or user-facing capability changes.

<details>

New test in `launcher-duplicate-spawn.test.ts`: a live bystander whose command line does not match the recorded bundle, recorded as the tracked autospawn server, with `process.platform` stubbed to `win32` for the `ensureToolsServer` call.

Fails on the unfixed source:

```
FAIL test/launcher-duplicate-spawn.test.ts > ensureToolsServer - duplicate-spawn prevention >
     still terminates the wedged server it respawns over where `ps` cannot identify it
AssertionError: expected true to be false
 ❯ expect(launcher.isToolsServerProcessAlive(wedgedPid)).toBe(false);
```

Passes with it. The pre-existing `never signals a recycled pid` test - same setup, real platform - still passes, so the non-Windows guard is unchanged.

```
npx tsc --build                                  exit 0
npm run typecheck:tests -w @argent/tools-client   exit 0
npm test -w @argent/tools-client                  12 files, 197 tests passed
npx prettier --write <changed files>              unchanged
npx eslint packages/argent-tools-client/src/launcher.ts   clean
```

Not run on a real Windows host; the platform branch is exercised through the stub.
</details>

### DIFF
--- a/packages/argent-tools-client/src/launcher.ts
+++ b/packages/argent-tools-client/src/launcher.ts
@@ -842,15 +842,23 @@ export async function ensureToolsServer(paths: ToolsServerPaths): Promise<ToolsS
     //   • bundlePath === ours — never a different version's server, which may be
     //     healthy and serving another project's session;
     //   • a command-line identity match, re-confirmed by terminatePid right
-    //     before each signal.
+    //     before each signal — only where `ps` exists. On Windows the check
+    //     always fails, so the kill stays unguarded there rather than leaving
+    //     the wedged server running, untracked, on a leaked port. The two
+    //     guards above still hold, so this site takes on less pid-reuse risk
+    //     than the other two kills that go unguarded on Windows.
+    const guarded = process.platform !== "win32";
     if (
       state &&
       state.managed === "autospawn" &&
       state.bundlePath === paths.bundlePath &&
       isProcessAlive(state.pid) &&
-      processCommandMatches(state.pid, state.bundlePath)
+      (!guarded || processCommandMatches(state.pid, state.bundlePath))
     ) {
-      await terminatePid(state.pid, () => processCommandMatches(state.pid, state.bundlePath));
+      await terminatePid(
+        state.pid,
+        guarded ? () => processCommandMatches(state.pid, state.bundlePath) : undefined
+      );
     }
     // Retire only OUR OWN record — another install's must survive so its server
     // stays reachable by its owner. The sweep then clears per-bundle files whose
--- a/packages/argent-tools-client/test/launcher-duplicate-spawn.test.ts
+++ b/packages/argent-tools-client/test/launcher-duplicate-spawn.test.ts
@@ -184,6 +184,55 @@ describe("ensureToolsServer — duplicate-spawn prevention (nvm node-version swi
   );
 
   it(
+    "still terminates the wedged server it respawns over where `ps` cannot identify it",
+    { timeout: 30_000 },
+    async () => {
+      // Windows has no `ps`, so processCommandMatches returns false for every
+      // pid. Gating the kill on it there would leave the wedged autospawned
+      // server holding its port after its record is unlinked, with nothing left
+      // to find it by. Its command line does not match the recorded bundle, so
+      // the guard would reject it on a platform that does have `ps`.
+      const wedged = spawn(process.execPath, ["-e", "setInterval(() => {}, 1e9)"], {
+        stdio: "ignore",
+      });
+      const wedgedPid = wedged.pid!;
+      spawnedPids.push(wedgedPid);
+      expect(launcher.isToolsServerProcessAlive(wedgedPid)).toBe(true);
+
+      await launcher.writeToolsServerState({
+        port: 1, // nothing listens here -> health check fails fast
+        pid: wedgedPid,
+        startedAt: new Date().toISOString(),
+        bundlePath: FAKE_BUNDLE,
+        host: "127.0.0.1",
+        token: "wedged-token",
+        managed: "autospawn",
+      });
+
+      const realPlatform = process.platform;
+      Object.defineProperty(process, "platform", { value: "win32", configurable: true });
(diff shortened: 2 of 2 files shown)


latekvo:
Not reproducible on current main: #713 (f106e486e, merged 2026-09-03) hoisted the win32 carve-out into `couldBeOurToolServer` (`if (process.platform === "win32") return true;`), which covers all three call sites including ensureToolsServer's kill-before-respawn. This branch was cut from a stale local main ref that predates it.

PR #1151, opened 2026-09-09
fix(screenshot): give ARGENT_SCREENSHOT_SCALE the 0.01 floor its parameter enforces

### LINKED ISSUE #816: ARGENT_SCREENSHOT_SCALE has no 0.01 floor and swallows out-of-range values, turning a diff into dimension_mismatch
## Summary

`getScreenshotScale()` accepts a band of `ARGENT_SCREENSHOT_SCALE` values that the `scale` parameter
of the same tool rejects, and silently swallows out-of-range ones. The low end is the damaging half:
a scale below the schema's `0.01` floor makes `screenshot-diff`'s live capture too small to compare
against any saved baseline, and the tool reports `dimension_mismatch` rather than a configuration
error.

`packages/tool-server/src/utils/simulator-client.ts:198-205`:

```ts
export function getScreenshotScale(): number {
  const v = process.env.ARGENT_SCREENSHOT_SCALE;
  if (v) {
    const n = parseFloat(v);
    if (!Number.isNaN(n) && n > 0 && n <= 1) return n;
  }
  return DEFAULT_SCREENSHOT_SCALE;
}
```

The guard is `n > 0 && n <= 1`. The `scale` parameter on `screenshot` is `z.number().min(0.01).max(1.0)`
(`packages/tool-server/src/tools/screenshot/index.ts:30-34`), so the two disagree over `(0, 0.01)`.

## Reproduction

```
$ npx tsx -e 'import { getScreenshotScale } from "./src/utils/simulator-client";
  for (const v of ["1e-3","0.0001","30","1.5","abc",""]) {
    process.env.ARGENT_SCREENSHOT_SCALE = v; console.log(JSON.stringify(v), "->", getScreenshotScale());
  }'
"1e-3"   -> 0.001
"0.0001" -> 0.0001
"30"     -> 0.3
"1.5"    -> 0.3
"abc"    -> 0.3
""       -> 0.3
```

Downstream, with the same file's `normalizeToCommonSize`:

```
scale 0.001:  1080x2400 vs 1x2 -> null (dimension_mismatch)
scale 0.0001: 1080x2400 vs 0x0 -> null (dimension_mismatch)
```

So with `ARGENT_SCREENSHOT_SCALE=1e-3`, a `screenshot-diff` live capture that falls back to the
tool-server scale comes back at 1×2, `normalizeToCommonSize` refuses the pair, and the caller is told
the two screenshots have mismatched dimensions — which is true but describes the configuration, not
the screens. At `0.0001` the fallback size rounds to 0×0.

## Two distinct halves

1. **No floor.** Anything in `(0, 0.01)` is accepted by the env parser and rejected by the parameter
   that documents the same range. Whatever floor `scale` enforces, the env var should enforce too.
2. **Silent fallback on an invalid value.** `30`, `1.5` and `abc` all resolve to `0.3` with no
   warning anywhere. `30` is a plausible misreading of the skills' own "30% of original resolution"
   phrasing, and the user gets the default while believing the var took effect.

## Related

#626, #815 — other cases of a screenshot geometry input being accepted and silently discarded.

latekvo:
Fixes #816

**Cause** - `getScreenshotScale()` guarded on `n > 0 && n <= 1`, while the `scale` parameter of the same tool is `min(0.01).max(1.0)`; a value in `(0, 0.01)` was accepted and shrank a capture towards zero pixels, so `screenshot-diff` refused the pair and reported `dimension_mismatch` rather than a configuration problem. An out-of-range value was also swallowed with no warning, so `ARGENT_SCREENSHOT_SCALE=30` silently read as the default.

**Fix** - the guard is now `n >= 0.01 && n <= 1`, and a rejected value is reported once per distinct value on stderr, naming the value and the accepted range.

**Verified** - the three new floor cases and the warning assertion fail on the unfixed function and pass with it; full `@argent/tool-server` suite green.

Docs: `packages/docs/docs/reference/configuration.mdx` - the `ARGENT_SCREENSHOT_SCALE` row said `0` to `1` and a `0.3` default; it now says `0.01` to `1`, default `0.25`, and that an out-of-range value is ignored with a warning. No other reference or feature page mentions the variable.

Class check: `getScreenshotScale()` is the only env-to-scale parser in the tool-server (iOS, Android, tvOS and Vega all read it), so there is no sibling site to fix.

<details>
<summary>Discriminating-test proof and checks</summary>

Source fix reverted, test file kept:

```
FAIL  test/screenshot-default-scale.test.ts > falls back to 0.25 below the floor for "0.009"
  expected 0.009 to be 0.25
FAIL  test/screenshot-default-scale.test.ts > falls back to 0.25 below the floor for "1e-3"
  expected 0.001 to be 0.25
FAIL  test/screenshot-default-scale.test.ts > falls back to 0.25 below the floor for "0.0001"
  expected 0.0001 to be 0.25
FAIL  test/screenshot-default-scale.test.ts > names the ignored value and the accepted range, once per value
  expected "warn" to be called 1 times, but got 0 times

Test Files  1 failed (1)
     Tests  4 failed | 9 passed (13)
```

Fix restored: 13 passed.

Checks run from the worktree root:

- `npx tsc --build` - clean
- `npm test -w @argent/tool-server` - 368 files, 4837 passed, 1 skipped
- `npx prettier --check` on the three changed files - clean

`npx eslint` on the changed files could not run locally: the root install skips `packages/docs`, so its `tsconfig.json` fails to resolve `@docusaurus/tsconfig` and the shared project parse errors out. Same reason `npx docusaurus build` was not run for the one-cell docs table edit.

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Screenshot scaling now rejects values below `0.01` or above `1` and falls back to the default when invalid.
  * Invalid screenshot scale values display a warning only once per distinct value.

* **Documentation**
  * Updated the documented default screenshot scale to `0.25`.
  * Documented the accepted range as `0.01` to `1`.

### DIFF
--- a/packages/docs/docs/reference/configuration.mdx
+++ b/packages/docs/docs/reference/configuration.mdx
@@ -175,15 +175,15 @@ An environment variable changes the behavior of the process that reads it. Expor
 
 ### Tool-server
 
-| Variable                      | Default                              | Description                                                                                         |
-| ----------------------------- | ------------------------------------ | --------------------------------------------------------------------------------------------------- |
-| `ARGENT_PORT`                 | `3001`                               | The port the tool-server listens on                                                                 |
-| `ARGENT_HOST`                 | `127.0.0.1`                          | The address the tool-server binds to                                                                |
-| `ARGENT_IDLE_TIMEOUT_MINUTES` | `0`                                  | Minutes without a request after which the tool-server exits. `0` disables the timeout               |
-| `ARGENT_EVENT_LOG`            | `~/.argent/tool-server-events.jsonl` | Path of the event log that the `tool-server-event-log` flag writes                                  |
-| `ARGENT_SCREENSHOT_SCALE`     | `0.3`                                | Scale factor of a captured screenshot, between `0` and `1`. A smaller value returns a smaller image |
-| `ARGENT_CHROMIUM_PORTS`       | empty                                | Comma-separated CDP ports that Argent probes for Chromium apps, in addition to `9222`               |
-| `ARGENT_CHROMIUM_PORTS_FILE`  | `~/.argent/chromium-cdp-ports.json`  | Path of the file that remembers the CDP ports of apps that Argent started                           |
+| Variable                      | Default                              | Description                                                                                                                                                   |
+| ----------------------------- | ------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `ARGENT_PORT`                 | `3001`                               | The port the tool-server listens on                                                                                                                           |
+| `ARGENT_HOST`                 | `127.0.0.1`                          | The address the tool-server binds to                                                                                                                          |
+| `ARGENT_IDLE_TIMEOUT_MINUTES` | `0`                                  | Minutes without a request after which the tool-server exits. `0` disables the timeout                                                                         |
+| `ARGENT_EVENT_LOG`            | `~/.argent/tool-server-events.jsonl` | Path of the event log that the `tool-server-event-log` flag writes                                                                                            |
+| `ARGENT_SCREENSHOT_SCALE`     | `0.25`                               | Scale factor of a captured screenshot, between `0.01` and `1`. A smaller value returns a smaller image. A value outside that range is ignored, with a warning |
+| `ARGENT_CHROMIUM_PORTS`       | empty                                | Comma-separated CDP ports that Argent probes for Chromium apps, in addition to `9222`                                                                         |
+| `ARGENT_CHROMIUM_PORTS_FILE`  | `~/.argent/chromium-cdp-ports.json`  | Path of the file that remembers the CDP ports of apps that Argent started                                                                                     |
 
 ### Devices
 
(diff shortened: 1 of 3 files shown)


packages/docs/docs/reference/configuration.mdx
```diff
@@ -183,25 +175,24 @@ An environment variable changes the behavior of the process that reads it. Expor
...
+| `ARGENT_PORT`                 | `3001`                               | The port the tool-server listens on                                                                                                                           |
+| `ARGENT_HOST`                 | `127.0.0.1`                          | The address the tool-server binds to                                                                                                                          |
+| `ARGENT_IDLE_TIMEOUT_MINUTES` | `0`                                  | Minutes without a request after which the tool-server exits. `0` disables the timeout                                                                         |
+| `ARGENT_EVENT_LOG`            | `~/.argent/tool-server-events.jsonl` | Path of the event log that the `tool-server-event-log` flag writes                                                                                            |
+| `ARGENT_SCREENSHOT_SCALE`     | `0.25`                               | Scale factor of a captured screenshot, between `0.01` and `1`. A smaller value returns a smaller image. A value outside that range is ignored, with a warning |
```
latekvo:
```suggestion
| `ARGENT_SCREENSHOT_SCALE`     | `0.25`                               | Scale factor of a captured screenshot, between `0.01` and `1`. A smaller value returns a smaller image. |
```

PR #1166, opened 2026-09-14
fix(tool-server): protect simulator HID from CoreDevice suppression

### LINKED ISSUE #932: iOS sim HID injection silently no-ops on macOS 27 beta 26A5416b (CoreDevice dtuhidd owns sim HID); #547 killall workaround ineffective

latekvo:
> **Reworked after live testing.** Two designs in this PR did not survive measurement, and the sections below are updated. Kept rather than rewritten, because the failures explain the shape of what ships.

## What ships

**Protection** — `boot-device` spawns simulator-server's new `hid_warmup` one-shot **before** `simctl boot`, concurrently with it (software-mansion/radon#218). Eagerly resolving the whole SimulatorServer service during boot, which this PR did before, could not win: standing up the process and its transports takes ~3 s while the window shuts by ~1.2 s, and every measured cold boot lost all three services. The one-shot retries its attach, so firing it at T0 lands the first event at ~300 ms.

**Detection** — a background probe, once per boot, surfacing a `hint` on `gesture-tap` / `button` / `keyboard` naming which services are not receiving input.

### Measured, iOS 18.6, through this path

48 cold boots: **92–95%** of boots with a winnable window kept their buttons and keyboard; digitizer **47/48**. Negative control with the warm-up neutered: buttons and keyboard dead **9/9**.

### The detection traces were wrong twice

Both mistakes are now pinned by `test/hid-suppression.test.ts` (table-driven over verbatim captured `backboardd` lines, no simulator needed — it fails on each of the old predicates):

- `contact 1 presence: none` and `Keyboard receives keyEvent` only ever appear for a **real press**. Keyed on those, `touch` could never be true and `keyboard` was true only by accident, via the unrelated key-down frame kick `Ios::new` fires on attach. **Every healthy simulator was told its taps were broken and to reboot.**
- Bare `downEvent:0` appears on touch lines *and* button lines, so an unscoped match reported a live digitizer whenever the buttons happened to be alive.

Detection now reads correctly on **19 of 19** boots across two runtimes: no hint on a healthy simulator, correct service naming on a broken one, once per boot, re-armed by a reboot.

### The remedy no longer loops forever

It used to say "repeating it usually does" fix it. On **iOS 26.5 (23F77)** that advice is an unbounded reboot loop that destroys app state every iteration and can never succeed: across 12 boots of a warmed device the buttons and keyboard were torn down **12 of 12**, window −289 ms to +23 ms (median −196 ms). It now caps at two reboots and says to stop. Docs carry the same caveat.

The cause there is an early flag, not late services — connect times match 18.6, while the flag lands at a median 1.119 s against 18.6's 1.24–2.53 s. (An earlier revision of this PR said the opposite; that reading came from a freshly created device, where first-run work delays the connects.) Touch keeps working on 26.5 because the flag beats `createDigitizerForTargetID:`, which then skips the connect — in 11 of 12 the digitizer was never registered at all. One device, one host, n=12; 26.3 and 27.0 untested.

### Also

The warm-up spawn is gated to non-tvOS — `bootIos` serves tvOS simulators too, which have no main-screen digitizer, and an Indigo event naming a target absent from the service table takes `backboardd` down. Only the detection probe had been gated. **Untested**: no tvOS runtime on this machine.

---

<details>
<summary>Original analysis (mechanism, escape hatch, scope) — still accurate, kept for reference</summary>

Fixes #932 (partially — see *Scope* below).

## What breaks, and why it is silent

A CoreDevice client on the host demand-starts `dtuhidd` inside the guest. DeviceHub is the usual culprit and it attaches to **every** simulator that boots, so simply leaving it open is enough. `dtuhidd` sets `com.apple.coredevice.dtuhidd.active`, and `backboardd` tears down three virtual HID services: the main-screen digitizer, the main-screen buttons, and the external keyboard.

The teardown is permanent for `backboardd`'s lifetime — `IOHIDEventSystemRemoveService` *terminates* the service rather than unregistering it, and the reconnect path re-adds the same dead object. Every injected touch, button and key is discarded from then on. And because `sendCommand` is fire-and-forget, `gesture-tap` still returns `{ tapped: true }` for all of them, which is why this presented as "taps do nothing, no error".

## What this PR does

`backboardd` exempts any service that has already seen **one** Indigo HID event, and skips exempt services when tearing down. So a single event per service, delivered before the flag rises, protects it for the whole boot — across later flag changes, app switches and SpringBoard restarts.

- **Warm up on every attach** (`simulator-server` blueprint). Invisible and idempotent, so it is unconditional.
- **Bring simulator-server up during `boot-device`** instead of lazily on the first gesture. `boot-device` resolves `NativeDevtools`, not `SimulatorServer` — they are separate services — so without this the warm-up would not happen until the first interactive tool call, leaving everything in between unprotected.
- **Detect and report.** A per-service probe runs once per boot, 3 s after attach, and the interaction tools surface the verdict as a `hint` naming exactly what is dead and telling the agent to reboot through argent. Said once per boot, re-armed by a reboot.
- **Detection helpers** in a new `utils/hid-suppression.ts`.

### The events are releases without a press

`touch Up` with no `Down`, `button Up`, `key Up`. They set the exemption flags but carry no semantics — no contact, no button press, no keystroke. Verified pixel-identical against a foregrounded app **with a modal sheet presented**. A stray Home release dismissing someone's alert would have been a serious regression, so this was tested specifically.

Known cosmetic side effect, commented at the call site: the keyboard warm-up makes `backboardd` log `missing a sequence for <senderID…>` on every attach. Harmless, but it will otherwise send someone debugging simulator logs on a detour.

### Detection: the obvious detector is inverted

`flag == 1` is the **normal state of a healthy simulator** whenever a CoreDevice client is running — measured on 4 of 5 fresh boots, with taps landing fine. A detector keyed on it would tell almost every affected-machine user their simulator was broken. `readSuppressionFlag` is therefore documented as diagnostic only. Two further traps are documented in the module: a host-side read of that key is a silent false negative (the guest has its own `notifyd`), and the `dtuhidd-suppressed` log line also appears in working states.

`probeHidServices` is the real signal. Because the warm-up deliberately produces **no** contact, it cannot be verified by counting contacts — so it looks for the orphan-release trace instead, which is present exactly when the services are alive. Verified both ways on hardware: healthy device → trace present and a real tap lands; suppressed device → trace absent.

It reports the digitizer and the buttons/keyboard pair **separately**, because they do not fail together. The digitizer survives boots that kill the other two, so a single boolean reports a healthy simulator on exactly the boots worth reporting. Two defects found while wiring it up:

- The original `probeHidDelivery` matched its predicate against raw `log show` output — whose preamble echoes the predicate text back, containing both literals it searches for. It would have returned `true` for every simulator, dead ones included.
- Probing from inside a tool call injects HID events mid-gesture; a stray touch `Up` between a drag's `Down` and its `Up` ends the drag early. The probe therefore runs only at attach, where the warm-up already fires, and tools read cached state.

A third defect, unrelated to detection: `sendHidWarmUp` ran for **every** spawned simulator-server, including Android emulators and tvOS, where this suppression does not exist. Now gated to iOS simulators.

## Scope — what this does *not* fix

**This cannot win a cold-boot race**, and I want to be explicit about that rather than have it discovered later. The window opens ~1.1 s after boot and can close 166 ms later; `simulator-server` does not accept a command over the WebSocket until ~7.1 s, roughly 5 s after the window has shut. Measured success of warm-up-on-attach on a cold boot: **0/6**.

The cold-boot fix belongs in `simulator-server`, which holds a usable HID bridge at ~1.0 s: **software-mansion/radon#218**. This PR is the complementary half — it covers a **mid-session demand-start**, i.e. someone opening DeviceHub while argent is already attached, which is otherwise an instant and silent break.

Deliberately **not** included:

- **Self-heal** (clear flag → `SIGTERM backboardd` → re-warm). It works and is validated, but it kills the user's foreground app and bounces SpringBoard, and it rests on a probe whose false-negative rate is still unmeasured. That is not something to land automatically in the same PR as the prevention; worth its own change once the probe has been characterised.
~~- **A user-facing error.**~~ **Now included.** The concern was false positives, and the fix for that was to stop keying on the flag: the `hint` fires on a probe that measures whether events actually arrive, never on `flag == 1`. It names which services are dead, and it is said once per boot — re-armed when the boot id changes, so an agent that reboots into another bad race is told again rather than left believing it is fixed.

## The zero-code workaround worth knowing

A running DeviceHub is what triggers this. **Quitting it returns machines to a healthy state with no code change at all**, and it reframes the affected population from "everyone on Xcode 27" to "people who leave DeviceHub open". Worth documenting for users regardless of what ships here.

## Residual: buttons and keyboard on a fast host

The warm-up only protects a service if an event reaches it **before** the flag rises. On a fast machine that is not always possible: measured per-service across 47 cold boots, 10 of 47 (21%) had a window at or below zero, and on those the buttons and external keyboard are torn down before anything can reach them.

The routes back are closed at the instruction level — `ensureServiceActiveForIndigoHIDData:` constructs nothing on a table miss, the suppressed-reconnect re-adds the same terminated object, and the inactive bit at `+0xE0` has no clearing instruction and sits behind a task port SIP does not grant. Detail in software-mansion/radon#218.

This is what the `hint` exists for. The advice converges rather than guaranteeing: a negative window is a per-boot race, so a reboot clears it ~80% of the time and a second ~96%. Being told again after an unlucky reboot is the point.

## Testing

- `tsc --noEmit` clean; `prettier` applied.
- **446/446 test files, 6307 tests passing.**
- **Five tests were already failing on this branch and passing at the merge-base** — this PR broke them and it had not reached CI. One was the Android warm-up bug above; the other four asserted on event ordinals that the warm-up shifted. Updated to count only what they mean to, plus a new assertion that `SimulatorServer` resolves *before* `bootstatus`, which is the ordering the `boot-device` change exists for.
- Warm-up mechanism verified on hardware: with all three sent, raising the flag produces **0** `Disconnecting service` lines; without them, **3**. Each event alone accounts for exactly its own service.
- `touch:"Move"` does **not** set the digitizer flag — only `Up` does.
- `notifyutil` get/set/post syntax verified inside the guest, including that `simctl spawn` requires a bare binary name (an absolute path fails with `SimXPCErrorDomain 111`).
- The probe signal verified to discriminate healthy from suppressed on real devices.

Not covered by automated tests — the behaviour only reproduces against a booted simulator with a CoreDevice client running.

## Background

- Root cause: https://claude.ai/code/artifact/31192f0a-ef14-4c03-811b-1f7dd1d84756
- How the host/guest boundary works: https://claude.ai/code/artifact/95305551-6692-4e3b-a85c-c6ee68df3c30
- Implementation plan and measurements: https://claude.ai/code/artifact/4c6ac6f7-132e-407e-bb6a-99bbd2c88151

https://claude.ai/code/session_01Kc7zdiDeePHMtT4LJWj7Pr

</details>


filip131311:
Pushed a917b0311 (+ merge of your 0cbdc1c49) that trims this to the protection half:

- **Dropped the detection half** — the per-attach probe, in-guest `log show`, per-boot health cache and the `hint` on gesture-tap / button / keyboard, plus the description, result-type, blueprint and MoQ-test changes that carried it. Its one correct case (a simulator booted outside argent, taps dead) is outweighed by what your own 26.5 numbers now say: buttons and keyboard are torn down 12/12 regardless, so on iOS 26+ the hint fires on every boot and prescribes reboots that cannot succeed. It was also only on three tools; swipe / custom / pinch / rotate / flows were never told.
- **Gated `startHidWarmUp` on `needsPreBoot`.** It ran unconditionally, so once the binary lands every non-force `boot-device` on an already-booted simulator would have waited ≥6 s while the one-shot fired release events at the live app. Test added for both the booted and the force paths.
- Docs section rewritten without the hint; header comment keeps your corrected 26.5 mechanism.

Live check on an Xcode 26.6 host (no DeviceHub, iOS 26.5 sim): the probe read healthy correctly there, so this was not about false positives on the common population.

Sequencing note: with the shipped 0.25.0 `simulator-server` the spawn logs `Unrecognized argument: hid_warmup` and is a no-op. This needs software-mansion/radon#218 released and a bin bump before merging, or it ships nothing.

Verified: `tsc --noEmit` clean, tool-server suite 447 files / 6316 tests green, prettier, knip, `docusaurus build`.

PR #1167, opened 2026-09-14
fix(android-devtools): reinstall the helper when am instrument reports it missing, and say why it failed

### LINKED ISSUE #1162: Android QA: live inspection works but flow helper fails instrumentation; readiness and capture feedback
During an Android QA session, live UI interaction worked but flow recording's Android helper failed. Some post-action screenshots also appeared to show the source screen while the subsequently returned/queried tree represented the destination, leading to extra inspection calls.

## Environment

- macOS host; installed Argent CLI reports 0.25.0 at report time (version was not separately recorded at session start).
- Dedicated headless Android emulator, Android16/API36, Google sdk_gphone64_arm64.
- Expo57 / React Native0.86 development client, local Metro on8091.
- Report-only QA: no helper repair/reinstall loop attempted, so root cause is not established.

## Flow helper failure

1. Launch a development app on an unlocked Android emulator and connect to Metro.
2. Confirm `gesture-tap`, `describe` (UIAutomator) and `debugger-component-tree` can inspect/interact with the app.
3. Start a fragment with `flow-start-recording` on an existing screen with testIDs.
4. Record `gesture-tap` using a target discovered from the live tree.
5. Record `await-ui-element` for a destination testID.

Observed on each of two short diagnostic fragments:

```text
selector capture failed (the android devtools helper is unavailable
[AndroidDevtools:<serial>] am instrument exited before becoming ready
(code=1 signal=null).
android.util.AndroidException: INSTRUMENTATION_FAILED:
com.argent.androiddevtools/com.argent.androiddevtools.SnapshotInstrumentation
```

The recorder kept raw coordinates. A destination visibility wait passed against the live tree but could not be re-verified against the runner tree. These recordings were therefore not treated as replay-verified tests.

Expected: expose a cheap per-device readiness/capability check before flow authoring that actually probes the runner hierarchy, with a concise actionable failure. Where feasible, let recorder/live interaction share a working inspection path; any fallback should disclose its selector coverage rather than silently weakening targets. Instrumentation registration alone is not proof the helper starts.

## Post-action capture observation (needs investigation)

Several navigation taps returned an auto-screenshot still showing the previous screen. A later component tree exposed the destination. This was especially noticeable through login and modal navigation. No timestamp/frame trace was collected, so this could be transition timing rather than a stale capture cache; please treat it as an observation, not a proven root cause.

Useful contract: return capture timestamps/transition status and an opt-in wait for a known destination selector before producing the screenshot/tree pair. Prefer a bounded wait with an explicit timeout over an unconditional sleep.

## Token/latency improvements

- Compact active-screen trees with necessary overlay context, avoiding inactive screens beneath the current route.
- Direct semantic targeting for live taps, or a supported way to reuse runner selector resolution.
- A small unified readiness report separating app connection, live inspection, runner inspection and recording capability.

The diagnostic evidence remains local because it contains app-specific navigation/account UI. Error text above is sanitized. No additional reproduction or benchmark run was launched for this report.

filip131311:
Closes #1162.

## What was wrong

`flow-start-recording` + `gesture-tap` on Android failed every selector capture with `am instrument exited before becoming ready … INSTRUMENTATION_FAILED`, while `describe` silently fell back to `uiautomator`. Reproduced on a Pixel 9 / API 36 google_apis_playstore emulator (the reporter's image family): `am instrument` fails this way **only when `com.argent.androiddevtools` is not installed** — a disabled or suspended package still starts. Two things made argent stay in that state:

- `android-helper-install.ts` memoized "installed" per serial for the life of the tool-server, so a wiped or snapshot-restored emulator (or a new instance reusing `emulator-5554`) never got the helper back.
- `spawnHelper` only reported the stderr tail (the Java stack trace); the device's own reason, `INSTRUMENTATION_STATUS: Error=Unable to find instrumentation info…`, is printed on **stdout** and was discarded.

## What changes

- Probe the installed package on every service instantiation instead of memoizing.
- Capture the `INSTRUMENTATION_STATUS` reason from stdout and settle from `close` (with a short grace after `exit`), so the message names the device's verdict.
- When the device says the instrumentation is missing: force a reinstall (`-r -t -d`) once and retry the spawn. If it still fails, `ANDROID_DEVTOOLS_HELPER_REPAIR_FAILED` carries the device's own `Error=` line; an install that fails is `ANDROID_DEVTOOLS_HELPER_INSTALL_FAILED` with the adb failure code. The `flow-execute` launch gate reports the same reason.
- Every failed selector capture returns one short line with the device's reason and nothing else.

Deliberately not done (and why): a per-serial failure cooldown (a failing install was never memoized before either, so that is a separate improvement); an optional `udid` on `flow-start-recording` (device-agnostic tool, 108 test call sites, and the skill calls it before any device exists); a `protocol` marker (needs a Java change in `argent-private`); `am force-stop` on dispose (would kill a co-tenant tool-server's helper).

## Verification

Unit: one new test file (`android-devtools-repair.test.ts`, 7 cases) + two string updates; `npm test -w @argent/tool-server` 443 files / 6258 passed; build, knip, prettier clean. Source diff is ~290 lines; an adversarial live review (races, mid-spawn uninstall, lingering stdout, refused installs, foreign-signed APK) found no functional defect.

Live on emulator-5554 (API 36) with the branch tool-server, adb calls counted through a shim:

| Scenario | Before | After |
|---|---|---|
| helper uninstalled behind a running server, then record a tap | `selector capture failed (… INSTRUMENTATION_FAILED …)`, coordinates kept, describe → uiautomator | one reinstall, selector `id: com.android.settings:id/recycler_view` recorded, describe → android-devtools |
| package `pm disable-user`'d | — | no reinstall (0 installs) |
| helper uninstalled and the bundled APK unavailable, 3 taps | — | one short line per step, no crash, describe falls back |

## Docs

`packages/docs/docs/features/flows.mdx` and the create-flow skill's recovery table describe the reinstall and the commands. `npx docusaurus build` could not run on this machine (docs deps not installed); the change is prose only.

https://claude.ai/code/session_013LVd4LMSLafzeF67iEuncN


filip131311:
Addressed both CodeRabbit comments in 311043d3a: the message now says to run the adb command from the host, and a present helper whose versionCode the `pm list packages` fallback cannot read is treated as current (a stale one is still caught by the forced reinstall). Docs: `npx docusaurus build` now runs clean with the docs deps installed.

Issue #1161, opened 2026-09-12
Orphaned Android screen-sharing agent keeps encoding after simulator-server exits (flattens a physical device)


kesha-antonov:
## What happens

On a **physical** Android device, the screen-sharing agent set up on `launch-app` is never torn down when the host-side connection dies. It keeps H.264-encoding into a socket nobody reads, at ~100% of one CPU core, indefinitely.

Measured on a Pixel 7 (Android 17, argent 0.25.0): orphaned for **19 hours**, **796 mAh** of a 3950 mAh battery. That is 30% of total drain and 71% of all CPU drain for the window. Phone went 100% -> 20% in 4h45m.

## Log

```
[15:35:32] [sim <serial>] Setting up screen sharing agent for device: <serial>
[15:35:33] [sim <serial>] Video channel established with codec: avc
[16:06:40] [sim <serial>] ERROR ...::protocol] Error reading control message: unexpected end of file
[16:06:40] [sim <serial>] ERROR ...::android_device] Video streaming error: early eof
[17:44:59] [argent] No activity for 30min — shutting down
```

argent shut down cleanly on the host. The device-side process kept running for another 19 hours.

## Why the existing cleanup cannot fire

The on-device wrapper is:

```sh
trap "... | xargs kill" SIGHUP SIGINT SIGTERM EXIT
CLASSPATH=... app_process ... com.android.tools.screensharing.Main ...
```

POSIX `sh` defers trap handlers until the foreground command returns, and `app_process` never returns, so the trap is unreachable for as long as the agent is alive.

Confirmed by hand: `kill <wrapper sh pid>` does nothing at all. Only killing the `app_process` pid works, and the wrapper then exits and runs its trap.

## Repro

1. `launch-app` against a physical Android device
2. kill the host-side simulator-server, unplug the cable, or let the host sleep
3. `adb shell "ps -A -o PID,%CPU,ARGS | grep [s]creensharing"` - still running at ~100% CPU

## Suggested fix

Reap the device-side agent by pid when simulator-server shuts down and on the idle timeout, instead of relying on the wrapper's trap. A stale-agent sweep at device attach would also catch orphans left by a previously crashed host.

This only has a visible cost on physical devices - an emulator has no battery to flatten, so it goes unnoticed there.

Possibly related: #885

PR #1168, opened 2026-09-14
fix(tool-server): reap the on-device screen-sharing agent when a physical Android session ends

filip131311:
Part of #1161 (the argent side; the simulator-server side is software-mansion/radon#219). This PR is valuable on its own: the simulator-server binary argent ships is a rolling, unpinned `radon-main` download, so argent cannot assume the device-side fix is present, and on SIGKILL, host sleep, or Windows (`TerminateProcess`) no host-side signal ever reaches the binary.

## What was wrong

On a physical Android device the simulator-server starts a screen-sharing agent over adb. When the host side went away, nothing told the device, and the agent kept H.264-encoding at ~100% CPU for hours (796 mAh in the report). The on-device wrapper's trap could never fire (POSIX defers a trapped signal while a foreground `app_process` runs), and argent's `dispose` was a bare `proc.kill()`.

## What changes

- `simulator-server.ts`: for the `android_device` subcommand only, snapshot `adb reverse --list` before spawn and once ready; the one new `localabstract:screen-sharing-agent-<port>` entry is this session's socket. A failed probe yields `null`, never an empty set, so a foreign agent is never attributed to us (Android Studio's "Running Devices" mirror uses the same class and path).
- `dispose` (explicit stop, stop-all, idle shutdown, or `terminated` after the process exited by itself): SIGTERM, wait up to 2.5 s for the binary's own teardown, then `pkill -f 'screensharing[.]Main.*--socket=<socket>'` (bracket-escaped: unescaped, `pkill -f` matches its own `sh -c` and exits 143) and `adb reverse --remove`. Each adb call is bounded at 5 s; pkill exit 1 (nothing to reap) is silent, exit 0 logs one line. The ready-timeout path reaps too. Emulators and iOS are untouched.
- New `utils/android-agent-reap.ts`.

## Verification

Unit: `test/simulator-server-blueprint.test.ts` 22 passed (7 new cases incl. a real-Registry `terminated` path, the failed-probe/foreign-socket case mutation-checked); full tool-server suite 6260 passed; build, knip, prettier clean.

Live: routed the API 36 emulator through the physical-device path (`adb tcpip 5555; adb connect 127.0.0.1:5555` → kind "device") and drove the branch tool-server over HTTP with both the installed 0.25.0 binary and the patched one. After `stop-simulator-server`, `kill -9` of the simulator-server (terminated path) and `stop-all-simulator-servers`: no agent, no wrapper shell, no reverse entry, no simulator-server process. With the old binary the TS reap reports "stopped a leftover screen-sharing agent"; with the patched binary the binary cleans up itself and the TS side stays silent.

## Docs

`packages/docs/docs/fundamentals/supported-platforms.mdx`: the physical-Android section now says the agent is stopped when the session ends and names the manual command for a host that died abruptly (with the Android Studio caveat). `npx docusaurus build` could not run here (docs deps not installed); prose only.

https://claude.ai/code/session_013LVd4LMSLafzeF67iEuncN


filip131311:
Addressed both CodeRabbit comments in 1feb9865a: the page now says the cleanup is best-effort and needs adb at that moment, and tells the user to reconnect the device before the manual command. `npx docusaurus build` passes with this page.

PR #1178, opened 2026-09-16
docs: enable Algolia DocSearch with the Argent index

kacperkapusciak:
image

## Summary

- Replace the placeholder Algolia credentials in `docusaurus.config.js` with the real DocSearch application . The search API key is public and read-only, so it ships in the bundle as Docusaurus recommends.
- Remove the CSS in `overrides.css` that hid the navbar search bar until credentials existed, together with the rule that pushed the color mode toggle right to fill the gap.

No docs content update is needed, this changes site config only.

## Test plan

- [x] `npx docusaurus build` succeeds and the built pages contain the DocSearch button
- [x] `npm run format` from the repo root reports no changes
- [ ] Search on the deployed preview returns results from the `argent` index


filip131311 (approved):
🚢 🇮🇹

PR #1179, opened 2026-09-16
docs: emit trailing-slash URLs so canonical tags match what GitHub Pages serves

kacperkapusciak:
## Problem

Google Search Console reports every docs page as "Alternate page with proper canonical tag" and the Algolia crawler lists the same pages under "Monitoring: Canonical URL". Neither indexes them.

GitHub Pages serves each page as `<route>/index.html` and answers `<route>` with a 301 to `<route>/`. Docusaurus was emitting canonical, hreflang, og:url and sitemap URLs without the slash, so a page at `/docs/features/flows/` declared `/docs/features/flows` as canonical. That URL redirects back to the slashed page, so Google never settles on an indexable URL and Algolia skips the page because its canonical differs from the crawled URL.

## Fix

- `trailingSlash: true` in the Docusaurus config. Canonical, hreflang, og:url, sitemap and internal link URLs now end with `/` and match what GitHub Pages serves. The OG image plugin already strips slashes when naming cards, so card file names are unchanged.
- Drop the `noindex` search page from the sitemap.

No docs content changes.

## After deploy

- Search Console: "Validate fix" on the canonical issue.
- Algolia dashboard: trigger a recrawl.


filip131311 (approved):
it keeps the people going

